PR #6, opened 2026-07-16
feat: real console backend — parsers, harness, ports/REST, GraphQL reads, metering, content

senamakel:
## Summary

Makes every operator-console surface real: swaps each seam from *client-only sample* to a live backend, and replaces the out-of-process OpenHuman JSON-RPC seam with **openhuman embedded as a library**. Implements the full plan in [`docs/plans/`](docs/plans/README.md) — server-side workstreams WS1–WS6, WS8, plus the GraphQL read plane (WS2), metering (WS5), and the docs updates (WS9).

Bundled as one PR at the maintainer's request; internally it is 35 commit-sized, conventional-prefixed slices.

## What's included

| WS | Delivered |
|---|---|
| **WS1** | Company parsers — workflow TOML, `SKILL.md` frontmatter, workspace walker + wikilinks, shared skills registry; content-validation walk over `companies/*` + `skills/*`. |
| **WS4** | Embedded openhuman harness under the new `openhuman` feature — `HarnessPool`, hosted/mock provider, `OcMemory` adapter, approval bridge, cost hook, `AgentReply` event. Default build stays offline/echo-brained. |
| **WS6** | OAuth / DNS / SMTP connection surfaces + inbox ingest; secrets confined to `SecretStore` (status-only responses); feature-gated networked impls (`dns`/`smtp`/`oauth`). |
| **WS3** | Six console-surface ports (Task / Workspace / Fact / Usage / SkillState / Inbox) × fs / sqlite / mongodb + conformance suite; `src/server/ops/` REST write plane (dual-scoped); idempotent workspace/skills seeding. |
| **WS8** | ~12-skill shared library + starter skills/workflow/workspace for all 18 previously-bare company templates. |
| **WS5** | Metering — pure `bucket_usage` / `finances_from` projections + 90-day usage retention/eviction across all backends. |
| **WS2** | GraphQL read plane — `graphql.rs` → module tree, `Company` aggregation root (async handle), schema built once at startup, `GqlAuth` context (shared `resolve_claims`), SDL snapshot test. |
| **WS9** | `docs/spec/*`, `docs/modules/*`, and `frontend/ARCHITECTURE.md` updated to match what shipped. |

## Known partial (tracked upstream)

[...]

PR #7, opened 2026-07-16
fix: address #6 review findings (P1 credential-route authz + serve-path resolution)

senamakel:
## Summary

Follow-up to #6. The Codex review on #6 raised 1 P1 + 5 P2 findings; the fixes for them were pushed to the branch **after** #6 was merged (merge captured `b086a12`, the CI fix, not the review fixes), so they did not land in `main`. This PR carries exactly those three fix commits.

**`main` currently contains a P1 cross-tenant authorization bug** (the WS6 credential routes still use bare `OperatorAuth`) — this PR closes it.

## Commits

- `fix(ops): platform-scope credential routes` — **P1.** WS6 by-id `domain`/`smtp`/`connections` routes were guarded only by `OperatorAuth` + `resolve(&state, &id)`, so in `platform_auth` mode a tenant token could act on another tenant's company by choosing its id. Migrated them to the WS3 `scoped()` + `ScopedCompany` extractor (`PlatformOrOperatorAuth` + `authorize_address`; single-company alias uses `OperatorAuth` + `registry.sole()`). Added `credential_route_rejects_foreign_tenant` (foreign tenant → 403). HMAC inbox ingest left as-is (per-company signed secret already gates cross-tenant use).
- `fix(ops): allow moving workspace nodes to root` — **P2.** `WorkspaceStore::rename_move` parent widened to `Option<Option<&str>>` (`None` = unchanged, `Some(None)` = detach to root, `Some(Some(id))` = reparent) across trait + fs/sqlite/mongodb; PATCH `parentId` uses a `double_option` deserializer so omitted vs explicit `null` are distinct. Conformance + HTTP tests added.
- `fix(runtime,graphql): resolve skills/workflows from company source dir` — **P2 ×4.** `register_company` now calls `with_seed_dir` (serve-path workspace seeding); a new `CompanyRuntime::source_dir() -> Option<&Path>` seam makes `Company.skills`, `Company.workflow`/`workflows`, and `skillRegistry` read the real `companies/<name>/{skills,workflows}` + repo-level `skills/`. Platform-provisioned companies (no source dir) degrade to manifest-derived/empty — no behavior change there.

## Verification (all green)

```
cargo fmt --all -- --check            # clean
cargo clippy --all-targets -- -D warnings   # clean
cargo test                            # 359 passed
cargo test --features sqlite          # 375 passed
cargo check --features mongodb        # ok
cargo check --features openhuman      # ok
cargo test --lib content              # 5/5
```

### DIFF
--- a/src/app/types.rs
+++ b/src/app/types.rs
@@ -126,6 +126,10 @@ pub struct AppState {
     /// selected (`OPENCOMPANY_STORAGE`). Provisioning injects these into each
     /// new company's builder; `None` means fs defaults.
     stores: Option<crate::store::StorageHandles>,
+    /// The repo-level shared skill library directory (`skills/`), set on the
+    /// serve path. `None` in platform-provisioned mode (no repo checkout), where
+    /// the `skillRegistry` query degrades to empty.
+    skills_root: Option<std::path::PathBuf>,
     /// Cache of the repo-level shared skill registry (`skills/*/SKILL.md`).
     /// Populated on first read via [`AppState::skill_registry`]; never
     /// invalidated because the repo's skill library is immutable at runtime.
@@ -163,6 +167,7 @@ impl AppState {
             home: std::path::PathBuf::from("."),
             ownership: Arc::new(RwLock::new(HashMap::new())),
             stores: None,
+            skills_root: None,
             skill_registry: Arc::new(OnceLock::new()),
             schema: crate::server::graphql::build_schema(),
             connections: crate::server::ops::ConnectionsRuntime::new(),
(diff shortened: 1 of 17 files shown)


senamakel:
Good catch — fixed in `44e1bd9`. `--company` accepts either the company directory or the manifest file inside it; the source-dir wiring recorded the file path directly, so seeding and the skill/workflow/registry resolvers looked under `company.toml/…`. Added a `company_source_dir()` helper that normalizes a manifest-file argument to its parent company directory, applied at both the `with_seed_dir` call and the `skillRegistry` root derivation (which was off by one level for the file form), plus a unit test and a `register_company` test that passes the file path and asserts the workspace still seeds. Gate green.

PR #43, opened 2026-07-22
fix(store): write each JSONL record as one atomic append

YellowSnnowmann:
## Summary

Hardens `append_line` (`src/store/fs.rs`) against a JSONL corruption that surfaces as a `serde_json` **"trailing characters"** parse error — the failure seen intermittently in `store::fs::test::conformance_export_totality` (which flaked #39's CI, unrelated to that PR's code).

**Root cause:** `append_line` wrote the record body and its `\n` as **two separate `write_all` calls**. The per-path locks that serialize appends (`PathLocks`) are **per store instance**, not process-global, so any concurrent append not covered by one instance's lock can interleave the two writes as `{a}{b}\n\n` — two JSON objects on one physical line — which `read_jsonl` then fails to parse.

**Fix:** build `line + "\n"` in one buffer and write it with a single `write_all`, so the whole record is one atomic `O_APPEND` write.

**Proof:** a new test appends 64 records concurrently and asserts each round-trips as its own line. It fails **20/20** runs with the two-write form and passes deterministically with the single write.

### Honesty note
`conformance_export_totality` cleared on a plain CI re-run, so I could not tie the specific flake to this with 100% certainty. But the two-write interleave is a **confirmed, reproducible** atomicity bug (the 20/20 test), and single-atomic-append is correct regardless of the exact trigger. Framed as a proper fix, not a guess. A follow-up could make `PathLocks` process-global (shared across store instances) to close the cross-instance gap entirely.

## API Or Behavior Changes

None (internal helper; same on-disk format).

## Tests

- [x] `cargo test` — 484 passed (+1 new `concurrent_appends_stay_one_record_per_line`)

## Documentation

Doc comment on `append_line` explains the single-write invariant and the interleave it prevents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent appends to JSONL files, preventing records from being merged or corrupted.
  * Ensured each appended record remains on its own complete line, even when multiple writes occur simultaneously.

* **Tests**
  * Added coverage verifying that concurrent appends are preserved accurately and can be read back successfully.

### DIFF
--- a/src/store/fs.rs
+++ b/src/store/fs.rs
@@ -58,6 +58,12 @@ impl PathLocks {
 }
 
 /// Appends one line (a `\n` is added) to `path`, creating the file if absent.
+///
+/// The line and its terminating newline are written in a **single** `write_all`
+/// so the whole record lands as one `O_APPEND` write. Splitting it into two
+/// writes let concurrent appends to the same JSONL file interleave as
+/// `{a}{b}\n\n` — two JSON objects on one physical line — which surfaces later as
+/// a `serde_json` "trailing characters" parse error in [`read_jsonl`].
 pub(crate) async fn append_line(path: &Path, line: &str) -> Result<()> {
     let mut file = tokio::fs::OpenOptions::new()
         .create(true)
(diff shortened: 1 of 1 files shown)


YellowSnnowmann:
@coderabbitai review

Issue #36, opened 2026-07-22
Memory system integration with OpenHuman + per-instance workspace folder with Docker and disk configuration


Al629176:
## Summary

Two tightly coupled tasks:

1. **Memory integration** — wire OpenCompany's memory system to OpenHuman so each company instance shares, reads from, and writes to a persistent memory store that the OpenHuman orchestration layer can use for decision-making.
2. **Workspace + Docker** — configure a per-instance workspace folder for each OpenCompany tenant and provision Docker + disk space correctly so the workspace is usable (mounted, size-bounded, and persisted across restarts).

---

### Current state

OpenCompany has a file-based store (`src/store/Memory`, `src/store/Context`, `src/store/EventLog`) scoped per company instance. OpenHuman has its own Brain/Memory system (vector store, memory tree, Data Sync pipeline). These two memory layers are currently disconnected — a company agent running inside OpenCompany cannot read from or write to the OpenHuman memory graph, and vice versa.

### Goal

A company agent operating inside OpenCompany should have the same memory capabilities as an agent in OpenHuman: compounding knowledge base, semantic retrieval, entity extraction, and memory-informed decision-making. The OpenHuman memory layer is the canonical implementation — OpenCompany should integrate against it rather than duplicate it.

### Integration design

```
OpenCompany CompanyRuntime
        │
        ▼
Memory Adapter (new)
        │
        ├─── Local path (file-based): reads src/store/Memory directly
        │    (no OpenHuman process; lightweight; used when OH runtime not embedded)
        │
        └─── OpenHuman embedded runtime path (see #26):
             reads/writes OpenHuman Brain via internal IPC or shared store path
```

**Memory adapter interface:**

```rust
trait MemoryAdapter {
    fn store(&self, tenant_id: &str, chunk: MemoryChunk) -> Result<()>;
    fn retrieve(&self, tenant_id: &str, query: &str, top_k: usize) -> Result<Vec<MemoryChunk>>;
    fn list_sources(&self, tenant_id: &str) -> Result<Vec<MemorySource>>;
}
```

Implementations:
- `FileMemoryAdapter` — reads/writes `workspace/<tenant>/memory/` directly (standalone mode)
- `OpenHumanMemoryAdapter` — delegates to embedded OpenHuman Brain via shared SQLite or IPC socket (embedded mode)

### What gets stored in memory

Per company instance, memory should accumulate:
- Inbound emails (from inbox pod — #22) summarised and stored as memory chunks
[...]

PR #44, opened 2026-07-22
feat(harness): intrinsic memory tools on embedded agents (#36 Phase 2)

YellowSnnowmann:
## Summary

`build_agent` passed an **empty tool vector** to `AgentBuilder`, so embedded company agents could converse but not act (Phase 2 of the openhuman↔opencompany wiring). This gives every agent the **intrinsic memory tools** — `memory_store` and `memory_recall` — over its own `OcMemory`.

- An agent can now **deliberately** save and look up facts, on top of the **automatic** retrieve→inject→store loop (#39). Both go through the same `ContextStore` the loop and the `OPENCOMPANY_MEMORY` overlay (#37) use, so deliberate and automatic memory share one store.
- These are **intrinsic and tenant-isolated** (an agent's memory is its company's `ContextStore`), so they're granted to every agent rather than gated on the manifest `[tools]` allow-list.

### Scope / why memory-only first
- **External/integration tools** (`web.*`, `docs.*`, file, shell) are *not* wired here: the manifest grant vocabulary (`web.*`/`docs.*`) doesn't map to openhuman's tool names, and file/shell tools need a `SecurityPolicy` + `RuntimeAdapter` sandbox not yet in `HarnessDeps`. That grant→tool bridge + sandbox is a tracked follow-up; the builder extends the same vector.
- **`MemoryForgetTool` excluded**: `OcMemory`'s append-only `ContextStore` cannot delete, so a forget tool would silently no-op.

## API Or Behavior Changes

- Behavior: embedded agents now expose `memory_store`/`memory_recall` tools. Feature-gated to `openhuman`; the default (echo) build is unchanged.
- No public API added.

## Tests

- [x] `cargo clippy` — CI (default features) clean; no findings in opencompany `src`. (The vendored `tinyagents` lint only appears under `--features openhuman` clippy, which CI does not run.)
- [x] `cargo test --features openhuman` — **549 passed**, incl. a new test asserting the tool surface and all existing agent/chat tests.

## Documentation

Module docs on `build_agent`/`memory_tools` explain the intrinsic-vs-granted split and the excluded forget tool.

## Summary by CodeRabbit

* **New Features**
  * Embedded agents now have built-in memory capabilities, allowing them to store and recall information within their assigned context.
  * Memory tools are available by default, while external and integration tools remain controlled through configuration.

* **Documentation**
  * Updated agent configuration guidance to describe the default memory tools and external tool access controls.

* **Tests**
  * Added coverage confirming the built-in memory tools are available.

### DIFF
--- a/src/harness/build.rs
+++ b/src/harness/build.rs
@@ -2,13 +2,13 @@
 //!
 //! [`build_agent`] turns one roster entry into a ready-to-run openhuman
 //! [`Agent`], injecting the harness's provider, the [`OcMemory`] adapter, the
-//! [`ApprovalPolicy`] tool policy, and a workspace directory. It is deliberately
-//! minimal for v1:
+//! [`ApprovalPolicy`] tool policy, and a workspace directory.
 //!
-//! * **Tools** start empty. The manifest `tools ∩ agent.tools` intersection and
-//!   the port-backed tool surface are ported from the legacy
-//!   `src/openhuman/tools.rs` in a follow-up (WS4 subtask, tracked seam) — the
-//!   builder accepts the vector so wiring them in is a one-line change.
+//! * **Tools**: every agent gets the intrinsic [`memory_tools`] (`memory_store`
+//!   + `memory_recall`) over its own company memory. External/integration tools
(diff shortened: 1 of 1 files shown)


YellowSnnowmann:
@coderabbitai review

Issue #56, opened 2026-07-23
Use OpenRouter instead of DeepSeek for all inference in OpenCompany


Al629176:
## Summary

OpenCompany should use **OpenRouter** (`openrouter.ai`) as its inference provider rather than calling DeepSeek directly. Since we are actively building OpenCompany now, this is the right time to set OpenRouter as the default from the start — single API key, access to hundreds of models, OpenAI-compatible.

## Why OpenRouter instead of DeepSeek direct

- **One key, all models** — OpenRouter gives access to DeepSeek, Claude, Gemini, Llama, Qwen, and hundreds more through a single `OPENROUTER_API_KEY`. No need to manage multiple provider keys as we experiment with models.
- **OpenAI-compatible** — drop-in replacement, no SDK changes needed
- **Flexibility** — per-company model selection via `company.toml` without adding new provider integrations
- **Staging-safe** — easier to swap models per-environment without credential sprawl

### `company.toml` default inference config

```toml
[inference]
provider     = "openrouter"
base_url     = "https://openrouter.ai/api/v1"
api_key_secret = "OPENROUTER_API_KEY"
model        = "deepseek/deepseek-chat"        # default; overridable per company
```

Each company template can override the model while sharing the same provider and key.

### OpenRouter headers

All requests to OpenRouter must include:
```
HTTP-Referer: https://opencompany.work
X-Title: OpenCompany
```

Inject these automatically in the inference client when provider is `openrouter`.

### Environment variable

```
OPENROUTER_API_KEY=sk-or-...
```

No `DEEPSEEK_API_KEY` or other per-provider keys needed.

### Model selection per company template

Update the 19 company templates to specify appropriate models via `company.toml`:

| Company type | Suggested model |
|---|---|
| Agentic marketing, content | `deepseek/deepseek-chat` |
| Coding / technical | `deepseek/deepseek-r1` |
| Research / analysis | `deepseek/deepseek-r1` |
| Fast / high-volume | `meta-llama/llama-3.1-8b-instruct` |
| Vision tasks | `qwen/qwen-vl-plus` |

All routed through the same OpenRouter endpoint — model is just a string in `company.toml`.

### Docker / deployment

Update `docker-compose.yml` and `deploy/` manifests:
- Remove any `DEEPSEEK_API_KEY` references
- Add `OPENROUTER_API_KEY` to required env vars
- Update `.env.example`

## Acceptance criteria

[...]

PR #60, opened 2026-07-23
feat(harness): refresh roster on skill-delta change so console skills reach the agent next turn (#41)

### LINKED ISSUE #41: skills: console-authored custom skills don't reach the embedded agent (only committed company-dir bundles surface)
Follow-up to #28 (epic #26). The console Skills tab (#34) correctly shows a company's effective skills, but the **embedded agent only sees committed company-dir `SKILL.md` bundles** — custom skills authored through the console never reach its effective set, even after a fresh process restart.

## Repro (verified 2026-07-22, main @ 8a32dcf)
1. Company `openhuman_demo` with a committed `skills/inbox-triage/SKILL.md`.
2. Via the console API: install a registry skill (`weekly-report`) and create a **custom** skill with a body (`standup-digest`, POST /skills → source=custom, custom_doc stored + parseable — GET /skills shows all three).
3. Restart the backend (fresh `HarnessPool`).
4. Dispatch a task to `engineer`: "List every skill you can use, by name."

**Expected:** agent lists `inbox-triage` + `standup-digest` (both have real bodies). Registry `weekly-report` legitimately has no local content.
**Actual:** agent lists only `inbox-triage`.

## Analysis
- `HarnessPool::ensure` (src/harness/mod.rs:178) fetches deltas via `deps.skills.list()` and passes them to `build_roster` → `build_agent` → `EffectiveSkills::materialize(source_dir, deltas)`.
- `materialize` (src/harness/skills.rs:90-105) is *written* to fold an enabled delta carrying `custom_doc` into the effective docs.
- Despite that, a custom-doc delta does not surface at runtime. Needs root-cause: is `deps.skills` wired to the same store the console writes (`runtime.skills()`), does `materialize` actually insert the custom doc, and does `catalogue()`/the read-tool tree include it?

## Also
Registry installs (`source=registry`, no `custom_doc`) have no local skill content, so they can never surface to the agent. Either fetch/store content on install, or make the console mark them as catalog-only until content exists.

Part of #26.

### Triage Assessment
This issue is **valid and actionable**. A partial fix in commit `e79d279` (referencing this issue) correctly added skill-delta freshness tracking via `CompanyRoster`, but the subsequent MCP-freshness commit `0f5829d` replaced that mechanism with MCP-only fingerprint checking, **regressing the skill-delta freshness**. The `EffectiveSkills::materialize` function in `src/harness/skills.rs` correctly handles `custom_doc` (verified by unit tests), and the console API handlers in `src/server/ops/skills.rs` correctly persist `custom_doc` for both registry installs and custom skills. The root cause is the `HarnessPool::ensure` fast-path in `src/harness/mod.rs:354-361` which only checks MCP fingerprints and returns early when MCP is unchanged, skipping the skill-delta re-fetch and roster rebuild.

**Impact**: medium-high. An operator who authors/enables/disables a skill in the console during a running session will not see that change reflected in agent behavior until the next backend restart. The "fresh restart" case from the original repro now works (materialize handles custom_doc correctly), but the live-refresh case is broken.

### Proposed Scope
- **In-scope**: Add skill-delta fingerprinting to `HarnessPool` so that `ensure()` rebuilds the roster when operator skill deltas change, even when MCP servers are unchanged.
- **Explicit non-goals**:
  - Skill execution (`RunWorkflowTool`): out of scope per the existing skills module docs.
  - Registry installs without `custom_doc`: this was addressed by the `install` handler in `src/server/ops/skills.rs:219-242` which generates a `SKILL.md` from registry metadata; issue #42 was closed as the separate tracking item.

### Implementation Steps
1. **Add skill-delta fingerprint field to `HarnessPool`** (`src/harness/mod.rs`)
   - Add a field `skill_fingerprints: RwLock<HashMap<CompanyId, u64>>` to the `HarnessPool` struct, alongside the existing `mcp_fingerprints`.
   - Add a private `skill_delta_fingerprint(deltas: &[SkillState]) -> u64` function, analogous to the existing `mcp_fingerprint()`. Hash the number of deltas and each delta's slug + enabled + source + custom_doc (or a hash of custom_doc). Sorting deltas by slug before hashing ensures a store-order-agnostic fingerprint.

2. **Update `HarnessPool::ensure()` fast-path** (`src/harness/mod.rs:349-385`)
   - Before entering the MCP fast-path check, fetch the current skill deltas (or a hash of them) and compute their fingerprint.
   - Add the skill-delta fingerprint to the fast-path condition: a cached roster is current only when both the MCP fingerprint AND the skill-delta fingerprint match.
   - When either fingerprint changed (or the roster is absent), proceed to rebuild the roster using the freshly fetched deltas and store both fingerprints.

3. **Store fingerprints on rebuild** (`src/harness/mod.rs:378-383`)
   - On the rebuild path, insert both the MCP fingerprint and the skill-delta fingerprint alongside the new roster.

4. **Add skill fixture helper for tests** (analogous to the existing `MockSkillStore` from the `e79d279` diff).

5. **Add tests** (in `src/harness/mod.rs` test module):
   - `late_authored_skill_reaches_agent_on_next_ensure`: author a custom skill after the first roster build; assert the second `ensure` rebuilds the roster and the skill body surfaces through `describe_workflow`.
   - `unchanged_deltas_reuse_the_same_roster`: assert `Arc::ptr_eq` identity across two `ensure` calls with no delta change.
   - `disabling_a_skill_drops_it_on_next_ensure`: disable a skill and assert the roster rebuilds and the materialized catalogue no longer contains it.
   - `ensure_rebuilds_on_skill_change_when_mcp_is_unchanged`: combine both MCP and skill-delta changes to prove the fast-path handles both independently.

### Verification
- `cargo test --features openhuman` — the new freshness tests in `harness::tests` and the existing `skills::tests::console_custom_docs_surface_content_through_read_tools` (which already proves `materialize` handles `custom_doc`).
- `cargo clippy --all-targets --features openhuman -- -D warnings`.
- `cargo fmt --all -- --check`.
- Manual: the repro script from the issue should now show the custom skill on the second `ensure` call with no restart.

### Risks and Open Questions
- **Concurrent access**: The existing `rosters` read-write-lock pattern from `e79d279` (double-check pattern inside the write lock) should be restored to prevent racing `ensure` calls from both rebuilding.
- **Skill fingerprint vs full delta comparison**: A hash-based fingerprint is cheaper than storing and comparing the full `Vec<SkillState>`. Risk of false negatives from hash collisions is negligible for this use case.
- **MCP-fingerprint interaction**: The fix must ensure both fingerprints are checked independently: a MCP change triggers a rebuild (which fetches current skill deltas), and a skill-delta change triggers a rebuild (which re-resolves MCP servers).

### Acceptance Criteria
- A skill authored in the console after the first `ensure()` call surfaces to every agent on the **next** `ensure()` call, with its full body content reachable through the agent's read tools, without a process restart.
- An unchanged delta set across two `ensure()` calls returns the cached roster identity (no rebuild).
- A disabled skill is dropped from the materialized catalogue on the next `ensure()` call.
- MCP-only and skill-only changes both independently trigger a rebuild.

oxoxDev:
## Summary

`HarnessPool::ensure` cached a company roster on first turn and never re-read skill deltas, so a skill authored in the console after the first turn was invisible until restart. Fix: the cache value becomes `CompanyRoster { agents, skill_deltas }`; `ensure()` re-lists deltas every call, fast-paths on equality, rebuilds under a double-checked write lock when they differ. Since `brain.rs` calls `ensure()` at the top of every cycle, a console skill edit reaches every agent on the NEXT turn — no restart.

`Closes #41`.

> Note: Touches the same `HarnessPool::run` cost-slug span as #56 (BYOK) — expect a ~2-line conflict; whichever merges second rebases (keep this PR's `CompanyRoster` body + #56's live-slug binding).

## API Or Behavior Changes

No HTTP API change. Behavior change: adding, enabling, or disabling a skill in the console now takes effect on the next agent turn instead of requiring a restart.

## Tests

- [x] `cargo test` (518 passed)

`cargo test --features mcp` = 608 passed, including 4 new freshness tests (late-authored skill, unchanged deltas, disabled skill).

Same pre-existing base caveat as the other cells: `cargo check --features openhuman` alone is broken on base (`server/ops/mcp.rs:388`), not caused by this change; validated via `--features mcp`.

## Documentation

- `src/harness/skills.rs` gained a "## Freshness" module note.
- Fixed the stale "Idempotent … no-op" doc on `ensure`.

### DIFF
--- a/src/harness/mod.rs
+++ b/src/harness/mod.rs
@@ -156,9 +156,25 @@ impl CompanyAgent {
     }
 }
 
+/// A company's cached roster together with the skill deltas it was built from.
+///
+/// Caching the deltas is what lets [`HarnessPool::ensure`] tell whether the
+/// operator's effective skill set has moved (a skill authored, enabled, or
+/// disabled in the console *after* the roster was first built) and rebuild only
+/// then — so a fresh skill reaches every agent on the next cycle without a
+/// process restart, and an unchanged set leaves the live agents (and their
+/// conversation state) untouched.
+struct CompanyRoster {
+    /// The live agents, one per manifest `[[agent]]`.
+    agents: Vec<Arc<CompanyAgent>>,
+    /// The skill deltas the agents were built from, sorted by slug so the
+    /// freshness comparison is store-order-agnostic.
+    skill_deltas: Vec<SkillState>,
+}
+
 /// A pool of live agents, one roster per company.
 pub struct HarnessPool {
-    agents: RwLock<HashMap<CompanyId, Vec<Arc<CompanyAgent>>>>,
+    rosters: RwLock<HashMap<CompanyId, CompanyRoster>>,
 }
 
 impl Default for HarnessPool {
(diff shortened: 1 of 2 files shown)


senamakel (approved):
Automated technical review: approved for the inspected head commit.

Issue #113, opened 2026-07-24
Epic: Workflows -> tinyflows parity (P1-P4)


oxoxDev:
Part of #26.

## Goal
Bring OpenCompany's workflow subsystem to feature parity with OpenHuman's **tinyflows** engine. Phased — each phase is a shippable PR-sized unit, additive (existing workflows keep working).

## Parity gaps (today OC has 6 node kinds, manual trigger only, one real capability)
tinyflows has 12 node kinds, per-node config + expressions, 9 trigger kinds, full capability traits, per-node error/retry, durable runs with resume/cancel, observability, versioning.

## Phases
- [ ] **P1 — Model config + real capabilities** (backend). Node-level `config` passthrough, `on_error`/`retry`/`requires_approval` fields; wire `ToolInvoker` (harness toolbelt), `HttpClient` (behind the web allowlist), `StateStore` (company KV). `code` stays unwired (no sandbox).
- [ ] **P2 — Full node catalog + expressions + sub-workflows** (backend + minor UI). `switch/merge/split_out/transform/output_parser/sub_workflow` kinds; `=expr` bindings; `WorkflowResolver` (workflow-calls-workflow with cycle guard).
- [ ] **P3 — Durable runs, resume/approvals, cancel, schedule + webhook triggers** (backend + UI). New run store port (Fs+Mongo), observer-fed; durable resume via checkpointer; cancel via run registry; schedule + webhook triggers wired to the existing scheduler/webhook route.
- [ ] **P4 — Authoring parity + polish** (UI-heavy). PUT/DELETE + enable/disable; validate endpoint (multi-error); node-contracts for config forms; persisted canvas positions; orchestrator `build_workflow` agent tool.

Strictly sequential P1->P4. Overall scope: large (4-6 PRs). Explicitly out: n8n import, drafts, langfuse export.

PR #124, opened 2026-07-25
fix: restore openhuman build (duplicate field + missing SteerAction import)

oxoxDev:
## Summary

`main` currently **does not compile under the `openhuman` feature** — two breaks slipped in via the #119/#120 squash-merges. CI stayed green because the default CI build doesn't enable `openhuman` (the known full-feature CI gap), so nothing compiled the affected paths. This fixes both so the real app + any `openhuman` build works again.

1. **Duplicate struct field** — `src/runtime/builder.rs` set `workflow_source_dir: self.seed_dir.clone()` **twice** in the same `HarnessDeps { … }` initializer (identical field + comment, a union-merge artifact). That's a hard `E0062` "field specified more than once" error under any build that compiles the production runtime builder. Removed the duplicate.
2. **Missing import** — `src/harness/mod.rs:748` uses `SteerAction::Cancel` in the run loop, but the module imported only `SteerControl` (the sole `SteerAction` import lived inside a `#[cfg(test)]` fn). Added `SteerAction` to the module-level `use`.

## API Or Behavior Changes

None — build-only fix. No functional change.

## Tests

- [x] `cargo build --lib --features openhuman,mcp,telegram` — **compiles clean** (was E0062 + unresolved `SteerAction` before). Only the pre-existing vendored-openhuman warnings remain.

## Documentation

None needed. (Root cause is the documented full-feature CI gap — CI never builds the `openhuman` combo, so these slipped through.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling so canceled turns do not save unnecessary memory or context.

* **New Features**
  * Media-generation capabilities are now available when explicitly configured through the runtime.
  * Media generation is unavailable by default when no backend is configured, preventing unintended access.

### DIFF
--- a/src/harness/mod.rs
+++ b/src/harness/mod.rs
@@ -66,7 +66,7 @@ use oh::inference::provider::Provider;
 use crate::company::Agent as ManifestAgent;
 use crate::company::Policy;
 use crate::company::mcp::McpServerDecl;
-use crate::company::steer::SteerControl;
+use crate::company::steer::{SteerAction, SteerControl};
 use crate::error::OpenCompanyError;
 use crate::harness::cost::{TurnUsage, record_turn_cost};
 use crate::harness::mcp_probe::McpFailureQueue;
--- a/src/runtime/builder.rs
+++ b/src/runtime/builder.rs
@@ -881,11 +881,6 @@ impl RuntimeBuilder {
                                 // closed — `build_agent` wires no media tools even
                                 // for a company that grants `media`.
                                 media: self.media_backend.clone(),
-                                // #113 P2: the company source dir so a workflow's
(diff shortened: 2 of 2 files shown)


oxoxDev:
Closing as obsolete — both fixes here are already on `main` (@ 1be90e9): the `SteerAction` import is present at `src/harness/mod.rs:71`, and `workflow_source_dir` appears exactly once in `src/runtime/builder.rs` (the legitimate #113 field for `sub_workflow` resolution — no duplicate). The union-build breakage this patched was a consolidated-integration-branch artifact, not main; applying it now would delete that legitimate field and break sub_workflow resolution (also why it shows as conflicting). No action needed.

PR #135, opened 2026-07-27
feat(connections): P2 — toast display name, 2nd-provider alignment, best-effort revoke

oxoxDev:
## Summary

Follow-on to PR #129 (Connections P1). Three tight fixes so the console connections flow is complete beyond GitHub.

**Fix 1 — connect toast shows the display name.** The OAuth bounce-back `useEffect` (`app-shell.tsx`) toasted the raw `?connected={provider}` id, so it read "Connected slack." It now maps the id through the `CONNECTION_PROVIDERS` catalog → "Connected Slack." (falls back to the raw id when unknown). The disconnect toast already used the display name.

**Fix 2 — catalog-id ↔ backend-key alignment (2nd provider connects end-to-end).** For a provider to connect, the manifest `[[connection]] provider`, the console catalog `id`, and the backend `well_known` key must be the *same* string. Slack now works end-to-end because its catalog `id === "slack"` matches the backend key (GitHub/Gmail already aligned). The invariant is documented on `ConnectionProvider.id`, and the `google-calendar` / `google-drive` tiles — whose ids map to no backend key — are flagged as dead in a code comment. No new backend providers were invented (comment-only for the google-* mismatch).

**Fix 3 — best-effort provider-side revoke on disconnect.** `do_disconnect` previously only blanked the local `oauth/{provider}` secret (SecretStore has no delete). It now attempts a real provider-side revoke *first*: GitHub → authenticated `DELETE /applications/{client_id}/grant`, Slack → `POST auth.revoke`. Any failure (unknown provider, unconfigured creds, network error, non-2xx) is logged with **provider + status only — never the token** — and swallowed, so the local secret-blank always runs. A new optional `OPENCOMPANY_OAUTH_<P>_REVOKE_URL` override mirrors the existing `_AUTHORIZE_URL` / `_TOKEN_URL` pattern.

### Host configuration (for a 2nd provider, e.g. Slack — nothing committed)
Set on the host: `OPENCOMPANY_OAUTH_SLACK_ID`, `OPENCOMPANY_OAUTH_SLACK_SECRET`, optional `OPENCOMPANY_OAUTH_SLACK_SCOPES` / `_REVOKE_URL`; the company `company.toml` must declare `[[connection]] provider = "slack"`. Same `OPENCOMPANY_OAUTH_REDIRECT_BASE` / `OPENCOMPANY_CONSOLE_URL` as P1.

## API Or Behavior Changes

- Disconnect now makes a best-effort provider-side token revocation before blanking the local secret. Behavior is unchanged when the provider has no known revoke endpoint or the call fails (secret still blanked).
- New optional env override `OPENCOMPANY_OAUTH_<P>_REVOKE_URL`.
- Console: connect toast shows the provider display name; a 2nd provider (Slack) connects end-to-end.
- No schema, persistence, or route change. No token material appears in any response or log (asserted by test).

## Tests

- [x] `cargo test` — **708 passed / 1 ignored / 0 failed**, incl. 3 new revoke tests: revoke invoked against a mock server (asserts the token is carried to the provider) + secret blanked; secret still blanked when the provider is unconfigured; secret still blanked when the revoke endpoint is unreachable
- [x] `cd frontend && npm run typecheck`
- [x] `cd frontend && npm run build`

## Documentation

Behavior carried by inline route/module docs. The `ConnectionProvider.id` end-to-end invariant and the dead-tile flags are documented inline; the host env recipe for a 2nd provider is in the Summary above.

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -52,6 +52,7 @@ import { useHashView } from "@/hooks/use-hash-view";
 import { toast } from "sonner";
 
 import { type ChatMessage, fromHistory, makeMessage } from "@/lib/chat";
+import { CONNECTION_PROVIDERS } from "@/lib/connections";
 import { DISCORD_INVITE_URL } from "@/lib/links";
 import { defaultThreads, threadsFromDesks } from "@/lib/threads";
 import { Overview } from "@/views/Overview";
@@ -236,7 +237,11 @@ export function AppShell({
       window.location.pathname + (query ? `?${query}` : "") + window.location.hash,
     );
     setView("connections");
-    toast.success(`Connected ${connected}.`);
+    // The callback param carries the raw provider id (e.g. "slack"); show the
+    // catalog display name ("Slack") when we know it, falling back to the id.
+    const providerName =
+      CONNECTION_PROVIDERS.find((p) => p.id === connected)?.name ?? connected;
+    toast.success(`Connected ${providerName}.`);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
--- a/frontend/src/lib/connections.ts
+++ b/frontend/src/lib/connections.ts
@@ -11,6 +11,16 @@ export type ConnectionCategory =
   | "Storage";
 
 export interface ConnectionProvider {
+  /**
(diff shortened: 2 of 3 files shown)


oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai review

Issue #105, opened 2026-07-24
Agent chat leaks raw <tool_call> markup (tinyagents 1.9 stopgap)


oxoxDev:
## Problem
DeepSeek-via-OpenRouter (and similar native models on some OpenAI-compatible routes) emit tool calls as **text** — `<tool_call id="...">{...}</tool_call>` — instead of the structured field. OpenCompany's pinned tinyagents (**v1.9.0**) has no tool-call recovery seam, so the raw markup leaks straight into agent chat replies.

## Why it's not a simple bump
The complete fix landed on **tinyagents main (v2.1.0)** in tinyhumansai/tinyagents#76 (terminal + streaming). OpenCompany cannot consume it: openhuman requires `tinyagents ^1.7`, and main is a breaking **2.x** major (see the 2.x upgrade issue).

## Stopgap (this issue)
Port the complete fix onto OpenCompany's **1.9** tinyagents line (proven compatible — the consolidated test build already ran a 1.9-patched tinyagents), then bump `vendor/tinyagents`:
- attribute/variant-tolerant `<tool_call ...>` matcher (Hermes/DeepSeek forms)
- native-mode fallback recovery (`should_recover`)
- streaming `ToolCallStreamScrubber` so live deltas don't leak either

## Scope
small–medium. Superseded by the tinyagents 2.x upgrade.

Related: tinyhumansai/tinyagents#76

PR #139, opened 2026-07-27
chore(deps): move onto tinyagents 2.x + migrate provider to ChatModel (#106)

### LINKED ISSUE #106: Upgrade tinyagents 1.9 -> 2.x (openhuman + tinyagents as a matched pair)
## Goal
Move OpenCompany onto **tinyagents 2.x** so it consumes the upstream fixes (incl. the `<tool_call>` leak fix, tinyagents#76) without a private downstream patch.

## Blocker
Both OpenCompany's top-level `vendor/tinyagents` and the vendored `vendor/openhuman/vendor/tinyagents` pin **v1.9.0** (`19dc2c4`). openhuman requires `^1.7`; main is **2.1.0** (breaking major). The two histories are **not linear** — this is a real migration, not a fast-forward.

## Spike result (openhuman compiled against tinyagents 2.1.0 = 31 errors)
- **~26 errors = embeddings API rewrite.** tinyagents 2.x restructured `harness::embeddings` — renamed every concrete model (`Voyage`/`Cohere`/`Ollama`/`Cloud`/`Noop`EmbeddingModel), changed the `TaEmbeddingModel` trait, removed constants + rate-limit/backoff helpers. openhuman's embeddings layer must be rewritten. **This is the bulk of the work and is unrelated to the leak.**
- **4 errors = `HarnessToolCall.invalid` field** removed in 2.x. Trivial adapter fix.

## Sequence
1. **openhuman repo**: migrate its tinyagents dependency to 2.x (embeddings rewrite + HarnessToolCall). Maintainer-owned.
2. **OpenCompany**: bump `vendor/openhuman` + `vendor/tinyagents` together to the 2.x-compatible pair.

Depends on upstream openhuman migration. Supersedes the 1.9 leak stopgap.

oxoxDev:
## Summary

Moves OpenCompany onto **tinyagents 2.x** so it consumes upstream fixes (including the `<tool_call>` leak fix, tinyagents#76) without a private downstream constraint.

The presumed blocker — openhuman's embeddings/provider rewrite being maintainer-owned — is **cleared**: openhuman `main` already migrated to tinyagents 2.1. But the rewrite reaches OpenCompany after all: openhuman **deleted its `Provider` trait** and rebuilt inference on tinyagents `ChatModel<()>` (`create_chat_model` now returns `Arc<dyn ChatModel<()>>`). So this is a coordinated submodule bump **plus** a provider-adapter migration, not a version-only bump.

Three commits:
1. **deps** — bump `vendor/openhuman` → main (carries the 2.1 migration) and `vendor/tinyagents` → the exact commit openhuman's own nested submodule pins (matched-pair invariant → one tinyagents copy); `openhuman` feature now enables `tiny` so the crate is in scope.
2. **provider migration** — the five harness providers move from `Provider` to `ChatModel<()>`.
3. **shim import fix** — the #105 `<tool_call>` shim's imports follow openhuman's relocated message types.

Closes #106.

## API Or Behavior Changes

- **`--features openhuman` only; default build byte-for-byte unaffected** (both deps stay optional).
- `HarnessDeps.provider` type changes from `Arc<dyn Provider>` to `Arc<dyn HarnessModel>`, where `HarnessModel: ChatModel<()>` is a thin supertrait re-adding `telemetry_provider_id` — the one method `ChatModel` drops, kept because per-turn BYOK **cost attribution** reads it live.
- **All provider behavior is preserved**: tier→model mapping, OpenRouter attribution headers, keyless-Ollama, bearer scrubbing in every error string, and `TenantProvider`'s per-turn BYOK re-resolution (a console provider switch still takes effect on the next turn with no rebuild).
- **Billing preserved.** openhuman's `pub(crate)` `merge_openhuman_usage_meta` is unreachable, so its shape is re-expressed locally: the response's `openhuman.billing.charged_amount_usd` is re-projected onto `ModelResponse.raw` under the identical `openhuman_usage_meta` key + fields (`charged_amount_usd`, `context_window`) + both-zero no-op rule that openhuman's own `usage_info_from_response` reader consumes — so the host cost layer still sees the charged USD and it flows to `last_turn_usage().cost_usd`. Cached-input tokens ride `Usage.cache_read_tokens` via the same `openhuman.usage.cached_input_tokens` precedence as before.
- The agent is now wired via openhuman's native `AgentBuilder::chat_model(Arc<dyn ChatModel<()>>)` seam (was `.provider_arc`).
- **The #105 `AttrTolerantXmlDispatcher` shim is retained, deliberately.** tinyagents#76's fix lives in tinyagents' own parser, which OpenCompany does not use — it dispatches through openhuman's `XmlToolDispatcher`, which still matches only bare-literal `<tool_call>` open tags (attribute tolerance exists only for `<invoke>`). Removing the shim would reopen the `<tool_call id="…">` leak against the staging model. Only its imports moved (`ChatMessage`/`ConversationMessage` → `agent::messages`; `ChatResponse` stays at `inference::provider`).

## Tests

`--features openhuman`-gated code, so gates run with the feature on; clippy scoped to the `opencompany` package because the vendored `tinyagents` crate trips a **pre-existing** `large_enum_variant` lint under `-D warnings` (its own `AgentStreamItem`/`Running`/`Draining` enum — unrelated to this diff, never CI-built in this repo):

- [x] `cargo build` (default features) — clean, unaffected
- [x] `cargo build --features openhuman,mcp,telegram` — clean (`--features openhuman` alone won't compile: pre-existing mcp-gated-symbol issue; the combo is the real target)
- [x] `cargo test --features openhuman,mcp,telegram --lib harness::` — **201 passed, 0 failed** (provider filter: 25 passed, 0 failed)
- [x] `cargo clippy -p opencompany --all-targets --features openhuman,mcp,telegram` — no warnings in any changed file
- [ ] N/A: `cargo clippy --all-targets -- -D warnings` — this code is `openhuman`-gated; see the tinyagents pre-existing-lint note above.

The provider tests keep the billing assertion alive: a `{ "openhuman": { "billing": { "charged_amount_usd": 0.0123 } } }` envelope must survive onto `raw.openhuman_usage_meta.charged_amount_usd`, and cached-input tokens onto `Usage.cache_read_tokens`.

## Documentation

No external doc surface changed. The rationale lives inline: the provider module docs explain the `HarnessModel` supertrait and the local re-projection of `openhuman_usage_meta` (with a pointer to openhuman's `pub(crate)` original), and the tool-dispatcher module docs already state the shim is retained because openhuman's parser is still bare-literal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5405,7 +5405,7 @@ dependencies = [
  "tar",
  "tempfile",
  "thiserror 2.0.19",
- "tinyagents 1.9.0",
+ "tinyagents 2.1.0",
  "tinyflows",
  "tokio",
  "toml 0.8.23",
@@ -5419,7 +5419,7 @@ dependencies = [
 
 [[package]]
 name = "openhuman"
-version = "0.63.0"
+version = "0.63.3"
 dependencies = [
  "aes-gcm",
  "aho-corasick",
@@ -5508,7 +5508,7 @@ dependencies = [
  "tar",
  "tempfile",
  "thiserror 2.0.19",
- "tinyagents 1.9.0",
+ "tinyagents 2.1.0",
  "tinychannels",
  "tinycortex",
  "tinyflows",
@@ -5525,7 +5525,6 @@ dependencies = [
  "tracing-appender",
  "tracing-log",
  "tracing-subscriber",
- "uiautomation",
  "unicode-width",
  "url",
  "urlencoding",
@@ -8250,17 +8249,13 @@ dependencies = [
 
 [[package]]
 name = "tinyagents"
-version = "1.9.0"
+version = "2.0.0"
 dependencies = [
  "async-trait",
  "bytes",
  "chrono",
- "chrono-tz",
- "dotenvy",
  "futures",
  "reqwest 0.12.28",
- "rhai",
- "rusqlite",
  "serde",
  "serde_json",
  "sha2 0.11.0",
@@ -8271,13 +8266,17 @@ dependencies = [
 
 [[package]]
 name = "tinyagents"
-version = "2.0.0"
+version = "2.1.0"
 dependencies = [
  "async-trait",
  "bytes",
  "chrono",
+ "chrono-tz",
+ "dotenvy",
  "futures",
  "reqwest 0.12.28",
+ "rhai",
+ "rusqlite",
  "serde",
  "serde_json",
  "sha2 0.11.0",
@@ -8371,7 +8370,7 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror 2.0.19",
- "tinyagents 1.9.0",
+ "tinyagents 2.1.0",
  "tracing",
 ]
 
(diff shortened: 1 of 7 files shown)


oxoxDev:
@coderabbitai review

Issue #154, opened 2026-07-29
feat(workflows): add a run-input field so a workflow can be given its topic/request


oxoxDev:
## Problem

On the Workflows tab, the Run result's **"Feature requested" field always shows "—"**, and there is **no input to supply what was requested before running**. In practice the workflow runs without a topic, then asks the operator to provide notes/topic — but there is no field to enter them, so the run cannot be completed with real input.

## Steps to reproduce

1. Open Workflows → Run a workflow.
2. There is no field to enter the request/topic.
3. The result's "Feature requested" shows "—"; the agent asks for the topic you have no way to provide.

## Expected

A workflow run should accept the operator's input (the "feature requested" / topic / notes) **before** it runs, thread it into the workflow's input, and echo it back in the run result (so "Feature requested" reflects what was actually asked).

## Fix

- Frontend (`WorkflowsView` / run dialog): add an input field for the request/topic; require it (or make optional per the workflow's declared inputs) before dispatching the run.
- Wire the value into the workflow run input payload so nodes receive it.
- Populate the result's "Feature requested" from that input instead of a hardcoded "—".

## Scope

Small–medium — a run-input field + plumbing into the workflow input + result display.

## Related

Depends on the workflow/delegation path being stable (the delegation stack-overflow crash) to actually complete a run end-to-end.

Issue #151, opened 2026-07-29
Delegated-agent replies never reach the human: no agent DM / no reply back to chat


oxoxDev:
## Problem

When a task is handed off to an agent from the company chat (e.g. `delegate_to_desk` / a hand-off), the agent runs but **there is no interface for the human to receive the reply**. The delegated agent's output is never routed back to a surface the operator can see — no agent DM thread, and the result does not post back into the chat where the task was created. The loop is open: you can dispatch work but never get the answer back.

## Steps to reproduce

1. In the Conversation/company chat, hand a task off to an agent / desk lead (delegation).
2. The sub-agent executes the turn (`run_background`).
3. Observe: no reply appears in the chat, there is no DM thread with that agent, and the delegated task's detail does not surface the agent's response/artifacts.

## Expected (per the product roadmap)

The roadmap defines the return paths this is missing:
- **Chat DMs with agents** and **§3.3 "Agent DM — the agent's console"**: a per-agent conversation thread (split view: conversation + agent card) where the operator sees the agent's replies and can follow up.
- **§3.2**: "agents post back into the channel where their task started" — e.g. *"Draft's ready, needs your approval"* as a task-card update in the originating thread.
- **§3.6 Task detail**: the delegated task's timeline shows the agent's output + artifacts.

The **hard rule** still holds: agent↔agent coordination stays in the trail, never in channels — this is specifically about the **agent→human reply** reaching the human who delegated.

## Proposed approach (combine / stage)

1. **Route the reply back to origin (MVP):** publish the delegated agent's turn output onto the originating conversation thread via the `turn_stream`/reply bus — as a message + a task-card status update ("done / needs approval") where the task was born. Closes the loop with the least new UI.
2. **Agent DM surface:** add a per-agent DM thread (Conversation → DMs → agent) so history is durable and the human can reply/steer. (roadmap §3.3)
3. **Task-detail output:** surface the sub-agent's result + artifacts in the delegated task's detail timeline.

## Dependencies / risks

- **Blocked by the delegation stack-overflow** (the `delegate_to_desk`/`run_workflow` re-entry currently aborts the host) — the hand-off path must be stable before replies can flow reliably. Fix that first / together.
- Reply routing must attribute cost per hop to the delegated task's trail (ties into the metering work).
- Respect the agent↔agent-not-in-channels rule: only the agent→human reply surfaces to chat/DM.

## Scope

[...]

PR #156, opened 2026-07-29
fix(harness): bound agent/workflow re-entry depth to stop stack-overflow crash (#155)

### LINKED ISSUE #155: Workflow / agent-delegation run overflows the stack and aborts the host
## Problem

Running a **workflow** or **delegating a task to an agent** from chat crashes the host: a tokio worker **overflows its stack and the process aborts**, taking the whole tenant server down. The console then shows *"cannot reach the company host at this origin"* (the host died mid-request). Simple single-turn chat is unaffected.

```
thread 'tokio-rt-worker' (…) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

## Steps to reproduce

1. From the company chat, run a workflow (or hand a task off to an agent / desk lead — `delegate_to_desk` / `run_workflow` / `spawn_task`).
2. The host aborts with the stack overflow above; subsequent requests fail with "cannot reach the company host".

## Root cause (hypothesis — under investigation)

Believed to be **unbounded async recursion on the delegation path, newly reachable after native tool-calling landed (#145)**. Before #145 the inference provider was text-only, so the orchestrator's delegation tools never actually executed and the path never recursed. Now that tools execute, `delegate_to_desk`/`run_workflow` run a sub-agent turn (`run_background`) which can itself delegate → the re-entry appears to recurse in-stack without a `Box::pin` boundary or a hop-depth guard → stack overflow. Note: the previously-known `task_local::scope` nesting overflow is already `Box::pin`-guarded (`src/harness/mod.rs` ~319–385) — this is a **different** cycle in the orchestration/delegation path.

## Impact

**HIGH / regression.** #145 is already merged to `main`, so this crash is now **latent on main** and will crash orchestration/workflows on staging once the tenant image is redeployed. It blocks the entire Workflows feature and the agent hand-off loop.

## Proposed fix (direction)

Bound the delegation re-entry — `Box::pin` the recursive future and/or add a delegation-depth guard, or dispatch the sub-agent turn via a fresh spawn instead of an in-stack recursive call. Add a regression test that runs a delegating/workflow turn and asserts it completes without overflow.

## Related

- Blocks #154 (workflow run input) and #151 (delegated-agent reply routing) — the same hand-off → run → reply loop.
- Exposed by #145 (native tool-calling).

oxoxDev:
## Summary

Bound agent/workflow **re-entry depth** so a workflow that (directly or transitively) runs an orchestrator agent can no longer recurse without limit and overflow the worker stack / abort the host. Closes #155.

## Problem

`run_workflow` runs the tinyflows engine **inline inside the tool call**, and the engine runs `agent` nodes **inline back into the harness** (all awaited in-stack, no spawn). When a workflow `agent` node resolves to the **orchestrator** — which itself carries `RunWorkflowTool` over the same shared runner — it calls `run_workflow` again → the engine runs the orchestrator node again → unbounded depth → `tokio-rt-worker` stack overflow → the tenant host aborts (console then shows "cannot reach the company host"). Reachable with shipped seed companies (e.g. the first roster agent defaulting to orchestrator while also being a pipeline node).

Existing guards don't catch it: tinyflows' `MAX_SUB_WORKFLOW_DEPTH=8` counts `sub_workflow` nodes *within one* `engine::run`, but each `run_workflow` tool call starts a **fresh** engine (depth reset); the `Box::pin` in `run_with_steer` bounds future *size*, not recursion *depth*. Made live by native tool-calling (#145) — before it the orchestrator tools never executed.

## Solution

Add a re-entry **depth guard that spans the harness↔engine boundary**:
- A `tokio::task_local!` counter `AGENT_REENTRY_DEPTH` + `const MAX_AGENT_REENTRY_DEPTH = 8` (matches tinyflows' sub-workflow bound).
- `HarnessPool::run_inner` (the single choke point for `run`/`run_background`/`run_steered*`) reads the task-local (defaulting to 0 outside any scope), returns a clean `Harness("agent/workflow re-entry depth exceeded")` once depth hits the cap, else runs the body inside `AGENT_REENTRY_DEPTH.scope(depth+1, Box::pin(..))`. The `Box::pin` at the scope boundary mirrors the existing `with_stop_hooks` pattern to avoid the nested-`task_local::scope` worker-stack trap.
- The error surfaces gracefully: `run_agent` maps it to `EngineError::Capability`, the engine run fails, and `RunWorkflowTool::execute` returns a `ToolResult::error` — the calling orchestrator turn replies normally and **the process stays up**. No changes needed at those call sites (they already handle a `crate::Error`).

## API Or Behavior Changes

No public API change. A workflow that self-recurses past 8 levels of agent-driven re-entry now fails that run with a clear tool-error instead of crashing the host. Legitimate nesting stays well under 8.

## Tests

- [x] `cargo build --features openhuman`
- [x] `cargo test --features openhuman` — `workflows::runner` 16/16 (incl. new `workflow_agent_reentry_is_bounded_not_a_stack_overflow`), `harness::` 211/211
- [x] `cargo clippy --features openhuman,composio --no-deps -- -D warnings` — clean

Regression test reproduces the overflow pre-fix (unbounded chain exhausts even a 32 MiB thread) and asserts a bounded clean refusal post-fix.

## Documentation

None needed — internal harness safety guard.

## Related

Closes #155. Exposed by #145 (native tool-calling). Unblocks the Workflows feature + the hand-off loop (#154, #151).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added protection against unbounded workflow and agent recursion.
  * Recursion now stops with a clear error instead of risking crashes or stalled runs.
  * Top-level workflow runs continue completing normally when recursive re-entry is detected.

* **Tests**
  * Added regression coverage for recursive workflow execution and clean error handling.

### DIFF
--- a/src/harness/mod.rs
+++ b/src/harness/mod.rs
@@ -82,6 +82,30 @@ use crate::ports::{
 };
 use crate::runtime::builder::agent_effective_grants;
 
+tokio::task_local! {
+    /// Re-entry depth across the harness↔workflow-engine boundary (issue #155).
+    ///
+    /// Incremented once per [`HarnessPool::run_inner`] entry — the single choke
+    /// point every agent turn flows through (`run` / `run_background` /
+    /// `run_steered` / `run_steered_background`). A workflow `agent` node
+    /// re-enters the pool via [`run_background`](HarnessPool::run_background);
+    /// when that node resolves to an orchestrator carrying the `run_workflow`
+    /// tool, it can call the engine again and re-enter here, so this counter is
+    /// the only thing that spans the harness→engine→harness cycle. Tinyflows'
+    /// own `MAX_SUB_WORKFLOW_DEPTH` counts `sub_workflow` nodes *within one*
+    /// `engine::run` and resets to 0 on every fresh `run_workflow` tool call, so
+    /// it cannot see the cycle. Reading this outside any scope defaults to 0.
+    static AGENT_REENTRY_DEPTH: usize;
+}
+
+/// The maximum harness↔engine re-entry depth before a turn is refused
+/// (issue #155). Matches tinyflows' `MAX_SUB_WORKFLOW_DEPTH` (8): a
+/// deep-but-legitimate workflow→agent→workflow nesting stays well under it,
+/// while the unbounded self-delegation cycle (an `agent` node that resolves back
+/// to the orchestrator, which runs the same workflow again) is cut before it can
+/// overflow the Tokio worker stack and abort the host.
+const MAX_AGENT_REENTRY_DEPTH: usize = 8;
+
 /// Shared dependencies every harness-built agent draws on.
 #[derive(Clone)]
 pub struct HarnessDeps {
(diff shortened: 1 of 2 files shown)


oxoxDev:
Superseded by #164, which bounds the same workflow re-entry recursion at the tinyflows runner (the only unbounded path per root-cause). Closing this harness-level duplicate in favor of #164.

PR #161, opened 2026-07-29
fix(usage): emit OauthCall samples so oauthCalls/connections aren't always zero

### LINKED ISSUE #152: fix(usage): emit OauthCall samples so oauthCalls/connections aren't always zero
## Problem

The Usage page shows **`oauthCalls` and `connections` = 0** (and the by-provider breakdown empty) despite connected-tool / OAuth activity.

## Root cause

The read/aggregation side is already correct: `src/metering/usage.rs` sums `SampleKind::OauthCall` into `oauth_calls`, `by_provider`, and `connections`. The gap is on the **write side — `SampleKind::OauthCall` is never emitted from any live execution path.** `src/server/ops/usage.rs` documents this explicitly:

> `SampleKind::OauthCall` … are not yet emitted anywhere, so `byProvider` / `oauthCalls` read zero until the Phase 2 [wiring].

So the counters are structurally always zero regardless of activity.

## Fix

Emit an `OauthCall` `UsageSample` at the connected-tool / OAuth execution site (where a composio/MCP OAuth-backed tool call actually runs), tagged with the provider, so the existing aggregation surfaces it. Add a test that a connected-tool execution records one `OauthCall` sample.

## Scope

Small — one emit site + a test; the read model already handles it.

M3gA-Mind:
## Summary

Fixes #152.

`bucket_usage` has always summed `SampleKind::OauthCall` into `oauthCalls`,
`byProvider`, and `connections` — but **nothing ever emitted one**, so all three
counters were structurally zero regardless of connected-tool activity.
`server/ops/usage.rs` documented its own gap ("not yet emitted anywhere … until
the Phase 2 emit work lands"). This is that emit side.

`metering::oauth` mints the sample and records it; `composio_execute` calls it
once per completed action, tagged with the action's toolkit
(`GMAIL_SEND_EMAIL` → `gmail`).

### Three judgement calls

**The logic lives in `metering` (always compiled), not at the call site.** The
connected-tool execution sites are behind non-default features (`composio`,
`mcp`) and CI builds only the default set, so code written *at* those sites is
neither compiled nor tested by CI. Keeping the sample shape and the
record-and-swallow behaviour beside the aggregation that reads it means the
contract is unit-tested on every CI run and the gated site is a single call.

**Metered on success only.** `connections` in the read model is the *count of
providers seen*, so counting a failed call would report a connection for a
provider this company may not be connected to at all. A call rejected by the
toolkit allowlist, or one with malformed arguments, never reaches the provider
and is not metered.

**Providers are normalised** (trimmed, lowercased) before recording. `byProvider`
groups on the raw string and `connections` is the count of those groups, so
`GitHub` and `github` would otherwise read as two connections with one call each
instead of one connection with two.

Metering never fails the work it meters: `record_oauth_call` logs and swallows a
meter error. A full disk must not turn a Gmail send that already happened into a
tool-call failure the agent then retries.

### Knock-on fix: zero-token samples in "tokens by teammate"

`bucket_usage` fed *every* sample into the per-teammate token accumulator. That
branch was unreachable by a zero-token sample before this change, because none
were ever emitted. Now that they are, an agent whose whole window was
connected-tool calls would render as a 0-token bar in a token chart. Guarded,
with a test.

## API Or Behavior Changes

- New `metering::oauth` module: `oauth_call_sample`, `record_oauth_call`,
  `normalize_provider`, `UNKNOWN_PROVIDER`. Additive.
- `harness::composio::composio_tools` takes a second argument, the new
  `ComposioMetering { company, agent, meter }`. A `None` meter leaves metering
  off and the tools behave exactly as before. Only reachable under
  `--features composio`.
- `bucket_usage` no longer creates a `byAgent` row for a sample carrying no
  tokens. No existing behaviour changes on a token-bearing sample.
- No wire-format change. `GET …/usage` keeps its shape; `byProvider` /
  `oauthCalls` simply stop being permanently zero on a build that has a
  connected-tool surface.

## Tests

Always compiled, so **CI runs these**:

- `metering/oauth.rs` — sample is zero-token/zero-cost and correctly kinded;
  provider normalisation collapses casing; a blank provider falls back to
  `unknown` rather than dropping the call; `record_oauth_call` writes one sample
  per call; a failing meter never surfaces to the caller; and
  `emitted_samples_reach_the_console_counters` drives emitted samples through
  `bucket_usage` to assert `oauthCalls` / `byProvider` / `connections` are the
  numbers the issue says should be non-zero.
- `metering/usage.rs` — `oauth_only_agents_do_not_appear_as_zero_token_teammates`.

Behind `--features composio` (**not built by CI** — see limitations):

- `allowlist_rejection_records_no_sample`, `missing_tool_argument_records_no_sample`.

### Limitations, stated plainly

- **CI does not compile the `composio` feature**, so the wiring in
  `harness/composio.rs` and `harness/build.rs` is not covered by a green check on
  this PR. It needs review eyes. (Pre-existing repo-wide gap — the same is true
  of every `openhuman` / `mcp` / `media` path, and is already noted in
  `harness/build.rs`.)
- The **positive** end-to-end path (a real `composio_execute` → provider →
  exactly one sample) is not unit-tested: it requires a live `ComposioClient`
  round-trip. The two rejection paths are tested because they return before any
  network call.
- **MCP is deliberately out of scope.** `harness/mcp.rs` maps both
  `AuthMaterial::OAuth` and a static bearer onto the same
  `McpAuthConfig::BearerToken`, so the call site cannot distinguish an
  OAuth-backed server from a statically-credentialed one — metering all of them
  as OAuth calls would over-count `connections`. Threading the original
  `AuthMaterial` through to `OcMcpCallTool` is a separate change; Composio alone
  is enough for the counters to stop being structurally zero.

Run locally:

- [x] `rustfmt --edition 2024 --check` on every changed Rust file
- [ ] `cargo fmt --all -- --check` — cannot run locally: `cargo metadata` fails
      because the `vendor/` submodules are not initialised in this checkout.
      Covered by CI.
- [ ] `cargo clippy --all-targets -- -D warnings` — same; relying on CI.
- [ ] `cargo test` — same; relying on CI (covers `metering::oauth`, not the
      composio wiring).

## Documentation

`server/ops/usage.rs` module docs and the `byProvider` field doc updated: the
"not yet emitted anywhere / Phase 2" note is replaced with the accurate state —
samples are emitted per completed connected-tool call, and a default build reads
zero because it compiles in no connected-tool surface, not because a hook is
missing. New `metering/oauth.rs` carries its own module docs covering the
placement rationale and the never-fail-the-caller contract.

### DIFF
--- a/src/harness/build.rs
+++ b/src/harness/build.rs
@@ -261,7 +261,18 @@ pub fn build_agent(
     #[cfg(feature = "composio")]
     if crate::company::grants_composio_explicit(grants) {
         match &deps.composio {
-            Some(config) => tools.extend(crate::harness::composio::composio_tools(config)),
+            // The metering handle lets `composio_execute` record an `OauthCall`
+            // usage sample per completed call, so the Usage view's
+            // calls-by-provider chart reflects real connected-tool activity
+            // (issue #152). A `None` meter simply leaves metering off.
+            Some(config) => tools.extend(crate::harness::composio::composio_tools(
+                config,
+                crate::harness::composio::ComposioMetering {
+                    company: company.clone(),
+                    agent: manifest_agent.id.clone(),
+                    meter: deps.meter.clone(),
+                },
+            )),
             None => tracing::warn!(
                 company = %company,
                 agent = %manifest_agent.id,
--- a/src/harness/composio.rs
+++ b/src/harness/composio.rs
@@ -147,7 +147,7 @@ fn slug_toolkit(slug: &str) -> String {
 }
 
 #[cfg(feature = "composio")]
-pub use live::composio_tools;
+pub use live::{ComposioMetering, composio_tools};
 
 #[cfg(feature = "composio")]
 mod live {
@@ -158,11 +158,31 @@ mod live {
     use serde_json::{Value, json};
 
     use crate::harness::mcp_probe::scrub;
+    use crate::metering::record_oauth_call;
+    use crate::ports::UsageMeter;
+    use crate::ports::now_millis;
 
     use oh::composio::ComposioClient;
     use oh::integrations::IntegrationClient;
     use oh::tools::traits::{PermissionLevel, Tool, ToolResult};
     use openhuman_core::openhuman as oh;
+
+    /// What `composio_execute` needs to meter a call it just made: the company
+    /// the sample belongs to, the agent that made it, and the meter to write to.
+    ///
+    /// Bundled because these three travel together and are individually
+    /// meaningless — and because [`composio_tools`] would otherwise grow four
+    /// positional parameters at a call site that already has many.
+    #[derive(Clone)]
+    pub struct ComposioMetering {
+        /// The company the sample is scoped to.
+        pub company: CompanyId,
+        /// The agent whose turn made the call.
+        pub agent: String,
+        /// The usage meter. `None` leaves metering off entirely (the harness
+        /// wires no meter in some embeddings) — the tools still work.
+        pub meter: Option<Arc<dyn UsageMeter>>,
+    }
 
     /// Build the five per-tenant Composio tools over the tenant's bearer token.
     ///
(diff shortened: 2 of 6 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai full review

oxoxDev (approved):
Solid, well-tested. Right call keeping the emit logic in always-compiled `metering/oauth` (CI only builds the default feature set) — the bucket_usage round-trip proves the three counters stop being zero. Success-only metering and the swallow-on-meter-failure are both correct. Nice catch on the zero-token guard in usage.rs so an OAuth-only agent doesn't render as an empty token bar. One note: `composio_tools`'s new signature only compiles under the `composio` feature, which CI doesn't build — I'll verify the `openhuman,composio` combo builds locally before merge. LGTM.

PR #163, opened 2026-07-29
fix(delegation): answer in the thread a dispatched task came from (#151 part b)

M3gA-Mind:
**Part (b) of #151** — the §3.2 MVP return path, no new UI. First of three PRs:

| | Part | Status |
|---|---|---|
| 1 | **(b)** dispatched-task post-back + adapter drop | **this PR** |
| 2 | (a) delegation re-entry guard (the Path-A symptom) | next |
| 3 | (c) agent DM surface (§3.3) | after (a) |

**#151 stays open until all three land** — only PR 3 will close it.

## Summary

Two places a delegated reply was being lost. Neither is the `delegate_to_desk` chat path, which — see below — already works.

**1. A dispatched card's answer never left the board.** `spawn_task` creates a card and returns no reply by design; the work happens later, when `TaskDispatched` fires. `run_task` folded the result into `card.note`, moved the column, and returned `()`. Nothing was pushed onto the cycle's channel responses, so the only way to see the answer was to go to the board and open the card.

`TaskRecord` now remembers the conversation it came from (`origin_chat_id`, stamped by `spawn_task` from the delegating turn's chat id), and a finished card posts one bubble back to that thread — the card's title, where it landed, and the result — attributed to the responder and threaded with `reply_to`. The note is still written and remains the durable record; the post-back is additive.

The post-back is derived from the **card**, not the raw turn reply, so a cancelled or paused run reports what the board actually shows instead of claiming an answer it does not have.

**2. Agent-addressed replies were dropped on any real channel adapter.** `route_response` matched `OutboundMessage.channel` against registered adapters and, on no match, dropped the message — the comment said *"drop silently in Phase 1"*. But an agent reply is addressed by **agent id**, not adapter id: a delegated desk bubble (`brain.rs`, `channel: member`) and this new post-back both do that so the console can attribute them, and no adapter answers to an agent id.

The operator REST route never noticed, because it reads `CycleReport.responses` directly and bypasses `route_response` entirely. A company reached over a real channel adapter, though, received the orchestrator's reply and lost every delegated one. Unmatched responses now fall back to the operator adapter and are forwarded **unchanged**, so `channel` still names the agent and attribution survives.

## Scope note — what this PR deliberately does not claim

While tracing the reply path I could not find a drop on the **synchronous `delegate_to_desk`** path described in the issue:

```
run_delegation → run_steered → OutboundMessage{channel: <agent_id>}   brain.rs
  → channel_responses.extend(delegated)                               brain.rs
  → operator.rs journals AgentReply{chat_id: desk, agent_id} for EVERY response
     and returns them all in the HTTP body
  → Conversation.tsx renders ALL reply.responses, attributed by channel
```

That path appears to deliver the reply into the originating thread already. **The live symptom could not be reproduced locally** — this stack is not runnable here — so this PR fixes the two drops that are demonstrable from the code and leaves the reported Path-A symptom to PR 2, where the likeliest cause (unbounded `run_workflow` → orchestrator-agent-node re-entry aborting the host) is addressed. If the host crashes mid-turn, no reply arrives regardless of routing.

## API Or Behavior Changes

**Additive, and fails safe in both halves.**

- `TaskRecord` gains `origin_chat_id: Option<String>` with `#[serde(default, skip_serializing_if = "Option::is_none")]`, so stored records round-trip byte-identically and the key is absent on the wire unless set. A card with no remembered origin — created straight on the board, or written before this field existed — posts back nowhere and behaves exactly as before.
- `HarnessBrain::run_task` returns `Result<Option<OutboundMessage>>` instead of `Result<()>` (private method, one call site).
- `route_response` no longer drops unmatched responses when an operator adapter is registered. A runtime with **no** operator adapter keeps the previous non-delivery, now logged rather than silent.

## Tests

Added:

- `postback_reports_the_card_title_status_and_result`
- `postback_reflects_the_landing_column_not_a_presumed_success` — paused and cancelled runs must not read as success
- `postback_without_a_note_is_still_a_complete_sentence` — including a whitespace-only note
- `a_card_with_no_origin_posts_back_nothing` — the compatibility guarantee
- `a_card_with_an_origin_posts_back_to_that_thread` — attribution + `reply_to` + empty steps
- `a_reply_addressed_by_agent_id_reaches_the_operator_channel` — drives a real cycle through a recording `OperatorChannel` and asserts both replies land with the delegated one's `channel`/`reply_to` intact

Commands run locally:

Since none of this is compiler-checked locally, the mechanical part carries the most risk and is worth a reviewer's eye: adding a field to `TaskRecord` required edits at **9 construction sites across 6 files**. A scripted first pass mis-fired and inserted the field into two `-> TaskRecord {` *function bodies* in `brain.rs`; both were caught and reverted by inspection. Two new tests initially called the non-async `brain_with_tasks()` with `.await` and were corrected. All added lines are within the width limit (the long lines the diff touches are pre-existing).

## Documentation

No separate doc change: the new field, the post-back rule, and the adapter fallback are documented inline where the behaviour lives, each explaining what used to happen and why the new path fails safe.

Refs #151

## Summary by CodeRabbit

* **New Features**
  * Task completion updates now post results back to the conversation where the task originated.
  * Completion messages indicate the task’s final status and include relevant notes or results.
  * Delegated tasks retain their originating conversation for accurate follow-up.
* **Bug Fixes**
  * Responses are no longer dropped when a destination channel is unavailable; they are delivered through the operator channel while preserving reply context.

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -849,7 +849,10 @@ description = "Builds it."
         let (brain, tasks) = brain_with_tasks(dir.path());
         let mut c = card("t-no-origin", "maya");
         c.origin_chat_id = None;
-        tasks.upsert(&CompanyId::new("acme"), &c).await.expect("seed");
+        tasks
+            .upsert(&CompanyId::new("acme"), &c)
+            .await
+            .expect("seed");
 
         let posted = brain.run_task("t-no-origin").await.expect("run");
         assert!(
@@ -868,7 +871,10 @@ description = "Builds it."
         let (brain, tasks) = brain_with_tasks(dir.path());
         let mut c = card("t-origin", "maya");
         c.origin_chat_id = Some("strategy".to_string());
-        tasks.upsert(&CompanyId::new("acme"), &c).await.expect("seed");
+        tasks
+            .upsert(&CompanyId::new("acme"), &c)
+            .await
+            .expect("seed");
 
         let posted = brain
             .run_task("t-origin")
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -72,9 +72,17 @@ impl HarnessBrain {
     /// onto the board — moved to `in_review` on success, back to `backlog` with
     /// the error noted on failure. A missing task store or a card that has since
     /// vanished is a silent no-op.
-    async fn run_task(&self, task_id: &str) -> Result<()> {
+    /// Runs a dispatched card to completion and, when the card remembers the
+    /// conversation it was spawned from, returns the reply to post back there
+    /// (issue #151 §3.2).
+    ///
+    /// Before this the answer only ever reached `card.note`: the card runs
+    /// asynchronously, long after the turn that spawned it has answered, so the
+    /// operator had to know to go and look. The note is still written — it stays
+    /// the durable record — and the post-back is additive.
+    async fn run_task(&self, task_id: &str) -> Result<Option<OutboundMessage>> {
         let Some(tasks) = self.deps.tasks.as_ref() else {
-            return Ok(());
+            return Ok(None);
         };
         let Some(mut card) = tasks
             .list(&self.record.id)
@@ -82,7 +90,7 @@ impl HarnessBrain {
             .into_iter()
             .find(|t| t.id == task_id)
         else {
-            return Ok(());
+            return Ok(None);
         };
 
         let responder = self.task_responder(&card.assignee);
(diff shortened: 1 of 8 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@greptileai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Part of **tinyhumansai/opencompany#151** (part b — reply in the originating thread). #151 tracks 3 PRs: #163 (b), #164 (a), #165 (c); it's closed by the final one (#165), not this PR.

PR #200, opened 2026-07-30
ci: build and publish this repo's tenant image to staging

oxoxDev:
## Summary

This repo owns the tenant workload, so it should own the tenant image. Adds `deploy-staging.yml`: build → push `opencompany-tenant` to ECR → point the manager's `OCM_TENANT_IMAGE` at the immutable `staging-<sha>` tag.

Companion PR: `tinyhumansai/opencompany-microservice#7` — that repo drops the tenant build and keeps only its own manager image.

## API Or Behavior Changes

None in the crate. This is CI only. Two operational behaviours worth stating explicitly:

- **New and woken tenants** boot on the new image as soon as `OCM_TENANT_IMAGE` moves.
- **Already-running tenants keep their current image** until recreated. That is deliberate — restarting them interrupts live agent work. `workflow_dispatch` exposes `refresh_all_tenants` to roll them all when you actually want that.

`TENANT_FEATURES` is carried over verbatim from the pipeline this replaces (`mongodb,openhuman,openhuman-rpc,tinyplace,github,smtp,dns,tinyhumans,webhooks,platform-jwt,export`). Nothing is added or removed, so the shipped feature set is unchanged. Editing that list is a product change, not a CI tweak, and the workflow says so.

## Problem

The tenant image was built from the platform superproject, which cloned this repo as a submodule and rebuilt on pointer movement. Two consequences:

1. **A change here shipped only when that repo's cron noticed** — this repo had no deploy path of its own.
2. **It had not shipped anything since 2026-07-20.** That pipeline's checkout passes `token: ${{ secrets.SUBMODULES_PAT }}`, the secret was never set, and every scheduled run failed with `Input required and not supplied: token`.

The PAT existed because the superproject cloned `backend` and `tinyhuman-landing` (both private) purely to sync their submodule pointers — neither is compiled by any build. Everything the tenant image actually compiles is public: this repo, `openhuman`, and openhuman's nested vendor crates. So building here needs **no** cross-repo token; the default `GITHUB_TOKEN` clones the whole graph.

## Tests

- `deploy-staging.yml` parses as valid YAML.
- Verified against the live staging cluster (read-only) that the target objects match what the workflow assumes: namespace `opencompany-system`, deployment `opencompany-manager`, ECR pull via the `ecr-creds` `dockerconfigjson` Secret, and tenant StatefulSets labelled `app=opencompany`.
- The build itself is unchanged from what the superproject ran — same `Dockerfile`, same `FEATURES`, same context — so this is a relocation of a working build rather than a new one.
- **Not exercised end to end** until the prerequisites below exist. First push to `main` is the real test; `workflow_dispatch` allows a deliberate trigger.

## Documentation

The workflow header documents the one-repo-one-image split, the required secrets, and why no PAT is needed. Deployment semantics (immutable tags, pull-secret refresh, tenant-restart policy) are documented in the inline comments at the point they matter, and in the companion PR's `deploy/ci/README.md`.

## Notes for review

**The submodule fetch is path-limited on purpose:**

```
git submodule update --init --recursive vendor/openhuman vendor/tinyagents
```

Recursing from the repo root would also pull openhuman's desktop-only submodules — `tauri-cef` is a heavy CEF checkout, and `tauri-plugin-notification` — which no server build compiles. Both are needed by `--features openhuman` only insofar as cargo must read the manifests behind `[patch.crates-io]`, which this covers.

**SSH rather than a kubeconfig** matches `tinyhumansai/backend` and is load-bearing: the kubelet's ECR pull secret expires after 12h and is refreshed by `/usr/local/bin/ecr-refresh.sh` **on the node**, which cannot be invoked through the Kubernetes API.

**Immutable tag, not `:staging`.** Tenant pods pin `imagePullPolicy: IfNotPresent`, so re-pushing a moving tag is silently ignored. The manager is pointed at `staging-<sha>`; `:staging` is a convenience alias only.

## Prerequisites (blocking — repo/AWS admin)

| What | Where |
|---|---|
| `AWS_ROLE_ARN` secret | this repo |
| OIDC trust policy updated to allow `repo:tinyhumansai/opencompany:*` | AWS |
| `SSH_PRIVATE_KEY`, `SSH_HOST` secrets | this repo |

If those secrets live in a GitHub Environment rather than at repo level, the job needs a matching `environment:` key — say so and I will add it.

## Related

- Companion: `tinyhumansai/opencompany-microservice#7`.
- Best landed together: #7 removes the old tenant pipeline, this adds the replacement.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -1,0 +1,159 @@
+# Build this repo's OWN image (the tenant workload) and put it in front of new
+# tenants on k3s staging.
+#
+# One repo, one image. This used to be built from the platform superproject
+# (tinyhumansai/opencompany-microservice), which meant a change here only shipped
+# when that repo's cron happened to notice the submodule pointer move. Now this
+# repo builds and publishes its own workload image, and the superproject builds
+# only the manager control plane.
+#
+# Required repository secrets:
+#   AWS_ROLE_ARN     — role assumed via OIDC to push to ECR. Its trust policy
+#                      must allow `repo:tinyhumansai/opencompany:*`.
+#   SSH_PRIVATE_KEY  — key trusted by root@ on the staging node.
+#   SSH_HOST         — staging node address.
+#
+# No PAT is needed: this repo and every crate the build compiles
+# (openhuman and its nested vendor graph) are public, so the default
+# GITHUB_TOKEN clones them all.
+name: Deploy tenant image to staging
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'src/**'
+      - 'frontend/**'
+      - 'companies/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'Dockerfile'
+      - '.dockerignore'
+      - 'vendor/**'
+      - '.github/workflows/deploy-staging.yml'
+  workflow_dispatch:
+    inputs:
+      refresh_all_tenants:
+        description: 'Also restart already-running tenants onto the new image (disruptive: wakes idle tenants)'
+        type: boolean
+        default: false
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  id-token: write # assume the AWS role via OIDC to push to ECR
+
+env:
+  AWS_REGION: us-east-1
+  ECR_REPOSITORY: tinyhumansai/opencompany-tenant
+  K8S_NAMESPACE: opencompany-system
+  MANAGER_DEPLOY: opencompany-manager
+  # The feature set the hosted tenant ships with. Kept verbatim from the
+  # pipeline this replaces — changing it changes what every tenant can do, so
+  # treat edits here as a product change, not a CI tweak.
+  TENANT_FEATURES: mongodb,openhuman,openhuman-rpc,tinyplace,github,smtp,dns,tinyhumans,webhooks,platform-jwt,export
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


oxoxDev:
Added 3c2e38e — a required fix, not a refinement.

Pushing to ECR is not sufficient to make a tenant image runnable. Tenant StatefulSets carry no `imagePullSecrets`, tenant namespaces have no `ecr-creds`, and there is no node-level `registries.yaml`, so a tenant pod pulling from ECR gets `403 Forbidden` and lands in `ImagePullBackOff`. I hit this for real while restoring the smoke1 tenant:

```
Failed to pull image ".../tinyhumansai/opencompany-tenant:staging": ...
unexpected status from HEAD request to https://<acct>.dkr.ecr.us-east-1.amazonaws.com/v2/tinyhumansai/opencompany-tenant/manifests/staging: 403 Forbidden
```

Worth being explicit that this is pre-existing, not introduced here: `OCM_TENANT_IMAGE` already points at ECR on staging, so **any** tenant wake currently fails this way — `john` and `tinyhumans-marketing` would too; nobody noticed because they are scaled to zero. The workflow would simply have kept publishing images no tenant could pull.

The fix pre-pulls the pushed image into the node's containerd using the credential the kubelet already holds, so every tenant's `imagePullPolicy: IfNotPresent` resolves locally and never authenticates. It fails the deploy if the image is not in containerd afterwards, rather than leaving tenants to discover it at wake time.

**This is single-node-staging specific and I would not call it the real fix.** The durable one belongs in the manager: provision a pull secret per tenant namespace and set `imagePullSecrets` on the tenant pod spec, with the refresh timer covering those namespaces. That is an `opencompany-microservice` change, so it does not belong in this PR — happy to file it as an issue there if you would like it tracked rather than remembered.

PR #212, opened 2026-07-31
fix(telegram): receive inbound over getUpdates polling, not a webhook

### LINKED ISSUE #203: telegram: inbound webhook is unreachable on local/self-host — mirror OpenHuman's getUpdates polling + managed-bot link
## Symptom
The console's **Telegram** channel card tells the operator to create a bot with @BotFather, paste its token + a webhook secret, and point the bot's webhook at:

```
http://127.0.0.1:8091/hooks/e2e-harness-co/telegram
```

On any **local or self-hosted** instance (and any deploy without a public, Telegram-reachable URL) this **cannot work** — Telegram's servers deliver webhooks by POSTing to the URL, and they can't reach `127.0.0.1` (or any private host). So inbound Telegram DMs never arrive; the whole "receive and reply" flow is dead in these deployments. The URL is derived from the public host + `/hooks/{company}/telegram`, so it's only viable on a hosted tenant with a real public URL — and even there it diverges from how OpenHuman does it.

Found while UI-testing the consolidated build on the `e2e_harness` company.

## How OpenHuman does it (the reference to mirror)
OpenHuman does **not** use per-user inbound webhooks. It uses:
- **`getUpdates` long-polling** — a persistent outbound listener, works behind NAT/localhost with no public URL. See `vendor/openhuman/src/core/runtime/services.rs:178` ("Realtime channel listeners (Telegram getUpdates, Discord gateway, …)") and the `TelegramChannel` connector under `src/openhuman/channels/providers/telegram/`, brought online by the channel supervisor.
- **A managed shared bot** rather than BYO-BotFather — `alphahumantest_bot` (staging) / `openhumanaibot` (production), see `src/openhuman/channels/controllers/ops/telegram.rs:16-33`.
- **An account-link / pairing flow** — RPCs `telegram_login_start` / `telegram_login_check` (`channels/README.md`), so the user links their Telegram to the managed bot in one step instead of creating a bot and configuring a webhook.

Net: OH's model needs no public endpoint and no per-user bot creation; OC re-implemented Telegram as a per-company BYO-bot + inbound webhook, which breaks off-cloud and adds setup burden.

## Current OC wiring (for whoever picks this up)
- `src/server/hooks.rs` — `POST /hooks/{company}/telegram`, constant-time `X-Telegram-Bot-Api-Secret-Token` check → `run_cycle(WebhookReceived)`.
- `src/server/ops/channels.rs` — `webhook_url()` = `{public_host}/hooks/{id}/telegram`; `POST …/channels/telegram/webhook` calls `setWebhook`.
- `src/company/telegram.rs` — `TelegramApi::set_webhook` (passes `secret_token`).
- `frontend/src/views/connections/ChannelsSection.tsx` — the Telegram setup card (token + secret + the pasted webhook URL).

## Suggested direction (not started — for discussion)
Mirror OpenHuman:
1. Add a per-company **`getUpdates` polling listener** (reuse/adapt OH's `TelegramChannel`) so inbound works without any public URL.
2. Prefer a **managed bot + account-link** flow over BYO-BotFather+webhook; keep BYO as an optional advanced path.
3. If the webhook path is retained for hosted mode, **only surface it when a real public `OPENCOMPANY_PUBLIC_URL` is configured** — never present a `127.0.0.1` webhook the operator can't use.

## Repro
1. Run a local instance (`--features openhuman,…,telegram`), open the console → Connections → Telegram.
2. The card shows a `http://127.0.0.1:<port>/hooks/<company>/telegram` webhook URL.
3. There is no way for Telegram to deliver to it → inbound DMs never reach the agent.

M3gA-Mind:
## Summary

Fixes #203.

Telegram's webhook can only reach a host on the public internet, but `webhook_url()` derived its URL from `AppConfig::host_base_url()` — which falls back to `http://{bind}`. So the console told every operator to point BotFather at `http://127.0.0.1:8091/hooks/<company>/telegram`. On any local or self-hosted instance that URL is undeliverable, so inbound DMs never arrived and the whole receive-and-reply flow was dead.

Mirroring OpenHuman, the host now dials **out** instead of waiting to be called.

- **`runtime::telegram_poller` (new)** — a per-company `getUpdates` long-poll, spawned in `serve` beside `MailboxPoller` and stopped by the same shutdown `Notify`. It idles until a bot token is stored and picks one up on the next tick, so pasting a token into the console starts inbound with no restart; it skips while the company is asleep (scale-to-zero). The poll offset is **in-memory by design**: Telegram holds the confirmed watermark, so a fresh process asking with no offset gets only what was never acked — a restart resumes exactly, without replaying an answered DM and without any local persistence. Pacing comes from Telegram's server-side long-poll hold, so only an idle or failed tick backs off.
- **`TelegramApi`** grows `get_updates` / `get_webhook_info` / `delete_webhook`; `HttpTelegramApi` routes every method through one token-scrubbing `call` helper.
- **The webhook becomes an optional hosted fast-path**, surfaced only when `OPENCOMPANY_PUBLIC_URL` is a public **https** URL (new `AppConfig::public_webhook_base_url`, deliberately with no bind fallback). Elsewhere `webhookUrl` is `null` and `POST …/channels/telegram/webhook` is refused with `400` and an explanation. Refusing matters: a registration Telegram can't deliver to *also* blocks `getUpdates`, so accepting one would take down the path that does work.
- **The two paths never both consume an update.** The poller checks `getWebhookInfo` before polling: on a publicly reachable host it stands by (re-checking each tick, so unregistering resumes polling); on a host with no public URL a registration can only be a dead endpoint, so it clears it and takes inbound back — `deleteWebhook` keeps pending updates, so anything queued arrives in the first batch. Both paths run the same turn and share the extracted `telegram::deliver_replies`.
- **Console** — asks for a bot token and says polling needs no public URL; the webhook block (URL + secret + register) renders only when the host advertises a usable URL.

## API Or Behavior Changes

Yes — three, all intended by the issue:

- `GET …/channels/telegram` → `webhookUrl` is now `string | null` (`null` on any host with no public https URL, so the console never shows an undeliverable URL). `SetWebhookResult.webhookUrl` likewise.
- `configured` now means **a bot token is stored**, no longer "token *and* webhook secret" — a token alone is a working channel via polling. A new `polling: boolean` field reports whether the host has the outbound transport wired.
- `POST …/channels/telegram/webhook` returns `400` on a host with no public https URL instead of registering a loopback URL.

New env var: `OPENCOMPANY_TELEGRAM_POLL_SECONDS` (default `30`).

Not in scope: the managed shared-bot + `telegram_login_start`/`_check` account-link flow (the issue's suggested direction #2). That needs a backend-side managed bot and a pairing surface this repo doesn't have; the issue's own refinement comment scopes the minimal fix to the personal-bot path, which is what this PR ships.

## Tests

12 new tests, 2 existing updated to the new contract; 910 passing.

- Poller (`src/runtime/telegram_poller.rs`): `polls_updates_and_replies_without_any_webhook` (the issue's ask — a token alone receives and answers a DM), `clears_an_unreachable_webhook_and_takes_over_inbound` (the exact broken state from the repro), `stands_by_for_a_webhook_on_a_publicly_reachable_host` (and resumes when it is deleted), `idles_without_a_token_then_picks_one_up_live`, `skips_while_the_company_is_not_running`, `poll_errors_are_scrubbed_and_rearm_the_handshake`, `a_new_token_resets_the_offset`.
- Server (`src/server/ops/`): `telegram_set_webhook_is_refused_on_a_host_telegram_cannot_reach`, `telegram_is_configured_by_a_bot_token_alone`, `status_omits_the_webhook_url_on_an_unreachable_host`; `telegram_set_webhook_registers_the_public_url` and `telegram_config_is_write_only_and_status_reads_back` updated.
- Config (`src/app/types.rs`): `public_webhook_base_url_requires_an_explicit_https_url`.

Commands run locally:

- [x] `cargo clippy --all-targets --features telegram -- -D warnings`
- [x] `cargo test` (910 passed, 0 failed)
- [x] `pnpm typecheck` (frontend)

## Documentation

- `docs/modules/runtime/README.md` — new "Background listeners" section covering both pollers and how the two Telegram inbound paths divide.
- `docs/modules/server/README.md` — new "Inbound channel webhooks" section explaining that `/hooks/{company}/telegram` is the optional hosted fast-path and when it is surfaced.
- `gitbooks/developers/configuration.md` — new "Channels: Telegram" section (operator-facing setup + `OPENCOMPANY_TELEGRAM_POLL_SECONDS`).

## Summary by CodeRabbit

* **New Features**
  * Added automatic Telegram long-polling for local and self-hosted deployments.
  * Added optional webhook support for deployments with a publicly reachable HTTPS URL.
  * Webhooks and polling now coordinate automatically, including fallback when webhooks are unavailable.
  * Telegram connections can be configured with a bot token alone.
  * Added transport status indicators and clearer setup guidance.

* **Documentation**
  * Documented polling intervals, webhook requirements, configuration, coexistence rules, fallback behavior, and shutdown handling.

### DIFF
--- a/docs/modules/runtime/README.md
+++ b/docs/modules/runtime/README.md
@@ -48,6 +48,31 @@ count dropping to zero — so a schedule saved later onto a still-unwired compan
 is reported rather than swallowed by the latch. A company with no scheduled
 workflows and no runner is not misconfigured and stays silent.
 
+## Background listeners
+
+Two per-company background loops sit beside the scheduler, both spawned in
+`serve` and stopped by the same shutdown `Notify`:
+
+- `mailbox_poller.rs` — the IMAP mailbox poll (feature `imap`), on a fixed
+  interval (`OPENCOMPANY_MAIL_POLL_SECONDS`, default `60`).
+- `telegram_poller.rs` — Telegram `getUpdates` long-polling (feature
+  `telegram`), the inbound path that **needs no public URL**. It dials out to
+  `api.telegram.org`, so it works on localhost, behind NAT, and on any
+  self-hosted box — where Telegram's servers can never reach an inbound
+  `/hooks/{company}/telegram` route. Setup is the bot token alone; the loop
+  idles until one is stored and picks up a token pasted into the console on its
+  next tick, with no restart. Long-poll hold and idle back-off are
+  `OPENCOMPANY_TELEGRAM_POLL_SECONDS` (default `30`).
+
+The webhook route (`server::hooks`) stays as an optional hosted fast-path, and
+is offered only when `OPENCOMPANY_PUBLIC_URL` is a public **https** URL. The two
+paths never both consume an update: Telegram refuses `getUpdates` while a
+webhook is registered, so the poller checks `getWebhookInfo` first and stands by
+on a publicly reachable host — while on a host with no public URL a registered
+webhook can only be a dead endpoint, so it clears it and takes inbound back.
+Both paths run the same turn and share `telegram::deliver_replies`, so which one
+delivered an update is invisible downstream.
+
 ## Harness pool (`src/harness/`, feature `openhuman`)
 
 `src/harness/` embeds `openhuman_core` as a library (see
--- a/docs/modules/server/README.md
+++ b/docs/modules/server/README.md
@@ -122,3 +122,17 @@ Relevant configuration:
   `https://api.tiny.place`).
 - `OPENCOMPANY_PUBLIC_URL` — public host base embedded in published Agent Card
   endpoints. When unset, the endpoint falls back to `http://{bind}`.
+
+## Inbound channel webhooks (`hooks.rs`)
+
+`POST /hooks/{company}/telegram` is the **optional hosted fast-path** for
+Telegram inbound, not the default. Telegram can only deliver to it from the
+public internet, so it is surfaced only when `OPENCOMPANY_PUBLIC_URL` is a
+public **https** URL (`AppConfig::public_webhook_base_url`); otherwise
+`GET …/channels/telegram` reports `webhookUrl: null` and
+`POST …/channels/telegram/webhook` is refused with `400`, rather than handing an
+operator a `http://127.0.0.1:<port>/hooks/…` URL that can never receive a
+delivery. Everywhere else — local and most self-hosted deployments — inbound
+arrives over `getUpdates` long-polling
+([`runtime::telegram_poller`](../runtime/README.md#background-listeners)) and a
+bot token is the whole setup: no webhook secret, no `setWebhook`, no public URL.
--- a/frontend/src/api/channels.ts
+++ b/frontend/src/api/channels.ts
@@ -4,22 +4,34 @@
 //
 // The bot token and webhook secret are write-only: they are sent on PUT and
 // stored in the host's secret store; neither is ever returned. The read shape
-// (`TelegramChannelStatus`) carries only presence booleans and the public
-// webhook URL. Mirrors `api/mcp.ts` — standalone functions over the shared
-// client, so no change to `OpenCompanyClient` is needed.
+// (`TelegramChannelStatus`) carries only presence booleans, how inbound
+// arrives, and the webhook URL when the host has a usable one. Mirrors
+// `api/mcp.ts` — standalone functions over the shared client, so no change to
+// `OpenCompanyClient` is needed.
+//
+// Issue #203: a bot token alone is a working channel. Inbound arrives over
+// `getUpdates` long-polling, which dials out and so needs no public URL; the
+// webhook is an optional hosted fast-path the host advertises (a non-null
+// `webhookUrl`) only when Telegram could actually deliver to it.
 
 import type { OpenCompanyClient } from "./client";
 
 /** The non-secret status of a company's Telegram channel. */
 export interface TelegramChannelStatus {
-  /** True once both the bot token and the webhook secret are stored. */
+  /** True once a bot token is stored — that alone is a working channel. */
   configured: boolean;
   /** Whether a bot token is stored (never the token itself). */
   tokenSet: boolean;
   /** Whether a webhook secret is stored (never the secret itself). */
   secretSet: boolean;
-  /** The URL to register with Telegram (`setWebhook`) / paste into BotFather. */
-  webhookUrl: string;
+  /**
+   * The URL to register with Telegram (`setWebhook`), or `null` when this host
+   * has no publicly reachable https URL — every local and most self-hosted
+   * deployment. Never show a webhook affordance while this is `null`.
+   */
+  webhookUrl: string | null;
+  /** Whether this host long-polls Telegram for inbound updates. */
+  polling: boolean;
 }
 
 /** The write-only config body. Only present, non-empty fields are applied. */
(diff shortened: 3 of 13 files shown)


oxoxDev (commented):
## Walkthrough

Replaces the undeliverable webhook URL with a per-company `getUpdates` long-poll that dials out, keeps the webhook route wired as an optional fast-path, and makes `configured` track the bot token alone. CI is green but builds default features only; I built this branch locally under `--features openhuman`.

The diagnosis is right and the fix is the correct shape — handing the operator a `http://127.0.0.1:8091/hooks/…` URL was never going to receive a delivery, and sharing `deliver_replies` between both inbound paths is what keeps them honestly identical. Three things.

## Actionable comments (3)

### Major

- `src/runtime/telegram_poller.rs:162` — the handshake **deletes a webhook it did not register**, automatically, based only on this host's own config. `webhook_capable` is `public_url` set to an https value; anything else means "not capable", and the poller then clears the registration and takes inbound over.

  That is a destructive write to state shared outside this process. Run the same bot token in two places — a staging box with `public_url` set and webhook mode working, plus any instance without it (a laptop, a container that lost its env, a self-hosted copy) — and the second one silently deletes the first's webhook and hijacks its DMs. The first host keeps running with no error anywhere; its inbound simply stops.

  Distinct bots per environment (`alphahumantest_bot` and friends) make this unlikely rather than impossible, and an operator who forgets `public_url` on a genuinely public host gets a silent downgrade to polling instead of a clear misconfiguration. Worth gating the delete behind an explicit opt-in — a config flag, or a console action — rather than doing it on the first tick.

### Minor

- `src/runtime/telegram_poller.rs:218` vs the module docs — they disagree. The docs say "the offset advances only after a batch has been processed, so a crash mid-batch redelivers that batch (at-least-once)"; the code advances the offset **per update, before** running the cycle, and the inline comment says so deliberately (a poison update must not strand the poller). Both behaviours are defensible, but the consequence the docs hide is that a cycle which returns `Err` is logged, skipped, and never retried — that update is at-most-once, not at-least-once. Since this is the delivery-semantics section a future reader will trust, it should describe the per-update pre-ack.

### Nit

- `src/runtime/telegram_poller.rs:260` — `tick()` is a `tokio::select!` branch racing `shutdown.notified()`, so shutdown can cancel it *mid-`run_cycle`*, aborting a company turn part-way. The doc note ("being dropped mid-tick is safe — an unacked batch is simply redelivered") covers the Telegram side, but not a cycle abandoned between its effects and its persistence. Draining the in-flight tick before breaking would make the claim complete.

Verified: keeping the offset in memory is the right call and the reasoning holds — Telegram owns the confirmed watermark, so a fresh process asking with no offset gets exactly what was never acked, and no local persistence can drift out of sync with it. Resetting the offset on a token change (offsets being per-bot) is the kind of detail that is normally found in production. `scrub_token` on every error path, including the `getUpdates` failure that becomes a returned `Store` error, keeps the token out of logs.

oxoxDev:
Rebased onto `main` (`f9e82f2`) to clear the conflict — the branch was based on `1ef0413`, before #211 and #234 landed. Force-pushed `9763776` → `adb402e`; the approval is intact (no branch rulesets, so no stale-review dismissal).

One real conflict, in `docs/modules/runtime/README.md`. It was purely additive — #211 added the `workflow_scheduler.rs` section and this branch added **Background listeners** — so both are kept, scheduler section first. `src/bin/opencompany.rs` and `src/runtime/mod.rs` auto-merged.

Verified locally on the rebased tip rather than trusting the clean auto-merge: `cargo check --features telegram --lib` clean, `cargo fmt --all -- --check` clean, `cargo test --features telegram --lib telegram` 21/21. CI is re-running now.

No behaviour change from the rebase — only the conflict resolution.

Issue #205, opened 2026-07-31
tasks: the working agent is never linked to the card's assignee; blank/invalid/overlay assignees silently fall back to the orchestrator with no validation


oxoxDev:
## Symptom
- A task assigned to **"Shane"** (not a roster agent) sat in in_progress; on the current build it silently dispatches to the **orchestrator (CEO)** instead, and the card keeps \`assignee="Shane"\` while the timeline reads "reply from ceo". No feedback that "Shane" isn't a valid agent.
- Tasks the CEO works keep \`assignee=""\` even though the CEO is doing the work — the working agent is never linked to the card.

## Root cause
\`task_responder\` (\`src/harness/brain.rs:427-435\`) resolves the assignee to an agent, but on ANY miss (blank \`""\`, unknown name like "Shane", an overlay-only teammate, or a desk id) it **silently falls back to \`self.responder\`** (the orchestrator). It then **never writes \`card.assignee\`** — all journaling attributes to the local \`responder\` var (\`AgentReply\`/\`DeskTaskCompleted\`/note attribution @300-328, 678-685), not to the card. So the card's \`assignee\` stays blank/invalid while a different agent works it.

Two aggravating factors:
- **Narrow roster lookup** — matches only \`self.record.manifest.agents\` (\`brain.rs:430\`), NOT \`overlay_agents\`, desks, or \`CompanyRecord::is_roster_agent\` (\`src/ports/types.rs:1247-1250\`). Operator-added teammates fall back too.
- **No column/assignee validation** — \`create_task\`/\`patch_task\` (\`src/server/ops/tasks.rs:193-263\`) accept any \`column\` and any \`assignee\` string verbatim; only the exact literal \`"in_progress"\` triggers dispatch (\`src/company/runtime.rs:32-39\`). A typo'd column or bad assignee is persisted silently.

## Desired
- On dispatch, **write the resolved working agent to \`card.assignee\`** so the board reflects who's actually working it.
- Resolve assignees against the FULL roster (manifest + overlay + desks via \`is_roster_agent\`).
- **Validate** assignee + column on write; reject or surface unknown values instead of silent fallback.

## Files
- \`src/harness/brain.rs:427-435\` (\`task_responder\`), dispatch/settle path @109/147/154-169
- \`src/ports/types.rs:1247-1250\` (\`is_roster_agent\`)
- \`src/server/ops/tasks.rs:193-263\` (create/patch — no validation)
[...]

Issue #185, opened 2026-07-30
feat(tasks): per-task event correlation + task-detail read API


oxoxDev:
## Problem

The Task Detail screen (and the lineage + timeline it renders) needs a **per-task event stream**, and nothing assembles one today:

- Events are journaled **company-scoped**. `CompanyEvent::TaskDispatched { task_id }` carries the id, but the events a dispatch then produces (tool calls, agent replies, waits) are **not tagged with `task_id`**, so they cannot be filtered back to their task.
- There is **no task-completion event** — the 14 `CompanyEvent` variants (`src/ports/types.rs`) have nothing for "this task finished", so a lineage rollup / timeline terminal has no anchor.
- There is **no read route** that assembles a per-task view. `GET …/tasks` lists board cards only; the timeline primitive (`AgentReply.steps: Vec<TurnStep>`, scrubbed, persisted, rehydrated) is keyed to a chat bubble, not a task.

## Proposal

Backend foundation for the Task Detail screen (#TASK_DETAIL_EPIC):

1. **Thread `task_id` onto dispatch-produced events** so a task's tool calls / replies / waits can be filtered out of the journal. Additive + `skip_serializing_if` so every already-persisted event round-trips byte-identical (same pattern as `OperatorMessage.by`/`chat`).
2. **Add `CompanyEvent::DeskTaskCompleted { task_id, desk, output }`** — the completion anchor the timeline terminal and lineage rollup need. (#171's done-transition can consume it; this issue only adds the event, not the column write.)
3. **`GET …/tasks/{id}`** read route that assembles: header fields, the timeline (journal filtered by `task_id`, reusing the scrubbed `TurnStep` shape — never raw args/output/ids), lineage (parent/children), and the approvals trail (from the `runtime/journal.rs` immutable ledger).

## Alternatives

- Derive the timeline purely client-side from chat history — rejected: a board-native card has no chat thread, and it loses task correlation entirely.
- One giant per-company event feed the frontend filters — rejected: leaks cross-task events, no server-side lineage.

## Scope / Related

Medium (backend only). Prereq for the Task Detail screen. Refs #176 (brain-agnostic seam / PR #178), #171 (done terminal consumes the completion event), #183 (MVP task board).

PR #213, opened 2026-07-31
fix(tasks): a dispatched task that delegates runs the delegate and links it as the assignee

### LINKED ISSUE #204: tasks: a dispatched task that delegates goes straight to in_review under the delegator — the delegate never runs or gets linked
## Symptom
A board task (\`Fetch my activity, from opencompany repo\`) that the **CEO delegated to a web-searcher** went straight to **in_review under the CEO**, with a blank \`assignee\`, and the web-searcher never ran. Observed timeline: \`dispatched → reply from ceo → completed on ceo → in_review\`. Expected: the delegate (web researcher) becomes the assignee, is shown working, and the card only moves to in_review **after the delegate produces output**.

## Root cause
\`run_task\` runs exactly **one** background turn (\`src/harness/brain.rs:147\`, \`run_steered_background\`) and **never drains the delegation queue** — confirmed by the comment at \`src/runtime/delegation.rs:286-288\` ("run_task never drains the delegation queue").

So when the dispatched responder (the orchestrator/CEO, which carries \`delegate_to_desk\`/\`spawn_task\` — \`orchestrator.rs:118-125\`) calls a delegation during the task turn, that \`Delegation\` is pushed onto the queue and **silently dropped** — nothing in \`run_task\` executes or clears it. The turn still returns \`Ok\` → \`TaskRunEnd::Completed\` (\`brain.rs:154-164\`) → \`settle\` → \`landing_column\`; the board card has no \`origin_chat_id\`, so \`success_terminal_column\` → **in_review** (\`lifecycle.rs:151-157\`). Card lands in_review under the delegator; the delegate never runs.

There is also **no lifecycle distinction between "I finished the work" and "I handed it off"** — \`TaskRunEnd\` (\`lifecycle.rs:67-81\`) only has \`Completed/Failed/Cancelled/Paused/RedirectsExhausted\`.

## Desired flow
When a dispatched task turn delegates to another agent:
1. Actually **execute** the delegation (drain the queue in the task path, like the operator-chat path does — \`delegation.rs:188-248\`).
2. **Reassign** the card to the delegate (write \`assignee\`) and show it working.
3. Keep the card in **in_progress** (a "delegated / awaiting delegate" state) — do NOT move to in_review yet.
4. Move to **in_review only after the delegate returns output**.

## Files
- \`src/harness/brain.rs\` — \`run_task\` (dispatch turn @147; settle @154-169), \`task_responder\` @427-435
- \`src/runtime/delegation.rs\` — \`run_task never drains the delegation queue\` @286-288; operator-path drain @188-248; \`delegate_to_desk\` (harness) @299-339 writes no card
- \`src/harness/lifecycle.rs\` — \`TaskRunEnd\` @67-81, \`landing_column\`/\`success_terminal_column\` @151-171

Found while UI-testing the consolidated build (e2e_harness co).

M3gA-Mind:
## Summary

Fixes #204.

A board task the CEO delegated to a web-searcher went straight to `in_review` under the CEO with a blank `assignee`, and the delegate never ran.

**Root cause.** `run_task` ran exactly one background turn and never drained the delegation queue. The dispatched responder is the orchestrator, which carries `delegate_to_desk` / `spawn_task`, so the `Delegation` its turn queued was pushed and silently dropped. The turn still returned `Ok`, so it settled as `Completed` → `landing_column` → `in_review` under the delegator, with nobody having done the work. `TaskRunEnd` also had no way to say "I handed it off" as distinct from "I finished it", so there was no state for a card whose work had only changed hands.

**The fix.**

- `TaskRunEnd::Delegated` lands a card in the new `COLUMN_IN_PROGRESS` — a hand-off is explicitly not a terminal.
- `DelegationRunner::handle_task_delegations` drains the queue in the task path. The first hand-off to a desk with a resolvable lead **owns the card**: it is reassigned to that lead and persisted in `in_progress` *before* their turn starts, so the board shows them working it while they work it. Every other delegation (`spawn_task`, `assign_task`, `review_task`, a hand-off to a desk nobody leads) executes for its side effect; a second hand-off's answer is appended to the note rather than dropped. The write goes through the `TaskStore` port, **not** `CompanyRuntime::upsert_task`, so it cannot re-fire the `column → in_progress` dispatch edge.
- `run_task` settles the card from the delegate's output and credits the delegate in every downstream write — note, artifact, journal, relay.
- The delegation queue is cleared before each dispatched turn, matching the operator path, so nothing a prior turn left behind can hijack a card.
- A `spawn_task` queued inside a dispatch now stamps `parent_task_id` with its parent card — the #185 lineage the old comment marked as unreachable *precisely because* the task path did not drain.

Two edge cases resolved deliberately: an **unresolvable desk** settles under the delegator exactly as before (no stranding in `in_progress` waiting on a delegate that can never run), and a hand-off **cancelled in flight** returns the card to `backlog` rather than reading as finished.

## API Or Behavior Changes

- **New `TaskRunEnd::Delegated` variant** and new `lifecycle::COLUMN_IN_PROGRESS` constant. `TaskRunEnd` is matched exhaustively in `landing_column`, which is pinned by a test, so a future variant still cannot be added without a deliberate column decision.
- **Behavior:** a dispatched turn that calls `delegate_to_desk` now runs a second turn (the delegate's) and writes the card twice — once mid-run to reassign it, once to settle it. Previously it ran one turn and wrote once. A dispatched turn that delegates nothing is byte-for-byte unchanged.
- **Behavior:** a card spawned by a dispatched turn now carries `parent_task_id`; one spawned from an operator chat turn still writes `None`.
- No REST/GraphQL schema changes.

## Tests

Both behaviour tests were verified to **fail** against the pre-fix path (drain stubbed back out) and pass after — the two compatibility tests pass either way, as they should.

New coverage:

- `a_dispatched_turn_that_delegates_runs_the_delegate_and_links_them_to_the_card` — the delegate actually runs (2 turns, not 1), is linked as `assignee`, the card reaches `in_review` on **their** output, and — via a per-invoke board snapshot — their turn ran against a card already reassigned and still `in_progress`.
- `a_dispatched_turn_that_delegates_nothing_settles_exactly_as_before` — the compatibility half.
- `a_hand_off_to_an_unknown_desk_settles_under_the_delegator` — no stranding.
- `a_task_spawned_by_a_dispatched_turn_records_its_parent_card` — #185 lineage.
- `a_hand_off_keeps_the_card_in_progress_rather_than_finishing_it` plus the extended exhaustive `landing_column` mapping, in both card shapes.

Commands run locally:

- [x] `cargo test` — 899 + 6 pass
- [x] `cargo test --features openhuman` — 1236 + 6 pass (the harness suite is where this code lives; the default build does not compile `src/harness/`)

Pre-existing and untouched by this PR: `--all-features` does not compile (`store/sqlite.rs` test method ambiguity), and clippy under `--features openhuman` flags `harness/composio.rs` + `harness/mcp_probe.rs`. Neither is in this diff and neither is in a CI lane.

## Documentation

No doc updates needed — the behaviour and its reasoning are documented in the rustdoc on `TaskRunEnd::Delegated`, `landing_column`, `handle_task_delegations`, and `hand_card_over`, which is where this codebase keeps its lifecycle decisions. The stale comment in `run_delegation` that asserted "`run_task` never drains the delegation queue" is rewritten in the same commit.

## Overlap with #205 (merge order)

#205 is a sibling bug in the same dispatch path and **touches two of the same files** — `src/harness/brain.rs` (`task_responder` and the dispatch/settle path) and `src/harness/lifecycle.rs`. Expect a conflict in `run_task` if both land.

This PR is written to be the *later* of the two:

- `task_responder` is **not touched here** — resolving blank/invalid/overlay assignees is #205's fix.
- The `assignee` write here happens only inside the hand-off branch, and lands *after* where #205's dispatch-time write would go. The delegate overriding the resolved responder is the correct final state, so the two writes compose rather than fight.
- The compatibility test here deliberately asserts **nothing** about `assignee` on the non-delegating path, so it will not pin the blank that #205 is fixing.

Either merge order works; if #205 lands first, the conflict in `run_task` is mechanical.

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -128,7 +128,12 @@ impl HarnessBrain {
         }
         // A blank assignee is the one legitimate miss: nobody was named, so the
         // orchestrator picks it up.
-        let responder = resolution
+        //
+        // Not final: a turn that hands the work off (issue #204) reassigns the
+        // card to the delegate, and from that point the delegate is the
+        // responder every downstream write credits — the note, the artifact,
+        // the journal, and the relay.
+        let mut responder = resolution
             .working_agent()
             .unwrap_or(&self.responder)
             .to_string();
@@ -186,6 +191,11 @@ impl HarnessBrain {
         // record exactly the text the note does rather than a second, divergent
         // rendering of the same run.
         let result_text = loop {
+            // Start each turn from an empty queue so nothing a prior turn (this
+            // cycle's operator message, or an earlier redirect rerun) left
+            // behind can hijack this card — the same guard
+            // `handle_operator_message` opens with.
+            self.deps.delegations.clear();
             let outcome = run_turn
                 // A dispatched task card carries no chat bubble (its steps are
                 // discarded into the note), so its live turn frames must not leak
@@ -199,13 +209,71 @@ impl HarnessBrain {
                     // A dispatched task discards its steps — the note is text-only.
                     match outcome {
                         Ok(outcome) => {
-                            // `settle` writes the note (attributed to the
-                            // assignee) and the landing column via the #186
-                            // lifecycle seam; the loop still yields the reply so
-                            // the #185/#190 completion events report the same
-                            // text that landed in the note.
-                            let result = outcome.reply;
-                            settle(&mut card, TaskRunEnd::Completed, &responder, &result);
+                            // Issue #204: the turn may have DELEGATED rather
+                            // than done the work. The dispatched responder is
+                            // the orchestrator, which carries `delegate_to_desk`
+                            // / `spawn_task`, and nothing here used to drain
+                            // what those queued — so the hand-off was dropped,
+                            // the turn still read as a clean completion, and the
+                            // card landed in `in_review` under the delegator
+                            // with the delegate never having run. Draining here
+                            // runs the delegate, reassigns the card to them, and
+                            // settles it from THEIR output.
+                            let handoff = self
+                                .delegation_runner(&run_turn)
+                                .for_task(&card.id)
+                                .handle_task_delegations(&mut card, &responder)
+                                .await?;
+                            // `settle` writes the note (attributed to whoever
+                            // actually produced the text) and the landing column
+                            // via the #186 lifecycle seam; the loop still yields
+                            // the reply so the #185/#190 completion events
+                            // report the same text that landed in the note.
+                            let result = match handoff {
+                                // The delegate answered: they own the card, and
+                                // every downstream write credits them.
+                                Some(handoff) => {
+                                    responder = handoff.delegate;
+                                    match handoff.reply {
+                                        Some(reply) => {
+                                            settle(
+                                                &mut card,
+                                                TaskRunEnd::Completed,
+                                                &responder,
+                                                &reply,
+                                            );
+                                            reply
+                                        }
+                                        // The hand-off ran but produced nothing
+                                        // (an operator cancelled it in flight).
+                                        // Partial work is discarded and the card
+                                        // returns to the backlog, exactly as a
+                                        // cancelled dispatch does — it must not
+                                        // read as finished, and it must not
+                                        // strand in `in_progress` either.
+                                        None => {
+                                            let reply =
+                                                "the delegated run was cancelled before it \
+                                                 produced anything"
+                                                    .to_string();
+                                            settle(
+                                                &mut card,
+                                                TaskRunEnd::Cancelled,
+                                                &responder,
+                                                &reply,
+                                            );
+                                            reply
+                                        }
+                                    }
+                                }
+                                // Nothing was handed off — the responder did the
+                                // work itself, as before.
+                                None => {
+                                    let result = outcome.reply;
+                                    settle(&mut card, TaskRunEnd::Completed, &responder, &result);
+                                    result
+                                }
+                            };
                             break result;
                         }
                         Err(err) => {
@@ -3101,6 +3169,19 @@ members = ["eng1", "eng2"]
         queue: orchestrator::DelegationQueue,
         pushes: StdMutex<VecDeque<Option<Delegation>>>,
         calls: std::sync::atomic::AtomicUsize,
+        /// The same task store the brain writes through, so each invoke can
+        /// snapshot the board **as that turn sees it** — the only way a test can
+        /// observe a dispatched card mid-run (issue #204).
+        tasks: Arc<FsOps>,
+        /// `(column, assignee)` of the company's card at each invoke, in order.
+        board: StdMutex<Vec<(String, String)>>,
+    }
+
+    impl DelegatingProvider {
+        /// The board snapshot each turn ran against, in invoke order.
+        fn board(&self) -> Vec<(String, String)> {
+            self.board.lock().unwrap().clone()
+        }
     }
 
     #[async_trait]
@@ -3111,6 +3192,15 @@ members = ["eng1", "eng2"]
             request: ModelRequest,
         ) -> tinyagents::Result<ModelResponse> {
             self.calls.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+            let snapshot = self
+                .tasks
+                .list(&CompanyId::new("acme"))
+                .await
+                .ok()
+                .and_then(|cards| cards.into_iter().next())
+                .map(|card| (card.column, card.assignee))
+                .unwrap_or_default();
+            self.board.lock().unwrap().push(snapshot);
             if let Some(Some(delegation)) = self.pushes.lock().unwrap().pop_front() {
                 self.queue.push(delegation);
             }
@@ -3140,10 +3230,13 @@ members = ["eng1", "eng2"]
         pushes: Vec<Option<Delegation>>,
     ) -> (HarnessBrain, Arc<DelegatingProvider>) {
         let queue = orchestrator::DelegationQueue::default();
+        let tasks = Arc::new(FsOps::new(dir));
         let provider = Arc::new(DelegatingProvider {
             queue: queue.clone(),
             pushes: StdMutex::new(pushes.into_iter().collect()),
             calls: std::sync::atomic::AtomicUsize::new(0),
+            tasks: tasks.clone(),
+            board: StdMutex::new(Vec::new()),
         });
         let deps = HarnessDeps {
             provider: provider.clone(),
@@ -3153,7 +3246,7 @@ members = ["eng1", "eng2"]
             meter: None,
             workspace_root: dir.to_path_buf(),
             model_override: None,
-            tasks: Some(Arc::new(FsOps::new(dir))),
+            tasks: Some(tasks),
             skills: None,
             skills_source_dir: None,
             mcp_servers: Vec::new(),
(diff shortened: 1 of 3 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (commented):
## Walkthrough

`run_task` now drains the delegation queue, so a dispatched turn that hands off actually runs the delegate, reassigns the card to them mid-flight, and settles from *their* output. New `TaskRunEnd::Delegated` keeps a hand-off out of the terminal columns. CI is green but builds default features only, so none of this file is compiled by it — I union-built `main + #212 + #213` and ran the gates locally: `cargo check --features openhuman` clean, 272 harness + 82 runtime tests pass.

## Actionable comments (3)

### Major

- `src/harness/brain.rs:178` — the delegate's run is propagated with `?`, so an errored hand-off strands the card in exactly the state this issue exists to eliminate. `hand_card_over` has already persisted the card as `in_progress` reassigned to the delegate (`delegation.rs:364`); if `run_delegation` then returns `Err`, `?` unwinds out of `run_task` before the settle and before the final `upsert`. The card sits in `in_progress` under the delegate with no result, and nothing re-dispatches it — `task_enters_in_progress` only edge-fires on the *transition* into that column, and the card is already there. Note the non-delegating path one arm down (`:232`) handles this properly: `Err` → `settle(Failed)` → `backlog`. The hand-off path wants the same arm.

### Minor

- `src/runtime/delegation.rs:332` — a `None` `desk_reply` is reported to the operator as `"the delegated run was cancelled before it produced anything"`, but the code cannot actually distinguish a cancellation from a delegate whose turn simply yielded nothing. If a non-cancel path can produce `None`, the card settles `Cancelled` with a message that names a cause that did not happen. Worth either narrowing the claim or asserting the invariant.
- `src/runtime/delegation.rs:327` — when the *first* hand-off yields `None`, `handoff` is still `Some { reply: None }`, so a later successful hand-off in the same drain takes the `owns_card == false` arm: its reply is appended to the note, but the card still settles `Cancelled` → `backlog` off the first one. Work that ran and produced output ends up filed under a card marked cancelled.

## Questions

- What happens when `desk_lead` resolves to the delegator itself — a desk whose lead *is* the orchestrator? As written the card is "handed over" to the agent that already had it, the note records `delegated to <self>`, and the turn runs again. Harmless in one pass, but it reads as a hand-off in the timeline when nothing changed hands.
- `src/harness/brain.rs:151` — the per-turn `delegations.clear()` is on the queue shared through `HarnessDeps`. Within a cycle that carries both an `OperatorMessage` and a `TaskDispatched` event, is the operator handler guaranteed to have drained before the task handler clears? The comment says it mirrors that guard, so I assume yes — confirming the ordering is structural rather than incidental.

Verified: the `DelegatingProvider::board()` snapshot is the sharp part of the test design — asserting the delegate's own turn ran against a card *already* reassigned and still `in_progress` is what makes "shown working it while they work it" a real invariant rather than an end-state coincidence. `a_dispatched_turn_that_delegates_nothing_settles_exactly_as_before` deliberately not pinning `assignee` (left to #205) is the right call.

oxoxDev:
Picking up the three actionable items from the review above so this isn't blocked on you — shout if you're already on it and I'll back off. Will push to this branch and reply per finding.

oxoxDev:
All three addressed — pushed to this branch, no rebase.

**MAJOR — errored hand-off strands the card** → `a0e0bc2`
The `?` didn't just skip the settle, it failed the whole cycle: `run_cycle` returned `Err` and the card stayed `in_progress` under the delegate with no result. Now takes the same arm the non-delegating path does — `settle(Failed)` → `backlog`, reason on the note, attributed to the delegator (not `operator`; a cancellation is the only ending recorded as the operator's act). The delegate stays the assignee on the way back, since the hand-off did happen and a re-dispatch should start from who it was given to. Regression test makes the delegate's provider error and asserts `backlog` + the reason; it fails pre-fix with the cycle erroring out. It fails *from* an invoke rather than *on* one — openhuman's agent loop retries a failed provider call inside the same turn, so failing a single invoke just succeeds on the retry.

**MINOR — `None` desk_reply reported as cancellation** → `da6ba5b`
You asked which was true: the invariant does hold. `run_delegation`'s `DelegateToDesk` arm has exactly two empty exits — an unresolvable desk and a mid-flight cancel — and the first is unreachable here because the lead resolved before the run and `desk_lead` is pure over the same record. So today it genuinely can only be a cancellation. But that's an argument spanning two functions, not something the code knows, so I made it carry the fact: `DelegationOutcome::cancelled` is set by the cancel exit, and a reply-less `TaskHandoff` is only built from that flag. Not a behaviour change. If the other exit ever becomes reachable it now reports *no hand-off*, so the delegator's own reply settles the card exactly as an unresolvable desk already does, instead of inventing a cancellation. Test: a cancelled hand-off still lands in `backlog` reported as a cancellation, attributed to the operator.

**MINOR — later hand-off filed under a cancelled card** → `e7c6d5d`
The card now belongs to the first hand-off that actually *produces* something. An empty one holds it only provisionally and a later one that answers takes it over — reassigned to that delegate, settled from their output. Once a hand-off has answered, every further one behaves exactly as before. Test: two hand-offs in one drain, first cancelled, second answering; card reaches `in_review` under the delegate with their output and no cancellation block. Fails pre-fix with the card in `backlog`.

---

**Q: `desk_lead` resolving to the delegator itself.**
Nothing structural breaks, but it's worse than a confusing timeline entry — it's a wasted turn and a discarded answer:
- No loop. `queue.drain(cap)` snapshots up front and clears the remainder, so anything the "delegate" turn queues isn't drained by that loop, and both `run_task` and `handle_operator_message` clear on entry. Bounded to exactly one extra turn.
- No re-dispatch. `hand_card_over` writes through `TaskStore::upsert`, not `CompanyRuntime::upsert_task`, so no `column → in_progress` edge fires — and the card is already there.
- The real cost: the same agent runs twice, and the second run's prompt is the `delegate_to_desk` `instruction`, not `task_instruction(&card)`. The card then settles from that second reply, so the delegator's own first-turn output is thrown away in favour of a re-roll against a narrower brief. Doubled tokens, possibly worse answer.

Left as-is since it's out of scope for this PR, but the cheap guard is to treat a lead that equals the delegator as no hand-off at all — drop straight into the unresolvable-desk arm and let the delegator's own reply settle the card. Happy to fold it in here or file it separately, your call.

**Q: `delegations.clear()` ordering with a mixed `OperatorMessage` + `TaskDispatched` cycle.**
Structural, on three counts — and the safety doesn't actually rest on the ordering:
1. `run_cycle` iterates `req.events` with a sequential awaited `for` loop. No fan-out, so a handler runs to completion before the next event's does.
2. `handle_operator_message` *ends* on `DelegationQueue::drain`, which takes up to the cap and clears the remainder — it returns with the queue empty. There's nothing left for `run_task`'s `clear()` to destroy. Desk members carry no delegation tools, so nothing gets pushed after the drain either.
3. Cycles never overlap per company — `CycleRunner::run` holds the per-company serial lock for the whole cycle (there's a `cycles_are_serial_per_company` test). Workflow runs build their own `DelegationQueue` rather than sharing the brain's.

And it's belt-and-braces: both entry points clear on entry, so the reverse event order (`TaskDispatched` first) is equally safe. The one thing that *is* load-bearing and unguarded: it's a property of the loop shape, not the types. If `run_cycle` ever fanned events out concurrently, a dispatched card's `clear()` would wipe an operator turn's queued delegations mid-flight, and nothing would catch it.

---

Gates (CI builds default features and never compiles this code, so all run locally with `--features openhuman`):
- `cargo check --features openhuman` — clean (2 pre-existing `dead_code` warnings in `harness/composio.rs`, untouched)
- `cargo test --features openhuman --lib harness::brain::` — 47 passed, 0 failed
- `cargo test --features openhuman --lib harness::lifecycle::` — 13 passed, 0 failed
- `cargo test --features openhuman --lib runtime::` — 75 passed, 0 failed
- `cargo clippy --features openhuman --all-targets` — no new warnings (3 pre-existing, all in files this branch doesn't touch)

Both new behaviour tests were verified to fail against the pre-fix code, not just pass against the new.

oxoxDev:
Heads-up @M3gA-Mind — **I rebased this branch onto `upstream/main` and force-pushed** (`e7c6d5d` → `716ba8f`). #214 landed as `eb6fb98` and put the PR in `CONFLICTING`; it's `MERGEABLE` again. Re-pull before you touch it: `git fetch && git reset --hard m3g/fix/204-delegate-never-runs`.

Your commit kept its authorship. Two conflicts:

**`src/harness/lifecycle.rs`** — #214 moved the column vocabulary to `ports::tasks` and re-exports it (this module is `openhuman`-gated, so the REST write boundary couldn't see the constants). Your `COLUMN_IN_PROGRESS` was a local `const`. Resolved by dropping the local const and adding `COLUMN_IN_PROGRESS` to the re-export list — `ports::tasks` already defines it — with your doc sentence folded into the re-export's comment. Everything else of yours (`TaskRunEnd::Delegated`, the `landing_column` arm, the `relay_text` arm, the tests) is additive and kept as-is.

**`src/harness/brain.rs`** — the real one. #214 rewrote the head of `run_task` from a pre-#204 base, so main's `run_task` has **no delegation drain at all** and `task_responder` is gone. Kept #214's block whole — `assignee::resolve`, the `rejection()` refusal, and the `links_working_agent()` write-back — and took only `mut` on the responder binding from your side, since the hand-off still reassigns it:

```rust
let mut responder = resolution.working_agent().unwrap_or(&self.responder).to_string();
```

Nothing of mine or yours read the deleted `task_responder`; `hand_card_over` uses `desk_lead` directly, which is what `assignee::resolve` itself calls. The rest of your commit auto-merged. Post-rebase my three fixes diff byte-identically to pre-rebase (334+/37−).

One thing worth a look, not changed: `hand_card_over` writes the desk **lead** to `card.assignee`, while #214's rule is that a desk assignment stays a desk (`links_working_agent()` is false for `Desk`). I read these as different situations — #214 governs the write-back of a card's *existing* desk assignee, whereas a `delegate_to_desk` hand-off is an explicit ownership transfer to a person, which is what #204 asks for and what your test asserts. So I left it. All of #214's assignee tests pass alongside it, including `task_dispatch_routes_a_desk_assignee_to_its_lead_but_keeps_the_desk`. Flagging in case you meant the other reading.

Gates on the rebased branch (CI builds default features and compiles none of this):
- `cargo check --features openhuman` — clean
- `cargo clippy --features openhuman --all-targets` — no new warnings
- `harness::brain::` **53 passed** (was 47; +6 from #214), `harness::lifecycle::` **13**, `runtime::` **110** (was 75) — 0 failed

I re-verified both behaviour tests still **fail** with their fix reverted on the rebased tree, so they're passing for the right reason and not because a resolution quietly dropped one.

Unrelated, pre-existing on main: `cargo clippy --features openhuman` flags `redundant_guards` at `src/runtime/delegation.rs:575` (`Some(canonical) if canonical.is_empty()` → `Some("")`), verbatim from `eb6fb98` and untouched here. Means main isn't clippy-clean under `--features openhuman` right now. Left it alone to keep this diff scoped — say the word and I'll fold in the one-liner.

Issue #171, opened 2026-07-29
fix(tasks): dispatched tasks stick in "In review" and never reach "Done" — no code path writes the done column


oxoxDev:
## Summary
When an orchestrator delegates work to a teammate via `spawn_task` (e.g. a "Backend Engineer status request"), the card runs to completion, the assignee's reply is folded into the card note as an agent-to-agent handoff block, and the card lands in the **In review** column. It stays there forever. Nothing in the harness ever advances a dispatched card to **Done** — the only writes of the `done` column in the whole backend are a manual operator drag on the board and test/sample fixtures.

## Reproduction
1. In a chat thread, the orchestrator calls `spawn_task` (title "Backend Engineer status request", assignee = the engineer desk/agent).
2. `run_delegation` opens a `backlog` card stamped with `origin_chat_id` (`src/harness/brain.rs:361-374`).
3. The card enters `in_progress`; this edge fires the dispatch cycle (`src/company/runtime.rs:326-338`).
4. The `TaskDispatched` cycle runs the assignee turn (`src/harness/brain.rs:522-523` → `run_task`), the reply is appended to the note (`src/harness/brain.rs:143-148`), and the card is set to `in_review` and upserted (`src/harness/brain.rs:148,215`).
5. Per #163 a post-back bubble is sent to the origin thread (`src/harness/brain.rs:228-236`).
6. Observe the board: the card sits in **In review** indefinitely. No agent, cycle, or event ever moves it to **Done**.

## Root cause
The dispatched-task state machine has no terminal `done` transition. Every status mutation in `run_task` writes a non-terminal column:
- success → `card.column = "in_review"` (`src/harness/brain.rs:148`)
- redirect-budget exhausted → `in_review` (`src/harness/brain.rs:200`)
- failure → `backlog` (`src/harness/brain.rs:156`); cancel → `backlog` (`:169`); pause → `paused` (`:182`)

[...]

PR #215, opened 2026-07-31
feat(tasks): To-do column and a To-do-only add button (#206)

### LINKED ISSUE #206: task board: add a To-do column, restrict the + add-task button to To-do only, and auto-assign+link an agent on move-to-in_progress
## Ask (from product)
Three changes to the task board:

1. **Add a "To-do" column.** Keep Backlog; add To-do as a distinct column. (Current columns: \`backlog, in_progress, paused, in_review, done\` — \`frontend/src/lib/tasks-sample.ts:21-28\`, backend constants \`src/harness/lifecycle.rs:47-56\`.) Confirm the intended order/semantics of Backlog vs To-do with product (this issue assumes both exist; new tasks land in To-do).

2. **Restrict the "+" add-task button to the To-do column only.** Today it renders on **every** column unconditionally — \`frontend/src/views/TasksView.tsx:233-239\` (inside the \`TASK_COLUMNS.map\`). Only To-do should show it.

3. **On move-to-in_progress, the orchestrator assigns AND links an agent.** When a card enters in_progress with no explicit (or invalid) assignee, the orchestrator should pick a valid agent, **write it to \`card.assignee\`** (link it), and dispatch — the working agent must be visible on the card. Today the dispatch falls back to the orchestrator silently and never writes \`assignee\` (see #205). The in_progress edge fires at \`src/company/runtime.rs:32-39\` → \`dispatch_task\` @361-385 → \`run_task\`.

## Notes
- Adding a column touches: \`TASK_COLUMNS\` (frontend), any backend column constants + the create default (\`src/server/ops/tasks.rs:211\` defaults to \`backlog\` → likely \`todo\`), GraphQL/docs (\`src/server/graphql/tasks.rs:21\`, \`schema.graphql:651\`).
- Part 3 overlaps the fix in #205 (assignee linking) — coordinate.

Found while UI-testing the consolidated build (e2e_harness co).

M3gA-Mind:
## Summary

Closes #206. Rebased onto `main` after #214 landed; two commits.

### A To-do column, and it is the board's only entry point — `2738603`

New work had nowhere of its own to land. `backlog` served double duty — the unqueued pool *and* where the lifecycle returns work that needs another pass (a failed dispatch, an orchestrator `revise` verdict) — so "nobody has picked this up" and "someone just asked for it" sat in one column.

The issue asked to confirm Backlog vs To-do semantics with product. Resolved as the issue itself assumed, and now written into `docs/spec/runtime/ports.md`:

- **Backlog** — the unqueued pool, and the lifecycle's return column. Unchanged; every path that places a card there still does.
- **To-do** — queued next, sitting between Backlog and In progress. `POST …/tasks` defaults here, while an explicit `column` still wins, so `spawn_task`, `revise`, and a failed run keep placing their own card.

The `+` button now renders on To-do alone (`ADD_TASK_COLUMN`). That is not only tidiness: the old unconditional button let an operator create a card straight into `in_progress` — skipping the dispatch edge, since only the *transition* into that column fires it — or into `in_review`/`done`, fabricating a terminal for work that never ran.

### The column-order test says what it actually pins — `e586b43`

`columns_are_ordered_and_unique` claimed it existed because the console mirrors the list by hand, but it only asserts the Rust array against a Rust literal — it cannot read `tasks-sample.ts`, so the very drift it named as its reason keeps it green. The comment now states that it pins the Rust order only, that cross-language drift is unguarded, and that closing the gap needs codegen this crate has no build step for.

### Reconciliation with #214

This PR originally carried a second slice that resolved and wrote back a card's `assignee` on dispatch. #214 (#205) owns that, and now that it has merged, **this PR no longer touches `src/harness/brain.rs` at all**. Two things were reconciled rather than landed twice:

- **One owner for assignee resolution.** The `task_responder` extension here was dropped. It resolved a *desk* assignee to the desk's lead and the dispatch wrote that back, so a card the operator assigned to the Strategy desk silently became assigned to `engineer` on first dispatch and the desk assignment was gone from the board. `runtime::assignee::resolve` on `main` keeps desk ownership intact — its write-back is gated on `links_working_agent()`, true only for an unassigned card or a teammate assignment — and it *refuses* an assignee that names nobody instead of routing it to the orchestrator.
- **One column list, including `todo`.** Both PRs hoisted the column literals into `ports::tasks` under different names and different membership: `TASK_COLUMNS` with six entries here, `BOARD_COLUMNS` with five and no `todo` on #214, wired into `validate_column` at the REST write boundary. Each was green alone; together they were a live bug — `POST …/tasks` defaults a new card to `todo` while `validate_column` rejects `todo` as not a board column, so **every card creation 400s**. Confirmed by building the union locally before #214 merged: dropping `todo` from the list fails 5 write tests, `created_tasks_default_to_the_todo_column` and `tasks_crud_round_trips_under_both_scopes` among them. This PR now adds `COLUMN_TODO` to `main`'s `BOARD_COLUMNS` and keeps `is_board_column` as the single predicate, so the validator and the create default agree.

## API Or Behavior Changes

- **`POST …/tasks` with no `column` now returns `todo`, not `backlog`.** A client relying on the old default will see new cards in a new column. An explicit `column` is unaffected, and stored boards need no migration.
- **`todo` is a new value of `Task.column`** on both the REST and GraphQL read surfaces (SDL snapshot regenerated). Columns are now `backlog | todo | in_progress | paused | in_review | done`, and `todo` is accepted by the `validate_column` write boundary added in #214.
- The `+` add-task button no longer appears on Backlog, In progress, Paused, In review, or Done.
- No change to assignee resolution, dispatch, or the harness — that surface is entirely #214's.

Deliberately **not** changed: agent-spawned cards (`spawn_task`), operator-chat-opened cards (`operator.rs`), and the lifecycle return paths all still land in `backlog`. The issue scoped the change to the create default at `server/ops/tasks.rs`, and those three are documented #171/#186 decisions. Happy to move the operator-chat one to To-do as a follow-up if product wants it.

## Tests

2 new, 1 updated.

- `server::ops::write_test` — `created_tasks_default_to_the_todo_column` (the default, and that an explicit column still wins); `tasks_crud_round_trips_under_both_scopes` updated for the new default. Both now run in the same binary as #214's `task_writes_reject_a_column_the_board_cannot_render`, which is the union that would have 400'd every card creation had the lists not been reconciled.
- `ports::tasks` — `columns_are_ordered_and_unique` (pins the Rust order; see the commit note on what it does *not* pin), `is_board_column_accepts_only_board_columns` extended to cover `todo` and near-miss typos.

Commands run locally, on the rebased branch:

- [x] `cargo test` (956 passed, 0 failed)
- [x] `cargo test --features openhuman` (1294 passed, 0 failed)
- [x] `cargo check --features openhuman`
- [x] `npm ci && npm run typecheck && npm run build` (frontend; this package has no lint or test script)
- [ ] `cargo clippy --all-targets --features openhuman -- -D warnings` — **fails on `main` too**, in vendored `tinyagents` (`vendor/tinyagents/src/harness/agent_loop/stream.rs:48`, `large size difference between variants`). This PR touches no vendored file.

## Documentation

- `docs/spec/runtime/ports.md` — the `TaskStore` section now lists `todo` in the column set and explains the Backlog-vs-To-do split and To-do's role as the single manual-entry column. #214's assignee-resolution paragraph is left as merged; it is the authority on that behaviour.
- Module docs for the column vocabulary live with the code in `src/ports/tasks.rs`.

### DIFF
--- a/docs/spec/runtime/ports.md
+++ b/docs/spec/runtime/ports.md
@@ -335,11 +335,20 @@ pub trait TaskStore: Send + Sync {
 `TaskRecord` carries `{id, title, note, column, priority, assignee,
 updated_at}`.
 
-`column` ∈ `backlog|in_progress|paused|in_review|done` — the `BOARD_COLUMNS`
+`column` ∈ `backlog|todo|in_progress|paused|in_review|done` — the `BOARD_COLUMNS`
 constant in `src/ports/tasks.rs`, which is the one authority the REST write
-boundary, the dispatch edge and the harness lifecycle seam all read. (`paused`
-arrived with steering, issue #111; this line used to omit it.) Entering
-`in_progress` is what dispatches the card; nothing dispatches out of `done`.
+boundary, the dispatch edge and the harness lifecycle seam all read, and which
+the console mirrors in the same order. (`paused` arrived with steering, issue
+#111; this line used to omit it.) Entering `in_progress` is what dispatches the
+card; nothing dispatches out of `done`.
+
+`backlog` and `todo` are both "not started", and the split is deliberate:
+`backlog` is the unqueued pool — and where the lifecycle returns work that needs
+another pass (a failed dispatch, an orchestrator `revise` verdict) — while
+`todo` is what has been queued up next. `todo` is the board's one manual-entry
+column: the console's `+` button lives there alone and `POST …/tasks` defaults
+to it (issue #206), so an operator cannot create a card straight into
+`in_progress` or a terminal column.
 
 `assignee` names a **roster teammate id, a desk, or nobody** (`""`), resolved by
 `crate::runtime::assignee` against the full roster — manifest agents, operator
--- a/frontend/src/lib/tasks-sample.ts
+++ b/frontend/src/lib/tasks-sample.ts
@@ -18,14 +18,36 @@ export interface TaskColumn {
   label: string;
 }
 
+/**
+ * The board's columns, left to right. Must stay in step with the backend's
+ * `BOARD_COLUMNS` (`src/ports/tasks.rs`), which is the source of truth — and
+ * which the REST write boundary rejects a card against, so a column that
+ * exists only here is one the host will refuse.
+ */
 export const TASK_COLUMNS: TaskColumn[] = [
+  // The unqueued pool, and where the lifecycle returns work that needs another
+  // pass (a failed dispatch, an orchestrator `revise` verdict).
   { id: "backlog", label: "Backlog" },
+  // Queued to be worked next (issue #206). The board's manual-entry column:
+  // `ADD_TASK_COLUMN` below restricts the `+` button to it, and it is what the
+  // host's `POST …/tasks` defaults to.
+  { id: "todo", label: "To-do" },
   { id: "in_progress", label: "In progress" },
   // Where a steered-to-pause run parks (issue #111); a card here offers Resume.
   { id: "paused", label: "Paused" },
   { id: "in_review", label: "In review" },
   { id: "done", label: "Done" },
 ];
+
+/**
+ * The one column that offers the "+" add-task button (issue #206).
+ *
+ * New work enters the board in exactly one place. Offering `+` on every column
+ * — as the board used to — let an operator create a card straight into
+ * `in_progress`, `in_review`, or `done`, which either skips the dispatch edge
+ * or fabricates a terminal state for work that never ran.
+ */
+export const ADD_TASK_COLUMN = "todo";
 
 const STRATEGY = { name: "Strategy desk", tone: "sky" };
 const CREATIVE = { name: "Creative studio", tone: "violet" };
--- a/frontend/src/views/TasksView.tsx
+++ b/frontend/src/views/TasksView.tsx
@@ -26,7 +26,7 @@ import {
 import { Skeleton } from "@/components/ui/skeleton";
 import { Textarea } from "@/components/ui/textarea";
 import { cn } from "@/lib/utils";
-import { PRIORITY_STYLES, TASK_COLUMNS } from "@/lib/tasks-sample";
+import { ADD_TASK_COLUMN, PRIORITY_STYLES, TASK_COLUMNS } from "@/lib/tasks-sample";
 import { toast } from "sonner";
 import { TaskDetailView } from "./TaskDetailView";
 
@@ -230,13 +230,16 @@ export function TasksView({
                   <span className="text-sm font-medium">{col.label}</span>
                   <span className="text-xs text-muted-foreground">{items.length}</span>
                 </div>
-                <button
-                  className="text-muted-foreground hover:text-foreground"
-                  aria-label={`Add task to ${col.label}`}
-                  onClick={() => setCreatingIn(col.id)}
-                >
-                  <Plus className="size-4" />
-                </button>
+                {/* New work enters the board in one place only (issue #206). */}
+                {col.id === ADD_TASK_COLUMN && (
+                  <button
+                    className="text-muted-foreground hover:text-foreground"
+                    aria-label={`Add task to ${col.label}`}
+                    onClick={() => setCreatingIn(col.id)}
+                  >
+                    <Plus className="size-4" />
+                  </button>
+                )}
               </div>
               <div className="flex min-h-0 flex-1 flex-col gap-2 overflow-y-auto px-2 pb-2">
                 {loading && items.length === 0 ? (
--- a/src/ports/tasks.rs
+++ b/src/ports/tasks.rs
@@ -139,10 +139,16 @@ pub trait TaskStore: Send + Sync {
 mod test {
     use super::*;
 
-    /// The board's columns, in the order the console renders them. Pinned
-    /// because `frontend/src/lib/tasks-sample.ts` mirrors this list by hand: a
-    /// column added here and not there (or vice versa) files cards into a
-    /// column the operator cannot see.
+    /// Pins the **Rust** list's ids and their order against a literal, so a
+    /// reorder or a rename is a deliberate two-place edit rather than a
+    /// side effect.
+    ///
+    /// It does **not** protect against drift from the console's mirror in
+    /// `frontend/src/lib/tasks-sample.ts` — a Rust test cannot see the TS list,
+    /// so a column added on one side and not the other keeps this green.
+    /// Closing that gap means generating one list from the other (a build step
+    /// this crate does not have, across a separate npm build), so for now the
+    /// mirror is maintained by hand and the two lists are reviewed together.
     #[test]
     fn columns_are_ordered_and_unique() {
         assert_eq!(
--- a/src/ports/tasks.rs
+++ b/src/ports/tasks.rs
@@ -1,8 +1,8 @@
 //! The [`TaskStore`] port: the company's durable Kanban board.
 //!
-//! Tasks are the operator-visible work items the console's board renders
-//! (backlog / in-progress / in-review / done). They are hand-curated state, not
-//! cycle working memory — the brain's per-cycle task results live in
+//! Tasks are the operator-visible work items the console's board renders (see
+//! [`BOARD_COLUMNS`]). They are hand-curated state, not cycle working memory —
+//! the brain's per-cycle task results live in
 //! [`MemoryStore`](crate::ports::MemoryStore). Each record is keyed by a stable
 //! id within the company.
 
@@ -23,8 +23,13 @@ use crate::ports::types::CompanyId;
 // unchanged, and `CompanyRuntime`'s dispatch edge reads `COLUMN_IN_PROGRESS`
 // from here, so each literal exists in exactly one place.
 
-/// Where new work waits, and where a failed, cancelled or revised run returns.
+/// The unqueued pool: cards nobody has committed to yet, plus the ones a failed,
+/// cancelled or revised run returns.
 pub const COLUMN_BACKLOG: &str = "backlog";
+/// Queued to be worked next (issue #206). This is the board's manual-entry
+/// column — the `+` button lives here alone, and it is what `POST …/tasks`
+/// defaults to.
+pub const COLUMN_TODO: &str = "todo";
 /// The dispatch column: entering it hands the card to its assignee.
 pub const COLUMN_IN_PROGRESS: &str = "in_progress";
 /// Where a steered-to-pause run parks. Resume is a `column → in_progress` PATCH.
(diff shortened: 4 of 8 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (commented):
## Walkthrough

Adds a `todo` column as the board's only manual entry point, hoists the column vocabulary into the always-compiled `ports::tasks`, and links the working agent to the card on dispatch. CI is green but builds default features only, so the `brain.rs` half is not compiled by it; I built this branch locally under `--features openhuman`.

The To-do slice is clean and well-argued — the Backlog-vs-To-do semantics are settled in the spec rather than left implicit, and the `+`-button restriction closes a real hole (creating a card straight into `in_review` or `done` fabricates a terminal state for work that never ran). My concerns are all about how this lands next to #213 and #214.

## Actionable comments (4)

### Major — cross-PR

- **This PR and #214 both implement "link the working agent to the card", in the same function, two different ways — and #215's version reinstates the bug #214 exists to fix.** `brain.rs:480` extends `task_responder` to resolve desks and the overlay roster, then `:126` writes the result back. #214 replaces the same call site with `assignee::resolve`, which additionally *refuses* an assignee that names nobody. `task_responder_resolves_desks_and_the_full_roster` asserts `task_responder("Shane") == "ceo"` — the exact silent misroute #205 is about. So if #214 merges first and this one's `task_responder` wins the conflict resolution, #205's fix is quietly reverted with all its tests still passing, because they test the resolver module rather than this call site. Worth deciding explicitly which PR owns assignee resolution and dropping the other's half.

- **The column constants collide with #214's, and the combination breaks card creation.** Both PRs add a column list to `src/ports/tasks.rs`: this one `TASK_COLUMNS` (6, including `todo`) with `is_known_column`; #214 `BOARD_COLUMNS` (5, no `todo`) with `is_board_column`, wired into a `validate_column` on the write boundary. Land both and `POST …/tasks` defaults every new card to `todo` (`ops/tasks.rs:219`) while `validate_column` rejects `todo` as not a board column — **every card creation 400s**. Each PR is green alone; the failure only exists in the union. Whichever lands second needs to reconcile to one list under one name.

### Major

- `src/harness/brain.rs:126` — the write-back overwrites a **desk** assignment with the desk's lead. `task_responder("strategy")` returns `engineer`, so a card the operator assigned to the Strategy desk silently becomes assigned to `engineer` on first dispatch and the desk assignment is gone from the board. Desk ownership and "who ran this turn" are different facts; only the resolved-from-blank case wants writing back. (#214 has the identical problem from the other direction — its `canonical()` explicitly promises a desk assignment stays a desk assignment, and then its dispatch path breaks that promise.)

### Nit

- `src/ports/tasks.rs:47` — `columns_are_ordered_and_unique` says it exists because `frontend/src/lib/tasks-sample.ts` mirrors the list by hand, but the test can only see the Rust side: it pins this list against a literal, so the two can still drift apart and the test stays green. The comment promises drift protection the assertion cannot deliver. Either generate the TS list, or say plainly that it pins the Rust order only.

Verified: routing `company::runtime`'s dispatch edge at the re-exported `COLUMN_IN_PROGRESS` rather than its own literal is the right consolidation — the transition-not-state trigger meant a typo there failed silently, and there is now one definition. Re-exporting from `lifecycle` so existing `lifecycle::COLUMN_*` call sites are untouched keeps the diff honest. `is_known_column`'s near-miss cases (`to_do`, `To-do`, `inprogress`) are the right things to test.

oxoxDev:
### Heads-up: #214 and #215 do the same job twice

Flagging across both threads so whichever gets picked up first sees it — these two overlap in a way neither PR can see on its own.

**1. Both link the working agent to the card.** #214 replaces the `task_responder` call site in `run_task` with the new `assignee::resolve`; #215 extends `task_responder` itself and writes the result back at the same spot. Same behaviour, two implementations, same function.

They are not equivalent: #215's version keeps the silent fallback — `task_responder_resolves_desks_and_the_full_roster` asserts `task_responder("Shane") == "ceo"` — which is exactly the misroute #205 exists to fix. So if #214 lands first and #215's version wins the conflict resolution, #205's fix is reverted **with all its tests still green**, because they cover the resolver module rather than the call site.

**2. Both add a column list to `src/ports/tasks.rs`, and together they break card creation.** #214 adds `BOARD_COLUMNS` (5 entries, no `todo`) + `is_board_column`, wired into a `validate_column` on the write boundary. #215 adds `TASK_COLUMNS` (6 entries, with `todo`) and defaults `POST …/tasks` to `todo`. Land both and every new card is created in a column the validator rejects — **`POST …/tasks` 400s on every request**. Each PR is green alone; the failure exists only in the union.

**What would help:** decide which PR owns assignee resolution and drop the other's half, and reconcile the two column lists to one name before the second one rebases. Otherwise the second rebase silently undoes part of the first.

(Also worth knowing: all three of #213/#214/#215 were cut from the same `main` tip and are not stacked, so they report MERGEABLE individually while conflicting with each other — #214 vs #215 collide on four files. Suggested order is #212 → #213 → #214 → #215.)

oxoxDev:
Picking up the review items here too — shout if you're already on it.

Flagging the ordering up front: two of the findings are cross-PR with #214, and the union of the two as they stand breaks card creation (this PR defaults new cards to `todo`, #214's `validate_column` rejects `todo` as not a board column — each is green alone). So this one reconciles **onto** #214 rather than in parallel: #214 keeps ownership of assignee resolution (`assignee::resolve`, which refuses an assignee naming nobody), and the column vocabulary lands as one list under one name. Detail in the commits.

oxoxDev:
All four findings addressed. Pushed `5274fb7`, `5b919e7`, `4cb8da6` — no rebase, no force.

- **MAJOR — write-back overwrites a desk assignment with the desk's lead** (`src/harness/brain.rs:126`) — `5274fb7`. Dropped, not patched: `brain.rs` is byte-identical to its pre-PR state again. `task_responder("strategy")` returned `engineer` and `run_task` persisted that, so a card the operator assigned to the Strategy desk silently became assigned to `engineer` on first dispatch. Verified no path still writes a lead over a desk — the write-back site is gone entirely, and #214's is guarded by `AssigneeResolution::links_working_agent()`, true only for `Unassigned | Agent`.
- **Cross-PR — two owners for assignee resolution** — `5274fb7`. Same commit: the `task_responder` desk/overlay-roster extension and its four tests are gone, including the one that asserted `task_responder("Shane") == "ceo"` — the silent misroute #205 exists to eliminate. #214's `runtime::assignee::resolve` is now the only resolver, and it *refuses* an assignee that names nobody rather than routing it to the orchestrator. Also dropped the `assignee` paragraph from the TaskStore spec, since #214 documents the resolution it actually implements. Nothing in the To-do slice reads assignee resolution, so the column work is untouched.
- **Cross-PR — one column list, one name** — `5b919e7`. Reconciled onto #214's name with this PR's membership: `BOARD_COLUMNS` / `is_board_column`, six entries including `todo`. Chose `BOARD_COLUMNS` because it says what the list *is* (the columns the board renders) where `TASK_COLUMNS` only says what the values belong to, and because it reads back correctly in the write boundary's operator-facing 400 (`"…" is not a board column`). Kept six entries because `todo` is what `POST …/tasks` defaults to. The console keeps its own `TASK_COLUMNS` name — pre-existing, used by four views — but its doc comment now points at `BOARD_COLUMNS` and says the host validates against it.
- **NIT — test promises drift protection it cannot deliver** (`src/ports/tasks.rs:47`) — `4cb8da6`. Took the honest-comment option; generating the TS list would need a build step this crate does not have, across a separate npm build. `columns_are_ordered_and_unique` now states that it pins the Rust order only, that drift from `frontend/src/lib/tasks-sample.ts` is unguarded, and why.

**Why the union with #214 is now safe.** One column list including `todo`, under one name, and one owner for assignee resolution. Verified by actually building the union locally rather than reasoning about it — merged this branch with #214's head in a scratch worktree, resolved (take this PR's `BOARD_COLUMNS`; take #214's `validate_column` + `resolve_assignee`; default `create_task` to `COLUMN_TODO`), then `cargo check --features openhuman` clean and `cargo test --features openhuman --lib server::ops::write_test` **43 passed / 0 failed**, including #214's `task_writes_reject_a_column_the_board_cannot_render` and this PR's `created_tasks_default_to_the_todo_column` side by side.

Counterfactual, to show the union bug was real and is what got fixed: dropping `todo` back out of the list in that same union makes **5 write tests fail**, `created_tasks_default_to_the_todo_column` and `tasks_crud_round_trips_under_both_scopes` among them — `POST …/tasks` defaults to `todo` while `validate_column` rejects it, so every card creation 400s. Neither PR's CI can see that, since each is green alone.

Two notes for whoever merges second: `src/ports/tasks.rs`, `src/server/ops/tasks.rs`, `src/server/ops/write_test.rs`, `src/company/runtime.rs`, `src/harness/lifecycle.rs` and `docs/spec/runtime/ports.md` conflict textually — loudly, which is what we want. In `create_task` the union wants **#214's `validate_column` + `resolve_assignee` with this PR's `COLUMN_TODO` default** (not `COLUMN_BACKLOG`); `src/harness/brain.rs` now merges clean, since this PR no longer touches it.

Gates on this branch: `cargo fmt --all -- --check` clean; `cargo check --features openhuman` clean (2 pre-existing `dead_code` warnings in `harness/composio.rs`, untouched); `cargo test --features openhuman --lib` — `ports::tasks` 2/2, `harness::brain` 40/40, `harness::lifecycle` 12/12, `server::ops` 121/121, `company::runtime` 1/1. Frontend: `npm ci`, `npm run typecheck` and `npm run build` all clean (no lint or test script in this package).

oxoxDev:
Rebased onto `main` now that #214 has merged (`eb6fb98`). `MERGEABLE` / `CLEAN`. Title and body updated — the old title's "and agent linking on dispatch" described work that is now #214's.

The SHAs in my previous comment are superseded. New history is two commits:

- `2738603` (@M3gA-Mind) — the To-do column and the To-do-only `+` button
- `e586b43` — the column-order test's honest comment

The two commits that cancelled out (`1268328` linking, `5274fb7` reverting it) were dropped rather than replayed: replaying the linking half onto a `main` that already contains #214's `run_task` rewrite would have meant hand-resolving a heavy conflict for code the next commit deletes, leaving a broken intermediate commit in the history. `5b919e7` (the `TASK_COLUMNS` → `BOARD_COLUMNS` rename) dissolved too — `main` now owns that name, so all that survives is adding `COLUMN_TODO` to it.

How the six conflicts resolved:

- `src/ports/tasks.rs` — `main`'s `BOARD_COLUMNS` / `is_board_column`, widened to 6 with `COLUMN_TODO`. Kept the rationale for why `todo` must be *in the validated list*, not merely defined.
- `src/server/ops/tasks.rs` — `main`'s `validate_column` + `resolve_assignee` structure, defaulting to `COLUMN_TODO` rather than `COLUMN_BACKLOG`. This is the line the union bug turned on.
- `src/server/ops/write_test.rs` — union, both sides' tests kept.
- `src/harness/lifecycle.rs`, `src/company/runtime.rs` — comment-only conflicts over an identical `use` line; took `main`'s. This PR adds nothing there.
- `docs/spec/runtime/ports.md` — took `main`'s side for the assignee paragraph, since #214 documents the behaviour it implements; added `todo` to the column enumeration and the Backlog-vs-To-do paragraph.

`src/harness/brain.rs` no longer appears in the diff at all.

Gates on the rebased branch: `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo build --all-targets`, `cargo check --features openhuman` all clean; `cargo test` 956 passed, `cargo test --features openhuman` 1294 passed, 0 failed. Frontend `npm ci` + `npm run typecheck` + `npm run build` clean. `cargo clippy --all-targets --features openhuman` fails in vendored `tinyagents` on `main` as well — untouched here, noted in the body.

PR #231, opened 2026-07-31
fix(runtime): don't lose a shutdown that arrives during a tick (#229)

### LINKED ISSUE #229: CompanyScheduler loses a shutdown notification that arrives during a tick
## Problem

`CompanyScheduler::spawn` (`src/runtime/scheduler.rs`) watches for shutdown with

```rust
tokio::select! {
    _ = shutdown.notified() => break,
    _ = tokio::time::sleep(...) => { self.tick().await; }
}
```

The `Notified` future is rebuilt on every loop iteration, so it is only registered as a waiter while the loop is sleeping. Boot signals shutdown with `notify_waiters()`, which wakes only the waiters registered at that instant — so a notification delivered **while `tick()` is running** is dropped, and the scheduler sleeps another full minute before it notices it was asked to stop.

## Impact

Low but real: up to a minute of extra shutdown latency per scheduler, and one more tick executed after shutdown was requested. It matters most when a tick is slow, which is exactly when a shutdown is most likely to land inside one.

## Fix

Create the `Notified` future once before the loop and `tokio::pin!` it so it stays registered across iterations, latching a notification that arrives mid-tick.

## Provenance

Found while addressing review on PR #211, which fixed the identical bug in the new `WorkflowScheduler` (commit `a996f7c` there is the reference implementation). This one was pre-existing and outside that PR's diff, so it was deliberately not bundled.

## Pointers

- `src/runtime/scheduler.rs` — `CompanyScheduler::spawn`
- `src/runtime/workflow_scheduler.rs` — the already-fixed equivalent to copy

oxoxDev:
## Summary

`CompanyScheduler` drops a shutdown notification that arrives while a tick is running, so it keeps ticking and only stops a minute later.

Closes #229.

## API Or Behavior Changes

None externally. Shutdown latency for the company scheduler drops from up to a full sleep interval to the current tick, and no extra tick runs after shutdown was requested.

## Problem

The loop rebuilt its `Notified` future on every iteration:

```rust
tokio::select! {
    _ = shutdown.notified() => break,
    _ = tokio::time::sleep(...) => { self.tick().await; }
}
```

A `Notified` is only registered as a waiter while it is being awaited, and shutdown is signalled with `notify_waiters()`, which wakes only the waiters registered at that instant. So a notification delivered *during* `tick()` reaches nobody and is lost — the scheduler sleeps another full interval before noticing.

The premise was verified rather than assumed. Across `src/` there is exactly one `Notify` construction and exactly one signal site, and no `notify_one()` anywhere in the repository:

- `src/bin/opencompany.rs:829` — `let shutdown = Arc::new(Notify::new());`
- `src/bin/opencompany.rs:864` — `shutdown.notify_waiters();`, the sole signal, in the Ctrl-C handler

It matters most when a tick is slow, which is exactly when a shutdown is most likely to land inside one.

## Solution

Build the `Notified` once before the loop and `tokio::pin!` it, so it stays registered across iterations and latches a notification that arrives mid-tick.

This is the same fix already applied to the newer `WorkflowScheduler` in #211 (`a996f7c`). The two loops were compared before reusing the approach — `CompanyScheduler`'s tick arm additionally calls `tick_maintenance()` and log-warns on a `Result`, but the `select!` / `Notified` structure is identical, so it transfers exactly.

## Tests

- [x] `cargo test` — 900 passed, 0 failed

Also run: `cargo check --all-features` and `cargo check --features openhuman,mcp,telegram`. No frontend surface, so no frontend gates.

`shutdown_during_a_tick_stops_the_loop` was written first and **failed against the unfixed code**:

```
panicked at src/runtime/scheduler.rs:545: scheduler kept running after a shutdown
delivered mid-tick: Elapsed(())
```

It deliberately avoids a timing race. `tokio`'s `test-util` feature is not enabled (so paused virtual time is unavailable, and auto-advance could fire the bound during legitimate `tokio::fs` work), so the assertion is **loop termination**, not elapsed time: a `oneshot` handshake parks the brain mid-cycle so the shutdown is provably delivered while `tick()` is in flight, and the `FakeClock` sits at `09:00:59.999` so the next sleep is 1ms. A scheduler that drops the signal never joins; a correct one exits on its next iteration. The 5s bound only caps how long a *failure* takes to report — the passing case finishes in 0.01–0.02s, run 5×.

## Documentation

None needed — the fix is internal to the scheduler loop, and the reason it is written this way is in a code comment beside the pinned future.

## Related

- Found while addressing review on #211, which fixed the identical bug in the newer `WorkflowScheduler`.
- **`src/runtime/mailbox_poller.rs:102` is very likely a third instance** — same `spawn(shutdown: Arc<Notify>)` signature, sharing the same `Notify`. Not inspected or touched here, since it is outside this issue's scope; worth its own look.

### DIFF
--- a/src/runtime/scheduler.rs
+++ b/src/runtime/scheduler.rs
@@ -176,10 +176,21 @@ impl CompanyScheduler {
     /// shared `shutdown` so the scheduler stops cleanly when the server does.
     pub fn spawn(mut self, shutdown: Arc<Notify>) -> JoinHandle<()> {
         tokio::spawn(async move {
+            // The `Notified` future is built ONCE and pinned across iterations,
+            // not rebuilt inside the `select!`. Boot signals with
+            // `notify_waiters()`, which wakes only the waiters registered at
+            // that instant — a future created fresh each iteration is not
+            // registered while `tick` is running, so a shutdown arriving
+            // mid-tick would be dropped and the scheduler would sleep another
+            // full minute before noticing. Polled once here, this one stays
+            // registered, and a notification delivered during `tick` is
+            // latched: the next `select!` sees it immediately.
+            let notified = shutdown.notified();
+            tokio::pin!(notified);
             loop {
                 let sleep_ms = millis_to_next_minute(self.clock.now_millis());
                 tokio::select! {
-                    _ = shutdown.notified() => break,
+                    _ = &mut notified => break,
                     _ = tokio::time::sleep(Duration::from_millis(sleep_ms)) => {
                         if let Err(err) = self.tick().await {
                             tracing::warn!(company = %self.runtime.id(), %err, "scheduled cycle failed");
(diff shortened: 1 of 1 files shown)


oxoxDev:
CI red here is a pre-existing flake, not this change: `runtime::journal::test::expired_record_removes_parked_and_survives_reload` failed with `Serde(Error("trailing characters", line: 1, column: 211))`, in a file this diff does not touch (it changes only `src/runtime/scheduler.rs`).

Root cause filed as #235: the journal tests build their temp filename from `generate_id()`, which is `millis` plus a **process-local** atomic counter, inside the shared `std::env::temp_dir()`. Two test processes starting in the same millisecond both begin at counter 0, produce the same filename, and append to the same JSONL file — so a line ends up holding two JSON objects and the reader fails exactly as observed. Invisible locally (one binary at a time), rare in CI, and capable of failing any PR at random.

Locally: 5/5 in isolation, 900/900 in a full run at `--test-threads=32`. `main` is green. Re-running the failed job.

Issue #276, opened 2026-08-02
feat(workflows): enable/disable without deleting, and disarm on edit


oxoxDev:
### Context

Two related gaps #259 left open, filed together because settling either requires settling the same decision.

**(a) There is no way to pause a workflow.** The only way to stop a schedule firing is to delete the workflow, which loses the graph.

**(b) An edit can arm a schedule with no review.** OpenHuman's `flows_update` enforces "B29 Rule 1": when an edit turns a manual or absent trigger into an automatic one, it forces `enabled = false` in the same store write, so the user must explicitly re-arm after reviewing. Its own docs record this firing live — a flow started running on an unreviewed 8am schedule.

### Why both are the same decision

`WorkflowScheduler::tick` **deliberately does not gate on `[workflows].enabled`** (reasoned at `src/runtime/workflow_scheduler.rs`: a trigger `schedule` is itself the operator's "run this on a cron" statement, so re-asking the enabled list adds a second switch for one decision).

Given that, (a) has no mechanism and (b) has no lever — writing `enabled = false` would stop nothing today. Both need the gating decision reversed first.

Note (b) is **not a regression from #259**: `create_company_workflow` already persists a live schedule the same way, so an edit is no riskier than authoring the same graph fresh. Whatever rule is adopted must cover create and update together, or it just moves the hole.

### Scope

Decide whether `enabled` gates the scheduler.

**If yes:** make the tick honour it, add `PUT …/workflows/{wid}/enabled`, surface a toggle in the console, adopt the disarm-on-manual-to-automatic rule in both create and update, and pin it in `workflow_scheduler.rs` — note `adding_a_schedule_by_edit_arms_the_workflow_on_the_next_tick` currently pins the *opposite* behaviour and would need inverting.

**If no:** document why explicitly and close, since (a) then has no answer short of delete.

oxoxDev:
**Related — `cluster:workflows`**

Workflow engine, run lifecycle, scheduling, and canvas.

*Gated on the vendored engine surface — read #499 first:*
- #382 — the canvas cannot show which node is executing, because `RunObserver` has no `on_step_start`.
- #398 — cancelling a run drops the engine future, because tinyflows cannot observe and cancel at once.
- Both need a `tinyflows` change, and `tinyflows` only moves with the `openhuman` pin. **#499 (PR #505, open) is what unblocks them** — #499 names #382 as the first issue to hit it.

*Both fell out of #259 (closed) and settle the same decision:*
[...]

Issue #275, opened 2026-08-02
feat(ports): a retention policy for the company event journal


oxoxDev:
### Context

#259 deletes a workflow but deliberately leaves its `WorkflowRunFinished` entries in the journal. Nothing in OpenCompany ever removes a journal entry, so the log grows without bound for the life of a company.

### What OpenHuman does

`flow_runs` is a flow-owned table, so `flows_prune_runs` is one bounded `DELETE` keeping `MAX_FLOW_RUNS_PER_FLOW`, plus `ON DELETE CASCADE` when the flow goes.

### Why that shape does not port (verified by grep, not assumed)

The `EventLog` port is exactly three methods — `append`, `read_from`, `subscribe` (`src/ports/events.rs:12-24`). **There is no delete of any kind**, and there are three production implementors: `store/fs.rs:351`, `store/sqlite.rs:433`, `store/mongodb.rs:371`.

Our run outcomes are not a flow-owned table; they are `CompanyEvent` rows interleaved with chat and audit in one append-only log. So this is not "add a delete flag" — it is a retention *design*: a new port method, three backends, conformance tests, and a per-event-kind policy (a run outcome may be prunable where a `LifecycleChanged` or an approval verdict is not).

### Questions to settle first

- Is the journal an audit log — append-only, never pruned? If so the answer is archival or compaction, not deletion.
- If prunable, is policy by age, by count, or by event kind?
- Does pruning renumber `EventSeq`, and what does that do to `read_from` cursors and the SSE stream?

oxoxDev:
**Related — `cluster:journal`**

The company event journal: what it guarantees, what it keeps, and what it can be made to forget.

- #392 — decide whether an append should survive losing the **host**, not just the process. #386 made the write syscall complete before returning; the bytes still sit in the page cache, so a power loss can lose a record reported durable. Matters most for the at-most-once contract on `EffectExecuted`.
- #450 — the embedded writer reports an append failure **per queued item**, unbounded and undeduped, so an unwritable root turns the container log into one repeated line. #446 removed the trigger; this is the amplifier that is still armed.
- #275 — a retention policy: nothing in OpenCompany ever removes a journal entry, so the log grows for the life of a company.

**Sequencing note:** #358 (PR #504, open) adds a withdrawal path for discussion posts — the first mechanism that mutates journaled history. Whoever picks #275 should read what #358 lands first; a retention policy and a redaction path want the same primitive, and building them twice would be waste.

Issue #274, opened 2026-08-02
feat(workflows): revision history and rollback for an edited workflow


oxoxDev:
### Context

#259 made a saved workflow replaceable (`PUT …/workflows/{wid}`), guarded by an optimistic-concurrency token. What it does not do is keep the graph it replaced. An edit that drops a node or mangles a cron is unrecoverable except by re-authoring from memory.

### What OpenHuman does

`flow_revisions` is a dedicated SQLite table. Every `store::update_flow_graph` inserts the prior graph as a revision inside the same guarded `UPDATE`, then prunes to `MAX_REVISIONS_PER_FLOW = 20`. `flows_get_history` lists them newest-first; `flows_rollback` restores one *through the normal update path*, so a rollback is itself snapshotted and therefore undoable.

### Why it was not done in #259

Our overlay bodies live inside `CompanyRecord`, which is loaded and saved **whole** on every write. A 20-snapshot ring per workflow would ride that hot path on every unrelated company save. This needs its own store surface (a new port plus the three backends), not a field on the record.

### Shape

#259's full-replace-plus-version-token is exactly the shape a snapshot-on-update slots into: the update already knows the prior TOML under the write lock, which is the only moment a snapshot can be taken race-free.

Suggested: a `WorkflowRevisions` port (`push` / `list` / `get`, bounded per workflow), `GET …/workflows/{wid}/revisions`, and `POST …/workflows/{wid}/rollback` routed through `update_company_workflow` so a rollback is undoable and re-validated.

### Out of scope

Diffing or merging revisions.

oxoxDev:
**Related — `cluster:workflows`**

Workflow engine, run lifecycle, scheduling, and canvas.

*Gated on the vendored engine surface — read #499 first:*
- #382 — the canvas cannot show which node is executing, because `RunObserver` has no `on_step_start`.
- #398 — cancelling a run drops the engine future, because tinyflows cannot observe and cancel at once.
- Both need a `tinyflows` change, and `tinyflows` only moves with the `openhuman` pin. **#499 (PR #505, open) is what unblocks them** — #499 names #382 as the first issue to hit it.

*Both fell out of #259 (closed) and settle the same decision:*
- #274 — revision history and rollback for an edited workflow.
- #276 — enable/disable without deleting, and disarm on edit.

*Bounds and durability:*
- #401 — nothing bounds how many runs of a company can be in flight (the accidental socket-based cap died with #383).
[...]

Issue #271, opened 2026-08-02
bug: Playwright global-setup intermittently gets no dev_code and aborts the suite


oxoxDev:
## Summary

`frontend/test/e2e/global-setup.ts` intermittently fails to obtain a `dev_code`, which aborts the entire Playwright suite before any spec runs.

The same request issued with `curl` against the same host succeeds seconds later, so the host is healthy and the endpoint is working. The failure is specific to the `request.newContext()` path Playwright uses.

The obvious suspect is the 60-second resend throttle on `POST …/auth/request`, which returns `{"sent": true, "dev_code": null}` rather than an error. **That was tested and disproved** — it still failed after 75 seconds of quiet against a brand-new `--home`.

Impact is out of proportion to the symptom: the suite is the only way to verify anything that is a DOM event or a network round trip, and three recent branches all needed it. An intermittent global setup makes it unreliable for everyone and tempts people to skip it.

## Reproduction

1. Start a host with a fresh `--home`.
2. `npm run e2e`.
3. Intermittently, global setup reports no `dev_code` and the run aborts. Immediately afterwards the identical `curl` request returns one.

## Expected behavior

Global setup either obtains a code or fails with a diagnosable message.

Worth investigating in this order:

- Whether the response was actually a throttle (`dev_code: null` with `sent: true`) or something else — the current failure does not distinguish them, which is why the throttle theory took a while to rule out.
- Whether `request.newContext()` differs from curl in a way the host reacts to — headers, connection reuse, or the loopback check that gates the `dev_code` echo in the first place. That gate is the interesting one: it turns the code off for anything the host does not consider local.
- Retry with backoff in global setup regardless, since a flaky bootstrap is worth defending against even once the root cause is known.

PR #279, opened 2026-08-02
feat(workflows): delete a saved workflow, and replace one with conflict detection (#259)

### LINKED ISSUE #270: bug: the Workflows picker lists workflows by id instead of name
## Summary

The Workflows picker lists each workflow by its **id** rather than its name. A workflow saved as "Daily Standup" with id `daily-standup` shows as `daily-standup`; one created through the console with a generated id shows as that.

Both fields are already on the wire — `WorkflowSummary` carries `id`, `name` and `description`, and the detail panel renders the name correctly. Only the list is wrong.

Verified identical on `1423a83`, so this is pre-existing rather than introduced by any of the recent workflow work.

## Reproduction

1. Console → Workflows.
2. Compare a row in the picker against the name shown once that workflow is selected.

## Expected behavior

The picker shows the name, with the id available where it is actually needed — an id is what `POST …/workflows/{wid}/run` takes and what appears in run history, so it is worth keeping visible, just not as the primary label. A name with the id as secondary text would serve both readings.

oxoxDev:
## Summary

Closes #270. Advances #259 — a saved workflow could not be edited or deleted, so a mistyped cron fired forever and the only recovery was authoring a second workflow beside the broken one.

This lands the whole backend, the API client, and **Delete** in the console. **Edit's button is deliberately not in this PR** — it needs the create dialog's edit mode, which collides with #277 (`WorkflowCreateDialog.tsx`); a visibly broken Edit is worse than none. `PUT` is fully built and tested at the API level, and `editable` plus the conflict banner are live and exercised *through* Delete, so nothing here is dead code waiting on a follow-up. The dialog commit adds a button and one `updateWorkflow` call.

**Conflict detection, not last-write-wins.** `GET` returns a `version` — a hash of the stored graph — and `PUT`/`DELETE` accept it back. Mismatched means someone else changed the workflow since you loaded it, answered 409 with a reload-and-reapply message. The comparison happens inside the same write lock as the save, so it is race-free. Omitting the token writes unconditionally, which keeps curl usable; the console always sends it. This follows OpenHuman's `flows_update`, which takes an optional `expected_version` and enforces it in a guarded UPDATE — reading that is what overturned an earlier plan that had settled for last-write-wins on the grounds that versioning would ripple through three storage backends. A content hash needs no schema change at all.

**Only console-created workflows are editable.** Ones defined in a company's source tree answer 409 with an actionable message, and render disabled with a tooltip. The seed file wins the read union, so an edit would persist and never display; and the boot rebuild reconstructs `enabled` from seed ids, so a delete would resurrect on restart. Rejecting cleanly beats half-working.

**Deleting keeps run history.** The `EventLog` port has no delete across any of its three backends, and run outcomes are interleaved with chat and audit in one append-only journal. Filed as #275 rather than bodged.

**The scheduler needs no changes**, and that is now proven rather than assumed. OpenHuman needs `reconcile_schedule_triggers_on_boot` because a scheduled flow binds a row in a separate `cron.db` that can drift from `flows.db`. We persist no registration at all — the tick re-reads the record and re-derives the schedule set every minute, so the tick *is* the reconcile. Three tests pin it.

Follow-ups filed: #274 revision history and rollback · #275 journal retention · #276 enable/disable plus disarm-on-edit.

## API Or Behavior Changes

- **New** `PUT /workflows/{wid}` — full replace, same validation as create. Optional `expectedVersion` in the body. `body.id` must equal `{wid}`; rename via PUT is rejected 400, since the id keys the read union, the scheduler and run history. Unknown 404, seed-backed 409, stale token 409.
- **New** `DELETE /workflows/{wid}` — 204. Optional `?expectedVersion=`. Removes the overlay body and the id from `[workflows].enabled` in one save.
- `WorkflowSummary` and `WorkflowGraph` gain `editable: bool`; `WorkflowGraph` also gains `version: String` when overlay-backed. Both additive.
- **New events** `CompanyEvent::WorkflowUpdated` / `WorkflowDeleted`, carrying id and name only. A test asserts no graph body can reach the shared journal.

## Tests

- [x] `cargo test` — 1048 passed / 0 failed (+6)

Also `cargo check --all-features` and `cargo check --features openhuman,mcp,telegram` — the new event variants cross feature boundaries, and `--all-features` is what located all three gated match sites; the default combos surface none. Frontend: `npm ci && npm run typecheck && npm run build`.

Live curl over every path: PUT happy / stale-409 / id-mismatch-400, DELETE happy-204 / stale-409 / unknown-404, the token moving on each write, a real seed-backed workflow refusing both writes, and **restart survival** — a deleted workflow did not resurrect.

**4 new Playwright specs, 4/4**, including forcing a genuine conflict by editing out-of-band over HTTP while the console held a stale token. Not run by CI (no `webServer`).

**One gate not completed:** the full-suite regression baseline. Playwright's `global-setup` intermittently gets no `dev_code` while curl to the same host succeeds seconds later. The 60s resend throttle was the obvious suspect and was disproved — it still failed after 75s of quiet on a brand-new `--home`. Filed as #271 rather than papered over.

## Documentation

`docs/modules/` workflow page and the `src/server/ops/workflows.rs` module header cover the PUT/DELETE contract, the version token, the seed-vs-overlay rule, and that run history is orphaned rather than reaped.

**#270 is now fixed here too.** The Workflows picker rendered the raw id in its closed trigger while the popup showed the name — `<SelectItem value={w.id}>` with a bare `<SelectValue />`, which renders the value rather than the item's children. It is the same picker this PR's Delete affordance sits next to, and the delete specs had to key on the id to describe the console honestly; both now read the name.

One pre-existing bug found and filed, **not** fixed here: confirming an `AlertDialog` leaves the modal mounted (#268). This PR's own delete flow works around it by controlling `open`, but the cause is in the shared `AlertDialogAction` primitive and the same pattern is at three more sites on `main` — so it is fixed once at the primitive on its own branch. Whichever merges second, the local workaround here is belt-and-braces, not a conflict.

One gap worth naming: OpenHuman's `flows_update` enforces that an edit turning a manual trigger automatic forces `enabled = false`, so a schedule cannot go live unreviewed. **We have no lever for that** — our scheduler deliberately does not gate on `enabled`, so writing `false` would stop nothing. Not a regression, since create already behaves identically, but it is documented, pinned by a test, and folded into #276.

### DIFF
--- a/docs/modules/server/README.md
+++ b/docs/modules/server/README.md
@@ -46,7 +46,88 @@ address check for `{id}`, operator + `sole()` for the alias).
 | `domain` | `PUT …/domain`, `POST …/domain/verify` |
 | `smtp` | `PUT …/smtp`, `POST …/smtp/test` |
 | `connections` (feature `oauth`) | `POST …/connections/{provider}/start\|disconnect`, `GET /api/v1/oauth/callback` |
-| `workflows` | `POST …/workflows`, `GET …/workflows`, `GET …/workflows/{wid}`, `POST …/workflows/{wid}/run` |
+| `workflows` | `POST …/workflows`, `GET …/workflows`, `GET …/workflows/{wid}`, `PUT …/workflows/{wid}`, `DELETE …/workflows/{wid}`, `POST …/workflows/{wid}/run`, `GET …/workflows/runs` |
+
+### Editing and removing a workflow (issue #259)
+
+A saved workflow used to be write-once: a typo'd cron or a node pointed at the
+wrong teammate was permanent, and the only recovery was to author a second
+workflow and leave the broken one firing. `PUT` replaces a graph wholesale and
+`DELETE` removes it.
+
+**Only overlay-backed graphs are writable.** A workflow defined by a file in the
+company source tree (`companies/<name>/workflows/<wid>.toml`), or enabled by
+name with no saved graph at all, answers `409` — it is not squeamishness about
+touching disk, it is the reader's rules restated. `load_workflow_union` gives a
+seed file precedence on an id collision, so an edit stored behind one would
+never be served; and `merge_enabled_workflows` re-derives `[workflows].enabled`
+from seed ids at boot, so a "delete" of a seed-backed workflow would come back
+on the next restart. The same invariant is what makes an overlay delete durable.
+
+The read routes project that predicate as `editable`, so the console can grey
+its buttons out instead of surfacing a 409 after the click.
+
+**The version token.** `GET …/workflows/{wid}` returns an opaque `version` for
+an editable graph. Echo it back — in the `PUT` body as `expectedVersion`, or on
+the `DELETE` as `?expectedVersion=` — and the write is refused with `409` if the
+graph moved in between, so one console cannot silently overwrite another's edit.
+The comparison happens under the same per-company write lock as the save, so it
+is a real guard rather than a check-then-act race. **Omitting it is an
+unconditional write**, which keeps `curl` usable without a read-modify-write
+dance; the console always sends it. Never parse the token — the contract is
+"echo back what the read returned", which is what lets the algorithm change
+without a client migration.
+
+**The id may not change.** A `PUT` whose body `id` differs from `{wid}` is a
+`400`. The id keys the union read path, the scheduler and every journalled run,
+so a rename would silently orphan all three. A rename is a create plus a delete.
+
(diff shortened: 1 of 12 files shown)


oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

PR #282, opened 2026-08-02
fix(console): refuse an inference save that would discard a typed key (#265)

### LINKED ISSUE #265: bug: saving inference with provider=managed silently discards the key
## Summary

In the console's Connections → Inference section, saving with the provider left on **managed** silently discards whatever key was typed, then reports success.

`save()` in `frontend/src/views/connections/InferenceSection.tsx` branches on the provider:

```js
if (provider === "managed") {
  await revertInference(client, company);   // key never sent
} else {
  await setInference(client, company, { provider, baseUrl, models, key: key.trim() || undefined });
}
```

The reasoning behind the branch is sound — "managed" means *use the platform default*, which is a revert rather than an override. The problem is the form does not reflect that: the key input stays rendered and editable while managed is selected, accepts the credential, and the save path drops it. The toast then says "Inference updated. Agents use it on their next turn."

The result is an operator who believes they have configured a credential and has not. On a fresh company this is easy to hit, because managed is the default selection and the status panel shows a managed base URL already populated — so attaching a key to it looks like the intended action.

Observed: after the flow above, `GET …/inference` still returns `keyConfigured: false`, `cognition: "echo"`, and no write appears in the server log.

## Reproduction

1. Boot a company with no inference configured.
2. Console → Connections → Inference. Leave the provider on **managed**.
3. Type any value into the key field. Save.
4. Observe the success toast.
5. `curl …/api/v1/companies/<id>/inference` → `keyConfigured: false`.

## Expected behavior

The key must not be silently discarded. Any of these would do, roughly in order of preference:

- Hide or disable the key input while managed is selected, with a line explaining that managed uses the platform credential and that a key requires choosing a provider.
- Keep the field enabled but block Save with a validation message naming the fix ("managed uses the platform credential — choose `openai_compatible` or `openrouter` to bring your own key").
- Let managed accept a key, if a managed credential is a thing the backend can actually store. That is a product question, not a UI one — the other two are safe regardless.

At minimum the toast must not claim success for a save that discarded input.

oxoxDev:
## Summary

Connections → Inference could report a successful save that had thrown the operator's API key away.

The key input is rendered only for the bring-your-own-key providers, but the value the operator typed lives in form state that survives a switch back to **managed**. Managed means "use the platform default", so the save path takes the revert branch (`DELETE …/inference`), which carries no credential — the pending key was dropped and the toast still said `Inference updated. Agents use it on their next turn.` The operator walks away believing a credential is configured; `GET …/inference` still reports `keyConfigured: false`.

Of the three fixes the issue lists, this PR takes **"block the save with a validation message"**, and completes the "explain it in the form" half that was already partly there:

- The key input was **already** hidden for managed on `main`, so the issue's literal repro (typing into the field with managed selected) is not reachable — but hiding alone does not close the hole, because the state behind the hidden field is what gets discarded. The reachable path is: pick OpenRouter → type the key → switch back to managed → Save.
- Auto-switching the provider when a key is typed was rejected: it silently overrides an explicit dropdown choice, and this console never moves a control the operator set.
- Clearing the key on every provider switch was rejected too: it is the same silent discard, just moved earlier and hidden behind a field that disappears.

The invariant pinned here is that **a save which reports success must never have discarded operator input**. Managed now (a) says it runs on the platform credential and that a key needs a different provider, (b) flags a pending key with an explicit `Discard key` action, and (c) disables Save until the operator resolves it either way. `save()` carries the same guard, so the invariant holds even if the button state is bypassed.

Frontend-only — zero Rust diff.

## API Or Behavior Changes

No API change. The `…/inference` routes, request shapes, and the write-only key contract are untouched.

Console behavior, Connections → Inference:

- Selecting **managed** now shows a line explaining that it runs on the platform credential and that a key requires OpenRouter or Custom (OpenAI-compatible).
- With a key still pending from a previous provider selection, managed shows a warning with a `Discard key` button and **Save is disabled**. Clicking Save in that state (or reaching `save()` any other way) produces an error message, never a revert and never a success toast.
- Every other path is unchanged: managed with no pending key still saves as a revert, and BYOK saves still send the key exactly as before.
- Known and deliberately left alone: `Reset to managed` also clears a pending key while reporting success. Its label announces that it throws the in-progress configuration away, so it is an explicit act rather than a silent one — but it is the nearest sibling of this bug and worth a look if the maintainers disagree. Separately, a key typed under OpenRouter and then saved under **Ollama** (whose key field is also hidden) is *stored* rather than discarded — surprising, but no input is lost, so it is out of scope here.

## Tests

Frontend gates (`frontend/`):

- `npm ci && npm run typecheck` — clean.
- `npm run build` — clean (pre-existing chunk-size warning only).
- Note: the console has no lint or unit-test script; `typecheck` + `build` are the whole gate. The e2e specs are outside `tsconfig.app.json`'s `include`, so they are exercised by running them, not by `typecheck`.

Browser proof (`frontend/test/e2e/inference.spec.ts`, new). Playwright drives a real Chromium against a live host serving the built bundle, on an isolated `--home`. **This spec is not part of CI** — `playwright.config.ts` declares no `webServer`, so it is an executable reproduction plus documentation of the invariant, not a merge gate.

- Against the **unpatched** bundle, a scratch version of this flow reproduced the bug: the success toast appeared and `keyConfigured` stayed `false`.
- Against the patched bundle, both specs pass (`2 passed`): the blocked-save case (warning shown, Save disabled, stored-credential flag unchanged, `Discard key` unblocks it, and returning to a BYOK provider offers an empty field) and the positive case (the same key, saved under `openai_compatible`, does reach the host).

Rust gates — run despite the empty Rust diff:

- [x] `cargo test` — `1009 passed; 0 failed; 1 ignored`, plus the binary's `6 passed`

## Documentation

`docs/spec/runtime/manifest.md` — the `[inference]` bullet already described console-vs-manifest precedence; it now also records that saving `managed` from the console is a revert that carries no credential, and that the console refuses such a save while a key is typed rather than dropping it.

Closes #265

### DIFF
--- a/docs/spec/runtime/manifest.md
+++ b/docs/spec/runtime/manifest.md
@@ -119,6 +119,9 @@ prompt = "Weekly review and operator digest"
   neither. The status route reports that state as `restartRequired` — a resolved
   config next to a non-harness `cognition` — and the console says "restart"
   instead of "next turn" for it (issue #266).
+  Saving `managed` from the console is a *revert* (`DELETE …/inference`) and
+  carries no credential, so the console refuses that save while a key is still
+  typed in the form rather than dropping it and reporting success (issue #265).
 - **`[channels.*]`** enables `ChannelAdapter`s. Unknown channels are a
   validation error; disabled OpenHuman means non-operator channels degrade
   with a boot warning, never a failure.
--- a/frontend/src/views/connections/InferenceSection.tsx
+++ b/frontend/src/views/connections/InferenceSection.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useState } from "react";
-import { BrainCircuit, Check, Loader2, RotateCcw, Save, Zap } from "lucide-react";
+import { AlertTriangle, BrainCircuit, Check, Loader2, RotateCcw, Save, Zap } from "lucide-react";
 import { toast } from "sonner";
 
 import type { OpenCompanyClient } from "@/api/client";
@@ -136,6 +136,18 @@ export function InferenceSection({
 
   async function save() {
     if (busy) return;
+    // Managed is a revert, and a revert cannot carry a credential — so a key
+    // typed under a different provider and left behind by a switch would be
+    // dropped by the save below while the toast claimed success (issue #265).
+    // Refuse instead: a save that reports success must never have discarded
+    // what the operator typed. The Save button is disabled in this state; this
+    // is the guard that makes the invariant hold regardless of the button.
+    if (provider === "managed" && key.trim()) {
+      toast.error(
+        "Managed uses the platform credential and can't store a key. Choose OpenRouter or Custom (OpenAI-compatible) to save it, or discard it first.",
+      );
+      return;
+    }
     setBusy("save");
     try {
       // "Managed" means "use the platform default" — that's a revert, not a
@@ -219,6 +231,10 @@ export function InferenceSection({
   if (load === "unavailable") return null;
 
   const modelRows = status ? Object.entries(status.models) : [];
+  // A credential typed under a BYOK provider survives a switch to managed (the
+  // input is hidden, the state is not). Managed has nowhere to put it, so this
+  // is the one combination that would silently lose operator input on save.
+  const managedWouldDiscardKey = provider === "managed" && key.trim().length > 0;
 
   return (
     <section className="space-y-3">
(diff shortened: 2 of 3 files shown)


oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai review

PR #284, opened 2026-08-02
fix(auth): stop the resend throttle blocking the echoed dev login code (#271)

### LINKED ISSUE #271: bug: Playwright global-setup intermittently gets no dev_code and aborts the suite

oxoxDev:
## Summary

The Playwright bootstrap did not fail intermittently — it failed **every time the
suite was run within 60 seconds of the previous run**, which is what an iterating
developer does. The 60s resend throttle on `POST …/auth/request` applied even on
a loopback host with no mail transport, where the login code is echoed in the
response rather than mailed. Because only the code's *hash* is stored, a
throttled answer cannot re-echo the still-live code: the caller gets
`202 {"sent": true}` and no way in, for a minute after every single sign-in.

The throttle exists to stop `auth/request` being a mail cannon aimed at an
invited mailbox. Where nothing is mailed there is no cannon — so it is now
skipped on exactly the shape the dev echo is already gated on (loopback bind, no
`public_url`, no mail transport wired). Every other host keeps it unchanged.

This is a root-cause fix, not a retry: no backoff loop was added, because after
the change there is nothing to retry around. First-run-on-a-fresh-host was never
the failing case (verified below), so there is no host-readiness race to defend
against either.

Separately, the bootstrap's failure message was actively misleading. It died on a
bare `expect(devCode).toBeTruthy()` with "must echo dev_code on a loopback bind",
which reads as a broken host. `auth/request` answers `202 {"sent": true}`
identically for an unknown address, a throttled resend, a routable host and a
mailing host — deliberately, so the route is not a membership oracle — so the
client cannot be told which happened and must say so. It now reports the
endpoint, the status and the verbatim body, then enumerates the conditions that
suppress the echo, in the order worth checking.

Closes #271

## API Or Behavior Changes

`POST …/auth/request` on a **loopback-only host with no mail transport** now
mints and echoes a fresh code on every request instead of returning
`{"sent": true}` with no code inside the 60s window. The reissue still
invalidates its predecessor, so "one live code per address" holds.

No change on any other host. A routable host, or a host with
`OPENCOMPANY_PUBLIC_URL` set, or a host with `OPENCOMPANY_MAIL_*` wired, keeps
the throttle exactly as before and still never echoes a code. The response shape
is unchanged, the generic-`202` rule is unchanged, and no new membership signal
is introduced — on a loopback/no-mail host the echo is already the oracle, by
design.

## Tests

- `cargo build --all-targets` — clean
- `cargo test` — 1011 passed, 0 failed, 1 ignored
- `cd frontend && npm ci && npm run typecheck && npm run build` — clean

New Rust coverage in `src/server/users/routes_test.rs`:

- `a_loopback_host_with_no_mail_reissues_inside_the_resend_window` — two
  sign-ins back to back both get a code, the codes differ, the second verifies
  and the first is invalidated.
- `a_routable_host_still_throttles_even_with_no_mail` — the exemption is scoped
  to the echo path; a routable host keeps the throttle and leaves the live code
  in place.
- The existing `a_second_link_within_the_throttle_window_is_not_sent` (mail
  wired) is untouched and still green.

End-to-end verification, driving the real `global-setup.ts` against a live host
booted with an isolated `--home`:

| Host binary | Setup | Result |
|---|---|---|
| `upstream/main` | 12 consecutive runs, one fresh host | **1 / 12 passed** — run 1 OK, runs 2–12 all failed |
| `upstream/main` | 6 fresh hosts, 1 run each | 6 / 6 passed — the first run is never the failure |
| this branch | 12 consecutive runs, one fresh host | **12 / 12 passed** |
| this branch | 5 fresh hosts, 4 runs each | **20 / 20 passed** |

32 clean bootstrap runs after the change; the reproduction on `main` is
deterministic rather than intermittent once the trigger is understood.

## Documentation

`docs/spec/runtime/users.md` — the "Abuse and exposure" section now says the
resend throttle covers one *mailed* link per address per minute, and states why
it does not apply where the code is echoed instead.

### DIFF
--- a/docs/spec/runtime/users.md
+++ b/docs/spec/runtime/users.md
@@ -194,10 +194,14 @@ can send mail never echoes it.
 
 ## Abuse and exposure
 
-- **Resend throttle**: one link per address per minute. A throttled request
-  returns the same `202` as a sent one (or the throttle is itself a membership
-  oracle) and leaves the live code alone (or anyone could invalidate a victim's
-  link on demand).
+- **Resend throttle**: one *mailed* link per address per minute. A throttled
+  request returns the same `202` as a sent one (or the throttle is itself a
+  membership oracle) and leaves the live code alone (or anyone could invalidate
+  a victim's link on demand). It does not apply where the code is echoed rather
+  than mailed — a loopback-only bind with no transport wired. There is no
+  mailbox to spare there, only the plaintext's hash is stored (so a throttled
+  answer cannot re-echo the live code), and throttling would lock the sole local
+  sign-in path for a minute after every use.
 - **Login codes are never echoed** from a host that is reachable from anywhere
   else. `dev_code` appears only on a loopback-only bind with no mail
   transport. A routable host with broken mail lets nobody in rather than
--- a/frontend/test/e2e/global-setup.ts
+++ b/frontend/test/e2e/global-setup.ts
@@ -1,34 +1,129 @@
-import { expect, request, type FullConfig } from "@playwright/test";
+import {
+  request,
+  type APIRequestContext,
+  type APIResponse,
+  type FullConfig,
+} from "@playwright/test";
 
 const ADMIN_EMAIL = "harness-e2e@tinyhumans.ai";
+const REQUEST_PATH = "/api/v1/company/auth/request";
+const VERIFY_PATH = "/api/v1/company/auth/verify";
 
+/**
+ * Authenticates once and shares the session with every spec through Playwright
+ * storage state, so the suite signs in a single time.
+ *
+ * Every failure here aborts the whole run before a single spec executes, so the
+ * message has to carry enough to diagnose it without a second run: the endpoint,
+ * the status, and the body the host actually returned. `auth/request` answers
+ * `202 {"sent": true}` for *every* outcome by design — it refuses to say whether
+ * an address is a member — so a missing `dev_code` is otherwise indistinguishable
+ * from a broken host, which is exactly what issue #271 sent people chasing.
+ */
 export default async function globalSetup(config: FullConfig) {
   const storageState = process.env.PW_STORAGE_STATE;
   if (!storageState) return;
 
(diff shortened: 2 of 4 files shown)


oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

PR #285, opened 2026-08-02
fix(console): dismiss the alert dialog when its action is confirmed (#268)

### LINKED ISSUE #268: bug: confirming an AlertDialog leaves the modal mounted and blocks the console
## Summary

Confirming an `AlertDialog` leaves the modal mounted. `AlertDialogAction` is a plain `Button` — only `AlertDialogCancel` wraps `AlertDialogPrimitive.Close`. So a confirm handler runs, the underlying action succeeds, and the dialog stays open with its inert backdrop swallowing every subsequent click. The console is effectively frozen until the user reloads.

Found while building #259: confirming a workflow delete left `Delete ""?` rendered over a console whose workflow was already gone — the title read empty because the selected workflow had been removed from state while the dialog it belonged to was still on screen. Fixed on that branch by controlling `open`/`onOpenChange`, and pinned with an explicit `toBeHidden` assertion.

The same uncontrolled pattern is on `main` at three sites:

- `frontend/src/views/TaskDetailView.tsx:855`
- `frontend/src/views/SettingsView.tsx:251`
- `frontend/src/views/TaskEditDialog.tsx:192`

Every confirm flow in the console is likely stuck the same way. Each is a one-line fix.

## Reproduction

1. Console → a view with a destructive confirm (e.g. Task detail → delete).
2. Confirm.
3. The action takes effect, but the dialog remains and the page stops responding to clicks.

## Expected behavior

Confirming closes the dialog. Either control `open` explicitly and set it false in the handler, or wrap the action in `AlertDialogPrimitive.Close` so it behaves like Cancel does today.

Worth a follow-up sweep for any `AlertDialogAction` added since — this is easy to reintroduce, because the component name reads as though it closes.

oxoxDev:
## Summary

Confirming an `AlertDialog` left the modal mounted. `AlertDialogAction` rendered a plain `Button`, while `AlertDialogCancel` rendered `AlertDialogPrimitive.Close` — so only *cancelling* dismissed the dialog. Confirming ran the handler, the underlying action succeeded, and the popup stayed on screen with its backdrop marking the rest of the console inert. Every click after that went nowhere until a reload.

Fixed once in the shared primitive rather than three times at the call sites, since all three inherit the same component: `TaskDetailView`, `SettingsView`, `TaskEditDialog`.

Closes #268

## API Or Behavior Changes

**Behavior.** `AlertDialogAction` now dismisses the dialog when clicked, matching what its name has always implied and what `AlertDialogCancel` already did. No call site changes.

The props it accepts are unchanged in practice — `onClick`, `className`, `disabled`, and the `Button` `variant`/`size` all still reach the rendered button. Its type moves from `ComponentProps<typeof Button>` to `AlertDialogPrimitive.Close.Props & Pick<…, "variant" | "size">`, exactly mirroring `AlertDialogCancel`.

**Handler ordering, deliberately.** Base UI merges event handlers right-to-left and the caller's `onClick` sits to the right of the primitive's own close handler, so the confirm handler runs *first* and the close runs after it. Because every confirm on `main` is fire-and-forget (`onClick={() => void remove()}`), the handler returns at its first `await` and the close still runs synchronously inside the same click — it cannot be outrun by a continuation that later tears the dialog's owner down. The reverse direction is safe too: React batches state updates until the handler returns, so a synchronous unmount could not beat the close either. Nothing the handlers need is unmounted early; on the delete path the surrounding view unmounts only when the request lands, well after the dialog is gone. A handler that wants to *keep* the dialog open (a failed validation, say) can call `event.preventBaseUIHandler()`.

**No call site was working around this.** None of the three controls `open` itself. The one workaround that exists is on an unmerged branch — #279 controls `open`/`onOpenChange` locally in `WorkflowsView` for the same reason. This PR does not touch that file. Once both land the local control is redundant but harmless (the handler's `setConfirmOpen(false)` and the primitive's close drive the same state), so whoever merges second is looking at belt-and-braces, not a conflict. Removing it is a small follow-up, not something to do from here.

## Tests

- [x] `cargo test` — 1015 passed, 0 failed

Frontend: `npm ci && npm run typecheck && npm run build` — clean. There is no Rust diff in this PR; the Rust gates are run as a matter of course.

New Playwright spec, `frontend/test/e2e/alert-dialog-confirm.spec.ts`, driving a local host against the built console. It covers two of the three call sites — the task delete in `TaskEditDialog` and the lifecycle `ConfirmAction` in `SettingsView` — because the point is that one primitive serves all of them. (`TaskDetailView`'s `ConfirmButton` is the same shape but needs a live in-flight run to reach.)

Each test asserts both halves of the bug: the dialog is gone from the DOM, **and** the surface underneath takes clicks again. A popup that has animated out of sight but left an inert backdrop behind is the same bug, so the second assertion clicks through with actionability checks on rather than trusting a visibility check.

Both tests hold the confirm handler's request open, so the assertion lands in the window between the click and the response. That is where the bug lives, and holding it also removes what masked it: on the delete path the console navigates away once the request lands, unmounting the stuck dialog, so a naive "hidden eventually?" check passed even on `main`. The lifecycle test then *fails* its held request rather than releasing it, so the company is never actually suspended and later specs still find it running.

Ran the same spec both ways against a host on the default (offline) build:

| Console bundle | `alert-dialog-confirm.spec.ts` | Full suite |
| --- | --- | --- |
| `upstream/main`, unmodified | **2 failed** — `[data-slot="alert-dialog-content"]` expected 0, received 1, on both tests | 8 failed / 1 passed |
| this branch | **2 passed** | 6 failed / 3 passed |

The 6 failures common to both runs are environmental, not from this change: that host has no mocked LLM backend, which those specs need. They are identical before and after.

Not a CI gate — `playwright.config.ts` declares no `webServer`, so like the rest of `test/e2e` this needs a host brought up by hand. It is a reproduction and a written-down contract.

Also hit the known #271 global-setup flake (`auth/request` not echoing `dev_code`) several times; retried past it, and PR #284 is fixing it.

## Documentation

No doc file changes. Nothing under `docs/` or `frontend/ARCHITECTURE.md` describes the UI primitives — `ARCHITECTURE.md` is about which surfaces need which endpoints — so there was no stale sentence to correct. The contract instead lives as a doc comment on `AlertDialogAction` itself, next to the code that has to keep it, covering why it renders `Close` and why the handler ordering is what makes an async confirm safe. The issue notes this is easy to reintroduce because the component name reads as though it already closes; the comment is aimed squarely at that.

### DIFF
--- a/frontend/src/components/ui/alert-dialog.tsx
+++ b/frontend/src/components/ui/alert-dialog.tsx
@@ -139,14 +139,33 @@ function AlertDialogDescription({
   )
 }
 
+/**
+ * The confirm button. Like {@link AlertDialogCancel}, it renders the primitive's
+ * `Close` so confirming dismisses the dialog — otherwise the popup stays mounted
+ * and its inert backdrop swallows every later click on the app behind it.
+ *
+ * Base UI merges event handlers right-to-left, and the caller's `onClick` sits to
+ * the right of the primitive's own close handler, so the confirm handler runs
+ * *first* and the close runs after it. A handler that wants to keep the dialog
+ * open (a failed validation, say) can call `event.preventBaseUIHandler()`.
+ *
+ * That ordering is what keeps an async handler safe. Every confirm here is
+ * fire-and-forget (`onClick={() => void remove()}`), so it returns at its first
+ * `await` and the close still runs synchronously inside the same click — it can
+ * never be outrun by a continuation that later tears the dialog's owner down.
+ */
 function AlertDialogAction({
   className,
+  variant,
+  size,
   ...props
-}: React.ComponentProps<typeof Button>) {
+}: AlertDialogPrimitive.Close.Props &
+  Pick<React.ComponentProps<typeof Button>, "variant" | "size">) {
   return (
-    <Button
+    <AlertDialogPrimitive.Close
       data-slot="alert-dialog-action"
       className={cn(className)}
+      render={<Button variant={variant} size={size} />}
       {...props}
     />
   )
--- a/frontend/test/e2e/alert-dialog-confirm.spec.ts
+++ b/frontend/test/e2e/alert-dialog-confirm.spec.ts
@@ -1,0 +1,168 @@
+import { expect, test, type Page, type Route } from "@playwright/test";
+
+/**
+ * Regression proof for #268 — confirming an `AlertDialog` must dismiss it.
+ *
+ * `AlertDialogAction` used to render a plain `Button` while `AlertDialogCancel`
+ * rendered the primitive's `Close`, so only *cancelling* closed the dialog.
+ * Confirming ran the handler, the underlying action succeeded, and the popup
+ * stayed mounted — its backdrop marking the rest of the console inert and
+ * swallowing every subsequent click until a reload.
+ *
+ * Both tests hold the confirm handler's request open, because the interesting
+ * moment is the one *between* the click and the response. That window is where
+ * the bug lives, and holding it makes the assertion deterministic rather than a
+ * race: while the request is in flight the dialog must already be gone. It also
+ * removes the thing that used to mask the bug — on the delete path the console
+ * navigates away once the request lands, which unmounts the stuck dialog and
+ * makes a naive "is it hidden eventually?" assertion pass even on `main`.
+ *
+ * The dialog being *hidden* is only half of it. A popup that has animated out of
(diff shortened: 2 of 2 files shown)


oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

Issue #228, opened 2026-07-31
Scheduled workflow runs discard delivery outcomes


oxoxDev:
## Context

Two in-flight changes intersect here, and neither branch can fix this on its own:

- PR #211 (#169) adds a workflow scheduler that fires saved workflows on a trigger cron. It calls `WorkflowRunner::run` and logs the run's completion.
- PR #226 (#170) adds output-node delivery, whose per-destination outcomes ride on the returned `WorkflowRun.deliveries`.

## The problem

The scheduler discards the `WorkflowRun` value it receives. Once both land, a **scheduled** run that delivered nothing — a `denied` row because the company lacks the `email` grant, a `skipped` row because the recipient is cold, a `failed` row because the mail transport refused, or the loud "delivery is not wired" row — will be completely invisible. The run logs as finished and the report silently goes nowhere.

This matters more for scheduled runs than for anything else: an operator-clicked run surfaces its delivery rows in the run drawer, so a failure is visible immediately. A scheduled run has nobody watching, which is exactly why delivery was placed in the shared runner path rather than the HTTP handler.

## Fix

In the scheduler's fire path, include the delivery outcomes in what it logs — at minimum a warning when any row is not `sent`, naming the workflow, the node, and the status. Consider a company-visible surface too (an event, or the operator channel) since a log line on the host is not something a tenant operator reads.

## Why it is filed separately

`WorkflowRun.deliveries` does not exist on #169's branch, and the scheduler does not exist on #170's. Whichever merges second is the first point at which this is fixable.

## Pointers

- `src/runtime/workflow_scheduler.rs` — the `tokio::spawn`ed fire site that matches on `runner.run(...)`
- `src/ports/workflow_runner.rs` — `WorkflowRun.deliveries`, `DeliveryReport`, `DeliveryStatus`

oxoxDev:
Scope update: the **visibility** half of this is now in PR #226 (`9bcd24c`). After that PR's rebase onto merged #169, the scheduler and `WorkflowRun.deliveries` live in one tree, so the constraint that originally forced this into its own issue — neither branch could fix it alone — no longer applies. The scheduler now logs a `warn` per undelivered row (company, workflow, node, kind, target, detail) and per-status counts on the info line, with tests pinning both the warning and the no-warning-on-a-clean-run case.

[...]

Issue #288, opened 2026-08-02
ci: --all-features never compiles the test target, hiding two E0034 errors in sqlite.rs


oxoxDev:
## Summary

`cargo check --all-features --all-targets` fails on `main` with two `E0034` errors. Nothing in CI runs that combination, so the breakage has been sitting on the default branch unnoticed.

## Problem

```
error[E0034]: multiple applicable items in scope
    --> src/store/sqlite.rs:1988:27

error[E0034]: multiple applicable items in scope
    --> src/store/sqlite.rs:2009:27

error: could not compile `opencompany` (lib test) due to 2 previous errors
```

Both sites are a bare `store.list(&id, "")`, and `SqliteStore` implements two traits that each have a `list` method:

- `impl ports::context::ContextStore for sqlite::SqliteStore` — `src/store/sqlite.rs:602`
- `impl ports::store::CompanyStore for sqlite::SqliteStore` — `src/store/sqlite.rs:315`

rustc suggests the fix itself: `ports::context::ContextStore::list(&store, &id, "")`.

## Why CI does not see it

The failing target is **`lib test`**, and it needs the intersection of two things CI never combines:

- `ci.yml`'s `rust-gated` job runs `cargo check --all-features`, which checks lib and bins only. The test target is never compiled under the all-features set.
- `cargo test` in the `rust` job runs default features, where `ContextStore` and `CompanyStore` are not both in scope for `SqliteStore`, so the ambiguity does not arise.

Neither job compiles test code with all features on. That is the gap.

This is the same shape as #202, which added the gated job because the default `Rust` job could not see feature-gated code — and the same shape as #281, where no job compiles the console at all. Each time the lesson is the same: a green check is only evidence for the surface it actually compiles.

## Confirmed pre-existing

Line 2009 is present verbatim at `1423a83` via `git show`. It is not introduced by any branch currently in flight; it was found by a branch that does not touch `sqlite.rs` at all.

## Solution

Two parts, and the second is the one that matters:

1. Disambiguate both call sites with the fully-qualified form rustc suggests.
2. Change the `rust-gated` job's all-features step to `cargo check --all-features --all-targets`, so the test target is compiled under the full feature set and this class of breakage cannot return silently.

On (2): `--all-targets` here means *compile*, not *run*. The existing job already comments that its `--all-features` step is check-only by design, because that set enables `mcp` and would change tool-belt expectations in tests that pin them. Compiling the test target does not run it, so that constraint is unaffected.

[...]

PR #289, opened 2026-08-02
fix(workflows): log a delivery's failure class, not the transport's reply (#248)

### LINKED ISSUE #248: fix(workflows): transport error text can carry a recipient address into host logs
## Context

PR #226 (#170) added output-node report delivery and, in the scheduler, a warning
that names each report a scheduled run failed to deliver. During review that
warning was found to log `report.target` directly — the recipient's address on an
`email` destination — and it was scrubbed before merge, replaced with a
non-sensitive `target_configured` boolean.

CodeRabbit agreed on that thread that a second, indirect path remains and is worth
tracking separately. This is that follow-up.

## The problem

The scheduler's undelivered-report warning still logs `detail`:

- `src/runtime/workflow_scheduler.rs:324` — `detail = %report.detail`

On most paths `detail` is static prose written by us ("this recipient has never
written to the company…"). On the transport-failure arms it is not — it
interpolates the raw transport error:

- `src/workflows/delivery.rs:366` — `format!("the mail transport refused the message: {err}")` (the `email` destination)
- `src/workflows/delivery.rs:259` — the same string on the `owner` destination, where the address is a company admin's
- `src/workflows/delivery.rs:535` — `format!("the channel refused the message: {err}")` (the `channel` destination)

An SMTP rejection routinely quotes the address it rejected — a 550 or 553 reply is
commonly of the form `550 5.1.1 <ada@example.com>: Recipient address rejected`. So
on a scheduled run whose delivery fails at the transport, the recipient's address
reaches host stdout through the error string even though the `target` field is now
scrubbed.

## Why it matters

This line goes to the host's stdout. On a self-hosted deployment that is the
operator, who is entitled to the address. On a hosted tenant it is the platform,
which is the distinction `src/runtime/workflow_scheduler.rs` module docs already
draw. Scrubbing `target` while leaving `detail` closes the direct path and leaves
the indirect one, which is the harder one to notice later.

## Why it is not simply "drop `detail`"

`detail` is the operator-actionable half of that warning — it is the part that says
what to fix, and the existing test asserts it for exactly that reason. Blanking it
would make transport failures undiagnosable, trading one problem for another. The
fix wants to *sanitize*, not remove.

## Proposed direction

Options, in rough order of preference — worth a decision rather than a default:

1. **Sanitize on the way to the log.** Strip address-shaped substrings out of the
   transport error before it reaches `tracing`, keeping the status code and reason
   text. Preserves diagnosability; needs a matcher that is not itself a source of
   surprises.
2. **Map transport failures to a closed set of safe prose** at the point the
   `DeliveryReport` is built ("the recipient address was rejected", "the mailbox is
   over quota"), keeping the raw text only on the `DeliveryReport` returned to the
   operator over HTTP, where the operator is entitled to it, and never on the
   host-log path.
3. **Split the field** — a `detail` for the operator response and a `log_detail`
   that is always safe.

Option 2 keeps the on-demand run's drawer as informative as it is today, which is
probably the property worth protecting.

## Acceptance criteria

- A scheduled run whose delivery fails at the transport emits a warning that names
  the company, workflow, node, kind and status, and does **not** contain the
  recipient address, on the `email`, `owner` and `channel` arms.
- The warning still carries enough to diagnose the failure class.
- The `DeliveryReport` returned on an on-demand run is unchanged, or changed
  deliberately with the drawer's usefulness preserved.
- A regression test asserts the address is absent from the emitted line for a
  transport failure, mirroring the existing `target` guard in
  `a_scheduled_run_reports_an_undelivered_report`.

## Pointers

- `src/runtime/workflow_scheduler.rs:318-326` — the warning, including the already-scrubbed `target_configured` field
- `src/workflows/delivery.rs:259`, `:366`, `:535` — the three sites that interpolate a transport error into `detail`
- `src/ports/workflow_runner.rs` — `DeliveryReport`, `DeliveryStatus`
- PR #226 discussion: https://github.com/tinyhumansai/opencompany/pull/226#discussion_r3695344171

oxoxDev:
## Summary

The workflow scheduler's undelivered-report warning logged `report.detail`. On
most paths that is static prose this codebase wrote, but on the transport-failure
arms it interpolates the transport's own words — and a mail transport quotes the
mailbox it refused (`550 5.1.1 <someone@example.invalid>: Recipient address
rejected`). That line goes to host stdout, which on a self-hosted deployment is
the operator but on a hosted tenant is the platform. PR #226 already removed
`report.target` from this line and replaced it with a `target_configured`
boolean; `detail` was the remaining path by which a recipient address could
reach the same place.

Rather than filter the string on its way to `tracing`, this splits the row's
reason in two so the loggable half is loggable by construction:

- `DeliveryReport::detail` — unchanged. Free text for the operator's own
  surfaces. It still quotes the transport verbatim, which is what makes a
  refused send fixable.
- `DeliveryReport::reason` — new. A `DeliveryReason` enum: a closed set of
  outcomes with **no `String` payload**, so no `format!` can produce one and no
  transport-supplied text can reach it without changing its type. This is the
  only half the scheduler logs.

Both are set at the same construction site, so the classification is made where
the outcome is known instead of being recovered later by pattern-matching prose.
A regex over the transport reply was the alternative and was rejected: the input
is unbounded (every MTA words a refusal differently), a matcher tuned to today's
replies rots quietly, and a near-miss leaks the whole line rather than failing
loudly.

### Sites covered

This is the family, not the one line.

- `runtime/workflow_scheduler.rs` — the undelivered-report warning now carries
  `reason` instead of `detail`. The only behaviour change in the PR.
- `workflows/delivery.rs` — all three arms that interpolate a transport error
  (`owner` mail, `email` mail, `channel` post) plus the `owner`
  operator-channel fallback now classify where the row is built.
  `post_to_channel` returns `Result<(), (DeliveryReason, String)>` so the two
  channel failure shapes — "you named a channel that does not exist" and "the
  channel said no" — stay distinguishable without the caller parsing a message.
- `brain/medulla/effects.rs` and `harness/orchestrator.rs` — the journal's two
  non-tenant readers. Both already fold a finished run to counts and exclude
  `target`/`detail`; #228 wrote them that way. Neither exclusion was pinned by a
  test, so both now are. An unpinned rule is one refactor from not being true.

### Why `detail` keeps the address, and the check behind that

`detail` is only safe to leave alone if its readers are genuinely tenant-scoped,
so this was traced rather than assumed. `WorkflowRunFinished` is appended
through `EventLog` to a **per-company** `companies/<slug>/events.jsonl` — the
journal is shared across event kinds (chat, audit, run history) within one
company, not across companies. Reading the full rows back:

| Reader | Guard |
|---|---|
| `GET …/workflows/runs` | `ScopedCompany` |
| operator SSE `GET {scope}/events` | `ScopedCompany` |
| manual run HTTP response | `ScopedCompany` |

`ScopedCompany` resolves the company, then runs `authorize_address`: a user
principal must belong to that company; a platform principal must either hold
platform scope or own the company and be permitted to address it. A
platform-scope token can therefore read these rows — deliberately, and the same
token already reads that company's chat history, memory and manifest. That is a
credentialed administrative boundary, and a different thing from a log line on
host stdout, which needs no credential at all. The two readers that are *not*
tenant-scoped — the inference-sidecar wire-out and the orchestrator's insight
summary — never saw `detail`, and now have tests saying so.

So the split stands, and the companion test pinning that `detail` keeps the
address is pinning the intended behaviour rather than a defect.

Sibling sites reviewed and deliberately left alone: the remaining
`error = %err` interpolations in the delivery path (a park failure, the
approval-gate rollback, an unreadable user directory, a failed outbound-mail
record) carry store, queue or serialization errors rather than transport
replies, so none can quote a recipient. The scheduler's `scheduled run failed`
warning carries a runner error, and a delivery failure never fails a run, so a
transport reply cannot reach it either.

Closes #248

## API Or Behavior Changes

**Behavior.** One log line changes. For a scheduled run whose delivery the mail
transport refuses, the warning was:

```
... report was NOT delivered company=acme workflow=digest node=owner_summary
kind=email target_configured=true status=Failed detail=the mail transport
refused the message: 550 5.1.1 <someone@example.invalid>: Recipient address
rejected: User unknown in local recipient table
```

and is now — copied from the event the new test captures:

```
2026-08-02T20:11:38.553244Z  WARN opencompany::runtime::workflow_scheduler:
workflow scheduler: a scheduled run's report was NOT delivered
company=transport-refusal-co workflow=digest node=owner_summary kind=email
target_configured=true status=Failed reason=the mail transport refused the
message
```

Everything needed to route the problem is still there — which company, which
workflow, which node, which destination kind, that a target resolved at all, the
status, and the failure class. What is gone is the transport's own sentence,
which is the only part that could name a person. That sentence still reaches the
operator in full, through the run response and through the
`WorkflowRunFinished` event their own console reads back (#228).

**API.** Additive and backward compatible. `DeliveryReport` gains `reason` on
the wire (`"reason": "mail-transport-refused"`). It is `#[serde(default)]`, so a
`WorkflowRunFinished` event journaled before this change still deserializes, as
`"unspecified"`. The console's `DeliveryReport` type declares it optional;
nothing renders it, because `detail` remains strictly more informative on a
tenant surface. No route, status value, or existing field changed.

## Tests

- [x] `cargo test` — 1012 passed, 0 failed

Also run, because the default feature set compiles neither
`workflows/delivery.rs` nor `harness/orchestrator.rs` and would have hidden
every change in them:

- [x] `cargo check --features openhuman,mcp,telegram --all-targets`
- [x] `cargo test --features openhuman,mcp,telegram` over the touched modules
- [x] `cd frontend && npx tsc -b --noEmit`

New tests:

- `a_transport_failure_does_not_log_the_address_the_transport_quoted`
  (`runtime::workflow_scheduler`) — the regression guard the issue asked for,
  mirroring the existing `target` guard. It drives a scheduled run whose
  delivery row carries a realistic SMTP refusal with the address embedded, then
  **asserts over the emitted `tracing` event** — the captured subscriber output,
  not an intermediate string — that the line contains neither the address, nor
  its local part alone, nor any of the transport's text, while still naming the
  workflow, node, kind, status and failure class.
- `the_operator_facing_detail_keeps_the_transport_text`
  (`runtime::workflow_scheduler`) — the other half, pinned so a later "scrub it
  everywhere" sweep has to argue with a test.
- `a_refusal_that_quotes_the_address_keeps_it_out_of_the_loggable_half`
  (`workflows::delivery`) — the same property at the source, through the real
  `deliver_outputs` path with a sender that refuses the way an MTA does.
- `a_finished_run_wires_out_counts_without_the_recipient_or_the_transport_text`
  (`brain::medulla::effects`) — the journal boundary to the inference sidecar.
- `a_finished_run_summarizes_to_counts_without_the_recipient_or_transport_text`
  (`harness::orchestrator`) — the same for the insight summary a model reads.
- `channel_that_is_not_wired_fails_with_the_wired_list` and
  `a_send_failure_is_reported_and_does_not_abort_delivery` extended to assert
  the classification, including that a channel id stays off the loggable half.

Fixtures use `example.invalid` (RFC 2606 reserved), so a fixture that ever
escapes into a log names nobody.

## Documentation

- `docs/modules/server/README.md` — the delivery section now states which of the
  two reasons is logged and why.
- Module docs updated where the rule has to be found: `workflows::delivery`
  gains a "Two reasons per row" section, and `runtime::workflow_scheduler`'s
  existing note about host stdout being a platform surface now spells out what
  that permits the warning to carry.
- `DeliveryReason` and both `DeliveryReport` reason fields carry doc comments
  naming which reader each half is for, including the enumerated read paths
  behind `detail`.

## Summary by CodeRabbit

* **New Features**
  * Added structured delivery outcome reasons for workflow runs.
  * Delivery reports now retain operator-facing details while providing stable, log-safe reason labels.
  * Older saved workflow data remains compatible.

* **Bug Fixes**
  * Prevented recipient addresses and transport error text from appearing in scheduler and host logs.
  * Preserved detailed delivery information in operator-facing workflow views.

* **Documentation**
  * Clarified the separation between operator-facing delivery details and safe scheduler log reasons.

### DIFF
--- a/docs/modules/server/README.md
+++ b/docs/modules/server/README.md
@@ -73,6 +73,14 @@ diagnosable in the host's stdout, which is not the same as operator-visible.
 Surfacing those outcomes is issue #228; the durable record it needs is the
 first-class `Run` tracked by issue #242.
 
+That distinction decides what the scheduler's log line may say. Every row
+carries two reasons: `detail`, the free text the run response and the console
+render, and `reason`, a closed set (`DeliveryReason`). Only `reason` is logged.
+On the transport-failure arms `detail` interpolates the transport's own reply,
+and a mail transport quotes the mailbox it refused — so `detail` on host stdout
+would put a recipient's address on a platform surface. `reason` says what class
+of thing failed and has no field that could carry the address (issue #248).
+
 Authoring a destination and reading the result back:
 
 Both routes go through `ScopedCompany`, so both need an operator credential —
--- a/frontend/src/api/workflows.ts
+++ b/frontend/src/api/workflows.ts
@@ -112,8 +112,24 @@ export interface DeliveryReport {
   /** The address or channel actually addressed, when there was one. */
   target?: string;
   status: DeliveryStatus;
-  /** An operator-readable reason — populated even on success. */
+  /**
+   * An operator-readable reason — populated even on success. This is the half
+   * the drawer renders: it may quote the transport verbatim, which is what
+   * makes a refused send fixable, and this console is a tenant surface.
+   */
   detail: string;
+  /**
+   * The same outcome as a stable token out of a closed set (issue #248) —
+   * `"mail-transport-refused"`, `"channel-not-wired"`, and so on.
+   *
+   * It exists because the host's own logs may not carry `detail`: on the
+   * transport-failure arms `detail` interpolates the transport's reply, which
+   * routinely quotes the recipient's address, and host stdout is a platform
+   * surface rather than a tenant one. Nothing renders it today — `detail` is
+   * strictly more informative here — so it is optional on the type (not the
+   * wire), and a response from a host predating #248 still parses.
+   */
+  reason?: string;
 }
 
 /** The result of a run: the engine's final state and any pending approvals. */
--- a/src/brain/medulla/effects.rs
+++ b/src/brain/medulla/effects.rs
@@ -158,6 +158,11 @@ pub(crate) fn wire_event(seq: u64, event: &CompanyEvent) -> WireEvent {
         // recipient's email address and its detail can quote one, and this body
         // is wired out to the inference sidecar; the console reads the full
         // rows from the journal instead, where they belong to the operator.
+        //
+        // Issue #248 pinned this with a test rather than leaving it a comment:
+        // the exclusion is the journal-boundary half of the same rule the
+        // scheduler's log line follows, and an unpinned rule is one refactor
+        // away from not being true.
         CompanyEvent::WorkflowRunFinished {
             workflow_id,
             scheduled,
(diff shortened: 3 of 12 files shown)


oxoxDev:
The pre-existing `E0034` finding in the last section already has an issue: **#288**. Same diagnosis independently — `ContextStore::list` vs `CompanyStore::list` ambiguity in test code, invisible because CI runs `--all-features` without `--all-targets`. No need to file a second one.

PR #293, opened 2026-08-02
fix(delegation): ground the delegation target and make an undelivered hand-off visible (#272)

### LINKED ISSUE #272: bug: delegation invents desk ids and an unknown one fails silently
## Summary

When the orchestrator delegates, it invents desk ids that do not exist, and an invented one fails silently.

Observed on a live harness company whose desks are `engineering`, `content` and `legal`. Asked to hand work to a desk without naming an id, the orchestrator answered:

> "[ceo] Handed off to the **writer** desk — their lead will draft the release note and respond this turn."

`writer` is not a desk. It is an *agent*, and a member of the `content` desk. Per #213, a hand-off to a desk with no resolvable lead "settles under the delegator exactly as before" — so the card landed in `in_review` assigned to `ceo`, with a note claiming a hand-off that never happened.

Naming the desk explicitly in the same company works correctly: with "desk id: content" in the instruction, the card came back assigned to `writer`, the content desk's lead, with the delegate's real output in the note.

So the delegation machinery is fine. What is missing is grounding — nothing constrains the delegation target to the actual roster, and nothing tells the operator when the target was rejected.

## Why this matters more than a bad guess

The silent-failure shape is the problem. A delegation to a non-existent desk is indistinguishable, on the board, from a delegation that simply produced nothing:

- the note says the work was handed off
- the card is assigned to the delegator
- no error surfaces anywhere

An operator reading the board sees a plausible hand-off and an unworked card, with nothing connecting the two.

## Expected behavior

Two parts, and both are worth doing:

**Ground the target.** `delegate_to_desk` should be constrained to the company's real desk ids — the roster is already queryable, and the orchestrator already has `query_company` for exactly this kind of grounding. Rejecting an unknown id at the tool boundary with the valid set in the error would let the model correct itself in-turn.

**Make the fall-through loud.** When a hand-off cannot resolve a lead, the card should say so — in the note, and ideally in the run trail — rather than settling under the delegator with a note that claims otherwise. Settling under the delegator is the right *behaviour* (#213 chose it deliberately to avoid stranding cards); it just needs to be visible.

## Related

The same class of ungrounded action is why #267 proposes triaging an incoming message before it becomes board activity.

oxoxDev:
## Summary

Closes #272.

When the orchestrator hands work to a desk it never had to name a desk that
exists. It invented one, the invented one was accepted, and the hand-off then
failed with nothing anywhere saying so.

**Reproduced first, on a live host.** A harness company with desks
`engineering`, `content` and `legal`; `writer` is a *teammate* who leads
`content`. A dispatched board card whose turn calls
`delegate_to_desk(desk: "writer")` produced, before this change:

- tool result: `Delegated to the writer desk. Its lead will answer this turn.`
- the delegation queued and drained (`queued=1`), and no delegate turn ever ran
- the card settled in `in_review`, `assignee: ceo`, note:
  `[ceo] Handed off to the **writer** desk — their lead will draft the release note and respond this turn.`

which is the report verbatim. On the board a delegation to a desk that does not
exist is indistinguishable from a delegation that produced nothing.

**Two changes, matching the two halves of the issue.**

*Ground the target.* `delegate_to_desk` now resolves `desk` against the
company's real desks before anything is queued, and refuses what it cannot
deliver — an id that matches no desk, and a desk with no member on the roster,
which no turn could ever run for. The refusal names the valid desk ids, and when
the invented target is a teammate it says so and points at the desk that
teammate is on:

```
There is no "writer" desk — "writer" is a teammate, not a desk.
The desk they are on is "content". Valid desk ids: engineering, content, legal.
```

`query_company` — the surface the orchestrator's own brief tells it to consult —
listed the roster but not the desks, so there was no authoritative place to read
a desk id in the first place. It now has a `## Desks` section listing each desk
by the id `delegate_to_desk` takes, with its lead, and marking a desk nobody
leads as unable to take work. That is the preventive half; the refusal is the
corrective half.

*Make the fall-through loud.* Settling under the delegator stays exactly as
#213 chose it — a card is never stranded waiting on a delegate that cannot run.
It is only no longer silent. A dispatched card now records the attempt:

```
[ceo] hand-off to the "writer" desk was not delivered — it is not a desk this
company can hand work to. Nothing was delegated; this card is still with ceo.
```

This is written from a refusal recorded on the delegation queue, **not** from
the turn's own account of itself — the turn is free to keep claiming a hand-off
happened, and on the live company it did exactly that. The same note covers the
other route to an undelivered hand-off: a desk that had a lead when the tool
accepted it and lost it before the drain.

## API Or Behavior Changes

- **`delegate_to_desk` can now fail.** An unknown desk, or a desk with no
  roster-backed lead, returns a `ToolResult::error` and queues nothing. It
  previously always succeeded. `delegate_to_desk` is already in
  `harness::steps::INTRINSIC_TOOLS`, so the OC-authored refusal reaches the
  operator's run trail verbatim (bounded to 200 chars) as well as the model.
- **Fails open.** If the company record cannot be read, the hand-off is queued
  ungrounded exactly as before. A store hiccup does not take delegation offline.
- **`DelegateToDeskTool::new` and `orchestrator::delegation_tools` take the
  company id + store.** The store is read at call time, never snapshotted at
  agent-build time — see the trade-off below.
- **`DelegationQueue` carries refused desk keys** alongside queued delegations,
  drained by `handle_task_delegations` and wiped by the same `clear()`.
- **`query_company` gains a `## Desks` section** and a `desks` count in its JSON
  payload. Additive; no existing section changed.
- **Hosted path**: the unknown-desk tool error now carries the valid desk ids
  instead of only `no desk matches "x"`. A real desk with no lead is deliberately
  *not* refused there — a hosted hand-off is a durable card assigned to the desk,
  visible on the board whether or not anyone leads it yet, so nothing is silent.
- No REST/GraphQL schema changes, and no frontend changes.

## Choices that differ from the issue, and why

**Schema enum vs. tool-boundary rejection.** The issue proposes rejection; I
checked first whether the schema could carry the desk ids as an enum so the
model is constrained rather than corrected. It cannot, for two reasons worth
recording: `Tool::parameters_schema` is synchronous, so it cannot read the
company store, and desks are dynamic per company — an operator can create one
mid-session through the desk overlay. Any enum baked at agent-build time would
go stale and refuse a desk that genuinely exists, which is a worse failure than
the one being fixed. Rejection at the boundary, reading the store fresh on each
call, is the version that stays correct.

**No new `CompanyEvent` variant.** The trail loudness the issue asks about is
already there: `delegate_to_desk` is an intrinsic tool, so its own refusal text
is surfaced on the run trail verbatim rather than collapsed to a generic class.
Against that, `CompanyEvent` is documented in `runtime/journal.rs` as "a closed,
binding enum with no marker variants", it is projected in several places, and
the brain-agnostic `DelegationRunner` holds no event-log handle — so a variant
would mean threading one across that seam for something the trail and the card
note already carry. If a reviewer wants the event anyway it is a small follow-up,
not a rewrite.

**Refusing a leadless desk is mine, not the issue's.** The issue asks only that
unknown ids be refused. A desk with no roster member can never run a turn, and
the tool's success message ("its lead will answer this turn") is exactly the lie
this issue is about, so it is refused too — with a distinct message, and only on
the harness path, where a hand-off is a live turn rather than a durable card.

## Does the in-turn self-correction actually work?

**The mechanism does, verified live** rather than assumed. Driving a real host
with a scripted OpenAI-compatible endpoint, so the model's choices are
deterministic and the rest of the path is real: the orchestrator calls
`delegate_to_desk(desk: "writer")`, gets the refusal back as a failed tool result
inside the same turn, calls `delegate_to_desk(desk: "content")`, and the hand-off
completes — card reassigned to `writer` (the content desk's lead) and settled on
the delegate's own output. The loop does feed the error back, and does act on a
corrected retry.

**What that does not prove** is that a real model chooses to correct itself. The
scripted endpoint decides to retry; a live model might instead answer in prose
without retrying — which the second scripted run exercises, and which also lands
correctly: the refusal recorded on the card, no phantom hand-off. Both outcomes
are honest on the board now, so the fix does not depend on the model cooperating.
That is precisely why the loud note is written from the recorded refusal rather
than from what the turn says about itself.

## Tests

New coverage:

- `reject_desk_target` / `unknown_desk_message`: the valid set in the refusal,
  the teammate-is-not-a-desk case, the leadless-desk case offering only desks
  that can take work, a company with no desks, and the desk-list cap.
- `delegate_to_desk_tool_refuses_a_desk_that_does_not_exist` — nothing queued,
  valid ids and the teammate hint in the error, refusal recorded for the drain.
- `delegate_to_desk_tool_refuses_a_desk_with_no_roster_lead`.
- `delegate_to_desk_tool_queues_ungrounded_when_no_record_is_readable` — the
  fail-open contract.
- `query_company_tool_lists_the_desks_delegation_accepts` — each desk by id with
  its lead; a leadless desk says it cannot be handed work.
- `a_refused_hand_off_is_recorded_on_the_card` — the card carries the refusal
  even though the turn claimed a hand-off; still settles under the delegator.
- `a_hand_off_that_cannot_be_delivered_says_so_on_the_card` — the drain-time
  route (a lead lost between the tool call and the drain).
- `a_spawn_task_never_records_an_undeliverable_hand_off` — a delegation with no
  desk target must not pick up the line.

The existing #213 pins (`a_hand_off_to_an_unknown_desk_settles_under_the_delegator`
and the cancellation / card-ownership set) are untouched and still pass — the
settle behaviour is unchanged.

Commands run locally:

- [x] `cargo build --all-targets` — via `cargo check --all-targets` on both
      feature sets below, plus a full `cargo build --features openhuman,mcp,telegram`
      for the live host
- [x] `cargo test` — 1015 + 6 pass
- [x] `cargo test --features openhuman` — 1397 + 6 pass (the harness suite is
      where this code lives; the default build does not compile `src/harness/`)
- [x] `cargo check --features openhuman,mcp,telegram --all-targets`
- [x] Live verification on a running host (isolated `--home`, harness company with
      `engineering` / `content` / `legal`): bug reproduced before the change, then
      all three post-change paths — invented desk refused, leadless desk refused,
      corrected retry delivering to the right lead.

N/A: no `frontend/` files are touched by this diff, so the console gates
(`npm run typecheck` / `npm run build`) were not run.

## Documentation

No doc file names `delegate_to_desk` or the delegation contract — those decisions
live in rustdoc, which is where this codebase keeps them (as with #204 and #213).
The rustdoc on `reject_desk_target`, `DelegateToDeskTool`,
`DelegationQueue::refused` and `undeliverable_handoff` records the grounding
rules, the fail-open contract, the read-the-store-at-call-time reasoning, and why
the hosted path refuses only the unknown-desk half. The orchestrator persona
brief and the `query_company` tool description are updated to point at the desk
list.

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -3237,6 +3237,10 @@ members = ["eng1", "eng2"]
         /// Invokes that CANCEL their own in-flight delegation mid-run, so the
         /// delegated reply is discarded exactly as an operator cancel does.
         cancel_on: Vec<usize>,
+        /// Desk keys the first turn's `delegate_to_desk` calls named and the
+        /// tool REFUSED (issue #272). A refusal never becomes a `Delegation`,
+        /// so this is how a test reproduces one without standing up the tool.
+        refused_on_first: Vec<String>,
     }
 
     impl DelegatingProvider {
@@ -3283,6 +3287,11 @@ members = ["eng1", "eng2"]
             self.board.lock().unwrap().push(snapshot);
             for delegation in self.pushes.lock().unwrap().pop_front().unwrap_or_default() {
                 self.queue.push(delegation);
+            }
+            if invoke == 1 {
+                for desk in &self.faults.refused_on_first {
+                    self.queue.push_refusal(desk.clone());
+                }
             }
             let message = request
                 .messages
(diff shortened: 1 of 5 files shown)


oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

PR #298, opened 2026-08-03
feat(chat): wire the chat-to-card edge in both directions (#246)

### LINKED ISSUE #246: fix(chat): close the chat-to-card provenance and visibility gaps left by #126
## Summary

#126 shipped orchestrator-spawned task cards from chat, but the chat↔card edge is still missing in both directions: the REST create path drops provenance entirely, and a card the orchestrator opens during a turn is invisible in the transcript. Requests typed into desk and DM threads structurally cannot become cards at all.

## Problem

Three verified gaps against HEAD:

1. **REST create drops provenance.** `src/server/ops/tasks.rs:215` hardcodes `origin_chat_id: None`. The tool-spawn path does stamp it (`src/runtime/delegation.rs:285`), so the field works — the REST surface just never sets it.
2. **A spawned card surfaces nothing.** `run_delegation`'s `SpawnTask` arm returns `DelegationOutcome::default()`, documented as surfacing nothing to relay or bubble. The wire `OutboundMessage` carries no task id, so the operator gets a reply with no indication a card was opened.
3. **Only the orchestrator can spawn at all.** Desk leads and DM-routed roster agents carry no delegation tools (`src/runtime/delegation.rs:258`, "no sub-agent re-delegation in v1"), and `HarnessBrain::responder_for` (`src/harness/brain.rs:454`) routes desk and DM threads to those agents. So a request typed into a desk thread cannot become a card by any path.

Steps to reproduce:

1. Type an actionable request into a desk or DM thread.
2. No card appears, and there is no affordance to make one.
3. Repeat in the orchestrator thread: a card may be spawned by judgement, but the reply says nothing about it and the card has no link back.

## Solution

1. **Explicit operator action as the headline.** A per-message "Add to board" hover action in the transcript, creating a card via the existing REST `createTask`. It works on **every** thread — desk, DM, orchestrator — because it goes through REST rather than the responder's toolbelt, which makes the V1 guarantee true in one click.
   Rejected: NLU auto-detection (false positives litter the board, an inference call per message, non-deterministic) and widening `spawn_task` into desk/DM toolbelts (breaches the v1 depth-1 delegation design and collides with #204).
2. **Land in intake, never `in_progress`.** Dropping into `in_progress` **is** the dispatch trigger (`src/company/runtime.rs:346`) — it spends money. The create call **omits `column`** so the server default decides, keeping the human drag as the spend gate. When #206 introduces its To-do intake lane, chat-created cards follow the server default with no change here. `assignee` is likewise omitted (#205's territory).
3. **Card → chat link.** `TaskRecord.origin_chat_id` already exists. Add additive `originChatId` to the create body (Rust and TS), stamp it instead of the hardcoded `None`, and echo it in the DTO.
4. **Chat → card link, reusing `AgentReply.task_id`.** `DelegationOutcome` gains `spawned_task: Option<{id, title}>` so the brain stamps the journaled `AgentReply.task_id` — today stamped only on dispatch turns. Bonus: the per-task timeline filters the journal on `task_id`, so the reply that *created* the card appears in the card's lineage for free. Multi-spawn turns stamp the first id (documented limitation; the steps timeline still shows every call).
5. **What the operator sees.** Explicit path: a "Task created: *title*" chip linking to `#/tasks/<id>`, plus a toast in the board's vocabulary; failures use the existing error pattern without losing the draft. Orchestrator path: `OutboundMessage` gains additive `taskId`, and bubbles carrying it render a "card opened" chip. History DTOs project `taskId` from `AgentReply.task_id` in **both** REST and GraphQL, per the #65 shared-projection rule, so the chip survives reload.

## Acceptance criteria

- [ ] Every transcript message — desk, DM and orchestrator threads — offers an "Add to board" action that creates a card titled from the message, with the full text in the note.
- [ ] A chat-created card lands in the board's intake lane, unassigned, and does **not** dispatch: no agent turn runs until it is moved to `in_progress`.
- [ ] The created card persists `originChatId` for the originating thread, and task detail can read it back.
- [ ] After creation the transcript shows a confirmation chip linking to the card; failure surfaces the existing error pattern without losing the draft.
- [ ] When the orchestrator spawns a card during a chat turn, the reply bubble carries a "card opened" chip, live **and** after a transcript reload, with REST and GraphQL history agreeing.
- [ ] The journaled `AgentReply` for a card-spawning turn carries the new card's `task_id`, and the card's timeline includes that reply; pre-existing journals round-trip byte-identical.
- [ ] No changes to board lanes, the `+` button, assignee validation, or the dispatch/review lifecycle.

## Notes and scope

**False positives:** none on the explicit path, and the orchestrator judgement path is deliberately unchanged, so there is no *new* surface where idle chat becomes cards. The trade-off is that unaided auto-conversion stays judgement-quality — accepted, because auto-classification risks board litter and per-message inference cost.

**Instant-dispatch spend** is prevented structurally: creation always lands via the server's intake default, and `in_progress` entry remains the only dispatch trigger. A regression test asserts a chat-created card fires no `TaskDispatched`.

**Free-form `column`:** there is no enum, so omitting `column` client-side pins behaviour to one server-side default. If #206 renames the intake lane while lanes stay client-hardcoded (`frontend/src/lib/tasks-sample.ts:21`), a card could land in a column the board doesn't render — mitigated by a single-default constant and a heads-up on #206.

**`AgentReply.task_id` semantic widening:** a "creating" reply now appears in the card's timeline alongside dispatch replies. Accepted as lineage signal and documented on the field.

Scope: **medium**.

## Related

Completes #126, which opened dashboard task cards for actionable chat requests but left provenance and visibility unwired. Builds on #185 (per-task event correlation) and #65 (shared history projection).

Deliberately stays out of the way of work currently in flight: **#206/#215** (board columns, `+` button, auto-assign on move), **#205/#214** (assignee linking and validation), **#204/#213** (delegated dispatch). This issue adds a creation entry point and changes no lanes, no `+` button and no assign-on-move. If #206 lands first, the server default absorbs it with no retarget.

oxoxDev:
## Summary

Closes #246.

#126 opened task cards from chat but left the chat↔card edge unwired in both
directions. This closes it:

1. **`POST …/tasks` carries the thread a card came from.** It hardcoded
   `origin_chat_id: None` while the tool-spawn path stamped the same field
   correctly, and no DTO projected it — so a card opened from a conversation had
   no way back to it, and the #151 "answer where you were asked" post-back could
   never fire for anything the REST surface created. The create now accepts an
   optional `originChatId`, and both reads (board list, task detail) hand it
   back.

2. **A chat turn that opens a card says so.** `run_delegation`'s `SpawnTask` arm
   returned `DelegationOutcome::default()` — the card landed on the board and
   the reply carried nothing tying the two together. It now reports the card's
   id, the harness brain stamps it onto the operator bubble
   (`OutboundMessage.task_id`), and the chat cycle journals it onto that turn's
   `AgentReply.task_id`, so the console's chip survives a transcript reload.

3. **Any message, in any thread, can become a card.** Only the orchestrator
   carries the delegation tools, so a request typed into a desk or DM thread
   could not become a card by *any* path. A per-message "Add to board" action
   creates through REST rather than through the responder's toolbelt — which is
   what makes it true on every thread without widening the v1 depth-1 delegation
   design.

### What I verified against the issue, and what had gone stale

The issue is unusually precise; three of its claims held exactly, three had
aged:

| Claim | Status |
|---|---|
| REST create hardcodes `origin_chat_id: None` | **Held** (line ref drifted; the code was as described) |
| `SpawnTask` arm returns `DelegationOutcome::default()`, wire carries no task id | **Held** |
| Desk leads / DM-routed agents carry no delegation tools | **Held** |
| "there is no enum [for `column`]" | **Stale.** #205 landed `BOARD_COLUMNS` + a write-boundary `validate_column`; an unrenderable column is now a `400`, not a silently-persisted string. The issue's "Free-form `column`" worry — that a chat-created card could land in a lane the board does not render — is obsolete, and the mitigation it proposes (a single-default constant plus a heads-up on #206) is unnecessary. |
| "When #206 introduces its To-do intake lane…" | **Stale.** #206 has landed. `COLUMN_TODO` is already the create default, so "omit `column` and let the server decide" resolves to the intake lane **today**. The spend gate is therefore provable now rather than pending on a future issue — which is why the regression test below is a real assertion and not a placeholder. |
| `frontend/src/lib/tasks-sample.ts` lanes are client-hardcoded | Still hardcoded, but the file now documents the backend as the source of truth and the write boundary rejects any mismatch, so the "card lands in a column the board can't render" risk is closed from the other end. |

### The spend-gate proof

The create call omits `column` on purpose, so this rests on the server default
not being the dispatch trigger. That is two facts in two modules, so a change to
either alone would silently open the gate. Both are now pinned:

- `create_task` defaults an omitted `column` to `COLUMN_TODO`.
- `CompanyRuntime::upsert_task` edge-fires a dispatch only through
  `task_enters_in_progress`, and only on entry into the literal `in_progress`.
- `the_column_a_chat_created_card_defaults_to_does_not_dispatch` asserts
  `!task_enters_in_progress(None, COLUMN_TODO)` **plus** a positive control
  (`task_enters_in_progress(None, COLUMN_IN_PROGRESS)`) so the test cannot pass
  by the trigger simply never firing.
- `a_chat_created_card_lands_off_the_dispatch_trigger` asserts the same at the
  HTTP boundary, and that no `TaskDispatched` reaches the journal.

Honest caveat on that last one: in the default (harness-less) build
`dispatch_task` is already a no-op, so the journal assertion is belt rather than
braces — it would catch a create that started journaling a dispatch of its own.
The load-bearing assertion is the landing column, tied to the same predicate the
runtime dispatches on.

### Where I chose differently from the issue

**Multi-spawn turns — first, not "whichever won."** The issue accepts stamping
only the first task id. I kept that (the journal field is a single optional id;
widening it would break the byte-identical round-trip every stored reply relies
on) but made *first* an explicit contract rather than an accident of iteration
order, with a test pinning it. The claim "this reply opened that card" is then
incomplete but never false, and the bubble's step timeline still shows every
spawn. Stamping the last would have made the reported card depend on the model's
queue order, which is not a contract anyone can rely on.

**The wire carries the id only, not `{id, title}` — the issue's own design
would have failed the issue's own acceptance criterion.** #246 proposes
`spawned_task: Option<{id, title}>` and, separately, requires the chip to render
"live **and** after a transcript reload". Those two cannot both hold: the
journal field the chip is rehydrated from (`AgentReply.task_id`) can persist only
an id, so a title available live and absent after reload would make the two chips
render differently — which is precisely the failure the criterion exists to
prevent. The wire therefore carries the id alone, and the chip's wording is
derived from whose message it is ("Card opened" on a company reply, "Added to the
board" on your own). Both paths then render identically, live and rehydrated.

**`AgentReply.task_id` semantic widening.** No serde change was needed: the
field already existed from #185 and this only populates it on a new occasion.
Pre-existing journals therefore round-trip byte-identically by construction, and
the existing #185 test `task_id_correlation_is_additive_and_omitted_when_absent`
— which asserts an untagged reply serializes to the exact legacy string — still
passes unchanged. The widening is real but only in meaning: "the dispatch that
produced this reply" becomes "the card this reply is about", so a card-creating
reply now also appears on that card's timeline, which is lineage an operator
wants at no schema cost.

**`OutboundMessage`'s wire name is pinned to `taskId`.** That struct carries no
`rename_all`, unlike the two sibling surfaces the console reads the same card on
(the SSE `agent_reply` frame and the `chat/history` DTO). Letting this one alone
be `task_id` would be a trap for the next reader.

**One asymmetry, stated rather than hidden.** The chip on a *company reply* is
rehydrated from the journal and survives reload. The chip on *your own* message
(the "Add to board" path) is session-only, because the operator message carries
no card id in the journal — the durable record of that link is `originChatId` on
the card, which task detail now reads back and offers a jump from. Making the
operator-message chip durable would mean a new journal field, which is more
schema than this edge is worth.

### Collision with #293 (issue #272)

#293 is open on `src/runtime/delegation.rs`, `src/harness/brain.rs`,
`src/harness/orchestrator.rs`, `src/runtime/cycle.rs` and
`src/runtime/delegation_tools.rs`. I read its diff before starting.

Both PRs edit the same `match` in `run_delegation`, in adjacent arms: #293 adds
a refusal path (a `refused` list on `DelegationQueue`, drained in
`handle_task_delegations`, plus a `tracing::warn!` in the `DelegateToDesk` arm's
no-lead branch); this adds a `spawned_task` field to `DelegationOutcome`, set in
the `SpawnTask` arm. Only this PR changes the struct, so #293's bare
`DelegationOutcome::default()` returns keep compiling.

The two hunks are close enough to share context lines — my last changed line is
the `SpawnTask` arm's return, #293's first is three lines later inside
`DelegateToDesk` — so a clean auto-merge is likely but not guaranteed, and a
clean merge is not the thing to check anyway. Whoever lands second should re-run
`cargo test --features openhuman,tinycortex --lib` rather than trust the merge,
since both PRs change what that `match` reports and only the harness tests
exercise it.

We also both touch `src/runtime/cycle.rs` and `src/harness/brain.rs`, without
overlap: in `cycle.rs` my edits are struct-literal `task_id: None` fills at the
construction and test sites while #293's are at the imports and `CycleHostImpl`;
in `brain.rs` we both append tests at the tail, which is the likelier textual
conflict of the two and a trivial one. I touch neither `orchestrator.rs` nor
`delegation_tools.rs`.

Also checked: **#278** (assignee picker) touches `TaskDetailView.tsx` and
`TasksView.tsx`, which I also edit — but its hunks are inside `ControlBar` and
`CreateTaskDialog`, while mine are the props block, the render slot and a new
component. Adjacent, not overlapping. **#289** (`DeliveryReason`) does not
intersect.

### Found but not fixed

Deliberately out of scope; each is a separate decision:

- **`run_chat`'s deterministic intent card** (`src/server/operator.rs`) — #126's
  own auto-created card — still writes `origin_chat_id: None` and
  `column: "backlog"` even though `message.chat` is in scope one line away.
  Stamping it is a one-liner, but it would change behaviour: #151's post-back
  would start firing into that thread on completion. That is arguably the right
  behaviour, but it is a behaviour change no acceptance criterion asks for, and
  it does not belong in a visibility PR.
- **The hosted-brain spawn path** (`CycleHostImpl::spawn_task`,
  `src/runtime/cycle.rs`) also writes `origin_chat_id: None` and reports nothing
  to a bubble. It has no chat id in scope and no `DelegationOutcome` equivalent,
  so it needs its own plumbing.
- **`frontend/src/api/types.ts` mis-declares `OutboundMessage.replyTo` /
  `ReplyTo.chatId`.** The Rust structs carry no `rename_all`, so the wire is
  `reply_to` / `chat_id`. Harmless today because nothing reads them — which is
  exactly why it will bite whoever does first. Pre-existing.
- The GraphQL `Task` type carries neither `parentTaskId` (#185) nor
  `originChatId`. Left alone: that type is a deliberately lean read and the
  console reads the board over REST.

## API Or Behavior Changes

All additive; no existing wire shape changes.

- `POST …/tasks` accepts an optional `originChatId`. A blank value is normalised
  away rather than persisted as a thread id matching nothing.
- `TaskCard` (both `GET …/tasks` and `GET …/tasks/{id}`) gains `originChatId`,
  omitted when absent.
- `OutboundMessage` gains `taskId` (wire name pinned), omitted when absent. A
  bubble that opened no card serializes byte-for-byte as before.
- `GET …/chat/history` DTO and the GraphQL `Message` type both gain `taskId`,
  projected from the one shared `MessageView` field (#65's rule), so the two
  surfaces cannot disagree. GraphQL SDL snapshot regenerated.
- `AgentReply.task_id` is now also written for a chat turn that opened a card.
  No serde change; stored journals are untouched.
- Console: a per-message "Add to board" action on every thread; a chip on any
  message tied to a card, linking to `#/tasks/<id>`; a jump from task detail
  back to the originating conversation.

No changes to board lanes, the `+` button, assignee validation, or the
dispatch/review lifecycle.

## Tests

Backend gates, run on this branch:

- [x] `cargo test` — 1015 passed, 0 failed
- [x] `cargo test --features openhuman,tinycortex --lib` — 1431 passed, 0 failed
- [x] `cd frontend && npm ci && npm run typecheck && npm run build`
- [x] `npx playwright test test/e2e/chat-to-card.spec.ts` — 2 passed, against a live host

The harness-feature runs are listed because the delegation and journal paths
this touches do not compile under default features, so `cargo test` alone never
executes the three `harness::brain` tests below. `--no-deps` scopes clippy to
this crate: the vendored `openhuman` checkout emits its own warnings, and
`-D warnings` applies to the primary crate, not to a path dependency.

New tests:

| Test | What it pins |
|---|---|
| `a_task_created_from_a_thread_remembers_and_reads_back_that_thread` | the write keeps `originChatId`, both reads return it, a blank one is normalised away, and a card with no thread does not grow the key |
| `a_chat_created_card_lands_off_the_dispatch_trigger` | the HTTP-level spend gate |
| `the_column_a_chat_created_card_defaults_to_does_not_dispatch` | the default column and the dispatch trigger, together, with a positive control |
| `outbound_message_task_id_is_additive_and_omitted_when_absent` | a bubble with no card keeps the pre-#246 wire form; a legacy payload still loads |
| `a_turn_that_opens_a_card_reports_it_on_the_operator_bubble` | the reported id is the card actually on the board |
| `a_turn_that_opens_several_cards_reports_the_first` | the documented limitation, as a contract rather than prose |
| `a_turn_that_opens_no_card_reports_none` | no bubble claims a chip it has not earned |
| `chat_history_route_rehydrates_the_card_a_reply_opened` | the chip survives a reload over REST |
| `chat_history_projects_the_card_a_reply_opened` | …and over GraphQL, from the same shared field |

End-to-end, `frontend/test/e2e/chat-to-card.spec.ts` — **2 passed**. A real host
built `--features openhuman,tinycortex` on an isolated `--home`, serving the
built console, with only the model's *choices* stubbed behind a scripted
OpenAI-compatible endpoint; the harness, tool plumbing, cycle, journal and HTTP
surface all real.

The stub's call log is the part worth reading, because it evidences the two
claims that motivated the feature:

- on the **desk** thread the responder was offered only
  `['memory_recall', 'memory_store']` — no `spawn_task` — which is precisely the
  case where a card was previously unreachable by any path. "Add to board" works
  there anyway because it creates over REST rather than through the responder's
  toolbelt.
- on the **orchestrator** thread the toolbelt included `spawn_task`, and the
  following call arrives with `tool_result=True` — a real tool round-trip, not a
  canned reply. The chip that turn produced is then still present, pointing at
  the same card id, after a full reload.

## Documentation

- `docs/spec/runtime/ports.md` — `OutboundMessage.task_id`, the
  `AgentReply.task_id` widening and its multi-spawn limitation, and why the
  "Add to board" action leans on the `todo` create default as the spend gate.
- `docs/spec/runtime/api.md` — `originChatId` on the create route.

### DIFF
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -43,7 +43,7 @@ below). Anything a build doesn't serve `404`s — the console treats that as "no
 wired yet".
 
 ```text
-POST   …/tasks                              create a task card
+POST   …/tasks                              create a task card (`originChatId` records the thread it came from, #246)
 PATCH  …/tasks/{taskId}                      edit / move a task
 DELETE …/tasks/{taskId}                      delete a task
 POST   …/memory                             add a memory fact
--- a/docs/spec/runtime/ports.md
+++ b/docs/spec/runtime/ports.md
@@ -176,6 +176,18 @@ task-to-task edge that `origin_chat_id` (a *conversation*, shared by every
 sibling spawned in that thread, and absent entirely on a board-native card)
 cannot express. It is the parent half of the Task Detail screen's lineage.
 
+`OutboundMessage` gains `task_id` on the same contract (issue #246): the card a
+chat turn **opened**, so the console can say a card exists instead of leaving an
+operator to notice it on the board. It is journaled onto that turn's
+`AgentReply.task_id`, which widens that field's meaning from "the dispatch that
+produced this reply" to "the card this reply is about" — a card-creating reply
+now also appears on that card's timeline, which is the lineage an operator
+wants and costs no schema change. A turn that opens several cards reports the
+**first**: the journal field is a single optional id, and widening it would
+break the byte-identical round-trip, so the claim is incomplete but never wrong.
+Both `chat/history` surfaces (REST and GraphQL) project it from the shared
+`MessageView`, so the chip survives a transcript reload on either.
+
 ## MemoryStore
 
 The equivalent of Medulla's `CyclePersistence`; TinyCortex is the target
@@ -363,7 +375,10 @@ another pass (a failed dispatch, an orchestrator `revise` verdict) — while
 `todo` is what has been queued up next. `todo` is the board's one manual-entry
 column: the console's `+` button lives there alone and `POST …/tasks` defaults
 to it (issue #206), so an operator cannot create a card straight into
-`in_progress` or a terminal column.
+`in_progress` or a terminal column. The transcript's "Add to board" action
+(issue #246) relies on exactly that default: it omits `column` so the *server*
+decides where a chat-created card lands, which is what keeps the human drag into
+`in_progress` the only thing that spends an agent turn.
 
 `assignee` names a **roster teammate id, a desk, or nobody** (`""`), resolved by
 `crate::runtime::assignee` against the full roster — manifest agents, operator
--- a/frontend/src/api/tasks.ts
+++ b/frontend/src/api/tasks.ts
@@ -20,6 +20,12 @@ export interface Task {
    * shape is unchanged.
    */
   parentTaskId?: string;
+  /**
+   * The chat thread this card was opened from (issue #246), when it came from
+   * a conversation rather than the board's `+` button. Omitted otherwise, which
+   * is every card created before this shipped.
+   */
+  originChatId?: string;
 }
 
 /** The create body; the host defaults column→`backlog`, priority→`medium`. */
@@ -29,6 +35,16 @@ export interface CreateTask {
   column?: string;
   priority?: string;
   assignee?: string;
+  /**
+   * The chat thread this card is being opened from (issue #246). Set by the
+   * transcript's "Add to board" action; the board's `+` button omits it.
+   *
+   * Note what is deliberately NOT sent alongside it: `column`. Dropping a card
+   * into `in_progress` is what dispatches an agent turn — it spends money — so
+   * the server's intake default decides where a chat-created card lands, and
+   * the human drag stays the only spend gate.
+   */
+  originChatId?: string;
 }
 
 /** A partial update; any omitted field is left as-is. A drag sends `{column}`. */
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -65,6 +65,16 @@ export interface OutboundMessage {
   steps?: TurnStep[];
   /** Channel-specific reply addressing (Telegram). Absent on operator messages. */
   replyTo?: ReplyTo;
+  /**
+   * The board card this turn opened, when it opened one (issue #246). Drives
+   * the reply bubble's "card opened" chip. Absent when the turn opened nothing
+   * — which is every reply the host sent before this field existed.
+   *
+   * Only the *first* card of a turn that opened several: the journal field this
+   * is persisted into is a single id, so the claim is incomplete but never
+   * wrong. The bubble's `steps` timeline still shows every spawn.
+   */
+  taskId?: string;
 }
 
 /** Channel-specific reply addressing. Mirrors `ReplyTo` in `src/ports/types.rs`. */
@@ -141,6 +151,13 @@ export interface ChatHistoryMessageDto {
    * empty (operator messages, tool-less replies).
    */
   steps?: TurnStep[];
+  /**
+   * The board card this reply is about (issue #246) — the card the turn opened,
+   * or the dispatched card it ran for (#185). Projected from the same shared
+   * `MessageView` field the GraphQL `Chat.history` resolver reads, so the chip
+   * renders identically whichever surface hydrated the transcript.
+   */
+  taskId?: string;
 }
 
 /** Response of `/chat` and approval-resolution routes. */
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -358,7 +358,13 @@ export function AppShell({
         if (dup) return t;
         return {
           ...t,
-          messages: [...t.messages, makeMessage("company", event.text, { channel: event.agentId })],
+          messages: [
+            ...t.messages,
+            makeMessage("company", event.text, {
+              channel: event.agentId,
+              taskId: event.taskId,
+            }),
+          ],
         };
       }),
     );
@@ -546,7 +552,19 @@ export function AppShell({
             />
           )}
           {view === "inbox" && <InboxView client={client} company={company} />}
-          {view === "tasks" && <TasksView client={client} company={company} />}
+          {view === "tasks" && (
+            <TasksView
+              client={client}
+              company={company}
+              // Issue #246: the card → chat half of the round trip. A card
+              // opened from a conversation remembers which one, so its detail
+              // screen can put the operator back in that thread.
+              onOpenThread={(threadId) => {
+                setActiveThreadId(threadId);
+                setView("conversation");
+              }}
+            />
+          )}
           {view === "team" && <TeamView client={client} company={company} />}
           {view === "desks" && <DesksView client={client} company={company} />}
           {view === "people" && <PeopleView client={client} company={company} />}
--- a/frontend/src/hooks/use-events.ts
+++ b/frontend/src/hooks/use-events.ts
@@ -11,7 +11,20 @@ import type { DeliveryReport } from "@/api/workflows";
  * raw third-party payload is ever on the wire.
  */
 export type CompanyStreamEvent =
-  | { type: "agent_reply"; seq: number; atMillis: number; chatId: string; agentId: string; text: string }
+  | {
+      type: "agent_reply";
+      seq: number;
+      atMillis: number;
+      chatId: string;
+      agentId: string;
+      text: string;
+      /**
+       * The board card this reply is about (issue #246/#185) — the card the
+       * turn opened, or the dispatched card it ran for. Absent on an ordinary
+       * chat reply.
+       */
+      taskId?: string;
+    }
   | { type: "task_dispatched"; seq: number; atMillis: number; taskId: string }
   | { type: "task_steered"; seq: number; atMillis: number; taskId: string; action: string }
   | {
@@ -79,6 +92,8 @@ export interface AgentReplyEvent {
   chatId: string;
   agentId: string;
   text: string;
+  /** The board card this reply opened (issue #246), when it opened one. */
+  taskId?: string;
 }
 
 interface Options {
@@ -260,7 +275,15 @@ function handleEvent(
       onTaskEvent?.(event);
       break;
     case "agent_reply":
-      onAgentReply?.({ chatId: event.chatId, agentId: event.agentId, text: event.text });
+      onAgentReply?.({
+        chatId: event.chatId,
+        agentId: event.agentId,
+        text: event.text,
+        // Issue #246: a reply injected from the stream — one this console did
+        // not POST for, e.g. an inbound Telegram turn — carries its "card
+        // opened" chip too, rather than only the locally-awaited copy.
+        taskId: event.taskId,
+      });
       break;
     case "approval_resolved":
       toast(event.verdict === "approve" ? "Approval granted" : "Approval denied", {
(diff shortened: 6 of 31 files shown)


oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai review

oxoxDev:
@coderabbitai full review

Issue #239, opened 2026-07-31
fix(skills): registry install persists the description as the skill body, discarding the real procedure


oxoxDev:
## Summary

Installing a skill from the registry gives the agent a one-line description instead of the procedure. The `## Steps` and `## Output` sections — the entire value of the skill — are discarded at install time.

## Problem

`install()` builds the persisted `SKILL.md` from client-supplied metadata, with the description doubling as the body:

```rust
// src/server/ops/skills.rs (~:230)
// The registry ships metadata only. Persist a real `SKILL.md` built from it
// (the description doubles as the body) so `EffectiveSkills::materialize`
// surfaces the skill to the agent instead of skipping a content-less delta.
let doc = skill_md(&name, &description, meta.category.as_deref(), &description);
```

The server already holds the full documents — `AppState::skill_registry` (`src/app/types.rs:388`), bodies preserved verbatim by `parse_skill_md` (`src/company/skill_file.rs:33`) — and the install handler never consults them. The root cause is that **install trusts client-supplied metadata** rather than resolving content from the server's own registry.

Steps to reproduce:

1. Install `competitor-scan` from the console registry.
2. Read the materialised `skills/<slug>/SKILL.md` in the agent's scratch tree.
3. The body is one sentence. The five-step procedure is gone.

Two adjacent defects compound it:

- The console's registry tab renders a **hardcoded 6-entry client array** (`frontend/src/lib/skills.ts:25`) instead of the live 14-skill registry, so what an operator can browse and what the backend can serve have already diverged.
- There is **no `version` field** anywhere — not in frontmatter, `SkillDoc`, `RegistrySkillGql`, or `SkillState`. Publisher is the constant `"OpenCompany"`. An installed skill cannot be pinned or updated.

## Solution

1. **Make install server-authoritative.** Add a `State<AppState>` extractor to `install()`; resolve the slug against the server's registry and persist the **full original document** (frontmatter + body verbatim) into `SkillState.custom_doc`. Registry content always wins over client metadata. If the registry is non-empty and lacks the slug, return `404` so typo'd installs fail loudly. Keep the client-meta fallback only when `skills_root` is `None` (platform-provisioned mode, where the registry has nothing to serve).
2. **Add a frontmatter renderer.** `render_skill_md(&SkillDoc) -> String` in `src/company/skill_file.rs`, reconstructing frontmatter and appending the body verbatim, with a parse→render→parse round-trip test.
[...]

Issue #237, opened 2026-07-31
feat(harness): surface the company workspace to agents as live read/write tools


oxoxDev:
## Summary

The company workspace — the shared note tree that is meant to be every agent's single source of truth — is invisible to every agent. An operator can create a "Brand voice" folder, fill it with the guidance all agents are supposed to follow, and no agent will ever read a word of it.

## Problem

`WorkspaceStore` (`src/ports/workspace.rs:46`) is real and well-built: ULID node ids, `[[wiki links]]`, seeded once from `companies/<name>/workspace/**` by `seed_workspace` (`src/runtime/builder.rs:1247`), persisted in `workspace_nodes` (`src/store/sqlite.rs:155`) across all three backends. The `agentic_software_company` template already ships `Playbooks/`, `Product/`, `Standards/`.

But it is referenced **only** by `src/runtime/builder.rs` and `src/server/ops/workspace.rs`. Nothing under `src/harness/` touches it. `build_agent` (`src/harness/build.rs:128`) wires memory, file, exec, MCP and orchestrator tools — and the only "files" an agent sees are its own private sandbox at `<workspace_root>/<company>/<agent>/workspace` (`build.rs:144-171`).

Two adjacent surfaces are disconnected the same way:

- Agent sandbox files are never bridged back into `WorkspaceStore`.
- Humans write facts via `POST {scope}/memory` (`src/server/ops/memory.rs:49`) into `FactStore`, which is **not** the store agents recall from (`src/harness/memory.rs:148`). The human-writes/agent-reads loop is broken in both directions.

Steps to reproduce:

1. Add a note under `Standards/` via the workspace REST API.
2. Ask any dispatched agent what that standard says.
3. The agent has no tool that can reach it and answers as if it does not exist.

This blocks any workflow premised on shared source-of-truth notes — brand voice, research findings, rate cards, playbooks.

## Solution

Three tools in a new `src/harness/workspace_tools.rs`, each holding `Arc<dyn WorkspaceStore>` + `CompanyId` and hitting the store **live at `execute()` time**:

- `workspace_list` — bounded path index (path, kind, id, `updated_at_millis` as revision token), optional `prefix` for subtree listing.
- `workspace_read` — by path or id, content capped (~64 KiB) with an explicit truncation marker.
[...]

Issue #188, opened 2026-07-30
feat(agents): OpenCompany agent toolbelt + capability tiers (metered OH-subset)


oxoxDev:
## Problem

When a task is dispatched to a roster agent (#186 / `run_task`), **what tools that agent actually has is undefined and inconsistent**. An OpenCompany agent is a **constrained, company-scoped, metered derivative** of an OpenHuman agent — never a superset — but nothing enforces that shape:

- The **hosted** build wires a near-empty `StubToolProvider` — the agent can narrate but can't actually do work.
- The **harness** (`openhuman`) build inherits OH's toolbelt with **no per-tier metering and no OC-specific filtering** — over-broad (raw browser, web-search, Node exec, recursive sub-agent delegation, skill execution) and uncapped.
- No capability tiers by company plan, no enforced delegation-depth cap, and memory recall is unwired (`src/harness/memory.rs` `OcMemory` is append-only, no vectors, `forget` is a no-op).

## Proposal

Define and enforce the OpenCompany agent toolbelt for dispatched agents:

1. **Curated OH-subset toolbelt (v1):** file / shell / code / web + media / composio (per-company bearer). **Stripped for v1:** raw browser, web-search, Node exec, **recursive sub-agent delegation (depth capped at 1 — "no re-delegation in v1")**, skill execution, raw memory-tree tool surface.
2. **Per-tier token metering** — the capability OC adds over OH: cap tool + model spend by company plan tier at the dispatch/tool boundary.
3. **Memory recall capability — in-pod, NO backend service.** OH's memory engine (the `tinycortex` crate) is an **unconditional dep** of `openhuman_core`, so it is **already compiled into any core-in-pod build** — just **never constructed** by OC today (OC instead wired a weaker parallel path: `OcMemory` append-only over `ContextStore` + an imagined remote `CortexClient`/`HttpCortexClient` that is inert). The engine is designed to be **embedded with compute injected** — "TinyCortex never makes a network call"; the host supplies embeddings/LLM. So the recall wiring is entirely in-pod:
   - Construct the engine via `memory_config_from(config, workspace)` pointed at the pod's `/data` workspace (local SQLite vectors/kv/entity + YAML vault + summary trees).
   - Inject OC's **existing** embedding + LLM providers through the seam adapters (`SeamEmbedder`, `SeamChatProvider`) — OC already calls the TH endpoint for these. Embedding **compute** is remote; the **vector index is stored locally** in the engine's SQLite.
[...]

PR #299, opened 2026-08-03
feat(harness): admit a scoped, metered web_search for the research skills (#238)

### LINKED ISSUE #238: feat(harness): admit scoped, metered web_search for research skills
## Summary

Three shipped skills instruct agents to research and cite sources, but no discovery tool exists. `web-research`, `seo-audit` and `competitor-scan` run on a toolbelt that cannot search — which invites fabricated URLs and citations.

## Problem

Agents hold `web_fetch`, `http_request`, `curl` and `image_info` (`src/harness/toolbelt.rs`), SSRF-guarded with a per-company allowlist. Those read a **known** URL. They cannot discover one.

`web_search`, `search` and `google_search` are pinned absent by a negative contract test (`src/harness/build.rs:811`), and the toolbelt comment defers them for exactly one recorded reason — *"search tools (need engine keys)"* (`src/harness/toolbelt.rs:38`).

The result is a live contradiction: `skills/web-research/SKILL.md` tells the agent to "search broadly; gather several independent sources" and cite them, on a belt with no way to do either.

Steps to reproduce:

1. Ask an agent to research a competitor's pricing.
2. It has no search tool, and the skill instructs it to cite sources.
3. It either declines or invents plausible URLs.

## Solution

Add a metered, company-scoped `web_search`, admitted **scoped, granted and metered** rather than by reopening the toolbelt.

1. **Citation-first signature**, inherited from OpenHuman's `web_search_tool` rather than written fresh. Input `{ query, max_results (default 5, cap 10), freshness? }`; output `[{ title, url, snippet, published?, source_domain }]` plus `{ provider, result_count }`. Results are wrapped in an explicit untrusted-data frame with snippets capped (~300 chars) to bound injection payload and token cost. Discovery and retrieval stay layered — search returns metadata only; page content still flows through the SSRF-guarded `web_fetch`.
2. **No new provider abstraction.** OpenHuman's `src/openhuman/search/engines/` already has one file per engine and owns the canonical `web_search_tool` slot. OpenCompany selects an engine; it does not implement one.
3. **Key model: adopt OpenHuman's, do not invent one.** OpenHuman already owns this problem — `src/openhuman/search/registry.rs` builds the tool surface from `Config.search`, with engines `managed` / `parallel` / `brave` / `querit` / `exa` / `disabled`. `managed` is **backend-proxied**, so there is no platform API key to hold, and a BYO engine with no key configured falls back to managed, making managed the effective default until a tenant saves one. `disabled` registers nothing, so search is absent from agent prompt context entirely.

   OpenCompany threads a per-company `Config.search` (engine plus optional BYO key) rather than adding an `OPENCOMPANY_SEARCH_*` credential of its own. This also means cost recovery should mirror **managed inference** — the backend charges, and OpenCompany recovers the amount the way `openhuman_usage_meta` already does — rather than a locally configured `price_per_call_usd`.
4. **Explicit grant; `*` does not confer it.** New `search` namespace beside media/composio, because each call spends real platform money. Added to `GATEABLE_NAMESPACES` so capability-tier budgets can shed it under spend pressure. Granted-but-uncredentialed wires nothing and warns (media's shape exactly).
5. **A paid call is never free.** New `SampleKind::SearchCall` (additive on `src/ports/usage.rs:36`) plus `src/metering/search.rs`. Reusing `OauthCall` is wrong twice: it hardcodes `cost_usd: 0.0` by design (`src/metering/oauth.rs:55`), and it would inflate the connected-accounts chart. `bucket_usage` learns a `searchCalls` counter, surfaced in the console Usage view.
6. **Loud cap, not silent.** Metering is a rear-view mirror, so the tool enforces a per-company daily call cap; over-cap returns a well-formed "search budget exhausted for today" error so the agent reports the constraint instead of fabricating.
7. **Surgical test edit.** Remove only `web_search` from the forbidden array at `build.rs:811`; keep `search`, `google_search` and every other deferred family pinned. Add mirror-image positive tests: `["*"]` + credentialed → absent; `["search"]` + config → present; `["search"]` + no config → absent.
8. **Close the contradiction from both ends.** The three skills name the tool ("discover with `web_search`, read with `web_fetch`") and gain a degradation clause: if search is unavailable, say so and work from operator-supplied URLs — **never fabricate citations**.

## Acceptance criteria

- [ ] `web_search` is wired only when the manifest grants `search` explicitly **and** a platform `SearchConfig` resolves; `["*"]` alone never wires it; granted-but-uncredentialed warns and wires nothing.
- [ ] Output is structured citations — title, url, snippet, published (when available), source domain — framed as untrusted third-party data with capped snippets.
- [ ] Every completed search records exactly one `SampleKind::SearchCall` with nonzero `cost_usd`; a failed search records nothing; the console Usage view shows a `searchCalls` counter and rolls cost into period totals.
- [ ] `build.rs:811` drops only `web_search`; all other deferred families stay pinned, with new positive tests covering the three gate states.
- [ ] The daily per-company cap is enforced and returns an explicit "budget exhausted" tool error, never a silent drop.
- [ ] `search` is in `GATEABLE_NAMESPACES` and sheds under capability-tier budget exhaustion.
- [ ] The three research skills name `web_search` and instruct graceful degradation with no fabricated citations.

## Notes and scope

**Signed off by the #188 owner on 2026-08-01 (see the decision comment below). No longer blocked.**

The argument that was accepted, for treating search differently from its fellow exclusions: browser was stripped for backend and attack surface, Node exec for arbitrary code under multi-tenancy, sub-agent spawn for budget bypass — all safety rationales. Search was deferred for one infrastructure reason, "need engine keys", and the managed-platform-credential pattern from #109 has since dissolved that blocker. Search is read-only, session-less, per-call-bounded, and strictly less dangerous than the paid media tools already admitted.

The honest counter-argument: v1 minimalism has intrinsic value, and every tool widens the injection surface operators must reason about.

Most contestable sub-decision, flagged for that review: individual searches do **not** park under `ApprovalPolicy` the way media and composio calls do. Five to ten approval clicks per research task would kill the feature; the explicit grant is the consent act and the daily cap is the boundary.

Residual risks stated plainly: the in-process cap resets on restart and does not span multi-process deployments (accepted for v1); SEO-poisoned snippets are a real injection carrier, mitigated by untrusted framing and truncation but not eliminated.

**Scope note:** adopting OpenHuman's search domain removes the provider trait, the engine implementation and the platform credential from this issue. What remains is OpenCompany-side: un-strip the tool from the deferred family, add the `search` namespace and its explicit grant, thread per-company `Config.search`, meter and cap, and update the three skills. Closer to **small-medium** than the medium originally estimated.

Ships opt-in. Making it default-on is a one-line follow-up gated on the same sign-off.

Scope: **small-medium** (revised down — see scope note above).

## Related

Revisits the web-search exclusion in #188 — **decided 2026-08-01, admitted**. Key model follows OpenHuman's `search.engine` (managed-by-default, BYO override); grant model follows the explicit-grant pattern of #110.

Coincides with any issue touching `src/harness/build.rs` or `HarnessDeps`.

### LINKED ISSUE #297: test: dispatched_desk_agent_tool_belt_is_pinned fails under --features openhuman,mcp,telegram
## Summary

`cargo test --features openhuman,mcp,telegram` fails one test on `main`: `dispatched_desk_agent_tool_belt_is_pinned`. No CI lane runs that feature combination, so it has been failing unseen.

## Problem

The assertion diff is exactly two entries — `mcp_registry_list_tools` and `mcp_registry_tool_call`. Both are pushed onto the tool belt unconditionally under `#[cfg(feature = "mcp")]`, while the test's own doc comment states it assumes `--features openhuman`. So enabling `mcp` alongside `openhuman` adds two tools the pin does not expect, and the pin fails.

Found while implementing #237 (workspace tools), on a branch that adds no tool to that belt — the diff contained no workspace tool at all, which is what identified it as pre-existing rather than introduced. Worth an independent confirmation on a clean `main` checkout before fixing, since it was observed from a feature branch.

## Why CI does not see it

`ci.yml` runs `cargo test` (default features) and `cargo test --features openhuman,tinycortex --lib`. Neither enables `mcp`. The gated job's comment explains the choice deliberately — one combination rather than a matrix, on the grounds that `tinycortex` adds no new compilation over `openhuman`. That reasoning is sound for compilation and does not hold for *behaviour*: `mcp` genuinely changes what the agent's tool belt contains.

This is the third instance of the same shape, after #288 (`--all-features` never compiles the test target) and #281 (no job builds the console). A green check is only evidence for the surface it actually exercises.

## Solution

Two parts:

1. Make the pin feature-aware so it asserts the right belt under each combination, rather than assuming one. Adding the two entries unconditionally would make the test wrong under plain `--features openhuman`, which is the combination CI actually runs — so the fix has to branch on `cfg(feature = "mcp")`, not just extend the array.
2. Decide whether a lane should exercise `openhuman,mcp,telegram`. That combination is what a full local build uses and what the tenant image ships, so nothing in CI currently tests the belt real deployments get.

## Coordination

The four-line fix touches the same pinned array that **#238** (scoped, metered `web_search`) will edit when it adds its tool. Whoever gets there first should take it; doing both separately guarantees a conflict in a five-line region. #237's author deliberately left it alone for that reason.

## Acceptance criteria

- [ ] Confirmed failing on a clean `main` checkout with `cargo test --features openhuman,mcp,telegram`
- [ ] The pin passes under `openhuman`, `openhuman,tinycortex`, and `openhuman,mcp,telegram`
- [ ] A CI lane covers the belt that shipped deployments actually build, or the decision not to is written down in `ci.yml` beside the existing rationale

Scope: **small** for the fix, **small-to-medium** including the CI lane.

oxoxDev:
## Summary

Adds `web_search` — the discovery tool the agent belt never had. Agents hold
`web_fetch` / `http_request` / `curl`, which read a URL they already have;
nothing could *find* one, while three shipped skills (`web-research`,
`seo-audit`, `competitor-scan`) instruct the agent to "search broadly" and cite
its sources. The only way to satisfy that instruction was to invent plausible
URLs.

The tool runs on the managed platform's backend-proxied search surface (the same
identity as managed inference), behind three gates: an **explicit** `search`
grant that a `*` wildcard never confers, a managed credential, and a
**per-company daily call cap** enforced before the request. Every completed
search records one priced `SampleKind::SearchCall` usage sample carrying the cost
the backend reported.

Also takes **#297** (see *API Or Behavior Changes*), because its four-line fix
edits the same pinned array this issue extends and doing them separately
guarantees a conflict.

`Closes #238`
`Closes #297`

### What I verified against the code before implementing

The issue's premise held up in the parts that matter and was wrong in two
places. Each is implemented the way the code supports, not the way the issue
proposed:

1. **The deferral reason was real and is gone.** `toolbelt.rs` deferred search
   for exactly one recorded reason — *"need engine keys"* — and OpenHuman's
   managed engine is backend-proxied, so there is no engine key to hold. The
   other deferred families (browser, Node exec, sub-agent spawn) were excluded
   for *safety* reasons that still hold; they stay pinned absent.
2. **`budget_usd_daily` really does enforce nothing.** Verified: stored on
   `ApprovalPolicy`, exposed by a getter, **zero call sites**. This PR does not
   fix that — see *Not fixed here*.
3. **The issue's approval decision was wrong in both directions.** It proposed a
   flat carve-out making individual searches never park. Half right:
   - Not parking under `supervised` is **correct, and load-bearing**. openhuman
     resolves a `RequireApproval` inline and never re-dispatches the call
     (documented in `harness::policy`), so parking a search is not "the operator
     approves it later" — it is "the search never happens", which leaves the
     agent in exactly the no-discovery state that makes it fabricate. Since
     `supervised` is the manifest default, a parked search would have shipped the
     feature dead.
   - A flat carve-out under **`readonly`** is wrong. That tier's whole contract
     is that nothing is spent, and a search spends. `web_fetch` — its sibling in
     the same research loop — is denied there; a *priced* discovery call has no
     business being more permissive than the free retrieval call it feeds.

   So the classifier grew a third bucket rather than a carve-out: `is_metered_read`
   ("reaches outside and costs money, but changes nothing") relaxes **only**
   `supervised`. `readonly` still denies via the ordinary external-effect path.
   All three modes are pinned by test, plus the operator escape hatch
   (`[policy].always_approve = ["web_search"]` overrides every tier, and the
   parked effect is projected as `Spend`, not the catch-all `Other`).
4. **The proposed `freshness` argument would not have worked.** The backend's
   declared body has no freshness parameter — OpenHuman had to *remove* an
   undeclared `timeoutSecs` field for that reason. A schema advertising a filter
   the backend never applies is worse than none: the agent believes it
   constrained the search and did not. Dropped; each result's `published` date is
   surfaced instead so the model can weigh recency itself.
5. **No per-tenant BYO engine key in v1.** A tenant Brave/Exa key is a secret and
   the manifest is not a secret store. Managed is the effective default upstream
   anyway (a BYO engine with no key falls back to it), so this ships the default
   and leaves the override to the console credential surface `composio` already
   uses.

### The approval-classification decision (the #237/#296 trap, re-decided)

`policy::is_external_effect` sees only a tool name. #296 added a read-only
carve-out for `workspace_list` / `workspace_read`; copying that shape here would
have been wrong, because those are free and a search is not. The rule the
existing arms actually encode is *free ⇒ waved through* (`media_list_models`,
`composio_list_*`) and *costs money ⇒ gated* (`media_generate_*`,
`composio_execute`). `web_search` sits on the money side, so it stays external —
and the new metered-read bucket relaxes only the tier where parking would be
useless rather than protective.

### Metering — what it enforces and what it does not

- **New `SampleKind::SearchCall`, not `OauthCall`.** `oauth_call_sample` hardcodes
  `cost_usd: 0.0` by design, and it is what mints the calls-by-provider chart
  whose *row count* is the connections KPI. Reusing it would have thrown away the
  only cost figure we have and reported a connected account the company never
  set up. Verified both by test.
- **Cost is recovered, not configured.** The backend reports `costUsd` per search
  and that is what lands in the sample, mirroring managed inference. A
  non-positive reported cost floors to the documented list price rather than
  zero — a completed *paid* call recorded free is the failure this feature exists
  to end.
- **What actually stops spend is the cap, not the meter.** Metering is a
  rear-view mirror. `[tools].search_daily_calls` (default 200, `0` pauses search
  without editing `allow`) is a per-company, UTC-day-bucketed reservation
  counter checked *before* the request. A reservation is refunded if the search
  never reached the backend, so an unreachable endpoint cannot burn a company's
  day. Over-cap returns a loud, specific error naming the ceiling and forbidding
  fabrication — never an empty result set, which is the shape that makes a model
  invent sources.
- **Honest limits.** The ledger is in-process: it resets on restart and does not
  span replicas. That is the v1 position the issue records. It is a ceiling on
  runaway spend, not an accounting boundary — the usage samples are the durable
  record.

### #297 — the pinned belt was failing unseen

`dispatched_desk_agent_tool_belt_is_pinned` fails under
`--features openhuman,mcp,telegram`: `build_agent` pushes two `mcp_registry_*`
tools unconditionally under `#[cfg(feature = "mcp")]`, and no CI lane ran that
combination — the one a full local build and the shipped tenant image both use.
Fixed by making the pin **feature-aware** (branching on the same `cfg` as the
wiring), not by extending the array, which would only have moved the failure onto
plain `--features openhuman` that CI does run. Reproduced on this branch by
disabling the new `cfg` block: the diff is exactly `mcp_registry_list_tools` +
`mcp_registry_tool_call`.

Also added a narrow CI step running the belt-contract module under
`openhuman,mcp,telegram` — #297's third acceptance criterion, and the third time
this shape has bitten (after #288 and #281). Deliberately scoped to that one
module rather than the whole suite, since the feature set is a fourth distinct
compile.

### Also worth knowing

- **No new Cargo feature.** `web_search` rides the plain `openhuman` harness
  feature, unlike `media` / `composio`. It needs only the always-compiled
  integrations client, and CI's gated lane builds `openhuman,tinycortex` — so the
  tool and its tests are actually compiled and run. Hiding a real-money surface
  behind a feature no CI job builds is how #288 / #281 / #297 each happened.
- **Ships opt-in.** `search` is deliberately **not** in the default
  `[tools].allow` list, unlike `media` / `composio`, per the #188 sign-off.
  Making it default-on is a one-word change.
- **Results are framed as untrusted.** Snippets are capped at 300 chars via
  OpenHuman's UTF-8-safe truncator, whitespace-flattened so a snippet cannot fake
  a numbered result entry, and wrapped in a fence carrying a per-call nonce (a
  page indexed yesterday cannot contain a token minted on this call). Discovery
  and retrieval stay layered: search returns metadata only; page content still
  flows through the SSRF-guarded `web_fetch`.
- **`GATEABLE_NAMESPACES` grows to 7.** `search` is uncapped under `unlimited`
  and denied by `free`/`starter`/`pro` unless opted into — the media/composio
  shape. Note a *token* budget on `search` is a blunt cross-subsidy gate; the real
  ceiling is the daily call cap, because a search costs money per call and no
  tokens at all.

### Not fixed here (found, deliberately left)

- **`budget_usd_daily` still enforces nothing.** Confirmed zero readers of the
  getter. Wiring a per-agent USD gate is a Control-pillar change of its own and
  would have been invisible inside this diff.
- **Approvals still do not resume.** A parked tool call records the verdict but is
  never re-dispatched, so under `supervised` the `web` namespace (`web_fetch`,
  `curl`, `http_request`) is effectively unusable today. That is pre-existing and
  is the reason the search classification matters; fixing it is separate work.
- **No per-tenant BYO engine key** and **no cross-replica cap** — both stated
  above with their reasons.

## Tests

Nothing here uses or requires a real search-provider key or inference
credential. The turn tests stub two loopback endpoints — a scripted
OpenAI-compatible `/chat/completions` for the model's choices, and the managed
backend's own `/agent-integrations/parallel/search` route — so the real harness
path runs end to end with only the network boundary faked.

New end-to-end turn tests (`harness::search_turn_test`), driving a real
`HarnessPool` / `build_agent` / `ApprovalPolicy` / `UsageMeter`:

- a real **supervised** turn searches, the citations reach the *model's* context,
  the managed bearer is presented, and exactly one priced sample lands in the
  meter the console reads;
- a turn past the daily cap is refused **without reaching the backend**, the
  refusal reaches the model, and nothing is metered;
- a `*`-granted turn is never offered the tool;
- a `readonly` desk is **denied** the call even though the tool is wired.

Plus unit coverage for the ledger (cap, UTC-day reset, per-company isolation,
refund, no-underflow, shared-handle semantics), credential redaction, argument
handling, rendering (fence nonce, snippet cap/flattening, empty-result honesty,
local result-count cap, domain derivation), the three build-time gate states,
the three policy modes, the sample shape, and the REST usage projection.

Commands run locally:

- [x] `cargo test`
- [x] `cargo test --features openhuman,tinycortex --lib` (the CI gated lane)
- [x] `cargo check --features openhuman,mcp,telegram`
- [x] `cargo test --features openhuman,mcp,telegram --lib harness::build::tests` (the new lane)
- [x] `cd frontend && npm ci && npm run typecheck && npm run build`

Two pre-existing failures outside this change were not touched:
`cargo check --all-features --all-targets` (two `E0034` in the sqlite store,
#288) and `clippy -D warnings` unscoped under harness features (promotes
`large_enum_variant` inside vendored `tinyagents`; scoped with `--no-deps` as CI
does).

## Documentation

- `docs/spec/runtime/manifest.md` — the `search` namespace: explicit-grant rule,
  managed credential, the daily call cap, why individual searches do not park,
  why `readonly` still denies, and why there is no `search` Cargo feature. The
  `[tools]` example now shows `search` and `search_daily_calls`.
- `skills/{web-research,seo-audit,competitor-scan}/SKILL.md` — each names
  `web_search` for discovery and `web_fetch` for reading, and gains an explicit
  degradation clause: if search is unavailable, say so and work from
  operator-supplied URLs — **never fabricate citations**.
- Module docs on `harness::search`, `harness::toolbelt` (search removed from the
  deferred list, with the reason), `harness::build`, `metering::search`, and the
  contract test explaining exactly which entry left the forbidden array and why.

### Coordination

- **#296** (`#237`, open) — touches `HarnessDeps`, `build_agent`, `policy.rs`,
  `build.rs` tests. Read its diff before writing this; same seam, additive
  fields/blocks, no logical conflict. Textual conflicts are likely in the
  `HarnessDeps` struct literals and the `build.rs` test module; both resolve by
  keeping both fields.
- **#294** (`#239`, open) — `skills_registry` on the same struct literals; same
  keep-both resolution.
- **#295** (`#281`, open) and **#292** (`#251`, open) — both edit `ci.yml`; this
  adds one step at the end of the `rust-gated` job.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in web search with managed credentials and configurable daily call limits.
  * Search usage is metered separately and reported through the usage dashboard and API.
  * Capability status now displays search availability, grant status, credentials, and budget.
  * Added explicit search support to competitor, SEO, and web-research workflows.

* **Policy**
  * Search is allowed in supervised and full modes, but denied in read-only mode.
  * Search remains unavailable without an explicit grant and configured credentials.

* **Documentation**
  * Updated manifest and workflow guidance for search permissions, limits, citations, and fallback behavior.


oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai full review

PR #345, opened 2026-08-04
feat(console): edit a saved workflow (#259)

### LINKED ISSUE #259: bug: a saved workflow cannot be edited or deleted
> **Scope update (2026-08-04).** PR #279 shipped most of this. The backend, the API client and the delete affordance are all live on `main`. **The one remaining piece is edit mode in the workflow dialog** — and it was left out deliberately, not forgotten. The body below has been rewritten to describe only that remaining work, with everything the implementer needs already established.

## What already exists on `main` — do not rebuild it

| Piece | Where | State |
|---|---|---|
| `PUT /workflows/{wid}` | `src/server/ops/workflows.rs` | Live, validated, version-guarded |
| `DELETE /workflows/{wid}` | same | Live, durable across restart |
| `updateWorkflow(client, company, wid, graph, expectedVersion)` | `frontend/src/api/workflows.ts` | Exported, unused |
| `deleteWorkflow(...)` | same | Exported, wired to the UI |
| `editable` / `version` on the read | `getWorkflow` | Live |
| Delete button + confirm + 409 conflict banner | `frontend/src/views/WorkflowsView.tsx` | Live |

**No Edit button ships today, by design.** Edit needs the dialog to accept an existing graph, and a visibly-broken Edit button is worse than none. `updateWorkflow` is built and tested at the API level; the `editable` flag and the conflict banner are already exercised *through Delete*, so none of it is dead code waiting on this issue.

## The remaining work

Give `WorkflowCreateDialog` an edit mode, add the Edit button in `WorkflowsView`, and call the `updateWorkflow` that already exists.

### The contract, already settled

- **Hydrate from `getWorkflow`**, which returns `editable` and `version` alongside the graph.
- **Send `graph.version` as `expectedVersion`** on save. On `ApiError.status === 409`, call `setConflict(e.message)` — the banner, its Reload button and the `graphTick` refetch already exist in `WorkflowsView` and need no change.
- **Treat only `editable === false` as a refusal.** A host predating #259 sends no field at all, and absent must not read as false.
- **The id may not change through `PUT`** — the host answers `400`. If the dialog exposes the id field in edit mode it must be read-only. A rename is a create plus a delete.

### Two traps that will cost you an afternoon if missed

**1. `fieldErrors` must be reset on hydrate.** The dialog keeps `fieldErrors: Record<string, string>` keyed `` `${node.key}:${field}` `` (helper `errorKey`, field ∈ `schedule` | `destinationTarget`). It is cleared in the dialog-open effect. An edit mode that populates the draft *without* going through that effect will carry stale errors across graphs.

**2. Mint fresh `key`s — never reuse saved ids as keys.** `fieldErrors` is keyed on `node.key`, the stable row key, **not** `node.id`. Hydration must therefore allocate new keys via `blankNode()` / `nextKey()` rather than reusing the saved ids. Reusing them collides the error map across graphs.

Also: `changeKind()` and the `fieldErrors` clearing inside `updateNode()` are a matched pair. Anything you add to `DraftNode` behind a `node.kind === …` control belongs in **both**.

### Shapes you are hydrating into

Unchanged since before #279 — no field was added, removed or renamed:

```ts
DraftNode { key, id, kind, name, summary, agent, schedule, destinationKind, destinationTarget }
DraftEdge { key, from, to, label }
```

`NodeRow` and `ScheduleField` gained required props (`client`, `company`, `errors`, `onValidateField` / `error`, `onBlurValidate`). Any new render site must pass them.

`WorkflowsView` already carries state you should reuse rather than duplicate: `deleting`, `conflict`, `graphTick`. The graph-load effect clears `conflict` on any successful read.

## One open design question — decide it, don't skip it

OpenHuman's `flows_update` enforces a rule this repo has no equivalent of: an edit that turns a **manual** trigger into an **automatic** one forces `enabled = false`, so a schedule cannot go live without review.

OpenCompany cannot do that today — `WorkflowScheduler::tick` deliberately does not gate on `[workflows].enabled` at all, so writing `false` would stop nothing. The current behaviour is pinned by a test named `adding_a_schedule_by_edit_arms_the_workflow_on_the_next_tick`.

Note this makes an *edit* no riskier than a *create*, which already persists a live schedule the same way. So it is defensible to leave as-is — but say which you chose in the PR rather than leaving it implicit.

## Acceptance

- [ ] An author can open a saved workflow, change a schedule, a node kind, an agent or a destination, and save it
- [ ] The id field is not editable
- [ ] A stale save surfaces the existing conflict banner rather than silently overwriting
- [ ] A seed-file-backed workflow (`editable: false`) offers no Edit button, with the same explanation Delete already gives
- [ ] Field errors from one graph never appear on another
- [ ] Verified **in a browser**, not only by curl — the whole remaining scope is UI

## Verification notes

There is a Playwright harness at `frontend/test/e2e/` with an existing `workflow-edit-delete.spec.ts` to extend. CI does not run it (no `webServer`), but it is how UI behaviour actually gets verified here.

Isolate any host you run with the `--home <dir>` CLI flag. `OPENCOMPANY_DATA_DIR` alone does not move the company store, and a host started without `--home` writes into a bundle other sessions share.

## Original report

A workflow was write-once: the ops router exposed only create, list, read and run, so a typo in a cron expression, a wrong destination, or a node pointed at the wrong agent was permanent. The only recovery was to author a second workflow and leave the broken one in the list, where it still showed in the picker and still fired on its schedule. That lands harder once a trigger can carry a schedule (#169) and an output node a destination (#170) — exactly the fields an author gets wrong on the first pass.

## Related

- #279 — shipped the backend, the API client and delete
- #169 / #170 — schedule and destination, the fields this makes correctable

M3gA-Mind:
## Summary

Closes the last piece of #259: **edit mode in the workflow dialog**, wired to the
`updateWorkflow` and the `version` token that #279 already shipped. Nothing on the
backend, the API client or Delete was rebuilt.

`WorkflowCreateDialog` now takes an optional `workflow`. When it is set the same form
hydrates from that saved graph and saves through `updateWorkflow(client, company,
graph.id, graph, graph.version)`; `null` is the create path, unchanged. `WorkflowsView`
gains an Edit button beside Delete, behind the same `editable === false` predicate and
the same explanation, and a save swaps the graph in place with its fresh token.

Four things worth reading closely, three of them called out by the issue:

* **Hydration goes through the dialog-open effect** — the one place `fieldErrors` is
  cleared. A draft populated anywhere else carries the previous graph's errors onto rows
  that no longer exist.
* **Fresh row keys.** Every hydrated row is minted by `blankNode()`/`nextKey()`. Reusing
  saved node ids as keys would collide the error map across graphs, and most graphs have
  a trigger node called `start`.
* **The id is read-only in edit mode**, not merely rejected on save. The host answers 400
  to a rename, so offering the field would be a trap; a line under it says why.
* **An edit no longer silently deletes what the form cannot show.** An overlay graph can
  carry `config`, `onError`, `retry` and `requiresApproval` (`POST`/`PUT` accept them and
  the orchestrator's own `create_workflow` tool writes them). Rebuilding a node from the
  visible controls alone would drop a retry policy or an approval gate on the first save,
  so those fields ride on the draft row and go back out verbatim. `config` is
  kind-specific and is dropped by `changeKind()` along with every other kind-conditional
  field — the matched pair the issue names.

**On the 409.** The dialog stays open with the host's message inline, so the author keeps
what they typed, *and* calls `setConflict(e.message)` so the existing persistent banner
and its Reload sit behind it with the way out. The banner, its Reload and the `graphTick`
refetch are untouched. Reload re-hydrates an open dialog with the fresh graph and token,
which discards the edit — the honest outcome, since keeping it would keep the stale token
with it. `PUT` also answers 409 for a display name already taken; that lands in the same
banner rather than being sniffed out of the message text, which would couple the console
to host copy.

## The open design question

**Decided: an edit does not disarm a schedule.** OpenHuman's `flows_update` forces
`enabled = false` when an edit turns a manual trigger automatic. This host gains no
equivalent here, because `WorkflowScheduler::tick` deliberately does not gate on
`[workflows].enabled` at all — writing `false` would stop nothing, and
`adding_a_schedule_by_edit_arms_the_workflow_on_the_next_tick` pins that. An edit is
therefore exactly as live as a create, which already persists a running schedule the same
way. Reversing it means first making the scheduler honour `enabled`, which is the
enable/disable follow-up already listed on #259. Written down in
`docs/modules/server/README.md` rather than left implicit.

## API Or Behavior Changes

Console only; no route, payload or Rust behavior changed.

* Workflows view: an **Edit** button beside Delete, disabled with the same explanation for
  a source-defined workflow (`editable === false`; an absent field still reads as
  editable).
* The workflow dialog reads "Edit workflow" / "Save changes" when editing, with a
  read-only id.
* `WorkflowCreateDialog` props: `onCreated` is now optional, and `workflow` / `onSaved` /
  `onConflict` are new. Both existing call sites are in this repo.
* The Delete tooltip now reads "can't be changed or removed" rather than "can't be
  removed" — one refusal, one sentence, matching the host's own message.

## Tests

Extended `frontend/test/e2e/workflow-edit-delete.spec.ts` (the harness the issue points
at) with four specs: an edit round trip the host actually stored, the read-only id, a
stale save that surfaces the conflict banner and writes nothing before succeeding after a
Reload, and a field error raised on one graph that must not appear on another. The whole
remaining scope is UI, and three of those four are observable nowhere else.

Run locally, from `frontend/`:

- `npm ci`
- `npm run typecheck` — clean
- `npm run typecheck:e2e` — clean (this is what compiles the new specs)

**Not run locally, stated plainly:** the Playwright suite and the browser walk-through the
issue's acceptance list asks for. Both need a live host, which means a full local build
matrix that is prohibited on this machine. The specs are written for the harness and are
compiled by CI's `typecheck:e2e`; the `Console` job also runs `npm run build`.

No Rust file is touched by this branch.

## Documentation

`docs/modules/server/README.md`, in the existing "Editing and removing a workflow (issue
#259)" section: what the console now offers, and the schedule decision above.

Closes #259.

## Summary by CodeRabbit

* **New Features**
  * Added workflow editing from the workflow list.
  * Edit dialogs load existing workflow details and keep workflow IDs read-only.
  * Workflow changes preserve advanced node settings and update schedules on the next scheduler cycle.
  * Added conflict handling for stale edits without automatic retries.

* **Bug Fixes**
  * Prevented editing and deletion for workflows marked as non-editable.
  * Improved error handling so validation errors remain isolated to the affected workflow.

### DIFF
--- a/docs/modules/server/README.md
+++ b/docs/modules/server/README.md
@@ -165,6 +165,26 @@ curl -X DELETE "$HOST/api/v1/company/workflows/weekly_digest?expectedVersion=a60
      -H "Authorization: Bearer $TOKEN"
 ```
 
+**In the console.** The Workflows view offers Edit and Delete side by side, both
+greyed out with the host's own explanation when `editable` is `false` — an
+explicit `false` only, since a host predating this sends no such field and
+`undefined` must not read as a refusal. Edit reopens the create dialog hydrated
+from the saved graph, with the **id read-only** (a rename is a `400`, so
+offering the field would be a trap) and `version` sent as `expectedVersion`. A
+`409` leaves the dialog open with the host's message and raises the same
+persistent conflict banner Delete uses, whose Reload re-reads the graph and its
+token; nothing is retried without one.
+
+**An edit does not disarm a schedule.** OpenHuman's `flows_update` forces
+`enabled = false` when an edit turns a manual trigger into an automatic one, so
+a new schedule cannot go live unreviewed. This host has no equivalent and
+deliberately gains none here: `WorkflowScheduler::tick` does not gate on
+`[workflows].enabled` at all, so writing `false` would stop nothing, and
+`adding_a_schedule_by_edit_arms_the_workflow_on_the_next_tick` pins that. An
+edit is therefore exactly as live as a create, which already persists a running
+schedule the same way. Reversing this means first making the scheduler honour
+`enabled`, which is the enable/disable follow-up below.
+
 **Deliberately not in #259**, each its own follow-up: revision history and
 rollback (OpenHuman keeps a bounded snapshot ring in a dedicated table; our
 overlay bodies live inside `CompanyRecord`, which is saved whole on every write,
--- a/frontend/src/views/WorkflowCreateDialog.tsx
+++ b/frontend/src/views/WorkflowCreateDialog.tsx
@@ -3,6 +3,12 @@
 // are restricted to the ones the engine actually executes today
 // (`CREATABLE_NODE_KINDS`); `tool_call`/`http_request` stay off the palette
 // until they're wired (see `src/workflows/caps.rs`).
+//
+// It is also the EDITOR (issue #259): pass a `workflow` and the same form
+// hydrates from that saved graph and saves through `updateWorkflow`, carrying
+// the graph's `version` as the optimistic-concurrency token. One component
+// rather than two because an edit is the same form with the same rules — a
+// second one would drift the moment either side grew a field.
 
 import { useEffect, useId, useState } from "react";
 import { Plus, Trash2 } from "lucide-react";
@@ -11,12 +17,14 @@ import {
   CREATABLE_NODE_KINDS,
   DESTINATION_KINDS,
   createWorkflow,
+  updateWorkflow,
   type WorkflowDestination,
   type WorkflowEdge,
   type WorkflowGraph,
   type WorkflowNode,
 } from "@/api/workflows";
 import type { OpenCompanyClient } from "@/api/client";
+import { ApiError } from "@/api/types";
 import { CronPreviewLine } from "@/views/CronPreviewLine";
 import type { TeamMemberDto } from "@/api/types";
 import { Alert, AlertDescription } from "@/components/ui/alert";
@@ -57,6 +65,24 @@ interface DraftNode {
   destinationKind: "" | WorkflowDestination["kind"];
   /** The address (`email`) or channel id (`channel`). Unused for `owner`. */
   destinationTarget: string;
+  /**
+   * Fields the form has no control for, carried through an edit **verbatim**
+   * (issue #259).
+   *
+   * An overlay graph can carry them — `POST`/`PUT` accept them and the
+   * orchestrator's own `create_workflow` tool writes them — so a graph authored
+   * outside this dialog can reach it. Rebuilding the node from the visible
+   * controls alone would then quietly delete a retry policy or an approval gate
+   * on the first save, which is a worse bug than the write-once one this fixes.
+   *
+   * They ride on the ROW, not on the node id, so they follow the row when its
+   * id is edited. `config` is the exception: it is kind-specific, so
+   * {@link changeKind} drops it along with every other kind-conditional field.
+   */
+  config?: unknown;
+  onError?: string;
+  retry?: WorkflowNode["retry"];
+  requiresApproval?: boolean;
 }
 
 /** "No schedule" — the workflow runs only when something starts it. A sentinel
@@ -194,7 +220,18 @@ function nextKey(): string {
  * this reset.
  */
 function changeKind(kind: string): Partial<DraftNode> {
-  return { kind, agent: "", schedule: "", destinationKind: "", destinationTarget: "" };
+  return {
+    kind,
+    agent: "",
+    schedule: "",
+    destinationKind: "",
+    destinationTarget: "",
+    // Kind-specific by definition (a `switch`'s cases, a `sub_workflow`'s
+    // target), so it means nothing on the new kind — and unlike the fields
+    // above there is no control that could ever show what was dropped. The
+    // kind-agnostic policies (`onError`, `retry`, `requiresApproval`) are kept.
+    config: undefined,
+  };
 }
 
 /** A blank node row, so every construction site stays in step as the shape grows. */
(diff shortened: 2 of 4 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Addressed the nitpick: the edit round-trip spec now creates its fixture with `config`, `onError`, `retry` and `requiresApproval` on the report node, and asserts all four survive a schedule-only edit. That is the behaviour the draft rows carry those fields for, so it is worth a pin rather than a comment.

@coderabbitai review

oxoxDev (commented):
## Claims

- Hydration sits inside the dialog-open effect (`frontend/src/views/WorkflowCreateDialog.tsx:361-384`), the only `setFieldErrors({})` site (`:369`); `workflow` is in the deps, so Reload re-hydrates rather than leaking a draft past the clear. ✔
- `draftNodes` / `draftEdges` (`:265-291`) both mint keys via `blankNode()` → `nextKey()`; no saved node id is reused as a row key. ✔
- The id is genuinely `readOnly` in edit mode with an explanation line (`:634`), not rejected at save. ✔
- No silent structure loss. Every field of the read DTO (`src/server/ops/workflows.rs:200-222` — `summary`/`agent`/`schedule`/`config`/`onError`/`retry`/`requiresApproval`/`destination`) rides on the draft row and is re-emitted at `WorkflowCreateDialog.tsx:529-559`; the write DTO (`CreateNode`, `:425`) accepts the same camelCase set, `retry.maxAttempts` included. `changeKind` (`:222-235`) drops only `config`, deliberately. Edge and graph levels have no unmodelled fields. ✔
- Stale token: the host answers 409 "changed since you loaded it… reload", nothing is written, and the message lands both inline and in the persistent banner. No lost update. ✔

## Findings

- **major** — `frontend/src/views/WorkflowCreateDialog.tsx:806`: the Node kind `<Select>` only offers `CREATABLE_NODE_KINDS`, so a saved node of kind `tool_call` / `http_request` / `switch` / `split_out` / `output_parser` / `sub_workflow` renders an **empty** control. Those are exactly the graphs carrying the `config` this PR protects — `POST` validates against `WORKFLOW_NODE_KINDS`, not the palette, and the orchestrator's `create_workflow` writes them. An untouched save preserves the kind, but a blank control invites a pick, and any pick runs `changeKind` → `config: undefined`, dropped silently with nothing on screen that ever showed it. Render an off-palette kind as a disabled item (or plain text).
- **nit** — `frontend/src/views/WorkflowsView.tsx:638`: `open={editOpen && graph !== null}` force-closes without an `onOpenChange`, so `editOpen` stays `true`. A company switch clears `graph`, the list effect auto-selects the first workflow (`:148-151`), and the dialog re-opens on a graph nobody asked to edit. Clear `editOpen` when `graph` goes null.
- **nit** — `frontend/src/views/WorkflowCreateDialog.tsx:603`: the name-already-taken 409 lands in the same banner whose only action is Reload, which re-hydrates and discards the edit without fixing the clash. Not sniffing host copy is right; the banner action is still a dead end on that arm.
- **nit** — `frontend/src/views/WorkflowsView.tsx:374`: the explanation asserts a source-tree file, but `is_editable` is `has-overlay && !seed-file`, so a manifest-enabled id with no saved body is also `editable === false` with no `workflows/<id>.toml` to edit. Pre-existing wording, now carried by two buttons.

## Against the issue

- **Acceptance, one by one:** edit-and-save a schedule / agent / destination ✔ (node kind only for on-palette kinds — see the `major`); id not editable ✔; stale save surfaces the existing banner rather than overwriting ✔; a seed-backed workflow refuses Edit with the same treatment Delete gets ✔ (disabled + title, matching Delete rather than hiding it); field errors never cross graphs ✔ (spec written).
- **major** — the sixth criterion, "verified **in a browser**, not only by curl", is unmet and the author says so plainly. The whole remaining scope is UI, and the off-palette blank kind control is precisely what one pass over an orchestrator-authored graph surfaces. Run it or waive it explicitly before merge.
- **Earlier pieces verified on main, not assumed:** `update_workflow` / `delete_workflow` / `is_editable` / version token in `src/server/ops/workflows.rs`; `updateWorkflow` + `deleteWorkflow` in `frontend/src/api/workflows.ts`; Delete, confirm gate and conflict banner testids in `WorkflowsView.tsx`. #279 merged 2026-08-03. Main carries no Edit affordance and no `updateWorkflow` call site, so #259's premise holds and this is genuinely the remainder.
- **The fourth callout is the author's, not the issue's.** #259 names only the two traps (effect-scoped hydration, fresh row keys) and the `changeKind`/`updateNode` matched pair; it never asks about `config` / `onError` / `retry` / `requiresApproval` surviving a round trip. Catching that unprompted is the strongest thing in the PR.
- **The open design question is answered, not skipped:** an edit does not disarm a schedule, grounded in `WorkflowScheduler::tick` not gating on `[workflows].enabled` and the test that pins it. Correctly changes no Rust.
- **No scope creep.** Outside the dialog and the view: a docs paragraph (#259 asked for the decision to be stated), the Delete tooltip re-word, and `onCreated` becoming optional — both call sites in-repo.

Issue #318, opened 2026-08-03
connections: OAuth state signing key falls back to a shipped literal when unconfigured


oxoxDev:
## Summary

The signing key for the OAuth `state` nonce has a default value baked into the source. When its environment variable is unset — which is the state of any host that has not been explicitly configured — every `state` is signed with a value that is public, identical across all such deployments, and present in the repository.

The nonce is what binds an in-flight authorization to one company, one provider and an expiry, and the callback verifies it before exchanging the code. With a known key that verification confirms only that the value is well-formed, not that we issued it.

## Why it matters

The `state` check is the flow's CSRF defence. Its purpose is to ensure the callback we handle belongs to a handshake we started. If the key is knowable, a `state` can be constructed rather than obtained, and the check stops distinguishing the two.

This is a configuration-hardening gap rather than a defect in the verification logic — the signing and decoding are fine, and a properly configured host is unaffected. What is wrong is the failure mode: an unconfigured host silently degrades to a weaker guarantee instead of refusing to operate.

## Suggested direction

Fail closed rather than falling back:

- With the variable unset, refuse to start the OAuth flow and say why, instead of signing with a default. Connections are already read-only without the feature configured, so a clear refusal is consistent with how the surface already behaves.
- Alternatively, derive a per-installation value at first boot and persist it, so a self-hosted operator gets a unique key without extra setup. Either is acceptable; the shared literal is not.
- Whichever is chosen, add a test asserting that no signing occurs with the fallback value.

## Notes

Found while working on #300 and deliberately kept out of that PR — it is unrelated to the redirect handling and did not belong in a change about onboarding.

Scope: **small**.

## Related

- #300 — OAuth redirect handling in the same module
- #316 — connection ownership and account rules

PR #347, opened 2026-08-04
fix(platform-auth): authenticate every tenant token before trusting it (#322)

### LINKED ISSUE #322: platform auth: the offline verifier accepts unsigned caller-supplied claims
## Summary

The default platform-token verifier parses an unsigned, caller-supplied payload directly into platform claims. A request carrying a hand-constructed bearer of that shape is accepted as a platform principal, with whatever tenant, scopes and company allow-list the caller wrote into it.

This is the verifier the default build ships. The signature-checking implementation exists, but it is constructed only in test code, and CI never exercises it: the lane that compiles the relevant feature set is check-only and never runs the tests.

## Where it is

`StaticPlatformVerifier` has two arms. One is an exact-match shared secret — genuinely authenticated, since it is authenticated by *knowledge of the secret*. The other decodes an `oc_tenant.<base64url(json)>` bearer straight into claims with no authentication at all. The file describes that second arm as insecure in its own comment.

The downstream gates are sound — the platform scope gates provision/pause/resume/suspend/archive, and address authorization plus the company allow-list are enforced on top. The machine-versus-human split is type-enforced: claim resolution can never return a user principal. **The gates are fine; the input they trust is not.**

## Direction — settled, don't re-litigate

1. **The exact-match shared-secret arm stays in production.** It is the same pattern the control plane itself uses for its admin token, and it is the only machine-credential flow anything actually uses today.
2. **The unsigned `oc_tenant.` arm is removed from production entirely** and survives only as a test-constructed type. Not a `cfg(test)` branch *inside* `verify` — that would make the production rejection unobservable to unit tests, since the test build would still accept unsigned tokens. Separate **types**.
3. **Tenant-scoped machine tokens in production become signed JWTs.**
4. **The signing feature joins the default feature set.**

Rejected: a loud env opt-in. There are **zero** production consumers of unsigned tenant tokens — the control plane injects no platform credential at all, and nothing in the repo, the superproject, the frontend, the e2e harness, the scripts or the docs mints a token of that shape. An off-switch for authentication would preserve a footgun nobody needs, which is the classic bypass-flag band-aid this repo bans.

Rejected: full deletion of the codec. The offline encoder is what lets the scope / allow-list / ownership suites run without signing machinery, and those suites should keep passing **unchanged**.

## The CI half is part of the fix, not a footnote

Because cargo features are additive, moving the signing feature into the default set makes **both** existing test lanes execute the JWT tests. No new CI lane is needed.

**Prove the gate by commit ordering.** Push the feature-widening commit *plus* the new regression tests **alone**, first. That run must go **red** — demonstrating both that the lane now executes those tests and that the bug is real. Only then push the fix. Do not fake a break, and do not claim the gate works without showing it failing. Flag the deliberate red in the PR body so a reviewer reads it as the proof rather than as breakage.

## Fail closed, loudly

A host with a signing secret set but the feature absent must **refuse to boot** with a named message — the same precedent the storage backend already sets, where selected-but-unavailable aborts rather than silently falling back. Print a boot line naming the active mode (shared-secret / jwt / both). Mode only — **never** secret material.

There must be no configuration state that fails open: no credential → 401; unsigned tenant bearer → 401; signing secret on a feature-less build → refuses to boot; wrong secret → 401.

## Key handling

The signing secret has **no shipped literal and no fallback** — unset means the path does not exist. Do not introduce a default; that is the shape of the separate #318 problem and it must not be recreated here.

Note this issue is *not* that shape — the platform token has no baked default either. Its flaw is an unauthenticated **parse**, which is worse in kind.

**Never print, paste or commit a real secret** in a commit, a PR body, a doc or an issue comment. Docs use placeholders.

## Known collision

**PR #325 (issue #315) touches the same startup wiring.** Whichever lands second needs a small rebase. Check before starting.

## Acceptance

- [ ] A hand-constructed unsigned tenant bearer is rejected with 401 by the shipped default build
- [ ] The exact-secret flow still works verbatim
- [ ] Signed tenant tokens verify, and expired or tampered ones are refused
- [ ] A signing secret set without the feature refuses to boot with a message naming the cause
- [ ] The existing scope / allow-list / ownership suites pass with only construction-line changes
- [ ] The regression test demonstrably ran red before the fix
- [ ] Boot logs name the auth mode and never the secret

## Two limits to state rather than fix

- **The signing is symmetric** — a workload that can verify could also mint. Acceptable while the trust boundary is a single workload; asymmetric keys via the control plane's issuer are the follow-up. Do not widen this PR into that.
- **A token with no expiry currently never expires**, because the verifier clears its required-claims set. Leave as-is and name it as a follow-up; changing it is a separate policy decision.

## PR framing

Neutral, per repo convention: describe the behaviour change plainly. **No** "vulnerability", "exploit", "CVE" or "GHSA" language in commit messages or the PR body.

## Gates

`cargo fmt --all -- --check`; `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`; `cargo test --locked --features openhuman,tinycortex --lib`; `cargo check --locked --all-features --all-targets`; plus plain `cargo test --locked` — that is the lane whose behaviour this changes.

Ships as a standalone PR: it changes what a default build accepts, so it wants its own review.

M3gA-Mind:
## Summary

Closes #322.

`StaticPlatformVerifier` had two arms. The exact-match shared secret is genuinely authenticated — by knowledge of the secret. The other decoded an `oc_tenant.<base64url(json)>` bearer straight into `PlatformClaims`, so a caller who knew nothing could hand-build a bearer that resolved to whatever tenant, scopes and company allow-list they wrote into it. The downstream gates were sound — the platform scope gates provision/pause/resume/suspend/archive, address authorization and the allow-list are enforced on top, and claim resolution is type-prevented from returning a user principal. The gates were fine; the input they trusted was not.

**What changes**

1. **The shared-secret arm stays in production**, verbatim. It is the same pattern the control plane uses for its own admin token, and it is the only machine-credential flow anything uses today.
2. **The unsigned arm is removed from production entirely.** It survives only as `UnsignedTenantVerifier`, a `cfg(test)` type. A separate **type**, not a `cfg(test)` branch inside `verify` — a branch there would make the production rejection unobservable to unit tests, since the test build would still accept unsigned tokens. The offline encoder is what lets the scope / allow-list / ownership suites run without signing machinery, so those suites keep their coverage with construction-line changes only.
3. **Tenant-scoped machine tokens in production are signed JWTs.** `platform_auth::configure` selects the verifiers from the environment; both credentials may be set, and a bearer accepted by either is accepted.
4. **`platform-jwt` joins the default feature set** — see the CI section below.

**Fail closed, loudly.** No configuration state fails open: no credential → `401`; an unsigned tenant bearer → `401`; a wrong secret → `401`; a signing secret on a build compiled without `platform-jwt` **refuses to boot** with a message naming the variable and the missing feature, rather than degrading to the weaker check — the same precedent the storage backend sets, where selected-but-unavailable aborts instead of silently falling back. Boot prints the active mode (`platform auth: shared-secret | jwt | both`). Mode only, never secret material.

**Key handling.** `OPENCOMPANY_PLATFORM_JWT_SECRET` has no shipped literal and no fallback — unset means the path does not exist. Deliberately not the shape of #318; no default is introduced.

## Deliberate red run — please read this as the proof, not as breakage

The CI half is part of the fix. Because `platform-jwt` was opt-in and no lane enabled it, the signed verifier's tests were compiled by the check-only `cargo check --locked --all-features --all-targets` step and **run by nothing**. Cargo features are additive, so moving `platform-jwt` into the default set makes *both* existing test lanes execute them. No new lane is needed.

The two commits are ordered so the gate proves itself:

- **`be77d63`** — the feature widening plus the new regression tests, **alone**. This run is **red**, on `hand_constructed_tenant_bearer_is_refused`. That single failure demonstrates both that the widened lane really executes these tests and that the bug was real. Nothing was faked to produce it.
- **`0e2a94e`** — the fix. Green.

Run for the red commit: https://github.com/M3gA-Mind/opencompany/actions/runs/30906312125

**Read the `Rust` job, not the run badge.** The run as a whole is labelled *cancelled*, because the follow-up push landed while `Rust (openhuman, tinycortex)` was still building and CI cancels in-progress runs on the same ref. The `Rust` job itself ran to completion and concluded `failure`:

```
6. Format: success
7. Clippy: success
8. Test:   failure

thread 'server::platform_auth::test::hand_constructed_tenant_bearer_is_refused' panicked at
  src/server/platform_auth.rs:468:9
test result: FAILED. 1177 passed; 1 failed; 1 ignored
```

Two things are visible there. `Format` and `Clippy` pass, so the red is the assertion and not a compile or lint error. And in the same run:

```
test server::platform_auth::test::jwt_verifier_round_trips_signed_claims ... ok
test server::platform_auth::test::jwt_verifier_refuses_an_expired_token ... ok
test server::platform_auth::test::jwt_verifier_refuses_a_tampered_payload ... ok
```

`jwt_verifier_round_trips_signed_claims` already existed on `main` and previously ran in **no lane at all**. Seeing it execute is the CI-hole half of the proof; seeing `hand_constructed_tenant_bearer_is_refused` fail is the bug half. The gated lane runs the same tests, so cancelling it costs no evidence.

## API Or Behavior Changes

**A default build no longer accepts an `oc_tenant.<base64url(json)>` bearer.** Any caller sending one now gets `401`. There are zero production consumers of this shape — the control plane injects no platform credential at all, and nothing in this repo, the superproject, the frontend, the e2e harness, the scripts or the docs mints a token of that shape — so this is expected to be a no-op in practice. A loud env opt-in was considered and rejected: an off-switch for authentication preserves a footgun nobody needs.

Public API on `server::platform_auth`:

- Removed: `StaticPlatformVerifier::TENANT_PREFIX`, `StaticPlatformVerifier::tenant_token`.
- Added: `configure`, `PlatformAuthMode`, `PLATFORM_TOKEN_ENV`, `PLATFORM_JWT_SECRET_ENV`.
- `JwtPlatformVerifier` is unchanged but now compiled by the default build.

New environment variable `OPENCOMPANY_PLATFORM_JWT_SECRET` (HS256, tenant-scoped tokens). `OPENCOMPANY_PLATFORM_TOKEN` is unchanged in meaning.

Default features change from `["oauth"]` to `["oauth", "platform-jwt"]`, adding one pure dependency (`jsonwebtoken`) already present in `Cargo.lock`. `Cargo.lock` is untouched.

**Two limits stated rather than fixed**, both documented in the module header and `docs/spec/runtime/config.md`:

- The signing is **symmetric** — a workload that can verify could also mint. Acceptable while the trust boundary is a single workload; asymmetric keys via the control plane's issuer are the follow-up. Deliberately not widened into this PR.
- A token with **no `exp` never expires**, because the verifier clears its required-claims set. Left as-is; changing it is a separate policy decision.

## Tests

New tests in `src/server/platform_auth.rs`:

| Test | Asserts |
| --- | --- |
| `hand_constructed_tenant_bearer_is_refused` | The regression test. Builds the bearer literally from the wire shape — no helper, no shared constant — and asserts the shipped verifier rejects it |
| `jwt_verifier_refuses_an_expired_token` | A past `exp` is refused |
| `jwt_verifier_refuses_a_tampered_payload` | Claims rewritten to grant `platform`, original signature kept, refused |
| `no_credential_configures_no_platform_auth` | Neither variable set → no machine credential. A blank value is unset, not an empty accepted bearer |
| `shared_secret_alone_configures_shared_secret_mode` | Exact secret works verbatim; anything else `401`s |
| `signing_secret_alone_configures_jwt_mode` | Signed token verifies; the shared-secret arm is not silently wired |
| `both_secrets_accept_either_credential` | Either credential is accepted; an unsigned or unknown bearer still is not |
| `signing_secret_without_the_feature_refuses_to_boot` | Boot aborts, the message names the variable and the missing feature, and does **not** contain the secret |
| `auth_mode_names_are_stable` | The boot line's mode names |
| `unsigned_codec_round_trips_for_the_gate_suites` | The `cfg(test)` codec the gate suites depend on |

`signing_secret_without_the_feature_refuses_to_boot` reaches the refusal through a private `configure_with(.., jwt_available: bool)` seam, so the fail-closed path stays observable on a build that *has* the feature. A build without the feature reaches the identical error through `jwt_verifier`.

The scope / allow-list / ownership suites in `src/server/provision/test.rs` (4 sites) and `src/server/ops/write_test.rs` (9 sites) pass unchanged apart from swapping the verifier they construct. `src/server/users/auth_test.rs` is untouched — it only ever used the exact secret, which still works verbatim.

Nothing was run locally: this machine prohibits the full local build matrix, and `cargo fmt --all` cannot run in this clone at all because `vendor/openhuman` is not initialized. `rustfmt --edition 2024 --check` was run directly on the four changed Rust files and is clean. Everything else is verified in CI.

## Documentation

- `docs/spec/runtime/config.md` — both platform variables added to the environment table, plus a new **"How a platform bearer is authenticated"** section under Authentication: the two credential shapes, the boot line, the full no-fail-open matrix, and the two stated limits. Placeholders only; no secret material.
- `src/server/platform_auth.rs` module header rewritten to describe what a shipped build accepts, why the test-only codec is a separate type, and the two limits.
- `Cargo.toml` — the `default` feature comment records why `platform-jwt` ships and which CI blind spot that closes.

## Summary by CodeRabbit

* **New Features**
  * JWT authentication is now enabled by default for platform verification.
  * Added flexible platform authentication with shared bearer tokens, JWTs, or both simultaneously.
  * Added configuration options: `OPENCOMPANY_PLATFORM_TOKEN` and `OPENCOMPANY_PLATFORM_JWT_SECRET`.

* **Bug Fixes**
  * Invalid authentication configurations now prevent startup with clear diagnostics.
  * Authentication failures consistently return `401 Unauthorized`.
  * Improved protection against unsigned, tampered, and expired JWTs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -218,7 +218,17 @@ rand_core = { version = "0.6", default-features = false, features = ["getrandom"
 # (`github`, `telegram`, `medulla`, …) also depend on, so the default build
 # gains no new heavy native dependency. The read-only status route
 # (`ops::connections_read`) is un-gated and available regardless.
-default = ["oauth"]
+#
+# `platform-jwt` ships in the standard build because it is the *only* way a
+# shipped build can accept a tenant-scoped machine credential: the platform
+# verifier authenticates a bearer either by knowledge of the shared secret or by
+# an HS256 signature, and nothing else. Leaving it optional made the signed path
+# a build-time opt-in, and — because cargo features are additive and no CI lane
+# enabled it — its tests were compiled by the check-only `--all-features` step
+# and *run* by nothing. Moving it into the default set makes both `cargo test`
+# lanes execute them, which is why this needs no new CI lane. It costs one pure
+# dependency (`jsonwebtoken`), already in `Cargo.lock`.
+default = ["oauth", "platform-jwt"]
 tiny = ["tinyagents"]
 # WS4: embed `vendor/openhuman` (`openhuman_core`) directly as the tenant agent
 # harness via `AgentBuilder`. `src/harness/` compiles only under this feature;
--- a/docs/spec/runtime/config.md
+++ b/docs/spec/runtime/config.md
@@ -99,6 +99,8 @@ layer set it, and what is missing for each optional capability.
 | `OPENCOMPANY_MAIL_USERNAME` / `_PASSWORD` | — | SMTP auth. Redacted from `Debug` and never logged |
 | `OPENCOMPANY_MAIL_FROM_NAME` | — | Display name on the `From` header |
 | `OPENCOMPANY_CORS_ORIGINS` | — (CORS off) | Comma-separated exact origins allowed to send the session cookie cross-origin, e.g. `http://localhost:5173`. `*` is refused: a wildcard is illegal with credentials |
+| `OPENCOMPANY_PLATFORM_TOKEN` | — (no machine credential) | The shared platform secret. A bearer equal to it is the `tenant:platform` principal, with the `platform` scope |
+| `OPENCOMPANY_PLATFORM_JWT_SECRET` | — (signed tenant tokens not accepted) | HS256 secret that signs tenant-scoped machine tokens. No shipped literal and no fallback: unset means the path does not exist. Set on a build without `platform-jwt` (in the default feature set) and the host **refuses to boot** |
 
 ### Outbound mail
 
@@ -155,6 +157,33 @@ whole story, and HTTP provisioning is unavailable by construction (load
 companies with `serve --company <dir>`). A company with no admin in its
 manifest's `[users]` cannot be reached until one is listed — that is the
 bootstrap, and it is deliberate.
+
+#### How a platform bearer is authenticated
+
+Exactly two shapes, selected by the two environment variables above:
+
+- **Shared platform secret** (`OPENCOMPANY_PLATFORM_TOKEN`) — an exact match.
+  Authenticated by knowledge of the secret, the same pattern the control plane
+  uses for its own admin token. Grants the single `tenant:platform` principal.
+- **Signed tenant token** (`OPENCOMPANY_PLATFORM_JWT_SECRET`) — HS256 over the
+  `tenant` / `scopes` / `companies` claims, `exp` honored when present. This is
+  how a tenant-scoped machine token is issued.
+
+Both may be set; a bearer accepted by either is accepted. Boot prints the active
+mode (`platform auth: shared-secret | jwt | both`) and never the secret.
+
+No state fails open. No credential means every machine route answers `401`; an
+unauthenticated bearer of any shape is `401`; a wrong secret is `401`; and a
+signing secret on a build compiled without `platform-jwt` aborts boot with a
+message naming the variable and the missing feature rather than degrading to a
+weaker check.
+
+Two limits worth naming. The signing is **symmetric**, so a workload that can
+verify a token could also mint one — acceptable while the trust boundary is a
+single workload; asymmetric keys via the control plane's issuer are the
+follow-up. And a signed token carrying **no `exp` never expires**, because the
+verifier clears its required-claims set; tightening that is a separate policy
+decision.
 
 ### What changed, and why it mattered
 
--- a/src/bin/opencompany.rs
+++ b/src/bin/opencompany.rs
@@ -882,19 +882,24 @@ async fn main() -> Result<()> {
                 state = state.with_memory_overlay(overlay);
                 println!("memory backend: {:?}", storage_settings.memory_backend);
             }
-            // Platform (multi-tenant) auth: a shared platform token enables the
-            // provisioning/lifecycle surface. Without it the prosumer operator
-            // path stays in force. Real signed JWT is `platform-jwt`.
-            if let Some(token) = std::env::var("OPENCOMPANY_PLATFORM_TOKEN")
-                .ok()
-                .filter(|v| !v.trim().is_empty())
+            // Platform (multi-tenant) auth: either credential enables the
+            // provisioning/lifecycle surface. Without both the prosumer operator
+            // path stays in force. A signing secret this build cannot verify
+            // aborts boot here rather than silently degrading — same precedent
+            // as a selected-but-unavailable storage backend above.
             {
                 use opencompany::server::platform_auth::{
-                    PlatformAuthConfig, StaticPlatformVerifier,
+                    PLATFORM_JWT_SECRET_ENV, PLATFORM_TOKEN_ENV, configure,
                 };
-                state = state.with_platform_auth(PlatformAuthConfig::new(Arc::new(
-                    StaticPlatformVerifier::new(token),
-                )));
+                if let Some((platform_auth, mode)) = configure(
+                    std::env::var(PLATFORM_TOKEN_ENV).ok(),
+                    std::env::var(PLATFORM_JWT_SECRET_ENV).ok(),
+                )? {
+                    state = state.with_platform_auth(platform_auth);
+                    // The mode only — never the secret, or anything derived
+                    // from it.
+                    println!("platform auth: {mode}");
+                }
             }
             // Outbound webhooks: a URL wires the HTTP sink under `webhooks`;
             // without the feature the request is warned and dropped.
--- a/src/server/ops/write_test.rs
+++ b/src/server/ops/write_test.rs
@@ -887,13 +887,13 @@ async fn memory_operator_fact_is_injected_into_the_agent_turn() {
 #[tokio::test]
 async fn memory_is_isolated_between_companies() {
     use crate::server::platform_auth::{
-        PlatformAuthConfig, PlatformClaims, StaticPlatformVerifier,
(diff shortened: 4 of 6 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Pushed `8055cf7` addressing the review comment on `src/server/platform_auth.rs`.

Adds `jwt_verifier_refuses_an_unsigned_token` beside the existing expired/tampered tests, under `#[cfg(feature = "platform-jwt")]`. The token is assembled literally from the wire shape — `base64url({"alg":"none","typ":"JWT"}).base64url(claims).` with an empty signature — so nothing in the test picks the algorithm on its behalf.

It carries no `exp` and no registered claims, so the assertion does not lean on expiry or on `required_spec_claims`; the only thing left to reject it is the `Validation::new(Algorithm::HS256)` pin. A second assertion signs the *same* claims over HS256 and expects success, which is the control: it rules out the payload as the reason for the first rejection.

No other change in the push.

Issue #242, opened 2026-07-31
feat(runs): first-class run records with persisted trace, status, and cost


oxoxDev:
## Summary

`Run` — one attempt at a task — is one of the five core objects (Node, Task, Run, Trail event, Policy) and the only one with no type, no row and no id. There is no way to list a task's attempts, see why one failed, or say what a single attempt cost.

## Problem

Today an attempt is an ephemeral cycle: `CycleRequest`/`CycleResult` (`src/ports/types.rs:672`/`:696`) whose `Vec<TurnStep>` (`:903`) is folded into the `AgentReply` event and discarded as a unit. No run id, no status, no list endpoint.

Consequences that show up in the product:

- **Attempts are invisible.** A task that failed twice and succeeded on the third try looks identical to one that succeeded immediately.
- **Cost is unattributable.** Inference spend lands on `UNATTRIBUTED_AGENT = "company"` (`src/metering/inference.rs`) with no per-attempt breakdown.
- **Artifacts hang off Task only.** `ArtifactRecord` (`src/ports/artifacts.rs:103`) has `task_id` but no `run_id`, and its optional `step_seq` is never read.
- **The console is built for data that does not exist.** Task detail already renders a grouped timeline from `TimelineEntry {seq, atMillis, kind, label, detail}` (`frontend/src/api/tasks.ts:51-70`) with a lineage rail — designed for exactly this, with no run concept to hang it on.

## Solution

1. **`RunRecord` + `RunStatus`** in a new `src/ports/runs.rs`: run id, company, task, `attempt: u32`, status, `trigger_event_seq`, started/finished timestamps, error, token and USD totals, step count. `RunStatus = Pending | Running | WaitingApproval | Succeeded | Failed | Cancelled` (`Cancelled` reserved; no endpoint sets it in v1). Trace steps are a **separate** `RunStepRecord { run_id, step_seq, at_millis, step }` using a run-scoped dense sequence, deliberately **not** `EventSeq`.
2. **`RunStore` port** following the `TaskStore` pattern, with a transition API rather than free-form updates — `create_run` (Pending), `begin_run` (Pending→Running), `finish_run(outcome)` (terminal status + error + usage + timestamp), `append_run_step`, `get_run`, `list_runs(filter)`, `list_run_steps`, `list_stale_active`. Backends stay dumb and illegal transitions become impossible.
[...]

PR #348, opened 2026-08-04
feat(tasks): give the per-task Discussion tab a backend (#335)

### LINKED ISSUE #335: tasks: the per-task Discussion tab has no backend
## Summary

The Task Detail page ships a **Discussion** tab that is permanently empty. It renders an honest empty state — *"No discussion yet — a per-task discussion thread has no backend yet"* — and there is a code comment saying the same.

So this is not a bug. It is a shipped stub with no tracking issue, which is the reason to file it: right now the only record that it is unfinished lives in a comment.

Raised by QA on smoke1, who correctly read it as a hole rather than a feature.

## Why it matters

A task is where a piece of work is discussed, and today every conversation about a card has to happen somewhere else — which means the card stops being the record of its own work. It also undercuts the tab bar: three tabs (Approvals, Artifacts, Discussion) and one of them can never contain anything.

## Design questions to settle before building

- **Is a task discussion a distinct channel, or a view onto the existing company chat filtered to a task?** The second is much cheaper and keeps one message store; the first gives a cleaner thread. This choice drives everything else.
- **Do agents participate, or is it operator-only?** If agents post, this becomes a delegation surface and needs to interact with the harness rather than just persist rows.
- **How does it relate to the timeline the other tabs read?** A discussion entry and a timeline entry are both "something happened on this task", and two parallel notions of that will drift.
- Ordering, editing, deletion, and whether a message can reference an artifact or an approval.

Worth deciding alongside **#242** (first-class runs) — if a task gains a real run object, a discussion anchored to the task versus to a run are different products.

## Acceptance

- [ ] An operator can post to a task and see it persist across reload and across browsers
- [ ] The empty state remains honest for a task with no discussion
- [ ] Whatever is chosen above is written down in the module doc, not just implied by the schema

## Related

- #242 — first-class run records
- #184 — Task Detail screen epic
- #183 — MVP spec

M3gA-Mind:
## Summary

Closes #335 — the Task Detail **Discussion** tab shipped as an honest stub ("a per-task discussion thread has no backend yet"). This gives it one.

**The design decisions the issue asked to settle, and what was chosen** (all written down in `docs/modules/server/README.md`, per the third acceptance box, not just implied by the schema):

- **A task discussion is its own thread, not company chat filtered to a card.** Chat is addressed to a *desk* and is a conversation with the company; a discussion is addressed to a *card* and is a note about one piece of work. Filtering one into the other makes every thread a slice of a stream written for a different audience, and the card stops being the record of its own work the moment somebody replies in the desk instead.
- **It is not a second message store, either.** A post is journaled as `CompanyEvent::TaskDiscussionPosted`, in the same `EventLog` the timeline is folded from, and `GET …/tasks/{task_id}` serves both out of **one** traversal (`fold_task_journal`). That is what keeps the two notions of "something happened on this task" from drifting — same log, read two ways. They stay two arrays rather than one merged list because the timeline is what the *company* did on the card and the discussion is what *people* said about it; merged, an operator's aside would sit between a dispatch and its completion and read as part of the run.
- **Operator-only in v1; agents do not participate.** Posting journals a message and nothing else: no cycle, no dispatch, no spend. The projections enforce that rather than merely documenting it — `wire_event` (inference sidecar) and `summarize_event` (orchestrator insight) both name the *card* and deliberately omit the *text*, so the tab cannot become an unannounced prompt surface, and the deny-by-default SSE projection drops the variant. Agent participation is a decision to take with first-class runs (#242): a discussion anchored to a task and one anchored to a run are different products.
- **Ordering / editing / deletion / references.** Oldest-first by journal sequence (also the console's render key). Append-only, so no edit and no delete in v1 — a retraction would need a tombstone event and is not one. Plain text: no artifact or approval references yet.
- **Attribution.** The signed-in user rides on `by` and resolves to a roster label on read (display name, or an email's local part); a departed user reads as `someone`, a machine credential as `operator`. A user id or email address never reaches the wire — a thread is read by every member of the company.

**Wire surface**

```
POST …/tasks/{taskId}/discussion   {"text": "blocked on the API key"}   -> 201 {seq, atMillis, author, text}
GET  …/tasks/{taskId}                                                   -> {task, timeline, discussion[], lineage, waitingSince?}
```

`400` on empty/whitespace-only text (there is no delete, so a blank row would be permanent noise), `404` on an unknown card, and over-long text is truncated to `MAX_DISCUSSION_CHARS` (4000, codepoint-safe) rather than refused.

**Console.** `TaskDetailView`'s stub becomes a real thread plus a composer (Enter posts, Shift+Enter for a newline). It reads off the parent's single `GET …/tasks/{id}` and therefore rides its existing 4s visibility-gated poll, which is what makes another operator's post appear without a reload. The empty state stays honest for a card nobody has posted on.

**Acceptance**

- [x] An operator can post to a task and see it persist across reload and across browsers — the message lives in the company journal, not the posting session; the tests assert it through a second, independent request rather than off the POST's echo.
- [x] The empty state remains honest for a task with no discussion.
- [x] The choices above are written down in the module doc.

## API Or Behavior Changes

- **New route** `POST …/tasks/{taskId}/discussion` under both scope forms.
- **`GET …/tasks/{taskId}` gains `discussion[]`** — additive; every existing field is unchanged.
- **New `CompanyEvent::TaskDiscussionPosted`** — additive on the wire (`by` is `skip_serializing_if`), so every persisted log loads unchanged and no stored record needs migrating.
- **`ScopedCompany` now carries `actor: Option<Actor>`** — the principal the extractor already resolved, so a route that journals an operator action can attribute it instead of landing an anonymous `by: None`. No existing handler reads it, so no behavior changes elsewhere.

## Tests

Added with the change (CI runs them):

- `task_discussion_posts_persist_and_are_scoped_to_their_card` — post → re-read; oldest-first order and strictly increasing `seq`; another card's message never leaks in; a card with no posts reads back an empty thread; both scope forms serve the same thread; and a post never lands on the run timeline.
- `task_discussion_rejects_an_empty_message_and_an_unknown_card` — empty and whitespace-only are `400`, an unknown card is `404`, an over-long paste is accepted and capped, and the refusals journal nothing.
- `cap_discussion_is_codepoint_safe` — truncation on a character boundary, never a byte one.
- `task_discussion_posted_round_trips_and_omits_an_absent_actor` — the additive wire contract.

Intentionally untested: the console half. This repo has no unit-test runner for the frontend, and CI does not execute the Playwright suite (it needs a live host, #281) — the tab is covered by `typecheck` / `typecheck:e2e` / `build`.

## Documentation

- `docs/modules/server/README.md` — new "The per-task discussion (issue #335)" section recording every decision above, plus the `tasks` route row brought up to date (it still listed only `POST`/`PATCH`/`DELETE`).
- `docs/spec/runtime/api.md` — the new route in the console write-plane list.
- `docs/spec/runtime/ports.md` — `TaskDiscussionPosted` in the `CompanyEvent` catalogue.


M3gA-Mind:
@coderabbitai review

oxoxDev (changes requested):
The journal-as-one-store call is sound and it's actually written down in the module doc — that acceptance box is genuinely met, not hand-waved. Two things stop the change being additive, both cheap to fix.

**Majors**

- `major` `src/harness/orchestrator.rs:553` — `QueryCompanyTool` surfaces the **last 10** journal events (`RECENT_EVENTS`, line 77; read at 357–368). A discussion is a user-driven, high-frequency writer into that same log, so ten messages on one card evict every dispatch, reply and approval from the orchestrator's activity view and replace them with ten lines it can't act on. "Agents do not participate" holds for the *text*, not for the *slot*. Filter the variant out of the `recent` tail, or fold it to one "N discussion posts" line, instead of a row per post.
- `major` `src/server/ops/tasks.rs:593` — the thread comes back whole and uncapped on every 4s poll of an open detail screen, forever. `MAX_DISCUSSION_CHARS`' own doc concedes the thread is "read back in full on every detail poll"; at 4000 chars a post that is a large response on a busy card, re-sent every four seconds per browser. `chat_history::history_for_desk` already has the shape to copy — a `first` tail cap plus a `before_seq` cursor.

**On the design call, and what it costs**

- The fold itself is not new: `GET /tasks/{id}` already paged the whole journal per read before this PR, so the discussion rides an existing cost rather than introducing one. Fair.
- The prior art cuts the other way on *writes*, though. `src/ports/runs.rs:12` — landed in #342, after this branch was cut — says plainly that "folding the whole journal per read is what makes the existing workflow-run list expensive", and #242 answered that by giving runs their own store rather than more events. This PR points a chattier writer at the same log, and there is no retention or compaction for it anywhere in the tree. Worth one honest sentence in the module doc; the current text presents the shared log purely as a consistency win.
- Append-only for human prose is defensible and is documented, but there is no redaction path at all — no edit, no delete, no tombstone, and the log is what export/import ships. The PR's own fixture text ("blocked on the API key") is exactly the shape of message where somebody pastes the key itself. Please open the tombstone follow-up as an issue rather than leaving it a paragraph in the doc.

## Against the issue

- **Acceptance 1 — "post persists across reload and across browsers"**: met for reload; `task_discussion_posts_persist_and_are_scoped_to_their_card` asserts through a second independent `GET`, which is the right proof shape. "Across browsers" holds by construction (journal, not session state) but no test posts as a *second* principal — `nit` `src/server/ops/write_test.rs:3289`, every post is `harness-admin`, so the `someone` (departed user) and `operator` (machine credential) branches of `into_message` are unexercised.
- **Acceptance 2 — honest empty state**: met. `t-quiet` asserts an empty array server-side, and `frontend/src/views/TaskDetailView.tsx:1095` renders `EmptyState` on `messages.length === 0` with copy that no longer claims there is no backend.
- **Acceptance 3 — decisions written down**: met, in `docs/modules/server/README.md`.
- **Design questions**: all four #335 asked are actually asked and actually settled the way the PR body says — distinct thread over filtered chat, operator-only, same log two arrays, and ordering/edit/delete/references. No skipped question.
- `nit` **the #242 coupling #335 explicitly flagged**: the issue said to decide task-anchored vs run-anchored *alongside* #242, and the PR defers it. Reasonable when the branch was cut — but #342 landed the `RunStore` on main since, so runs are first-class now and the deferral is weaker than the PR body implies. One line acknowledging that would keep the doc honest.
- **Beyond scope?** No. `orchestrator.rs` and `effects.rs` both `match` `CompanyEvent` exhaustively with no `_` arm, so a new variant *forces* both arms — this is an obligation, not agent-facing plumbing the author chose to add. Agents still can't post: no tool, and nothing builds a `CycleRequest` carrying the variant. `ScopedCompany.actor` is genuinely new but small and justified. The consequence of the arms' *content* is the insight-tail major above.
- **Premise still true**: yes — `frontend/src/views/TaskDetailView.tsx:392` on main still carries the stub and its comment, and `grep -rn discussion src/` on main is empty. Nothing has closed part of #335 since filing.

**Nits**

- `nit` `src/brain/medulla/effects.rs:171` — the PR body says the projections "enforce that rather than merely documenting it". Not yet: nothing routes this variant into a cycle, and no test asserts the wired body omits `text`. The sibling exclusion just below it, `a_finished_run_wires_out_counts_without_the_recipient_or_the_transport_text`, exists precisely so a later "just include the detail" fails CI. Same pin, one assert.
- `nit` `src/server/ops/tasks.rs:940` — the `201` echo costs a second roster read, and `DiscussionTab` (`frontend/src/views/TaskDetailView.tsx:1116`) discards the response and refetches anyway. Either render the echo or drop the read; `at_millis: now_millis()` at :943 is also a different stamp from the journaled one.

Auth reads correctly: `ScopedCompany` (`src/server/ops/scope.rs:107`) enforces address + temp-password before the actor is derived, and `Admin`/`Member` are the only roles, so "any member may post" matches the other task writes.

M3gA-Mind:
Both majors are fixed, all three nits are done, the follow-up is filed as #358, and the branch is merged up to `main` (#357) so it is mergeable again.

**major — `orchestrator.rs:553`, the insight tail.** Took the fold rather than the filter. The tail now walks the log skipping `TaskDiscussionPosted` — so a post never holds one of the ten slots — and appends a single `- 20 discussion posts on task cards (text not shown)` line, counted over the span the tail covers so it still reads as *recent* activity. `discussion_posts_fold_to_one_line_instead_of_evicting_the_activity_tail` puts twenty posts between a dispatch and a completion and asserts both run events survive, that there is one folded line rather than twenty rows, and that no message text appears. The `summarize_event` arm stays (the match is exhaustive) with a comment saying it is no longer the caller and why the no-quoting rule is inherited rather than re-decided.

**major — `tasks.rs:593`, the uncapped thread on a 4s poll.** Copied `history_for_desk`'s shape: `DISCUSSION_PAGE = 50` tail cap plus a `?discussionBefore=<seq>` cursor, and `discussionHasMore` so a truncated thread says it is truncated. The fold drops out-of-window posts as it reads them, so a long thread is traversed but never resident — the same "bounded per page, not per request" discipline `TIMELINE_PAGE` already had. The timeline is deliberately *not* paged: it is bounded by what the company did on one card. Test asserts the newest page comes first, the cursor is exclusive (no message served twice, none dropped between pages), and `discussionHasMore` is false at the start of the thread and on a short one.

Capping surfaced a console bug worth naming: the page *slides*, so one new post would push the oldest visible message out from under a reader mid-thread. `DiscussionTab` now accumulates what it has been shown, deduped by `seq` and keyed per card, so the thread on screen only ever grows — and it has a "Load earlier messages" control.

**nit — `write_test.rs:3289`, one principal.** Added a test that posts through the real write path with a **tenant token**, which also pins `ScopedCompany`'s keep-the-person-drop-the-credential rule (a platform post that started attributing itself to something would show up as a label that is not `operator`), and journals a **departed** user id to exercise the `someone` branch, asserting the raw id never reaches the wire.

**nit — `effects.rs:171`, "enforce rather than document".** Fair — it was documented, not asserted. `a_discussion_post_wires_out_the_card_without_the_message_text` is the same pin as its `WorkflowRunFinished` sibling, one variant over.

**nit — `tasks.rs:940`, the echo.** Kept the echo and made the console *render* it, so the roster read is earned rather than discarded. The echoed row is now the journaled one, read back at its own `seq` instead of re-stamped with `now_millis()`, so it carries the same `atMillis` the next poll returns under the same key.

**#358 — the redaction follow-up**, filed on this repo referencing #335 and #348: no edit, no delete, no tombstone, and the log is what export/import ships. It sketches a tombstone superseding an earlier post (keeping the log append-only), the read fold honouring it so redaction is a property of every read surface, export/import carrying the tombstone, who may redact, and the real tension — scrub-at-rest is the only thing that helps for a leaked secret and is the part that conflicts with an immutable log.

**The doc sentence.** Now stated plainly, and it got stronger while this sat: `src/ports/runs.rs` (#342) says folding the whole journal per read is what made the run list expensive, #242 answered that by giving runs a store of their own instead of more events, and #357 has since landed the write path — so that is a conclusion the runtime has already acted on, not a future concern. The discussion goes the other way knowingly (one small event per post, no per-step fan-out, paged read), but it is a shared log with an extra writer on it and no retention or compaction anywhere in the tree. The #242 deferral paragraph now says the same thing about itself: with runs first-class on `main` the question is answerable, so the deferral is on scope, not on missing prior art.

**Merge.** One conflict, in `fold_page`: #357 gave `TaskDispatched` a `run_id` and moved the window anchor behind a `&mut`, where this branch had already folded that state into `TaskFold`. Kept `TaskFold` and took the widened pattern — the timeline projection anchors on the *instant*, not the attempt, which is now a comment so it does not get re-added.

@coderabbitai review

Issue #266, opened 2026-08-02
bug: inference save claims agents use it next turn, but echo→harness needs a restart


oxoxDev:
## Summary

Saving an inference configuration toasts "Inference updated. **Agents use it on their next turn.**" That is true for some transitions and false for the one that matters most: going from no credential to a credential.

Brain selection happens once, at company-build time (`src/runtime/builder.rs` ~:917, resolving the effective inference config at ~:945). A company built with no inference source gets the offline echo brain, and `wf_runner` is left `None` (~:835) — it is only populated inside the harness arm. Storing a credential afterwards updates the secret store but does not rebuild the company, so:

- agents keep echoing
- the workflow runner stays unwired, so `POST /workflows/{id}/run` keeps returning `not_wired`
- the scheduler keeps logging `N scheduled workflow(s) will not fire — no workflow runner is wired`

Only a restart picks it up. The per-tenant provider genuinely does re-resolve each turn — so changing a *model* or swapping a key while the harness is already active is live, and the toast is accurate there. The inaccurate case is the echo → harness transition, which is exactly the first thing a new operator does.

Observed on a local build (`openhuman,mcp,telegram`): key stored, toast shown, agents continued echoing and the scheduler kept warning, until the process was restarted.

## Reproduction

1. Boot a company with no inference configured. Console shows `cognition: "echo"`.
2. Connections → Inference → choose `openai_compatible`, set a base URL and key. Save. Toast says agents use it next turn.
3. Send an agent a message. It still echoes.
4. `POST …/workflows/{id}/run` → `{"error":"workflow execution is not wired in this deployment"}`.
5. Restart the server. Both now work.

## Expected behavior

Either the transition is live, or the console says what is actually required.

- **Preferred:** rebuild the company runtime when inference goes from unresolved to resolved, so the promise the toast makes is kept. This is the same seam a hosted tenant needs anyway — an operator setting up a fresh tenant should not have to get the pod restarted.
- **Minimum:** make the message conditional. When the prior state had no credential, say that the change takes effect after a restart rather than on the next turn — and surface it in the status panel too, not only in a toast that disappears.

The status route already knows enough to tell the difference: it reports `cognition` and `keyConfigured`, so a stored credential sitting next to `cognition: "echo"` is precisely the detectable "restart pending" state.

Issue #208, opened 2026-07-31
Boot rebuild overwrites the persisted record's manifest with the seed company.toml


oxoxDev:
## Problem

Every runtime mutation of a company's manifest is silently lost on the next boot. The rebuild path reads the existing record for lifecycle, ledger, overlays, and provenance, then saves the record back with `manifest: self.manifest.clone()` — the manifest freshly parsed from the seed `companies/<name>/company.toml`, not the one that was persisted.

`src/runtime/builder.rs:1121`

The comment directly above it (lines ~1115-1119) says "a rebuild never rewrites the version-controlled manifest". That is true of the file on disk and false of the record: the record's manifest is overwritten wholesale on every launch.

## Evidence

The store persists the manifest correctly — `FsCompanyStore::save` serializes it to the bundle's `company.toml` (`src/store/fs.rs:248-250`). The loss happens after, at rebuild:

```
save/load round-trip:   ["greeter"]     <- store persists the runtime-added id
after boot rebuild:     []              <- overwritten by the seed manifest
```

## Impact

`[workflows].enabled` is the runtime-mutated field found so far. Workflow creation pushes the new id onto it and saves; after a restart the id is gone. This is visible today as `Company.workflows.enabled` (GraphQL) reading `true` immediately after creation and `false` after a restart.

Nothing currently consults `enabled` to decide whether a workflow may run — the REST route, the orchestrator tool, and the `sub_workflow` resolver all load a graph by id from the seed ∪ overlay union — so no feature is broken by it right now. It becomes load-bearing the moment something enumerates enabled workflows, which is exactly what the scheduler proposed in #169 would do.

## Why it needs a design decision, not a one-line fix

Making the persisted record's manifest win at boot would mean an operator editing `company.toml` and restarting gets silently ignored — arguably worse than the current behavior. The fix has to answer, per field, whether the seed file or the persisted record is authoritative, including the fields where seed-wins is the desired behavior.

Suggested direction: an explicit merge at rebuild that preserves the known runtime-mutable fields (starting with `[workflows].enabled`) and lets the seed win everywhere else, plus a test that a runtime manifest mutation survives a rebuild.

## Found via

PR #207 (#168) — workflow graph bodies now persist as a record overlay, which made the enabled-id loss observable. That PR does not depend on this being fixed; its read paths are union-driven.

PR #353, opened 2026-08-04
feat(runtime): rebuild a company runtime in place when inference resolves (#290)

### LINKED ISSUE #290: Rebuild the company runtime when inference goes unresolved to resolved, so first-time BYOK setup needs no restart
## Summary

#266 asked for one of two things: rebuild the company runtime when inference goes
unresolved → resolved, or say honestly that a restart is required. #266 shipped the
honest-message half (`restartRequired` on the inference status, a persistent console
banner, a `restart_required` response on `POST …/workflows/{id}/run`). This issue is
the half it did not ship, plus what a rebuild actually costs.

A rebuild is still the better answer, for the reason #266 gave: it is the same seam a
hosted tenant needs. An operator setting up a fresh tenant should not have to get the
pod restarted, and today there is no way to give them that — `POST /api/v1/companies`
refuses an existing id, archive removes a company with no unarchive, and nothing else
in the process ever replaces a registered runtime.

## What a rebuild costs today

Rebuilding means calling `RuntimeBuilder::build()` again for a registered company and
swapping the result into `CompanyRegistry`. That is not the hard part — every builder
input is recoverable from `runtime.source_dir()` + `AppState` + the process env, with
one exception (below). The hard part is that the old runtime does not go away, and
two live `CompanyRuntime`s for one company id are not safe.

**1. The journal forbids a second writer, and the failure is permanent.**
`src/runtime/journal.rs` documents that exactly one process may write a journal file:
`append` emits the record and its newline as two separate writes under a *per-instance*
lock. A rebuild constructs a second `RuntimeJournal` over the same path. While any
holder of the old runtime is still alive, an interleave leaves two records on one line,
which fails to parse on replay — bricking the *next* boot and `CompanyRuntime::recover`.

**2. At-most-once effect execution breaks.** `is_executed` reads a per-instance
in-memory set populated at `load()`. Runtime A commits an effect key; runtime B never
sees it, so the effect — a send, a spend — runs twice.

**3. The two mutual-exclusion locks are per-instance.** `serial` guards a whole cycle
(an LLM turn) plus five desk read-modify-write routes; `task_writes` guards a
documented parent-cycle invariant across the three board-write handlers. Two runtimes
means two locks, so both invariants lapse across the boundary, against a company store
whose `save` writes the whole record.

**4. Nothing hands the old runtime's work over — there is no teardown seam.**
`CompanyScheduler`, `MailboxPoller` and `TelegramPoller` each snapshot an
`Arc<CompanyRuntime>` at boot and never re-read the registry. A registry swap leaves
all three driving the old echo brain forever — including the cron scheduler, which is
one of the surfaces #266 was reported against. `WorkflowScheduler` already does this
correctly (it re-reads the registry every tick) and is the model to copy.

**5. In-flight work is genuinely torn out.** `upsert_task` clones the Arc into a
spawned cycle that lives for a whole agent turn. `WorkflowScheduler` clones the runner
and event log out of the then-current runtime before spawning. The orchestrator's
runner handle is a `Weak` whose only strong reference is the runtime, so dropping the
old runtime makes its in-flight `run_workflow` tool silently no-op.

**6. `build()` is not side-effect-free, so a rebuild re-fires them.** It re-runs the
going-public path (network calls to tiny.place), and re-spawns the MCP boot, which
re-dials every installed server into a *process-global* connection map keyed by server
id — replacing connections the old runtime's agents are mid-call on. The feedback
filer's rate limiter is in-memory, so a rebuild loop is a rate-limit bypass.

**7. On the default filesystem path the stores are per-instance too.** `FsEventLog`
derives an event's `seq` from a line count under a per-instance lock, so two instances
mint duplicate sequence numbers. An open console SSE connection is subscribed to the
old broadcast sender and goes permanently deaf.

**8. One builder input is genuinely unreachable.** `serve --discoverable` exists only
in the `serve` stack frame; it mutates the manifest before the build and is retained
nowhere. Re-reading the manifest from `source_dir` would silently drop it.

## What this issue is actually asking for

Not "call build() twice". A runtime-lifecycle seam:

- [ ] **Quiesce**: a way to stop a runtime accepting new cycles and drain the one in
      flight, so the swap happens at a point with no live turn.
- [ ] **Hand over, don't reconstruct**: the journal, approval gate, event log, harness
      pool and the two serialising mutexes must be *moved* to the new runtime, not
      rebuilt. (Reusing `runtime.harness()` also preserves per-agent conversation
      history, which a fresh pool would drop.)
- [ ] **Make the three pollers registry-driven** the way `WorkflowScheduler` is, so a
      swap actually reaches cron, mail and Telegram.
- [ ] **Make `build()` idempotent on rebuild**: skip going-public and MCP re-boot when
      rebuilding an already-registered company.
- [ ] **Stash the boot-only inputs** (`--discoverable`, and anything else added later)
      somewhere a rebuild can reach.
- [ ] Then, and only then, trigger it from `PUT …/inference` when
      `restartRequired` would otherwise become true, and drop the banner.

## Why it is worth doing anyway

The hosted control plane has no "restart this tenant" button an operator can press,
and the tenant container is the unit of restart. Until this exists, first-time BYOK
setup on a hosted tenant requires an operator with pod access. That is the actual
product gap behind #266; the banner only makes it visible.

## Related

- #266 — the honest-message half, and where the console reports `restartRequired`

M3gA-Mind:
## Summary

Fixes #290. Builds the half #266 did not ship: rebuild the company runtime when inference goes unresolved → resolved, so first-time BYOK setup needs no restart.

The restart is the actual product gap. A hosted tenant's unit of restart is its container, and the control plane has no button an operator can press: `POST /api/v1/companies` refuses an existing id, archive removes a company with no unarchive, and nothing in the process ever replaced a registered runtime. So this is not "call `build()` twice" — it is the runtime-lifecycle seam the issue asked for.

### 1. Quiesce (`CompanyRuntime::quiesce`)

Sets a flag that makes every cycle entry point (`run_cycle`, `resolve_approval`, `resolve_approval_amended`) refuse with a new `OpenCompanyError::Quiescing` → **`503`**, then waits on the per-company `serial` lock, which a cycle holds for a whole turn. Both halves are load-bearing: the flag alone leaves a turn mid-cycle; the wait alone races a cycle already queued behind the one in flight.

The check on `resolve_approval*` is deliberately at the `CompanyRuntime` entry point rather than inside the follow-up cycle: a resolution that journaled the verdict and then failed to run its follow-up would leave the brain permanently unaware of an approval the operator had granted.

`Quiescing` is distinct from `LifecycleConflict` (`409`) on purpose. That one is a durable state an operator chose; this one clears itself within a turn, so it is a retry-me.

### 2. Hand over, don't reconstruct (`src/runtime/handover.rs`)

`RuntimeHandover` moves the per-instance state a second runtime must not duplicate. Each of these is a correctness matter, not an optimisation:

| Piece | Why a second copy breaks |
|---|---|
| `RuntimeJournal` | Single-writer. `append` writes a record and its newline separately under a per-instance lock, so two journals on one path interleave onto a single line and fail to parse on replay, bricking the **next** boot |
| At-most-once effect keys | In memory, populated at `load()`. A second instance never sees the first's commits, so a send or a spend runs twice |
| `serial` / `task_writes` | Per-instance mutexes. Both changed to `Arc<Mutex<_>>` so the successor adopts them; two of either and both invariants lapse at the swap |
| `FsEventLog` | Derives `seq` from a line count under its own lock, and its broadcast sender is what an open console SSE stream is subscribed to |
| `FeedbackFiler` | In-memory rate limiter; a fresh one makes a rebuild loop a rate-limit bypass |
| `HarnessPool` | Each agent's conversation history. Also injected back into `self.harness` **before** brain selection, so the successor's brain and its runtime share one pool rather than the brain getting a freshly minted one |
| `McpRuntime` | Dials a process-global connect map keyed by server id; re-booting replaces connections other agents may be mid-call on |
| `OpsStores` (incl. the `RunStore`) | Carried whole (#242), so the dispatch choke point, the successor's `HarnessBrain` (via `with_runs`) and the suppressed reaper all address one set of attempt rows. A second store strands every live run and restarts attempt ordinals at 1 |
| `ManifestApprovalGate` + `GrantSet` | Parked approvals and unredeemed grants. Rehydrating a fresh copy from the journal resurrects what the live one has already resolved |

Deliberately **not** carried over: the brain, tools, channels, workflow runner and economy (replacing those is the point), and the in-flight steer registry — the successor's harness deps mint their own, which is sound *only* because the swap happens after the drain, so the outgoing registry is empty at that point.

### 3. `build()` is idempotent on rebuild

A handover's presence is the "this is a rebuild" signal, suppressing every boot-only side effect: journal `load()`, approval rehydration, grant rehydration, **orphan-run reaping**, workspace seeding, going-public, and the MCP re-boot.

Orphan-run reaping is the subtle one, and the rebase onto #242 PR-2 is what made it worth stating exactly. Its own doc comment rests the argument on "nothing from this process can be in flight yet" — true at boot, false the moment a company has been serving.

The exposure is narrower than "a rebuild would kill live runs", and stating it loosely hides where the real edge is:

* **No `Running` row survives the drain.** `quiesce` waits on `serial`, and *both* `begin_run` and the terminality backstop sit inside that lock.
* **`Pending` rows do survive.** The dispatch choke point mints its row **outside** `serial` (in `upsert_task`, before the spawn), and board writes are not gated on the quiesce.
* **Reaping one of those is the actual hazard.** It stamps the wrong reason on it, and if the rebuild then *fails* and `resume()` puts the company back to work, that row is already terminal — its cycle's `begin_run` is rejected by the transition check and a genuinely live attempt runs with no record at all.

So it stays suppressed, and deliberately **not** on the strength of the drain: leaning on the drain would rest correctness on the current call order in `rebuild_company` rather than on the invariant `reap_orphaned_runs` states. That function's doc comment now says the proof is about boot and only boot, so a future second caller has to make the argument again rather than inherit it.

### 4. The three pollers are registry-driven

`CompanyScheduler`, `MailboxPoller` and `TelegramPoller` each snapshotted an `Arc<CompanyRuntime>` at boot and would have driven the replaced (and now quiesced) runtime forever. Each gained a `following(registry)` opt-in that the boot path uses, copying `WorkflowScheduler`. The snapshot stays as the fallback, so every existing caller and test is unaffected — this is an opt-in, not a signature change.

Cron matters most: "scheduled workflows never fire" is one of the two surfaces #266 was reported against, so a rebuild that did not reach the scheduler would look fixed and not be. Covered by a test.

### 5. Boot-only inputs are stashed

`AppState::set_boot_inputs` records `BootInputs { source_dir, discoverable }` at registration. `--discoverable` is the case that forces this to exist: it lives only in the `serve` stack frame and mutates the manifest before the build, so a rebuild re-reading `company.toml` would silently un-publish a public company.

The manifest itself comes from the **persisted record**, not a fresh `company.toml` read, so console-created workflows merged into `[workflows].enabled` (#208) and the `[place]` fields survive — and a platform-provisioned tenant with no `company.toml` is rebuildable at all.

### 6. Triggered from `PUT …/inference`

When `restartRequired` would otherwise become true, the route rebuilds and re-reads the status **from the successor**, so it can never claim a success the runtime cannot back up. A host that wired no rebuilder, or a rebuild that fails, falls through to the existing `RESTART_NOTE` — the save landed and the company is still on its old brain, which is exactly what that note describes.

`RuntimeRebuilder` is a trait implemented by the binary, because the builder inputs (harness pool, OpenHuman RPC, managed media/search backends, the injected per-tenant mailbox) are assembled there. `register_company`'s assembly is extracted into `company_builder` and shared, so a rebuild cannot drift from boot. `AppState` is passed *into* `rebuild`, not captured, so there is no `AppState` ↔ rebuilder `Arc` cycle.

## What happens to in-flight work

**The cycle in flight completes.** It is not cancelled and its effects are not replayed: the executed-key set travels with the journal, so an effect the outgoing runtime committed stays committed for the successor. Anything the outgoing runtime spawned detached (a task dispatch, a scheduled workflow run) goes through `run_cycle` and therefore through `serial`, so it is either already finished or it *is* the turn being drained.

**Cycles arriving during the window are refused (`503`), not queued.** A queued cycle would acquire the shared lock after the swap and run on the brain the rebuild was replacing. A caller retries against the successor instead.

**The journal and approval queue are continuous across the swap.** Parked approvals keep their ids, parked effects and TTLs; resolving one after the swap runs its follow-up cycle on the *new* brain. Unredeemed single-use grants survive too, so an operator who approved a blocked tool call moments before the rebuild is not asked again.

**Attempt rows (#242) get the same answer, extended.** The attempt *in flight* is the cycle in flight, so it completes and settles itself on the outgoing runtime, against the store the successor adopts; its trace keeps writing through the drain, because the sink holds an `Arc<dyn RunStore>` from the inherited `OpsStores` rather than a handle to the runtime.

The case that needed a code change is **the attempt that never starts**, and it is a real defect this rebase surfaced. Board writes are deliberately not gated on the quiesce — only cycles are — so a card dragged into `in_progress` during the window still mints its `Pending` row at the choke point. Its cycle is then refused by `ensure_accepting`, which runs *before* `CycleRunner` takes the serial lock and therefore before `begin_run` — putting it out of reach of the terminality backstop, which only settles rows that cycle started. Every other dispatch failure is already covered in there; this is the one that escapes. And the rebuild has just switched off the one sweep that would have caught it. Net: a row claiming to be pending for the rest of the process's life, on a card that reads as under way by an attempt that never began, which nothing re-drives. Pre-#357 that window merely dropped the dispatch with a `warn`; #357 makes it leave a durable lie.

The refused dispatch now settles its own row: `Pending` → `Failed` carrying a new `ports::runs::RUNTIME_REPLACED_ERROR`, kept distinct from `ORPHAN_ERROR` so a run list can tell *"we swapped your runtime, retry"* apart from *"the host died"*. `Pending` → terminal is already the legal move the status table names for **"a dispatch that failed before the first turn"**, so this needs no new state — only a caller willing to use it. `Failed` rather than `Cancelled` for the same reason the reaper picks `Failed`: the runtime went away underneath a card an operator had just dispatched, and that wants to be visible, not filed away as an intentional stop. The card stays in `in_progress`, which is where every other failed dispatch cycle leaves it.

**The window bounds the triggering request.** `PUT …/inference` blocks until the in-flight turn drains. That is a deliberate trade, stated in `docs/spec/runtime/rebuild.md`: swapping mid-cycle is precisely the two-live-runtimes hazard the handover exists to avoid, and a bounded wait that gave up and swapped anyway would trade a slow response for a corrupted journal. The transition this exists for (first-time setup on an echo-brain company) has no expensive turn to wait on.

**A failed rebuild resumes the company it quiesced.** A company left quiesced would refuse every cycle forever, which is strictly worse than the stale brain the rebuild was replacing. Covered by a test.

## API Or Behavior Changes

- **Behavior:** `PUT …/inference` on a company whose config was stranded behind a boot decision now rebuilds that company's runtime in place instead of asking for a restart, and returns a note saying so. A host with no rebuilder wired, or a failed rebuild, keeps the previous response verbatim.
- **Behavior:** a cycle attempted against a quiescing/replaced runtime returns `503` with code `quiescing`. Previously no runtime was ever replaced, so this state did not exist.
- **Behavior:** `serve` now runs its cron scheduler, mailbox poller and Telegram poller off the registry rather than a boot snapshot. Identical unless a swap happens.
- **API (new):** `ports::runs::RUNTIME_REPLACED_ERROR`; `OpenCompanyError::Quiescing`; `CompanyRuntime::{quiesce, resume, is_quiesced, handover, adopt_locks, journal}`; `runtime::{RuntimeHandover, BootInputs, RebuildRequest, RuntimeRebuilder, rebuild_company}`; `RuntimeBuilder::with_handover`; `AppState::{with_rebuilder, rebuilder, set_boot_inputs, boot_inputs}`; `CompanyScheduler/MailboxPoller/TelegramPoller::following`.
- **API (changed):** `CompanyRuntime`'s `serial` and `task_writes` are `Arc<Mutex<()>>` rather than owned. Both are `pub(crate)` and every call site is `.lock().await`, which is unchanged by the `Arc`.
- **No frontend change.** The `restartRequired` banner in `InferenceSection.tsx` is retained on purpose: after this it is only reachable when a rebuild is genuinely unavailable, which is the situation it was written for, and it self-clears because the PUT response and the GET both read the successor. Deleting it would re-introduce a silent failure on exactly the hosts that still need a restart.

## Tests

- [x] `cargo fmt --all -- --check` — N/A - full local build matrix is prohibited on this machine; verified in CI (ran `rustfmt --edition 2024` on every touched Rust file; clean)

Tests added with the behavior change (CI runs them):

`src/runtime/rebuild.rs` — `a_rebuild_swaps_the_registered_runtime_and_hands_state_over` (asserts pointer identity of the journal, approval gate, store, event log and **both** mutexes across the swap), `the_successor_runs_cycles_the_outgoing_runtime_now_refuses` (the replaced runtime stays quiesced and answers `quiescing`; the successor is live), `a_failed_rebuild_resumes_the_company_it_quiesced`, `a_host_with_no_rebuilder_says_so_instead_of_quiescing` (and proves the check runs *before* the quiesce), `rebuilding_an_unregistered_company_is_a_not_found`.

`src/company/runtime.rs` — `a_dispatch_refused_by_a_quiescing_runtime_settles_its_attempt`, with a positive control: the same call on a **live** runtime goes down the ordinary path (the cycle runs, the backstop settles the row), so the assertion below it is specifically about the quiesce refusal. Then quiesce, re-dispatch, and assert the row is `Failed` / `RUNTIME_REPLACED_ERROR` with no start time and a finish time.

`src/runtime/scheduler.rs` — `following_the_registry_reaches_a_rebuilt_runtime` (drives a real quiesce → handover → swap and asserts the scheduled cycle lands on the successor), `without_following_a_scheduler_drives_its_snapshot` (the opt-in is real, and the refusal surfaces rather than being swallowed).

`src/server/ops/inference.rs` — `configuring_inference_after_boot_rebuilds_instead_of_asking_for_a_restart`: with a rebuilder wired, the save clears `restartRequired`, reports the successor's cognition, swaps the registry, leaves the outgoing runtime quiesced and the successor accepting, and a plain `GET` agrees. The pre-existing `configuring_inference_after_boot_reports_restart_required` is untouched and now also pins the no-rebuilder fallback.

The inference test's rebuilder runs the **real** builder over the **real** handover and injects a brain that reports the harness cognition path, rather than constructing a live `HarnessBrain` (which would resolve MCP servers, Composio and an agent roster inside a unit test). Only the cognition path is stubbed; the swap, the handover and the route logic are the real thing.

## Documentation

- `docs/spec/runtime/rebuild.md` (new) — the full sequence, the inherit-vs-reconstruct table with the reason each piece cannot be duplicated, the suppressed boot-only side effects, what happens to in-flight work, the registry-driven pollers, and the wiring.
- `docs/spec/runtime/README.md` — indexed.
- `docs/spec/runtime/manifest.md` — the `[inference]` section's restart paragraph now points at the rebuild and explains why `restartRequired` stays on the read shape.
- Module docs on `src/runtime/handover.rs` and `src/runtime/rebuild.rs` carry the rationale; `build()`'s existing boot-replay and orphan-reaping comments are extended in place rather than duplicated.

## Checklist against the issue

- [x] **Quiesce**: stop accepting new cycles and drain the one in flight
- [x] **Hand over, don't reconstruct**: journal, approval gate, event log, harness pool and the two serialising mutexes are moved
- [x] **Make the three pollers registry-driven**, the way `WorkflowScheduler` is
- [x] **Make `build()` idempotent on rebuild**: going-public and MCP re-boot skipped (plus journal replay, rehydration, orphan-run reaping, workspace seeding)
- [x] **Stash the boot-only inputs** (`--discoverable`, source dir) where a rebuild can reach them
- [x] **Trigger it from `PUT …/inference`** when `restartRequired` would otherwise become true. The banner is retained rather than dropped, for the reason given under *API Or Behavior Changes*.

### DIFF
--- a/docs/spec/runtime/README.md
+++ b/docs/spec/runtime/README.md
@@ -11,6 +11,8 @@ Supporting docs:
 - [ports.md](ports.md) — the port trait contracts (normative)
 - [manifest.md](manifest.md) — `company.toml` schema
 - [lifecycle.md](lifecycle.md) — company state machine and durability
+- [rebuild.md](rebuild.md) — replacing a registered runtime in place (quiesce →
+  hand over → swap), so a first-time inference config needs no restart
 - [api.md](api.md) — HTTP routes and auth
 - [config.md](config.md) — configuration and the one-key story
 - [users.md](users.md) — human collaborators: magic-link/password sign-in,
--- a/docs/spec/runtime/manifest.md
+++ b/docs/spec/runtime/manifest.md
@@ -156,6 +156,14 @@ prompt = "Weekly review and operator digest"
   neither. The status route reports that state as `restartRequired` — a resolved
   config next to a non-harness `cognition` — and the console says "restart"
   instead of "next turn" for it (issue #266).
+  Since issue #290 that save **rebuilds the company's runtime in place** rather
+  than asking for a restart the operator may have no way to perform: a hosted
+  tenant's unit of restart is its container, and the control plane has no button
+  for it. `PUT …/inference` quiesces the running runtime, hands its live state to
+  a successor, and swaps the registry — see
+  [runtime rebuild](rebuild.md). `restartRequired` remains on the read shape and
+  stays honest: it is still `true` on a host that wired no rebuilder, which is
+  when a restart genuinely is the only route.
   Saving `managed` from the console is a *revert* (`DELETE …/inference`) and
   carries no credential, so the console refuses that save while a key is still
   typed in the form rather than dropping it and reporting success (issue #265).
--- a/docs/spec/runtime/ports.md
+++ b/docs/spec/runtime/ports.md
@@ -657,6 +657,11 @@ at least one approval finishes `waiting_approval`, not `succeeded` — a person
 must act. A failed, cancelled or paused run keeps the reason it stopped;
 relabelling it "waiting on you" would hide that reason.
 
+**A runtime being replaced is a fifth writer.** Writer 1 mints the row before
+writer 2 can start it, so a runtime swapped in between refuses the cycle and
+settles the row itself — and the boot reaper is *not* a safe fallback mid-life.
+See [rebuild.md](rebuild.md#attempt-rows-242).
+
 #### Correlation fields elsewhere
 
 Four additive `Option<String>` fields point back at a run. All are
--- a/docs/spec/runtime/rebuild.md
+++ b/docs/spec/runtime/rebuild.md
@@ -1,0 +1,134 @@
+# Rebuilding a company runtime in place
+
+Issue #290. Which brain a company runs is chosen **once**, in
+`RuntimeBuilder::build`. A company that resolved no inference source at boot is
+on the offline echo brain with an unwired workflow runner, and a credential
+written afterwards reaches neither. #266 shipped the honest half of that: a
+`restartRequired` flag on the inference status, a persistent console banner, and
+a `restart_required` answer from `POST …/workflows/{id}/run`.
+
+The restart is the problem. A hosted tenant's unit of restart is its container,
+and the control plane has no button an operator can press: `POST
+/api/v1/companies` refuses an existing id, archive removes a company with no
+unarchive, and nothing else in the process ever replaced a registered runtime.
+So first-time BYOK setup needed someone with pod access.
+
+This is the seam that removes it.
+
+## The sequence (`src/runtime/rebuild.rs`)
+
+`runtime::rebuild_company(&state, id)`:
+
+1. **Read the manifest** from the company's persisted record — before anything
+   else, so a store failure can never leave a company parked. The record's
+   manifest is the *materialized* one, which matters for two things a fresh
+   `company.toml` read would drop: console-created workflows merged into
+   `[workflows].enabled`, and the `[place]` fields `serve --discoverable`
+   mutated before the original build.
+2. **Quiesce** (`CompanyRuntime::quiesce`). Sets a flag that makes every cycle
+   entry point refuse with `Quiescing` (`503`), then waits on the per-company
+   `serial` lock, which a cycle holds for a whole turn. Both halves are
+   necessary: the flag alone leaves a turn mid-cycle, and the wait alone races a
+   cycle queued behind the one in flight.
+3. **Hand over** (`CompanyRuntime::handover`). Snapshots the per-instance state a
+   second runtime must not duplicate — see below.
+4. **Rebuild** through the host's `RuntimeRebuilder`, which runs the *same*
+   builder assembly boot ran (`company_builder` in `src/bin/opencompany.rs`).
+5. **Swap** the successor into the registry.
+
+## What must be inherited, not rebuilt (`src/runtime/handover.rs`)
+
+Most of a runtime is `Arc<dyn Port>` handles that are safe to share. These are
+not; a second copy of any of them is a correctness bug:
+
+| Piece | Why a second copy breaks |
+|---|---|
+| `RuntimeJournal` | Single-writer. `append` writes a record and its newline separately under a per-instance lock, so two journals interleave onto one line and fail to parse on replay — bricking the *next* boot |
+| At-most-once effect keys | Held in memory, populated at `load()`. A second instance never sees the first's commits, so a send or a spend runs twice |
+| `serial` / `task_writes` | Per-instance mutexes. Two of either and both invariants lapse across the swap |
+| `FsEventLog` | Derives `seq` from a line count under its own lock, and its broadcast sender is what an open console SSE stream is subscribed to |
+| `FeedbackFiler` | In-memory rate limiter; rebuilding it makes a rebuild loop a rate-limit bypass |
+| `HarnessPool` | Holds each agent's conversation history |
+| `McpRuntime` | Dials a *process-global* connection map keyed by server id; re-booting it replaces connections other agents may be mid-call on |
+| `ManifestApprovalGate` + `GrantSet` | Parked approvals and unredeemed single-use grants. Rehydrating a fresh copy from the journal resurrects what the live one has already resolved |
+
+Deliberately **not** carried over: the brain, tools, channels, workflow runner
+and economy (replacing those is the point), and the in-flight steer registry
+(the successor's harness deps mint their own; sound only because the swap
+happens after the drain, so the outgoing registry is empty).
+
+## Boot-only side effects, suppressed on a rebuild
+
+`build()` treats the presence of a handover as "this is a rebuild" and skips:
+
+- **Journal replay** (`load()`) and **approval rehydration** — the inherited
+  journal and gate are already live.
+- **Orphan-run reaping.** At boot it is sound because nothing from this process
+  can be in flight; mid-life that premise is false, and reaping would settle
+  live run records as dead.
+- **Workspace seeding** — the company is already running.
+- **Going public.** A paid, networked handle claim. A company that is already
+  public does not become more public by claiming again.
+- **MCP boot** — see the table above.
+
+## What happens to in-flight work
+
+The cycle in flight at the quiesce **completes** on the outgoing runtime,
+against the same journal, approval queue and stores the successor adopts. It is
+not cancelled, and its effects are not replayed: the executed-key set travels
+with the journal.
+
+Cycles arriving *during* the window are refused with `503 quiescing` rather than
+queued, so a caller retries against the successor instead of silently getting
+the brain the rebuild was replacing. The window is one turn wide.
+
+That is also the bound on the triggering request: `PUT …/inference` blocks until
+the in-flight turn drains, so a save landing mid-turn waits for it. This is a
+deliberate trade. Swapping while a cycle runs is precisely the two-live-runtimes
+hazard the handover exists to avoid, and a bounded wait that gave up and swapped
+anyway would trade a slow response for a corrupted journal. In practice the
+transition this feature exists for — first-time BYOK setup on a company running
+the echo brain — has no expensive turn to wait on.
+
+Parked approvals and unredeemed grants survive untouched, with their ids, parked
+effects and TTLs. Resolving an approval after the swap runs its follow-up cycle
+on the *new* brain — which is the desired outcome, not a compromise.
+
+A rebuild that fails leaves the outgoing runtime registered and **resumes** it. A
+company stuck quiesced would refuse every cycle forever, which is strictly worse
+than the stale brain the rebuild was trying to replace.
+
+## Registry-driven pollers
+
+A registry swap only helps surfaces that read the registry. `WorkflowScheduler`
+already re-read it every tick; `CompanyScheduler`, `MailboxPoller` and
+`TelegramPoller` each snapshotted an `Arc<CompanyRuntime>` at boot and would have
+kept driving the replaced — and now quiesced — runtime forever. Each gained a
+`following(registry)` opt-in that the boot path uses; the snapshot remains the
+fallback, so existing embedders are unaffected.
+
+Cron is the one that matters most here: "scheduled workflows never fire" is one
+of the two surfaces #266 was reported against, so a rebuild that did not reach
+the scheduler would look fixed and not be.
+
+## Wiring
+
+`RuntimeRebuilder` is a trait implemented by the **binary**, because the builder
+inputs (harness pool, OpenHuman RPC transport, managed media/search backends,
+the manager-injected per-tenant mailbox) are assembled there from the process
+environment and feature flags. `AppState::with_rebuilder` installs it, and
+`AppState::set_boot_inputs` stashes what a rebuild cannot recover any other way
+(`--discoverable`, the source directory).
+
+A host that wires **no** rebuilder keeps the pre-#290 behaviour exactly: the
+status still reports `restartRequired`, and the console still says so. That is
+the honest answer when a rebuild genuinely is unavailable, so the flag and the
+banner are retained rather than deleted.
+
+## Related
+
+- #266 — the honest-message half, and where the console reports `restartRequired`
+- [manifest.md](manifest.md#inference) — `[inference]` precedence and the
+  next-turn guarantee
+- [lifecycle.md](lifecycle.md) — company lifecycle states (`paused`, `archived`),
+  which are durable and unrelated to the transient quiesce window
--- a/docs/spec/runtime/rebuild.md
+++ b/docs/spec/runtime/rebuild.md
@@ -51,6 +51,7 @@ not; a second copy of any of them is a correctness bug:
 | `HarnessPool` | Holds each agent's conversation history |
 | `McpRuntime` | Dials a *process-global* connection map keyed by server id; re-booting it replaces connections other agents may be mid-call on |
 | `ManifestApprovalGate` + `GrantSet` | Parked approvals and unredeemed single-use grants. Rehydrating a fresh copy from the journal resurrects what the live one has already resolved |
+| `OpsStores` (incl. the `RunStore`) | Carried whole, so the dispatch choke point, the successor's `HarnessBrain` and the (suppressed) reaper all address one set of attempt rows. A second store would strand every live run and restart attempt ordinals at 1 |
 
 Deliberately **not** carried over: the brain, tools, channels, workflow runner
 and economy (replacing those is the point), and the in-flight steer registry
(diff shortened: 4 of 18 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Rebased onto `main` (`eba82bb`), which now carries **#342** (#242 PR-1: the `RunStore` port + boot reaper) and **#357** (#242 PR-2: the run write path). The textual rebase was clean; the semantic collision was not, and it turned up one real defect. Answering the two questions raised on the rebase directly.

## Q1 — does `build()` re-run `reap_orphaned_runs` on a rebuild?

**No, and that was already deliberate — but the argument in the comment was weaker than it should have been, so it has been made exact.**

The suppression is `handover.is_none() && …` at `src/runtime/builder.rs`. The exposure is narrower than "a rebuild would kill live runs", and stating it loosely hides where the real edge is:

* **No `Running` row survives the drain.** `quiesce` waits on `serial`, and *both* `begin_run` and the terminality backstop sit inside that lock. So at the point `build()` runs on a rebuild, no row from this process is `Running`.
* **`Pending` rows do survive.** The dispatch choke point mints its row **outside** `serial` (in `upsert_task`, before the spawn), and board writes are not gated on the quiesce. A card dragged into `in_progress` during the window leaves a `Pending` row behind.
* **Reaping one of those is the actual hazard.** It stamps the wrong reason on it, and if the rebuild then *fails* and `resume()` puts the company back to work, that row is already terminal — its cycle's `begin_run` is rejected by the transition check and a genuinely live attempt runs with no record at all.

So it stays suppressed, and deliberately *not* on the strength of the drain: leaning on the drain would rest correctness on the current call order in `rebuild_company` rather than on the invariant `reap_orphaned_runs` actually states. `reap_orphaned_runs`' own doc comment now says the proof is about boot and only boot, so a future second caller has to make the argument again instead of inheriting it.

## Q2 — what happens to a run record mid-attempt when the runtime is swapped?

Three cases; the third needed a code change.

**The attempt in flight completes and settles itself**, on the outgoing runtime, against the store the successor adopts. Its trace keeps writing through the drain — the sink holds an `Arc<dyn RunStore>` from the inherited `OpsStores`, not a handle to the runtime.

**The store is inherited whole.** `RuntimeHandover` carries `OpsStores`, so the `RunStore` the dispatch choke point mints into, the one the successor's `HarnessBrain` gets via `with_runs`, and the one the (suppressed) reaper would sweep are one instance. A second store would strand every live run and restart attempt ordinals at 1. Added to the inherit-vs-reconstruct table.

**The attempt that never starts was a real leak, now fixed.** This is the defect the rebase surfaced:

1. Board writes are deliberately not gated on the quiesce — only cycles are — so `upsert_task` in the window still passes the choke point and still mints its `Pending` row.
2. Its cycle is then refused by `ensure_accepting`, which runs in `CompanyRuntime::run_cycle` **before** `CycleRunner` takes the serial lock, and therefore before `begin_run`.
3. The cycle's terminality backstop only settles rows *that cycle* started, so it never sees this one. Every other dispatch failure is already covered in there; this is the one that escapes.
4. And the rebuild has just switched off the one sweep that would have caught it.

Net: a row claiming to be pending for the rest of the process's life, on a card that reads as under way by an attempt that never began, which nothing re-drives. Pre-#357 that same window just dropped the dispatch with a `warn`; #357 makes it leave a durable lie.

The refused dispatch now settles its own row — `Pending` → `Failed` carrying a new `ports::runs::RUNTIME_REPLACED_ERROR`, kept distinct from `ORPHAN_ERROR` so a run list can tell *"we swapped your runtime, retry"* apart from *"the host died"*. `Pending` → terminal is already the legal move the status table names for **"a dispatch that failed before the first turn"**, so this needed no new state, only a caller willing to use it. `Failed` rather than `Cancelled` for the same reason the reaper picks `Failed`: the runtime went away underneath a card an operator had just dispatched, and that wants to be visible, not filed away as an intentional stop. The card stays in `in_progress`, which is where every other failed dispatch cycle leaves it.

The spawned cycle body is split out of the `tokio::spawn` as `run_dispatch_cycle` so the path is reachable from a test.

## Mechanical rebase notes

* `dispatch_task` is now `async fn(self: &Arc<Self>, task: &TaskRecord)`. This PR adds no call site of it, and holds no guard across it.
* `CompanyEvent::TaskDispatched` gained `run_id`. No pattern match in this PR's code touches it.
* Everything else in `#353` applied unchanged. `HarnessBrain::new(…).with_runs(ops.runs.clone())` lands on the inherited `ops` on a rebuild, so the successor's brain writes to the same rows.

## Tests

Added `a_dispatch_refused_by_a_quiescing_runtime_settles_its_attempt` in `src/company/runtime.rs`, with a positive control: the same call on a **live** runtime goes down the ordinary path (cycle runs, backstop settles the row), so the assertion below it is specifically about the quiesce refusal. Then quiesce, re-dispatch, and assert the row is `Failed`/`RUNTIME_REPLACED_ERROR` with no start time and a finish time.

Per this repo's constraint the local build matrix is not run here — `rustfmt --edition 2024` is clean on every touched file, and CI compiles and runs the rest.

M3gA-Mind:
@coderabbitai review

Issue #301, opened 2026-08-03
board: add Planning column, collapse Backlog into Todo, and reduce New Task to a single prompt box


oxoxDev:
Part of the MVP spec epic #183 (§3 Task Board). Filed separately because it is concrete, small, and unblocked — unlike §4, which waits on #243.

## Current state

`TASK_COLUMNS` today is `backlog · todo · in_progress · paused · in_review · done`. Against the spec's `Todo · Planning · In Progress · Paused · In Review · Done`, that means **only Planning is missing**, plus the Backlog→Todo collapse. The To-do column and the `+`-restricted-to-Todo rule already shipped in #206/#215, and Paused already exists (#111).

## Work

1. **Add `planning`** between `todo` and `in_progress`. `BOARD_COLUMNS` in `src/ports/tasks.rs` is the source of truth and the REST write boundary rejects unknown columns, so the host and console must change together.
2. **Collapse Backlog into Todo.** Read the note below first — this one has a consequence.
3. **Reduce the New Task modal to a single free-text prompt and a Create button.** It currently carries Title, Note and Priority. The assignee picker added in #278 lives on the *edit* dialog and card detail; that stays.

## The coupling nobody has flagged

`backlog` is documented as where the lifecycle **returns work that needs another pass** — a failed dispatch, or an orchestrator `revise` verdict. Collapsing it into Todo removes the column that failure path currently targets.

So this issue must decide where a revised or failed task goes: back to Todo alongside genuinely new work, or to a distinct state. Shipping the collapse without answering that silently changes failure behaviour. Say which was chosen and why in the PR.

## Acceptance criteria

- [ ] Board renders the six spec columns in order; host and console agree on the set
- [ ] `+` appears only on Todo (already true — assert it still holds)
- [ ] New Task modal contains only a prompt input and Create; a created card lands in Todo
- [ ] A created card does **not** dispatch — entering `in_progress` remains the only dispatch trigger, and a regression test asserts no `TaskDispatched` fires on create
- [ ] The revise/failed-dispatch destination is explicitly chosen, implemented, and tested
- [ ] Existing cards in `backlog` are migrated, not orphaned

## Notes

The spend gate matters here: dropping a card into `in_progress` is what spends money. Nothing in this issue may weaken that. Auto-advance is §4's subject and is blocked on #243.

Scope: **medium**.

Issue #269, opened 2026-08-02
bug: TaskEditDialog renders Column and Priority blank on open


oxoxDev:
## Summary

`TaskEditDialog`'s Column and Priority selects render blank every time the dialog opens, even though the task being edited has both.

`draft` starts as `{}`, so `<Select value={draft.column}>` sees `undefined` on first render. Base UI latches the select as *uncontrolled* at that point and then ignores the seeding effect that runs a tick later. Confirmed by DOM probe: `#task-column` renders `<span data-placeholder>` with no value.

The assignee select in the same dialog is immune only because it binds `draft.assignee ?? ""` rather than a bare `undefined`.

Pre-existing on `upstream/main`; found while working #263 and deliberately left alone as out of scope there.

## Reproduction

1. Console → Tasks → open any card's edit dialog.
2. Column and Priority both show placeholder text rather than the card's actual values.
3. Saving without touching them is fine (the diff-PATCH from #263 omits unchanged fields), but the operator cannot see what the current values are.

## Expected behavior

Both selects show the task's current column and priority on open.

One-line fix each — bind through the seeded task the way the assignee field already does:

```tsx
value={draft.column ?? task.column}
value={draft.priority ?? task.priority}
```

Worth checking any other select in the console that binds a possibly-`undefined` draft field, since the failure is silent and looks like a styling issue rather than a state bug.

PR #354, opened 2026-08-04
fix(tasks): move a card from In review to Done by dragging it (#334)

### LINKED ISSUE #334: tasks: a card cannot be moved out of In review — drag to Done fails silently
## Summary

A task that reaches **In review** cannot be moved to **Done** from the console. Dragging the card onto Done appears to work and then does nothing — the column count does not change and the card stays put. No error, no toast, nothing in the UI to say why.

Found by QA on smoke1.

## Scope — read this before starting

The original report also said the Edit dialog has no status control. **That part does not reproduce.** `TaskEditDialog` renders Column, Priority and Assignee selects (`#task-column`, `#task-priority`, `#task-assignee`) on the build QA tested, and a related bug where those rendered blank was fixed in #269. So the Edit dialog is a working fallback today.

This issue is **only** the drag interaction, plus the silence around it.

## Why the silence is the worse half

The board writes a column change with `PATCH …/tasks/{id}`, and the host validates the target against `BOARD_COLUMNS`. A rejected or failed write currently produces no operator-visible signal on the drag path — the card simply snaps back. That is indistinguishable from "the app is frozen", and it is what makes this report read as *"there is no way to do this"* rather than *"this one gesture failed"*.

Worth checking as part of this: whether the drop is rejected client-side before any request is sent, or whether the request goes out and fails. The two need different fixes and only one of them is a host problem.

`in_review` and `done` are both legitimate columns in `BOARD_COLUMNS`, so the transition is not forbidden by the model.

## Acceptance

- [ ] Dragging a card from In review to Done moves it, and the counts update
- [ ] A drop the host refuses surfaces a reason rather than silently reverting
- [ ] The Edit dialog's Column control continues to work as the alternative path
- [ ] Verified in a browser — this is a drag interaction and is not observable any other way

## Verification

There is a Playwright harness at `frontend/test/e2e/`; CI does not run it, but it is how UI behaviour actually gets checked here. Isolate any local host with the `--home <dir>` flag.

## Related

- #269 — the Edit dialog blank-select bug (fixed), which is why the dialog is not in scope here
- #301 — board column rework (Planning column, Backlog→Todo); this should stay independent of it

M3gA-Mind:
Fixes #334

## Summary

A card in **In review** could not be dragged to **Done**: the drop appeared to work, the counts did not change, and nothing said why.

The issue asked which side refuses the drop — the host, or the console before any request goes out. I reproduced it in a headless Chromium with every drag event instrumented: **the console, before any request**. `PATCH …/tasks/{id}` is never sent. A host refusal, when one does happen, already surfaced correctly. Three separate causes, all console-side:

**1. The last column was mostly off-window.** `SidebarInset` is a flex item beside the sidebar, and a flex item's default `min-width: auto` floors it at its content's min-content width. That floor won: the inset measured a full window wide while starting a sidebar's width in from the left, so its trailing strip hung outside the window inside a wrapper that clips and cannot scroll. Those pixels were unreachable by any means. At 1280x800 the strip is 256px, and on the board it held the Done column. **Scrolled hard right, Done showed a 64px sliver** hugging the window edge. Every view was losing the same strip; the board is where #301's sixth column pushed real content into it.

**2. Nothing scrolled the board during a drag.** HTML5 drag-and-drop does not scroll a nested scroll container. Measured against a live drag session: the pointer parked on the board's right edge for 1.8s moved `scrollLeft` **0 → 0** of a possible 832. A column parked off-screen could not be reached by the very gesture the board's own hint recommends.

**3. Every miss was silent.** `moveTo` read the dragged card's id out of React state alone and returned wordlessly when it was absent — which it is for any drag that did not begin on a card. And the pixels between and around the columns had no drop handler at all, so a near miss did nothing and said nothing. With a 64px target, a near miss is the common case, not the rare one. That is what made one failed gesture read as *"there is no way to do this"* rather than *"this one gesture failed"*.

### What changed

* **`frontend/src/components/app-shell.tsx`** — `min-w-0` on `SidebarInset` removes the min-content floor, so the inset measures the space it actually has and the board's own `overflow-x-auto` handles the columns.
* **`frontend/src/views/TasksView.tsx`**
  * The board scrolls itself while a drag sits in an edge band, ramping from nothing at the band's inner lip to full speed hard against the edge (rAF-driven, torn down on dragend / dragleave / unmount).
  * A card stamps its id onto the drag's `dataTransfer` and the drop reads it back, with the old React state as the fallback. Filling the data store matters on its own too: a drag carrying nothing is aborted outright by Firefox and Safari, so the board's one documented gesture never even started there.
  * Columns claim their own drops; the board handles what is left over — the gap between two columns, the leading padding, the trailing gutter — and says the card has to land on a column.
  * Every exit from the move now speaks: a drop carrying no card, a card that has left the board, and a host refusal, which now names the card and the target column and carries the host's own words as the reason. The one deliberate silence is a card dropped back into its own column, which is a no-op rather than a refusal.

Every `dataTransfer` access is optional-chained: a real drag always carries one, a synthesized `DragEvent` does not, and the existing suite drives these handlers with synthesized events. Verified that path still moves the card.

### Measurements, before and after

| | before | after |
|---|---|---|
| Done column visible width, max scroll @ 1280px | 64px | 288px |
| `SidebarInset` right edge vs. window (1280px) | 1536 | 1280 |
| `scrollLeft` riding the right edge mid-drag | 560 → 560 | 560 → 832 |
| Drag In review → Done | no PATCH, no toast, counts unchanged | `PATCH {"column":"done"}`, In review 1→0, Done 1→2 |
| Released beside a column | silence | "Drop the card on a column to move it." |
| Host refuses with a 400 | bare message | names card + target column + host's reason, card rolls back |

### Scope

Only the drag interaction and the silence around it, as the issue directs. `TaskEditDialog`'s Column control is untouched and still works as the alternative path (#269). Independent of #301: this fixes the container the columns live in, not the column vocabulary.

## API Or Behavior Changes

No API change. The wire contract is untouched — the board still writes a column with `PATCH …/tasks/{id}` and the host still validates against `BOARD_COLUMNS`.

Console behavior changes:

* The app shell no longer extends past the right edge of the window. This affects every view, but only by removing overflow — no view loses width. Spot-checked Tasks, Team and Workflows: all sit exactly inside the window.
* The board scrolls itself when a drag is held near its left or right edge.
* A drop that cannot be applied now raises a message instead of reverting in silence. A card dropped back into its own column stays silent, as before.

## Tests

Run locally:

* `pnpm typecheck` — clean
* `pnpm typecheck:e2e` — clean
* `rustfmt --edition 2024 --check src/server/ops/write_test.rs` — clean

Added with the change:

* **`frontend/test/e2e/board-drag.spec.ts`** (new, 4 tests). The issue calls out that a drag is not observable any other way, and `frontend/test/e2e/` is where UI behaviour gets checked here. Drives a real browser against a live host: the shell no longer overshoots the window and Done is a full-width drop target; a card rides the board's edge from In review into Done, with the counts following and the host's own record confirming the move; a drop that misses every column surfaces a message and moves nothing; a refused move names the card, the column and the reason, and rolls back. CI does not run this suite.
* **`src/server/ops/write_test.rs::a_card_moves_from_in_review_to_done`** — pins the half of the report's question that *is* a host question: `in_review → done` is exactly the PATCH the drag sends, the write boundary accepts it, and the card lands in the terminal column and reads back from it. This one does run in CI.

## Documentation

No docs needed. The REST contract is unchanged, and no doc describes the console's drag mechanics — `docs/spec/runtime/ports.md` covers the column vocabulary and the write boundary, both untouched here. The reasoning for each fix is in the code comments and in the e2e spec's header, next to what it guards.

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -536,7 +536,15 @@ export function AppShell({
         <SidebarRail />
       </Sidebar>
 
-      <SidebarInset className="min-h-0">
+      {/* `min-w-0`: the inset is a flex item beside the sidebar, and a flex
+          item's default `min-width: auto` floors it at its content's
+          min-content width. That floor won — the inset measured a full window
+          wide while sitting a sidebar's width to the right of the origin, so
+          its last ~256px hung past the right edge of the window inside a
+          wrapper that clips and cannot scroll. On the task board that clipped
+          strip held the "Done" column, which is why a card could not be dragged
+          into it (issue #334); every view was losing the same strip. */}
+      <SidebarInset className="min-h-0 min-w-0">
         <header className="flex h-14 shrink-0 items-center gap-2 border-b px-4">
           <SidebarTrigger className="-ml-1" />
           <h1 className="text-sm font-semibold">{TITLES[view]}</h1>
--- a/frontend/src/views/TasksView.tsx
+++ b/frontend/src/views/TasksView.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from "react";
+import { type DragEvent, useCallback, useEffect, useRef, useState } from "react";
 import { Loader2, Play, Plus } from "lucide-react";
 
 import { createTask, listTasks, patchTask, type Task } from "@/api/tasks";
@@ -42,8 +42,44 @@ function readTaskDetailId(): string | null {
 /** How often to re-poll the board, so a dispatched card's result appears. */
 const POLL_MS = 4000;
 
+/**
+ * The MIME type a dragged card stamps its id onto (issue #334).
+ *
+ * The drop used to read the dragged id out of React state alone. That is a
+ * silent single point of failure: a drag that began on anything other than a
+ * card leaves the state null, and the drop handler then returned without a
+ * word. Putting the id on the drag itself makes a drop self-describing.
+ *
+ * Filling the data store at all matters for a second reason: a drag whose store
+ * is empty is aborted outright by Firefox and Safari, so the board's one
+ * documented gesture never even started there.
+ *
+ * Every read and write of it is optional-chained. A real drag always carries a
+ * `dataTransfer`; a synthesized `DragEvent` — which is how the e2e suite drives
+ * these handlers — does not, and the `dragId` fallback covers that case.
+ */
+const CARD_MIME = "application/x-opencompany-task";
+
+/**
+ * How near the board's left or right edge a drag has to come before the board
+ * starts scrolling itself, and how fast it goes once hard against that edge.
+ *
+ * The board is a horizontal scroller and, at six columns, wider than an
+ * ordinary window: the last column sits off the right edge. HTML5
+ * drag-and-drop does not scroll a nested scroll container on its own — a drag
+ * parked on the edge moves it zero pixels — so without this the far column
+ * cannot be reached by the very gesture the board's own hint recommends.
+ */
+const EDGE_BAND_PX = 72;
+const EDGE_SPEED_PX = 16;
+
 function priorityStyle(priority: string): string {
   return PRIORITY_STYLES[priority as keyof typeof PRIORITY_STYLES] ?? PRIORITY_STYLES.low;
+}
+
+/** A column's board label, for messages the operator reads. */
+function columnLabel(id: string): string {
+  return TASK_COLUMNS.find((c) => c.id === id)?.label ?? id;
 }
 
 /**
@@ -80,6 +116,14 @@ export function TasksView({
   // A real HTML5 drag fires a trailing click; suppress it so a drag never also
   // opens the detail dialog.
   const dragged = useRef(false);
+  // The horizontal scroller holding the columns, so a drag near its edge can
+  // scroll it (issue #334).
+  const boardRef = useRef<HTMLDivElement | null>(null);
+  // Pixels per frame the board is currently scrolling itself by, and the frame
+  // that is doing it. Both refs rather than state: this is driven by dragover
+  // at pointer rate and must not re-render the board on every move.
+  const edgeSpeed = useRef(0);
+  const edgeFrame = useRef<number | null>(null);
 
   const refresh = useCallback(async () => {
     try {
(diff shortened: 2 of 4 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (commented):
Reproduced the layout claim in isolation (headless Chromium, 1280x800, the real `sidebar-wrapper` → `SidebarInset` → board class stack): inset `width 1280`, `right 1536`, overshoot **+256px**, Done visible at max scroll **64px**. Adding `min-w-0`: inset `1024`, overshoot **0**, Done **288px**, board max scroll 576 → 832. The `min-width: auto` diagnosis holds exactly, and `min-w-0` on the flex item is the conventional fix, not a narrower workaround. All three console-side causes are implemented, and each has an e2e test behind it.

- `nit` `frontend/src/views/TasksView.tsx:370` — the comment claims the cursor says "nothing over the dead pixels", but the board's own `onDragOver` (`:336`) calls `preventDefault()` unconditionally, so the gaps and trailing gutter *are* valid drop targets and inherit `effectAllowed="move"` — the cursor reads "move" there too. Behaviour is right (it's what lets the near-miss toast fire at all); the comment isn't.
- `nit` `src/server/ops/write_test.rs:321` — `tasks_crud_round_trips_under_both_scopes` (`:136`) already PATCHes `{"column":"done"}`, asserts 200 + the echo, and re-reads the board; its own inline comment labels that step "Drag (PATCH column)". The only delta here is seeding at `in_review`, and the boundary validates by `is_board_column` set membership with no transition rules — so this adds no coverage. The one novel claim is in the doc at `:319` ("no dispatch fires behind it"), which the test never asserts. Assert the journal has no `dispatched` entry (as `board-columns.spec.ts` does for Planning) and it earns its place; otherwise drop it.
- `nit` PR body, "API Or Behavior Changes" — "only by removing overflow — no view loses width" understates it. The inset's layout box goes 1280 → 1024 at a 1280px viewport, so every view reflows into 256px less; what is unchanged is *visible* width. Net improvement, since that strip was clipped before — but three views spot-checked out of ~25, for a global shell change with no e2e in CI, is thin. Worth a pass over the denser ones: Workflows canvas, Usage, Finances, Memory.
- `nit` `frontend/test/e2e/board-drag.spec.ts:85` — these drive real `page.mouse` drags, while `board-columns.spec.ts` (#301) states the opposite — that Playwright "does not drive React's HTML5 drag handlers reliably here" — and synthesizes `dragstart`/`dragover`/`drop`. No coverage conflict (Planning vs Done, and the `?? dragId` fallback keeps the synthesized path green — good call), but one of the two comments is now wrong. If real drags work, say so and retire #301's workaround; if they flake, these will, and CI runs neither suite so nobody will see it.

Issue #333, opened 2026-08-04
tasks: a task's Approvals tab never shows approvals, because approvals carry no task id


oxoxDev:
## Summary

The Task Detail page has an **Approvals** tab that is empty for every task, always — including tasks that demonstrably have an approval pending.

Found by QA on smoke1: a "Describe Workflow" request was sitting on the main Approvals page and was approved there, and the originating task's own Approvals tab still read *"No approvals in this run."*

## Why it happens

The tab is not querying approvals at all. It filters the task's **timeline** for entries of kind `approval`:

```
<ApprovalsTab entries={detail.timeline} />
...
const approvals = entries.filter((e) => e.kind === "approval")
```

and the timeline is derived from the task's own run window. The code says so directly in a comment a few lines above:

> *"Waiting is correlated to this task's run window — approvals are not linked"*

So this is a known-by-construction gap rather than a regression. **An approval record carries no task id**, and nothing correlates the two. The tab can only ever show what the timeline happens to contain, which is nothing.

## Why it is not a small fix

There is no object to hang the link on. An approval is parked against an agent and a tool call; a task is a board card. The thing that would join them — a **run** with an id, a task id, an agent, and a trace — does not exist yet.

That is **#242** (first-class run records with persisted trace, status and cost). Attempting this before #242 means inventing a correlation key that #242 will then replace.

## Acceptance

- [ ] An approval raised while working a task appears on that task's Approvals tab
- [ ] Approving from either surface reflects on both
- [ ] A task with no approvals still shows the honest empty state
- [ ] The correlation is a real id, not a time-window heuristic — a second task worked in the same window must not absorb the first's approvals

## Related

- **#242** — first-class run records. **This is blocked on it**; do not start before it lands.
- #243 — approval re-dispatch via single-use grants (shipped), which is what makes an approved call resume
- #183 §4 — the task lifecycle, where "In Review · waiting for your approval" is a station

Issue #305, opened 2026-08-03
tasks: show working time separately from waiting, and render waiting proportionally


oxoxDev:
Splits out of #184. **Not blocked — the blocker it cites already merged.**

## Summary

The Task Detail header is meant to answer the question the cost-free spec is built around: *how much of this task's elapsed time was the agent actually working, and how much was it sitting waiting on us?* Today it shows worked time only, and prints a placeholder saying waiting time arrives when approval parking lands.

Approval parking landed on 30 July (#172). The runtime journals it. So the placeholder is no longer describing a blocker — it is describing unfinished work, while reading to anyone looking at the screen as though the feature is still gated on something else.

## Scope

1. **Header: working vs waiting.** Derive waiting spans from the parked/resolved approval record the runtime already writes, and show them beside worked time rather than folded into one elapsed figure.
2. **Timeline: proportional waiting.** The timeline is currently a uniform list of fixed-height rows, so a task that waited four hours on one approval looks identical to one that waited four seconds. Waiting periods should occupy visual space in proportion to their duration — that contrast is the whole point of the view.

## Why it matters

This is the number that makes the screen worth opening. Without it, a task that took a day looks the same whether the agent was busy for that day or idle for it waiting on a human — which is exactly the diagnosis an operator needs and the only thing this screen can uniquely tell them.

## Acceptance

- [ ] The header reports worked time and waiting time as separate figures
- [ ] A task with no waiting shows no waiting figure, rather than a zero
- [ ] Waiting periods in the timeline are visually proportional to their duration
- [ ] A test covers a task with an approval that parked and later resolved, asserting the split is correct
- [ ] The placeholder text referencing #172 is gone

## Notes

Both halves are frontend work over data the backend already produces — no new route is expected. If the parked/resolved pair turns out not to be correlatable per-task from what's journalled today, say so rather than inferring the span, and the read side becomes part of this issue.

Issue #184, opened 2026-07-30
feat(tasks): Task Detail screen (09 spec, cost-free) — EPIC


oxoxDev:
## Problem

There is no screen that answers, for any piece of work, **what happened and who decided what**. The board (`TasksView`) shows cards; clicking one gives no detail — no timeline, no lineage, no approval record, no way to watch a run. This is the most important screen in the product: the Board is a list of these, the Inbox is the subset that need you. If this screen is good, the product works.

This epic tracks the **Task Detail screen** per the v1 `09 — Task Detail` product spec, **with the cost / currency dimension removed** (no per-line cost, no per-child cost, no total-cost header, no budget control/gate). The **elapsed working-vs-waiting split** becomes the single headline number.

## Proposal

Build the Task Detail screen (cost-free):

- **Header:** title · status · which node · autonomy mode · owner · origin (a link back to the message that created it) · **elapsed, split into working time vs waiting-on-a-human time** — the headline number.
- **Lineage:** parent above, children below, the node that handled each; hand-offs shown inline.
- **Timeline:** plain-English, past-tense lines written at event-creation time and rendered verbatim; **every line expands** to its scrubbed payload (prompt / tool args / response — never raw ids); **waiting rendered as a proportional line, not a gap**; retries grouped and collapsed.
- **Artifacts tab:** versioned outputs + the human-edit diff — see #187.
- **Approvals tab:** every gate the task hit (what was asked, the preview, who decided, how long, what changed) — from the immutable approval ledger (`runtime/journal.rs`); see #172.
- **Discussion tab:** human comments only (agents post to the timeline, never here).
- **Controls:** Stop / Retry / Reassign / Take-over / Change-autonomy / Cancel / Export. Retry-after-an-irreversible-action names what already fired and requires confirmation. (Raise-budget control removed with cost.) Steer verbs (pause/cancel/redirect) already exist.
- **Live:** the timeline appends live for a running task (reuse `turn_stream`).

## What already exists (reuse, don't rebuild)

- Timeline primitive: `AgentReply.steps: Vec<TurnStep>` — scrubbed, expandable, persisted + rehydrated + live-streamed.
- Approval record: `runtime/journal.rs` append-only ledger (`ApprovalParked/Resolved/Expired/Amended`) → satisfies the "no control can alter a past event" criterion for free.
- Controls: `listInflight` + `steerTask` (pause/cancel/redirect, #111).

## Acceptance (cost-free)

[...]

PR #355, opened 2026-08-04
feat(tasks): export a task's record as a readable document (#352)

### LINKED ISSUE #352: tasks: a task's record cannot be exported, so it never leaves the console
## Summary

There is no way to get a task's record out of the console. No export button on the Task Detail screen, no export endpoint behind `GET …/tasks/{id}`, nothing anywhere in the console or the API that renders a task's history to a file.

This is the other unmet criterion of epic #184: *"Export produces something a non-technical person can read unaided."*

## Why it matters

Task Detail is meant to answer, for any piece of work, **what happened and who decided what**. Today that answer only exists inside a React view, behind console access, live. Every case where the record has to leave the screen — handing an audit trail to a client, attaching the decision history to a dispute, briefing someone who does not have a console login, keeping a copy of a run after the company is archived — has no path at all.

The bar in the epic is deliberately not "dump the JSON". It is a document someone non-technical can read **unaided**.

## What already exists to export

`GET …/tasks/{id}` (`src/server/ops/tasks.rs`, `TaskDetail`) already assembles everything the document needs, and the console already renders all of it:

- the header — title, status, node, autonomy mode, owner, origin link, and the **worked / waiting** split (#305)
- the timeline — journal-ordered entries with a past-tense `label` and an optional **already scrubbed** `detail`
- waiting spans with `waitedMillis`
- artifact versions and the human-edit diff (#187)
- lineage — the neighbouring cards

So this is a rendering and delivery problem, not a data-collection problem.

## Design questions to settle before building

- **Format.** Markdown is the cheapest thing that reads well unaided and diffs; PDF is what a client actually asks for; a self-contained HTML file sits between them and keeps the waiting-band proportions the screen uses. Pick one and say why.
- **Where it renders.** Server-side behind an endpoint (works for automation, one implementation, needs the renderer in the Rust crate) or client-side from the detail the view already holds (no new endpoint, no export for a script). Prefer the server if a scheduled or API-driven export is plausible.
- **Redaction.** `detail` is already scrubbed on the way out of the host. The export must not reach around that — it should consume the same read the console consumes, not a privileged one.
- **Approvals.** Whether the export includes the approval record depends on #333; until an approval carries a task id, an exported "approvals" section would be as empty as the tab.

## Scope

**In scope** — one export of one task, covering header, timeline (including collapsed failures, expanded), waiting periods, artifact versions and lineage, in a format a non-technical reader can open and understand with no explanation.

**Out of scope** — bulk or board-level export, scheduled delivery, and any new data collection. If something is missing from the document, fix the read it comes from.

## Acceptance

- [ ] A task can be exported from the Task Detail screen in one action.
- [ ] The output opens without a developer tool and needs no key to interpret — no raw ids, no effect keys, no JSON.
- [ ] It carries the worked/waiting split, the full ordered timeline with details expanded, artifact versions, and lineage.
- [ ] Scrubbing is inherited from the existing read, not bypassed.
- [ ] Exporting alters nothing — no journal entry, no column change, no state on the task.

Sub-issue of #184.

M3gA-Mind:
## Summary

Closes #352. A task's record could not leave the console: no export button, no endpoint, nothing anywhere that rendered a task's history to a file. The answer to *"what happened and who decided what"* existed only inside a React view, live, behind a login.

`GET …/tasks/{taskId}/export` now answers **one self-contained HTML file** carrying what the Task Detail screen shows: the header and the worked/waiting split, the whole ordered timeline with its details expanded, the **text of every artifact revision** plus the human-edit diff, and the neighbouring cards. The screen gets an **Export** button next to Edit.

The issue left three questions open. Here is what was decided and why.

### 1. Format: a self-contained HTML file

The bar is "a non-technical person can read it unaided", which already rules out the JSON the screen fetches.

- **HTML** opens by double-click in any browser on any machine — no reader to install, no tool to explain. It is the only one of the three that keeps the **proportional waiting bands**: #305 exists so a four-hour wait does not look like a four-second one, and that is exactly the information a text format flattens away. It prints: the reader's own browser turns the file into the PDF a client asks for. And it costs **no new dependency** — the document is `format!` plus inlined CSS. No templating engine, no PDF crate, no headless browser.
- **Markdown** loses the proportions, and opened in a text editor by the non-technical reader this is *for*, shows raw `##` and `|` rather than a document.
- **PDF** is what gets asked for by name, but generating one server-side means a new rendering stack — and the browser's Print dialog already produces it from this file, which is the same PDF without the dependency.

The file is inert: no `<script>`, no external fetch, no webfont. It survives being emailed as one attachment and archived after the company is gone.

### 2. Where it renders: the server

One implementation serves the console button *and* an automated caller, so an audit or scheduled export needs no second renderer in whatever client wants it. The issue's own tiebreak ("prefer the server if a scheduled or API-driven export is plausible") points the same way. A client-side document would live only inside the React view — the same place the record is stuck today. The console's job here is delivery, not layout.

### 3. Redaction: structural, not procedural

The handler renders [`tasks::assemble_detail`] and [`artifacts::artifacts_for_task`] — **the same values** `GET …/tasks/{id}` and the Artifacts tab return. That is why `task_detail` was split into a handler plus an assembly function rather than the export growing its own read: an exporter that touched the journal itself would be a second, unreviewed path to the same events, one scrub away from disagreeing with the console about what an operator may see. There is nothing to keep in step here — the document renders what the screen renders because it is handed the identical value. `detail` text is already scrubbed at source before either caller sees it, and nothing in this module reads the event log.

Everything interpolated is HTML-escaped. A card titled `<script>alert(1)</script>` renders as text in a file that will be opened in a browser and forwarded to a client; that is pinned by a test.

### 4. Approvals: not in this PR — deferred to #333 (PR #349)

The first revision shipped a Sign-offs section correlated by run window, with the imprecision stated in the document. **That was wrong and it is gone.** #352 deferred it explicitly, and an approximate attribution is least defensible in the one artifact that leaves the console for a client or an auditor — a caveat sentence does not fix a fact that may not be true, it just moves the burden onto the reader.

The document now has no Sign-offs section. When #349 lands, `TaskDetail` gains an `approvals[]` keyed by a real id and the section arrives reading that — including the *pending* sign-offs which have no resolution event and so could never have appeared under the correlated version at all. Recorded in the module doc next to where the section will go.

Note the console's own Approvals tab is untouched: it is live, behind a login, and correctable in place. The constraint is on the exported artifact.

### Acceptance

- [x] A task can be exported from the Task Detail screen in one action — the Export button in the controls bar.
- [x] It carries the origin: the header names the conversation the card was opened from (`Opened from: the strategy conversation`) when `originChatId` is set. It names rather than links — the card carries a chat id, not a URL, and the document is read far from any console.
- [x] The output opens without a developer tool and needs no key to interpret — plain-English headings ("What happened, in order", "What this task produced"), human column labels (`In review`, not `in_review`), readable dates (`5 Aug 2026, 09:00 UTC`), no JSON, no effect keys. **Deviation, stated rather than unremarked:** the criterion says "no raw ids", and the footer prints one — the card's id, once, labelled `Reference:`. An audit or client copy needs a stable handle to quote back, and every *other* id is gone (lineage is by title, columns by label, authors by name). It is the only raw id in the file and it is not asked to carry meaning. Flagging it as a conscious exception to the criterion as written, not an oversight.
- [x] It carries the worked/waiting split, the full ordered timeline with details expanded, artifact versions and lineage. **Every revision's text is printed**, not just the latest — a metadata row saying revision 1 existed, without what it said, did not meet "artifact versions". The console collapses repeat failures and hides details behind a disclosure; a reader holding a file has nothing to click, so everything is printed once, in order.
- [x] Scrubbing is inherited from the existing read, not bypassed.
- [x] Exporting alters nothing — asserted by comparing the board rows and the journal length across the call.

### The worked/waiting split is computed once, host-side

The first revision hand-mirrored `workedMillis` / `waitingMillis` from `TaskDetailView.tsx` into Rust. They agreed, but nothing failed if they stopped, and the failure mode is an exported record disagreeing with the screen about how long a person was waited on.

Both are now computed in `TaskDurations` (`src/server/ops/tasks.rs`) and carried on `TaskDetail`; the console and the exporter read the same numbers. `TaskDurations::compute` is the single construction site, so a `TaskDetail` cannot carry durations that disagree with its own timeline.

The one thing a snapshot cannot carry is a *still-open* span, which the screen ticks every second. So the struct carries `workedLive` / `waitingLive` and `asOfMillis`, and a caller wanting a live figure adds `now - asOfMillis` to the live half.

**That extension is exact, not an approximation** — worth stating plainly, because the shape invites the opposite assumption. It is *not* "sealed total plus a delta", which would be wrong: a still-parked approval can overlap a resolved one, so adding a live span to an already-summed set double-counts the overlap. What makes it exact is that the merge happens at assembly, with the live span included and clamped to `asOfMillis`. Past that instant every closed span has already ended, so the merged total has exactly one interval still growing, and it grows second for second. Adding elapsed wall-clock is therefore the whole of the remaining arithmetic.

That is a four-line `extend()` in the console; the window pairing and the interval merge no longer exist client-side. Net −60 TS lines, and the merge is pinned by tests in `tasks.rs` where it now lives — including `live_spans_extend_exactly_and_sealed_ones_do_not`, which is the property above.

`waiting_band_height` stays mirrored, as suggested — it is presentation, and a drifted pixel curve misleads nobody about a fact.

`column_label()` is `pub(crate)`, not `pub`: presentation is not part of the port's contract and the exporter is its only consumer.

## API Or Behavior Changes

- **New route** `GET …/tasks/{taskId}/export` under both scope forms. `text/html; charset=utf-8`, `Content-Disposition: attachment; filename="task-<slug>.html"`, `404` for an unknown card. Read-only.
- **`ports::tasks::column_label()`** — new **`pub(crate)`** helper mapping a column id to its human label.
- **`TaskDetail.durations`** — new field on `GET …/tasks/{id}`: `{workedMillis, workedLive, waitingMillis, waitingLive, asOfMillis}`. Additive; every existing field is unchanged. The console now reads it instead of deriving the split itself.
- `ArtifactView`, `TaskDetail` and its members are now `pub(crate)`, and `assemble_detail` / `artifacts_for_task` exist as crate-internal functions. No wire shape changed; `GET …/tasks/{id}` and `GET …/tasks/{id}/artifacts` return exactly what they returned before.
- Console: `OpenCompanyClient.getDocument()` for document responses (returns the body *and* the host's `Content-Disposition` filename, so the console stops inventing a second name — `fileSlug` is deleted) (the JSON path turns a non-JSON body into an `{error, code:"unparseable"}` object rather than throwing), and an Export button that downloads through a blob so the platform-token deployment's `Authorization` header still travels — a plain `<a href>` would drop it.

## Review follow-ups (@oxoxDev)

| Ask | Done |
| --- | --- |
| `major` Sign-offs override #352's deferral | **Dropped from v1.** Section removed; returns on #349. |
| `major` `worked_millis`/`waiting_millis` hand-mirror the console | **Computed host-side** on `TaskDetail.durations`; console reads it. |
| `major` only the latest revision's body is printed | **Every revision's text** now printed. |
| `nit` origin carried but never rendered | **Rendered** as `Opened from`. |
| `nit` footer prints a raw task id | **Kept, and now stated explicitly** as a deviation (see Acceptance). |
| `nit` `column_label` new `pub` on a port | **Narrowed to `pub(crate)`.** |
| `nit` artifact path not covered by the escaping test | **`artifact_text_cannot_become_markup` added.** |
| `nit` unbounded `String` assembly | **Bounded** per revision body and per diff; both cuts announced in the document. |
| `nit` `fileSlug` is a third slug impl | **Deleted**; the console uses the host's `Content-Disposition` name. |
| `nit` `ArtifactKind` never printed | **Printed**, so a reference artifact is not a bare path. |

## Tests

Added with the change (CI runs them). The renderer is a pure function of its inputs including `now`, so the document tests assert against a fixed clock:

- `the_document_reads_as_prose_not_as_data` — the section headings, human labels rather than wire words, a readable date, no epochs, no camelCase field names, a complete document with styles inlined and nothing fetched or scripted.
- `every_timeline_entry_and_detail_is_printed` — nothing is collapsed or hidden.
- `the_worked_and_waiting_split_is_carried` / `a_task_still_waiting_says_so` / `the_document_prints_the_hosts_figures` (feeds durations no correct derivation would produce, so a renderer that recomputed would fail) / `the_waiting_band_curve_mirrors_the_console`.
- In `src/server/ops/tasks.rs`, where the arithmetic now lives: `worked_accumulates_every_dispatch_window`, `a_completion_without_a_dispatch_is_not_counted_from_zero`, `overlapping_waits_are_merged`, `an_instant_or_unknown_sign_off_is_not_a_wait`, `live_spans_extend_exactly_and_sealed_ones_do_not`, `a_reader_clock_behind_the_host_does_not_go_backwards`.
- `user_text_cannot_become_markup` — a title, a note and a reply each carrying a tag.
- `artifact_text_cannot_become_markup` — the artifact path too (`author_id`, version note, body, diff line text), which is the largest agent-written surface and was the one gap in the previous escaping test.
- `an_untouched_task_exports_an_honest_document` — empty sections explain themselves; Related work is omitted entirely rather than left as an empty heading.
- `the_attachment_is_named_after_the_task` — including the fallback when a title has nothing filename-safe in it.
- `artifacts_carry_their_versions_and_the_human_edit` — every revision, who wrote it, **the text of each**, and the diff.
- `a_reference_artifact_says_what_it_is` — an `image` / `file` artifact names its kind instead of rendering a bare URL or workspace path.
- `an_oversized_revision_is_bounded_and_says_so` — the output bound, and that the cut is announced rather than silent.
- `the_header_names_the_conversation_the_task_came_from` — the origin, including the absent and whitespace-only cases.
- `task_export_serves_a_readable_document_and_alters_nothing` (over the real router) — content type, attachment filename, content, the 404, and the board-and-journal comparison across the call.
- `every_column_has_a_human_label` — the new label map, asserted against literals so a rename stays a deliberate two-place edit.

Intentionally untested: the console's Export button. This repo has no frontend unit-test runner and CI does not execute the Playwright suite (it needs a live host, #281), so the console half is covered by `typecheck` / `typecheck:e2e` / `build`.

## Documentation

- `docs/modules/server/README.md` — a new "Exporting a task's record (issue #352)" section carrying all three decisions above, the pure-read guarantee, and the approvals-until-#333 note; plus the `tasks` route row updated and a `task_export` row added.
- `docs/spec/runtime/api.md` — the route in the console write-plane list.
- The module header of `src/server/ops/task_export.rs` carries the format/rendering/redaction rationale next to the code it explains.


M3gA-Mind:
@coderabbitai review

oxoxDev (changes requested):
Necessity holds up — Export is not speculative. Epic #184 lists it in the Controls bar and as an open acceptance criterion, and #352 is a faithful decomposition of it. What the diff adds beyond that is a second renderer for a screen that already exists, and two parts of it read as premature rather than wrong. The security work is genuinely solid; my asks are about scope and timing.

## Against the issue

- `major` — #352 explicitly deferred this: *"Whether the export includes the approval record depends on #333; until an approval carries a task id, an exported 'approvals' section would be as empty as the tab."* `task_export.rs:360` ships one anyway, correlated by run-window, with the imprecision written into the document itself. #349 is open right now. This is the issue's own guidance being overridden, not a gap it left open.
- `major` — Acceptance *"carries … the full ordered timeline, artifact versions, and lineage"* is only partly met on artifacts: `task_export.rs:416` prints a metadata row per revision but `:438` prints only the latest revision's body. The PR body says "every artifact revision plus the human-edit diff", which promises more than the file contains. Either print each revision's text or reword the claim.
- `nit` — #352 names the header as including the origin link. `TaskCard.origin_chat_id` is carried through `assemble_detail` and then never rendered — the exported record can't point back at the message that created the task. (`node` / `autonomy mode` from the issue text don't exist on the card at all, so those aren't on this PR.)
- `nit` — Acceptance *"no raw ids"*: the `task_export.rs` footer prints `Reference: <task id>`. Defensible for an audit copy and labelled as such, but it is the criterion as written — worth an explicit line in the PR body rather than an unremarked deviation.
- The other three criteria are met in the diff, not just asserted: the one-action button (`TaskDetailView.tsx:1094`), scrubbing inherited via the `assemble_detail` / `artifacts_for_task` refactor, and alters-nothing pinned by the board+journal comparison in `write_test.rs`. All three of #352's open questions (HTML, server-rendered, shared read) are implemented as described.
- `nit` — Scope creep is small but real: `column_label` is a **new public** function on `src/ports/tasks.rs:102` for a rendering concern, and `tasks.rs` / `artifacts.rs` widen six types plus all their fields to `pub(crate)` for one consumer. Nothing from #352's out-of-scope list (bulk export, scheduling, new data collection) leaked in. Premise still stands — main has no export route as of `4e50fcd`.

## Scope / sequencing

- `major` — `src/server/ops/task_export.rs:360` — per the bullet above, drop Sign-offs from v1 and add it on #349, or land this after it. An imprecise attribution is least acceptable in the one artifact that goes to a client or an audit.
- `major` — `src/server/ops/task_export.rs:524,556` — `worked_millis` / `waiting_millis` hand-mirror `workedMillis` / `waitingMillis` at `frontend/src/views/TaskDetailView.tsx:94,124`. I diffed both: they agree today, including the interval merge. But nothing fails if they stop, and the failure mode is the exported record disagreeing with the screen about how long a person was waited on. This PR's own argument for sharing `assemble_detail` — "hand both the same value rather than keep two reads honest by review" — applies to the derived durations too: compute host-side, put them on `TaskDetail`, have the screen read them. Removes the drift and ~60 lines. `waiting_band_height` is presentation; fine to keep mirrored.

## Security — no blockers found

Every interpolated value goes through `escape` (`task_export.rs:664`); the only unescaped interpolations are `u32` / `usize` / rounded `f64`. The document is inert — no script, link, img, `url()` or webfont. `ScopedCompany` gives the route the same authorization as the detail read it mirrors, and the card is resolved within `company.id()`, so it cannot reach another company's task. `text/html; charset=utf-8` plus `Content-Disposition: attachment` means it downloads rather than rendering in-origin. Artifacts carry references, not bytes, so nothing binary is embedded.

- `nit` — `src/server/ops/task_export/test.rs:216` — `user_text_cannot_become_markup` covers title, note and detail but not the artifact path (`author_id`, version note, body, diff line text — `task_export.rs:416-455`), the largest agent-written surface. Correct today; a later edit there wouldn't fail a test.
- `nit` — `src/server/ops/task_export.rs:95` — the document is assembled as one unbounded `String`; a long timeline plus several large artifact bodies has no ceiling on the response.
- `nit` — `frontend/src/views/TaskDetailView.tsx:1147` — `fileSlug` is a third slug implementation; the server already sends the filename in `Content-Disposition` and `getText` discards the headers.
- `nit` — `src/server/ops/task_export.rs:396` — `ArtifactKind` is never printed, so an `image` / `file` artifact renders a bare URL or workspace path with nothing saying what it is.

M3gA-Mind:
Thanks — both majors were right, and the second one changed the shape of the PR rather than just patching it. Pushed in e494bb1. Nothing declined.

**`major` Sign-offs override #352's deferral.** Dropped from v1 rather than landing after #349, so this isn't blocked on it. The section is gone, not softened — a caveat sentence doesn't fix a fact that may be untrue, it just moves the burden onto the reader, and you're right that an audit or client copy is the worst place for that. What lands on #333 is recorded in the module doc and `docs/modules/server/README.md`, including the detail that makes it a deferral rather than a hole: the *pending* sign-offs have no resolution event, so they could never have appeared under the correlated version at all. The console's own Approvals tab is untouched — it's live and behind a login; the constraint is on the exported artifact.

**`major` `worked_millis`/`waiting_millis` hand-mirror the console.** Done as described: both now live in `TaskDurations` (`src/server/ops/tasks.rs`), carried on `TaskDetail`, built by a single `TaskDurations::compute` so a detail can't carry durations that disagree with its own timeline. The console reads them. `waiting_band_height` left mirrored, as you said — presentation.

One wrinkle worth flagging, since the screen ticks a 1s clock and a plain snapshot would freeze it. `TaskDurations` carries `workedLive`/`waitingLive` and `asOfMillis`, and the console adds `now - asOfMillis` to a live half. **That is exact, not an approximation**, and it is deliberately *not* "sealed total plus a delta" — that would be wrong, because a still-parked approval can overlap a resolved one, so adding a live span to an already-summed set double-counts the overlap. It is exact because the merge happens at assembly with the live span included and clamped to `asOfMillis`: past that instant every closed span has already ended, so the merged total has exactly one interval still growing, second for second. Pinned by `live_spans_extend_exactly_and_sealed_ones_do_not`. Net effect is that the window pairing and the interval merge leave the client entirely — what's left is a four-line `extend()`, −60 TS lines.

`TaskDetail` gains a `durations` field on `GET …/tasks/{id}`. Additive; no existing field changed, and no test asserts an exact key set.

**`major` only the latest revision's body.** Every revision's text now prints, each under its own author/when/why header. I replaced the metadata table rather than adding bodies beneath it, so the row and the text it describes travel together — the old split was exactly what let a reader learn revision 1 existed without learning what it said.

**Nits**

- *Origin never rendered* — now in the header as `Opened from: the <chat> conversation`. It names rather than links: the card carries a chat id, not a URL, and this document is read far from any console. Absent and whitespace-only cases covered.
- *Raw id in the footer* — kept, and now stated explicitly in the PR body as a conscious deviation from "no raw ids" rather than an unremarked one. It's the only raw id in the file; lineage is by title, columns by label, authors by name.
- *`column_label` new `pub`* — narrowed to `pub(crate)`. Presentation isn't part of the port's contract and the exporter is its only consumer.
- *`test.rs:216` misses the artifact path* — `artifact_text_cannot_become_markup` added, covering `author_id`, the version note, the body and diff line text. Agreed this was the gap that mattered: it's the largest agent-written surface.
- *Unbounded `String`* — this got worse before it got better, since every revision now prints. Each revision body is capped (`MAX_BODY_CHARS`) and each diff is capped (`MAX_DIFF_LINES`), and both cuts are announced in the document — a reader must never be left believing they hold the whole text. A truncated diff loses nothing recoverable, since both revisions are printed above it.
- *`fileSlug` is a third slug impl* — deleted. `getText` became `getDocument`, which returns the body alongside the host's `Content-Disposition` filename, so the console stops inventing a second name that could disagree with what `curl -OJ` saves.
- *`ArtifactKind` never printed* — now named, so an `image`/`file` artifact says what it is instead of rendering a bare URL or workspace path.

Not compiled locally (full local build matrix is prohibited on this machine); CI covers fmt/clippy/test.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Both Rust lanes were red on e494bb1 — two unrelated failures, fixed in dcbe91d.

**`cargo fmt --check`** (`Rust` lane). One call in `render_artifacts` had a 62-character argument list, over rustfmt's `fn_call_width` of 60, so it wants the vertical form. My pre-push check measured against `max_width` (100) and missed the tighter sub-limit. I re-measured every call I added against 60 and moved two neighbours off the boundary — one sat at exactly 60.

**`E0063: missing field run_id`** (`Rust (openhuman, tinycortex)` lane). #357 added `run_id` to `CompanyEvent::TaskDispatched` on main. CI compiles the branch merged with main, so this shows up even though the PR is mergeable and needs no rebase. Main updated its own four literals in `write_test.rs`; the export test's dispatch is a fifth, added by this PR, so it needed the field too.

`run_id: None` is the honest value rather than a placeholder: the fixture journals a dispatch directly instead of going through the choke point that mints a run row (#242), and the exported document renders the timeline, which never reads `run_id`.

Verified by merging `upstream/main` into the branch locally with `--no-commit`, auditing the merged tree for any remaining `TaskDispatched` / `Effect` literal without the field (none), then aborting — no merge commit was pushed. Local builds are prohibited on this machine, so that is a source-level check, not a compile.

M3gA-Mind:
@coderabbitai review

PR #356, opened 2026-08-04
fix(tasks): name what already fired before re-entering a run (#351)

### LINKED ISSUE #351: tasks: Retry re-enters a run without naming the irreversible effects that already fired
## Summary

On the Task Detail screen, a task that is not in flight shows a single **Retry** button (**Resume** when the column is `paused`). It calls `patchColumn()` — a bare `PATCH …/tasks/{id} {"column":"in_progress"}` — with no confirmation step and no statement of what the previous attempt already did.

The same control bar gates **Cancel** behind `ConfirmButton`. Retry, which can re-fire real-world effects, is ungated.

This is the last unmet safety criterion of epic #184: *"Retry after an irreversible action names what already fired and requires confirmation."*

## Why it matters

The runtime already knows which effects are irreversible, and it already knows which ones fired. Neither fact reaches the operator at the moment they matter.

- **Classification exists.** `PolicyGate::evaluate_supervised` (`src/policy/gate.rs`) sorts effects by group and comments the rule directly: *"Irreversible / identity-touching effects always park"* for `EffectGroup::Sign | Publish | Identity`. `Spend` parks at or over the cap; `Send` parks for a first-time counterparty.
- **The record exists.** `src/runtime/journal.rs` keeps an append-only executed-key set (`record_executed` / `is_executed`), replayed from disk at boot, precisely so an effect is not executed twice.

So after a run that signed something, published something, paid a supplier or sent a first email to a new counterparty, the operator can re-enter that run with one unguarded click and no idea any of it happened. The board's own dispatch path has the same shape, but Task Detail is the screen that claims to be the record of what happened — the gap is worse here.

## Where it is

| Piece | Location |
|---|---|
| The ungated Retry | `frontend/src/views/TaskDetailView.tsx` — `ControlBar` → `patchColumn()`, and `resumeLabel` |
| The confirm pattern to reuse | same file — `ConfirmButton`, already wired for Cancel |
| Irreversibility taxonomy | `src/policy/gate.rs` — `evaluate_supervised`, `EffectGroup` |
| What actually fired | `src/runtime/journal.rs` — executed-key set |
| The read the screen consumes | `src/server/ops/tasks.rs` — `TaskDetail`, `GET …/tasks/{id}` |

## The read side is the real work

The console cannot classify anything from what it is served today. A `TimelineEntry` carries `seq`, `atMillis`, `kind` (`dispatched | reply | tool_failed | approval | completed`), a `label`, an optional scrubbed `detail` and `waitedMillis`. There is no effect group and no executed-effect list, so the frontend has no way to tell a sent invoice from a web search.

Expect the change to be: expose the task's executed irreversible effects on the task-detail read, then have the console name them in a confirm dialog before re-dispatch.

## Scope

**In scope**

- Surfacing, per task, which irreversible effects already executed.
- A confirmation on Retry/Resume that names them in plain English ("This task already sent an invoice to Acme and published a release note") and requires an explicit confirm.
- Retry with nothing irreversible behind it should stay one click — do not add friction where there is no risk.

**Out of scope**

- Changing what the policy gate parks. The taxonomy is settled; this issue only reads it.
- Idempotency or effect-skipping on retry. This issue is about *naming and confirming*, not about suppressing the second execution. If a retry should skip an already-executed effect, that is a separate question and the journal already holds the key set to answer it.

## Acceptance

- [ ] Retry on a task with executed `Sign` / `Publish` / `Identity` / capped `Spend` / first-contact `Send` effects opens a confirmation that names each one.
- [ ] Retry on a task with no irreversible history stays a single click.
- [ ] The names are plain English, not effect keys or tool ids.
- [ ] The list is drawn from the journal's executed record, not re-derived from the timeline labels.
- [ ] Dismissing the confirmation does not change the task's column.

Sub-issue of #184.

M3gA-Mind:
## Summary

Fixes #351. Retry on Task Detail was a bare `patchColumn()` with no confirmation and no statement of what the previous attempt already did. Cancel on the same control bar has always been gated by `ConfirmButton`; Retry, which can send a second payment or submit a second filing, was not.

As the issue says, the read side is the real work. Both facts needed to warn already existed in the runtime and neither reached the operator:

* `evaluate_supervised` classifies `Sign` / `Publish` / `Identity` / capped `Spend` / first-contact `Send` as the effects it refuses to wave through;
* the journal's executed-key set records what actually fired.

But an idempotency key is opaque. It answers "has this run?" and nothing else, so a `TimelineEntry` carrying `seq` / `atMillis` / `kind` / `label` / `detail` / `waitedMillis` gave the console no way to tell a sent invoice from a web search.

**What the record now carries.** `EffectExecuted` gains an optional `ExecutedEffect`: the effect kind, its amount, the board task it ran for, and whether the gate called it irreversible. Three deliberate choices:

* **Classified at execution time**, by a new `ManifestApprovalGate::is_irreversible` that delegates to `evaluate_supervised` rather than restating the rules. Two copies of "which effects are irreversible" would drift, and the copy in a warning dialog would drift silently — nobody notices a warning that stopped naming something.
* **Mode-independent.** A `full`-mode company executes a filing without ever parking it, so the retry dialog is the first and only place anyone learns it happened. A predicate that read the company's mode would report "nothing to warn about" in exactly the configuration that needs the warning most.
* **No payload.** The journal keeps the classification facts and not the contents, so a recipient or a message body cannot reach this operator-facing response even by accident.

**Attribution follows the cycle.** Under `supervised` — the default — an irreversible effect never executes in the cycle that emitted it: it parks, and the operator's approval opens a fresh cycle carrying only `ApprovalResolved`. `ApprovalParked` therefore gains an optional task id, and the approved execution reads the card back off it. Without that, every effect that went through the approval gate the way the policy intends would be attributed to nothing and named on no dialog. A batch naming two different cards stamps nothing rather than guessing: a false alarm is how a safety dialog stops being read.

**The console.** `GET …/tasks/{id}` gains `irreversibleEffects[]`; `ControlBar` renders `RetryButton`, which is the plain one-click button when that list is empty and a `ConfirmButton` naming each effect when it is not. Plain-language wording lives in `frontend/src/lib/language.ts` (`effectDone`) beside the existing `effectAction`, per that file's glossary rule that an operator never sees a runtime internal.

Acceptance criteria:

- [x] Retry on a task with executed irreversible effects opens a confirmation naming each one (`task_detail_names_only_this_task_s_irreversible_effects`, plus `RetryButton`)
- [x] Retry with no irreversible history stays a single click (`a_task_with_no_irreversible_history_reports_an_empty_list`; `RetryButton` returns the bare `Button`)
- [x] Plain English, not effect keys or tool ids (`effectDone`: "Sent a payment of $2,400")
- [x] Drawn from the journal's executed record, not the timeline labels (`RuntimeJournal::irreversible_effects`)
- [x] Dismissing does not change the column — Cancel closes the dialog and `patchColumn()` only runs from `onConfirm`

Out of scope and untouched, as the issue directs: what the policy gate parks, and idempotency / effect-skipping on retry. This names and confirms; a second run still re-executes.

## ⚠️ Overlap with #349 (issue #333)

Both PRs branch off `main` and both touch the task-detail read, so they will conflict textually. Flagging rather than reshaping #349, per instruction. The overlap is:

| Both add | #349 (#333) | This PR (#351) |
|---|---|---|
| `task_id` on `JournalRecord::ApprovalParked` | to link an approval to its card for the Approvals tab | to let the approval follow-up cycle attribute the executed effect |
| `CycleHostImpl.task_id` + a `cycle_task_id(...)` derivation | identical shape and semantics | identical shape and semantics |
| a `record_parked` fourth argument | `Option<&str>` | `Option<&str>` |
| a field on `TaskDetail` | `approvals[]` | `irreversibleEffects[]` |

The two `TaskDetail` fields are independent and additive. The real merge work is one hunk each in `journal.rs` and `cycle.rs`; I wrote them with the same names and the same shape on both branches so the resolution is mechanical. **#349's retained index is the superset** — it keeps park instant, effect kind and task per approval, so if #349 lands first, this PR's narrower `approval_tasks` map should be dropped in favour of `origins`, and `approval_task(id)` becomes `origins.get(id).and_then(|o| o.task_id.clone())`. Whichever order they land in, the second one needs that rebase and nothing more. Happy to do it on whichever is second.

## API Or Behavior Changes

* `GET …/tasks/{task_id}` gains `irreversibleEffects: IrreversibleEffect[]` (`kind`, `atMillis`, optional `amountUsd`). Additive.
* Retry / Resume on Task Detail now opens a confirmation when the task has executed irreversible effects. Unchanged — a single click — when it has not.
* `ManifestApprovalGate::is_irreversible` is new and public on the concrete gate; the `ApprovalGate` port is untouched.
* `RuntimeJournal::record_executed` takes an `ExecutedEffect`; `record_parked` takes a fourth `task_id` argument. Both crate-internal.
* Journal records stay backward compatible: a pre-#351 `EffectExecuted` line replays as a committed key with no description, so the at-most-once guarantee is exact and it simply contributes no warning.

## Tests

Focused tests added with the behavior change; CI runs them.

* `src/policy/gate.rs` — the taxonomy holds in `supervised`, `full` and `readonly`; `always_approve` ("ask me first") deliberately does not widen "cannot be undone".
* `src/runtime/journal.rs` — the executed record filters by task and by irreversibility, survives a restart, and does not double-count a re-committed key; an approval's task outlives its resolution; a pre-#351 line still replays as a committed key.
* `src/runtime/cycle.rs` — an effect executed by a card's run is recorded against that card and its payload never reaches the journal record; the `supervised` park → approve → execute path attributes the effect to the card that parked it; an ambiguous batch attributes to no card.
* `src/server/ops/write_test.rs` — the read names only this task's irreversible effects (three negatives: a reversible effect on the same card, another card's payment, and an effect with no card behind it), and an empty list for a task that only read and replied.

- [x] `cargo fmt --all -- --check` — run locally, clean

Frontend `pnpm typecheck` could not run locally (no `node_modules` in this clone); verified in CI.

## Documentation

`docs/spec/runtime/ports.md` gains a "What a retry would repeat (issue #351)" section next to the existing per-task correlation notes: where the classification comes from, why it is made at execution time and mode-independent, why there is no payload, and the additive replay contract.


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

oxoxDev (changes requested):
The classification half is solid: `is_irreversible` (`src/policy/gate.rs:248`) calls `evaluate_supervised` (`:256`) rather than restating it, so the two cannot drift; it is stamped at execution time into an append-only record, so a later cap change cannot rewrite what a past effect is called; and Retry is now gated exactly as Cancel is — confirm is reachable only via `AlertDialogAction`, Cancel/Esc never call `patchColumn`. The problem is coverage: the list can be empty, or short, while a payment really did fire.

- **blocker** `src/runtime/cycle.rs:372-381` — an approved *agent tool call* (`Effect::agent = Some`) is settled by `mint_grant`, **not** `execute_effect_once`; the tool then runs inside the next turn through `tool_call` → `rt.tools.invoke`. Nothing is journaled as an `ExecutedEffect`, so an operator-approved `composio_execute` payment never reaches `irreversibleEffects`. If the same card also sent a native email, the dialog opens and names only the email — that is a confirmation understating what already fired. The journal already holds the facts: `ApprovalGranted { grant }` carries tool + args, `GrantConsumed` means "the tool ran", and `approval_tasks` gives the card. Fold them in via `effect_for` + `is_irreversible`, or the dialog must state that it covers only part of the history.
- **blocker (same fix)** `frontend/src/lib/language.ts:63` — `EFFECT_DONE_LABELS` ships `composio_execute` / `composio_authorize` / `mcp_registry_tool_call` / `media_generate_*` entries that no current code path can produce. Either the grant path is meant to be covered (see above) or those five entries are dead.
- **major** `frontend/src/views/TaskDetailView.tsx:1106` — the comment claims "an empty list means the journal has nothing irreversible against this card, not that the console failed to classify something." Untrue in two replay cases: a pre-#351 `EffectExecuted` line has no description (`journal.rs:253-258`), and a pre-#351 `ApprovalParked` has no `task_id`, so an approval parked before the upgrade and approved after executes attributed to nothing. A card with mixed history gets a dialog listing only the newer half. Set a flag when replay sees an undescribed executed key and, when set, always confirm with one line saying earlier activity cannot be described.
- **major** `src/runtime/cycle.rs:581-592` — `record_executed` commits *before* `perform_effect`, and a failed perform leaves the record standing. Right for at-most-once, but it makes the list "what was committed to run", not "what actually fired" as `ports.md` and `tasks.rs:590` both claim — and `effectDone` renders it as flat fact ("Sent a payment of …"). Safe direction, wrong wording; soften the docs or the string.
- **nit** `frontend/src/lib/language.ts:63-90` — `EFFECT_DONE_LABELS` is a hand-mirrored copy of `EFFECT_LABELS`; a kind added to one only falls back to `titleCase(kind)`, i.e. the raw effect key. `satisfies Record<keyof typeof EFFECT_LABELS, string>` (with `EFFECT_LABELS as const`) turns the omission into a typecheck failure.
- **nit** `src/runtime/journal.rs:372` — linear scan of every effect the company ever executed, on each task-detail GET; `tasks.rs:598`'s "costs nothing extra" is not accurate for a long-lived company. A `HashMap<task_id, Vec<_>>` keeps it flat.
- **nit** `src/runtime/cycle.rs:731` — every production call site passes a single-event vec, so the ambiguous-batch arm is unreachable today; when it does fire the effect belongs to no card and warns nowhere. A `tracing::warn!` costs nothing.
- **nit** `frontend/src/lib/language.ts:86` — `money(2400)` renders `$2,400.00`; the doc comment and PR body both say `$2,400`.

i18n: this frontend has no i18n framework — `language.ts` is the locale layer, and `effectDone` is correctly placed there. The dialog's own sentences (`TaskDetailView.tsx:1145`, "Leave it alone", `${label} anyway`) stay inline, consistent with the pre-existing "Keep running"; noting, not asking.

## Against the issue

- **major** — AC1 ("names *each* one") is not met: satisfied for native effects, missed for approved agent tool calls and for pre-#351 history (blocker above). AC2, AC4, AC5 are met as written (`RetryButton:1125` returns the bare `Button`; the read is `journal.irreversible_effects`; Cancel does not patch). AC3 is met only for the 16 mapped kinds — `effect_for` sets `kind` to the *tool name*, so an unmapped tool renders a title-cased tool id, which AC3 explicitly forbids.
- **nit** — the issue names the board's dispatch path as having "the same shape" and does not put it out of scope; the PR neither gates it nor says why. Dragging a card into `in_progress` still re-dispatches with no warning (`company/runtime.rs` `upsert_task` → `dispatch_task`), and the read this PR adds is now available to it. Say "Task Detail only, board follow-up" or gate it.
- **nit** — the issue's "the read side is the real work" question is answered directly and well: journal-sourced, classified once at the gate, no payload retained (the `annual-return` assertion in the cycle test pins it).
- No scope creep found — `ApprovalParked.task_id`, `approval_task` and `cycle_task_id` are all load-bearing for attribution, and the `ConfirmButton` `cancelLabel`/`details` additions keep their previous defaults.
- Premise still holds on `main`: Retry is still a bare `patchColumn()` at `TaskDetailView.tsx:660`, and the issue has no comments amending it.

Overlap with #349: same `journal.rs` / `cycle.rs` / `tasks.rs` cluster, both add `task_id` to `ApprovalParked` and a field to `TaskDetail` — textual conflict is certain but semantics are additive, and the PR's own read that #349's `origins` is the superset matches the code.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Thanks — the coverage read was right, and both blockers were real. All four review items and all four nits are addressed in `abbe4e8`; `8215181` merges `main` after #357.

### blocker — `cycle.rs:372-381`, approved agent tool calls

`GrantConsumed` now carries the same optional `ExecutedEffect` that `EffectExecuted` does, so an approved `composio_execute` reaches `irreversibleEffects` like a native effect.

Two calls worth stating, since neither is forced by the shape:

**Described at redemption, not at minting.** Minting a grant only means the agent is now *allowed* to make the call. A grant that then expires unredeemed is a call that never ran, and warning about it would be exactly the false alarm this feature is careful about. `GrantConsumed` is the line that means "the tool ran", so it is the line that carries the description.

**Read back from the park record, not re-projected from `grant.tool` / `grant.args`.** The projection to an `Effect` (and therefore the `EffectGroup`) already happened once, in `ApprovalPolicy::effect_for`; re-deriving it at redemption would be a second copy of the classification, which is the drift you already flagged on the label tables. So `State` retains the parked effect past resolution — payload replaced with `Null` on the way in, since classification reads only kind/group/amount/counterparty flags and this map outlives the queue entry — and an approve-with-edit overwrites it, because the grant is minted against the amended arguments and reporting the pre-edit amount would name a payment nobody approved.

The three facts then join on the `ApprovalId` the redemption reports: the retained effect, the same `approval_task` lookup the native approved path uses, and the same `ManifestApprovalGate::is_irreversible` asked at the moment the tool ran. `None` when the park record is not recoverable — journaled anyway, contributes no warning, same additive degradation a pre-#351 `EffectExecuted` line has.

### blocker (same fix) — `language.ts:63`

Four of the five are now producible: `composio_execute` → `Send`, `composio_authorize` → `Identity`, both `media_generate_*` → `Spend`, all classifying irreversible. `mcp_registry_tool_call` maps to `EffectGroup::Other` and so can never be irreversible — it stays because of your nit at the same lines, which I took: `EFFECT_DONE_LABELS` is now

```ts
} satisfies Record<keyof typeof EFFECT_LABELS, string>;
```

with `EFFECT_LABELS` un-annotated so its keys stay literal. That makes the mirror exhaustive both ways, and it also settles what the table's contract is: *every kind the glossary names*, not *every kind observed in a journal*. Pruning to the latter would break the moment a classification changed. One `labelFor()` helper widens the literal key types back to a lookup, so the widening is a deliberate single line rather than an annotation that quietly disables the check.

### major — `TaskDetailView.tsx:1106`, the untrue comment

Replay raises a company-wide flag on an undescribable executed key, surfaced as `historyIncomplete` on `GET …/tasks/{task_id}`. With it set, Retry confirms regardless and the dialog says the earlier activity cannot be described — "this list may be incomplete" when there is a list, "nothing here can be listed" when there is not. The comment now says an empty list means nothing irreversible *only when that flag is false*.

**One piece of this I am declining, and want to be explicit about it.** The flag fires on undescribed executed keys only — as you specified — and **not** on a `task_id`-less `ApprovalParked`. A pre-#351 park record is byte-identical on the wire to a legitimately card-less park written today (an approval parked from operator chat), so flagging it would show the warning on every company that has ever done that. Distinguishing them needs `Option<Option<String>>` on `ApprovalParked` with the field always serialized, and that breaks the deliberate shape-alignment with #349 on that exact field — the one thing keeping the merge mechanical. In practice a company old enough to hold a pre-#351 park also holds pre-#351 executed lines, so the flag is set and the same warning shows. Documented on `has_undescribed_history` rather than left implicit.

### major — `cycle.rs:581-592`, committed vs fired

You offered docs or the string; I did both, narrowly. Every "what actually fired" claim is now "committed to run" (`ports.md`, `tasks.rs`, `ExecutedEffect`, `tasks.ts`, `execute_effect_once`), stating that the record is written before the perform, that a failed perform leaves it standing, and that the runtime never re-attempts it — so an entry is something an operator must *assume* happened, not something proven to have completed. The dialog gains one line rather than a rewrite:

> Each is recorded the moment it was committed, so one that was interrupted still appears — nothing is ever retried on its own.

The labels stay past-tense. "Must be assumed to have happened" is what the operator has to act on, and hedging every sentence would cost more trust than the edge case is worth.

### nits

- `EFFECT_DONE_LABELS` mirror — done, above.
- `journal.rs:372` — now a `HashMap<task_id, Vec<_>>` index. Reversible and card-less effects are dropped on the way in, so it is strictly *less* memory than the flat `Vec` it replaces, and the "costs nothing extra" claim in `tasks.rs` is now accurate for a long-lived company.
- `cycle.rs:731` — `tracing::warn!` on the ambiguous arm, naming both cards and saying the consequence (effects attributed to neither, named on no dialog — which looks exactly like a cycle that did nothing irreversible).
- `money(2400)` — the doc comment says `$2,400.00`.

### against the issue

**AC1** — met once the grant path and the replay flag land; those were the two gaps. **AC3** — met generally now, not just for the 16 mapped kinds: an unmapped kind falls back to "Did something that cannot be undone" (with the amount when there is one) instead of title-casing a tool id. A tool name in that dialog is a runtime internal in front of an operator, which the glossary rule forbids and which reads as a system leak in the one dialog that has to be trusted; saying less is the better failure.

**Board dispatch** — taken as *Task Detail only, board follow-up*, stated in `ports.md` and on the `RetryButton` doc rather than left unsaid. Happy to open the follow-up if you want it tracked.

### tests

`journal.rs` — a redeemed grant names itself against its card and survives a restart; an undescribed redemption still journals and warns about nothing; the park effect outlives resolution without its payload (asserted live *and* replayed); an amendment supersedes it; a pre-#351 line sets the flag and a fully-described journal does not. `cycle.rs` — the whole park → approve → mint → redeem path for a `composio_execute`, asserting the `GrantConsumed` line carries no arguments. `write_test.rs` — `historyIncomplete` false on a clean journal, true over a hand-written pre-#351 line.

### #357 / #349

`8215181` merges `main`. One conflict, in `CycleRunner::run`'s think step, where both sides had added. Kept both: `run_id` and `task_id` are different granularities — which attempt, and which card — so `CycleHostImpl::new` still takes the cycle's card and `backstop_dispatched_runs` still settles the attempt rows before the `?`. `cycle_task_id` matches `TaskDispatched { task_id, .. }` and ignores `run_id` on purpose: an effect belongs to the card across every attempt, and a payment sent on attempt one is exactly what attempt two must warn about. The ambiguity test now stamps distinct `run_id`s on two dispatches of the same card to pin that.

#349's overlap is unchanged by any of this — `approval_effects` is a new map beside `approval_tasks`, so if #349 lands first both collapse into `origins` the same way.

M3gA-Mind:
@coderabbitai review

Issue #388, opened 2026-08-05
store: the fs stores' write locks are per-instance, so two stores over one bundle serialise against nothing


oxoxDev:
Follow-up from #386, deliberately left out of that PR's scope.

## What

`PathLocks` in `src/store/fs.rs` is a **field** on `FsCompanyStore`, `FsEventLog`, `FsMemoryStore`, `FsContextStore` and `FsInboxStore`, and each `new` builds a fresh one. Two instances over the same bundle therefore share no lock — the same shape #386 fixed on `RuntimeJournal`, where the lock was a field and two journals over one path serialised against nothing.

`FsOps` (`src/store/fs_ops.rs`) has the same field-scoped lock across its read-modify-write-then-`write_atomic` sequences. `FeedbackStore` (`src/feedback/store.rs`) is weaker still: a bare per-instance `TokioMutex<()>`, not even path-keyed.

## Why this is not a torn-line bug

Every one of these writes through `append_line` or `write_atomic`, both of which are atomic, so no line can tear. What is at risk is mutual exclusion:

- **`FsEventLog::append`** derives the next `EventSeq` by counting the file's lines under the lock. Two instances mean two events with the same `seq`.
- **`FeedbackStore::update_status`** rewrites the whole file via `write_atomic` under that bare mutex. A concurrent `append` from a second instance is not torn, it is erased.
- The `FsOps` read-modify-write pairs are lost-update rather than corruption.

## Why it was not fixed in #386

It is not currently reachable. Rebuild inherits store handles via handover rather than rebuilding them, provisioning rejects duplicate ids against the registry before building, and one runtime is registered per company — so each store is a singleton by discipline today and the locks do hold.

It is fragile rather than broken: nothing in the type system enforces it, the constructors are `pub`, and the CLI's export/import path freely mints a second set. Six structs across three modules is also more than a bug fix should carry.

## Suggested shape

Promote to a `static LazyLock<PathLocks>` for the fs stores, and a path-keyed static for `FeedbackStore`. `JOURNAL_WRITE_LOCKS` in `src/runtime/journal.rs` is the precedent, including keying on the absolutised path so a relative and an absolute spelling of one file match.

Worth stating in the fix that this is in-process only by construction: a second process on the same data directory is outside any lock's reach, and what keeps *that* case safe is the atomicity of `append_line` and `write_atomic`, not the lock.

[...]

Issue #387, opened 2026-08-05
store: a single malformed JSONL line stops a company booting


oxoxDev:
Follow-up from #386, deliberately left out of that PR's scope.

## What

`read_jsonl` (`src/store/fs.rs`) parses with `serde_json::from_str(line)?` inside its loop, so one malformed line fails the whole read. That is fine for the request-time readers, but one caller is on the boot path:

`FsCompanyStore::load` reads `ledger.jsonl` strictly, and `RuntimeBuilder::build` propagates the error with `?`. One damaged line in a company's ledger and the company does not start.

This is the same failure mode #386 just removed from `RuntimeJournal::load`, on a different file — an operator cannot repair a ledger through a console that will not start.

## Why it was not fixed in #386

The journal's strict read had to be fixed there because the journal's *writer* was producing torn lines, so real damaged files exist. `ledger.jsonl` has been written through `store::fs::append_line` since #43, so there is no known corruption source for it — this is a latent robustness gap, not a live defect. Fixing it means changing `read_jsonl`'s contract for roughly thirty call sites across `src/store/fs.rs` and `src/store/fs_ops.rs`, which did not belong in a bug fix.

The safety argument is easier here than it was for the journal: a skipped ledger row is a lost accounting line, not an un-committed at-most-once effect key, so tolerating it is unambiguously right.

## Suggested shape

Mirror `RuntimeJournal::load`: skip the line, log it against the file and line number, and report the count to the caller rather than returning `Err`. `CorruptLine` in `src/runtime/journal.rs` is the precedent — it carries line number, byte length and parse error but never the line's contents, because these reports get logged.

Worth deciding at the same time:

- Whether the request-time readers (events, traces, context index, inbox, feedback) should be tolerant too. Lower stakes — a 500 on one endpoint, not a dead tenant.
- Whether the import path in `src/store/export.rs` should stay strict. Refusing to half-import is defensible, but it should be a decision rather than an accident.

## Related

#386 — the journal half of this, already fixed.

Issue #243, opened 2026-07-31
feat(approvals): re-dispatch approved tool calls via single-use grants


oxoxDev:
## Summary

Approving a parked tool call records the verdict and clears the queue, but never re-dispatches the tool. The operator clicks Approve, nothing happens, and they have to re-ask from scratch — losing the run's context.

## Problem

The code says so plainly (`src/harness/policy.rs`):

> **Still out of scope:** resume-after-approval. openhuman resolves the decision inline, so approving a parked tool call records the verdict and clears the queue but does not re-dispatch the tool — the operator re-asks.

Two halves, both verified:

- The effect executor `perform_effect` (`src/runtime/cycle.rs:432`) only understands ledger, `{channel,text}` and `email.send` payloads. A harness tool effect — `kind` = tool name, payload = raw args — silently no-ops on approve.
- `resolve_approval` (`src/runtime/cycle.rs:277`) already runs a follow-up cycle, but the `ApprovalResolved` event renders as just `"approval Approve"` (`src/harness/orchestrator.rs:450`). Even if the model re-tried the tool, the stateless `ApprovalPolicy::check` would park it **again** — an infinite approval loop.

Steps to reproduce:

1. Run a company in supervised mode so a tool call parks.
2. Approve it in the Approvals view.
3. The card clears. The tool never runs. The operator re-asks.

The request side is otherwise good — `ManifestApprovalGate` (`src/policy/gate.rs:53`) evaluates `never_do` → `always_approve` → mode with a 7-day TTL that default-denies on silence, durable via journal replay.

## Solution

Close the loop with a **durable single-use grant plus guided re-issue**, not out-of-band dispatch. Tools live inside per-agent openhuman sessions with workspaces, MCP clients and credentials; running one outside a session would mean the model never observes its own tool result. The existing follow-up cycle is the resume vehicle.

1. **Capture the replay envelope.** `Effect` gains a serde-default `agent: Option<String>`; the approval request records the parking agent. Pre-upgrade journal lines decode to `None` and mint no grant — legacy re-ask behaviour, fail-safe.
2. **Mint the grant on approve.** When the resolved effect is a harness tool effect, journal `ApprovalGranted{approval_id, agent, tool, args, at_millis}` **before** inserting into a live shared `GrantSet`, and **skip** `execute_effect_once` for it — no dual execution path. For amended approvals the grant carries the **amended** args, so the operator's edit becomes the only executable version. Native effects keep their existing at-most-once path untouched.
[...]

PR #389, opened 2026-08-05
fix(runtime): make a journal append actually durable before it returns (#386)

### LINKED ISSUE #386: runtime: the journal can write a torn line, and a torn line makes the rest of it unreadable
## What happens

CI failed on an unrelated pull request with a journal line that could not be parsed:

```
runtime::journal::test::a_parked_approval_carries_its_task_across_a_restart_and_a_resolution
panicked at src/runtime/journal.rs:1226:31:
called `Result::unwrap()` on an `Err` value: Serde(Error("trailing characters", line: 1, column: 237))
```

"Trailing characters on line 1" means the first line of the journal held **two records with no newline between them**. The branch that failed does not touch `journal.rs` at all, and the same suite passes six times out of six locally — so this is not that branch's defect and not a one-off assertion. It is the journal writing a corrupt line.

The test is the messenger. The problem is the write path, and the journal is the durable record of every parked approval, every resolution and every run outcome.

## Why it happens — two independent faults in one function

`append` (`src/runtime/journal.rs:804-823`):

```rust
let _guard = self.write_lock.lock().await;
let mut file = OpenOptions::new().create(true).append(true).open(&self.path).await?;
file.write_all(line.as_bytes()).await?;
file.write_all(b"\n").await?;
Ok(())            // <- file dropped here
```

**1. The handle is dropped without a flush.** `tokio::fs::File` buffers writes and does **not** guarantee a flush on drop — its own documentation says a file should be flushed before it goes out of scope, because outstanding operations complete on a background blocking task that nobody awaits. `rg '\.flush\(\)\.await' src/` returns **nothing across the entire crate**, so no write path here flushes. Two appends in quick succession open two handles whose background flushes are not ordered against each other, which is exactly how a record lands mid-line.

**2. The write lock is per-instance, not per-path.** `write_lock` is a `TokioMutex<()>` constructed fresh in `RuntimeJournal::new` (`:369,:380`). Two `RuntimeJournal` values over the same file serialise against nothing — and the code does construct two: the failing test opens the path, writes, then opens it again as `reloaded` and writes more. Nothing in the type prevents a caller from doing the same in production.

Either fault alone can split a record; together they explain the observed line.

## Why this is not just a test flake

The journal is the crash-durability substrate. `#242`'s boot reaper, `#243`'s grant replay, `#228`'s run outcomes and `#333`'s approval-to-task links all reconstruct state by replaying it. A torn line does not degrade gracefully: `load()` fails the parse, and every record after the corruption is unreachable. On a hosted tenant that is a company losing its approval and run history on restart, with the only symptom being a startup error nobody sees.

The failure has been observed in CI. Nothing about the mechanism is specific to tests — it needs concurrent or rapid appends, which production has and the test suite merely reaches sooner.

## What the fix should do

1. **Flush before the handle is dropped**, and confirm the write actually reached the file rather than trusting drop.
2. **Write the record and its terminator as one write**, so a torn pair is not representable rather than merely unlikely.
3. **Serialise on the path, not the instance** — or make it impossible to hold two writers for one journal file. A per-instance lock guarding a shared file is a lock that does not guard anything.
4. **Make a torn line survivable on read.** `load()` should report the corruption with the file and line, and callers should decide whether to stop or continue, rather than an `unwrap` deep in the parse taking the process with it.
5. Audit the rest of the crate for the same pattern — no write path in `src/` currently flushes.

## How to verify

- A test that appends from two `RuntimeJournal` values over one path concurrently, then reloads and asserts every record parses and none is missing. It should fail before the fix.
- A test that appends many records in rapid succession and reloads, run repeatedly.
- A test that a deliberately truncated final line is reported as corruption rather than panicking.
- The existing `a_parked_approval_carries_its_task_across_a_restart_and_a_resolution` stays green under repeated full-suite runs.

## Related

Surfaced by CI on PR #375, which does not touch this file.
Affects everything that replays the journal: #228, #242, #243, #333.

oxoxDev:
## Summary

The runtime journal was reporting writes as durable before they reached the file. **199 of 200 appends returned `Ok(())` with their bytes still queued on tokio's blocking pool** — measured, not inferred.

The journal's central promise is that an effect's key is durable *before* the side effect runs. That promise was not being kept, and a real write failure — `ENOSPC`, for instance — reached nobody. The torn line that surfaced this in CI is the rare visible symptom of a window that was open essentially always.

## API Or Behavior Changes

**`append` is now one durable write.** It delegates to `store::fs::append_line` — a single blocking `O_APPEND` `write_all` whose syscall has completed before the call returns. This follows the fix already established twice in this crate rather than inventing a fourth variant.

**The write lock is process-wide, keyed on the absolutised path**, instead of per-`RuntimeJournal`-instance.

**`load()` no longer fails the boot.** A **merged** line now replays *both* records through serde_json's stream deserializer. That is a correctness requirement, not a nicety: skipping a merged line would un-commit an `EffectExecuted` key and let an at-most-once effect fire a second time. Only an unrecoverable line — a truncated tail — is skipped, logged with its file and line number, and reported through a new `corruption()` accessor.

`CorruptLine` carries the line number, byte length and parse error, and deliberately **never the line's contents**. The journal holds effect payloads, and this report exists to be logged.

## Tests

- [x] `cargo test --locked --features openhuman,tinycortex --lib` — **1881 passed, 0 failed**
- [x] `cargo test --locked --lib` (default features) — **1332 passed, 0 failed**
- [x] `git diff --stat Cargo.lock` — empty

Both suites were run repeatedly rather than once, because the reported symptom is timing-dependent and a single green run proves nothing: **15/15 green with features, 12/12 green on default features**, with `a_parked_approval_carries_its_task_across_a_restart_and_a_resolution` — the test that failed in CI — passing every run.

The durability window is now pinned by a **deterministic** regression test rather than a racy one. The two-instance tear is covered separately.

## Documentation

No spec doc changes. The durability contract and the corruption-recovery rule are documented at the code that enforces them.

## Impact

**Three corrections to the issue this closes**, all found by probing rather than accepting the write-up:

**1. The issue's second fault does not explain the failure it was written to explain.** The CI failure was three *sequential* appends on one journal — `reloaded` only writes *after* the `load()` that panicked. The two-instance race is real and is fixed here, but it was not the cause. The single-instance sequential tear could not be reproduced in ~100 attempts, which is what prompted measuring the mechanism directly and finding the durability window instead.

**2. `rg '\.flush\(\)\.await' src/` returning nothing is not a bug signal.** I cited it as one. There is no async buffered writer left in `src/` to flush, so its absence is the expected shape, not evidence of a missing call.

**3. This bug had already been fixed twice in this crate** — PR #43 in `store::fs::append_line`, then the feedback store, whose own comment calls itself "the remaining twin". The journal was the third twin and nobody noticed.

**On the lock, a deliberate choice against the issue's preferred option.** The issue offered "make two writers impossible to construct" as an alternative. A singleton `open()` cannot deliver what it promises — the hosting model cannot stop a second *process* on the same data dir, and a type asserting "impossible" when it is not is worse than a lock stating "in-process only" and meaning it. It would also have gutted the suite: roughly 30 tests construct a second instance to model a restart, and a singleton would hand back the same instance with its memory intact, so `reloaded.load()` would stop testing replay at all. The cross-process guarantee comes from the atomic write, which is free.

**Boot behaviour: skip loudly, do not refuse.** Refusing to boot converts one lost record into a lost tenant, and an operator cannot repair a journal through a console that will not start. The trade is not symmetric — a dropped resolution leaves an approval parked, which a person can still deny, whereas a dropped `EffectExecuted` lets an effect re-run. That asymmetry is exactly why merged lines are recovered rather than skipped. Boot logs one line naming the company and what the gap costs.

**Audit result: the journal was the last one.** There are zero remaining `tokio::fs::File` / `AsyncWriteExt` write sites in `src/` — every match is now a doc comment. Everything else is a whole-file one-shot or a blocking write inside `spawn_blocking`. The tar writer in `export.rs` calls `finish()` explicitly.

Two adjacent findings were **filed rather than folded in**, to keep this PR to one defect:

- **#387** — `read_jsonl` is strict and `FsCompanyStore::load` reads `ledger.jsonl` on the boot path, so one bad line stops a company. Same failure mode, different file — but that file's writer was already correct, so there is no known corruption source, and fixing it changes the contract for ~30 call sites.
- **#388** — the fs stores' and feedback store's `PathLocks` are per-instance. Not a tearing risk, since those writes are atomic, but `FsEventLog::append` derives `EventSeq` by counting lines under that lock and `FeedbackStore::update_status` rewrites the whole file. Currently unreachable — singleton by discipline rather than by type.

## Related

Closes #386

Surfaced by CI on PR #375, which does not touch this file. **Merging this unblocks that PR's Rust job.**
Affects everything that replays the journal: #228, #242, #243, #333.
Follows the same fix as PR #43, which corrected the first of these three twins.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when runtime history contains damaged, truncated, invalid, or combined entries.
  * Preserved valid records surrounding corrupted entries so startup can continue.
  * Improved reliability of rapid and simultaneous history updates, reducing missing or interleaved records.
  * Added clearer startup warnings when committed effects or pending approvals could not be replayed.
* **Tests**
  * Expanded coverage for corruption recovery, replay behavior, record visibility, and concurrent updates.

### DIFF
--- a/src/runtime/builder.rs
+++ b/src/runtime/builder.rs
@@ -923,14 +923,14 @@ impl RuntimeBuilder {
         // `Arc`s go to the delivery deps and to the runtime — one gate, one
         // journal, one approvals queue.
         //
-        // On a rebuild the journal is **inherited, never reopened**. It is the
-        // one piece where a second instance is not merely wasteful: `append`
-        // writes a record and its newline as two writes under a per-instance
-        // lock, so two journals on one path interleave onto a single line and
-        // fail to parse on replay — bricking the next boot, not just this
-        // process. `load()` is skipped for the same reason it is not repeated at
-        // boot: the inherited journal is already replayed, and re-reading it
-        // would re-apply records the live instance has since resolved.
+        // On a rebuild the journal is **inherited, never reopened**, and the
+        // reason is now the in-memory state rather than the file. Since #386 a
+        // second instance on one path cannot corrupt it — appends are whole
+        // `O_APPEND` writes serialised on a process-wide per-path lock — but it
+        // is still wasteful, and `load()` is skipped for the reason it is not
+        // repeated at boot: the inherited journal is already replayed, and
+        // re-reading it would re-apply records the live instance has since
+        // resolved.
         let journal = match handover.as_ref() {
             Some(h) => h.journal.clone(),
             None => {
@@ -938,6 +938,26 @@ impl RuntimeBuilder {
                     Bundle::new(home.clone(), &id).journal_jsonl(),
                 ));
                 journal.load().await?;
+                // Issue #386: a damaged line no longer fails the boot, which
+                // means the company can come up on an incomplete history. That
+                // is the right trade — an operator cannot repair a journal
+                // through a console that will not start — but it is only
+                // defensible if somebody is told. `load` already logged each
+                // line; this is the one line that names the company, because
+                // the effect keys behind it are what the at-most-once guarantee
+                // is made of.
+                let corruption = journal.corruption();
+                if !corruption.is_empty() {
+                    tracing::error!(
+                        company = %id,
+                        lines = corruption.len(),
+                        first_line = corruption[0].line,
+                        "journal lines could not be replayed; this company booted \
+                         without them, so committed effects may be missing from \
+                         the at-most-once set and approvals may be missing from \
+                         the queue",
+                    );
+                }
                 journal
             }
         };
--- a/src/runtime/journal.rs
+++ b/src/runtime/journal.rs
@@ -273,10 +273,30 @@ pub struct ExecutedEffect {
     pub irreversible: bool,
 }
 
+/// A journal line [`load`](RuntimeJournal::load) could not replay (issue #386).
+///
+/// Deliberately carries **no line content**. The journal holds effect payloads —
+/// recipients, message bodies, arguments — and a corruption report exists to be
+/// logged and read by an operator, which is the one place [`ExecutedEffect`]
+/// goes to some trouble to keep those out of. The line number locates it in the
+/// file, the byte length separates a merged pair (long) from a truncated tail
+/// (short), and the parse error names the column without quoting it.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CorruptLine {
+    /// The line's 1-based number in the journal file.
+    pub line: usize,
+    /// The line's length in bytes.
+    pub bytes: usize,
+    /// What the parse rejected.
+    pub message: String,
+}
+
 /// In-memory state rebuilt from (and kept in sync with) `journal.jsonl`.
 #[derive(Default)]
 struct State {
     executed: HashSet<String>,
+    /// Lines the last replay could not read — see [`CorruptLine`].
+    corrupt: Vec<CorruptLine>,
     /// Every irreversible effect that ran for a board task, indexed by that
     /// task and oldest first within it (issue #351).
     ///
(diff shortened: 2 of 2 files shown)


oxoxDev:
Both findings were valid. Addressed in 161e82d.

**The UTF-8 one was the important catch.** `load` used `read_to_string`, which fails on the first invalid byte anywhere in the file — and a torn write splits a multi-byte codepoint, so the damaged line isn't merely bad JSON, it isn't valid UTF-8 at all. That turned exactly the damage this recovery path exists for back into the whole-boot failure it exists to prevent, and no amount of per-line JSON handling downstream could have saved it. Now reads bytes and decodes per line, so a mangled line lands on the `CorruptLine` path with the rest of the journal intact.

Covered by a new test that was verified to gate: reverting just the read path fails it with `stream did not contain valid UTF-8`.

**On durability** — narrowed the claim rather than adding `sync_data`. The doc now states process-crash durability explicitly and says what it does not cover. Adding an `fsync` per append is a real cost on a path that runs on every parked approval, resolution, effect and run outcome, and it defends a failure mode the rest of the system doesn't defend either — the boot reaper and the interrupted-run sweep are both built around a process dying, not a host disappearing. That's a decision worth making deliberately rather than as a review fix, so it's filed as #392 with what would settle it.

Gates re-run: fmt, clippy `-D warnings`, 1882 tests with features, 1333 on default features, `--all-features` check. `Cargo.lock` unchanged.

Issue #448, opened 2026-08-06
workflows: a run cancelled before it starts can report itself as not cancelled


oxoxDev:
## What happens

`run_workflow` decides whether a run was cancelled with an unbiased `tokio::select!`:

```rust
let outcome = tokio::select! {
    outcome = &mut engine => Some(outcome),
    () = ctx.cancel.cancelled() => None,
};
```

`tokio::select!` polls its branches in **random order** and picks among whichever are ready. On a token that is already cancelled, the `cancelled()` arm is ready on the very first poll. If the engine future is also ready on that poll, which arm wins is a coin flip — and when the engine arm wins, the run settles as a completed run with `cancelled: false`.

So a run that was stopped can come back reporting it was not stopped. Both `select!` sites in the runner have this shape.

## Evidence

`workflows::runner::tests::a_cancelled_run_parks_no_gates` fails on `assert!(run.cancelled)` at `src/workflows/runner.rs:1977`. The test cancels **before** calling `run_workflow`, so there is no race in the test itself — the nondeterminism is entirely inside the runner.

Measured on clean `main` @ `67fe0e9`:

| Condition | Runs | Result |
|---|---|---|
| Test in isolation | 24 | 24 pass / 0 fail |
| **Full `--lib` suite** | 5 | **2 pass / 3 fail** |
| Full suite, `biased;` with the cancel arm first | 5 | 5 pass / 0 fail |

Isolated runs hide it, which is why it reads as a flaky test rather than as a defect. Anyone who reproduces it the obvious way will conclude it is not real.

## Why it matters beyond the test

- **The behaviour is user-visible.** An operator stops a run and the run reports itself completed. Everything keyed off `cancelled` — the settle path, the report routing that a cancelled run is explicitly not supposed to trigger — then takes the wrong branch.
- **`main` is red about 3 runs in 5 on a full-suite gate.** Every branch's "tests pass" claim is currently a coin flip, and the habit that forms is re-running until green, which is how the next real break gets waved through.
- The two `select!` sites already carry careful comments about drop-ordering on the cancel arm — the cancellation path was thought about. The branch *priority* was not.

## What the fix should do

Make cancellation win deterministically when it is already signalled: `biased;` with the cancel arm first, at both sites. That preserves the existing drop-ordering (which the borrow checker currently enforces) and changes nothing when the token is not cancelled.

[...]

Issue #445, opened 2026-08-06
artifacts: publishing outside a task run reports success and records nothing


oxoxDev:
## What happens

An agent publishes a file it produced. The tool returns **success**. The agent tells the operator the deliverable is ready. It is not recorded anywhere the operator can reach — not on the Artifacts tab, not on a card, nowhere. The file exists only inside that agent's private sandbox directory on the server.

The operator goes looking for the deliverable and cannot find it. The agent, reading its own successful receipt, insists it is there.

Observed on a staging tenant, 2026-08-06.

## Why it happens

Publishing is staged, not written. The tool pushes the file onto a pending queue, and that queue is drained in exactly **one** place: the settle path of a **task** run, and only when the run succeeded.

A chat turn never reaches that path. So a publish made during a conversation is staged and then:

- never drained, because no task settles, and
- **cleared** by the next turn or the next task dispatch — the clear is deliberate and correctly reasoned, with a comment explaining that a chat turn's staged file must never be attributed to a card that did not produce it.

Every individual piece is right. Together they mean a publish outside a task run is a silent no-op.

## The receipt is the sharpest part

The tool's success message reads, in substance: *published, and it appears on this task's Artifacts tab when the run finishes.*

In a chat turn there is no task and no run that will finish. The message names a destination that does not exist for this caller, asserts a future that will not happen, and nothing anywhere checks whether the caller is in a position for it to be true.

An agent has no way to discover it was told something false. It reports the deliverable as published because it was told it was published — so the failure is laundered through the agent into a confident, wrong statement to the operator.

## Why this is worse than a missing tab entry

The deliverable is the point. Everything else — the plan, the gate, the board — exists to get to a thing the operator can open. This path produces the thing, tells everyone it worked, and drops it.

It also compounds the sibling issue about work running without a card: an agent asked for something in a conversation cannot get a card, so it takes the only path available, and that is precisely the path where publishing silently fails. The two defects meet exactly where the deliverable should have been.

Note that the file is not lost from disk — it sits in the agent's own sandbox. But the sandbox is not the company workspace the console shows, so an operator looking in the obvious place correctly finds nothing, and the agent describing "the workspace" means a different directory than the operator does. That naming collision is worth fixing alongside.

## What the fix should do

- **A publish outside a task run must not report success.** Either it records the artifact somewhere reachable, or it fails and says why — but it must never return a receipt promising a destination the caller cannot reach.
[...]

PR #456, opened 2026-08-06
fix(spine): work handed to an agent opens its card by construction (#442, #419)

### LINKED ISSUE #419: spawn_task / delegate_to_desk promise work that the per-turn cap then silently throws away
Found while auditing #410. Same family — a bound that nothing announces — but what gets dropped here is an *effect*, not a rendering, which makes it worse.

## What happens

`DelegationQueue::drain(cap)` takes `len().min(cap)` and then calls `guard.clear()`. The remainder is destroyed, not deferred to the next turn. `MAX_DELEGATIONS_PER_TURN` is **3**.

Meanwhile each queuing tool has already returned an unqualified success to the model:

- `spawn_task` → `Queued a task card: "{title}". It will be opened on the board this turn.`
- `delegate_to_desk` → `Delegated to the {desk} desk. Its lead will answer this turn.`

Ask an orchestrator to "open cards for each of these five items" — an entirely ordinary request — and it will report five cards opened. Two exist. Nothing anywhere says otherwise, and the operator's own record of the turn is the agent's confident (wrong) summary.

`ApprovalRequestQueue::drain` has the identical `min` + `clear()` shape with `MAX_APPROVAL_REQUESTS_PER_TURN = 8`; the same reasoning applies, though a dropped approval fails safe where a dropped card fails silently.

## Suggested direction

Pick one and make it true:

- **Refuse at the boundary** — once the cap is reached, the tool returns an error saying so, so the model knows the card was not queued and can say so.
- **Carry over** — keep the overflow for the next turn instead of clearing it, and tell the model it was deferred.

Either is fine; what is not fine is a success message for work that is about to be discarded. Whichever is chosen, the drain should also report what it dropped rather than clearing in silence.

Scoped out of #410 deliberately: this is a semantics decision about delegation, not a rendering fix.

### LINKED ISSUE #442: spine: work runs without ever opening a card — two paths, neither of which tracks it
## What happens

Ask for a real piece of work. The agent does it — reads a repository, gathers data, writes the file you asked for. **No card is ever created.** The board stays empty and the work exists only as a conversation that scrolls away.

Reported from a staging tenant, 2026-08-06, twice, by two different routes.

## Two mechanisms, one symptom

**Path one — you ask a desk directly.** The tools that open a card are attached only when the agent being built is the orchestrator. Every other roster agent, including every desk lead, is built without them. A desk agent asked to do a whole piece of work has no way to open a card even if it wanted one, so it does the only thing available: the work itself, inline.

**Path two — you ask the orchestrator, and it hands off.** The orchestrator has two ways to get work done and they behave completely differently:

| Tool | What it does |
|---|---|
| `spawn_task` | writes a card into To-do — tracked, plannable, gated, ends in a linked deliverable |
| `delegate_to_desk` | runs the desk lead's turn **inline, in this turn**, and creates no card at all |

Both are described to the model as delegation. When it picks the second, the work happens and nothing is tracked — which is the observed case: the requested file was produced, correctly, and no card exists.

There is a third way to lose one. When a desk delegation answers, a relay turn runs so the orchestrator can hand the answer back in one coherent reply — and **anything that relay turn queues is deliberately discarded**. If the orchestrator decides at relay time that this should have been a card, that card is dropped by design.

## Why the model can't be blamed for choosing wrong

The orchestrator's brief says to delegate *"when a request belongs to a specialist desk **or** should be tracked as work"* — one verb for two different decisions. A request that is **both** (a specialist should do it *and* it should be tracked) has no stated answer, and the tool described as actually performing the work is the one that skips the board.

Nothing downstream corrects it either: no path turns an inline desk delegation into a card after the fact.

## Why it matters more than a routing detail

The promise is *ask for something, watch it become work with an output you can open*. That promise silently does not apply depending on which agent answered and which of two similar-sounding tools it reached for.

The failure is silent in the worst direction: the agent is visibly busy and productive, and the deliverable is real, so nothing looks wrong until you go to the board and find nothing there. Everything downstream — planning, the prerequisite check, the gate, the settled-run mover, the link to the deliverable — is skipped in one step, because all of it hangs off a card that was never created.

## What the fix should do

The decision to settle first: **is a card the default for anything substantial, or is it opt-in?**

If default — the honest reading of the promise — then:

- A desk delegation opens a card by construction, rather than depending on the model choosing the card-shaped tool.
- A desk asked directly either opens its own card or says plainly that this conversation will not be tracked, **before** the work starts.
- The relay turn's discard stops being able to swallow a card.

If opt-in, then the console has to say so: a channel that will not track work should not look identical to one that will.

Either way, separate the two decisions in the orchestrator's brief. "Which desk should do this" and "should this be tracked" are independent questions, and one word for both is why they collapse.

## How to verify

- Asking the orchestrator for a substantial piece of work produces a card, whichever hand-off it chooses.
- Asking a desk directly produces a card, or a clear statement before the work begins that it will not be tracked.
- A trivial question does not mint a card on either path.
- A card cannot be lost because the turn that wanted it happened to be a relay.
- Whatever the console promises about a channel matches what that channel does.

## Related

The delegation tools, the drain, and the agent-build seam are the surfaces involved. Adjacent to the planning station, which only ever sees work that already became a card — and therefore never saw any of this.

oxoxDev:
## Summary

Ask for a real piece of work. The agent does it — reads the repository, gathers the data, writes the file you asked for. **No card is ever created.** The board stays empty and the work exists only as a conversation that scrolls away.

The failure is silent in the worst direction: the agent is visibly busy and the deliverable is real, so nothing looks wrong until you go to the board and find nothing there. Everything downstream — planning, the prerequisite check, the gate, the settled-run mover, the link to the deliverable — hangs off a card that was never created, so all of it is skipped in one step.

Three mechanisms produced it. Card-opening tools were attached only when the agent being built was the orchestrator, so no desk lead could open one even if it wanted to. `delegate_to_desk` ran the desk lead's turn inline and created no card while `spawn_task` did, and both were described to the model as delegation. And the relay turn's queue was drained and discarded **by design**, so a card decided at relay time was dropped.

Closes #442. Closes #419.

## API Or Behavior Changes

**A card is now the default for anything substantial.** This is a proposed product decision, not an established one — a reviewer should feel free to push back on the decision rather than the code. The alternative (opt-in, with the console saying plainly that a channel is not work intake) is cheaper but leaves the trap intact.

**A desk delegation opens its card by construction**, before the desk lead's turn runs — assigned to the resolved lead, in `in_progress`, settling through `lifecycle::landing_column` like every other card. Nothing asks the model to choose the card-shaped tool. It declines in exactly four cases: no task store, already inside a dispatched card (that card *is* the tracking), nothing substantial asked, or the chat handler already carded it.

The card id rides back on the existing `spawned_task` field, so the console's "Card opened" chip now tells the truth on desk and DM threads. **No frontend change was required** — that is the answer to the issue's last verification point.

**Substantial vs trivial**: the default is to track, and the carve-out needs positive evidence. Over 25 words, or containing a produce-verb (`write`, `draft`, `implement` — deliberately excluding noun-ish `build`/`report`/`review`) counts as work. Otherwise a trailing `?` or a leading wh-word, or a greeting that stays short, does not. `do the quarterly close` tracks; `what's the status of the build?` does not.

**The relay can no longer swallow a card.** The blanket discard is replaced by a loop that drops only `DelegateToDesk` and executes the rest, so re-delegation is still impossible while a card the relay opened survives.

**#419: a delegation past the per-turn cap is now refused at the tool boundary**, as an error inside the model's own turn naming the bound, rather than returning *"It will be opened on the board this turn"* for work the drain was about to destroy.

**The orchestrator brief and tool descriptions now separate the two decisions.** "Which desk should do this" and "should this be tracked" were one word, which is why they collapsed.

## Tests

- [x] `cargo test --locked --features openhuman,tinycortex --lib` — 2234 passed

`Cargo.lock` unchanged. The intermittent `a_cancelled_run_parks_no_gates` failure is pre-existing (#448, fixed in #455) — demonstrated by reverting these files to base and re-running, where it failed 4 of 6 on untouched code.

The test that carries the most weight reads the board **from inside the delegate's own turn** and asserts `(engineer, in_progress)`. A card written after the answer came back would satisfy every count-based test and still leave the work invisible while it was happening.

**Verified live** on a real host with a real store, real auth and real HTTP, driven by a scripted OpenAI-compatible endpoint so a real model turn could actually emit `delegate_to_desk`. Final clean-board walk: 7 messages produced 3 cards — hand-off substantial ✓, hand-off question ✗, desk-direct substantial ✓, question ✗, two small talk ✗, imperative → exactly one card. #419: five requested, three opened, items four and five refused as errors in the model's own turn.

**Unit-tested only:** that the relay cannot re-delegate, and cancellation returning a card to To-do.

## Documentation

The brief and tool descriptions are the documentation for this behaviour and are updated in place.

## Notes for review

**Live testing caught two things eighteen unit tests did not**, and both are worth a reviewer's attention:

`thanks!` opened a card. The chat layer appends an open-work briefing to desk-addressed messages *before* the brain sees them, so the classifier was scoring the annotation rather than the operator's words — and it was self-amplifying, since each card lengthened the next briefing. Fixed by extracting the marker into a shared constant and building the test's input from that same constant, so a wording change fails CI instead of silently un-splitting the two.

One message produced two cards. `task_intent.rs` already opens a deterministic card for leading-imperative messages at the REST layer. This path now stands down when that one has fired.

**Sibling defects found and deliberately not fixed** (their files belong to concurrent work): `PendingPublishQueue` is never drained on the chat/desk path — which is exactly where this issue and #445 meet, and note that this fix now gives that deliverable somewhere to land; `ApprovalRequestQueue::drain` is byte-for-byte the pre-fix #419 shape and still fully silent while `require_approval` tells the model an operator will be asked; `WorkflowRefQueue` silently truncates at 20.

**On widening the `is_orchestrator` gate**: `query_company` and `spawn_task` are the two a desk lead arguably needs, since a desk lead currently cannot file work for *later*, only for the turn it is in. That was deliberately left alone — desk agents share one `DelegationQueue`, and a desk turn pushing onto it mid-drain needs its own design. `delegate_to_desk` must stay orchestrator-only regardless, for the depth cap.

## Summary by CodeRabbit

* **New Features**
  * Substantial delegated work now automatically creates and tracks work cards through completion or cancellation.
  * Work-card references are included in operator results for improved visibility.
  * Briefings clearly distinguish immediate desk hand-offs from deferred board tasks.

* **Bug Fixes**
  * Delegation limits are enforced when work is submitted, with clear refusal messages when caps are reached.
  * Prevented duplicate hand-offs and re-delegation loops.
  * Improved handling of dispatched work and relay operations.

### DIFF
--- a/src/harness/orchestrator.rs
+++ b/src/harness/orchestrator.rs
@@ -70,9 +70,15 @@ use crate::ports::{CompanyStore, WorkflowRun, WorkflowRunner, generate_id};
 /// The manifest cognition-tier that marks the orchestrator agent.
 pub const ORCHESTRATOR_TIER: &str = "orchestrator";
 
-/// Max delegations drained from a single orchestrator turn (v1 cap). Anything an
-/// over-eager turn queues beyond this is discarded — delegation is bounded so a
-/// runaway turn can't fan out unboundedly.
+/// Max delegations one orchestrator turn may make (v1 cap) — delegation is
+/// bounded so a runaway turn can't fan out unboundedly.
+///
+/// Enforced **at the tool boundary** since issue #419
+/// ([`DelegationQueue::push_within_cap`]): a call past the cap is refused in the
+/// model's own turn, naming the bound. It used to be enforced only in the drain,
+/// which took `min(cap)` and destroyed the rest — after the tool had already
+/// told the model the card would be opened this turn. Ask for five cards, get
+/// told five were opened, find two.
 pub const MAX_DELEGATIONS_PER_TURN: usize = 3;
 
 /// How many recent events [`QueryCompanyTool`] surfaces.
@@ -128,10 +134,36 @@ pub fn is_delegation_tool(tool: &str) -> bool {
 }
 
 /// The orchestrator persona brief, appended to the orchestrator agent's persona.
+///
+/// # The two decisions are named separately (issue #442)
+///
+/// This brief used to say "delegate when a request belongs to a specialist desk
+/// **or** should be tracked as work" — one verb covering two independent
+/// questions. A request that was *both* (a specialist should do it **and** it
+/// should be on the board) had no stated answer, and of the two tools the model
+/// could reach for, the one described as actually doing the work
+/// (`delegate_to_desk`) was the one that touched no board. So the work got done,
+/// well, and nothing tracked it.
+///
+/// The brief now separates them, and — because a brief is guidance and guidance
+/// is not a guarantee — the runtime no longer depends on the model getting the
+/// second one right: a substantial hand-off opens its card in
+/// [`DelegationRunner::run_delegation`](crate::runtime::delegation::DelegationRunner::run_delegation)
+/// whichever tool was chosen. What the brief has to do now is stop the model
+/// *double-tracking* (a `spawn_task` beside every hand-off), which is why it
+/// tells it what `spawn_task` is still for.
 pub fn orchestrator_brief() -> String {
     " You are also this company's orchestrator: the single point of contact for the operator. \
-Answer from whole-company context, and when a request belongs to a specialist desk or should be \
-tracked as work, delegate instead of guessing. Use `query_company` to ground answers in the \
+Answer from whole-company context. Two decisions come up constantly and they are INDEPENDENT — do \
+not collapse them into one. (1) WHO SHOULD DO THIS: when a request belongs to a specialist desk, \
+hand it to that desk with `delegate_to_desk` rather than answering from your own guess; when it is \
+yours to answer, answer it. (2) SHOULD THIS BE TRACKED: you do not have to decide this, and you \
+must not pick a tool in order to influence it. Anything substantial handed to a desk is opened as a \
+board card automatically, and so is anything substantial an operator asks a desk or teammate \
+directly — the hand-off IS the card, so never call `spawn_task` alongside a `delegate_to_desk` for \
+the same work, and never prefer one over the other to get something tracked. Reach for `spawn_task` \
+only for work that belongs on the board but must NOT start in this turn: something for later, for \
+somebody else, or waiting on a person. Use `query_company` to ground answers in the \
 company's durable facts, recent activity, saved workflows, team roster, and desks — it is the \
 source of truth for what workflows exist, who is on the team, and which desks can take work, so \
 consult it before answering \"what workflows/teammates do we have?\" or before delegating, rather \
@@ -139,7 +171,7 @@ than guessing or naming a skill \
 — `delegate_to_desk` to hand a turn to a desk's lead \
 member, naming the desk by an id `query_company` lists under Desks (a desk is not a person: \
 handing work to a teammate's name is not a delegation), \
-`spawn_task` to open a tracked task card, `run_workflow` to execute one of the \
+`spawn_task` to open a card for work that should wait rather than start now, `run_workflow` to execute one of the \
 company's saved workflows by id (for example to advance or finish a task that is waiting on a \
 workflow run) — you can run workflows yourself; never claim the run_workflow tool is unavailable — \
 `create_workflow` to author and save a brand-new workflow graph (a trigger plus agent / tool / \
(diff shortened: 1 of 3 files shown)


oxoxDev:
CI red here is #448, not this branch: `workflows::runner::tests::a_cancelled_run_parks_no_gates` failed `assertion failed: run.cancelled` at `runner.rs:1977`. This branch touches nothing under `src/workflows/`.

That test races an already-cancelled token against the engine future in an unbiased `tokio::select!`, so which arm wins is a coin flip. #455 fixes it. Re-running CI here after #455 merges should clear it — worth noting this is also the first time the race has been caught in CI rather than locally.

Issue #475, opened 2026-08-06
ci: the integration suite runs nowhere — feature-gated out of one job, --lib'd out of the other


oxoxDev:
## What happens

`tests/one_card_per_message.rs` — the 11-case regression suite added by #466, which proves one operator message opens exactly one card — **never runs in CI**. Not "runs and passes". Never built.

Measured, not inferred:

```
cargo test --locked --test one_card_per_message
  → 0 passed; 0 failed          # default features: the file cfg's away

cargo test --locked --features openhuman,tinycortex --test one_card_per_message
  → 11 passed; 0 failed         # the feature set that has it
```

And the two CI jobs:

| Job | Command | Result |
|---|---|---|
| `Rust` | `cargo test --locked` | default features → the file is `#![cfg(feature = "openhuman")]` → **0 tests** |
| `Rust (openhuman, tinycortex)` | `cargo test --locked --features openhuman,tinycortex --lib` | `--lib` excludes `tests/` → **never built** |

One job has the feature but not the target. The other has the target but not the feature.

## Why it happened

The `--lib` flag carries a comment marking it **load-bearing** — do not widen to a bare `cargo test` — and that reasoning is sound: a bare `cargo test` under `--all-features` pulls in combinations CI does not intend to build.

Its stated justification, however, is *"There is no `tests/` directory."* That was true when written. #466 added one. The comment is now false, and nothing failed when it became false.

## Why it matters more than one uncovered file

The suite exists **because** the defect it guards was invisible to every other kind of test. It was found by driving a live host, and it was shown failing on base — 4 of 9 cases red at the parent commit — precisely so a future revert would be caught.

As it stands, that revert would be caught by nothing. The test passes locally, the PR was green, and the guarantee is decorative.

This is the same shape as the defects it was written to prevent: **coverage of a mechanism is not coverage that the mechanism is reached.** Three instances this week — a test whose fixture blinded it to part of its own surface, a case whose premise went vacuous, an outbox drain called only from `cfg(test)` — and now a whole suite wired to no runner.

## What the fix should do

- Run the integration suite in the job that has the features it needs, without widening `--lib` in a way that reintroduces the combination problem the comment warns about. A separate step (`--features openhuman,tinycortex --test <name>`, or `--tests`) is the obvious shape.
- Correct the stale comment, and state the real reason `--lib` is used rather than a fact that can silently expire.
[...]

PR #478, opened 2026-08-06
fix(console): one control answers to 'New workflow'

### LINKED ISSUE #341: console: two 'New workflow' buttons make the workflows view ambiguous
## What happens

The workflows view renders **two** controls both accessible as `New workflow`. Every Playwright spec that targets one fails:

```
strict mode violation: getByRole('button', { name: 'New workflow' }) resolved to 2 elements
```

This blocks `workflow-*.spec.ts` outright — the specs cannot proceed past the first click.

## Why it is a real bug, not just a test problem

Playwright's strict mode is doing what a screen reader does: it found two things with the same accessible name and refused to guess. A keyboard or screen-reader user hits the same ambiguity — two identically-named buttons with no way to tell them apart.

So this is an accessibility defect that happens to be caught by the test harness, not a locator that needs a `.first()` bolted on. **Fixing it by narrowing the selector would hide it.**

## Where it came from

Surfaced while running the full e2e suite against a live host during #301. It is unrelated to that change — #301 touches the tasks board, and these specs have zero references to it.

## What the fix probably is

Find the two controls and give them distinct accessible names, or remove the duplicate if one is a leftover. Likely candidates are a header action and an empty-state call-to-action that were added independently. Then re-run `workflow-*.spec.ts` — they should proceed.

## Acceptance

- Exactly one control in the workflows view answers to the accessible name `New workflow`, or any second one has a distinct name.
- `workflow-*.spec.ts` gets past the click without a strict-mode violation.
- The fix is in the component, not the spec's locator.

---

## Testing — required, and this one is bound to the UI

**This issue has a user-facing surface, so it is verified on that surface.** A green unit or component test is not evidence that this is fixed, and does not close it.

Before this is done:

- Run the console against a live host and walk the exact path described above, by hand, in a browser.
- Exercise the **failure** this issue describes, not only the happy path — the fix is that the described behaviour is gone, so the old behaviour is what you have to try to reproduce.
- If a Playwright spec covers it, that spec must actually run and pass in a lane that builds it. A suite that is compiled out of CI proves nothing (see #475).
- State in the PR what you ran it against and what you saw. "Tested" on its own is not a result.

Backend-only changes may be verified from the backend. Anything with a screen is verified on the screen.

M3gA-Mind:
## Summary

Fixes #341.

The workflows view rendered **two** controls both accessible as `New workflow` — the toolbar action, and the empty state's call to action — so `getByRole("button", { name: "New workflow" })` resolved to 2 elements and aborted every `workflow-*.spec.ts` before its first click.

The fix is in the component, not the locator, because the ambiguity is real: a screen-reader or keyboard user meets exactly what Playwright's strict mode met, two identically named buttons with nothing to tell them apart.

- The **toolbar** action keeps the name `New workflow`. It is the control present in every state, so it is the one that should answer to that name.
- The **empty-state** call to action becomes **"Create a workflow"** and opens the same dialog. Named that rather than "Create the first workflow" because deleting the last workflow lands in this same state, and by then there is nothing first about it.
- Both gain a `data-testid` (`workflow-create`, `workflow-create-empty`) so future specs have a stable handle without needing the name to be unique twice over.

New spec `frontend/test/e2e/workflow-create-affordance.spec.ts` pins it, deliberately using the **unqualified** role locator the failure arrived on — no `.first()`, no test id — so a regression fails here as a strict-mode violation rather than silently picking one of two. It covers both list states; the empty branch is reached by stubbing the workflows list with `page.route` (the suite's existing idiom), because the harness company ships `workflows/committed.toml` and the empty state is otherwise unreachable by driving the console.

### Acceptance, item by item

- **Exactly one control answers to `New workflow`, or any second one has a distinct name.** — The toolbar action alone carries the name; the empty-state control is `Create a workflow`. Confirmed in a browser: 1 match in the populated state, 1 match in the empty state (where both buttons are on screen together).
- **`workflow-*.spec.ts` gets past the click without a strict-mode violation.** — The unqualified `getByRole(...).click()` succeeds and opens the create dialog in both states. Before the change, the same click against an empty list fails with the issue's exact message.
- **The fix is in the component, not the spec's locator.** — No existing spec's locator was touched. `workflow-dialog-feedback.spec.ts` still targets `New workflow` unqualified, and the new spec does the same on purpose.

## API Or Behavior Changes

One user-facing string: the workflows empty-state button reads **"Create a workflow"** instead of "New workflow". Same action, same dialog. No API change, no `src/` change — this is `frontend/**` and `test/e2e/**` only.

## Tests

Run in `frontend/`:

- `npm ci`
- `npm run typecheck` — clean
- Browser verification, before and after the change, in a real Chromium driving the real console bundle (Vite dev server) with the host API stubbed at the network layer, in both list states:

| | `role=button name="New workflow"` | unqualified `.click()` |
|---|---|---|
| **before**, list populated | 1 | opens the dialog |
| **before**, list empty | **2** | `strict mode violation: getByRole('button', { name: 'New workflow' }) resolved to 2 elements` — the issue's exact failure |
| **after**, list populated | 1 | opens the dialog |
| **after**, list empty | 1 (plus 1 × `Create a workflow`) | opens the dialog |

**What I could not run, stated rather than glossed:** the live-host walk this issue's `## Testing` section asks for. `test/e2e/host.sh` needs `target/debug/opencompany`, and a local `cargo build` is prohibited on this machine; the Playwright lane is not built in CI either (#475). So the evidence above is the real component in a real browser with a stubbed host — it reproduces the reported failure and shows it gone — but it is not a walk against a running binary. Whoever has a host up can confirm in one step: open Workflows on a company with no saved workflows and check that only one button reads "New workflow".

The Rust boxes below are ticked as required by the lane; this branch contains no Rust change at all.

## Documentation

No docs change needed. No doc references the workflows empty-state button label (`grep -rn "New workflow" docs/ README.md` is empty), and the reasoning for which control owns the name now lives in a comment at each of the two call sites.

## Summary by CodeRabbit

* **New Features**
  * Improved workflow creation controls with clearer labels for toolbar and empty-state actions.
  * Both creation actions now open the workflow creation dialog.

* **Bug Fixes**
  * Ensured workflow creation buttons have distinct identifiers and predictable behavior.

* **Tests**
  * Added end-to-end coverage for workflow creation actions, empty workflow lists, and first-run tour handling.

### DIFF
--- a/frontend/src/views/WorkflowsView.tsx
+++ b/frontend/src/views/WorkflowsView.tsx
@@ -1153,7 +1153,18 @@ export function WorkflowsView({
               </AlertDialogContent>
             </AlertDialog>
           </span>
-          <Button size="sm" variant="outline" onClick={() => setCreateOpen(true)}>
+          {/* Issue #341: THE control named "New workflow". It is in the
+              toolbar in every state, so it is the one an operator — or a
+              screen reader, or a spec — should find under that name. The
+              empty-state call to action below is named differently on
+              purpose; two buttons answering to one name is an ambiguity
+              nothing can resolve. */}
+          <Button
+            size="sm"
+            variant="outline"
+            onClick={() => setCreateOpen(true)}
+            data-testid="workflow-create"
+          >
             <Plus className="mr-1.5 size-4" />
             New workflow
           </Button>
@@ -1244,9 +1255,19 @@ export function WorkflowsView({
         ) : !selectedId ? (
           <div className="flex h-full flex-col items-center justify-center gap-3 px-4 text-center text-sm text-muted-foreground">
             <p>This company has no saved workflows yet.</p>
-            <Button size="sm" variant="outline" onClick={() => setCreateOpen(true)}>
+            {/* Issue #341: opens the same dialog as the toolbar button, and
+                therefore must NOT carry the same name. "Create a workflow"
+                rather than "Create the first workflow" because this state is
+                also where deleting the last workflow lands, and by then there
+                is nothing first about it. */}
+            <Button
+              size="sm"
+              variant="outline"
+              onClick={() => setCreateOpen(true)}
+              data-testid="workflow-create-empty"
+            >
               <Plus className="mr-1.5 size-4" />
-              New workflow
+              Create a workflow
             </Button>
           </div>
         ) : (
--- a/frontend/test/e2e/workflow-create-affordance.spec.ts
+++ b/frontend/test/e2e/workflow-create-affordance.spec.ts
@@ -1,0 +1,95 @@
+import { expect, test, type Page, type Route } from "@playwright/test";
+
+/**
+ * Issue #341: the workflows view had **two** controls both accessible as
+ * "New workflow" — the toolbar action and the empty state's call to action.
+ *
+ * This is pinned in a browser because it is a fact about rendered output and
+ * nothing else. Both buttons call the same `setCreateOpen(true)`, so no unit
+ * test of behaviour can tell the fixed view from the broken one; the only
+ * difference is what the accessibility tree holds, which exists only once the
+ * component has rendered into a document.
+ *
+ * The assertions are deliberately written the way the failure arrived —
+ * `getByRole("button", { name: … })` with no `.first()` and no test id — so a
+ * regression fails here exactly as `strict mode violation: … resolved to 2
+ * elements` rather than silently picking one. A test id would pass over the
+ * defect, which is the whole reason #341 says the fix is in the component and
+ * not in the locator.
+ *
+ * Runs against the live host the harness brings up (see `playwright.config.ts`).
+ */
+
+/** The workflows list route, and only it — not `…/workflows/{id}` or `/run`. */
+function isWorkflowList(url: URL): boolean {
+  return /\/api\/v1\/(company|companies\/[^/]+)\/workflows$/.test(url.pathname);
+}
+
+/**
+ * Dismisses the first-run tour if it is up — its overlay swallows pointer
+ * events, so without this every spec fails on an unrelated modal. Tolerates its
+ * absence; a company that has seen it never shows it again.
+ */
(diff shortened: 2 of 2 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (commented):
The fix is the right one — in the component, not the locator — and the reasoning for renaming the empty-state control rather than the toolbar one holds up: the toolbar action is the only one present in every state, so it should own the name.

Confirmed the lane actually ran it: `Console E2E` executed 81 tests, 75 passed / 6 skipped, and both new specs are named in the log. That is the part worth having.

**One finding, non-blocking: the first test would pass against unfixed `main`.**

`workflow-create-affordance.spec.ts:44` asserts `toHaveCount(1)` after navigating to `/#/workflows`. But the list effect auto-selects the first entry, and the harness company ships `workflows/committed.toml` — so `selectedId` is non-null on load, the `!selectedId` branch never renders, and the second button is not in the tree. The count is 1 whether or not the bug is present.

The second test is carrying the whole proof. Its stubbed empty list is the only state in which both controls coexist, which also means the comment justifying that stub is correct rather than over-cautious — the empty state genuinely is unreachable by driving this company.

Worth closing because the risk is specific: if someone later removes the stub from the second test as redundant, coverage drops to zero while two green specs still claim to pin the defect. Two options, either is fine:

- Narrow the first test's claim to what it checks — the toolbar control opens the dialog — and leave the name-collision assertion to the second.
- Or stub the empty list in the first test too, so both actually render the offending branch.

Prefer whichever reads better; the important thing is that no test claims to pin the collision unless it renders both buttons.

Two smaller notes:

- `data-testid` added but deliberately unused in the assertions is a good call, and the comment saying why is worth keeping.
- This PR and #479 both edit `WorkflowsView.tsx`. Each is mergeable against `main` and nothing has built them together. Whichever lands second wants a build of the pair first, not just its own green check.

Issue #428, opened 2026-08-05
ci: the Playwright suite is type-checked and never run, so it has been red for months unnoticed


oxoxDev:
## The problem

`.github/workflows/ci.yml`'s `Console` job runs `typecheck`, `typecheck:e2e` and
`build`. It does not run `test/e2e/`, and nothing else does either. The job's own
comment is explicit about it: *"This does NOT run the suite: that needs a live
host and is still out of CI, per #281."*

That was a reasonable place to stop when #281 closed the "no job compiles the
console at all" gap. It has since become its own blind spot, and a worse one,
because a type-checked suite **looks** covered.

Concretely, found while fixing #406:

* `workflow-edit-delete.spec.ts` selects a workflow named `Committed flow` and
  asserts the disabled Edit/Delete buttons name `workflows/committed.toml`. That
  fixture **was never committed to `companies/e2e_harness`**. Two of the file's
  eight tests had therefore never passed, in any environment, since the day they
  merged. Nobody reported it, because nobody ran it.
* `oauth-onboarding-resume.spec.ts:69` — "the tour resumes on the Connections
  stop after a redirect" — fails against a current host. It is a genuine product
  failure and it is not new.

Type-checking a spec proves it compiles. It says nothing about whether it holds.

## What has already been done (#406)

`npm run e2e` now brings its own host up. With `PW_BASE_URL` unset,
`playwright.config.ts` starts `test/e2e/host.sh` — the `e2e_harness` company, a
freshly built console bundle, an isolated data root under `target/e2e/` — and
signs in against it. Setting `PW_BASE_URL` opts out entirely and preserves the
older contract.

So the suite is now runnable by anyone with one command, and the missing fixture
is committed. What is still missing is anything that *makes* it run.

## Current state, measured

Against a `cargo build --locked --bin opencompany` host (default features):

**70 passed, 5 failed, 2 skipped, 4.7 minutes.**

The five, and why:

| Spec | Cause |
|------|-------|
| `wiring.spec.ts:20` | needs the `openhuman` harness + a mock inference backend echoing `__MOCK_LLM__` — the spec's own header says so |
| `chat-to-card.spec.ts:83` | needs an orchestrator that actually opens a card |
| `workflow-run-history.spec.ts:116` | needs a workflow run that actually executes |
| `mcp.spec.ts:3` | self-declared: `PW_MCP_SERVER must point at the simple MCP server` |
[...]

Issue #414, opened 2026-08-05
Settings → MCP is built on an API the host does not serve (crashes on open)


oxoxDev:
Split out of #407 rather than folded into its fix (PR #413), because it is not a missing guard — the whole view targets a non-existent API and needs a real port.

## Symptom

Opening **Settings → MCP** throws `TypeError: Cannot read properties of undefined (reading 'length')`. Reproduced in a browser against a real host; it fails identically before and after #413, so it is pre-existing and not a regression from that PR.

## Cause

`api/client.ts` carries a second, contradictory MCP surface next to the correct one in `api/mcp.ts`:

| | declared in `client.ts` | what the host serves |
|---|---|---|
| `GET .../mcp/servers` | `Promise<{ servers: McpServer[] }>` | bare array of `McpServerDto` |
| `GET .../mcp/servers/{name}/tools` | `Promise<{ tools: McpTool[] }>`, keyed by `server_id` | bare array of `McpToolInfo`, keyed by **name** |

`lib/mcp.ts`'s `McpServer` (`server_id`, `transport`, `command`, `args`, `env_keys`, `status`, `tool_count`) matches nothing the host sends — the real DTO is `{name, endpoint, description?, source, enabled, allowedTools, disallowedTools, timeoutSecs, authConfigured, health?}` (see `src/server/ops/mcp.rs`). `McpServersView` reads `.servers` off the wrapper, stores `undefined` into `useState<McpServer[]>([])`, and throws later on render — so the crash surfaces away from the fetch that caused it.

`installMcpServer` / `connectMcpServer` / `disconnectMcpServer` / `uninstallMcpServer` / `callMcpTool` are in the same family; `/connect` and `/disconnect` routes do not exist at all. `ConnectionsView` already uses the correct `api/mcp.ts` surface.

## Why not just guard it

`?? []` would stop the crash and leave a settings page that lists nothing and whose buttons hit absent routes — a broken page that no longer says it is broken.

## Suggested fix

Port `McpServersView` onto `api/mcp.ts` (as `ConnectionsView` already is), then delete the phantom MCP methods from `client.ts` and the unused `lib/mcp.ts` types so the contradictory pair cannot be picked up again.

## Related, same class

[...]

Issue #302, opened 2026-08-03
nav: remove Inbox, Desk, Brains and Finances from the sidebar


oxoxDev:
Part of the MVP spec epic #183 (§2 Navigation / IA). Filed separately because it is the smallest independent slice.

## Work

**Keep:** Workspace, Team Desks (plus global items — account, settings).

**Remove from nav:** Inbox · Desk (the nav item; the Team Desks section stays untouched) · Brains · Finances.

Routes must not remain reachable by URL either — 404 or redirect.

## Two things to decide rather than assume

**Removing the nav item is not removing the capability.** "Brains" is the agent-memory surface, and the engine behind it shipped recently (in-pod vector recall). Dropping the nav entry should not delete the store, the routes, or the tests — only the way in. Be explicit in the PR about what was removed versus hidden.

**Finances is the accountability surface.** Per-agent spend enforcement is currently a manifest field (`budget_usd_daily`) that is **read by nothing** — zero call sites. Removing the money view while enforcement is also absent leaves that area with neither a view nor teeth. That may be the right MVP call, but it should be a decision, not a side effect. If the intent is "not in v1 nav" rather than "not a goal", say so in the PR so it is not read later as descoped.

## Acceptance criteria

- [ ] Sidebar shows only Workspace and Team Desks, plus required global items
- [ ] Removed routes 404 or redirect; no dead link reachable by typing a URL
- [ ] No backend route, store or test is deleted as part of this — nav only
- [ ] The PR states, per removed item, whether the capability is hidden or retired

Scope: **small**.

Issue #483, opened 2026-08-06
e2e: the unread-badge spec fails under concurrent lanes because the operator's own message renders twice


oxoxDev:
## What happens

`chat-live-events.spec.ts:130` — *"a reply to a channel you are not on leaves an unread badge, and opening it clears the badge"* — fails when several `Console E2E` jobs run at the same time:

```
Error: strict mode violation:
  getByText('You said: unread-<marker>') resolved to 2 elements
```

The operator's own message is rendered **twice**. The assertion is not at fault: a marker is unique per run, so two matches means two nodes carrying the same text.

## Evidence that it is load-sensitive, not caused by any change

Measured across four runs on 2026-08-06, all from the same base:

| Run | Concurrent E2E jobs | Result |
|---|---|---|
| `main` @ current tip | 1 | `Console E2E` **success** |
| PR A, ran 09:32 alone | 1 | 75 passed / 6 skipped, **0 failed** |
| PR B, started 09:42:09 | 3 | **1 failed** — this spec |
| PR C, started 09:42:10 | 3 | **1 failed** — this spec, identical |
| PR D, started 09:42:10 | 3 | **1 failed** — this spec, identical |

One of those three PRs changes **only Markdown documentation**. It cannot affect a browser test, which rules out the change under test as the cause in at least that case, and the failure is byte-identical across all three.

## Why this is a product race, not a test bug

The duplicate is the operator's own bubble. This codebase already treats that as a known failure mode — the very next test in the same file is *"a turn sent from the composer renders exactly one company bubble"*. The suspected mechanism is the synchronous POST response and the live event frame both minting a bubble for the same message; under load the window between them widens enough for both to land.

If that is right, the flake is **evidence of a real defect** rather than noise, and the spec is doing its job. Confirming the mechanism is the first task here — the alternative (a rail preview and the transcript both matching the text) would make it a genuine test-selector problem instead, and that distinction decides the fix.

## Why it matters now

Three open PRs are red on this, none of them responsible. A red check that nobody caused is worse than a failing test: the next person to see it learns that red does not mean broken, and that habit is expensive to unlearn.

## What the fix should do

1. Determine which of the two mechanisms produces the second node — dedupe at the source (a bubble keyed by message id, so a POST result and its event frame collapse) or a selector scoped to the transcript. Prefer the former if the duplicate is real; it is a user-visible defect in its own right.
[...]

PR #484, opened 2026-08-06
test(e2e): run the four environment-gated specs in a live-brain lane

### LINKED ISSUE #467: e2e: four specs need a feature-gated host and an inference backend, so a default lane must skip them
## What this is

Split out of #428, which puts the Playwright suite into CI. That lane builds the
**default** feature set, and four specs cannot pass against it — not because
they are wrong, but because the thing they test is not compiled in.

They are now skipped behind `PW_LIVE_BRAIN` (`test/e2e/capabilities.ts`), with
this issue named in the skip reason so none of them is bare. This issue is what
stops that skip becoming permanent.

## The four

| Spec | What it needs |
|------|---------------|
| `wiring.spec.ts` | the agent harness plus a mock inference backend echoing `__MOCK_LLM__` |
| `chat-to-card.spec.ts:83` | an inference backend whose tool choices are scripted (`SPAWNONE` → `spawn_task`) |
| `workflow-run-history.spec.ts:116` | the workflow runner, which lives behind `openhuman` |
| `mcp.spec.ts` | `PW_MCP_SERVER` pointing at the simple MCP server |

The first three are the suite's most valuable specs: they exercise the product
end to end rather than the console's own rendering. Leaving them skipped means
the CI lane proves the console renders, not that the company works.

## Why a host probe could not be used instead

`GET /tiny` looks like the right probe and is not. `RuntimeModuleStatus::all()`
reports `openhuman` as `enabled: true` unconditionally — the field describes the
vendored *checkout*, not the cargo feature — so a default build answers
identically to a feature-gated one. A probe built on it would look
authoritative while distinguishing nothing.

Worth fixing on its own: that field is misleading to anything that reads it.

A CI lane that builds `--features openhuman,tinycortex`, stands up the mock
inference backend `wiring.spec.ts`'s header describes, exports
`PW_LIVE_BRAIN=1`, and runs the suite with these four included. #428's own
step 3 calls this "the more valuable version", and it is.

## Related

Split from #428.

---

## Testing — required

**Show the numbers, before and after.** This issue is about a lane not doing what it claims, so the evidence is a measured test count from a real run — not an assertion that the configuration now looks right.

Before this is done:

- Paste the before/after test counts from the actual command each CI job runs.
- Link a CI run on the branch where the intended tests are visibly executing, with their count.
- Where the change unblocks browser specs, run them against a live host and report the result.

Anything with a user-facing surface is verified on that surface, not only in CI.

M3gA-Mind:
## Summary

Closes #467.

#428's lane builds the **default** feature set, so four of the suite's most
valuable specs skipped themselves through `test/e2e/capabilities.ts` rather than
running. This is the lane that runs them, and the two fixtures that make it
possible.

Three things had to exist, and none did:

* **A host with the harness in it.** `rust-gated` now also builds
  `--features openhuman,tinycortex,mcp --bin opencompany` and uploads it as
  `opencompany-gated-binary`. `mcp` on top of the job's own feature set because
  `mcp.spec.ts` needs the `mcp_call_tool` belt entry, which is gated on it; the
  extra step is cheap because that job already compiles `openhuman_core` with
  `openhuman_core/mcp` for its tool-belt contract test, so what recompiles is
  this crate plus its bin, not the vendored tree. Built in `rust-gated` rather
  than in the new job for the same reason #428 put the default binary in `rust`:
  a second submodule checkout, apt install and cold openhuman compile is minutes
  of wall clock for a thing that job has already very nearly built.
* **Something for that harness to think with.** `frontend/test/e2e/mock-brain.mjs`
  is the mocked inference backend `wiring.spec.ts`'s header has described since
  the day it was written, and which nobody had ever committed: an
  OpenAI-compatible chat-completions and embeddings endpoint with no model
  behind it. Ordinary turns answer with a `__MOCK_LLM__` marker; a prompt
  carrying `SPAWNONE` calls `spawn_task` once; one carrying
  `__MOCK_TOOL_CALL__ {…}` calls exactly the tool it names. "Once" is
  load-bearing — the harness resends the whole transcript on every turn, so a
  directive counts as served when a tool result (or an assistant turn carrying
  tool calls) appears *after* it. Otherwise a single `SPAWNONE` message would
  open a new card on every subsequent turn in that thread, forever.
* **A tool server to call.** `frontend/test/e2e/mcp-server.mjs`, JSON-RPC over
  HTTP, two tools, no network.

`playwright.config.ts` starts both as `webServer` entries when `PW_LIVE_BRAIN=1`
and it is managing the host, and hands the host its inference endpoint through
`PW_HOST_PASSTHROUGH` — the escape hatch `host.sh`'s `env -i` allowlist keeps
for exactly this. `npm run e2e:live` is the whole command; CI runs it as
`Console E2E (live brain)`, a separate job so the fast default gate is not put
behind the gated compile, and so the product stays covered both with and without
these features.

### MCP: the console half landed upstream, the agent half is here

`mcp.spec.ts` had drifted three ways while nothing ran it, and each alone would
have failed it: it installed a **stdio** server at a host that rejects any
declaration carrying a `command`; it drove `#/mcp` and `McpServersView`'s form,
which posts `{transport, command, args, url}` at a route reading
`{name, endpoint}`; and it asked for upstream OpenHuman's
`mcp_registry_tool_call` rather than the `mcp_call_tool` this host's belt
carries.

While this PR was in flight, #414/#480 found the same divergence from the other
end and fixed the console page, rewriting that spec to cover the operator's side
against a default-feature host. **This branch takes their version wholesale** and
keeps only the half no console page can reach: an *agent* calling a tool over the
real transport, which needs the harness and therefore this lane. That is now
`mcp-agent.spec.ts`, and the split matches the lanes — console on the default
one, agent on the live one.

`PW_MCP_SERVER` comes back for it, carrying a **URL** rather than the stdio path
#414 rightly retired. `capabilities.ts` says why in place.

## API Or Behavior Changes

None in the product. Test-infrastructure contract changes:

* `PW_MCP_SERVER` returns, carrying an MCP server **URL** rather than the stdio
  script path #414 retired (that meaning could never have been installed by this
  host). Consumed by `mcp-agent.spec.ts` only.
* New: `PW_MOCK_BRAIN_BIND` (default `127.0.0.1:8099`) and `PW_MCP_FIXTURE_BIND`
  (default `127.0.0.1:8098`).
* New npm script `e2e:live`.
* New CI job `Console E2E (live brain)`. It is not `continue-on-error`; a lane
  allowed to be red is the state #428 was filed about.

## Tests

- [x] `cargo fmt --all -- --check` — N/A - full local build matrix is prohibited on this machine; verified in CI. (No Rust changed by this PR.)

Run locally in `frontend/`:

* `npx vitest run test/unit/mock-brain.test.ts` — **7 passed**. A new file, and
  the only new logic in this PR with branches worth naming. It spawns the real
  server on an ephemeral port and drives it over HTTP, because what has to be
  right about it is the shape it puts on the wire: `function.arguments` as a
  JSON *string*, `finish_reason` a sibling of `message`. It covers the plain
  reply, the non-echo of the operator's prompt, both directives (including the
  nested-object payload the brace scanner has to get right), serve-once, a fresh
  directive after a served one, and the 1024-wide embeddings the host validates.
* `npm run typecheck:e2e`, `npm run typecheck:unit` — both clean.
* `npx playwright test --list` with and without `PW_LIVE_BRAIN=1` — config loads
  in both modes, 80 tests in 23 files.
* Both fixtures smoke-tested by hand (`initialize` / `tools/list` / `tools/call`
  / `healthz`).

Not run locally: the suite itself, which needs a gated host, and building one is
prohibited on this machine. The measured numbers below come from CI.

### The numbers this issue asks for

All from [CI run 31110082447](https://github.com/tinyhumansai/opencompany/actions/runs/31110082447)
on this branch, which is green on all six checks.

**Before** — `npm run e2e`, the only lane that existed, against a default-feature
host. Four specs skip themselves because the thing they test is not compiled in:

```
7 skipped
89 passed (2.6m)
```

**After** — `npm run e2e:live`, the new lane, same suite against
`--features openhuman,tinycortex,mcp` with both fixtures behind it:

```
6 skipped
90 passed (2.8m)
```

The counts move by less than the work does, and that is the point: the same
suite runs twice against two different products, and the specs that skip are
different ones. What #467 asked for is that these four **execute**, and they do
([job log](https://github.com/tinyhumansai/opencompany/actions/runs/31110082447/job/92650397759)):

```
✓  42 chat-to-card.spec.ts:52   any message on a desk thread can be added to the board (5.9s)
✓  43 chat-to-card.spec.ts:85   a card the orchestrator opens is chipped in chat, and survives a reload (6.0s)
✓  49 mcp-agent.spec.ts:86      an agent calls a tool on a registered MCP server and shows the result (6.2s)
✓  69 wiring.spec.ts:44         operator console renders a mocked backend reply end to end (388ms)
✓  94 workflow-run-history:118  running a workflow adds it to the durable history and it survives a reload (11.1s)
```

Every one of those was skipped on `main` before this PR. The six that skip in the
live lane are the inverse gates that arrived with it — three in
`chat-live-events.spec.ts` (they assert the offline brain's `You said:` reply),
the Planning drag from #506, and two parked on their own issues (#302, #414).

The fixtures are visible in the same log, which is how the round trip is
evidenced rather than asserted:

```
[mock brain] listening on http://127.0.0.1:8099
[mcp fixture] listening on http://127.0.0.1:8098/mcp
[mock brain] tool call: spawn_task <spawn:SPAWNONE 1786…>
[mcp fixture] tools/call echo
```

**Browser verification**: the lane *is* the browser run — Chromium against a
live host, on every push. There is no separate surface to check by hand here;
what this PR adds has no UI of its own.

## Documentation

* `frontend/README.md` — the "four specs a default-feature host cannot run"
  section rewritten: what each proves, how to run them, what each fixture is.
* `frontend/test/e2e/capabilities.ts` — the skips now name the lane that runs
  them, and `MCP_SERVER` documents why it became a URL.
* `frontend/test/e2e/host.sh` — its "no binary" message now names the gated
  build and `npm run e2e:live`.
* `playwright.config.ts`, `wiring.spec.ts` — headers updated to say where the
  mocked backend comes from.

## Summary by CodeRabbit

* **New Features**
  * Added live-agent end-to-end coverage for chat, task spawning, tool calls, and HTTP MCP integrations.
  * Added support for testing agents with mock inference and MCP services.
  * Added an `e2e:live` command for running live-agent browser tests locally.

* **Documentation**
  * Expanded E2E setup guidance, feature requirements, environment configuration, and troubleshooting information.

* **Tests**
  * Added validation for mock inference responses, embeddings, tool execution, and MCP communication.
  * CI now runs the live-agent suite and uploads diagnostic artifacts when tests fail.


M3gA-Mind:
@coderabbitai review

oxoxDev (commented):
The fixtures are the right idea and most of the lane works — 74 passed, and the specs this issue exists to unblock are running against a real harness for the first time. The four failures are all one design decision, though, and it is worth fixing before this lands.

**The live-brain lane runs the whole suite, not the specs that need a live brain.**

Three of the four failures are `chat-live-events.spec.ts` (`:113`, `:130`, `:150`). Those specs were written against the **offline echo brain** and assert on its reply format — `You said: <marker>`. In this lane the brain is `mock-brain.mjs`, which by explicit design does not echo the prompt:

> *"The plain reply deliberately does NOT echo the operator's prompt back. A spec … would find two matches if the reply quoted it."*

That reasoning is sound. But it means every spec that depends on the echo format cannot pass in this lane, and running them here asserts a contract the fixture deliberately does not honour. They already pass in the default lane, so nothing is gained by running them twice with one of the runs guaranteed red.

The fourth is the same shape from the other side. `mcp.spec.ts:135` expects the reply to carry `echo: <marker>`, and the locator resolved to **14** elements, all reading `__MOCK_LLM__ mock inference backend reply.` — so `.last()` picked a plain reply, not one carrying a tool result. Either the tool-call branch did not fire, or the reply carrying the result is not the last match. Worth determining which before adjusting the assertion: if the tool call did not happen, the spec is correctly reporting a real gap in the fixture, and loosening the locator would hide it.

**Suggested shape:** make the live-brain lane run exactly the specs that need it — the four named in #467 — rather than the full suite. That is also what the issue asked for. A `--grep`, a project filter, or a directory split all work; whichever reads best.

Two smaller notes:

- The three `chat-live-events` specs are separately unstable under concurrency for an unrelated reason — see #483, where the agent's reply renders twice when a channel receives it while closed. Do not fold that in here; different cause, different fix.
- `mcp.spec.ts` asserting on the tool result is the most valuable assertion in this PR, since it is the only one proving the full remote-tool round trip. Keep it strict.

M3gA-Mind:
Pushed 534d8f5 after the first CI run of the new lane, which is exactly what the lane is for.

**Run 1 of `Console E2E (live brain)`: 74 passed, 4 failed, 2 skipped (4.8m).** All three specs #467 is about — `wiring.spec.ts`, `chat-to-card.spec.ts`'s card chip, `workflow-run-history.spec.ts`'s durable history — passed on the first attempt, as did the whole console half of the rewritten `mcp.spec.ts` (install an HTTP MCP server, live-probe it, list its tools). The four failures were two real defects in what I had written:

1. **A directive fired on every model call of a turn, not once.** A tool result is not always a `tool` message: OpenHuman's dispatcher (`to_provider_messages`) renders one as a **user** message reading `[Tool results]\n<tool_result id="…">…`, so `alreadyServed` saw nothing after the directive and re-emitted the call the moment the tool returned. The fixture log shows `spawn_task` four times in 60ms for one message, and it is also why the MCP agent test got the generic reply instead of the tool's output. `isToolOutput` now recognises both shapes, and the reply following a tool quotes its output with the wrapper stripped.
2. **The three `chat-live-events` specs find their own reply by the offline brain's `You said: <text>`** — which is the only way they can prove an SSE frame carried the answer to the message just sent rather than some other one. Rather than skipping them in this lane, the mock's plain reply now mirrors that shape and adds the marker, so one spec holds against both brains. It is safe against the specs that locate the operator's own bubble by its text: they match exactly, and the reply is never exactly the prompt.

Plus one locator fix: `mcp.spec.ts` resolved `getByText(/__MOCK_LLM__/).last()` to the chat rail's one-line thread preview rather than a transcript row.

Unit coverage added for the dispatcher's user-message tool-result shape, since that is the one that was wrong. 8/8 locally.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Pushed 6408268 after run 2 (default lane 72 passed / 1 failed / 7 skipped; live lane 74 / 4 / 2). Four findings, three mine:

1. **A directive still fired on every model call of a turn.** The history check I added in 534d8f5 is right but cannot be sufficient: `spawn_task` is serviced by the runtime's delegation seam, not the agent's own tool loop, so its result never enters the model-visible transcript and the history looks untouched on the next call. The fixture log shows `spawn_task` four times in 60ms in both runs. The mock now also remembers directive *identities* it has served — every directive a spec writes carries a `Date.now()` marker, so a genuinely new one is always a new key, and "once" holds whatever the transcript looks like.
2. **`You said:` quoted the memory preamble rather than the message.** `memory_loop::inject` wraps the operator's text as `## Relevant prior work … ## Task <text>`, so the reply read `You said: ## Relevant prior work…` and the three `chat-live-events` specs could not find themselves in it. Now takes the task section.
3. **`mcp.spec.ts` landed on `/#/conversation` without selecting a thread.** A composer is present either way so the fill succeeded, but the reply landed in a transcript the page was not showing — the only trace on screen was the rail's one-line preview, which is exactly what the assertion kept resolving to. It now selects the thread the way `chat-to-card.spec.ts` does. (The MCP round trip itself is working: run 2's fixture log shows `tool call: mcp_call_tool` → `tools/call echo` → a 42-character reply, which is `__MOCK_LLM__ echo: agent-mcp-…` exactly.)

And one that is **not** mine, which this lane surfaced:

4. `chat-live-events.spec.ts` matched `You said: <marker>` across the whole page, so it resolves to **two** elements — the transcript bubble and the rail's preview of the same message — as soon as the preview catches up. That failed the **default** lane on this run, on a commit that changes nothing it touches, having passed the run before. It is a latent flake in a required check, so I fixed it here rather than leave it: the three assertions now scope to a transcript bubble. Happy to split it into its own PR if you would rather keep this one to the lane.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Pushed 972a928. Run 4: **73 passed, 2 failed, 5 skipped** (default lane green).

The log line I added last push named the four-cards cause outright, which is exactly what it was for:

```
spawn_task <spawn:SPAWNONE 1786015999106>
spawn_task <spawn:SPAWNONE 1786015999106". It will be opened on the board this turn.>
spawn_task <spawn:SPAWNONE 1786015999106". It will be op...". It will be opened on the board…>
```

**The fixture was handing the directive back to itself.** It titled the card after the line the directive sits on, and the runtime reports a spawned card into the next prompt as `A card titled "<title>". It will be opened on the board this turn.` — so `SPAWNONE` went back in front of the model, re-wrapped and re-truncated each round, which is also why no identity key ever matched. Nothing the fixture writes may contain a directive; the title now has it stripped, and there is a unit test that fails if that regresses.

**`mcp.spec.ts` was cancelling its own turn.** A turn runs inside the request that started it and the host drops the work when the client disconnects, so the reload I added last push fired while the send was still in flight. Evidence rather than theory: run 4's fixture log carries **no model call at all** for that message, where run 3's carried the full `mcp_call_tool` → `tools/call echo` round trip. The POST is now awaited explicitly before the reload.

Still open, and I want to flag it rather than quietly absorb it: **`board-columns.spec.ts:104` ("dragging into Planning moves the card without dispatching it") has failed the live lane twice now** — expected `planning`, got `todo` — while passing the default lane on every run. It is not caused by anything in this PR's diff, and the difference between the lanes is that a live harness is doing real work in the background during the drag. Either a flaky drag that only loses the race when the host is busy, or something in the runtime touching the card. I have not chased it into this PR's scope; if it survives the next run I will file it separately with the evidence.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Pushed b572c5c. Run 5: **73 passed, 2 failed, 5 skipped**, and both remaining failures are now understood.

**`spawn_task` fired exactly once** — the title fix landed. And the MCP round trip completed end to end and survived the reload: the reloaded transcript carried the row. What it did not carry was the marker, spelled the way I expected:

```
Received string: "11:57 AMMOCK_LLM echo: agent-mcp-1786017420793"
```

A bubble renders its text as **markdown**, and the marker's own underscores are emphasis syntax — so `__MOCK_LLM__` reaches the DOM as `MOCK_LLM` inside a `<strong>`. Only the plain-text surfaces carry it verbatim: the rail's one-line thread preview, an API response. That explains every earlier symptom in one go, including why an unscoped `getByText(/__MOCK_LLM__/)` kept resolving to the sidebar — it is the only place the literal string exists.

**Worth knowing, and not something I changed here:** `wiring.spec.ts` asserts `page.getByText("__MOCK_LLM__").first()`, so by the same reasoning it can only be matching the rail preview, never the reply bubble. It still proves the chain — the preview only exists because the reply came back from the host — but it is a weaker assertion than it reads as, and tightening someone else's spec is not this PR's job. Flagging it for a follow-up.

**`board-columns.spec.ts:104` is now skipped in this lane only, on #501**, which I filed with the evidence. Three consecutive live-brain failures, every default-lane run in the same CI runs green. The drop is dispatched as three separate events with the dragged card's id in React state between them, and a live host pushes journal events that re-render the board; the card never moves at all rather than moving slowly. I filed it rather than absorbing it because if the board loses an in-flight drag whenever an agent turn refreshes it, that is an operator-facing bug and not only the spec's problem. The skip names the issue, per the rule this lane exists to enforce.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
**Green.** All six checks pass on [run 31110082447](https://github.com/tinyhumansai/opencompany/actions/runs/31110082447), and the PR body now carries the measured before/after this issue asked for.

Two things happened since the last comment.

**No CI run ever fired for `b572c5c`.** The PR head was right and Actions was busy with other PRs, but no workflow was created for that SHA. Re-triggered by the merge below.

**Merged `main`, which had moved 20+ commits and overlapped this work in three places.**

* **#506 corrects my #501, and I was wrong.** I read the live-lane board failure as a drag lost to a mid-drag re-render. The artifact says otherwise: the card carries `[system] planning could not read the model's answer as a plan, so nothing was written`, which is `harness/planning.rs` — so the drop reached the host, the card entered Planning, and the planning station settled it back. Correct behaviour on a planner-attached host. #501 is closed with that correction; this branch drops my skip and takes #506's spec wholesale.
* **#414/#480 fixed the console MCP page** from the other end while this was in flight. I took their `mcp.spec.ts` entirely and moved my agent-half coverage to a new `mcp-agent.spec.ts` — the part no console page can reach. The split now matches the lanes.
* **#502/#481 changed `ci.yml`.** Resolved by hand: upstream's `Test (tinyplace)` first, my gated binary build and upload last so the artifact carries the feature set the e2e lane needs.

Merged rather than rebased (six commits against three conflicting files); main squash-merges, so the branch history does not reach it. `board-columns.spec.ts` and `mcp.spec.ts` no longer differ from upstream at all — the diff is 12 files.

Ready for review.

M3gA-Mind:
Both CodeRabbit findings fixed in 2478408 — both were real.

**Test-owned MCP registration (Major).** Correct and worth having: a fixed server name plus a tolerated 409 let this spec adopt a leftover registration from an earlier run, pointing anywhere at all, pass against it, and then delete a server it never created. The name is unique per run now and a 409 is a setup failure. I checked the cleanup too, but only when the body got that far — an assertion thrown out of a `finally` replaces the real failure with a complaint about cleanup, which is the harder of the two to debug.

**README working directory (Minor).** The recipe ran `cargo build` from the repository root and `npm run e2e:live` from wherever you happened to be. Fixed with an explicit `cd frontend` and a note on which line runs where.

@coderabbitai review

M3gA-Mind:
This merged (83bd8ba) while two CodeRabbit fixes were still in flight, so they did not land with it. They are now #510:

* **Major** — `mcp-agent.spec.ts`'s server registration was not test-owned: a fixed name plus a tolerated `409` let it adopt a leftover registration from an earlier run, pointing anywhere, pass against it, and delete a server it never created.
* **Minor** — the README's live-lane recipe did not say that `npm run e2e:live` runs from `frontend/`.

Nothing else from the review is outstanding.

Issue #415, opened 2026-08-05
workflows: the copilot can describe a change but cannot propose one you apply


oxoxDev:
## What is missing

The per-workflow copilot shipped in #303 answers questions about a workflow from its graph and its recorded runs. It cannot propose an edit that the operator reviews and applies.

Acceptance criterion 4 of #303 asked for propose-then-apply and was deliberately deferred so the read-only half could ship and be verified on its own. This issue is that deferral, recorded rather than left as a comment on a merged PR.

## Why it is worth doing

Answering "why did this node fail" and then requiring the operator to translate the answer into canvas edits by hand is the gap between explaining a workflow and improving one. The copilot already holds the two things a proposal needs — the graph and the run history that shows what actually happened — so the reasoning is done and only the round trip is missing.

## What the fix should do

- The copilot returns a **proposed graph change**, not prose describing one.
- The operator sees the change **before** it applies: a diff against the current graph, not an accepted mutation with an undo.
- Applying goes through the same validation and persistence path as a canvas edit, so a proposal cannot produce a graph the editor could not have produced.
- A proposal made against one version of a graph that has since changed must be refused or rebased, never silently applied to a graph it was not reasoned about.

## Design questions to settle first

- **What is the unit of a proposal?** A whole replacement graph is simple to validate and hostile to review. A node-level or edge-level patch reviews well and needs a merge story when the graph moved underneath it.
- **Where does validation live?** The console can reject an obviously malformed proposal, but the host has to be the authority, or a proposal becomes a second way to write a graph that bypasses the first one's rules.
- **What happens to a proposal the operator rejects?** Discarded, or kept as part of the conversation so a later question can refer to it.

## Related

- #303 — the copilot, acceptance criterion 4
- PR #405 — shipped the read-only half

---

## Testing — required, and this one is bound to the UI

**This issue has a user-facing surface, so it is verified on that surface.** A green unit or component test is not evidence that this is fixed, and does not close it.

Before this is done:

- Run the console against a live host and walk the exact path described above, by hand, in a browser.
- Exercise the **failure** this issue describes, not only the happy path — the fix is that the described behaviour is gone, so the old behaviour is what you have to try to reproduce.
- If a Playwright spec covers it, that spec must actually run and pass in a lane that builds it. A suite that is compiled out of CI proves nothing (see #475).
- State in the PR what you ran it against and what you saw. "Tested" on its own is not a result.

[...]

Issue #457, opened 2026-08-06
approvals: a standing grant records a tool, so a Composio grant is broader than the sentence the operator agreed to


oxoxDev:
## What this is

Follow-up from the #441 / #443 / #444 lane, split out deliberately rather than half-done.

#441 asks that a grant be "narrow enough to be honest": *"'Composio, for a week' is too broad to mean anything; 'this teammate may read from GitHub for eight hours' is a sentence an operator can actually consent to."*

That lane got half of it. A standing grant on `composio_execute` now admits **reads only** — the policy re-classifies the live action slug at redemption, so a scope granted while listing pull requests cannot admit an outgoing email. What it does not do is scope the grant to a **toolkit**. The operator consented to "read from GitHub"; what they were given is "make Composio reads", which also covers Gmail, Slack and every other connected account.

Reads, never sends. But still broader than the sentence.

## Why it was not done in that lane

`StandingGrant` records `agent`, `tool` and an expiry, and matching is `(agent, tool, unexpired)`. Expressing "GitHub reads" needs a scope field on the record — and adding a field means touching every struct literal, two of which live in `src/harness/brain.rs`, which was owned by a concurrent lane at the time. Encoding the toolkit inside the existing `tool` string (`composio_execute:github`) was considered and rejected: it puts structure in a display string the console renders verbatim.

## What the fix should do

- Give `StandingGrant` an optional scope the match consults — a toolkit for a Composio grant, `None` for every other tool, `#[serde(default)]` so replayed lines keep working.
- Mint it from the action slug on the parked effect, so the grant records what the card showed.
- Surface it in the standing-permissions list, so "read from GitHub, expires 18:00" is what an operator revokes rather than "composio_execute".
- A grant with no scope keeps today's behaviour, so nothing already minted changes meaning.

## How to verify

- Granting a scope while reading a GitHub repository does not admit a Gmail read.
- It still admits a *different* GitHub read — the argument-exactness complaint in #441's comment must not come back.
- A grant replayed from a journal line written before the field existed still admits what it did.
- The standing-permissions list names the provider.

## Related

Sibling of #441. The read/send half shipped with it; this is the toolkit half.

Issue #453, opened 2026-08-06
orchestrator: review_task reports a card already moved to done, for work that is only queued


oxoxDev:
## What happens

`review_task` pushes a `Delegation::ReviewTask` onto the delegation queue and immediately returns, in the past tense:

> `Approved card {task_id}; it is complete and has moved to done.`

Nothing has moved. The card moves only if that queued delegation is later drained and run. The `Revise` arm has the same shape — *"Sent card {task_id} back to To-do"* — for a delegation that has not been sent anywhere yet.

## Why it is the same defect as #445 and #419

Three tools now share one shape: **stage the work, return an unqualified success naming a completed outcome, and depend on a drain that may not happen.**

- #445 — `publish` promised an Artifacts tab the caller could not reach.
- #419 — `spawn_task` / `delegate_to_desk` promise a card, and the drain destroys everything past a cap of 3.
- This — `review_task` promises a completed state transition for a queued intent.

The consequence is identical and it is what makes this class worth naming: the agent has no way to discover it was told something false, so it repeats the claim to the operator with full confidence. The operator's own record of the turn is the agent's wrong summary.

This one is compounded by #419: `review_task` shares the delegation queue, whose drain takes `len().min(MAX_DELEGATIONS_PER_TURN)` and clears the rest. A review approved as the fourth delegation in a turn is reported as done and then discarded.

## Also undrained on two paths

The delegation queue is not drained on the approval re-dispatch path, nor by the workflow agent node — both of which can run this tool. On those paths the receipt is wrong every time, not just past a cap.

## What the fix should do

- The receipt describes what actually happened: the review was **recorded and will be applied**, or the tool waits and reports the real outcome.
- No success message asserts a state transition that has not occurred.
- Paths that can run this tool either drain the queue or refuse the tool, in the same fail-closed direction #445 took — a turn that cannot deliver should not be able to promise.

## How to verify

- Approving a card reports a completed move only once the card has moved.
- A review queued past the per-turn cap is not reported as done.
- A review issued from the approval re-dispatch path and from a workflow agent node either takes effect or says it did not.

## Related

Found while fixing #445; the files belong to concurrent work on #442, so it was reported rather than fixed. Same family as #419.

Issue #449, opened 2026-08-06
harness: the agent-workspace mkdir is retried every dispatch, and a broken volume floods the log


oxoxDev:
## What happens

`src/harness/mod.rs` calls `build::ensure_agent_workspace` **on every dispatch**, and on failure logs at `error!` and continues. There is no memo, no backoff and no dedup, so a volume that cannot be written produces one error line per turn, indefinitely.

## Why it is not simply a bug

The non-fatal choice is deliberate and the comment explains it well: an agent with no file grant runs a perfectly good turn without this directory, and the `error!` level (rather than `warn!`) is chosen so the line is greppable next to the refusal it explains — a relative write later fails with a message that reads as a workspace escape when the real cause is the missing directory.

That reasoning holds. The gap is the **repetition**, not the level or the tolerance.

## Why it matters

This is the same shape as #446, one layer up: an unbounded identical error on a hot path. #446's harm was not the failed write, it was that the repeat filled a tenant's log and buried every other signal, so an unrelated investigation found one line repeated 200 times and nothing else.

The condition here cannot resolve itself mid-life either — if the mount is broken, every subsequent attempt fails the same way.

## What the fix should do

- Attempt once per (company, agent) and remember the outcome, rather than per dispatch.
- Keep the error visible and keep the turn running — both current behaviours are right.
- Ensure a broken volume produces a bounded number of lines, not one per turn.

## How to verify

- With an unwritable workspace root, a company running many turns emits the error a bounded number of times, not once per turn.
- The misdirecting-guard explanation is still reachable in the log when a relative write is refused.
- A healthy company is unaffected.

## Provenance

Found in the sibling sweep for #446. Not fixed there because `src/harness/mod.rs` was owned by concurrent work.

oxoxDev:
Fixed and on `main`.

Merged in PR #494 (`91611d66`). GitHub did not auto-close this one — a `Closes #a, #b, #c` list only closes the first entry — so closing by hand after verifying against `upstream/main` rather than trusting the PR body.

Verified: the failing-key set and its transition helper are in `src/harness/mod.rs`, so the first failure logs, repeats stay silent, and a recovery logs once.

Note the approach deliberately differs from the one this issue proposed. Memoising the *attempt* was rejected because it regresses in both directions — memoising success hides a later data-dir wipe, memoising failure means a late-mounting volume never recovers. The attempt still runs every dispatch; only the logging is edge-triggered.

PR #487, opened 2026-08-06
feat(workflows): confine the copilot's answering agent to one workflow

### LINKED ISSUE #416: workflows: the copilot's answers are grounded in one workflow, but the teammate answering is not confined to it
## The current behaviour, stated accurately

The per-workflow copilot (#303) sends its question to the company orchestrator on a dedicated conversation thread. Two properties hold and are verified:

- **Grounding** — the question carries that workflow's graph and its recorded runs, so the answer is about that workflow.
- **Transcript isolation** — the exchange stays on its own thread and does not appear in the company chat.

One property does **not** hold: the teammate answering is the company orchestrator, with the company's full context and tool surface. The thread selects the responder and the journal thread; it does not narrow what the responder can reach.

This is a scoping limit, not an authorization gap. The route authenticates before anything reaches the orchestrator, and an operator who can open the copilot can already address the same orchestrator with the same context from the Chat tab, from the chat endpoint directly, or through the workflow routes. The copilot therefore grants no capability its caller does not already have. PR #405 corrected the panel's own wording, which had claimed confinement it did not provide.

## What this issue proposes

A genuinely workflow-confined turn: an agent whose context and tool grants are narrowed to the workflow being discussed, so a copilot answer cannot draw on company material the operator did not ask about.

This is a **new capability**, not a restored check. It needs turn-level plumbing — a way to run an agent with a reduced grant set and a reduced context — which does not exist today and which the read-only copilot deliberately avoided introducing.

## Why it is worth having anyway

- It makes the panel's natural reading true, rather than requiring the wording to explain a limitation.
- A confined turn is the precondition for anything that lets a workflow-scoped assistant act rather than answer, including the proposal flow in the sibling issue.
- The same primitive would serve other places where an agent should reason about one object without the rest of the company in scope.

## Design questions to settle first

- **What does "confined" include?** Context only, tool grants only, or both. Narrowing context without narrowing tools leaves a turn that can fetch what it was not given.
- **Does the confinement live in the turn or in a distinct agent?** A per-workflow agent is easier to reason about and heavier to provision; a confined turn on an existing agent is cheaper and needs the grant reduction to be enforced host-side, not requested politely.
- **What does the copilot say when a question genuinely needs company-wide context?** Refusing is honest and unhelpful; answering unconfined defeats the purpose. The answer is probably to say which it did.

## Related

- #303 / PR #405 — the copilot as shipped, and the wording correction
- #415 — propose-then-apply, which would depend on this

---

## Testing — required, and this one is bound to the UI

**This issue has a user-facing surface, so it is verified on that surface.** A green unit or component test is not evidence that this is fixed, and does not close it.

Before this is done:

- Run the console against a live host and walk the exact path described above, by hand, in a browser.
- Exercise the **failure** this issue describes, not only the happy path — the fix is that the described behaviour is gone, so the old behaviour is what you have to try to reproduce.
- If a Playwright spec covers it, that spec must actually run and pass in a lane that builds it. A suite that is compiled out of CI proves nothing (see #475).
- State in the PR what you ran it against and what you saw. "Tested" on its own is not a result.

Backend-only changes may be verified from the backend. Anything with a screen is verified on the screen.

M3gA-Mind:
## Summary

Fixes #416. The per-workflow copilot's answers were grounded in one workflow; the agent answering them was not confined to it. The thread id (`workflow-copilot:<id>`) selected the responder and the journal thread and narrowed nothing else, so the responder was the company orchestrator, holding whole-company context and its full tool surface. "Answer only about this workflow" was a sentence in the composed prompt, which is advice.

This adds the missing primitive: a **confined turn**, enforced host-side.

### The new boundary, stated

A `/chat` turn addressed to `workflow-copilot:<workflowId>` now runs on an ephemeral agent built for that turn and dropped after it:

- **No tools.** Its toolbelt is empty. That includes the intrinsic `memory_store` / `memory_recall` every roster agent carries and the orchestrator's `query_company` / `spawn_task` / `delegate_to_desk`. It is deliberately **not** `build_agent` with empty grants, because that path wires the memory tools (and, under `mcp`, the MCP registry tools) onto every agent regardless of grants.
- **Enforced, not merely absent.** `ConfinedToolPolicy` denies every tool call by name, whatever the belt holds, with a refusal that names the workflow and tells the model to say what it would have needed. An empty belt is the ordinary case; the policy is what makes a tool wired here by a later change, or a name the model invents, a refusal rather than a call.
- **No company memory.** Its `Memory` is backed by `ConfinedContext`, a store that accepts a write, keeps nothing, and answers every read empty. The pool additionally skips retrieve→inject and the memory writeback for a confined turn, so no prior task outcome is prepended to the message and the exchange leaves nothing behind for a later company turn to retrieve.
- **No delegation.** The brain runs this turn directly instead of through the delegation runner, so it cannot hand off to a desk and have the desk do what it was not allowed to.
- **No board card.** `run_chat` no longer opens a `todo` card from a copilot message. This half is in the **default** build, not behind `openhuman`: a copilot question is phrased at the graph ("add a node that emails the report"), and the deterministic intent detector was reading that as a request to the company.
- **Not on the roster.** The confined agent id (`workflow-copilot`) is not addressable: dispatching or delegating to it is an unknown-agent error, exactly like any other non-teammate name.

### What a copilot turn can still reach

Exactly what the console put in the message, and nothing else:

- the workflow's name, id, description and editability;
- its nodes and edges, with each node's agent, schedule, approval flag, error/retry policy, destination and config;
- summaries of **that workflow's own** recorded runs (already a server-scoped read);
- the operator's question, and the earlier turns of this same thread that the provider carries.

It still cannot read the board, the roster, another workflow, the workspace tree, company memory, an MCP server, the web or a file, and it still cannot edit the workflow.

**When a question needs the wider company**, the issue's own design question, it answers what it can and says which part it could not answer here, pointing at the company chat. That is in both the host-side persona and the composed message, so what the panel discloses is built from the same text that was actually sent.

**This is a scoping boundary, not an authorization check, and must not be read as one.** `/chat` is already authenticated and company-scoped: an operator who can open the copilot can address the same orchestrator from the Chat tab. What changes is what one turn reaches, which is what makes #415 (propose-then-apply) safe to build on top — a copilot that can propose applyable edits while unconfined is worse than an unconfined answer.

### Where it lives

- `src/company/copilot.rs` (new, un-gated) — the thread convention. One function decides "this is a copilot turn"; both the harness and the chat route key off it.
- `src/harness/confine.rs` (new, `openhuman`) — `Confinement`, `ConfinedToolPolicy`, `ConfinedContext`, the persona, and the ephemeral agent builder.
- `HarnessPool::run_confined` — the turn itself. The plan-level total-token ceiling was extracted from `run_inner` into `total_ceiling_refusal` so both paths are gated by the *same* rule rather than by two copies: a turn that reaches nothing still spends model tokens.
- `HarnessBrain` — the branch in the `OperatorMessage` arm.
- Console: the panel's disclosure and the composed message state the boundary; the "asking for a change may open a card" line is gone, because it no longer happens.

## API Or Behavior Changes

- `POST {scope}/chat` with `chat: "workflow-copilot:<id>"` is answered by a confined agent instead of the orchestrator, and no longer opens a board card. Every other thread is byte-identical to before.
- No route, request or response shape changes. A host predating this answers the same requests the old way, so the console needs no capability check.
- New public surface: `company::copilot::{COPILOT_THREAD_PREFIX, workflow_of_thread, is_copilot_thread}`; `harness::confine::*`; `HarnessPool::run_confined`.
- The confined turn is metered and billed like any other turn (usage sample + ledger entry under the `workflow-copilot` agent id) and streams onto the same thread, so it stays visible in Usage rather than becoming an unattributed spend.

## Tests

Focused tests with the behaviour change, in the lanes CI runs:

- `src/company/copilot.rs` — the thread convention: reads the workflow id out, refuses a prefix that does not start the id, refuses a prefix naming no workflow, trims. Default build.
- `src/harness/confine.rs` — every tool denied (including `memory_recall`, `query_company`, `spawn_task`, `delegate_to_desk`, `mcp_registry_tool_call`, and a name no belt has ever carried); the refusal names the tool, the workflow and the way out; the persona states the boundary and the edge case; `ConfinedContext` stores nothing readable. `openhuman` lane.
- `src/harness/mod.rs` — `a_confined_turn_neither_reads_nor_writes_company_memory`, with the control in the same test: the same question on the ordinary roster path pulls a seeded company chunk into the prompt (visible in the mock provider's echo) and writes the outcome back; the confined path does neither, from the same store. Plus `the_confined_agent_is_not_addressable`.
- `src/server/ops/write_test.rs` — `a_copilot_thread_question_opens_no_board_card`, with the control: the identical actionable sentence on the ordinary thread still opens exactly one card. Default build, over the real HTTP stack.
- `frontend/test/unit/workflow-copilot.test.ts` — the thread id is the exact string the host confines on, and the composed message carries the boundary and the "say which part you cannot answer" instruction the panel's disclosure is built from.

Console gates run locally: `npm run typecheck`, `npm run typecheck:unit`, `npm run typecheck:e2e` clean; `npm test` passes apart from `tour-resume.test.ts`, which fails identically on unmodified `upstream/main` on this machine (local Node 25 jsdom `localStorage`; CI runs Node 22) and is untouched here. `rustfmt --edition 2024 --check` clean on every Rust file changed (`cargo fmt` itself cannot run without the vendored submodules).

**Browser walk** (`#/workflows` → Copilot, real Chromium against a stub host serving the console's documented shapes, because building the Rust host is prohibited on this machine):

- the panel's pre-question disclosure now reads "This turn runs confined to this workflow: no tools, no company memory, and no reach into the board, your teammates or another workflow. Ask something that needs the wider company and it will say so rather than guess.", and the old "asking for a change may open a card on the board" line is gone;
- asking "what is the rest of the team working on?" sends on thread `workflow-copilot:weekly_report` — the exact id `company::copilot::workflow_of_thread` parses, which is the console-side half of the trigger;
- the message actually sent begins with the boundary lines ("You have no tools and no access to the rest of the company…", "say which part you cannot answer here and that it has to be asked in the company chat");
- no uncaught page errors through the walk.

**What was not run here, stated plainly.** The host-side half (a confined turn refusing a tool call on a live `--features openhuman` host) was not exercised end to end on this machine: the local build matrix is prohibited, and the e2e lane's host is the default build whose echo brain the copilot panel deliberately refuses to send to. The boundary is covered by the Rust tests above in both CI lanes; no Playwright spec is added, because one would have to bypass the panel to reach the behaviour and would then be asserting what the HTTP test already asserts.

## Documentation

`docs/spec/runtime/api.md` — the `/chat` thread-addressing contract said thread addressing isolates transcripts and does not scope authority. It now records the exception: the copilot thread makes the turn confined, in two places that hold independently (the harness, and the card suppression in the default build), with links to `company::copilot` and `harness::confine`, and keeps the warning that confinement narrows a turn rather than granting or checking authority.

Module docs are in the code: `src/harness/confine.rs` opens with what was wrong, what a confined turn is, and what a copilot turn can still reach.

## Answering the issue's design questions

- **What does "confined" include?** Both, and the issue's warning is why: narrowing context alone leaves a turn that can fetch what it was not given. Context is narrowed (no retrieve→inject, no company memory backend, no writeback) and tools are narrowed to none, with the policy denying by name so the narrowing cannot be undone by a later wiring change.
- **Turn or distinct agent?** A confined **turn**, on an agent built for it and thrown away. No provisioning, nothing to keep in sync with the roster, and the grant reduction is host-side rather than requested politely. The `Confinement` enum is the reusable seam: the next surface that needs to reason about one object adds a variant.
- **What does it say when a question needs company-wide context?** It answers what it can, names the part it cannot answer here, and points at the company chat. Refusing outright is honest and unhelpful; answering unconfined defeats the purpose; saying which it did is the third option the issue predicted.

### DIFF
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -310,20 +310,32 @@ scope; SSE (`/chat` streaming, the `/events` work feed) is not yet wired.
   `owns` to match on prefix, would break that surface — see
   [`frontend/src/api/workflow-copilot.ts`](../../../frontend/src/api/workflow-copilot.ts).
 
-  **Thread addressing isolates transcripts. It does not scope authority.**
-  These are two different things and only the first is enforced here. The
-  thread id decides who answers and where the exchange is journaled; it does
-  **not** narrow the orchestrator's context or its tool grants, which stay
-  company-wide for every `/chat` turn whatever thread it names. A caller that
-  needs the responder confined to one subject has to constrain it in the
-  prompt — which is advisory — or build a genuinely scoped agent, which this
-  seam is not.
+  **Thread addressing isolates transcripts. For every thread but one, it does
+  not scope authority.** The thread id decides who answers and where the
+  exchange is journaled; for an ordinary thread it does **not** narrow the
+  responder's context or tool grants, which stay company-wide however the turn
+  is addressed.
 
-  That is a scoping property, not an authorization one: `/chat` is already
-  authenticated and company-scoped, so an operator addressing a workflow
-  thread gains nothing they could not get by opening the Chat tab or calling
-  the workflow routes directly. The copilot therefore adds no privilege; what
-  it adds is a transcript that stays out of the team's chat.
+  **The copilot thread is the exception (issue #416).** A `chat` id matching
+  `workflow-copilot:<workflowId>` ([`company::copilot`](../../../src/company/copilot.rs))
+  makes the turn **confined**, host-side, in two places that hold independently:
+
+  - the harness runs it on an ephemeral agent with **no tools, no company
+    memory and no delegation** ([`harness::confine`](../../../src/harness/confine.rs)),
+    and skips the retrieve→inject step and the memory writeback, so the turn
+    answers from the message it was sent and leaves nothing behind. Every tool
+    call is denied by the host with a reason, so an empty toolbelt is a
+    boundary rather than an absence;
+  - the `/chat` handler does not open a board card from a copilot message, so a
+    question phrased as a request cannot leave work on the board. That half is
+    in the default build, not behind `openhuman`.
+
+  Confinement narrows one **turn**; it is not an authorization check and must
+  not be read as one. `/chat` is already authenticated and company-scoped, so
+  an operator addressing a workflow thread gains nothing they could not get by
+  opening the Chat tab or calling the workflow routes directly. What the
+  copilot adds is a transcript that stays out of the team's chat and an answer
+  drawn from one workflow rather than from everything the company knows.
 
   Two more consequences worth knowing before reusing the seam. A chat turn
   runs the **whole** company cycle, so an actionable message also opens a
--- a/frontend/src/api/workflow-copilot.ts
+++ b/frontend/src/api/workflow-copilot.ts
@@ -22,15 +22,22 @@
 //    "no cross-workflow leakage" criterion, enforced server-side rather than by
 //    this file remembering to filter.
 //
-//    **Read that precisely: it isolates TRANSCRIPTS, not authority.** The
-//    thread id picks the responder and the journal; it does not narrow the
-//    orchestrator's context or tools, which stay company-wide. Asking this
-//    copilot about another workflow is not prevented by the transport — only
-//    by the instruction in the composed prompt, which is advisory. That is a
-//    deliberate limit of reusing the chat seam rather than building a scoped
-//    agent, and it costs nothing in privilege: the operator is already
-//    authenticated to the company and can reach the same orchestrator from the
-//    Chat tab and every workflow route. See `docs/spec/runtime/api.md`.
+//    **It isolates TRANSCRIPTS.** One workflow's copilot never sees another's,
+//    and none of it appears in the team's chat.
+//
+//    Until issue #416 that was the *whole* of the scoping: the thread id picked
+//    the responder and the journal and narrowed nothing else, so the teammate
+//    answering was the company orchestrator with its full context and tools,
+//    and "answer only about this workflow" was an instruction in the prompt —
+//    advice, not a boundary. It cost nothing in privilege (the operator can
+//    reach the same orchestrator from the Chat tab), but it meant an answer
+//    about a workflow could be drawn from anywhere in the company.
+//
+//    The host now reads the thread itself. `workflow-copilot:<id>` runs a
+//    **confined turn**: an ephemeral agent with no tools, no company memory and
+//    no delegation, whose whole world is the message composed below. See
+//    `src/company/copilot.rs` (the convention) and `src/harness/confine.rs`
+//    (the boundary), plus `docs/spec/runtime/api.md`.
 // 3. **It rehydrates for free.** Because the exchange is journaled under that
 //    thread, `GET {scope}/chat/history?desk=<thread>` replays it after a reload,
 //    with no new storage and no new route.
@@ -130,7 +137,13 @@ export function composeCopilotMessage(context: CopilotContext, question: string)
 
   const lines: string[] = [
     `You are the workflow copilot for this company, answering about ONE saved workflow.`,
-    `Answer only about this workflow. If the question is about a different workflow, or about the company generally, say so instead of guessing.`,
+    // The host confines this turn (issue #416): it runs on an agent with no
+    // tools and no company memory, so this is a description of the turn's real
+    // boundary rather than an instruction it could step outside of. Kept in the
+    // message as well as in the host-side persona so the disclosure the panel
+    // shows the operator is built from the same text that was actually sent.
+    `Everything you know about it is below. You have no tools and no access to the rest of the company, so answer from this material only.`,
+    `If the question needs a different workflow or the wider company, say which part you cannot answer here and that it has to be asked in the company chat. Do not guess, and do not claim to have looked anything up.`,
     ``,
     `## Workflow`,
     `Name: ${graph.name}`,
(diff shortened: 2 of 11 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev:
Heads up on file contention — this is my scoping error, not yours, but it needs handling before either side merges.

I split our work by file boundary — you on `frontend/**`, `test/e2e/**`, `ci.yml` and `docs/**`, me on `src/**` — so our branches would never need building together. That split assumed the console issues were console-only. #416 isn't: confining the copilot's answering agent genuinely needs harness work, so the boundary was wrong, not your change.

Three files in this PR are being edited right now by branches I have in flight:

| File | This PR | In flight |
|---|---|---|
| `src/harness/mod.rs` | +237/−38 | bounding a per-dispatch log flood (#449) |
| `src/harness/brain.rs` | +41 | a drain-commitment claim on the delegation queue (#453) |
| `src/server/operator.rs` | +14/−1 | a scope field on standing grants (#457) |

`src/harness/mod.rs` is the one that matters — both changes are substantial and in the same file.

**Nothing for you to do right now.** I own the merge ordering and I'll build the pairs together before anything lands; per-PR CI never builds the union of two open branches, and this project has had `main` broken by exactly that before. I'd expect this PR to go first, since it is further along, and my three to rebase onto it.

What would help: if a console issue turns out to need `src/**`, say so on the issue when you find out rather than at PR time. I can move my lane out of the way far more cheaply before an agent is mid-flight in the file.

`src/harness/confine.rs` and `src/company/copilot.rs` are new files, so no contention there.

I'll review the change itself separately once CI settles.

Issue #477, opened 2026-08-06
ci: no lane runs tinyplace-gated tests — the feature is type-checked and never executed


oxoxDev:
## What happens

No CI job ever **runs** a test with the `tinyplace` feature enabled. Every `tinyplace`-gated module is type-checked and never executed.

Measured against `ci.yml` on `main`:

| Job | Command | `tinyplace`? |
|---|---|---|
| `Rust` | `cargo clippy --locked --all-targets -- -D warnings` | no (default features) |
| `Rust` | `cargo test --locked` | no (default features) |
| `Rust (openhuman, tinycortex)` | `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets` | no |
| `Rust (openhuman, tinycortex)` | `cargo test --locked --features openhuman,tinycortex --lib` | no |
| `Rust (openhuman, mcp, telegram)` | `cargo test --locked --features openhuman,mcp,telegram --lib harness::build::tests` | no |
| — | `cargo check --locked --all-features --all-targets` | enabled, but **`check` only** — nothing is run |

Grep confirms it: of every `cargo test` / `cargo clippy` line in `ci.yml`, zero mention `tinyplace`.

## Why it matters

`--all-features` gives real coverage of *compilation*, which is why type errors in this code are caught. It gives zero coverage of *behaviour*. So any logic defect in a `tinyplace`-gated path ships green.

This is not hypothetical. `src/economy/outbox.rs` has a round-trip test for `Outbox::drain()` that passes, while the only caller of `drain()` in the entire tree is inside that same `#[cfg(test)]` module — there is no production drain at all (#454). A feature whose behaviour CI never executes is exactly where a test can assert something true about a function that nothing reaches.

## Relationship to #475

Same family, different mechanism. #475 is a test target that no job builds; this is a **feature set** that no job runs. Both produce the same illusion — a suite that reads as covered and is not. Worth fixing together, since both are edits to the same file.

## What the fix should do

Add a lane that runs, not merely checks, the `tinyplace`-gated tests — e.g. `cargo test --locked --features tinyplace --lib economy::`, following the precedent already set by the narrow `harness::build::tests` lane. Scope it deliberately rather than widening an existing job: the comments in `ci.yml` are explicit that `--lib` there is load-bearing and that the `--all-features` step must never become `cargo test`, and both warnings should be respected.

Whatever lane is added, prove it by the numbers — post the test count it executes, and confirm it is greater than zero.

## Acceptance

[...]

PR #493, opened 2026-08-06
fix(economy): a card queued during an outage is actually replayed, or the publish says it failed (#454)

### LINKED ISSUE #454: economy: cards queued while tinyplace is unreachable are never replayed — the outbox has no production drain
## What happens

When tinyplace is unreachable, `publish_card` queues the card and returns `Ok(())`:

```rust
Err(OpenCompanyError::Tinyplace { code, .. }) if code == "unreachable" => {
    // Offline: queue the card so it goes stale rather than erroring.
    self.outbox.enqueue(OutboxAction::PublishCard(card.clone()));
    Ok(())
}
```

The comment describes the intent exactly — degrade rather than error. But the **only** call to `Outbox::drain()` in the tree is inside a `#[cfg(test)]` module. Nothing replays the queue in a real build.

So a card enqueued during an outage is never published, and the caller was told it succeeded.

## Why it matters

The offline path is the one that most needs to work: it exists precisely for the case where the network is unreliable, which is when it will be exercised. Today an outage of any length means every card published during it is lost, silently, and the system reports a clean run throughout.

Nothing degrades — it just stops, while claiming otherwise. That is worse than the error the fallback was written to avoid, because an error is visible.

## Why it likely went unnoticed

The mechanism is complete apart from its trigger: the queue exists, `enqueue` works, `drain` works, and a test proves `drain` returns what was enqueued. Coverage of a replay path is not coverage of a replay *happening*. The missing piece is a caller, and no test asserts that one exists.

## What the fix should do

- Drain the outbox on reconnect, and on a bounded schedule, in a real build.
- Bound the queue, and decide explicitly what happens when it is full — dropping silently would reproduce this bug one level down.
- Do not report success for a card that is only queued unless the replay is guaranteed; if it cannot be, say so.
- Add a test that asserts the drain is reachable from production code, not only that it functions when called.

## How to verify

- A card published while tinyplace is unreachable appears once the connection returns.
- Restarting during an outage does not lose queued cards, or the loss is stated.
- Removing the production drain call fails a test.

## Related

Found in the sibling sweep for #445 — same class as #419, #445 and the `review_task` receipt: work staged behind an unqualified success, with a drain that does not run. This one is not agent-facing, so it produces no false statement to an operator; the loss is simply silent.

oxoxDev:
## Summary

Closes #454.

When tiny.place was unreachable, `publish_card` queued the card onto an outbox and returned `Ok(())`. **The only caller of `Outbox::drain()` in the whole tree lived inside its own `#[cfg(test)]` module.** The concrete economy is type-erased behind a trait with no flush surface, so no production code could ever reach that queue — every card published during an outage was lost, while the caller was told it succeeded.

This is the recurring defect class here: stage the work, return an unqualified success, depend on a drain that never runs. `publish.rs` (#452) is the merged template, and this follows it.

Two halves:

**A drain that exists.** A replayer is spawned where the economy is built — the single production attachment point — holding a weak reference so a runtime rebuild cannot leak it, and re-queuing on continued unreachability. The card slot is bounded to one entry, newest wins, because replay only ever needs the newest and the next boot republishes from the manifest anyway.

**A success that is true.** The unreachable arm now returns `Ok(())` only when a replayer is attached to send it. Without one it propagates the original error and queues nothing. A constructor path that forgets to spawn the replayer inherits a visible error instead of a silent drop — which is the guarantee, rather than fixing the two call sites that exist today.

`send_a2a_task` no longer queues a ghost copy alongside its `Err`. Background replay of a paid task is a double-send hazard with no budget context at flush time, so the contract splits deliberately: **a card publish degrades and replays; a task send fails and tells.**

## API Or Behavior Changes

**An offline card publish can now return an error where it previously always returned `Ok`.** That is the fix — the previous `Ok` was false whenever no replayer existed, which was always in production.

Two outbox variants with no production producer were pruned; the queue now holds exactly the one idempotent action it can honestly replay.

## Tests

- [x] `cargo test --locked --features openhuman,tinycortex --lib`

`Cargo.lock` unchanged. No user-facing surface — this is tiny.place wire traffic and logs — so backend proof is the whole of it.

**A test that the drain works when called is not coverage that the drain is reached**, which is exactly how this bug survived: the outbox already had a passing round-trip test and no production caller. So the assertions here are about reachability:

- built through the **production spawn path**, offline publish queues, then on reconnection the card is actually sent and the outbox empties;
- built through the **bare constructor** with no replayer, offline publish returns `Err` and queues nothing;
- a **builder-level** test asserting the discoverable path yields an economy that can degrade offline — a property that only holds while the spawn call is present.

Delete the spawn call and offline publishes start erroring instead of lying. That is the failure direction we want, and a test asserts it.

## Documentation

The outbox module doc now states the restart-loss position plainly: with a card-only queue the loss self-heals on the next boot, because the card is rebuilt from the manifest. Durable cross-restart replay remains a separate concern.

**One thing reviewers should know: CI compiles this code but never runs it.** No test or clippy lane enables the `tinyplace` feature — every job uses default features or `openhuman,tinycortex`, and the `--all-features` step is `cargo check` only. That gap is filed as #477. Until it is closed, these tests were run locally and CI's green says nothing about them.

### DIFF
--- a/src/economy/adapter.rs
+++ b/src/economy/adapter.rs
@@ -6,10 +6,11 @@
 //!   flag standing in for the Identity approval checkpoint) and funding covers
 //!   the registry fee — catching the `402` challenge, budget-checking, then
 //!   completing the paid registration;
-//! - publishes the Agent Card, queuing it to the [`Outbox`] (never erroring)
-//!   when tiny.place is unreachable;
+//! - publishes the Agent Card, parking it in the [`Outbox`] for the attached
+//!   replayer when tiny.place is unreachable — and erroring when no replayer is
+//!   attached, because then nothing would ever send it;
 //! - sends outbound A2A tasks, paying an x402 challenge under budget and
-//!   journaling the spend, or queuing the task when offline;
+//!   journaling the spend, and **failing** rather than queuing when offline;
 //! - quotes and pays firm requirements, **failing closed** the instant a
 //!   payment would exceed either the caller's [`BudgetScope`] or the company's
 //!   monthly ceiling, and journaling every in/out movement to the ledger.
@@ -207,6 +208,23 @@ impl AgentEconomy for TinyplaceEconomy {
         }
     }
 
+    /// Sends an outbound A2A task, and **fails** when tiny.place is unreachable
+    /// rather than deferring it.
+    ///
+    /// This is the deliberate other half of [`publish_card`](Self::publish_card)'s
+    /// contract, and the split is about money (issue #454). A card publish
+    /// degrades and replays: it is idempotent, it costs nothing, and the newest
+    /// card is always the right one to send whenever the network returns. A task
+    /// send does neither. It may carry an x402 payment, the budget scope that
+    /// authorised it belongs to the caller's cycle and is gone by flush time, and
+    /// a replay that lands after the caller already retried is a double-send —
+    /// which here means a double-spend. So the error goes back to the caller, who
+    /// is the only party holding the context to decide whether to retry it.
+    ///
+    /// Before #454 this arm did *both*: it pushed a copy onto the outbox **and**
+    /// returned the error. Nothing ever drained that copy, so it was pure
+    /// unreachable state; had anything drained it, it would have been a
+    /// background double-send with no budget behind it.
     async fn send_a2a_task(&self, to: &AgentAddr, task: A2aTask) -> Result<A2aTaskHandle> {
         let params = serde_json::json!({
             "id": generate_id(),
@@ -239,12 +257,8 @@ impl AgentEconomy for TinyplaceEconomy {
                 Ok(handle_from_response(&response, &rpc.id))
             }
             Err(OpenCompanyError::Tinyplace { code, message }) if code == "unreachable" => {
-                // Offline: queue the task and surface the error so the caller
-                // decides whether to retry.
-                self.outbox.enqueue(OutboxAction::SendTask {
-                    to: to.clone(),
-                    task,
-                });
+                // Offline: surface the error and queue nothing. The caller owns
+                // the retry decision for a task that may cost money.
                 Err(OpenCompanyError::tinyplace("unreachable", message))
             }
             Err(err) => Err(err),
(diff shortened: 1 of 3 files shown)


oxoxDev:
Two additions to the description, both worth knowing before review.

**The builder call was proved load-bearing, not assumed.** The spawn line was deleted from `maybe_build_economy`, the suite re-run, and `discoverable_path_builds_an_economy_that_can_degrade_offline` failed with `Tinyplace { code: "unreachable" }` — while all 51 economy tests still passed. Then it was restored and re-run green.

That distinction is the whole point of this PR: the outbox already had a passing round-trip test and no production caller, so "the economy tests are green" was true and meaningless. The only assertion that catches a deleted drain is one that goes through the discoverable construction path.

Local run against the feature CI never exercises (#477): `cargo test --locked --features tinyplace --lib economy::` → **51 passed, 0 failed**. Plus `cargo clippy --all-features --all-targets -- -D warnings` clean.

**One deviation I'd like a second opinion on.** The replayer timing test uses a real 20 ms interval polled against a 5 s budget rather than a paused clock, because `tokio`'s `test-util` feature is not enabled and `Cargo.toml` was outside this change's scope. The replay interval is therefore a parameter of the spawn function — production passes 30 s — which does let the test drive the *real* spawn path, flag and all, rather than calling `flush_outbox` directly. That part is a genuine improvement.

But a wall-clock budget is exactly the shape that fails under a loaded runner, and today has already produced a concurrency-sensitive E2E flake on this repo (#483). Adding `tokio = { version = "1", features = ["test-util"] }` to `[dev-dependencies]` would allow `start_paused = true` and remove the timing dependency entirely. It is a one-line change that does not move `Cargo.lock`. Happy to make it here if a reviewer prefers determinism over keeping the manifest untouched — I left it out only because the manifest was out of scope for the lane.

Also noted: `OutboxAction` survives as a single-variant enum rather than being deleted, because `src/economy/mod.rs` re-exports it and that file was out of scope. The invariant is documented in the file — an action belongs in the outbox only if replaying it unattended is safe.

Issue #440, opened 2026-08-05
workflows: the cron scheduler starts runs by its own path, not the shared one


oxoxDev:
## What is missing

Starting a workflow run carries a discipline: begin a supervised run, journal the outcome on both the success and failure arms, and make sure a run that ends — however it ends — is recorded rather than left looking live forever.

That discipline now lives in one shared place, used by the console's Run button and by the approval-resume path. **The cron scheduler still has its own copy**, because its body also claims the schedule and sweeps the delivery log, and folding those into the shared helper would have made it a union of two unrelated jobs.

That was the right call at the time. It leaves two implementations of the same discipline, and the usual outcome is that one of them gains a fix the other does not.

## Why it matters

The scheduler is the path with the least supervision — nobody is watching when a cron run starts at three in the morning. It is the worst place for the run-recording discipline to drift out of step with the path that gets exercised daily.

## What the fix should do

- The scheduler starts its runs through the same shared helper, so there is one definition of "a run was started and its end was recorded".
- The schedule claim and the delivery-log sweep stay the scheduler's own concern — they are genuinely its job, and pushing them into the shared path would recreate the problem in the other direction.
- A test that pins the two paths producing the same run bookkeeping, so a future change to one cannot silently skip the other.

## How to verify

- A cron-triggered run produces the same run record and journal entries as a console-triggered one.
- A cron run that fails is recorded as finished, not left appearing live.
- The scheduler's own responsibilities still happen exactly once per fire.

## Related

Deferred deliberately when the shared spawn path was introduced; recorded here rather than left as a comment.

oxoxDev:
Fixed and on `main`.

Merged in PR #496 (`5986640a`). GitHub did not auto-close this one — a `Closes #a, #b, #c` list only closes the first entry — so closing by hand after verifying against `upstream/main` rather than trusting the PR body.

Verified: the fire path calls `spawn.spawn(workflow, input, true)` in `src/runtime/workflow_scheduler.rs`. The three remaining `record_run_finished` mentions in that file are comments, not calls — checked, since a grep hit alone would not have distinguished them.

Found while doing this and **not** fixed: there is a third copy of the same discipline in the orchestrator's `run_workflow`, which still does its own supervisor registration and end-of-run journalling. This issue named only the scheduler, so that one is untouched and wants its own issue — otherwise the pattern looks resolved when it is not.

PR #496, opened 2026-08-06
fix(workflows): approving a paused run does not re-send what it already delivered (#438, #440)

### LINKED ISSUE #438: workflows: resuming an approved run re-executes upstream nodes, so a delivery can be sent twice
## What happens

A workflow that pauses for an approval is **settled**, not suspended — the engine finishes the run and reports which gates it stopped at. So resuming is necessarily a **re-run**: the workflow starts again from the beginning with the approval supplied.

That is the right primitive, and it is restart-durable for free. But it means every node upstream of the gate executes a second time. Agent nodes re-spend tokens. And an `output` node that already ran **delivers again** — the established-thread check is state-based rather than run-based, so a warm-recipient email will be re-sent.

## Why it has not bitten yet

A gate is normally placed *before* the node with the side effect — that is the point of gating it — so in the common shape the downstream never ran and there is nothing to repeat. The exposure is real but narrow: it needs a gate positioned after a delivery.

That is exactly the kind of thing that stays theoretical until someone builds a workflow that reviews a draft after it was sent.

## Options, none of them free

- **Checkpointed resume** — the engine can carry a checkpointer, which would let a resumed run skip completed nodes rather than repeat them. This is the correct long-term answer. It needs a durable checkpoint store that does not exist yet, and the engine's checkpointed entry point does not currently combine with the progress observer and the cancellation token the host relies on, so it is a multi-repo change.
- **Guard re-delivery from a continuation run** — mark a run as a continuation and refuse deliveries that a prior run in the same lineage already made. Cheaper, and honest about what it is: a targeted guard on the one side effect that is externally visible and irreversible.
- **Refuse to resume a workflow whose gate sits after a delivery**, and say so when the workflow is saved rather than when it runs. Prevention at the point the mistake is made, which is the pattern this codebase prefers — but it constrains a shape someone may legitimately want.

## What the fix should do

Whichever route: **an approval must never cause a message to be sent to a person twice.** Everything else here is cost and latency; that one is a user-visible mistake made in the company's name.

## How to verify

- A workflow whose gate follows a delivery node, approved, sends exactly one message.
- A workflow whose gate precedes its delivery behaves as it does today.
- The cost of a resumed run is stated somewhere the operator can see, so re-execution is visible rather than surprising.

## Related

Named as a known limitation when workflow approvals were first able to park at all.

oxoxDev:
## Summary

Closes #438 and #440.

**#438 — approving a paused workflow could send the same email twice.**

A workflow that pauses for an approval is *settled*, not suspended: the engine finishes the run and reports which gates it stopped at. So resuming is necessarily a re-run from the top, and every node upstream of the gate executes a second time. Agent nodes re-spend tokens. An `output` node that already ran **delivers again**, because the established-recipient check is state-based rather than run-based — so a warm-recipient report is re-sent on approval.

The re-run primitive is correct and restart-durable, and it is untouched. What is added is a delivery ledger: the run records what it actually delivered, that record rides on the parked approval card itself, and a reached output node whose id is in it is skipped rather than dispatched. Keeping it on the card preserves the property the resume design is built on — the card is self-contained, so this survives a restart the same way the rest of it does.

`Pending` counts as delivered, deliberately. A parked cold-send card is durable and approving it sends, so treating it as already-delivered also closes a second hole: cold-recipient parking had no dedupe, and a continuation stacked a second identical card for the same send.

**#440 — cron started runs by its own copy of the shared path.** The scheduler kept its own supervisor registration and end-of-run journalling, duplicating what the shared spawn primitive owns. Nothing was broken — the copies agreed — but a fix landing in one would not land in the other. The fire path now goes through the shared primitive; the schedule claim and the delivery-log sweep stay the scheduler's own.

## API Or Behavior Changes

**A resumed run no longer re-delivers reports its earlier attempt already sent.** Everything else about resume is unchanged.

The approval card now carries a plain-language note that approving re-runs the workflow from the start, that earlier steps run again and re-spend tokens, and that already-delivered reports will not be re-sent — the cost is stated where the decision is made.

**An honest limit, stated rather than hidden:** a *partial* fan-out is not retried. If a report reached two of three admins before the run paused, the third will not receive it on resume. Re-sending to two people in order to reach one is the worse outcome, so this takes the other side of that trade deliberately. The full answer is checkpointed resume, which needs work in the vendored engine — its checkpointed entry point does not compose with the observer and cancellation token this runner needs. The guard here is forward-compatible with that.

`DeliveryReason` gained a variant. Its kebab-case token is additive and the console types the field as a string, so no frontend change is owed.

## Tests

- [x] `cargo test --locked --features openhuman,tinycortex --lib`

`Cargo.lock` unchanged.

The decisive assertion for #438 is a counting mail sender across an approval lineage: the first run of an output-then-gate graph sends exactly one, the resumed run sends zero, and **the total across the lineage is exactly one**. That is the double-send claim itself rather than a proxy for it. Around it: sent-skipped, pending-skipped, an unlisted node still delivering, a non-continuation run untouched, and the ledger accumulating across two gates.

#440 is covered by the existing scheduler suite plus a parity test — one run driven through the scheduler and one through the shared primitive over a shared event log, asserting the journalled records differ only in the scheduled flag and the run id.

**A live-host check is owed and is not claimed here.** #438 has a user-facing surface — the Approvals card, the run drawer, and a real inbox. The sequence that proves it: author a workflow whose gate follows a warm-recipient delivery, run it, approve from the console, then confirm the recipient's mailbox holds **exactly one** message, the continuation's drawer shows the skipped already-delivered row, and the card carried the re-run cost note. A gate-before-delivery workflow must behave exactly as it does today.

## Documentation

The spawn primitive's module doc described the scheduler's copy as a deliberate omission; that is no longer true and has been rewritten.

### DIFF
--- a/src/ports/workflow_runner.rs
+++ b/src/ports/workflow_runner.rs
@@ -77,9 +77,10 @@ pub enum DeliveryStatus {
     /// then look at Approvals for what became of it.
     Pending,
     /// Deliberately not attempted — a policy precondition was unmet (no mailbox
-    /// configured, or a cold recipient on a runtime that cannot park). Not an
-    /// error; the report simply was not owed to that address under the current
-    /// rules.
+    /// configured, or a cold recipient on a runtime that cannot park), or a run
+    /// earlier in this lineage already delivered this node's report and the
+    /// continuation must not send it twice (issue #438). Not an error; the
+    /// report simply was not owed to that address under the current rules.
     Skipped,
     /// Refused by policy: the company does not grant what the destination needs.
     Denied,
@@ -146,6 +147,21 @@ pub enum DeliveryReason {
     /// The recipient is not an established thread and the send could not be
     /// parked for approval either.
     ParkingUnavailable,
+    /// A run earlier in this lineage already delivered this node's report, so
+    /// the continuation did not send it again (issue #438).
+    ///
+    /// Resuming an approved gate is a **re-run** — the engine settles when it
+    /// pauses, so continuing means walking the graph again from the trigger.
+    /// Every `output` node upstream of the gate is therefore reached a second
+    /// time, and without this the operator's approval would mail the same
+    /// report to the same person twice. The ledger of what a lineage has
+    /// already sent rides the continuation's trigger input; see
+    /// [`crate::runtime::workflow_resume`].
+    ///
+    /// A **parked** report counts as delivered too: the card is durable and
+    /// approving it sends, so re-parking would stack a second identical card
+    /// and approving both would send twice.
+    AlreadyDelivered,
     /// A `channel` report was posted to the wired adapter.
     ChannelPosted,
     /// The destination names a channel this deployment never wired. **Which
@@ -197,6 +213,10 @@ impl std::fmt::Display for DeliveryReason {
             Self::ParkingUnavailable => {
                 "the recipient has never written to the company, and the report could not be \
                  queued for approval either"
+            }
+            Self::AlreadyDelivered => {
+                "an earlier run in this workflow's approval lineage already delivered this report, \
+                 so it was not sent again"
             }
             Self::ChannelPosted => "posted to the channel",
             Self::ChannelNotWired => "the destination channel is not wired on this runtime",
--- a/src/runtime/workflow_resume.rs
+++ b/src/runtime/workflow_resume.rs
@@ -49,12 +49,40 @@
 //!
 //! **Upstream nodes re-execute.** That is the engine's documented semantic for
 //! resume, not something added here, but it is real: agent nodes re-spend
-//! tokens, and a reached `output` node **re-delivers** — a warm-recipient email
-//! will send a second time, because the established-thread check is state-based
-//! rather than run-based. This is acceptable for v1 because a gate normally
-//! sits *before* the side-effecting node it is gating, which is the entire
-//! reason to author one. It is not acceptable silently, which is why it is
-//! written here, in the PR, and nowhere hidden.
+//! tokens on every continuation. A gate normally sits *before* the
+//! side-effecting node it is gating — which is the entire reason to author one
+//! — so for most graphs the cost is tokens and wall-clock. It is not acceptable
+//! silently, which is why it is written here, in the parked card's own `note`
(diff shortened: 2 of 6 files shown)


oxoxDev:
Adding the evidence for the double-send claim, because "a test passes" is not the interesting part — this defect class survives precisely by having green tests that assert the wrong thing.

**The test was run as a negative control**, with the ledger bypassed, to prove it actually catches the bug rather than passing for an unrelated reason:

```
assertion `left == right` failed: one report, one send, across the whole lineage:
  [("acme@…", OutboundEmail { to: "ada@acme.test",
       subject: "[Acme] Lineage — Owner summary",
       body: "{\"request\":\"quarterly numbers\"}" }),
   ("acme@…", OutboundEmail { to: "ada@acme.test",
       subject: "[Acme] Lineage — Owner summary",
       body: "{…,\"approvals\":[\"gate\"],\"__opencompany_delivered\":[…]}" })]
  left: 2
 right: 1
```

The same person, the same report, twice — which is the user-visible bug stated as a test failure. With the guard in place: run 1 sends 1, run 2 sends 0, **total exactly 1 across the lineage**, and run 2's row reads `Skipped` / `AlreadyDelivered`.

The test drives a real engine with a counting mail sender and builds the continuation through the **production** continuation path — the same function the Approvals flow calls — rather than hand-constructing the resumed input.

Test count: 2344 passing on this branch against 2327 on base. The scheduler's cancel-in-flight and claim-release-on-panic tests both stay green; the panic now unwinds inside the spawn's inner task, so the join-error arm logs it and the claim still releases through its own guard.

**One thing found and deliberately not fixed here:** there is a *third* copy of the spawn discipline. `run_workflow` in the orchestrator does its own supervisor registration and two end-of-run journal calls — exactly the #440 pattern that this PR removed from the scheduler. #440 named only the scheduler, and that file belonged to another branch this round, so it is untouched. It wants its own issue; flagging it here so it is not lost when #440 closes and the pattern looks resolved.

Also for anyone rebasing: `deliver_outputs` gained a parameter and is re-exported, so a new call site passes an empty slice.

PR #503, opened 2026-08-06
fix(console): count the legacy import per kind instead of calling folders notes (#500)

### LINKED ISSUE #500: Workspace legacy-import toast counts folders as notes
## What happens

The Workspace's "import your local scratchpad" success toast reports `legacy.length` — the whole flat node list — as "N notes", but the import loops partition that list by kind.

- A scratchpad of 1 folder + 3 files toasts **"Imported 4 notes"** when three notes were imported.
- A folder-only scratchpad of exactly 1 folder toasts **"Imported 1 note"** having imported zero notes. The singular/plural branch keys off the same wrong number.

## Where

`frontend/src/views/WorkspaceView.tsx`, function `importLegacy`:

```tsx
toast.success(`Imported ${legacy.length} note${legacy.length === 1 ? "" : "s"}.`);
```

`legacy` is a flat `FsNode[]` holding both folders and files — the two import loops above it partition it by `n.kind`. The `IMPORT_FOLDER_NAME` root the import creates is packaging rather than imported content, and correctly stays outside any tally.

## Why no test caught it

The E2E fixture in `frontend/test/e2e/workspace.spec.ts` ("the retired local scratchpad is offered for import, its seed discarded") plants two entries, both `kind: "file"`. With no folders in the fixture, `legacy.length` and the file count coincide, so the assertion passes on the wrong expression. Nothing asserts the toast's number at all.

## What the fix should do

- Count per kind, and name both categories in the receipt, so the string is honest for every mix of input.
- Extend the E2E fixture with a folder (and a file nested under it) so the count/label drift cannot hide again.

## Acceptance

- The toast names what was actually imported for files-only, folders-only, and mixed scratchpads.
- A fixture containing at least one folder covers the import path.

oxoxDev:
## Summary

The Workspace's legacy-import toast reported `legacy.length` — the whole flat
scratchpad list — as "N notes", but the import loops partition that list by
kind. A scratchpad of 1 folder and 3 files toasted "Imported 4 notes" having
imported three; a folder-only scratchpad of one folder toasted "Imported 1
note" having imported none, singular/plural included.

The list is now partitioned once, at the top of `importLegacy`, and the loops
iterate those locals — so the counts and the work can no longer drift apart.
A small pure helper (`importSummary`) builds the receipt and **names both
categories**, because no single label is honest for every mix: calling the
whole list "notes" over-reports a mixed import, and counting only files makes
a successful folder-only import announce "Imported 0 notes." — a success that
reads as a failure and still never mentions the folders that arrived. The
`Imported from this browser` root the import creates is packaging rather than
content, and stays outside the tally as it always did.

Observed on a running console (real host, real browser), before → after:

| Scratchpad | Before | After |
|---|---|---|
| 3 notes | `Imported 3 notes.` | `Imported 3 notes.` |
| 1 folder | `Imported 1 note.` | `Imported 1 folder.` |
| 2 notes + 1 folder | `Imported 3 notes.` | `Imported 2 notes and 1 folder.` |

The common files-only case renders byte-identical to today's output — which is
the case where today's output was right.

Nothing caught this because the E2E fixture planted two entries, both
`kind: "file"`. With no folders, the list's length and the file count are the
same number, so an assertion on the wrong expression passes; and nothing
asserted the toast's number at all. The fixture now carries a folder plus a
note nested inside it, which also puts a real parent-child pair through the
import's `parentFor` remap for the first time, and the spec asserts the exact
rendered toast string.

## API Or Behavior Changes

User-visible copy only: the workspace legacy-import success toast now names
notes and folders separately. No API, route, or storage change. The failure
path (which deliberately leaves the legacy key intact), the seed sweep,
discard, and import ordering are untouched.

Adjacent and deliberately **not** fixed here: the migration *banner* still
labels the pending flat list "N notes" for the same reason the toast did — a
folder-only scratchpad is announced as "1 note … is not in the company
workspace yet". Out of scope for this issue; worth its own change.

## Tests

Console-side change; the Rust checklist below is not applicable.

Run locally from `frontend/`:

- `npx vitest run` — 72 passed (10 files), including the new
  `test/unit/workspace-import-summary.test.ts` (8 cases: files-only singular
  and plural, folders-only singular and plural, mixed, and the issue's own
  3 notes + 1 folder shape).
- `npm run typecheck`, `npm run typecheck:unit`, `npm run typecheck:e2e` — clean.
- `npm run build` — clean.
- `npx playwright test test/e2e/workspace.spec.ts` — 2 passed, against a live
  host (`cargo build --locked --bin opencompany`, default features), including
  the exact-toast assertion and the nested-note placement.
- Manual: drove the running console for the files-only, folders-only, mixed and
  single-note shapes and read the toast each time — the table above is what was
  observed, not what was predicted.

- [x] N/A: `cargo fmt --all -- --check` — no Rust changed.
- [x] N/A: `cargo clippy --all-targets -- -D warnings` — no Rust changed.
- [x] N/A: `cargo build --all-targets` — no Rust changed.
- [x] N/A: `cargo test` — no Rust changed.

## Documentation

None needed — no public API, route, or architecture change; the reasoning for
the wording lives beside the helper it governs.

Closes #500

### DIFF
--- a/frontend/src/views/WorkspaceView.tsx
+++ b/frontend/src/views/WorkspaceView.tsx
@@ -77,6 +77,30 @@ const AUTOSAVE_DELAY_MS = 800;
 /** The folder created to hold notes rescued from the retired local scratchpad. */
 const IMPORT_FOLDER_NAME = "Imported from this browser";
 
+/**
+ * The body of the import receipt, for a scratchpad of `files` notes and
+ * `folders` folders (issue #500).
+ *
+ * Names *both* categories rather than reporting one number, because the
+ * scratchpad is a flat list of two kinds and no single label is honest for
+ * every mix of them: calling the whole list "notes" over-reports a mixed
+ * import, and counting only files makes a folder-only import announce
+ * "0 notes" — a success that reads as a failure and still never mentions the
+ * folders that did arrive. The common files-only scratchpad renders exactly
+ * as it always did. The `IMPORT_FOLDER_NAME` root is packaging, not imported
+ * content, and is deliberately outside this tally.
+ */
+export function importSummary(files: number, folders: number): string {
+  const parts: string[] = [];
+  if (files > 0) parts.push(`${files} note${files === 1 ? "" : "s"}`);
+  if (folders > 0) parts.push(`${folders} folder${folders === 1 ? "" : "s"}`);
+  // Unreachable from the banner, which only renders for a non-empty
+  // scratchpad — but a receipt claiming an unqualified "Imported" would be the
+  // worst possible reading of an import that moved nothing.
+  if (parts.length === 0) return "nothing";
+  return parts.join(" and ");
+}
+
 /** What the editor's status line is currently reporting. */
 type SaveState = "idle" | "saving" | "saved" | "error";
 
@@ -328,6 +352,10 @@ export function WorkspaceView({ client, company }: Props) {
 
   async function importLegacy() {
     setImporting(true);
+    // Partitioned once, so the loops below and the receipt that reports them
+    // can never disagree about what was imported (#500).
+    const folders = legacy.filter((n) => n.kind === "folder");
+    const files = legacy.filter((n) => n.kind === "file");
     try {
       const root = await createNode(client, company, {
         name: IMPORT_FOLDER_NAME,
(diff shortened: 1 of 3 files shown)


oxoxDev:
Filed #507 for the migration banner, which carries the same defect on the pre-import surface and is deliberately out of scope here.

Note for whoever picks it up: the E2E fixture in this PR asserts the banner's current `3 notes` for a 2-file + 1-folder scratchpad. That is a pinned bug, not a regression — #507 has to update that assertion rather than treat it as a break.

PR #504, opened 2026-08-06
feat(tasks): a discussion message can be withdrawn, and stops shipping in the bundle

### LINKED ISSUE #358: tasks: journaled discussion posts have no redaction path — a pasted secret is permanent and ships in export/import
## Summary

The per-task discussion (#335, PR #348) is journaled as `TaskDiscussionPosted` in the company `EventLog` and is **append-only with no redaction path at all** — no edit, no delete, no tombstone. A message posted to a card is permanent, and because the journal is what export/import ships, it travels with the company bundle.

Raised in review on #348 (@oxoxDev), which asked for it as an issue rather than a paragraph in the module doc.

## Why it matters

The problem is not typos. It is that a discussion is the one surface in the product where a human types free prose about a piece of work that is *blocked*, and the natural next thing somebody pastes is the thing that unblocks it. The fixture text in #348's own tests — *"blocked on the API key"* — is exactly the shape of message where the next reply contains the key.

Today, once that is posted:

- there is no way to remove it from the card, for anybody, including an admin;
- it is in the append-only journal every read surface folds over, so it comes back on every detail read;
- it ships in export/import, so it leaves the instance with the bundle;
- `MAX_DISCUSSION_CHARS` is 4000, so a pasted credential, key block, or customer record fits comfortably.

Rotating the secret is the real-world remedy, but the record of it stays readable to every member of the company forever, and the operator has no way to act on their own mistake.

## Scope

A redaction path for journaled human prose. The discussion is the first writer that makes this urgent, but the mechanism is the journal's, not the tab's — `OperatorMessage` has the same shape of problem.

Sketch, to be settled in design rather than taken as the plan:

- a tombstone event (`TaskDiscussionRedacted { seq }` or a journal-level equivalent) that a later append supersedes an earlier post with — the log stays append-only, which is the property the rest of the runtime relies on;
- the read fold (`fold_task_journal`, `src/server/ops/tasks.rs`) drops or replaces the superseded row, so redaction is a property of every read surface rather than one of them;
- export/import must carry the tombstone with the post, otherwise a round trip resurrects the message;
- who may redact — the author, an admin, or both — and whether the fact that *something* was redacted stays visible (an audit trail that says "a message was removed" is a different product decision from removing it silently);
- whether the payload is scrubbed at rest or only hidden on read. Only the first actually helps for a leaked secret, and it is the one that conflicts with an immutable log — that tension is the real design work here.

## Acceptance

- [ ] An operator can remove a posted discussion message from every read surface
- [ ] A redacted message does not come back through export → import
- [ ] Whether a redaction is itself visible (and to whom) is decided and written down in `docs/modules/server/README.md`
- [ ] The chosen mechanism is stated as journal-wide or discussion-only, so the next human-prose writer inherits it

## Related

- #335 — the per-task Discussion tab (this is the follow-up it deferred)
- #348 — the PR that implemented it; the append-only decision is documented there and in `docs/modules/server/README.md`
- #242 — first-class runs, the other consumer of the journal's shape

M3gA-Mind:
## Summary

Fixes #358. A task discussion is the one surface where a person types free prose about work that is *blocked*, and the next thing pasted into a thread like that is often the thing that unblocks it — #348's own test fixture reads *"blocked on the API key"*. That post was append-only with no edit, no delete and no tombstone, and because the journal is what export/import ships, the message was not merely permanent: it was **portable**. It left the instance with the bundle.

`DELETE …/tasks/{taskId}/discussion/{seq}` withdraws one message. Below are the four decisions the issue asked to have settled, each with the argument rather than the assertion.

### 1. A tombstone, not a hard delete

`CompanyEvent::TaskDiscussionRedacted { task_id, seq, by }` **supersedes** the post at `seq`. The post event is never rewritten and never removed.

Three properties made this the only coherent shape, and a delete breaks all three for the sake of one tab:

- the log is append-only — *no control alters a past event* (epic #184);
- sequence numbers are stable ids other events name (a thread reply, a reaction), so removing an event shifts identities that are already referenced;
- import replays a bundle from zero to reproduce the original numbering, which a hole in the sequence would silently change.

**What a reader sees where the text was.** The row keeps its place in the thread, its `seq`, its author and its time. The text is replaced by `REDACTED_DISCUSSION_TEXT` ("This message was removed.") and the row carries `redacted: true` plus `redactedBy` — the label of whoever withdrew it. Deliberately not a silent disappearance: a message that vanishes without trace lets one member quietly rewrite what a thread said, which is a different product from "anyone may take back a mistake in the open". The substitution happens in the read fold, so the console thread and the task-detail export document (#352, which renders the *same* assembled value) cannot drift apart.

### 2. What export/import does with it — the half that actually closes the issue

`store::export::scrub_redacted_discussion` applies the same substitution to `events.jsonl`:

- **on the way out**, so a bundle never carries a withdrawn message;
- **on the way in**, so a bundle written by a host that predates this cannot smuggle one back through import — the resurrection the issue names, arriving through the one door the exporter cannot guard.

The **tombstone travels with the post**. Dropping it would leave the importing company with a post whose text is a placeholder and no record of why, and the fold would render it as an ordinary message reading "This message was removed." — a sentence nobody wrote. Carried through, the imported thread says exactly what the exporting one said, with the same attribution.

A redaction that stopped at the console would not make a pasted credential less exposed; it would move the exposure somewhere nobody is looking.

### 3. Who may redact

`ScopedCompany` — **any member of the company**, the same guard every other task write on this router carries. Not admin-only, and that is a match rather than an oversight: `DELETE …/tasks/{taskId}` already lets the same member remove the entire card, thread and all, so holding one message to a higher bar than the card that contains it would be an incoherent boundary. The withdrawal is attributed instead — `by` names whoever did it and the thread says so beside the row.

`404` when the card is unknown, or when `seq` names anything other than a discussion post **on this card** (another task's post, a dispatch, a chat reply) — a tombstone pointing at something else would be a way to write nonsense into the journal. Withdrawing an already-withdrawn message is a **no-op success**: the removal has still happened, and a `409` there would make a retry look like a failure.

### 4. Retrospective reach

- **Posts already journaled are fully covered.** The tombstone names a `seq`; nothing about the post has to change, so a message posted long before this shipped can be withdrawn today.
- **Bundles already exported are not**, and cannot be — they are copies in other people's hands.
- **This is not erasure at rest** on the instance that already holds the bytes, and the code says so rather than implying otherwise: that is precisely what the append-only property forbids. A leaked credential still has to be rotated. What the withdrawal guarantees is that the text stops being served by any read surface and stops leaving the building.

### Scope

Discussion-only, deliberately. `OperatorMessage` has the same shape of problem and is **not** covered here; generalizing before there is a second caller would fix the mechanism's shape around one use. `docs/modules/server/README.md` records the pattern the next human-prose writer should inherit — the pair (tombstone in the log, substitution in every fold *and* in the bundle), not the variant.

## API Or Behavior Changes

- **New route:** `DELETE …/tasks/{taskId}/discussion/{seq}` under both scope forms. `200` with the withdrawn row, `404` as above.
- **New journal variant:** `CompanyEvent::TaskDiscussionRedacted`. Additive — old logs never carry it and no existing variant's serialization changes. Its wire shape is pinned by a round-trip test because it is written into bundles and read back by a *different* instance.
- **`DiscussionMessage` gains `redacted` and `redactedBy`**, both `skip_serializing_if`, so an ordinary row keeps byte-for-byte the shape it had. A console that ignores the fields still cannot render a withdrawn message: the substitution is server-side.
- Agent-facing surfaces are unchanged and quote nothing: the sidecar wire event and the orchestrator insight line are card-only (`"Removed a discussion message on task t-1"`), which matters more here than for a post — the operator has just said that text should stop being readable.
- Console: the Discussion tab grows a per-message Remove control behind a confirmation, and renders a withdrawn row as withdrawn.

## Tests

**The export/import half, pinned hardest** — it is the part a reviewer should doubt:

- `a_withdrawn_discussion_message_does_not_survive_export_import` — asserts three separate leaks are closed: the **bundle bytes** (`events.jsonl` read back and grepped for the secret), the **imported journal** (carries the placeholder, not the text), and the **tombstone surviving** (so the imported thread can still report the withdrawal). A post nobody withdrew is asserted intact in the same bundle, so this is a substitution and not a filter that eats discussion history.
- `an_old_bundle_cannot_smuggle_a_withdrawn_message_back_in` — hand-edits an exported bundle back to its pre-#358 bytes (exactly the shape an older host writes) and proves import refuses to write the text into the fresh journal.

**The route and the fold:**

- `a_withdrawn_discussion_message_stops_being_served` — over the real HTTP stack: the row survives with its author and position, `redacted`/`redactedBy` are set, the secret is absent from the whole detail response, the other message is untouched, an ordinary row keeps its pre-#358 shape, both events are in the journal, and a second withdrawal is a no-op that appends nothing.
- `withdrawing_something_that_is_not_this_cards_post_is_refused` — another card's post and an empty sequence position are both `404`, and the other card's thread is untouched by either attempt.
- `a_withdrawal_survives_paging_back_past_it` — the trap this fold has: a tombstone is always *newer* than the post it withdraws, and the discussion arm skips events at or after `discussionBefore`. Applying the cursor to tombstones too would serve the original text to whoever scrolled far enough back — the one reader most likely to be looking for it.
- `task_discussion_redacted_round_trips_and_omits_an_absent_actor` — the tombstone's wire shape.

Local gates: `npm run typecheck` clean; `rustfmt --edition 2024 --check` clean on all eight Rust files touched.

**Browser walk** — real Chromium, against a stub serving the host's documented shapes (building the Rust host is prohibited on this machine):

- **The failure first, on `upstream/main`:** the Discussion tab renders `blocked on the API key: sk-live-7f3d-DO-NOT-SHIP` and there is **no remove control of any kind** — 0 by test-id, 0 by accessible name. That is the issue: no way to act on your own mistake, for anybody.
- **On this branch:** the control appears; the confirmation says what withdrawal actually does ("stops being readable here, on the exported record, and in this company's bundle", "cannot be undone", "you still need to rotate it"); after confirming, the row reads **"This message was removed. Removed by operator."**, carries `data-redacted="true"`, the secret is gone from the page, and the neighbouring message is untouched.
- **The host's copy** is the substituted row, and a **full page reload** still shows no secret — so it is the host's answer, not local state.
- No uncaught page errors in either walk.

One console-side trap worth naming, found during the walk: the tab's `absorb` deduped by `seq` and **ignored** any row it already held, so a withdrawal arriving on the poll would have left the original text on screen for as long as the tab stayed open. It now replaces a held row when the incoming one is withdrawn, and keeps the no-churn guard for every other case.

**Not run here:** the Rust half was not compiled or executed locally (the build matrix is prohibited on this machine), so CI is the first compiler to see it; and no Playwright spec is added, because the e2e lane's host is the default build and this behaviour is already asserted over the real HTTP stack in `write_test.rs`.

## Documentation

`docs/modules/server/README.md` gains **"Withdrawing a discussion message (issue #358)"**: the four decisions above with their reasoning, what a reader sees, what export/import does, who may, what it does not claim, and the explicit journal-wide-vs-discussion-only statement the issue's acceptance list asks for. The stale paragraph promising "no edit and no delete" is corrected rather than left to contradict the code, and the route table lists the new route.

## Acceptance

- [x] An operator can remove a posted discussion message from every read surface — the substitution is in the fold shared by the console read and the export document
- [x] A redacted message does not come back through export → import — asserted on the bundle bytes and on the imported journal, in both directions of the pipe
- [x] Whether a redaction is itself visible (and to whom) is decided and written down in `docs/modules/server/README.md` — visible to every reader of the thread, with the withdrawer named
- [x] The chosen mechanism is stated as journal-wide or discussion-only — discussion-only, with the pattern to inherit recorded for the next human-prose writer

### DIFF
--- a/docs/modules/server/README.md
+++ b/docs/modules/server/README.md
@@ -43,7 +43,7 @@ why this is not a read/write split: [authority.md](authority.md).
 
 | Surface (`ops::*`) | Routes |
 |---|---|
-| `tasks` | `POST …/tasks`, `GET …/tasks`, `GET …/tasks/{id}` (the Task Detail read, #185), `PATCH`/`DELETE …/tasks/{id}`, `GET …/tasks/inflight`, `POST …/tasks/{id}/steer` (#111), `POST …/tasks/{id}/discussion` (#335) |
+| `tasks` | `POST …/tasks`, `GET …/tasks`, `GET …/tasks/{id}` (the Task Detail read, #185), `PATCH`/`DELETE …/tasks/{id}`, `GET …/tasks/inflight`, `POST …/tasks/{id}/steer` (#111), `POST …/tasks/{id}/discussion` (#335), `DELETE …/tasks/{id}/discussion/{seq}` (#358) |
 | `task_export` | `GET …/tasks/{id}/export` (the task's record as a document, #352) |
 | `memory` | `POST …/memory`, `DELETE …/memory/{id}` (journals `MemoryFactDeleted`) |
 | `workspace` | `GET …/workspace`, `GET …/workspace/file/{id}`, `POST …/workspace`, `PUT …/workspace/file/{id}`, `PATCH`/`DELETE …/workspace/{id}` (the two `GET`s are REST twins of the GraphQL reads — the console has no GraphQL client, #177) |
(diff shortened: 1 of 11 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

Issue #337, opened 2026-08-04
spine: a task advances itself through the board


oxoxDev:
Part of the spine epic **#183**, §4. **UNBLOCKED as of 2026-08-05** — #242 shipped in three PRs (#342 the port and storage, #357 the write path, the read surface), so an attempt is now a durable thing with a status, a trace and a cost. #333 (approvals linked to their card) and #334 (the Done column was unreachable) have also closed since this was written.

With those in, this is the main remaining gap between "a person hand-carries every card" and the spine running itself.

## What is missing

A card moves because a person drags it. The one automatic edge in the system is `task_enters_in_progress` — entering In Progress dispatches a cycle. Nothing else moves on its own: a finished run does not advance its card, a blocked run does not park it, and nothing ever enters Planning.

So the spine's promise — type one sentence, watch it walk to a deliverable — currently requires an operator to hand-carry the card through every station.

## What this should do

Close the loop between a run's status and its card's column.

```
Todo ──plan──▶ Planning ──dispatch──▶ In Progress ──┬──▶ In Review   (a person must act)
                                                    ├──▶ Paused      (anything else)
                                                    ├──▶ Done        (finished, output attached)
                                                    └──▶ Todo        (failed or unplannable, reason on the card)
```

The mapping is not invented here — **#183** settles it and **#242** encodes it in `RunStatus`:

| Run status | Card lands in |
|---|---|
| `waiting_approval` | In Review |
| `paused` | Paused |
| `succeeded` | Done |
| `failed` | Todo, with `run.error` written to the card's note |
| `cancelled` | Todo, with the cancellation reason on the card |

The `failed` row is the settled decision from #183: **a task that cannot be planned goes back to Todo with the reason on the card**, not into a stuck column. `RunRecord.error` is what carries the reason — that field exists precisely so the card can say why.

---

## Planning — SETTLED (2026-08-04)

This was the open question. It is now decided: **Planning does real planning.** It is not a status-only station a card passes through.

### What the planning pass produces

Everything a person would want before work starts, written onto the card:

- **Decomposition** — the request broken into the steps that actually have to happen, in order.
- **Estimates** — what each step is likely to cost and how long it should take.
- **A detailed description on the doc** — the card stops being one sentence and becomes a brief someone could hand to a teammate. This is the main visible output.
[...]

Issue #183, opened 2026-07-30
EPIC: the spine — prompt → task → plan → build → approve → output link


Al629176:
> **Rewritten 2026-08-04 as a tracking epic.** The original was a flat spec with thirty unchecked boxes, several of which had already shipped, and a nav section that contradicted the rest of the document. This version records what is done, what is in flight, what is left, and — most importantly — **settles the three lifecycle questions the original left open**, because every remaining item depends on them.

## What this epic is

The **spine**: an operator types one sentence, and a task walks itself to a finished deliverable.

```
prompt box → Todo → Planning → In Progress → (Paused | In Review) → Done + output link
```

Everything else in OpenCompany is a surface. This is the product doing the thing. It is also the only part of the roadmap with a hard dependency chain, which is why it is tracked here rather than as loose issues.

---

## The three lifecycle decisions — SETTLED

The original left these open and they blocked every downstream issue. Decided 2026-08-04; build to them.

### 1. A task that cannot be planned goes **back to Todo, with the reason on the card**

Not a stuck state, not a silent dead-end. If Planning finds the request infeasible or needs more information, the card returns to Todo carrying the question, so the operator can revise the prompt and let it go again.

*Why:* a terminal "stuck" column is where work goes to be forgotten. The operator learns nothing and the card cannot move. Returning it with a reason keeps the loop closed and makes the failure legible.

### 2. **Paused vs In Review is decided by who unblocks it**

- If a **person** must act — an approval, an acceptance, a decision — it is **In Review**.
- Anything else — a dependency, a rate limit, a missing credential, a retry — is **Paused**.

One rule, no overlap, no judgement call at the boundary.

*Why:* the original asked for "a defined trigger condition" for Paused. Defining it by *who resolves it* is the only formulation that stays true as new blocking causes are added — it does not need updating each time.

### 3. A task may enter **In Review many times**

In Review is a re-enterable state, not a one-shot gate.

*Why:* modelling a single approval stop would force a run to batch its approvals into one prompt. That contradicts **#243**, where a grant is single-use and argument-exact — an operator approves *one call with these arguments*, not a bundle. Building a task around one stop would mean either abandoning that property or lying about what was approved.

---

### Shipped

| Area | Where |
|---|---|
| Nav cleanup — Inbox, Desk, Brains, Finances removed | #313 |
| Google / GitHub / X sign-in, proven end to end in a browser | #325 |
| A provisioned company's creator can sign in | #323 + manager #9 |
[...]

PR #506, opened 2026-08-06
test(board): assert what Planning means on each lane, and drag like a browser (#501)

### LINKED ISSUE #501: e2e: a board drag is lost when the host re-renders mid-drag, so the Planning drag fails only against a live harness
## What happens

`board-columns.spec.ts:104` — *"dragging into Planning moves the card without
dispatching it"* — fails against a host with the agent harness compiled in, and
passes against a default-feature one. Three consecutive runs of the new
`Console E2E (live brain)` lane (#467), same failure each time:

```
Error: expect(received).toBe(expected)
Expected: "planning"
Received: "todo"
```

The poll runs for 15 seconds and the card never leaves `todo`, so the drop did
not reach the host at all. The same spec passes on every run of the
default-feature `Console E2E` lane in the same CI runs.

Runs: [31094631249], [31096488117], [31098118431] — the `Console E2E (live
brain)` job in each.

## Why the two lanes differ

The only difference is what the host is doing while the drag happens. A
default-feature host answers a message from the offline echo brain and is
otherwise idle. A feature-gated one runs real agent turns, which push journal
events, which the board subscribes to.

The spec drives the drop synthetically, because `dragTo` does not drive React's
HTML5 handlers here:

```ts
await card.dispatchEvent("dragstart");
await planning.dispatchEvent("dragover");
await planning.dispatchEvent("drop");
```

Those three events are three separate round trips, and the dragged-card id lives
in React state between them. Anything that re-renders the board in that window —
a card list refreshed from an SSE event — drops it, and the `drop` handler then
has nothing to move. That is consistent with what the lane sees: not a slow
move, but no move.

## Why it matters beyond the spec

If that reading is right, the fragility is not only the test's. A board that
loses an in-flight drag whenever a background agent turn refreshes it would do
the same thing to an operator dragging a card while the company is working,
which is the normal case rather than an unusual one. Worth confirming against a
real drag before concluding it is test-only.

## What was done in the meantime

The spec is skipped in the live-brain lane only, naming this issue. It keeps
running on every push in the default lane, so the coverage is not lost — only
the feature-gated repeat of it.

Either the board holds a drag across a re-render (and the spec passes in both
lanes with the skip removed), or the drag state is confirmed safe and the spec
is made to drive a drag that does not straddle three round trips.

M3gA-Mind:
## Summary

Closes #501.

**The issue describes a bug that does not exist.** Its own failure artifact falsifies it, and the correction matters more than the fix: left as written, the next person reads the title and goes hunting for a drag that HTML5 loses on re-render. There is no such defect here.

### What the artifact says

`playwright-live-brain-failures` from run [31098118431], the page snapshot at the moment of failure. The card is in To-do, carrying this:

```
button "e2e planning drag 1786017364745 medium [system] planning could not read the model's
answer as a plan, so nothing was written — try again, or drag the card straight to In Progress
to run it unplanned"
```

That string is `src/harness/planning.rs:948`. **Only the host writes it.** So the drop reached the host, the console `PATCH`ed the card into `planning`, and the host put it back. The drag was never lost. The issue's central claim — "the drop did not reach the host at all" — is the one thing the evidence rules out.

### What actually happens

1. The drop lands; the console `PATCH`es `column: "planning"`; the host accepts it.
2. On a **planner-attached** host, entering Planning edge-fires one pass: `CompanyRuntime::plan_task` → `harness::planning::run_planning_pass`, a real model call. On a default-feature host `plan_task` is a `#[cfg(feature = "openhuman")]` no-op — "keeping the column inert", as its doc says.
3. The live-brain lane's brain echoes rather than plans, so `parse_draft` returns `None` → `PassFailure`.
4. `settle_failed` (`planning.rs:562`) appends that note **and sets `card.column = COLUMN_TODO`**.
5. The spec polls 15s and reads `"todo"`.

### Why the spec was wrong rather than the product

The spec's own comment scoped its no-dispatch assertion: it "lets the column ship **ahead of epic #183 §4's auto-advance**". §4 has since landed as the planning station (issue #337), whose module header states the contract:

| pass outcome | where the card lands |
|---|---|
| a plan, nothing blocking, a valid assignee | `in_progress` — **and the dispatch edge fires** |
| a plan, a hard prerequisite missing | `todo`, with the gap on the note |
| the pass itself failed | `todo`, with the reason on the note |

So on a planner-attached host a card **never stays in `planning`**, and the first row **does dispatch**. Both of the test's claims are false there by design, and both remain exactly right on the default lane. This is an assertion that outlived its own stated scope — the same shape as #475, one directory over.

## What changed, and why none of it is a relaxation

This is the part worth checking hardest, so it is itemised.

**1. The inert-Planning test is scoped to the lane where its contract holds.** It now carries `test.skip(LIVE_BRAIN, …)` — **inverted** from every other capability skip in the suite, because this test needs the *absence* of a brain. Every assertion inside it is byte-for-byte unchanged: still `toBe("planning")`, still zero `dispatched` timeline entries, still running on every push in the default lane. Nothing was loosened; a claim that is only true without a planner is now only made without a planner.

**2. The live-brain lane gains an assertion it never had.** This is the opposite of a relaxation and it is the substance of the change. That lane previously stated *no* contract for Planning — which is why it could report a failure and offer nothing to replace the claim with. The new test asserts what does hold there:

* the card **must not stay parked in `planning`** — the one outcome the product must never produce, and precisely what a genuinely lost drop or a stalled pass would look like;
* wherever it landed is one of the two documented columns (`todo` or `in_progress`);
* it carries a `[system]` note, so it says why it moved rather than moving silently.

It is deliberately agnostic about *which* landing. The lane's brain echoes, so today it always takes the failure row; pinning that would make the test a description of the mock rather than of the product.

**3. The drag is now the gesture a browser actually performs.** The three dispatches carried no `DataTransfer`, so `dragstart` could not `setData` the card id and `drop` could not `getData` it — the handler fell through to its React-state fallback (`moveTo` reads `dropped || dragId`). One shared `DataTransfer` handle across `dragstart`/`dragover`/`drop` puts the id where a real drag puts it. This is strictly *more* faithful than before: it exercises the production path instead of a fallback, and it is what makes the three round trips immune to anything that re-renders between them. It also closes the issue's own second suggested remedy properly rather than by hoping.

Net: no assertion weakened, one assertion added, one gesture made real.

## Is the drag itself safe? Yes, and here is why

The issue worries that "the dragged-card id lives in React state between them". In the product it does not. The column's `onDrop` reads the id from `e.dataTransfer`, which is browser-owned and survives any React re-render:

```ts
onDrop={(e) => { … void moveTo(col.id, e.dataTransfer?.getData(CARD_MIME) ?? null); }}
```

`dragId` state drives the drag affordance and serves as a fallback for browsers that mangle the payload. The only way to reach that fallback is a synthetic event with no `DataTransfer` — which is what the spec was doing and no browser does. So there is no operator-facing lost-drag defect to fix, and this PR does not pretend to fix one.

## Does the real failure deserve its own issue?

**No.** The real failure is this spec asserting a pre-auto-advance contract against a post-auto-advance host, and it is fully fixed here — the claim is scoped to the lane where it is true, and the lane where it is false now asserts what is. There is no residual product defect behind it, so a follow-up issue would have nothing to close.

**One product observation, which I have deliberately NOT fixed here.** The board header still reads *"Drag a card to move it; drop into 'In progress' to hand it to its assignee."* On a planner-attached host that is no longer the only way work gets handed over: a successful plan promotes the card to `in_progress` and the dispatch edge fires, so dropping into **Planning** can also spend an agent turn. The string is correct on the default build and misleading on a harness build. That is a one-line UI honesty question, not a test fix, and it belongs in its own issue rather than in this diff.

## API Or Behavior Changes

None. Test-only: one spec file. No product code is touched.

## Tests

- [x] `cargo fmt --all -- --check` — N/A - full local build matrix is prohibited on this machine; verified in CI.

No Rust file is touched by this change.

Verified locally: `npm run typecheck:e2e` clean (the spec suite is type-checked by `tsconfig.e2e.json`, which `tsc -b` does not reach).

The behavioural verification is the two CI lanes themselves, which is the only place a live host exists:

* `Console E2E` (default features) runs the existing test with its assertions unchanged, and now with a faithful `DataTransfer` drag. A pass there is the evidence that change 3 did not break the path that was working.
* `Console E2E (live brain)` skips that test and runs the new one. A pass there is the evidence that the drop reaches the host and the pass settles the card — the thing #501 believed was not happening.

Both are on this PR's run.

## Documentation

None needed: no product behaviour or API changed. The contract this spec now mirrors is already documented where it is implemented, in `src/harness/planning.rs`'s module header and on `COLUMN_PLANNING` in `src/ports/tasks.rs`; this change makes the spec agree with it rather than restating it.

### DIFF
--- a/frontend/test/e2e/board-columns.spec.ts
+++ b/frontend/test/e2e/board-columns.spec.ts
@@ -1,5 +1,7 @@
 import { expect, test, type Page } from "@playwright/test";
 
+import { LIVE_BRAIN } from "./capabilities";
+
 /**
  * Issue #301 — the board's shape, asserted against a live host.
  *
@@ -101,7 +103,39 @@ test("new work enters through one prompt box and lands in To-do", async ({ page,
   expect(created.note).toBe(long);
 });
 
+/**
+ * Issue #501. This test states a **no-planner** contract, and only a host
+ * without one keeps it.
+ *
+ * Its own comment used to say the no-dispatch assertion "lets the column ship
+ * ahead of epic #183 §4's auto-advance". §4 has since landed as the planning
+ * station (`src/harness/planning.rs`, issue #337), and a card entering Planning
+ * on a planner-attached host now edge-fires exactly one pass and is **settled**
+ * by it — never left sitting in `planning`:
+ *
+ * | pass outcome | where the card lands |
+ * |---|---|
+ * | a plan, nothing blocking, a valid assignee | `in_progress` — and the dispatch edge fires |
+ * | a plan, a hard prerequisite missing | `todo`, with the gap on the note |
+ * | the pass itself failed | `todo`, with the reason on the note |
+ *
+ * So on the live-brain lane both of this test's claims are false by design: the
+ * card does not stay in `planning`, and the first row dispatches. `plan_task`
+ * is a `#[cfg(feature = "openhuman")]` no-op without the harness, so the
+ * assertions below remain exactly right on the default lane, which is where
+ * this runs.
+ *
+ * The skip is therefore INVERTED — `skip(LIVE_BRAIN)`, not `skip(!LIVE_BRAIN)`.
+ * This test needs the absence of a brain, which is the opposite of every other
+ * capability skip in the suite. The harness contract is asserted by the test
+ * below instead, so the lane loses no coverage.
+ */
 test("dragging into Planning moves the card without dispatching it", async ({ page, request }) => {
+  test.skip(
+    LIVE_BRAIN,
+    "asserts Planning is inert, which is only true without a planner; the harness " +
+      "contract is covered by the live-brain test below. Issue #501.",
+  );
   const title = `e2e planning drag ${Date.now()}`;
   const seeded = await request.post(`${API}/tasks`, { data: { title } });
   expect(seeded.ok()).toBeTruthy();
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

PR #515, opened 2026-08-07
fix(workspace): a read the harness would cut can no longer invite a full-body overwrite (#417)

### LINKED ISSUE #417: workspace_read: a note between 16 KiB and 64 KiB is silently shortened, and the agent is then invited to overwrite it with what it saw
Found while auditing #410 (the Composio catalogue was cut by a 300-byte message cap). Same class, worse consequence: this one can destroy an operator's note.

## What happens

`workspace_read` clamps a note body at `MAX_CONTENT_BYTES` (64 KiB) and correctly announces its own cut. But the agent harness applies a *second*, smaller bound to every tool result — 16 KiB (`openhuman::context::DEFAULT_TOOL_RESULT_BUDGET_BYTES`; OpenCompany never overrides `tool_result_budget_bytes`). For any note between those two numbers:

1. The module thinks nothing was dropped (`dropped == 0`), so it emits the **write-eligible** branch: *"To revise it, call `workspace_write` with expected_updated_at=… and the complete new body."*
2. The harness then cuts the result at 16 KiB. The model receives roughly the first 16 KiB and an anonymous byte-boundary marker.
3. The agent, following the instruction it was given, calls `workspace_write` with "the complete new body" — which is the portion it actually saw. `MAX_WRITE_BYTES` is 64 KiB, so the write is accepted.

The rest of the note is gone. Nothing anywhere in the loop reports a loss.

This directly contradicts the invariant the module documents for itself — *"a truncated read can never become a write"* — because the write gate is measured against 64 KiB, not against the 16 KiB the model is actually handed.

## Two smaller problems on the same path

- The closing untrusted-content fence (`--- END WORKSPACE NOTE {nonce} ---`) is rendered after the body, so on exactly the notes large enough to hide a prompt injection the fence never reaches the model unterminated.
- The harness's own marker advises "re-run with a narrower query". `workspace_read` has no narrowing argument at all (`path`/`id` only), so the advice is unactionable and an identical re-run is the only move available — the retry loop #410 was about.

## Same-shape issue in `workspace_list`

`MAX_LIST_ENTRIES` is 300, but a rendered entry is ~90–105 bytes, so the 16 KiB budget bites at roughly 176 entries. Between ~176 and 300 the module's own (good) marker — *"… more entries not shown. Narrow the listing with the `prefix` parameter."* — is never generated, and above 300 it is generated and then truncated away, because it sits at the *end* of the body. The `unaddressable` notice has the same fate.

## Suggested direction

- Make the module's read bound the *effective* one (below the harness budget), so the module's own gate and the model's actual view agree — and keep the write-eligibility test on that same number.
- Put trailers that the model must act on where an outer cut cannot remove them (the header already carries the counts; the narrowing guidance should too).
- Consider making the harness budget explicit in OpenCompany rather than inherited, so these numbers are chosen in one place.

Scoped out of #410 deliberately: the fix is a design change to the read/write contract, not the same one-line mechanism.

oxoxDev:
## Summary

Closes #417.

`workspace_read` capped a note at `MAX_CONTENT_BYTES` (64 KiB) and emitted its
**write-eligible** branch whenever its own `dropped == 0` — telling the agent to
call `workspace_write` with "the complete new body". But the harness applies a
second, smaller bound to every tool result (`DEFAULT_TOOL_RESULT_BUDGET_BYTES`,
16 KiB), and `build_agent` never called `context_config`, so that default was
inherited silently.

For a note between the two numbers the module believed nothing was dropped, the
model saw roughly the first 16 KiB, and it complied — writing back the fragment
it had seen. `MAX_WRITE_BYTES` accepted it. **The rest of the note was destroyed
and nothing in the loop reported a loss.**

This restores the invariant the module documents for itself — *a truncated read
can never become a write* — by making the module's own bound the effective one,
so its gate and the model's actual view agree.

The same shape was fixed in `workspace_list`: its "narrow with `prefix`" guidance
sat at the end of the body, so it was cut away precisely when the listing was
long enough to need it. Trailers the model must act on now live in the header,
where an outer cut cannot remove them.

## API Or Behavior Changes

**Notes between 12 KiB and 64 KiB become agent-readable but not agent-writable.**
That window is the data-loss window itself, so the change is intended, but it is
user-visible: an agent that could previously revise such a note now receives
`Refused: … larger than the 12288-byte read limit`.

Operator edits are unaffected — the console and REST write path
(`server::ops::workspace::write_file`) does not go through `workspace_tools.rs`.
Verified directly rather than assumed: that module has no reference to either cap,
both are private consts with no re-export, and the frontend applies no size check.

`workspace_read`'s header now reports `returned X of Y bytes` when truncated.
`workspace_list` renders size-aware, keeping 300 as an absolute cap, with the
narrowing guidance and the `unaddressable` notice above the entries.

The tool-result budget is now stated explicitly in `build_agent` rather than
inherited. Behaviour is identical today; the point is that the number is chosen
here instead of by accident.

## Tests

49 tests in scope, up from 44, including a turn-level test through the real
middleware rather than a mock.

Every negative control was executed — break the fix, run, observe, restore:

| Control | Result |
|---|---|
| `MAX_CONTENT_BYTES` → 64 KiB | **build fails** on the const-assert |
| …assert neutralised so tests run | both read tests fail — one reports `rendered 66505 bytes, over the 16384-byte harness budget`, the other `a partial read still invited a full-body overwrite` |
| `READ_OVERHEAD_BYTES` → 0 | both read tests fail (17353 and 17038 bytes vs the 16384 budget) |
| delete the `current_len` write gate | write-refusal test fails |
| remove the size-aware list stop | list test fails: `rendered 23806 bytes, over the 16384-byte budget` |
| trailers moved back below entries | list test fails: guidance absent from the header at 184 of 301 entries |
| turn test under the old 64 KiB cap | fails — the model received a `[tool_result_preview]` envelope carrying the write-eligible instruction, an unterminated `--- BEGIN WORKSPACE NOTE`, and `[… 573 bytes truncated by tool_result_budget …]` with **no closing fence** |

That last one is the defect reproduced verbatim in the real pipeline.

Worth recording: the pre-existing test for this area passed with the bug present.
It exercised cap + 4 KiB, so it was structurally blind to the 16–64 KiB window.

- [x] `cargo test --features openhuman,tinycortex --lib harness::workspace` — 49 passed

Adjacent scopes re-run green: `harness::build` 19, `policy::` 92,
`runtime::builder` 23, `harness::toolbelt` 15.

`cargo build --all-targets`: N/A — the two clippy lanes above build all targets in
both feature configurations.

## Documentation

`docs/spec/runtime/manifest.md` — the 64 KiB figure was stale; the caps are now
documented as derived from the harness budget.

## Open question for review

`READ_OVERHEAD_BYTES` is 4096, but measured framing is ~969 bytes — about 4x
headroom, and every byte of it comes out of what an agent may write. Tightening to
2048 would still double the measured worst case and would raise the writable
ceiling from 12 KiB to 14 KiB. Kept at 4096 here; happy to change it if reviewers
prefer the tighter number.

Separately, and out of scope: the operator write path has **no** size limit of its
own — no handler validation, no body-limit layer, no store cap. The only backstop
is axum's implicit 2 MiB JSON extractor limit. Pre-existing, worth its own issue.

## Summary by CodeRabbit

* **New Features**
  * Workspace tool output limits now adapt to the available response budget.
  * Large workspace listings and notes are clearly marked when truncated, including returned and total sizes.
  * Path information is safely shortened to preserve useful content within the limit.

* **Bug Fixes**
  * Prevented writes to content that was only partially read, avoiding accidental data loss.
  * Improved boundary handling so outputs remain within the configured tool-result budget.

* **Documentation**
  * Updated workspace tool documentation to describe dynamic limits, truncation behavior, and write restrictions.

### DIFF
--- a/docs/spec/runtime/manifest.md
+++ b/docs/spec/runtime/manifest.md
@@ -268,10 +268,15 @@ prompt = "Weekly review and operator digest"
     `expected_updated_at` revision token taken from a prior read, so a note
     edited in the console since the agent read it is refused rather than
     clobbered; creating, renaming and deleting notes stay operator-only. Both
-    sides are capped at 64 KiB, which makes a larger note agent-read-only: the
-    agent sees a truncated body, and a write against it is refused rather than
-    discarding the part it could not see, so only an operator can edit it in the
-    console. Under
+    sides are capped at the agent harness's own per-tool-result byte budget
+    minus the framing a read wraps a body in (issue #417) — 12 KiB today,
+    derived rather than chosen so a full read always survives the harness cut
+    and the write gate is measured against the bytes the model actually
+    received. A larger note is agent-read-only: the agent sees a truncated body,
+    and a write against it is refused rather than discarding the part it could
+    not see, so only an operator can edit it in the console. **Operator edits
+    are not capped by this** — the console and the REST handlers write through
+    the `WorkspaceStore` port directly and never enter the agent tool path. Under
     `[policy].mode = "supervised"` (the default) a write additionally parks for
     approval, and under `readonly` it is denied — reads stay available in every
     mode. The namespace is **not** gateable by `[plan].token_budgets`: reads
--- a/src/harness/build.rs
+++ b/src/harness/build.rs
@@ -89,6 +89,37 @@ use crate::ports::skills_state::SkillState;
 use crate::ports::types::CompanyId;
 use crate::runtime::tools::extends_on_boundary;
 
+/// The per-tool-result byte budget every OpenCompany agent runs under.
+///
+/// The harness cuts **every** tool result to this many bytes on its way into
+/// the model's context — `ToolOutputMiddleware`, fed from
+/// `ContextManager::tool_result_budget_bytes`, which [`build_agent`] threads in
+/// below via [`AgentBuilder::context_config`]. It is the real ceiling on what a
+/// model ever sees from a tool, and it is *smaller* than the caps individual
+/// tools tend to write for themselves.
+///
+/// It is stated here rather than inherited on purpose (issue #417). Before this
+/// constant existed `build_agent` never called `context_config`, so the builder
+/// fell back to `ContextConfig::default()` and the 16 KiB became OpenCompany's
+/// effective bound by accident — invisible from this crate, and out of step
+/// with a tool that had capped itself at 64 KiB. `workspace_read` consequently
+/// told the model "nothing was dropped, you may write the complete body back"
+/// about a note the model had only seen the first 16 KiB of, and the resulting
+/// `workspace_write` silently destroyed the rest.
+///
+/// So any tool whose result carries a *contract* — "this is the whole thing",
+/// "act on the trailer below" — must size itself against this number, not
+/// against a cap of its own choosing. [`workspace_tools`] derives its read and
+/// write caps from it directly.
+///
+/// The value still tracks the vendored default: keeping the number identical
+/// makes adopting it a no-op today, so the behaviour change belongs to the
+/// consumers that now respect it, not to this line.
+///
+/// [`workspace_tools`]: crate::harness::workspace_tools
+pub(crate) const TOOL_RESULT_BUDGET_BYTES: usize =
+    oh::agent::context::DEFAULT_TOOL_RESULT_BUDGET_BYTES;
+
 /// Map a manifest cognition-tier hint to a hosted model/tier name.
 ///
 /// The manifest tier "never selects a model" (that is the TinyHumans backend's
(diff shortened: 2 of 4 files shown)


oxoxDev:
Addressed in 18f90d7f.

The finding was valid. The render loop bounds the listing by bytes and `break`s when the next line would exceed the budget — but the entry path was echoed unbounded. A node name is operator-supplied and no backend length-caps it, so one sufficiently deep path could render a line larger than the whole budget, fail the check on the loop's *first* iteration, and report a near-empty listing for a workspace that is almost entirely listable.

Fixed by bounding the echoed path with `echo_path`, the same clamp the header already uses. Two properties matter and both are covered:

- the clamp **announces its own drop** (`… (+N bytes)`) rather than presenting a shortened path as complete — the whole point of this issue
- `id=` is never clamped, so a bounded entry remains addressable

New test `one_oversized_path_does_not_hide_the_entries_behind_it`: an oversized path followed by 12 normal entries, asserting all 13 render, the clamp marker is present, and the id survives.

Negative control run — reverting the clamp fails that test alone (37 passed / 1 failed), restoring it returns 38. Scope is now 50 tests, both clippy lanes and fmt clean.

Not applied: skipping the entry instead of bounding it. That would drop a real node from the listing to save display width, and a silent omission mid-listing is the failure mode this issue exists to remove.

Issue #514, opened 2026-08-07
workflows: a company with no inference configured is told workflow execution is not wired in this deployment


oxoxDev:
## Summary

A company that has never had an inference source configured is told
`workflow execution is not wired in this deployment` when an operator clicks Run.
That is not true — the deployment is fine. The company simply has no inference
configured, which is something the operator can fix in the console in under a
minute.

Reported by the marketing pod: creating a content workflow, clicking Run, getting
an error toast, and finding zero runs in history with the page reporting the
workflow has never finished a run. Nothing was queued and nothing was kicked off,
so the operator has no signal about what to do next — and the message points them
at a deployment problem that does not exist.

## Why it happens

The workflow runner is built only inside the `#[cfg(feature = "openhuman")]` arm
of the runtime builder, and only when the company boots with a resolvable
inference source. With no source, `workflow_runner()` is `None`.

`run_workflow` then tries to tell two causes apart (#266):

```rust
let Some(runner) = company.runtime.workflow_runner() else {
    if super::inference::restart_pending_for(company.runtime.as_ref()).await {
        return Err(super::restart_required("workflow execution"));
    }
    return Err(super::not_wired("workflow execution"));
};
```

But `restart_pending` requires **all three** of: a tenant config that resolves
*now*, the company not already on the harness path, and the harness path
reachable. A company that never configured inference fails the first condition,
so `restart_pending_for` returns `false` and the route falls through to
`not_wired`.

There are three cases, and only two are covered:

| Situation | Message shown | Accurate? |
|---|---|---|
| Build genuinely has no workflow execution | not wired in this deployment | yes |
| Configured after boot, needs a rebuild | restart required | yes |
| **Never configured inference** | **not wired in this deployment** | **no** |

This is the exact failure the #266 comment set out to prevent — it says a
`not_wired` answer "sends the operator hunting a deployment problem that does not
exist." That fix covered the restart case and left this one.

Note that #290's rebuild-in-place work means configuring inference now recovers
the runtime without a restart, so the remaining defect is purely the message and
the dead end it creates.

## What the fix should do

Distinguish "no inference source is configured for this company" from "this build
has no workflow execution", and say so — naming the console screen that fixes it.
The operator's next action is completely different in each case.

Worth deciding at the same time: whether the Run control should be disabled with
an explanation when the company has no inference source, rather than presenting a
live-looking button that always fails. A first-time operator building their first
workflow currently gets an error with no route forward.

## How to verify

[...]

Issue #383, opened 2026-08-05
workflows: a run cannot be cancelled, and the console must hold a request open for its whole duration


oxoxDev:
## What the operator sees

Pressing **Run** holds a single HTTP request open until the workflow finishes.
A run with several agent nodes can take minutes, and for that whole time:

- there is no way to stop it — no cancel affordance exists, and none can, because
  the run is only reachable as an in-flight request;
- closing the tab or losing the connection abandons the response, though the run
  itself continues host-side and does still journal its outcome;
- any proxy or gateway idle timeout in front of the host can sever the request
  even though the run is healthy.

## Why it happens

`POST …/workflows/{wid}/run` is synchronous by design: the response body carries
the engine's final state, so the request cannot return until the run ends.

#371 made the *progress* asynchronous — a run's node events reach the console
over SSE while the request is still open — so the response is no longer the only
way to learn what a run is doing. That removes the reason the call had to be
blocking, but not the wire compatibility that requires it to stay available.

## What the fix should do

Add an asynchronous entry point alongside the existing one rather than replacing
it: a run POST that returns the `runId` immediately (#371 already mints one and
already returns it), with the outcome read back from the run history the same
`runId` keys. The blocking form stays the default for existing callers.

With a run addressable by id while it is in flight, cancellation becomes
possible: the engine already exposes `run_cancellable` with a cancellation
token, so a `POST …/workflows/runs/{runId}/cancel` can stop scheduling further
node work at the next node boundary. The run's `WorkflowRunFinished` should then
record that it was cancelled rather than that it failed — an operator stopping a
run is not an error, and the boot sweep's "interrupted by a host restart"
wording must stay distinct from it.

## Related

- #371 — put run progress on the event stream and minted the run id this needs.

oxoxDev:
## Correction to a premise here, from QA triage (2026-08-04 findings, re-checked against `main` @ `4a38e3b`)

This issue currently states:

> closing the tab or losing the connection abandons the response, **though the run itself continues host-side and does still journal its outcome**

That does not hold. The run dies with the connection, and its outcome is **not** journaled. This turns #383 from an ergonomics issue into a correctness one.

### Why

`run_workflow` awaits the engine **inside the request future**, unspawned — `src/server/ops/workflows.rs:809`:

```rust
let run = match runner.run(company.id(), &file, input, &ctx).await {
```

and `record_run_finished` runs only *after* that await returns (`:815` error arm, `:834` success arm).

[...]

PR #532, opened 2026-08-10
feat(console): show a workflow run's output, and an actionable no-inference banner (#528)

### LINKED ISSUE #528: workflows: a detached run's output is unreadable in the console
## Summary

An operator can run a workflow from the console and then has **no way to read what it produced** — the generated report / per-node agent text is unreachable after the fact.

Found in the workflow-subsystem gap audit. Related: #383 (detached runs), #418 (agent-tool truncation; this is the console twin, previously untracked).

## Why it happens

- `WorkflowsView.tsx` hardcodes `{ detach: true }` on every run; on any #383-capable host `isDetached(res)` is true and `setResult` is never called.
- `RunResultPanel` — the only surface rendering per-node agent text — is gated `{result && …}`, so it renders **only** on the legacy synchronous path.
- The full `output` exists **only** in the synchronous 200 body. Journal `WorkflowRunFinished`, SSE frames, and `GET …/workflows/runs` are structural-only (node output text deliberately absent); there is no run-detail route.

## Fix (approved — bridge now, durable follow-up tracked)

- **Now (frontend):** drop `{ detach: true }` so the settled body reaches `setResult` and `RunResultPanel` renders. Since #383 the sync path runs on a spawned server task and survives a dropped connection; the only cost is the body must arrive. Keep the `isDetached` branch as the compat seam; re-adopt the run id for the mid-run Stop button.
- **Follow-up (separate PR):** durable backend run-output store + `GET` run-detail route, so output survives a proxy idle-timeout on long hosted runs.

## Also here

Map #514's new `inference_required` code to a persistent actionable "Set a provider → Inference" banner instead of a vanishing raw-string toast.

oxoxDev:
## Summary

Two console fixes so an operator can actually see what a run does.

1. **Output was unreadable.** On any modern host every run is detached, and the only panel that renders a run's per-node output text mounts solely on the legacy synchronous path — so you could run a workflow and never read what it produced. The full output exists **only** in the synchronous response body; the journal, SSE frames, and runs-list carry no output text, and there is no run-detail route.
2. **The no-inference error was a dead end.** A run refused for missing inference showed a vanishing raw-string toast ("not wired in this deployment").

Closes #528. Consumes #514's `inference_required` contract.

## API Or Behavior Changes

- The console runs workflows **synchronously by default** so the settled body reaches the result drawer. Since #383 the server's sync path runs on a spawned task and survives a dropped connection, so the only cost of sync is that the body must arrive; the `detach` option is retained for other callers. The `isDetached` branch stays as a compatibility seam, and the mid-run Stop button is preserved by adopting the live run id.
- A run refused with `code: "inference_required"` or `"restart_required"` now renders a **persistent, actionable "Set up inference" banner** (`href="#/settings/connections"`), keyed on the structured `code` only — never the prose. `not_wired` is deliberately excluded (on a genuinely execution-less build the link would lie).
- A connection loss mid-run shows an informational toast and keeps the live view instead of a hard error.

## Tests

- [x] `npm run typecheck` · `typecheck:unit` · `typecheck:e2e`
- [x] `npm run build`
- [x] `npx vitest run` — 169/169 (10 new in `workflow-run-error.test.ts`)
- [ ] `cargo *` — N/A: frontend-only change.

New e2e: `workflow-run-result.spec.ts` (live-brain lane — asserts the drawer shows node text), `workflow-run-inference.spec.ts` (both console lanes — 409 injected via `page.route`). Manual live-host verification of both surfaces is pending (see checklist below).

## Documentation

`runWorkflow` API doc notes the sync-by-default rationale. A durable backend run-output store + run-detail read route is a tracked follow-up (out of scope here).

### DIFF
--- a/frontend/src/api/workflows.ts
+++ b/frontend/src/api/workflows.ts
@@ -308,10 +308,21 @@ export function getWorkflow(
 /**
  * Runs a workflow.
  *
- * With `detach` the host answers as soon as the run has an id (`202`), instead
- * of holding the request open for the whole run — which is what stops a proxy's
- * idle timeout from severing a healthy multi-minute run. The outcome then
- * arrives over the event stream and in {@link listWorkflowRuns}.
+ * **The console omits `detach` on purpose (issue #528).** A run's full `output`
+ * is carried ONLY by the settled `200` body: the journal, the SSE frames, and
+ * {@link listWorkflowRuns} are all structural (per-node status and timing, no
+ * agent text), and there is no run-detail route to fetch the output back later.
+ * The one surface that renders what a run produced mounts on that body, so the
+ * console must hold the request open to receive it. This is affordable because
+ * since #383 the host runs even the synchronous path on a spawned server task —
+ * the run itself survives a dropped connection; only the answer rides the wire.
+ *
+ * With `detach` the host instead answers as soon as the run has an id (`202`),
+ * without holding the request open — which stops a proxy's idle timeout from
+ * severing a healthy multi-minute run, at the cost of the console never seeing
+ * the output. The option is kept for callers that only need the run to start
+ * (e.g. a fire-and-forget trigger) and will read the outcome from the stream or
+ * the history.
  *
  * **Always check {@link isDetached} on the result**, whatever you asked for: a
  * host predating #383 ignores the flag and answers with the settled run, and
--- a/frontend/src/views/WorkflowsView.tsx
+++ b/frontend/src/views/WorkflowsView.tsx
@@ -16,6 +16,7 @@ import {
   Loader2,
   Pencil,
   Play,
+  Plug,
   Plus,
   RotateCw,
   Square,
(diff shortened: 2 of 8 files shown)


oxoxDev:
**Live smoke** — union of #531/#532/#533/#534 merged on fresh `main` (clean, zero conflicts), real `opencompany` binary built `--features openhuman,tinycortex,mcp` serving the built console, mock brain:

- `workflow-run-result` ✓ — ran a workflow end-to-end; the result drawer rendered per-node output. This is the G3 fix live — on a detached run (every modern host) that drawer never mounted before.
- `workflow-run-inference` ✓ — the persistent, actionable inference banner renders on the `inference_required` code (no vanishing toast).

`2 passed (7.9s)`, live-brain lane. The real backend `409 inference_required` path (not the page.route mock) is covered by #531's route-level test T1.

PR #533, opened 2026-08-10
fix(workflows): make the default marketing preset runnable, and wire web_search into tool_call nodes (#530)

### LINKED ISSUE #530: workflows: the default marketing preset workflow is authored broken and halts on Run
## Summary

`companies/agentic_marketing_agency/workflows/campaign_pipeline.toml` — the workflow behind the **default desktop preset** (`DEFAULT_PRESET_ID = "agentic_marketing_agency"`) — halts on the first Run, so it is the single most-likely first workflow a new operator runs and it cannot succeed. Of all 19 bundled templates it is the only one using a `tool_call` or `http_request` node, and both are placeholders.

Found in the workflow-subsystem gap audit (not previously tracked).

## Why it happens

- `research` (kind `tool_call`) has no config → slug falls back to node id `research`; `namespace_of("research")` is `None` → "not a wired workflow tool" (`src/workflows/caps/tools.rs`).
- `publish` (kind `http_request`) has no url → tool error.
- `gate` (kind `condition`) has no `field` → truthiness-checks the whole strategist envelope (always truthy) → every run takes the "yes" branch into `research` and halts under the default `on_error = stop`.
- Deeper: `WorkflowToolInvoker` never wires the metered `search` tool family, so even with `slug = web_search` + the existing `search` grant, `tools.get("web_search")` is `None`.

## Fix direction (approved)

- `research` → real `tool_call` slug `web_search`, `on_error = continue`, args `query = "=item.text"`.
- `publish` → `agent` node bound to `copywriter` (assemble the publish-ready post; sign-off is the human's job — no fake CMS POST).
- Wire the search tool family into `WorkflowToolInvoker::new`, mirroring `build_agent`'s two hard gates (explicit `search` grant, never `*`; managed backend present; fail-closed) — **mirroring OpenHuman's tool-wiring pattern**. Enables `web_search` for every company's workflow `tool_call` nodes.
- Add a regression test pinning every bundled workflow's `tool_call` slugs to known namespaces and `agent` refs to the company roster, so this bug class cannot recur.

oxoxDev:
## Summary

The default marketing preset's `campaign_pipeline` halted on the first Run — the most-likely first workflow a new operator ever runs. `research` (a `tool_call` with no config) fell back to the node id as its slug and resolved to no tool; `publish` (an `http_request`) pointed at a nonexistent CMS; and the real root cause: the workflow tool belt never wired the metered `search` family, so even a correct `web_search` slug failed "not available" despite the company granting `search`. This wires search into the workflow belt (mirroring the agent builder), makes the preset runnable end to end, and pins the whole class with a regression test.

Closes #530.

## API Or Behavior Changes

- `WorkflowToolInvoker` now wires the `web_search` tool for a `tool_call` node when the company **explicitly** grants `search` (never via `*`) **and** a managed search backend is present — fail-closed with a warn otherwise, mirroring `build_agent` (`build.rs:415-431`). This enables `web_search` in every company's workflow `tool_call` nodes under the same gates as agents.
- The invoke-time gate for the `search` namespace now requires an explicit grant (`grants_search_explicit`, was `grants_cover`), so the run-time gate matches construction.
- Preset graph: `research` → `tool_call` slug `web_search` (`on_error = continue`, degrades gracefully with no credential); `publish` → `agent` bound to `copywriter` (assemble the publish-ready post; sign-off stays the human's job).

## Tests

- [x] `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` — 0 warnings
- [x] `cargo test --features openhuman,tinycortex` — `workflows::` 189, `company::content_test` 9
- [x] `cargo check --features openhuman,mcp,telegram` · `cargo check --all-features`

New regression sweep: every bundled workflow's `tool_call` slugs resolve via `namespace_of` and every `agent` ref is on that company's roster — so an unrunnable preset cannot ship again. Manual end-to-end run of the preset is pending (see checklist).

## Documentation

Module/struct docs enumerating the wired tool families updated (were shell/code/web only).

### DIFF
--- a/companies/agentic_marketing_agency/workflows/campaign_pipeline.toml
+++ b/companies/agentic_marketing_agency/workflows/campaign_pipeline.toml
@@ -36,6 +36,21 @@ id = "research"
 kind = "tool_call"
 name = "Web research"
 summary = "Pulls sources and competitor takes."
+# `on_error` is a first-class node field, NOT a `config` key — the validator
+# rejects reserved keys inside `[node.config]`. `continue` means a search that
+# fails (e.g. no managed search credential on this install) becomes an error
+# item on the default port and the run proceeds to `copy`, instead of halting
+# the whole pipeline. web_search degrades gracefully with no credential.
+on_error = "continue"
+[node.config]
+# The metered discovery tool wired by the workflow tool belt (issue #238); the
+# `web` namespace can only read a URL the agent already has, never find one.
+slug = "web_search"
+[node.config.args]
+# Bind the strategist's reply as the query: `=`-expressions resolve against the
+# input item, and an agent node returns `{ text, agent_ref }` (see caps::mod).
+query = "=item.text"
+max_results = 5
 
 [[node]]
 id = "copy"
@@ -52,10 +67,14 @@ summary = "Generates the hero image."
 agent = "creative_director"
 
 [[node]]
+# No CMS exists to POST to, and the human sign-off is `[company].human_role`, so
+# this is an agent assembly step, not an HTTP call. Reusing `copywriter` across
+# two nodes is legal — the id stays `publish` so the edges are untouched.
 id = "publish"
-kind = "http_request"
-name = "Publish to CMS"
-summary = "POST the approved draft."
+kind = "agent"
+name = "Assemble final post"
+summary = "Assemble the publish-ready post and hero-image reference, then hand off for operator sign-off."
+agent = "copywriter"
 
 [[node]]
 id = "done"
--- a/src/company/content_test.rs
+++ b/src/company/content_test.rs
@@ -250,3 +250,117 @@ fn the_repo_skill_registry_parses() {
         );
     }
 }
+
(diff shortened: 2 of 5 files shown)


oxoxDev:
**Live smoke** — union built and booted (real binary, `--features openhuman,tinycortex,mcp`, mock brain). The workflow run path executes end-to-end (`workflow-run-result` ✓).

Coverage note for this PR specifically: the marketing preset is asserted **runnable** by the new `content_test` sweep (every bundled workflow's `tool_call` slugs resolve via `namespace_of` + every `agent` ref is on-roster), and the `web_search` wiring by the `caps` unit tests (explicit-grant + backend gates, fail-closed). A true end-to-end run of `campaign_pipeline` with a real managed `web_search` credential is credential-gated and not run here — but `on_error = continue` means the run degrades gracefully to the agent nodes and reaches `done` without it, which is the behaviour that was previously a hard halt.

PR #536, opened 2026-08-10
feat: split OpenCompany into an orchestration server and a desktop client

senamakel:
Splits OpenCompany into an orchestration **server** and a **desktop client**, per the design in this branch's plan. The console stays one codebase; the server stays one binary. What's added are the seams a desktop needs to hold several hosts at once, speak ACP to locally-installed harnesses, and run a host in-process.

## Why this shape

Three findings from reading `block/buzz`, `paperclipai/paperclip` and the ACP spec drove the design:

- **buzz is the cautionary example for multi-host.** Its desktop *looks* like this — a rail of workspaces — but only one is live: `AppState` holds a single `relay_url_override` and its retention DB is scoped per-community, so switching is a stateful re-apply. The failure isn't the rail, it's the singleton behind it. Everything here is keyed by connection, and the registry deliberately has **no "active connection"**.
- **paperclip is right about isolation.** Git worktrees per task, not a shared repos dir — with N hosts dispatching to one machine, a shared tree is a data race with no attribution.
- **ACP's `Proxy`/`Conductor` cannot carry runner dispatch.** Spiked it: `agent-client-protocol-conductor` takes `components: Vec<String>` where *"the final command must be the agent"* and spawns the chain at `initialize` — a fixed linear chain to exactly one upstream. Dispatch needs the upstream chosen **per session**. The routing is ours; `AcpAgent` is a port so a runner is just another implementation.

## Server

- **Header-form session auth.** A `SameSite=Lax` cookie is never sent cross-site, so a desktop had no way to authenticate *as a person* — only a platform bearer, which `ScopedCompany` maps to `actor: None`, writing to the journal anonymously. `resolve_claims` still structurally cannot return a human.
- **Device pairing** as a `SessionKind::Device` session rather than a 15th storage port. All three backends already persist `SessionRecord` as JSON, so this needed no migration — and, more importantly, `delete_for_user` (suspension, admin reset) already revokes devices with no second code path to remember. Pairing codes and login codes share a store but are separated by a hashing domain prefix, so neither can be redeemed as the other.
- **`/spec` identity.** A random, persisted `instance_id` so a client can tell two hosts apart across URL changes. Random rather than derived: `/spec` is unauthenticated, and a hostname-derived id publishes topology.
- **`/acp` reserved**, so a build without the feature 404s instead of answering an ACP client's probe with the console shell and a `200`.
- **`AcpRunTurn`** — a second `RunTurn` implementation over an `AcpAgent` port. Unlocks a desktop company with no inference key (runs on the operator's own Claude subscription), reverse dispatch, and any other ACP harness.
- **Runner lane** — outbound dial, Ed25519 handshake with owner attestation, presence-with-TTL, at-most-one-per-scope, and dispatch that rewrites session ids.

## Desktop (`src-tauri`)

A **separate workspace, not a member** — making it a member would put the whole Tauri tree into `cargo metadata --locked`, which CI runs first on a runner with no webkit/GTK. Verified: the root graph is 893 packages and `tauri` is not among them.

- An in-process host on `127.0.0.1:0`, holding the data root's lock.
- All HTTP and events **proxied through Rust**: no per-server CORS configuration (an onboarding cliff for the headline feature), and the device token never enters the webview.
- An ACP client over stdio, with **path confinement enforced below the UI** — a renderer decides what a person sees and must never decide what a model can reach.

## Console

- One `Transport` seam. `BrowserTransport` is a literal restatement of the previous inline `fetch`/`EventSource`; the browser build's behaviour is unchanged.
- N connections, each with its own status. A host being unreachable reddens one rail row instead of blanking the app.

## Bugs found and fixed along the way

These were pre-existing and are the part I'd most want reviewed:

1. **The SQLite conformance suite ran nowhere.** 24 tests covering isolation, append-only, export totality and eleven ports — read as a guarantee, executed by no lane (`cargo test` omits the feature; `--all-features` is check-only). Issue #475's pathology, in the backend the desktop makes primary. Now has a lane, and I confirmed the lane goes red when a port regresses.
2. **Nothing enforced the documented single-writer rule** on a data root. Two hosts would overwrite each other's companies. Now an OS advisory lock (`fs2`, already transitive — the lockfile delta was one line).
3. **`HOME`-only data-dir resolution** fell back to a *relative* path on Windows — resolved against the launcher's working directory, and silently different per launch.
4. **`app::journal::prepare`'s `set_var` is UB** in a process that already has threads. The embedded path uses the non-mutating half and returns the value for `main` to export.

## Verification

| | |
|---|---|
| Host, default | 1702 |
| Host, `openhuman,tinycortex --tests` | 2546 |
| Host, `sqlite` / `runner` | 24 / 71 |
| Desktop | 49 unit + 12 ACP + 7 proxy |
| Console | 182 unit, 3 typechecks, **94 e2e / 0 failed** |
| fmt + clippy | clean on default, sqlite, openhuman, runner, all-features |

Security-critical guards were **falsified**, not just asserted — path confinement, the pairing/login domain separation, the device-reproduction guard, connection-scoped storage keys and the `/acp` reservation each fail their tests when the guard is removed.

## Things needing a decision

- **Overlap with #526 (desktop-launcher).** `main.tsx` now calls a Tauri `desktop_config` command, but there is no `src-tauri` on main — the Rust half isn't in this repo. That's a *different* desktop approach: single host, magic-link `dev_code` redemption via `?api=&code=`. Mine is N hosts with device pairing and a Rust proxy. They don't break each other (`bootstrapDesktop` swallows the missing command), but the tree shouldn't keep both. **Which one do we want?**
- **`src-tauri` icons are placeholder** solid-colour PNGs. Real artwork is a design task.
- **Squashed to one commit deliberately.** The auto-commit hook produced 150 commits whose generated subjects describe the wrong changes — e.g. `ci: remove CI workflow` on a commit that is 32 insertions and zero deletions.

## Deliberately not done

- No entry point to add the *first* extra host in a browser (the rail hides below two connections; the degrade test seeds via `localStorage`). Belongs with the desktop, where multi-host is the product.
- `#/c/{connId}/…` URL grammar deferred — connection ids aren't portable between machines.
- Lifting `RunTurn` above `HarnessBrain`, so a runner-only server needn't link `openhuman`, remains the destination.
- **No incremental assistant text over ACP.** `stream_event_from` returns `None` for `TextDelta`; adding deltas is a change to what this host will stream, not plumbing.

## Summary by CodeRabbit

* **New Features**
  * Added a desktop application with embedded hosting, multi-connection support, connection status indicators, and host switching.
  * Added device pairing, device management, labels, and revocation.
  * Added ACP sessions with agent communication, approvals, workspace isolation, and filesystem protections.
  * Added runner identity verification, availability tracking, and task routing.
  * Added instance identity, display name, capabilities, and storage details to the service specification.
* **Bug Fixes**
  * Improved connection-specific error handling and isolated local settings across hosts.
  * Added Windows profile-directory fallback support.
* **Tests**
  * Expanded coverage for desktop connections, authentication, transport, ACP, proxying, storage, and device security.


oxoxDev (commented):
Superseded — re-filed the same review as changes-requested with a checklist at the top. See the next review on this PR.

oxoxDev (changes requested):
## Required before merge

Blocking on these; everything below is the reasoning.

- [ ] **Wire the desktop, or drop `src-tauri` from this PR.** Nothing calls `oc_connect`, so every proxied request fails — the shell cannot complete one round trip today.
- [ ] **Stop persisting the platform bearer to `localStorage`** (`connections/registry.ts:187`). Straight security regression against main.
- [ ] **Close the dangling-symlink write escape** in `acp/confine.rs`, with a test that plants one.
- [ ] **Add a CI lane that builds and tests `src-tauri`**, or drop it — 68 tests currently gated by nothing.
- [ ] **Don't let the webview override the credential header** (`proxy/mod.rs:178`).
- [ ] **Fix per-chunk `from_utf8_lossy`** on the SSE lane (`proxy/mod.rs:255`).
- [ ] **Set a CSP** in `tauri.conf.json`.
- [ ] **`git rm --cached instance-id` and `src-tauri/gen/`**, extend `.gitignore`.
- [ ] **Either mount `server::acp` / `acp_run_turn` / `runner`, or land them separately.** ~2.2k lines with zero callers, and the `acp` feature the comments reference does not exist.
- [ ] **Answer the hosted rolling-deploy question** on the home lock before it reaches a tenant.
- [ ] **Update `docs/spec/` + `docs/modules/`** per CLAUDE.md.

Not blocking: the nits and the two questions at the bottom.

---

## Walkthrough

Read the whole diff. The reasoning in the module docs is the best in this repo — the buzz/paperclip/ACP-spike writeup, the `SessionKind`-not-a-new-port argument, and the four pre-existing bugs are all correct and worth landing on their own.

Longer than my usual review because the PR is 22k lines across two workspaces, touches session auth, and adds a second HTTP client with a credential on it.

The split I'd draw: **the server + console halves are wired and good; the `src-tauri` and ACP/runner halves are not wired to anything and no CI lane builds them.** Concretely, ~2.2k lines of server code (`src/server/acp/*`, `src/harness/acp_run_turn.rs`, `src/runner/*`) have zero callers, and the desktop cannot complete a single request today. That's a scope call, not a nit — details below.

## Blockers

- **`src-tauri/src/lib.rs:104` — the desktop cannot make one successful request.** Nothing anywhere calls `oc_connect`, so `ProxyRegistry` is always empty and every `oc_request` from `ProxyTransport` (`frontend/src/api/transport/proxy.ts:68`) returns `UnknownConnection`. `oc_disconnect` / `oc_connections` / `oc_embedded` have no callers either, so the embedded host's `127.0.0.1:<port>` never reaches the console. Separately `main.tsx:20` invokes `desktop_config`, which isn't in `generate_handler!`. The desktop shell is scaffolding until a bootstrap registers the embedded host and any remembered profiles.

- **`frontend/src/connections/registry.ts:187` — the platform bearer is now persisted to `localStorage`.** `addConnection` writes the whole `Credential` through `saveProfile`, and `App.tsx:117` builds `{kind:"platform", token}` from `config.operatorToken`. On main that token was memory-only; `stripAuthParams` deletes `token` from the URL precisely because it's sensitive, and this writes it to `oc.connections.v1` where it survives forever. `profileStore.ts` opens with "What is deliberately not stored: Secrets." Suggest persisting a marker and re-reading the token from the URL/env each load:
  ```ts
  credential: connection.credential.kind === "platform" ? { kind: "cookie" } : connection.credential,
  ```

- **`src-tauri/src/acp/confine.rs:100` — a dangling symlink escapes the write confinement.** With `<root>/link -> /outside/planted.txt` (target absent), `path.canonicalize()` fails, so the parent branch runs, the parent is `<root>` and passes `require_inside`, and `resolve_write` returns `<root>/link`. `tokio::fs::write` then follows the link and creates the file outside the root. Verified the OS semantics directly. The existing symlink tests all use links whose target exists, so canonicalize succeeds and they take the other branch. Fix:
  ```rust
  let target = parent.join(name);
  // A DANGLING symlink does not canonicalize, so it arrives here looking like a
  // file that does not exist yet — and `fs::write` follows it out of the root.
  if std::fs::symlink_metadata(&target).is_ok_and(|m| m.file_type().is_symlink()) {
      return Err(ConfineError::Escapes);
  }
  Ok(target)
  ```
  Worth a test that plants a dangling link, and a note on the residual TOCTOU (a link planted between resolve and write still wins; `O_NOFOLLOW` on the final open closes it).

- **`.github/workflows/ci.yml` — no lane builds or tests `src-tauri`.** No job in any workflow references it, and it's a separate workspace so root `--all-features` doesn't reach it. That's ~2k lines of Rust and the 68 tests cited in the description, gated by nothing — the exact #475 pathology this PR diagnoses and fixes for the SQLite suite two hundred lines earlier. A `macos-latest`/`ubuntu` step running `cargo test --manifest-path src-tauri/Cargo.toml` (webkit/GTK deps on Linux) would close it.

## Major

- **`src-tauri/src/proxy/mod.rs:178` — the webview can override the credential it isn't supposed to see.** Caller-supplied `request.headers` are applied first, then `apply_credential` adds the session header. `reqwest`'s `RequestBuilder::header` **appends**, and axum's `headers.get(SESSION_HEADER)` returns the *first* value — so a `x-opencompany-session` in `req.headers` (passed straight through by `ProxyTransport`) wins over the connection's own credential. Same for `authorization` vs `Credential::Platform`. Build the `HeaderMap` yourself, drop `x-opencompany-session` / `authorization` / `cookie` / `proxy-authorization` from the caller set, `insert` the credential last, then `.headers(map)`.

- **`src-tauri/src/proxy/mod.rs:255` — SSE payloads with non-ASCII get corrupted.** `String::from_utf8_lossy(&chunk)` runs per TCP chunk. `SseDecoder` is carefully chunk-safe for frame boundaries, but a multi-byte codepoint straddling a read boundary is replaced with U+FFFD before the decoder ever sees it — so an agent reply containing an emoji or accented text silently mangles under load, which is the same failure mode the decoder's own module doc argues against. Either buffer trailing partial bytes, or take `&[u8]` in `SseDecoder::push` and decode after framing.

- **`src-tauri/tauri.conf.json:22` — `"csp": null`.** No Content-Security-Policy in a webview that renders agent-authored markdown, while `proxy/mod.rs:12` justifies the whole design with "a cross-site-scripting bug in rendered agent markdown cannot exfiltrate N hosts' credentials". True for the token, but with no CSP an injected script can load remote code and `invoke("oc_request", …)` against every connection — it doesn't need to read the token to use it. `gen/schemas/capabilities.json` is `{}` and there's no `capabilities/` dir either.

- **`src/server/acp/`, `src/harness/acp_run_turn.rs`, `src/runner/` — ~2.2k lines with no callers.** No `router()` exists in `src/server/acp/`, nothing merges it, and `/acp` is only in `RESERVED_PREFIXES` — so it 404s in *every* build, including all-features. `routes.rs:22` and `acp/mod.rs:17` both reason about "a build without the feature", but there is no `acp` feature in `Cargo.toml`; both modules compile unconditionally. `RunnerDispatch`/`RunnerRegistry`/`verify_hello` are referenced only from inside `src/runner/`. The description reads these as delivered ("Unlocks a desktop company with no inference key", "dispatch that rewrites session ids") — today they're a library nothing can reach.

- **`src/runner/attest.rs:104` — the attestation proves provenance but authorises nothing, and the named guard doesn't exist.** `verify_hello` verifies that *some* key signed "this runner acts for me"; nothing checks that `attestation.owner` is an owner of this host, and `conditions` (documented as carrying companies, an expiry, a parallelism cap) is never parsed. A self-generated keypair self-attesting passes. The doc says the caller must reject replayed challenges via `NonceCache` — `NonceCache` does not exist anywhere in the tree. Harmless while unwired, which is exactly the trap: the wiring commit inherits a checked-looking handshake with the checks outside it.

- **`instance-id` and `src-tauri/gen/schemas/*` are committed runtime/generated artifacts.** `instance-id` at the repo root is a real minted id (from a `serve` run with home=cwd) — so every clone now boots with the *same* public instance id, defeating the property the file exists to provide. `.gitignore` covers `frontend/src-tauri/gen/`, but the tree moved to `src-tauri/`, so 5,225 lines of generated schema landed too. Add `instance-id`, `.lock` and `src-tauri/gen/` to `.gitignore` and `git rm --cached` all three.

- **`src/bin/opencompany.rs:890` — the home lock changes hosted deploy behaviour.** During a rolling update the new tenant pod and the old one overlap on the same `/data`; the new one now fails at boot with a `Config` error instead of starting. Turning a silent race into a loud refusal is right, but it wants a check against the manager's deploy strategy (and probably a `Recreate` strategy or a note in `docs/spec/runtime/storage.md`) before it ships.

- **No documentation changes.** Zero files under `docs/`. `CLAUDE.md` asks for `README.md` / `docs/spec/README.md` / `docs/modules/` to stay aligned with code changes, and this is the largest architectural change in the repo — new session carrier, device pairing, `/spec` identity, the data-root lock, a second workspace.

## Nits

- `src/server/users/devices.rs:243` — `claim_device` mints a session without the `session_cookie_name` check `mint_session` does; a company id that can't carry a header gets a token that can never be presented.
- `Cargo.toml:65` — `fs2` was last published in 2020. `File::try_lock_exclusive` is in std since 1.89 and the toolchain here is 1.95, so the dependency is avoidable entirely.
- `src-tauri/src/proxy/mod.rs:129` — `reqwest::Client::default()`: no timeouts and default redirect-following. `x-opencompany-session` is a custom header, so reqwest does not strip it across a cross-host redirect — a host can bounce the proxy at a third party carrying the header. `redirect::Policy::none()` plus a request timeout.
- `src/server/users/devices.rs` — no per-user device cap and no rate limit on `/devices/claim`. 256-bit codes make guessing a non-issue; enumeration and unbounded enrolment aren't.
- One squashed 22k-line commit against `CLAUDE.md`'s "commit each coherent, validated slice on its own". The bad auto-generated subjects are a real problem, but rewriting ~10 subjects beats losing all bisect granularity on the largest change in the repo.

## Questions

- The embedded host binds loopback with `AppConfig::default()` — no CORS, and whatever the default auth posture is for a company with no `[users]`. Can any other local process on the machine drive that company without a credential?
- On #526: agreed the tree shouldn't keep both. My read is yours is the right target shape (N hosts, device pairing, Rust proxy) — but that's the argument for landing the server + console halves now and the desktop when it's wired, rather than merging both trees' worth of half-connected desktop.

## What I'd suggest

Split, don't shrink. Land as-is: the session header carrier, `users::devices`, `store::lock`, the `USERPROFILE` fix, `app::boot`'s `set_var` fix, `/spec`, and the SQLite conformance lane — that half is wired, tested by a lane that runs, and fixes four real bugs. Hold `src-tauri` until a bootstrap calls `oc_connect` and a CI lane compiles it; hold `server::acp` + `harness::acp_run_turn` + `runner` until something mounts them. Reviewing 4k lines of unreachable code is reviewing a design doc that happens to compile.

oxoxDev:
Pushed the review items as #543, based on `acp-split` so it folds straight in rather than competing with this.

Eleven commits, one concern each; every guard falsified (the test fails when the check is removed). The four blockers: the desktop could not complete a single request — nothing called `oc_connect`, and separately `isDesktopRuntime()` probed a different global than `bridge()` reads; the platform bearer was being written to `localStorage`; a dangling symlink escaped the ACP write confinement; and no CI lane reached `src-tauri` at all. Plus CodeRabbit's two frontend majors.

Three are decisions rather than repairs and want your eye specifically — dropping the `desktop_config` bootstrap from `main.tsx` (settles the #526 overlap the way your writeup argues), not forwarding a `device` credential until something resolves a keychain handle, and making `acp` a real feature so the modules match their own docs. All three are called out in the PR body.

The hosted rolling-deploy question about the data-root lock is written into `docs/spec/runtime/data-root.md` as an open question rather than answered — it needs whoever owns the manager.

Issue #450, opened 2026-08-06
runtime: the embedded writer reports an append failure per item, with no bound and no dedup


oxoxDev:
## What happens

The embedded runtime's durable-append worker reports a failure **per queued item**, via `eprintln!`, with no dedup and no backoff:

```rust
Msg::Item(item) => {
    if let Err(e) = append(item).await {
        eprintln!("tinyagents: {name} durable append failed: {e}");
    }
}
```

When the store's root is unwritable, every append fails identically and the container log becomes that one line.

## Why it needs its own issue

#446 removed the condition that triggered it — the journal root now resolves inside a writable mount and boot refuses if it is not writable, so the loop is never entered. That is a precondition fix, and it is the right one for the failure we actually had: the mount is read-only for the pod's lifetime, so the retry could never have succeeded.

It does **not** bound the writer. A volume that flips read-only *mid-life*, or fills, would still flood exactly as before — and that case cannot be fixed by a precondition, because there is no boot to check.

The reason it was not fixed in #446: this is `eprintln!`, not `log`/`tracing`, so no subscriber-side filter on the embedding side can reach it. The only fix is inside the writer.

## Why the harm is the visibility, not the write

The durability loss matters. The visibility loss matters more, and #446 is the proof: a 200-line tail from a live tenant contained this line and nothing else, so the investigation that went looking for something unrelated found nothing else could be seen.

A durable-write failure should also be loud **once**, not quiet and infinite. `eprintln!` is neither — it bypasses the log level entirely while still being drowned out by its own volume.

## What the fix should do

- Bound identical reports — collapse the repeat, or stop retrying a condition that cannot resolve itself.
- Route through the tracing subscriber so the embedding application can control it at all.
- Report the first failure at a level matching its severity.

## A precedent already in the tree

`vendor/openhuman/src/core/observability.rs` contains a five-minute repeat suppressor (`FS_ERROR_COOLDOWN` / `was_recently_reported`), wired behind the `crash-reporting` feature for the Windows `ERROR_FILE_SYSTEM_LIMITATION` case after an unthrottled retry loop produced 6,050 crash events from one user. The mechanism works and the motivating incident is the same shape as this one — it is scoped to one error code and one feature flag rather than being unavailable.

## Related

Follows #446, which removed the trigger without bounding the writer. This is a vendored change and needs an upstream conversation, not a patch in this repo.

oxoxDev:
**Related — `cluster:journal`**

[...]

PR #537, opened 2026-08-10
feat(ports): a retention policy for the company event journal

### LINKED ISSUE #275: feat(ports): a retention policy for the company event journal

M3gA-Mind:
## Summary

Closes #275.

### Read this part first: a latent correctness bug that had to land with the feature

`FsEventLog::append` derived the next sequence number from the **line count** of `events.jsonl`:

```rust
let seq = existing.lines().filter(|l| !l.trim().is_empty()).count() as u64;
```

Remove any line and every subsequent append reuses a sequence number that a *surviving* row still holds. Sequences are stable ids: `parent` on `OperatorMessage`/`AgentReply`, `ReactionToggled`'s target, and #358's `TaskDiscussionRedacted` tombstone all address a message **by its sequence**. A reused sequence therefore silently re-points a withdrawal tombstone — or a thread parent — at an unrelated message. Nothing errors; the log just quietly means something different.

Retention is the first feature that removes a line, so on the default backend it would have been unsound the day it shipped. That is why the fix is in this PR rather than a follow-up: the feature is not safe without it.

The fix is `max(seq) + 1`. **For a log that has never been pruned the two expressions are identical** — n lines numbered `0..n-1` both yield `n` — so no existing behaviour changes and no stored log needs migrating.

Full audit of the three backends, since the same question applies to each:

| Backend | Allocates from | Safe under pruning? |
| --- | --- | --- |
| `FsEventLog` | line count → **now** one-past-the-highest | **was unsafe**, fixed here |
| `SqliteStore` | `COALESCE(MAX(seq) + 1, 0)` | safe while anything survives |
| `MongoStore` | a persistent `counters` document | always safe |

The shared invariant that makes all three safe: **a pass never removes the highest-sequence entry**, whatever the policy says.

### The feature

Nothing in OpenCompany ever removed a journal entry, so the log grew for the life of a company. `EventLog` gains one method — `prune(id, &RetentionPolicy) -> Result<PruneReport>` — built so the irreversible direction is the hard one to reach:

| Guard | Effect |
| --- | --- |
| `RetentionPolicy::default()` | both bounds `None`, removes nothing — retention is opt-in per caller |
| `CompanyEvent::retention_class()` | exhaustive match with **no `_` arm**, so a new event variant fails the build until somebody classifies it, and `Permanent` is the answer for anything ambiguous |
| `plan_prune()` | one **pure** function all three backends route through, so fs / sqlite / mongodb cannot drift on the one operation that cannot be undone |
| watermark rule | the highest sequence is never removed, so sequence reuse is unreachable |
| default trait body | `Err(Unimplemented("EventLog::prune"))` — a backend without retention says so, rather than reporting a pass that removed nothing. The two are indistinguishable to a caller and only one is true. |

**Which of the three questions in the issue this answers, and how.**

*Is the journal an audit log — append-only, never pruned?* Mostly yes, and #504 (issue #358) already set the posture: withdrawing a discussion post appends a tombstone that supersedes the row rather than rewriting it. So only **4 of 26** kinds are prunable — `WorkflowRunStarted`, `WorkflowRunFinished`, `WorkflowNodeFinished`, `McpCallFailed`. Everything else is permanent, either because it is the audit trail (approvals, lifecycle, payments, tool-access) or because another entry addresses it by sequence.

`WebhookReceived` and `ScheduleFired` are deliberately **not** prunable despite being the highest-volume kinds: a webhook body is the only record of what a counterparty actually sent, and a schedule tick the only evidence a cron fired at all. Both are evidence.

*By age, by count, or by kind?* Kind first — it gates everything — then either bound may select within the prunable set. The count bound is **per kind** (mirroring OpenHuman's `MAX_FLOW_RUNS_PER_FLOW`) so a burst of one kind cannot evict the run outcomes this issue was filed about.

*Does pruning renumber `EventSeq`, and what does that do to cursors and SSE?* It does not renumber. Pruning leaves gaps, which `read_from`'s `seq >=` scan already tolerates — a cursor or SSE subscriber parked on a removed sequence resumes at the next survivor. Renumbering would instead invalidate every stored cross-reference listed above.

**No clock.** The age bound is measured back from the newest entry *in the log*, not from wall-clock now — the same anchoring `ports::usage::retention_cutoff` already uses for usage samples. It makes every case testable without a clock seam and cannot empty a dormant company's journal just because time passed. The journal does diverge from that precedent in one way, deliberately: `UsageMeter` evicts on write against a fixed 90-day window, and the journal never evicts implicitly, because it is what export/import ships and what boot replays.

### Scope, stated so it is not mistaken for unfinished work

**Nothing calls `prune`.** No scheduler, no boot hook, no `company.toml` key. That is the decision, not an omission: the requirement is an explicit policy rather than an implicit default that silently deletes history, and the strongest form of that is retention happening only when something asks for it. Wiring a caller or a config surface is a deliberate follow-up and should be its own reviewable decision.

**#450 is not folded in.** It is the embedded writer's unbounded, undeduped per-item append-failure logging — a log-amplification problem in a different code path, about the OpenHuman runtime journal rather than the `CompanyEvent` log. The two share no surface. It remains open and unassigned.

**No collision:** the journal-adjacent branch `fix/fs-append-atomic-newline` is 0 commits ahead of `main` and 524 behind — stale. Recorded here to save the next person the same check.

## API Or Behavior Changes

**Additive to the `EventLog` trait.** `prune` has a default body returning `Unimplemented`, so the five test-only `EventLog` doubles in `src/` compile unchanged. The three production backends implement it.

New public items, all re-exported from `crate::ports`: `RetentionPolicy`, `RetentionClass`, `PruneReport`, `plan_prune`. New inherent methods on `CompanyEvent`: `kind()` and `retention_class()`.

**One behaviour change:** `FsEventLog::append` now allocates from one-past-the-highest sequence instead of the line count. As above, the two agree exactly for any log that has never been pruned, so this is observable only after a prune. No wire format changes, no serialization changes, no stored log requires migration, and the `session_db` equivalents here — the `events` table/collection and `events.jsonl` — keep their existing shape.

**Nothing is deleted unless a caller asks.** No default policy prunes anything.

## Tests

- [x] `cargo fmt --all -- --check` — run, clean.
- [x] `cargo clippy --all-targets -- -D warnings` — **run as `cargo clippy --lib --tests -- -D warnings`, under both `--features sqlite` and `--features mongodb`; clean on both.** Not run with `--all-targets` (examples/benches) — disk is a shared constraint on this machine and the change touches no example or bench target. CI covers the full invocation.
- [x] `cargo build --all-targets` — **N/A — not run locally, deliberately.** No full build matrix on this machine (shared disk). Compilation of every changed path is covered by the `cargo test --lib` and `cargo check --lib --tests --features mongodb` runs below, both of which built the crate. CI is the gate for the full matrix.
- [x] `cargo test` — **run as `cargo test --lib --features sqlite`: 1680 passed, 1 failed.** The single failure is pre-existing and unrelated — see below. Integration targets under `tests/` were not run locally; this change adds none and touches none.

What was actually exercised:

- **12 unit tests on `plan_prune`** (`src/ports/events.rs`), against synthetic timestamps with no clock: default-is-a-no-op, empty log, newest-anchored cutoff, an entry exactly on the window edge being kept, permanent-survives-any-policy, per-kind budgeting, the watermark rule, the two bounds behaving as a union rather than a double count, input-order independence, and `kind()` pinned against the tag `serde` actually emits.
  - `an_age_evicted_entry_does_not_consume_a_count_slot` deliberately puts sequence order and timestamp order in **disagreement** (what backfill and clock skew produce), because that is the only arrangement that can distinguish the two plausible counting rules. With sequence and time correlated, both implementations agree and the test would prove nothing.
- **`assert_event_retention`**, a new backend-agnostic conformance assertion wired into all three backends: the default policy is inert, permanent kinds survive, surviving sequences are unrenumbered, `read_from` resumes past a gap, and a later append does not reuse a survivor's sequence.
- `cargo check --lib --tests --features mongodb` — clean.

Three things a reviewer should know rather than assume:

1. **The mongodb conformance test compiled but did not execute.** The suite self-skips without a live server (`let Some(s) = store().await else { return }`), so this is **not** a green run for that backend. Its `prune` is the same `plan_prune` + delete-by-exact-sequence shape as sqlite's, and both route through the same pure planner, but that is an argument from construction, not from execution.
2. **`openhuman::launcher::tests::tauri_cli_is_fresh_when_binary_is_newer_than_sources` fails, and is not introduced or addressed here.** It is a filesystem-mtime freshness check with no relation to this change; I confirmed it fails identically on a stashed, clean tree at the same base commit.
3. **The age bound is deliberately absent from the conformance suite.** `append` stamps `at_millis` from the wall clock, so a backend test cannot place an event in the past without a clock seam none of the three backends has. Age selection is a property of `plan_prune`, which is pure and pinned against synthetic timestamps instead; what a *backend* must prove is that it removes exactly the entries that function names, which is what the conformance assertion covers.

## Documentation

- `docs/spec/runtime/ports-state.md` — the `EventLog` contract gains a **Retention (issue #275)** section: the guard table, the classification and why only four kinds are prunable, the no-renumbering rule with the cursor/SSE consequence, and the newest-anchored age bound with its deliberate divergence from `UsageMeter`.
- `docs/spec/runtime/events.md` — the "properties that are load-bearing" list goes from two to three, adding that a sequence is a stable id and retention keeps it that way, with a pointer to the section above.

**On the 500-line convention:** `events.md` was **already 509 lines** before this PR, over the repo's cap; this change adds 6, taking it to 515. **This PR does not introduce the violation.** The 6 lines are kept because the per-variant retention rule genuinely belongs in the event-vocabulary doc, which is where a reader looks for what happens to old entries. Splitting that file is worth doing but is unrelated to retention and would bury this diff — offered as a follow-up rather than done here. (There is no CI lint for the cap; I checked.)

Rustdoc carries the operational detail that does not belong in the spec: the module docs on `src/ports/events.rs` explain why `prune` is off the append path and why sequences are not renumbered, and `retention_class`'s docs record the reasoning for each side of the split — including why `WebhookReceived` and `ScheduleFired` stay permanent despite being tempting targets.

### DIFF
--- a/docs/spec/runtime/events.md
+++ b/docs/spec/runtime/events.md
@@ -16,7 +16,7 @@ Append-only, replayable, **single-writer**, and company-scoped. Boot replays the
 tail to rebuild in-flight state. Every variant is serialized internally-tagged
 under `kind`, so each JSONL line is self-describing.
 
-Two properties are load-bearing everywhere below:
+Three properties are load-bearing everywhere below:
 
 * **Additive-only.** A new variant, or a new field carrying
   `#[serde(default, skip_serializing_if = …)]`, cannot change how an
@@ -30,6 +30,12 @@ Two properties are load-bearing everywhere below:
   interleave a record and its newline onto one line and brick the next replay).
   Several boot-time sweeps rest on this; see
   [Interrupted runs](#interrupted-runs-issue-371).
+* **A sequence is a stable id, and retention keeps it that way** (issue #275).
+  Only the three workflow-run kinds and `McpCallFailed` may ever be pruned;
+  everything else is permanent, because it is either the audit trail or the
+  referent of a stored `EventSeq` (a thread `parent`, a `ReactionToggled`
+  target, a `TaskDiscussionRedacted` tombstone). Pruning renumbers nothing and
+  leaves gaps. See [Retention](ports-state.md#retention-issue-275).
 
 ## Variants
 
--- a/docs/spec/runtime/ports-state.md
+++ b/docs/spec/runtime/ports-state.md
@@ -73,8 +73,47 @@ pub trait EventLog: Send + Sync {
     async fn read_from(&self, id: &CompanyId, seq: EventSeq, limit: usize)
         -> Result<Vec<StoredEvent>>;
     fn subscribe(&self, id: &CompanyId) -> BoxStream<'static, StoredEvent>;
+    async fn prune(&self, id: &CompanyId, policy: &RetentionPolicy)
+        -> Result<PruneReport>;
 }
 ```
+
+### Retention (issue #275)
+
+`prune` is the only operation that removes a journal entry. Nothing calls it on
+the append path — retention is an operator action, not a side effect of writing
+— and it is built so that the irreversible direction is the hard one to reach:
+
+| Guard | Effect |
+| --- | --- |
+| `RetentionPolicy::default()` | removes nothing; both bounds are `None` |
+| `CompanyEvent::retention_class()` | exhaustive match, no `_` arm — a new variant fails the build until classified, and `Permanent` is the answer for anything ambiguous |
+| `plan_prune` | one pure function all three backends route through, so fs / sqlite / mongodb cannot disagree about what a policy means |
+| the sequence watermark | the highest-sequence entry is never removed, whatever the policy |
+| default trait body | `Err(Unimplemented)` — a backend without retention refuses, rather than reporting a pass that removed nothing |
+
+Only four kinds are `Prunable`: `WorkflowRunStarted`, `WorkflowRunFinished`,
+`WorkflowNodeFinished` and `McpCallFailed`. Everything else is `Permanent`,
+either because it is the audit trail (approvals, lifecycle, payments) or
+because another entry addresses it *by sequence* — thread parents, reaction
+targets, and the `TaskDiscussionRedacted` tombstone from issue #358 all name a
+message by its `EventSeq`, so pruning a referent would dangle a pointer no fold
+can repair.
+
+**Sequences are never renumbered.** Pruning leaves gaps, which `read_from`'s
+`seq >=` scan already tolerates: a cursor or SSE subscriber parked on a removed
+sequence resumes at the next survivor. The watermark rule exists because
+`FsEventLog` and `SqliteStore` both derive the next sequence from the highest
+one present, so an emptied log would hand the next append a number a retained
+cross-reference still holds.
+
+The age bound is measured back from the newest entry in the log, not from
+wall-clock now — the same anchoring
+[`UsageMeter`](ports-console.md#usagemeter) uses, which keeps a pass
+deterministic, testable without a clock, and unable to empty a dormant
+company's journal just because time passed. Unlike `UsageMeter`, which evicts
+on write against a fixed 90-day window, the journal never evicts implicitly:
+it is what export/import ships and what boot replays.
 
 **The `CompanyEvent` vocabulary lives in
 [`events.md`](events.md)** — every variant, the additive-serialization contract
--- a/src/ports/events.rs
+++ b/src/ports/events.rs
@@ -1,10 +1,214 @@
 //! The [`EventLog`] port: append-only, replayable event history.
+//!
+//! # Retention
+//!
+//! The journal is append-only and grows for the life of a company (issue #275).
+//! [`EventLog::prune`] is the only operation that removes an entry, and it is
+//! deliberately awkward to invoke by accident:
+//!
+//! - It is **never** called on the append path. The [`UsageMeter`] precedent
+//!   evicts on write against a hard-coded window; a journal must not, because
+//!   the journal is what export/import ships and what boot replays. Retention
+//!   here is something an operator asks for, not something a write does.
+//! - It takes an explicit [`RetentionPolicy`], and the [`Default`] policy
+//!   removes nothing. A caller that forgets to configure a bound gets a no-op,
+//!   not a silent deletion.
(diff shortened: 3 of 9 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

Issue #529, opened 2026-08-10
workflows: a crashed run cannot resume and re-running re-delivers already-sent outputs


oxoxDev:
## Summary

A workflow run interrupted mid-graph (host crash, panic, mid-graph failure) is settled **Failed** and cannot resume. The operator's natural recovery — re-run — **re-delivers every already-sent report** (emails to real people), because the delivery idempotency ledger is not durable.

Found in the workflow-subsystem gap audit (not previously tracked). Extends #496 (in-lineage idempotency) to survive crashes and span re-runs.

## Why it happens

- The boot sweep (`src/runtime/workflow_outcome.rs`) stamps a synthetic `WorkflowRunFinished{error: INTERRUPTED_BY_RESTART}` — it settles, it does not re-drive.
- Delivery's only durable record is `WorkflowRunFinished.deliveries`, journaled **after** the run returns; the dispatch itself writes nothing durable, and the `Err`/panic paths write no deliveries at all.
- The #496 ledger rides one approval lineage's trigger input and is not persisted, so an independently re-triggered / manually re-run workflow re-delivers.

## Fix (approved — data-loss half; resume deferred)

- Add a durable `CompanyEvent::WorkflowReportDelivered { workflow_id, run_id, node, kind, target }`, appended **write-behind** at the instant each report is sent (and each durable park), in `src/workflows/delivery.rs`.
- On run start, fold the journal (`delivered_by_unsettled_runs`) — deliveries by the trailing suffix of *unclean* runs are owed nothing; the first clean finish clears the set (so periodic/scheduled delivery still fires every cadence). Union with the existing #496 lineage ledger in `src/workflows/runner.rs`.
- **Deferred, named follow-ups:** F1 checkpoint/auto-resume of interrupted runs (engine-level); F2 per-recipient ledger for partial fan-out; F3 detached-panic finish hardening (G10).

Mirror OpenHuman's run-journaling / idempotency patterns where they exist.

PR #549, opened 2026-08-10
fix(events): main is red under full-feature build — cover WorkflowReportDelivered in two matches

oxoxDev:
## Summary

**Main is red under the full-feature (tenant) build** — the staging deploy fails to compile. Two stacked PRs each merged green against their own base, but the merged tree is non-exhaustive:

- #534 (#529) added the `CompanyEvent::WorkflowReportDelivered` variant (unconditional) and updated `kind()`.
- #537 (#275) added a **new** exhaustive `retention_class()` match, branched before #534, so it has no arm for the new variant.

Together, `src/ports/types.rs` has two non-exhaustive matches:

```
error[E0004]: non-exhaustive patterns: `CompanyEvent::WorkflowReportDelivered { .. }` not covered
  --> src/ports/types.rs:1024  (fn kind)
  --> src/ports/types.rs:1079  (fn retention_class)
```

This fix adds the missing arm to both:
- `kind()` → returns the variant tag `"WorkflowReportDelivered"`.
- `retention_class()` → **Permanent**. The delivery ledger's whole purpose (#529) is to stop a re-run from re-sending a report; pruning it would reintroduce the duplicate it exists to prevent.

Surfaced by the `deploy-staging` run for `16a5d4f4` (build-and-push-to-ECR step); no tenant rolled — smoke1/marketing still on the prior image.

## API Or Behavior Changes

None. Exhaustiveness-only fix; no wire-format or behavior change.

## Tests

- [x] `cargo check --lib` — clean (the error is feature-independent; the deploy log listed these two E0004s as the only errors)

## Documentation

None needed — no public surface changed. Follow-up worth filing: the full tenant feature combo (`mongodb,…,mcp,media,composio,imap,telegram,medulla,tinycortex,sidecar`) is only built at deploy time, so exhaustiveness regressions like this pass PR CI and first fail in the deploy — a CI lane building that combo would catch it pre-merge (full-feature CI gap).

### DIFF
--- a/src/ports/types.rs
+++ b/src/ports/types.rs
@@ -1048,6 +1048,7 @@ impl CompanyEvent {
             Self::WorkflowRunFinished { .. } => "WorkflowRunFinished",
             Self::WorkflowRunStarted { .. } => "WorkflowRunStarted",
             Self::WorkflowNodeFinished { .. } => "WorkflowNodeFinished",
+            Self::WorkflowReportDelivered { .. } => "WorkflowReportDelivered",
         }
     }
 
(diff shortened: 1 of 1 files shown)


oxoxDev:
Duplicate of #545 (M3gA-Mind), which repairs the same non-exhaustive CompanyEvent matches and is already CI-green. Deferring to that one to avoid a competing build-fix. Closing.

PR #588, opened 2026-08-11
fix: land the remaining desktop-split review fixes on main

oxoxDev:
## Summary

Brings `acp-split` the rest of the way into `main`. **@senamakel — this is your branch; opened because the work on it missed a merge window, not to take it over. Close it if you'd rather land this differently.**

#536 merged into `main` on 10 Aug at 13:59. #543 (the review fixes) merged into `acp-split` afterwards, and #564 (the follow-up findings) after that. So `main` has the original desktop split, and 18 commits of fixes are sitting on `acp-split` with nothing scheduled to bring them across. `main` moved 134 commits in the meantime; the merge is clean.

Three of the four blockers from the review of #536 are still live on `main` today. The two you fixed yourself — reserved request headers and the platform-bearer redaction — did land, because you pushed them before the merge.

### What is on this branch and not on `main`

**Security**

- **A dangling symlink escapes the ACP write confinement.** `<root>/link -> /outside/x` with no target yet fails `canonicalize`, takes the does-not-exist-yet branch, and its parent is the root — so every check passes and `fs::write` follows the link out of the tree. The existing symlink tests all use links whose target exists, so they take the other branch. Verified against the OS, and the new test goes red when the guard is removed.
- **`csp: null`** in a webview that renders agent-authored markdown. `connect-src` now allows the IPC origin only; all host traffic goes through Rust and needs nothing else.
- **The device token now lives in the OS keychain.** Every layer was written as though it already did — `users::devices` says the client "stores it in the OS keychain", the proxy docs say the token "lives in the OS keychain and is read by this process" — but nothing implemented it, so `oc_connect` took the raw token from the webview. `oc_connect` no longer accepts one at all; the core resolves it by connection id. Shape follows `vendor/openhuman/src/openhuman/security/keyring/`.

**Correctness**

- **The desktop could not complete one request.** Nothing called `oc_connect`, so `ProxyRegistry` was always empty and every proxied call answered `no such connection`. Registration is now awaited by the transport, so a probe cannot overtake it.
- **SSE payloads with non-ASCII were corrupted.** `from_utf8_lossy` ran per chunk, so a codepoint straddling a read boundary became U+FFFD — an agent reply with an emoji arrives mangled, under load, unreproducibly. The decoder now holds the incomplete tail.

**Coverage**

- **No CI job built `src-tauri`** — not even `--all-features`, since it is not a workspace member. ~2k lines and 79 tests built by nothing. There is now a `Desktop` job, and it passes on a real runner.
- **`server::acp` and `harness::acp_run_turn` are behind the `acp` feature** their own comments already described. They compiled into every build, were mounted by no router and run by no lane. This does not mount the endpoint; it makes the tree honest and puts 31 tests under a lane.

**Housekeeping**

- `instance-id` untracked — it is still committed on `main`, so every clone shares one public instance id, which is the one thing that file exists to prevent. Same for `src-tauri/gen/`.
- `docs/spec/runtime/data-root.md` and `desktop.md`; session carriers and device pairing in `users.md`.

## API Or Behavior Changes

- `oc_connect` no longer takes a `deviceSession` argument. The core resolves a paired session from the keychain by connection id.
- `ProxyRegistry::connection` is `base_url`, returning a `String` — the credential does not leave the registry.
- `withGlobalTauri` is on, exposing the Tauri API namespace in the webview. `invoke` was already reachable through `__TAURI_INTERNALS__`, so the surface is materially unchanged, and it now sits behind a CSP.
- A rolling deploy that overlaps two tenant pods over one `/data` now **fails the new pod's boot** rather than silently racing, because of the data-root lock. Open question for the platform, tracked in tinyhumansai/opencompany-microservice#15 and written up in `docs/spec/runtime/data-root.md`.
- `forgetConnection` issues its disconnect on a later tick, sequenced behind any in-flight registration.

## Tests

Run against this branch merged with current `main`:

- [x] `cargo clippy --all-targets --no-deps -- -D warnings` (default and `acp`)
- [x] `cargo test` — 1682 passing
- [x] Feature lanes: `sqlite` 24, `acp` 20 + 11, `runner` 71
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --locked` — 81 passing
- [x] Console: `typecheck` clean, 198 unit tests, `npm run build` clean

CI was green on both #543 and #564, including `Desktop` and `Console E2E (live brain)`.

Not done: nobody has launched a packaged desktop build. The wiring is covered by unit tests against a stubbed bridge and by `proxy_parity.rs` against a real host, but that is not the same as running the app.

## Documentation

`docs/spec/runtime/data-root.md` (new), `docs/spec/runtime/desktop.md` (new), `docs/spec/runtime/users.md` (session carriers, device pairing), `docs/spec/runtime/storage.md` (the side-by-side section moved out and corrected — the lock changed that behaviour), `docs/spec/README.md` index.

`storage.md` was already over the repository's 500-line ceiling, so this splits rather than grows it, and leaves that file six lines shorter than it started.


oxoxDev:
@tinysweeper review

Head moved to `6871b731` (three commits addressing the four findings from the `fa664056` review). No check runs appeared on the new head, so re-reviewing on request.

PR #591, opened 2026-08-11
fix: address the tinysweeper findings on #588

oxoxDev:
## Summary

The four findings from `tinysweeper` on #588 (three of the six comments were the same unpinned-action point). Three commits.

## API Or Behavior Changes

- `registerConnection` now sequences behind whatever is parked under the same connection id, so its `oc_connect` is issued a microtask later rather than synchronously. `forgetConnection` already behaved this way.
- No user-visible change otherwise.

### The findings

**The keychain availability probe could not see the case it was written for.** `os_store_answers` was `OsKeychain::get(..).is_ok()`, and `get` folds `NoStorageAccess` into `Ok(None)` deliberately — for a read, a locked store and an absent entry both mean "no credential here". Reusing it as a probe collapsed the one distinction the probe exists to make: a locked keychain answered `Ok(None)`, the probe called it reachable, `select` committed to `OsKeychain`, and every later `remember_device` failed on write. The documented degrade-to-memory could not happen. The probe now inspects the raw error — `NoEntry` is reachable and empty, anything else is a store not to commit to.

**Registration did not chain onto a pending disconnect.** I fixed this in one direction in #564 and not the other. A re-register racing an in-flight `oc_disconnect` could land first and then be torn down by it, leaving the connection registered in the console and absent from the core.

**A failed unpair was invisible.** The `.catch()` I added in #564 set state that the paired branch cannot render — the error paragraph lived inside the unpaired branch only. A failed forget left the row saying "Paired" with nothing shown, which is the symptom the catch was added to prevent.

**`Swatinem/rust-cache@v2` is a moving tag**, pinned to 6323deb (v2.9.2). Both call sites, not only the desktop lane it was filed against: a file where one job is pinned and its neighbour is not invites the reader to assume the unpinned one was considered. I declined this on #564 as a repo-wide policy question; raised three times on one PR, it is cheaper to just do. `actions/checkout` and `dtolnay/rust-toolchain` are still on tags — a wider sweep is worth doing, but as its own change.

## Tests

- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --no-deps -- -D warnings`
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --locked` — 82 passing
- [x] Console: `typecheck` clean, 199 unit tests, `npm run build` clean

Worth naming: **my first version of the ordering test passed without the fix.** Both orderings produce the same final sequence — only the invoke timing differs — so asserting on the final order proved nothing. It now drains microtasks with the disconnect still parked and asserts the second connect has not been issued, which does go red when the chaining is removed. The keychain test pins why a read cannot answer the reachability question at all: a locked store and an empty one are indistinguishable until you write.

## Documentation

No spec change — these are corrections to behaviour already described in `docs/spec/runtime/desktop.md` and the `keychain` module docs, not new behaviour.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,7 +255,10 @@ jobs:
         with:
           components: clippy
 
-      - uses: Swatinem/rust-cache@v2
+      # Pinned to a commit, not the `v2` tag: a tag can be repointed by whoever
+      # owns the action, and whatever it then resolves to runs inside this
+      # workflow. This is v2.9.2. Bump it deliberately.
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6
 
       # Steps are ordered to fail fast and cheap: a compile break (the `E0062`
       # class of bug that motivated this job) surfaces before the lint pass
@@ -788,7 +791,10 @@ jobs:
       - run: npm run build
         working-directory: frontend
 
-      - uses: Swatinem/rust-cache@v2
+      # Pinned to a commit, not the `v2` tag: a tag can be repointed by whoever
+      # owns the action, and whatever it then resolves to runs inside this
+      # workflow. This is v2.9.2. Bump it deliberately.
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6
         with:
           workspaces: src-tauri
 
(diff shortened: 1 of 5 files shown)


oxoxDev:
Auto-closed as merged: the three commits were pushed straight onto `acp-split` (`fa664056..6871b731`), which is this PR's base, so GitHub recognised them as landed.

Recording why, since the outcome is right but the route was wrong: `acp-split` is #588's own head branch. Opening a PR into it to fix #588's review comments put an extra approve-and-merge in front of commits that were going to end up in exactly the same place. Review fixes belong on the PR's head branch; subsequent rounds on #588 go there directly.

Issue #460, opened 2026-08-06
workflows: a tool_call node never consults ApprovalPolicy, so its shell and http calls raise no operator approval


oxoxDev:
## What happens

A workflow `tool_call` node checks the grant namespace and then executes. `ApprovalPolicy` — the thing that parks a call for an operator and records the decision — is never consulted.

`src/workflows/caps/tools.rs`:

```rust
match toolbelt::namespace_of(slug) {
    Some(namespace) if crate::harness::build::grants_cover(&self.grants, namespace) => {}
    ...
}
let tool = self.tools.get(slug)...;
let result = tool.execute(args).await...
```

`WorkflowToolInvoker` builds the same `shell` / `code` / `web` toolbelt an agent gets. So a workflow node calling `shell` or `http_request` produces **no approval card**, on a company in the default `supervised` mode.

## Correction: `[policy].mode` is NOT inert on this path

An earlier revision of this issue claimed the mode has no effect here. **That is wrong, and the distinction matters** — anyone reading the first version would over-rate the exposure.

`caps/mod.rs` parses `[policy].mode` and feeds it to `toolbelt::exec_security`, which is a real, independent enforcement layer:

```rust
SecurityPolicy {
    autonomy: autonomy_for(mode),        // Readonly→ReadOnly, Supervised→Supervised, Full→Full
    workspace_only: true,
    block_high_risk_commands: true,
    require_approval_for_medium_risk: mode == PolicyMode::Supervised,
    allow_tool_install: false,
    auto_approve_all: false,
    ...
}
```

So a `readonly` company's workflow node runs under `AutonomyLevel::ReadOnly`, high-risk commands are blocked, and execution is confined to the workflow workspace. The mode is honoured — by a *different* mechanism than the one the console surfaces.

## What is actually missing

**The operator-facing half.** Two separate things enforce here and only one is wired:

| Layer | On the agent path | On the workflow `tool_call` path |
|---|---|---|
| Exec-security (`SecurityPolicy`, autonomy tier, high-risk blocking, workspace confinement) | yes | **yes** |
| `ApprovalPolicy` (parks for an operator, produces a card, records the decision, honours grants) | yes | **no** |

The consequence is not "unattended arbitrary shell". It is that a class of call an operator would be asked about on the agent path happens silently on the workflow path — no card, no journal entry of a decision, no standing-grant accounting. An operator auditing what was approved sees a hole rather than a risk they accepted.

[...]

PR #594, opened 2026-08-11
feat(workflows): read_run_output companion tool for full node output (#418)

### LINKED ISSUE #418: run_workflow returns 120 characters of the last item of each node, and there is no way to ask for the rest
Found while auditing #410. This is the closest structural twin of that bug still in the tree.

## What happens

`run_workflow` summarises a run for the orchestrator. Per node it renders `- **{name}** ({kind}): {count} item(s) — {preview}`, where the preview is `items.last()` clipped to `ITEM_PREVIEW_CHARS` (120) and terminated with a bare `…`.

So for any workflow whose terminal node produces prose — a generated report, an LLM node's answer, an HTTP node's response body — the orchestrator receives 120 characters of it. Three separate silences stack up:

1. Only the **last** item is previewed. Items 1..n-1 are never shown and never mentioned; the count is honest but the omission is not stated.
2. The clip is a bare `…` with no dropped-character count.
3. There is **no way to get the rest**. `run_workflow` takes a workflow id and nothing else, and its run summary is the only surface for the run's output — so unlike the Composio case in #410 the agent cannot narrow, page, or re-ask. It is not a retry loop; it is a dead end.

The count-of-items line means the agent can tell output exists, which is exactly the shape #410 called out: enough to know something is there, not enough to use it, and no stated route to more.

## Suggested direction

Give the run's output a durable home the agent can read — persist it and hand back a reference (`publish.rs` already does this for oversized artifacts), or add a `read_run_output(run_id, node)` companion tool. Either way the preview should say how much it dropped and name the follow-up.

Scoped out of #410 deliberately: needs a new read surface, not just a better marker.

oxoxDev:
## Summary

`run_workflow` handed the orchestrator agent only 120 chars of the **last** item per node, with a bare `…` and no way to get the rest — a dead end for any workflow whose terminal node produces prose. This adds a `read_run_output(run_id, node, offset?)` companion tool over a bounded in-process cache, and makes the preview honest about what it dropped.

- Preview now reads `… (+N chars)` with the exact dropped **char** count (codepoint-safe), and `last of N items` when a node emitted more than one.
- The run summary footer names the follow-up and carries the `run_id`.
- `read_run_output` renders every item (`Item i of n`), pages under the tool-result byte budget on char boundaries, and has actionable error arms (unknown run → console fallback; unknown node → lists valid ids + counts).

Everything lives inside the already-`openhuman`-gated `src/harness/`, so the default build compiles none of it. No journal/persistence-invariant change (the journal deliberately carries no node output); the cache is process-local and agent-facing only.

## API Or Behavior Changes

- New agent-facing tool `read_run_output` (`ReadOnly`); registered in `INTRINSIC_TOOLS` and the `policy::consequence` declared table (`Reach::Nothing`).
- `run_workflow`'s JSON payload gains `run_id`; its summary wording changes as above.
- No REST route or human-facing UI change — `RunResultPanel.tsx` already renders full node output for the operator.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` (default + `--features openhuman,tinycortex`)
- [x] `cargo build --all-targets` (`cargo check --all-features`)
- [x] `cargo test` (default + gated **2740 passed / 0 failed**; T1–T7: preview marker incl. multibyte, `last of N` + footer, cache populate on success only, agent round-trip returns every dropped char, paging concatenates with no split codepoint, error arms, eviction + oversized-run announce; lockstep + tool-declared gates)

## Documentation

`run_workflow` tool description updated to point at the companion tool.

Closes #418


M3gA-Mind (commented):
## Review — `read_run_output` companion tool (#418)

**Mode:** manual review (read the full diff plus the surrounding source at `pr/594`); no local build or test run.

### Walkthrough

The change closes a real dead end. `summarize_run` previewed only each node's *last* item, clipped to 120 chars with a bare `…`, so any workflow whose terminal node produces prose was unreadable from inside the turn. This adds a bounded in-process `RunOutputCache` that `run_workflow` fills on success, a `read_run_output` tool that pages a node's full output back out of it, and makes the preview honest (`… (+N chars)`, `last of N items`).

The shape is right: a cache, not the journal — the reasoning in the `RunOutputCache` doc comment (journal deliberately scrubs node output; the consumer is the same process; the durable record is the console run drawer) is correct and well argued. Registration is complete on both sides (`INTRINSIC_TOOLS` and `policy::consequence::DECLARED`), and `intrinsic_tools_covers_every_orchestrator_tool` was extended so it can't drift.

| Area | Verdict |
|---|---|
| `RunOutputCache` bounds + eviction | correct |
| `preview_item` codepoint safety | correct, well tested |
| `page_run_output` codepoint safety / forward progress | correct, well tested |
| Error arms | genuinely actionable |
| Node lookup in `read_run_output` | **🟠 mismatched with what the summary prints** |
| Paging cost at scale | 🟡 quadratic |

---

### 🟠 Major — `read_run_output` cannot resolve the node label the run summary prints

`src/harness/orchestrator.rs:2395`

The summary and the reader disagree about what a "node" is called.

- `summarize_run` labels each line with the node's **display name**: `let name = node.name.trim();` (`orchestrator.rs:2145`) → `- **{name}** ({kind}): …`. It keys the state map by `node.id` but never prints that id.
- `ReadRunOutputTool::execute` looks the node up in the cached map, whose keys are **node ids**, with only an `eq_ignore_ascii_case` fallback — also against ids.

So a `node` argument only resolves when the display name and the id differ by case alone. Both the tool description (`:2320`) and the `node` parameter description (`:2333`) tell the agent to pass the name "as named in the run summary", which makes the wrong call the instructed one.

This is provable against the PR's own fixture. `DEMO_WF` (`orchestrator.rs:3919`) declares:

```toml
[[node]]
id = "done"
kind = "output"
name = "Report"
```

The summary renders `- **Report** (output): 1 item(s) — …`, and `read_run_output { run_id, node: "Report" }` falls straight through to the unknown-node arm. The round-trip test T4 passes only because it uses the other node in that same fixture — `id = "worker"`, `name = "Worker"` — the one pair that differs by case alone, which is exactly the degenerate case the fallback was written for. T6 sidesteps it by passing the ids (`"worker"`, `"done"`) rather than the names the summary would have shown.

In real graphs ids are slugs and names are prose, so the first call fails essentially every time. It is recoverable — the unknown-node arm lists the valid ids, which is good design — but it costs a wasted round trip on the happy path, which is the same friction this PR exists to remove.

Smallest fix, which also makes the footer's `node: <name>` instruction true, is to surface the id in the summary line:

```diff
-                                    "- **{name}** ({kind}): last of {count} items — {preview}\n"
+                                    "- **{name}** (`{id}`, {kind}): last of {count} items — {preview}\n"
```

applied to all three arms (with `let id = node.id.as_str();` alongside `name`), and the footer reworded to `node: <id>`.

Alternatively, carry a name→id map into `CachedRunOutput` at `store` time and resolve display names in the reader — more code, but it keeps the summary as-is.

Either way, please add a T6 case that asserts `node: "Report"` resolves, so the fixture's non-degenerate pair is actually exercised. As written, reverting the `eq_ignore_ascii_case` fallback would still leave T4/T6 green if they were switched to ids — the name path has no coverage at all.

---

### 🟡 Minor — paging is quadratic in the number of pages

`src/harness/orchestrator.rs:2271`

Every `read_run_output` call rebuilds the entire joined string via `render_run_items`, and `page_run_output` then walks `full.chars().enumerate()` from index 0, `continue`-ing past everything below `offset`:

```rust
for (i, c) in full.chars().enumerate() {
    if i < offset { continue; }
```

Reading one cached run to the end at ~15 KiB per page re-renders and re-scans the whole output once per page. For a run near `RUN_OUTPUT_ENTRY_MAX_BYTES` (16 MiB) that is ~1000 pages, each re-materialising 16 MiB. Fine at demo sizes, but the cache's own ceiling admits inputs where it is not.

Resolving the offset to a byte index once and slicing is a small change:

```rust
let start_byte = full.char_indices().nth(offset).map(|(b, _)| b).unwrap_or(full.len());
let mut page = String::new();
for (rel, c) in full[start_byte..].char_indices() {
    if !page.is_empty() && page.len() + c.len_utf8() > budget {
        return (page, Some(offset + full[start_byte..start_byte + rel].chars().count()));
    }
    page.push(c);
}
```

`render_run_items` still rebuilds per call; caching the rendered string per `(run_id, node)` alongside the entry would remove that half too. Happy to leave this as a follow-up if you'd rather keep the PR tight — flagging so it's a decision rather than an oversight.

---

### 🟡 Minor — an unserializable node map is accounted as 0 bytes and fails open

`src/harness/orchestrator.rs:1759`

```rust
let bytes = serde_json::to_string(&nodes).map(|s| s.len()).unwrap_or(0);
```

On the error arm the entry is still cached, at a recorded size of 0, so it never counts toward `RUN_OUTPUT_CACHE_MAX_BYTES` and never contributes to byte-ceiling eviction. Serializing a `serde_json::Value` is effectively infallible, so this is theoretical rather than reachable — but the safe default is the other one:

```diff
-        let bytes = serde_json::to_string(&nodes).map(|s| s.len()).unwrap_or(0);
+        // An unserializable map can't be sized, so treat it as over the hard
+        // ceiling: refuse it rather than cache it at a false zero.
+        let bytes = serde_json::to_string(&nodes)
+            .map(|s| s.len())
+            .unwrap_or(usize::MAX);
```

which routes it through the existing `Oversized` arm and its honest footer.

---

### Questions for the author

**1. Cache contents vs. the workflow approval gap (`#460`).** `store` is called for every non-cancelled run, including runs that returned with `pending_approvals` non-empty, and it caches the engine's raw `run.output["nodes"]` wholesale.

`read_run_output` itself is classified correctly — `PermissionLevel::ReadOnly`, `Reach::Nothing` — it genuinely reaches nothing but this process's own memory, and I'm not arguing otherwise.

The note is about reach through it rather than of it: today the orchestrator sees 120 clipped chars of a node's last item; after this PR it sees all of every item, unbounded. Issue #460 (workflow `tool_call` nodes never consult `ApprovalPolicy`) is in flight right now, and that ungated tool output is precisely what lands in this cache. This PR does not create that gap, but it does widen what the agent can pull through it.

Two things worth your call: whether `store` should skip nodes still sitting in `pending_approvals`, and whether there is a preferred landing order relative to #460. Raising it so it's a decision, not a surprise.

**2.** `ReadRunOutputTool` holds `company` but uses it only in `tracing::debug!`; lookup is by `run_id` alone. I traced `HarnessDeps` construction to the per-tenant path (`src/runtime/builder.rs:1473`) and satisfied myself the cache is per-company in practice, so this is not a leak — but is the field intended as future scoping, or is it decorative? An assert or a comment would stop a later refactor from making a shared cache look safe.

---

### Verified / looks good

- **`preview_item` is codepoint-safe and honest.** `chars()`-iterated throughout, and `(+N chars)` counts codepoints. T1 covers both sides of the cap with ASCII and multibyte fills, and asserts the kept body is exactly `ITEM_PREVIEW_CHARS`, not one over.
- **`page_run_output` never splits a codepoint and always advances.** The `!page.is_empty()` guard means the first char is pushed unconditionally, so a caller looping on the returned offset makes progress even against a pathological budget — a real trap, correctly avoided and explicitly documented. T5 proves the pages reassemble to the original byte-for-byte.
- **Eviction order is right.** Run-count bound first, then the byte ceiling, with `q.len() > 1` so the run just stored is never the one evicted. The comment explaining why briefly overshooting the total beats dropping the entry a follow-up is about to read is the correct trade and worth having written down.
- **Oversized runs are announced, not silently dropped.** `RunOutputStored::Oversized` threading through to the footer means an agent facing a too-large run is sent to the console drawer rather than to a `read_run_output` call that would find nothing. This is the detail most implementations get wrong.
- **Failure arms are actionable.** Unknown run names the cache scope and the console fallback; unknown node lists the valid ids *with item counts*; an empty node returns a success that says so rather than an empty result. No panics, no bare empties.
- **Registration can't drift.** Both `INTRINSIC_TOOLS` (`steps.rs:490`) and `consequence::DECLARED` (`consequence.rs:176`) updated, `Reach::Nothing` is the right classification, and the lockstep test was extended to cover the new tool.
- **Cancelled and failed runs cache nothing** (T3 asserts `len() == 0` on the cancelled path).
- **Blast radius is contained** — everything lives under the already-`openhuman`-gated `src/harness/`, and the remaining 12 files are one-line `run_outputs:` additions to test dep structs.

### Not covered

I did not compile or run the test suite, so the `#[allow(clippy::too_many_arguments)]` placement and the exact `TOOL_RESULT_BUDGET_BYTES` arithmetic are read, not executed. CI is green on this PR, which covers both.

oxoxDev:
Thanks Elvin — sharp read on the node-lookup mismatch especially; the summary/reader disagreement was exactly the wasted-round-trip this PR is meant to remove. All three findings addressed, plus answers to both questions. Rebased on latest `main` first.

- 🟠 **node lookup: fixed in c3a29da5** — implemented both halves. Every summary line now prints the id next to the name (`- **Report** (`done`, output): …`) and the footer instructs `node: <id>`; and the cache carries a display-name → id map captured at `store` time, so the reader resolves a `node` arg that is *either* the id or the display name (the exact + case-insensitive id paths stay). Tool/param descriptions updated to match. Added **T4b** (`read_run_output_resolves_display_name_and_id`) asserting `node: "Report"` (id `done`, the DEMO_WF non-degenerate pair) resolves, with an id-based case alongside — the name path now has real coverage.
- 🟡 **paging: fixed in 77840211** — resolve `offset` to a byte index once via `char_indices().nth(offset)` and slice from there, so reading a run to its end is linear rather than quadratic. Codepoint safety and the `!page.is_empty()` forward-progress guard are unchanged; T5 still proves pages reassemble byte-for-byte.
- 🟡 **byte accounting: fixed in 50604c6c** — `unwrap_or(0)` → `unwrap_or(usize::MAX)` with the explaining comment, so an unsizable map routes through the existing `Oversized` arm instead of being cached at a false zero that never evicts.

**Q1 (cache vs #460 approval gap).** Left `store` as-is — it does not special-case `pending_approvals`, deliberately. What lands in the cache is `run.output["nodes"]`, i.e. the output of nodes that actually *executed*; a node still sitting in `pending_approvals` hasn't run, so it has no items to leak — the cached entries are its completed upstream nodes, and handing those back in full is precisely the tool's job (it's `ReadOnly` / `Reach::Nothing` — this process's own memory, nothing else). The reach you're flagging is real but it lives in #460 itself: the `tool_call` approval gate is what governs whether that ungated output should exist at all. So no change to this cache is owed; I'll coordinate landing order with #460 rather than pre-empt it here. If #460 lands a notion of "approved-but-unexecuted partial output" that a node could carry, that's the point to revisit skipping it — flagging so it's tracked, not dropped.

**Q2 (`company` field).** Intentional — it's scoping context, not a leak. Lookup is by `run_id` alone and that's safe *because* the cache is per-company by construction: `HarnessDeps` (and the `RunOutputCache` handle) is built once per tenant in `build_agent` (`src/runtime/builder.rs`), so a foreign run id can never be in this cache. Added a doc-comment on the field spelling exactly that out (**b4310030**), so a later refactor toward a shared cache has to re-derive scoping from the id rather than assume this one is already isolated.

Gates green locally: `fmt --check`, `clippy --all-targets -D warnings` (default + `--features openhuman,tinycortex --no-deps`), `test` (default + gated), `check --all-features`.

Issue #580, opened 2026-08-10
spine: nothing turns a task's plan into a workflow, so Done can only ever link to work done once


graycyrus:
## What exists, so this is not confused with it

| Path | What it does |
|---|---|
| `src/harness/planning.rs` | A card entering Planning gets one model call that writes a `TaskPlan` — a plan, not a workflow |
| `src/runtime/workflow_spawn.rs` | Starts a run of an **existing** workflow through the supervisor (#395, #440) |
| `frontend/src/api/workflow-proposal.ts` | The copilot proposes edits to an **already-open** workflow, which the operator applies (#415) |
| #339 | A card in Done links to what it produced — including *"a workflow it built or ran"* |

So the console can link to a generated workflow, and the copilot can edit one an operator is already looking at. **Nothing turns a task's plan into a new workflow.** #339 built the destination; the road to it does not exist.

## What this is

The task lifecycle in the v1 feature spec ends with the workflow builder picking up the plan and producing a workflow as the deliverable:

```
prompt → Planning (plan written) → In Progress (workflow built) → In Review → Done + link to the workflow
```

That is a different product claim from "an agent did the work once". It means asking for *"a daily digest of dev activity emailed to me"* yields a repeatable thing you can open, edit, schedule and re-run — not a single completed errand.

## What it needs

1. A path from a settled `TaskPlan` to a candidate workflow graph. The plan already carries decomposed steps and verified prerequisites, which is most of the input a graph needs.
2. Reuse of the proposal machinery from #415 — an operation list the operator reviews, not a replacement graph — rather than a second way to author a workflow.
3. A decision on the gate: does the generated workflow land in In Review for approval before it is runnable, or is it published disabled? #276 shipped enable/disable, so "published but disarmed" is available.
4. Reuse of #339's Done link. Nothing new is needed at the far end.

## Open questions before building

- Which tasks should produce a workflow at all? A one-off "summarise this document" should not.
- What happens when the plan is not automatable — does the card just finish as work done?
- Does the operator get to choose "do it once" versus "build me the workflow" at the prompt box?

Part of the spine — epic #183.

oxoxDev:
Both stages shipped and merged: Stage 1 backend bridge (#595) and Stage 2 operator UI (#642). A task can be built into a workflow proposal, reviewed in the In Review op-diff panel, and applied to a real workflow with the Done card linking it (#339). Closing.

PR #595, opened 2026-08-11
feat(workflows): plan→workflow bridge, Stage 1 backend (#580)

oxoxDev:
## Summary

Stage 1 of the spine (#580, epic #183): turn a task's settled plan into a **repeatable workflow**, so a prompt like "email me a daily digest of dev activity" yields a thing you can open, edit, schedule and re-run — not a one-off errand.

When a card's deliverable is "workflow", entering In Progress runs a builder pass (mirrors the planning station; mints a RunRecord since building the workflow *is* the work) that turns the plan into a candidate graph and settles the card into **In Review**. The operator applies or rejects; apply creates the workflow through the single authoring path and links it on the Done card.

Locked product decisions: (a) the operator **explicitly** picks once-vs-workflow — no auto-detection; (b) the graph lands in **In Review** for approval before it exists (no ghost drafts) and #276's disarm keeps any scheduled graph unarmed until enabled; (c) a non-automatable plan drops back to To-do with the model's reason.

Scope is the **backend bridge + minimal console API**. The operator UI (composer toggle, In Review op-diff panel) is Stage 2.

**Note:** the plan referenced a "#415 op contract" that does not exist (#415 is the copilot propose-a-diff feature). The real authoring contract is `{id,name,description,nodes,edges} → RawWorkflow`; the proposal stores that graph shape and a new host-authority helper (`raw_workflow_from_spec`) rebuilds it. Single authoring path, apply-time re-validation, and host authority are all honored.

## API Or Behavior Changes

- `TaskRecord` gains `deliverable` (default `Once`, skip-if-default → pre-existing cards byte-identical) and an optional `TaskWorkflowProposal`.
- New routes: `POST /tasks/{id}/workflow-proposal/apply` and `/reject`. `deliverable` accepted on task create/patch and the chat intent card.
- Builder spend metered as `SampleKind::Inference`, attributed to the card assignee + run_id.
- Without the `openhuman` feature the builder edge is inert (planner-less pattern) — default behavior unchanged.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` (default + `--features openhuman,tinycortex`)
- [x] `cargo build --all-targets` (`cargo check --all-features`)
- [x] `cargo test` (default **1851**; gated **2781**; 2 port-additivity, 5 metering, 7 bridge-route incl. apply creates + #276-disarm + roster-drift keeps In Review + reject→To-do + no-proposal→400, 12 gated builder-pass)
- [x] frontend `npm run typecheck` / `build`
- [ ] **Manual end-to-end (curl flow in PR discussion):** create workflow-deliverable card → In Progress → In Review proposal → apply → Done links the workflow (disabled) → arm.

## Documentation

`docs/spec/runtime/workflow-build.md` (new) + tasks/planning pages document the bridge and the review-before-creation gate.

Advances #580 (Stage 1 of 2). Stage 2 (operator UI) is a follow-up PR.


M3gA-Mind (commented):
## Review — plan→workflow bridge, Stage 1 backend (#580)

**Mode:** manual review (read the full diff plus surrounding source at `pr/595`); no local build or test run.

**Coverage.** I read `src/harness/workflow_build.rs` and its `test.rs` in full, `src/server/ops/tasks.rs` (apply/reject), `src/company/workflow_create.rs` (`raw_workflow_from_spec`, `courtesy_validate_draft`, `validate_draft_against_record`, `validate_tool_call_node`), `src/ports/tasks.rs`, `src/metering/workflow_build.rs`, `src/company/runtime.rs` and the runtime wiring, and traced the node-kind vocabulary out into `src/company/workflow_file.rs` and `src/workflows/translate.rs`. I read `frontend/src/api/tasks.ts` only as a type shim and did **not** review `src/server/ops/write_test.rs` (295 lines) or the docs pages line by line. Flagging that rather than implying a uniform pass over 3015 lines.

### Walkthrough

The architecture is the right one and the doc comments are unusually good — the "modeled on the planning station, with one deliberate difference" header earns its place. The three decisions that matter are all correctly implemented: the operator picks once-vs-workflow explicitly, the graph lands In Review as a *proposal* rather than a disabled ghost draft, and apply is host-authoritative (rebuilds from stored `ops`, never a client body, under the `task_writes` lock, re-validating through the single `create_company_workflow` path with #276's disarm intact).

My findings are all in one place: **the boundary between what the host controls and what the model controls.** The host correctly took ownership of the workflow *id*. It did not take ownership of the node *kind*, the `requires_approval` flag, or the workflow *name* — and those are the three that carry consequence.

---

### 🟠 Major — the builder offers the model the whole engine vocabulary but specifies only three kinds

`src/harness/workflow_build.rs:819`

```rust
let node_kinds = WORKFLOW_NODE_KINDS.join(", ");
```

That hands the model all twelve kinds — `trigger, agent, tool_call, http_request, condition, output, switch, merge, split_out, transform, output_parser, sub_workflow`. The prompt then gives shaping rules for exactly three: `trigger` (exactly one, optional cron), `agent` (must name a roster id), `output` (`destination.kind` from a fixed set). For the other nine the model is told the name and nothing else.

What actually happens per kind is uneven:

- **`tool_call` — well defended.** `validate_tool_call_node` (`workflow_create.rs:665`) requires a non-empty, un-padded `config.slug` naming a wired toolbelt namespace the company's `[tools].allow` grants. A model inventing a `tool_call` node almost certainly fails courtesy validation and the card lands back in To-do with a reason. Graceful — it costs a paid pass, not an unsafe graph. Credit where due; that check is doing real work here.
- **`http_request` and the seven structural kinds — no arm at all.** `validate_draft_against_record` matches `"agent"` and `"tool_call"` and falls through `_ => {}` (`workflow_create.rs:643`). An `http_request` node's `config` — URL, method, body — rides through `raw_workflow_from_spec` verbatim into the created workflow, and `translate.rs:102` maps it straight to the engine's `NodeKind::HttpRequest`.

The input side is what makes this worth raising rather than filing as tidiness. The card's title, note and plan are user-written text — your own system prompt says so and correctly instructs the model to treat them as data. So there is a path from card text to a model-synthesized `http_request` node with an attacker-influenced URL, landing in a graph that runs on a cron once armed.

Two genuine mitigations exist and I want to be fair about them: #276 lands any schedule-carrying graph **disarmed**, and apply requires a person. But Stage 1 ships the apply route with no UI — that's explicitly Stage 2 — so the person approving today issues a bare `POST` with nothing rendering the graph's URLs. The human gate is real but currently blind.

The narrow fix is to give the builder its own vocabulary instead of borrowing the engine's:

```rust
/// The node kinds the builder prompt actually specifies. Deliberately narrower
/// than `WORKFLOW_NODE_KINDS`: the model is only told how to shape these, so
/// offering it the rest invites a node nothing validates. Widening this is a
/// deliberate act that comes with a matching prompt section.
const BUILDER_NODE_KINDS: &[&str] = &["trigger", "agent", "condition", "output"];
```

use it in `system_prompt()`, and reject an out-of-vocabulary kind in `run_workflow_build_pass` before `courtesy_validate_draft` (settling to To-do with a reason, like the other refusals). That makes host authority consistent: the host already owns the id, and the kind vocabulary belongs on the same side of the line.

Worth noting the test suite can't currently catch any of this — every test in `workflow_build/test.rs` uses the same two-node trigger→agent fixture. There is no case exercising `tool_call`, `http_request`, or any structural kind. If you keep the wide vocabulary, please add a case per admitted kind; if you narrow it, a single "an out-of-vocabulary kind settles to To-do" test covers it.

---

### 🟠 Major — `requires_approval` is model-controlled, never mentioned in the prompt, and defaults to unpaused

`src/company/workflow_create.rs:460`

`WorkflowGraphSpec`'s node carries `requires_approval: Option<bool>`, and `raw_workflow_from_spec` passes it through verbatim:

```rust
requires_approval: n.requires_approval,
```

The system prompt never mentions the field, so a model will omit it — and per `translate.rs:172` the key is only written when `Some`, so an absent value means the engine does not pause (`runner.rs:489`). Every node in a model-built graph is therefore unpaused by default, and nothing prevents a model from emitting `"requires_approval": false` explicitly on a node a human author would have gated.

This is the part that touches live work rather than theory. #460 — workflow `tool_call` nodes never consult `ApprovalPolicy` — is being fixed right now, and #560 is adding a fourth autonomy tier. This PR doesn't create that gap, but it introduces the first path where those nodes are **model-authored from user-supplied text** rather than hand-written TOML, which changes who has to be wrong for it to bite.

Suggested fix, and it's small: have the host strip the field before assembling the draft, so a builder-authored graph inherits whatever default #460/#560 settle on rather than whatever the model happened to emit.

```rust
// The model does not get a vote on approval gating — the host drops the
// field so a builder-authored node inherits the platform default (#460).
for node in &mut spec.nodes {
    node.requires_approval = None;
}
```

right alongside the existing `spec.id = safe_workflow_id(…)` line in the `BuildOutcome::Graph` arm, which is already where the host asserts authority.

If you'd rather the builder be able to gate nodes *up* (set `true`, never `false`), that works too — but it should be a stated rule in the prompt with a test, not an unmentioned passthrough.

---

### 🟡 Minor — the host dedups the id but leaves the name to a polite request

`src/harness/workflow_build.rs:608`

`gather_evidence` collects `existing_names` and `existing_ids`. The ids get real treatment — `safe_workflow_id` slugs, bounds to 56 chars for suffix room, dedups to `-999`, then falls back to a timestamp. The names get a line in the prompt asking the model not to clash.

Name uniqueness is enforced at `create_company_workflow`, so a clash is caught — but it's caught at *apply*, which is the documented "a name has since been taken" path that keeps the card In Review and hands the operator a refusal on an otherwise-good graph. Since the host is already holding `existing_names` and already has the de-dup logic, running the model's name through the same suffix treatment would turn a manual round trip into a non-event. Genuinely minor, since the failure is loud and recoverable.

---

### 🟡 Minor — the plan is the one unbounded input on a paid path

`src/harness/workflow_build.rs:615`

`card_title` is capped at `MAX_SUMMARY_CHARS` (400) and `card_note` at `MAX_REASON_CHARS` (1200), but the plan is cloned whole:

```rust
plan: card.plan.clone(),
```

and `evidence_prompt` (`:878`) renders every step's `title` + `detail`, every prerequisite, and the full `verification` string with no bound on count or length. `MAX_OUTPUT_TOKENS` bounds the answer; nothing bounds the question, and input tokens are exactly what `record_workflow_build_usage` meters. Plans are host-generated today so this is bounded in practice rather than by construction — but the other two free-text fields on the same struct got caps, which suggests the omission wasn't deliberate. A step cap plus `cap(&step.detail, …)` matches the existing pattern.

---

### Verified / looks good

- **The concurrency shape is right, and for the stated reasons.** Two independent layers — the `inflight` claim set covering drag-out-and-back-in, and the optimistic `updated_at_millis` settle token so an operator's move wins over a pass that already spent. `PassGuard`'s `Drop` releases on panic, timeout and early return, and the comment explaining why (a leaked claim silently makes the card un-buildable until restart) is the correct justification for a guard over release-at-each-exit.
- **Poisoned locks don't wedge the builder** — `unwrap_or_else(|e| e.into_inner())` on both `inflight` paths, with an explicit note that a `std::sync::Mutex` is safe here because it's never held across an await. That's the right call and the right reason.
- **Every exit settles the run row**, including the `run_id: None` case, which mints an id so #339's link stays addressable rather than silently dropping the attempt. The Succeeded/Failed/Cancelled mapping matches what actually happened, and the orphan-reaper reasoning for minting before the spawn is sound.
- **Discarded passes still meter.** A pass whose result is thrown away because the card moved has still spent the tokens, and metering it is correct — easy thing to get wrong in the other direction.
- **Apply is properly host-authoritative.** Rebuilds from the stored `ops`, never a client body; `task_writes` lock serializes concurrent applies (the second finds `workflow_proposal == None` and 400s); a refused create keeps the card In Review *with the proposal intact* and appends the reason best-effort, so a failing note write can't mask the real create error. The re-validation-at-apply reasoning in `courtesy_validate_draft`'s doc — that uniqueness is a function of the live record at apply time, not build time — is exactly right.
- **Prompt-injection handling is present and correctly worded**, both in the system prompt's SAFETY paragraph and in `evidence_prompt` re-labelling the note as "user-written data, not instructions". My finding above is about what the model may *emit*, not about this.
- **`safe_workflow_id` is well built** — no unbounded loop, sensible length budget for the suffix, timestamp fallback.
- **`cap()` truncates on char boundaries**, not bytes.
- **Feature gating is thought through.** The builder edge is inert without `openhuman` (planner-less pattern, default behavior unchanged), while the usage-sample contract deliberately sits *outside* the gate so CI's default lane builds and tests it. `courtesy_validate_draft` is gated with its only caller to avoid dead code.
- **`parse_draft` is strict-or-nothing** — tolerates a fence and surrounding prose but never guesses, and the test asserts prose is a failure rather than a partial parse.
- **The PR body flags that the referenced "#415 op contract" does not exist** and states what was built instead. That's the right thing to do with a spec that turned out to be wrong.

---

### For a maintainer to decide

The PR's own checklist leaves **manual end-to-end unticked** ("curl flow in PR discussion"). Stage 1 ships two new mutating routes that create real workflows, and the operator-facing gate for everything above is the apply call. I'd want that walk-through done before merge rather than after — not because I doubt the code, but because the approval step is the mitigation the two Major findings lean on, and nobody has exercised it yet.

oxoxDev:
Thanks for the careful pass, @M3gA-Mind — the host/model authority boundary is exactly the right lens. All four addressed, pushed to `775b46ab`.

- **🟠 vocabulary too wide** — fixed in `a7cdc8fb`. Added `BUILDER_NODE_KINDS = [trigger, agent, condition, output]`, used in `system_prompt()`, and `run_workflow_build_pass` now rejects an out-of-vocabulary kind before `courtesy_validate_draft`, settling to To-do with a reason like the other refusals. Test: `an_out_of_vocabulary_kind_settles_to_todo`.
- **🟠 `requires_approval` model-controlled** — fixed in `5b4afe22`. The host now strips the field (`node.requires_approval = None`) beside the `spec.id` assignment in the `Graph` arm, so a builder-authored node inherits the platform default (#460/#560) rather than the model's choice. Went with the simple strip, not gate-up-only. Test covers both a model-emitted `true` and `false` being dropped.
- **🟡 name not deduped** — fixed in `72eb108d`. The model's name now runs through the same suffix-dedup as the id (using the `existing_names` the host already holds), so a clash is a non-event instead of an apply-time round trip. Test: `the_host_dedups_a_clashing_name`.
- **🟡 unbounded plan on a paid path** — fixed in `d513e086`. Step count capped and each `step.detail` run through `cap()`, matching the `card_title`/`card_note` pattern.

Gates green: fmt, default + gated clippy `-D warnings`, gated tests (21 workflow_build lib tests incl. the new ones), `cargo check --all-features`.

**Manual E2E** — noted, not skipping it. The two host-authority fixes above shrink what apply has to catch, but the create-card → In Review → apply → Done-links-workflow walkthrough still needs a live host; it's being arranged on staging and I'll confirm it here before merge, not after.

Issue #486, opened 2026-08-06
overview: draw ring 1 from the company's real desks instead of keyword-guessed departments


M3gA-Mind:
## Context

Follow-up to #311, which built the Company org chart over the structure the company actually declares. That makes a long-standing caveat on the Overview graph fixable.

## The problem, in the graph's own words

`frontend/src/views/overview/README.md` heads a section **"The derived rings — read this before trusting the org chart"** and says:

> A company manifest has **no department field and no per-agent tool list** ... Departments come from keyword-matching each teammate's role, falling back to Operations ... Both are deterministic, so nothing jumps between renders — but neither is something the company declared.

So ring 1 of the five-ring graph is invented. `assignDepartment` in `kg/adapter.ts` guesses a department from a role string, and `DERIVED_NOTICE` is the standing apology for it. The graph ignores `[[group_chat]]` entirely — the one place the company *does* declare how it is organised.

## What changed

#311 added `frontend/src/lib/org.ts`: a pure, tested reader that turns `GET {scope}/desks` + `GET {scope}/team` into company → desk → seat, with provenance and the lead read from `members[0]`. Desks are the real departments. The graph can now read structure instead of guessing it.

## What this issue is for

Replace the invented departments with the declared ones:

- Ring 1 becomes the company's desks, not keyword-matched role buckets.
- Delete `assignDepartment` — the README already asks for exactly this ("when `[[agent]]` grows `department` ... delete `assignDepartment` and `assignTools` and read them straight through"). Desks arrive at the same destination by a different route.
- Decide what ring 1 shows for a teammate on **no** desk. The org chart lists them beside the tree rather than placing them, precisely because the company declares no position for them; the graph needs an answer that is not a fabricated ring.
- Update `DERIVED_NOTICE` and the README section so the remaining caveats are only the ones still true (tools, workflow stages, a human's department).

Kept deliberately separate from #311: deleting the fiction belongs with its replacement, not with the foundation, and this touches the graph's layout maths.

## Not in scope

`assignTools` and the derived workflow stages. Neither has a declared source yet, so both keep their caveat until one exists.

## Related

- #311 — the org chart and `lib/org.ts` this builds on

oxoxDev:
**Related — `cluster:company-graph`**

Org structure, desks, and the Overview knowledge graph. Two follow-ups here were split out of #311 (in flight, PR #488).

*The manifest gap and its consequences — read these together:*
[...]

Issue #363, opened 2026-08-04
manifest: give [[agent]] a department and a tool list so the graph stops inferring them


oxoxDev:
## What is missing

`[[agent]]` has no `department` and no `tools`. The knowledge graph needs both, so it invents them — `assignDepartment` keyword-matches the role string, and `assignTools` deals the company-wide `[tools] allow` deterministically across the roster.

This is the fix that makes the inference unnecessary. #361's author already wrote the deletion for it: *"when `[[agent]]` grows `department` and `tools`, `assignDepartment` and `assignTools` come out and the fields are read straight through."*

## Why it is worth doing properly rather than quickly

**A per-agent `tools` field is a permission statement, not a display hint.** There is already a real grant path — `[tools] allow` on the company, matched by `grant_matches`, and the per-agent `tools` list the roster builder reads. A second, differently-shaped tool list that exists only to label graph nodes would be a trap: the graph would show one thing and the harness would enforce another, and the first person to hit that will reasonably assume the graph is authoritative.

So the question to settle before writing the field: **is this a new field, or is the graph simply reading the roster's existing per-agent tools?** If the roster already knows, the graph should ask it rather than the manifest gaining a parallel list.

`department` has no such hazard — it is genuinely new information nothing else holds.

## Also worth deciding

- **Is `department` free text or a declared set?** Free text is easy and drifts into "Eng" / "Engineering" / "engineering" as three departments. A declared `[[department]]` block with an id and an `order` is more ceremony and gives the graph its ring ordering for free — the current code already carries an `order` onto the node because the ported `GRAPH_DEPT_ORDER` hardcoded six ids that were not ours.
- **What happens to a company that declares neither?** Falling back to today's inference is reasonable, but then the derived-versus-real distinction (#362) has to survive per-company rather than being a global caption.

## Acceptance

- A manifest can state a teammate's department, and the graph renders it without inference.
- A teammate's tools on the graph agree with what the harness actually grants — one source, not two.
- `assignDepartment` and `assignTools` are deleted, not left as a parallel path that can disagree.
- A company that declares nothing still renders, and is still marked as inferred.

## Related

- #361 — introduced the inference and named this as the fix
- #362 — the derived data is currently unmarked on screen

oxoxDev:
**Related — `cluster:company-graph`**

Org structure, desks, and the Overview knowledge graph. Two follow-ups here were split out of #311 (in flight, PR #488).

*The manifest gap and its consequences — read these together:*
[...]

PR #602, opened 2026-08-11
overview: draw ring 1 from the company's real desks instead of keyword-guessed departments

M3gA-Mind:
## Summary

Ring 1 of the Overview knowledge graph was invented. `assignDepartment` keyword-matched each teammate's role string into one of five hardcoded buckets (`dept-product`, `dept-engineering`, …) and fell back to Operations when nothing matched, while the graph ignored `[[group_chat]]` entirely — the one place the company actually declares how it is organised.

Ring 1 is now the company's **desks**, read from `GET {scope}/desks`: the same declared source the Company org chart reads. `assignDepartment`, `ROLE_HINTS` and the hardcoded `DEPARTMENTS` list are deleted, as `frontend/src/views/overview/README.md` had been asking for.

Closes the ring-1 half of the caveat. `assignTools` and the workflow templates keep theirs — neither has a declared source yet.

### The route I took, and what it leaves for #363

Both this and **#363** delete `assignDepartment`, so whichever landed first was going to delete the other's target. This takes the **desks** route (structure the company already declares) rather than #363's manifest route (a new `[[agent]] department` field). #363 is not blocked — its scope shrinks to what it always uniquely owned:

- **`assignTools`** — still derived, still a deterministic deal from the company-wide `[tools] allow`. #363's own analysis of why a per-agent tool list is a permission statement rather than a display hint is untouched by this PR.
- **The derived workflow stages** — still templates; there is no flow API.
- **A per-agent `department` field**, if that is still wanted once desks are read. It would now be a *second* declaration of the same fact, which is worth deciding deliberately rather than by default.

⚠️ **#363 is assigned to @senamakel** (since 2026-08-04). There is no PR and no code activity on it, so nothing was in flight to collide with, but flagging it explicitly since this deletes code that issue names.

## API Or Behavior Changes

**No API changes.** `Overview` now calls the existing `client.listDesks()`; no new endpoint, no schema change.

Behaviour changes, all on the Overview graph:

1. **Pillars are desks.** A teammate's department is the desk that seats them, not their job title. A "Designer" on the Engineering desk now reads as Engineering.
2. **Nobody gets a position the company did not declare.** A teammate on no desk, and every human, are **unplaced**: their node is drawn on the worker ring in a sector of its own, hanging off the company core, with no pillar above them. This is the radial form of the org chart's existing `Unplaced` section ("Outside the tree on purpose … putting them under a node would be inventing structure").
3. **`assignHumanDepartment` is deleted.** Not scope creep — leaving it in would have made it *worse*. It spread humans across whatever departments existed; while those were invented buckets that was self-consistent fiction, but spreading them across **real desks** asserts a membership the desk's own member list denies, and the org chart two clicks away would contradict it on a fact the operator can check. The issue lists "a human's department" as a surviving caveat; I think that only held while ring 1 was fake, and I'd rather have this argued down in review than shipped silently either way.
4. **An unplaced teammate's open cards are dropped from the graph.** Ring 2 hangs off ring 1, and there is no honest desk to hang them from. This is a real loss of real data — stated in the code where the decision is made rather than left to fall out of a guard in `model.ts` — and the alternative was asserting a desk membership. The board still shows the card; seating the teammate brings it back.
5. **A company that declares no desks draws no pillars**, and everyone is unplaced. That is the true picture rather than a guessed one. (#363's acceptance criteria allow a declare-nothing company to fall back to inference; that is available to it, but it needs the inference this PR deletes, so it is #363's call to make, not this one's.)
6. **Workflows survive.** `WORKFLOW_TEMPLATES` were pinned to the deleted `dept-*` ids, so once the ids changed every template would have matched no department and the workflow ring would have **silently drawn nothing**. Routines are now dealt to desks by position (wrapping), with desk-scoped ids. The arbitrariness is deliberate and documented: the console does not know what a desk called "Front of house" runs, and the old per-department pinning read as if it did.
7. **A failed `/desks` degrades rather than throwing.** The org chart treats it as a hard error because desks *are* that page; here they are one ring of five, so a failure draws no pillars instead of taking the real rings down.

### Scope question the issue isolated — what ring 1 shows for a teammate on no desk

The issue asks for an answer "that is not a fabricated ring", and notes a radial layout has no natural "beside the tree". What I settled on:

- **Not a ring.** No "Unassigned" pillar is created. `UNPLACED` is a sentinel that never becomes a `team:` node, and it is namespaced (`desk:<id>` for real desks) so a manifest desk literally named `unplaced` cannot collide with it.
- **Visible, not absent.** `lib/org.ts` keeps unplaced teammates in the chart precisely so "the roster still adds up" and an unplaced teammate is "visible rather than absent". Dropping them from the graph would have been the cheaper change and I think the wrong one.
- **Attached to the core, not floating.** `forceRadial` pins nodes to their ring regardless of edges, so an edgeless node does not fly off — but it gets no rest position, drifts into pillar wedges, and reads as a rendering fault. A `member` edge to the core says the true thing (on the roster, on no desk) without naming a desk, and `radialRestLayout` gives them a sector with **nothing at ring 1** — the empty wedge is what "on no desk" looks like.

## Tests

New: `frontend/test/unit/overview-ring1.test.ts` (14 tests). The fixtures are adversarial on purpose — every teammate's role points at a *different* department than the desk that seats them, so a test can only pass by reading desk membership.

**Revert-and-check: 12 of the 14 were confirmed to fail with the fix reverted**, across 8 targeted reverts (each reverted, run, restored):

| Revert | Tests that failed |
|---|---|
| `model.ts` — drop the core attachment | 1 |
| `tree-layout.ts` — ignore the unplaced sector | 2 |
| `adapter.ts` — stop reading desk membership | 4 |
| humans dealt onto a real desk (old `assignHumanDepartment`) | 2 |
| workflows pinned to the old `dept-*` ids | 1 |
| desk ids not namespaced | 5 |
| draw a pillar for a desk nobody is on | 1 |
| re-sort desks by name instead of host order | 1 |

**One of these caught a vacuous test of mine.** The host-order assertion originally passed even when I re-sorted the desks by name, because my two fixture desks were already alphabetical. The fixture now serves them deliberately out of order, and the test fails under that revert. Reporting it rather than quietly fixing it.

The remaining **2 tests are boundary/regression guards that pass before and after by design**, and I am not claiming otherwise: "draws no pillars at all for a company that declares no desks" (empty-input path) and "does not move the pillars when there is nobody unplaced" (asserts this change does *not* perturb existing layouts).

Commands actually run, in `frontend/`:

- `pnpm test` → **2 failed | 23 passed** test files. The 2 failures (`tour-resume`, `connection-registry`, both `window.localStorage.clear is not a function`) are **pre-existing**: I stashed my entire diff and re-ran them on a clean `upstream/main` tree, and got the identical 23 failed tests. Not caused by, and not fixed by, this PR.
- `pnpm vitest run test/unit/overview-ring1.test.ts` → 14 passed
- `pnpm typecheck` → clean
- `pnpm typecheck:unit` → clean
- `pnpm typecheck:e2e` → clean
- `pnpm build` → clean

That is every frontend step CI runs (`ci.yml` runs `typecheck`, `typecheck:e2e`, `typecheck:unit`, `build`). Note `pnpm typecheck` alone covers **neither** test project — `test/unit` and `test/e2e` are standalone `tsc -p` projects — so the two extra typechecks are load-bearing here, not decoration.

**Not verified:** I did not run Playwright e2e (`pnpm e2e`) — it needs a live host. `test/e2e/org-tree.spec.ts` covers the org chart, which this does not touch. I also have not driven the graph in a browser, so the *visual* result of an unplaced sector (spacing, label crowding at a large roster) is argued from the layout maths and unit-tested positions, not seen. Worth a look from someone with the app running.

- [x] `cargo fmt --all -- --check` — **N/A, not run.** No Rust in this diff.
- [x] `cargo clippy --all-targets -- -D warnings` — **N/A, not run.** No Rust in this diff.
- [x] `cargo build --all-targets` — **N/A, not run.** No Rust in this diff.
- [x] `cargo test` — **N/A, not run.** No Rust in this diff. Ticked because CI counts the boxes; the frontend commands above are what was actually verified.

## Documentation

- `frontend/src/views/overview/README.md` — the "derived rings" section now says ring 1 is declared, and a new "Who the graph does not place" subsection states the unplaced rule for both teammates and humans, and why. The standing request to delete `assignDepartment` is replaced by the narrower one #363 still owns (`assignTools`).
- `DERIVED_NOTICE` (the on-screen legend caveat) no longer claims departments are placeholders, and now says anyone on no desk is shown unplaced.
- `Overview.tsx`'s module doc, and `adapter.ts`'s real/derived table at the top of the file, updated to match.
- The deleted `assignHumanDepartment` leaves a comment in its place explaining why it went, so the next reader does not reintroduce it.

Refs #486

## Summary by CodeRabbit

* **New Features**
  * Company desks now define graph pillars and department placement.
  * Workers without a desk appear in separate sectors around the company core.
  * Human profiles clearly show “Not on a desk” when applicable.
  * Workflows are positioned according to occupied desks.

* **Bug Fixes**
  * Prevented unplaced workers from appearing in phantom departments.
  * Tasks belonging to unplaced members are no longer displayed incorrectly.
  * Graph loading remains available when desk information cannot be retrieved.

### DIFF
--- a/frontend/src/views/Overview.tsx
+++ b/frontend/src/views/Overview.tsx
@@ -7,7 +7,7 @@ import { discoverMcpTools, listMcpServers } from "@/api/mcp";
 import { listMemory, type MemoryEntry } from "@/api/memory";
 import { listSkills, type Skill } from "@/api/skills";
 import { listTasks, type Task } from "@/api/tasks";
-import { ApiError, type McpServer, type McpToolInfo } from "@/api/types";
+import { ApiError, type DeskDto, type McpServer, type McpToolInfo } from "@/api/types";
 import { fromDto, starterTeam, type TeamMember } from "@/lib/team";
 import { adapt, buildMemoryGraph } from "./overview/kg/adapter";
 import { buildKnowledgeGraph } from "./overview/kg/model";
@@ -28,6 +28,16 @@ interface Props {
 interface Sources {
   tasks: Task[];
   team: TeamMember[];
+  /**
+   * The company's desks — ring 1 (issue #486).
+   *
+   * Best-effort like every other source here. A host that cannot serve them
+   * draws a graph with no pillars, which is the same picture as a company that
+   * declares no desks. The org chart treats a failed `/desks` as a hard error
+   * because desks *are* that page; here they are one ring of five, and failing
+   * the whole graph over them would take the real rings down with them.
+   */
+  desks: DeskDto[];
   people: Person[];
   skills: Skill[];
   memories: MemoryEntry[];
@@ -49,6 +59,7 @@ interface Sources {
 const EMPTY: Sources = {
   tasks: [],
   team: starterTeam(),
+  desks: [],
   people: [],
   skills: [],
   memories: [],
@@ -76,14 +87,15 @@ function worthAsking(server: McpServer): boolean {
  * The command centre: the company's knowledge graph, and nothing else.
  *
  * The page is the graph — no header, no strip, no top bar (the shell hides its
- * own for this view). The company sits at the core, its departments are the
- * pillars, the jobs hang off each pillar, the teammate who does each job sits
- * above it, and their tools are the outer ring.
+ * own for this view). The company sits at the core, its desks are the pillars,
+ * the jobs hang off each pillar, the teammate who does each job sits above it,
+ * and their tools are the outer ring.
  *
- * Two of those rings are **derived**, not declared: a company manifest carries
- * no department field and no per-agent tool list, so `kg/adapter.ts` invents a
- * plausible structure rather than leaving the graph three rings short. See
- * `DERIVED_NOTICE` there — it is the standing caveat on this whole surface.
+ * The pillars are **declared**: they are the company's own desks (issue #486),
+ * not the keyword-matched guess that used to stand there. What is still derived
+ * is the tool assignments and the workflow templates — there is no per-agent
+ * tool list and no flow API. See `DERIVED_NOTICE` in `kg/adapter.ts`; it is the
+ * standing caveat on what remains.
  */
 export function Overview({ client, company }: Props) {
   const [sources, setSources] = useState<Sources>(EMPTY);
@@ -91,9 +103,11 @@ export function Overview({ client, company }: Props) {
   useEffect(() => {
     let live = true;
     void (async () => {
-      const [tasks, roster, people, skills, memories, mcp] = await Promise.all([
+      const [tasks, roster, desks, people, skills, memories, mcp] = await Promise.all([
         listTasks(client, company).catch(() => [] as Task[]),
         client.listTeam(company).catch(() => null),
+        // Ring 1 (issue #486). Best-effort: see `Sources.desks`.
+        client.listDesks(company).catch(() => [] as DeskDto[]),
         // Only an admin may list people; a member just gets no humans on the
         // graph, which is the right amount of information for them to have.
         listPeople(client, company).catch(() => [] as Person[]),
@@ -138,6 +152,7 @@ export function Overview({ client, company }: Props) {
       setSources({
         tasks,
         team: roster?.length ? roster.map(fromDto) : starterTeam(),
+        desks,
         people,
         skills,
         memories,
@@ -155,6 +170,7 @@ export function Overview({ client, company }: Props) {
     () =>
       adapt({
         members: sources.team,
+        desks: sources.desks,
         tasks: sources.tasks,
         people: sources.people,
         skills: sources.skills,
--- a/frontend/src/views/overview/README.md
+++ b/frontend/src/views/overview/README.md
@@ -32,25 +32,46 @@ dragged, and re-framing (selecting a node, opening the core) resets it.
 
 ## The derived rings — read this before trusting the org chart
 
-A company manifest has **no department field and no per-agent tool list**, and
-`[tools] allow` is company-wide. Two of the five rings above are therefore
-invented by `kg/adapter.ts`:
+**Ring 1 is no longer one of them.** Departments are the company's **desks**
+(issue #486) — a `[[group_chat]]` in the manifest, or an operator-created
+overlay desk. That is the one place the company declares how it is organised,
+and it is the same source the Company org chart reads. `assignDepartment` and
+its keyword table are gone.
 
-- **Departments** come from keyword-matching each teammate's role, falling back
-  to Operations.
-- **Tool assignments** are a deterministic deal from the company-wide tool list.
-- **Workflows and their stages** are templates — one routine per department —
-  because the console has no flow API; the Workflows canvas draws a single
-  hard-coded sample. Stages are dealt round-robin across the department's
-  agents.
-- **A human's department** is derived too: signing in says who someone is, not
-  what they work on. Admins sit with Operations; everyone else is spread
-  deterministically by their address.
+What is still invented by `kg/adapter.ts`:
+
+- **Tool assignments** are a deterministic deal from the company-wide tool list,
+  because `[tools] allow` is company-wide and `[[agent]]` has no `tools` field.
+- **Workflows and their stages** are templates, because the console has no flow
+  API; the Workflows canvas draws a single hard-coded sample. One routine is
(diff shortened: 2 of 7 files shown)


oxoxDev (commented):
Replacing a keyword-guessed ring with the company's own declared desks is straightforwardly right, and the adversarial fixtures — every teammate's role pointing at a *different* department than the desk seating them — are what make the test suite meaningful rather than decorative. The revert-and-check table (12 of 14 confirmed failing across 8 targeted reverts) is the strongest evidence in the PR; more PRs should carry it.

You asked for two calls to be argued in review rather than shipped silently, so:

**`assignHumanDepartment`'s deletion — agree, and for your reason.** Spreading humans across invented buckets was self-consistent fiction; spreading them across *real* desks asserts a membership the desk's own member list denies, and the org chart two clicks away contradicts it on a fact the operator can check. A view that disagrees with itself is worse than a view that shows less. Deleting it is not scope creep — keeping it would have been the regression.

**Dropping an unplaced teammate's cards — agree it beats the alternative, but it shouldn't be silent.** Asserting a desk to hang them from would be the worse trade, no argument. But the graph now quietly shows fewer cards than the board does, and nothing on the surface says so. That is the one place this PR stops being honest with the operator, in a PR whose whole thesis is honesty about declared vs invented. A count — "3 cards not shown (teammate on no desk)" near the legend, or on the unplaced sector itself — turns a silent omission into the same visible-rather-than-absent treatment you gave the teammates themselves.

**Major — a failed `/desks` is indistinguishable from a company that declares none.**
`client.listDesks(company).catch(() => [] as DeskDto[])` collapses both into an empty array, and the graph draws no pillars either way. Degrading instead of taking the whole graph down is right — desks are one ring of five here, not the page. But an operator whose desks failed to load sees the picture that means "you have declared no desks" and may go looking for structure they think they lost. Keep the distinction: a `desksFailed` flag alongside the array, and a small "couldn't load desks" note where the pillars would be. Same cost, and it's the identical argument the rest of the PR makes.

**Question — is the routine-to-desk deal stable?** Routines are dealt to desks by position, wrapping. If `/desks` ever returns in a different order between loads, the same routine lands on a different desk visit to visit, which reads as a bug even though the assignment is admittedly arbitrary. If the endpoint's ordering isn't already guaranteed stable, sorting by desk id before dealing makes the arbitrariness at least consistent.

**On #363:** flagging the collision explicitly was the right thing to do. Worth also dropping a note on #363 itself and pinging @senamakel — it is assigned to them, and this PR deletes code that issue names. Better they hear it from the PR than discover it in a rebase.

**Minor:** `Rust (openhuman, tinycortex)` is still pending here. Frontend-only diff, so it should be a formality — just don't merge on the four green checks alone.

oxoxDev:
**Post-merge note on the toolchain, not on the change — the two test failures this PR reported as pre-existing were not.**

The body says:

> `pnpm test` → **2 failed | 23 passed** test files. The 2 failures (`tour-resume`, `connection-registry`, both `window.localStorage.clear is not a function`) are **pre-existing** … Not caused by, and not fixed by, this PR.

They are caused by **pnpm**, not by main. This repo's frontend is an npm project and CI runs `npm ci` exclusively (`ci.yml` uses it in every console job, with a comment insisting on `ci` over `install` for lockfile fidelity). I ran this branch's head under npm while checking the graph locally:

```
npm ci && npx vitest run
→ Test Files  25 passed (25)
   Tests     266 passed (266)
```

Both `tour-resume` and `connection-registry` pass. The `window.localStorage.clear is not a function` signature comes from pnpm's strict `node_modules` layout resolving a different jsdom/environment shim than the lockfile pins.

**The trap is that `frontend/` tracks two lockfiles** — `package-lock.json` *and* a stale `pnpm-lock.yaml` — and neither `package.json` nor a `packageManager` field says which one wins. `pnpm install` therefore succeeds and looks authoritative. Worth deleting the stale `pnpm-lock.yaml` (or adding `"packageManager"`) so the next person does not hit this; happy to open that as its own change if you would rather not carry it.

Nothing here affects the merged change — CI was green because CI used npm. Raising it only so a genuinely new failure does not get waved through as "pre-existing" later.

Issue #559, opened 2026-08-10
approvals: a Composio read parks, so reading a mailbox costs the operator an approval


oxoxDev:
## Symptom

An agent asked to check a mailbox, list a repo's pull requests, or read a calendar stops and asks the operator for permission. Every page of every list is its own card.

## Where it comes from

`composio_execute_consequence` in `src/policy/consequence.rs` classifies a Composio **read** as:

```
Consequence { group: Other, reach: Reach::Consequence, standing: Standing::Grantable }
```

`Reach::Consequence` is precisely the value `supervised` parks on. The accompanying comment states the intent plainly — *"a read still reaches a third-party account, so `readonly` denies it and `supervised` parks it the first time"* — and offers the standing grant as the remedy. In practice the first-time park is the whole cost: it interrupts a person, refuses the call, and dead-ends the turn (see the parent issue, point 3).

The read/write classification itself is already correct and already in place: the provider's vendored catalogue tags ~660 actions Read/Write/Admin, an unrecognised slug is treated as a send, and a missing or non-string `tool` argument is treated as a send. None of that needs to change.

## The change

A Composio read should not park under `supervised`. It should still be denied under `readonly`, because that tier's contract is that nothing is reached and nothing is spent.

`Reach` has three values and none of them expresses "reaches outward, changes nothing, costs nothing":

- `Nothing` — runs in every mode including `readonly`. Too permissive: a `readonly` desk should not be pulling a third party's data.
- `Money` — runs under `supervised`, denied under `readonly`. **This is already the exact shape we need**, but the name lies about what is happening.
- `Consequence` — parks under `supervised`. What we have today.

Two options, and the second is likely cleaner:

1. Reuse `Reach::Money` for Composio reads. Correct behaviour immediately, but leaves a value named for billing carrying calls that are not billed.
2. Add `Reach::ExternalRead` with the same two answers as `Money` (denied under `readonly`, allowed under `supervised`) and move Composio reads onto it. Keeps `costs_money()` honest for the spend cap, which reads that predicate.

Keep `Standing::Grantable` either way — it stops mattering for the supervised path but still governs `readonly` and any future tier.

## Acceptance

- Under `supervised`, a Composio action the catalogue tags `Read` runs without parking.
- Under `supervised`, a Composio action tagged `Write` or `Admin` still parks.
- An unrecognised slug, a missing `tool` argument, and a non-string `tool` argument all still park.
- Under `readonly`, a Composio read is still denied.
[...]

PR #603, opened 2026-08-11
tests: build Composio fixtures from the classifier's own action key

### LINKED ISSUE #470: tests: Composio fixtures use a  key the tool does not accept, so they exercise the unclassified path
## What this is

Several test fixtures build a `composio_execute` call as:

```rust
serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL" })
```

The real tool's schema declares `"tool"` — `required: ["tool"]`, `additionalProperties: false` — and `consequence_of` reads `COMPOSIO_ACTION_KEY = "tool"`.

So every one of those fixtures is a `composio_execute` call **with no recognisable action**. The classifier cannot find a slug, falls back to the cautious verdict, and the test passes — for the wrong reason.

Present in at least `src/runtime/journal.rs`, `src/runtime/cycle.rs`, `src/server/operator.rs` and `src/harness/policy.rs`.

## Why it matters

**Production is fine.** The tool and the classifier agree on `tool`; a live operator's approval card shows the slug being read correctly. Nothing is broken for users.

What is broken is the coverage. These tests read as *"a Composio send parks"* / *"a Composio call is journaled"*, and what they actually assert is *"an unrecognisable action parks"*. Both are true today, so nothing fails — which is exactly the problem: **if the read/send classification regressed, these tests would not notice**, because they never reach it.

It also means the cautious fallback is doing the work the classifier is credited with. A fixture that named a real slug would exercise the catalogue lookup; these never do.

## Why it is worth fixing rather than shrugging at

This is the third instance this week of a test that passes while asserting something other than what it claims:

- a coverage test whose fixture pinned `skills_source_dir: None`, blinding it to three tools it claimed to cover
- a card-count case whose seed verb made its premise vacuous under the fix, so it reported success on nothing
- these fixtures

The shape is consistent: the assertion is right, the *input* does not reach the code under test, and nothing tells you. That is worse than a missing test, because it occupies the space where a real one would go.

## What the fix should do

- Correct the fixtures to `"tool"` so they exercise the path they claim.
- Where a test's point is the read/send split, name a slug the catalogue actually classifies, and assert the classification — not just the outcome.
- Keep at least one fixture that deliberately sends an unrecognisable action, clearly named as such, since the cautious fallback deserves its own test.
- Consider making the fixture a helper rather than an inline literal repeated in four files, so the key cannot drift again.

## How to verify

- A fixture naming a real read slug is classified as a read; one naming a real send slug is classified as a send.
- Deliberately breaking the catalogue lookup fails at least one test.
- The unclassified-fallback test still passes and says in its name that is what it covers.

## Related

Found while adding a detached-resolve test for #431 — the existing `gated_tool_call()` fixture would have been refused a standing scope, because it classifies as per-call for this reason rather than on its merits.

M3gA-Mind:
## Summary

Fixes #470.

Composio test fixtures across five modules built their `composio_execute` arguments as
`{"tool_slug": "GMAIL_SEND_EMAIL"}`. Neither the tool nor the classifier reads that key:
`composio_execute`'s schema declares `"tool"` (`required: ["tool"]`,
`additionalProperties: false` — `src/harness/composio.rs:840-853`) and `consequence_of`
reads it back under `COMPOSIO_ACTION_KEY = "tool"` (`src/policy/consequence.rs:151`).

Every one of those fixtures was therefore a call with **no recognisable action**. The
classifier could not find a slug, fell through to the cautious "unknown is a send"
verdict, and the test passed — without ever reaching the catalogue lookup it was
credited with exercising. The assertions were right; the inputs never arrived.

**Why the wrong key looked plausible.** `tool_slug` is not a typo — it is the wire key of
a *different* tool. openhuman's own Composio direct tool declares `tool_slug` /
`action_name`
(`vendor/openhuman/src/openhuman/integrations/composio/tools/direct.rs:713-737`), while
opencompany's `composio_execute` declares `tool`. Two tools, two key names, and fixtures
borrowed from the wrong one. That is why this is fixed with a shared helper keyed off the
classifier's own constant rather than with fifteen careful edits.

### What changed

- **New `src/policy/test_support.rs`** (test-only) — the shared fixture builders. They
  construct the argument object from `COMPOSIO_ACTION_KEY` **itself**, not from a
  literal, so a fixture and the code under test can no longer disagree: renaming the wire
  key moves both together. `COMPOSIO_ACTION_KEY` became `pub(crate)` for this.
- **All 15 occurrences corrected** across `src/harness/policy.rs` (7),
  `src/harness/brain.rs` (2), `src/runtime/cycle.rs` (2), `src/runtime/journal.rs` (2)
  and `src/server/operator.rs` (2). The issue body names four files; `src/harness/brain.rs`
  — the re-dispatch path — is the fifth.
- **The two synthetic loops keep unrecognisable slugs**, as they should:
  `harness/policy.rs`'s over-cap drain test and `server/operator.rs`'s `MultiParkBrain`
  want a *number* of distinct calls and are indifferent to how any of them classify. They
  now say so, via a `composio_unclassified_args_numbered()` helper named for the intent —
  and their slugs land under the real key, so each one reaches the catalogue lookup and
  misses it, rather than carrying no action at all.
- **Action parameters moved under `arguments`**, where the schema puts them. Two fixtures
  spread `"to": "a@b.test"` at the top level, which `additionalProperties: false` rejects.
- **A doc comment that documented the bug as intent.** `src/server/operator.rs:3691` read
  "`gated_tool_call` deliberately is not [grantable]: its Composio payload names no action
  slug, so `consequence_of` cannot classify it as a read". It now names a catalogued
  `Write` action and is a send *on its merits* — which is exactly the "Related" note in
  #470 about `gated_tool_call()` being refused a standing scope for the wrong reason.
- **New coverage for the split** at the layer that projects a blocked call onto the
  operator's card: `a_composio_read_and_a_composio_send_are_classified_differently`
  (`src/harness/policy.rs`), plus per-lane classification pins on each fixture in
  `test_support`'s own test module.

### Deliberately asserting classification, not parking

The new tests assert what a call *is* (`EffectGroup`, `Standing`), not whether it parks.
That keeps them honest across #559, which changes whether a read parks but not what it is
— and #559's negative control is the reason this PR comes first.

### The two lanes

`src/harness/*` compiles only under `--features openhuman`, so 9 of the 15 fixtures do not
exist in the plain lane at all. The vendored catalogue is likewise linked in only under
that feature, so `composio_read_args()` classifies as a **send** in the default build.
That is now pinned explicitly in both directions rather than left to be discovered.

## API Or Behavior Changes

None. Test-only, plus `COMPOSIO_ACTION_KEY` widening from private to `pub(crate)`. No
production code path changes; the tool and the classifier already agreed on `tool`.

## Tests

All exit-code gated (unpiped redirect, `EXIT:$?` read directly).

- [x] `cargo clippy --all-targets -- -D warnings` — EXIT:0 (default lane).
      Gated lane run as CI runs it:
      `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`
      — EXIT:0.
- [x] `cargo build --all-targets` — ran as `cargo check --all-targets` in both lanes
      (default; and `--features openhuman,tinycortex`), plus the two clippy runs above,
      which build all targets. Narrower than a full `build` (no codegen), broader in
      feature coverage.
- [x] `cargo test` — `cargo test --lib` (default lane): **1832 passed, 0 failed**, EXIT:0.
      `cargo test --locked --features openhuman,tinycortex` (full gated lane, integration
      targets included): EXIT:0, all six targets green — **2738 + 10 + 1 + 11 + 2 passed,
      0 failed** (one target reports 0 passed / 1 ignored).

### Negative controls

Two, because they prove different things.

**A — revert the fix itself.** Point `composio_args()` back at `"tool_slug"`, change
nothing else. Gated lane, EXIT:101, **4 failed**:

```
harness::policy::tests::a_composio_read_and_a_composio_send_are_classified_differently
policy::test_support::tests::a_fixture_names_its_action_under_the_key_the_tool_declares
policy::test_support::tests::action_parameters_ride_under_arguments
policy::test_support::tests::the_read_fixture_classifies_as_a_read
```

The interesting part is what **still passed**: every pre-existing test whose fixture this
PR corrected — `require_approval_records_the_request_to_park`,
`a_retried_call_is_recorded_once`, `a_granted_call_is_allowed_once_and_then_parks_again`,
`grants_survive_a_queue_clear` — plus `the_send_fixture_classifies_as_a_send` and
`the_unclassified_fixture_falls_back_to_a_send`. That is issue #470 reproduced on demand:
with the wrong key, the send-shaped assertions are satisfied by the fallback and notice
nothing. Only a test that asserts a *read* can tell the difference.

**B — break the catalogue lookup**, as the issue's "How to verify" asks: in
`composio_action_is_read`, `Some(ToolScope::Read)` → `Some(ToolScope::Admin)`. Gated lane,
EXIT:101, **10 failed**, including both tests this PR adds:

```
harness::policy::tests::a_composio_read_and_a_composio_send_are_classified_differently   <- new
policy::test_support::tests::the_read_fixture_classifies_as_a_read                       <- new
harness::policy::tests::a_composio_read_may_be_granted_standing_and_a_send_may_not
harness::policy::tests::a_grant_from_before_scopes_existed_still_admits_its_tool
harness::policy::tests::a_grant_scoped_to_one_provider_does_not_admit_another_providers_read
harness::policy::tests::a_standing_grant_on_a_composio_read_does_not_admit_a_send
policy::consequence::tests::a_composio_read_is_grantable_and_a_send_is_not
policy::consequence::tests::lookup_ignores_case
runtime::cycle::test::a_composio_read_is_offerable_and_a_composio_send_is_not
runtime::cycle::test::a_standing_grant_on_a_composio_read_records_the_toolkit_it_was_shown_for
```

Both controls restored; the tree is green again (**2738 passed, 0 failed**, EXIT:0).

### One correction to the issue's framing

#470 says a regression in the read/send split "would not fail a single existing test".
Control B shows that is too strong: **8 pre-existing tests do catch it**, in
`policy::consequence`, `runtime::cycle` and `harness::policy` — they were already written
against `"tool"`. What the *15 fixtures* were blind to is exactly what control A isolates:
their key could drift back tomorrow and nothing except the tests added here would notice.
The coverage hole is real and this PR closes it; it was just narrower than the issue body
claims, and worth stating accurately since #559's negative control is being planned
against it.

## Documentation

The rationale lives in `src/policy/test_support.rs`'s module docs — why the module exists,
what stops the key drifting again, how to choose a slug, and why a read-verdict assertion
must be feature-gated. No user-facing docs affected.

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -5620,7 +5620,7 @@ members = ["eng1", "eng2"]
             None,
         )
         .with_requests(requests.clone());
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL" });
+        let args = crate::policy::test_support::composio_send_args();
         let request = ToolPolicyRequest::new(
             "composio_execute",
             args.clone(),
@@ -5856,7 +5856,14 @@ members = ["eng1", "eng2"]
         let log: Arc<dyn crate::ports::EventLog> =
             Arc::new(crate::store::FsEventLog::new(dir.path().to_path_buf()));
         let requests = crate::harness::policy::ApprovalRequestQueue::default();
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL", "to": "a@b.test" });
+        // Issue #470: a real catalogued send, with the action's own parameters
+        // under `arguments` where the tool's schema puts them — so the
+        // re-dispatch path this test covers carries a call the classifier can
+        // actually read.
+        let args = crate::policy::test_support::composio_args_with(
+            crate::policy::test_support::COMPOSIO_SEND_SLUG,
+            serde_json::json!({ "to": "a@b.test" }),
+        );
         requests
             .grants()
             .grant(crate::runtime::grants::GrantedCall {
--- a/src/harness/policy.rs
+++ b/src/harness/policy.rs
@@ -950,6 +950,15 @@ mod tests {
     use super::*;
     use oh::agent::tool_policy::{ToolCallContext, ToolPolicyRequest};
 
+    // Issue #470: the `composio_execute` fixtures are built here, from the same
+    // key the classifier reads, so a call in a test reaches the same catalogue
+    // lookup a call in production does.
+    use crate::policy::test_support::{
+        COMPOSIO_OTHER_SEND_SLUG, COMPOSIO_READ_SLUG, COMPOSIO_SEND_SLUG, composio_args,
+        composio_read_args, composio_send_args, composio_unclassified_args,
+        composio_unclassified_args_numbered,
+    };
+
     fn policy(mode: &str, always: &[&str], auto_under: Option<f64>) -> ApprovalPolicy {
         let p = Policy {
             mode: mode.to_string(),
@@ -1397,7 +1406,7 @@ mod tests {
     #[tokio::test]
     async fn require_approval_records_the_request_to_park() {
         let (p, queue) = queued_policy("supervised", &[]);
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL" });
+        let args = composio_send_args();
         assert!(matches!(
             p.check(&request("composio_execute", args.clone())).await,
             ToolPolicyDecision::RequireApproval { .. }
@@ -1413,6 +1422,65 @@ mod tests {
             queued[0].reason.contains("supervised"),
             "the operator-facing reason rides along: {}",
             queued[0].reason
+        );
+    }
+
+    /// Issue #470, at the layer that projects a blocked call onto the effect
+    /// the operator's card is built from.
+    ///
+    /// The fixtures above used to name their action under a key nothing reads,
+    /// so every Composio test in this module classified through the
+    /// unknown-is-a-send fallback and none of them ever reached the catalogue.
+    /// Read and send came out identical, and a regression in the split would
+    /// have failed nothing here. This asserts they come out different, and
+    /// asserts the classification rather than the parking decision — so it
+    /// stays honest across issue #559, which changes whether a read parks but
+    /// not what it is.
+    #[tokio::test]
+    async fn a_composio_read_and_a_composio_send_are_classified_differently() {
+        let read = composio_read_args();
+        let send = composio_send_args();
+
+        assert_eq!(
+            classify_group("composio_execute", &read),
+            EffectGroup::Other,
+            "`{COMPOSIO_READ_SLUG}` is tagged `Read` in the vendored catalogue; \
+             if this fails the lookup is not being reached"
+        );
+        assert!(
+            grantable("composio_execute", &read),
+            "a read scoped to one connected account is what a standing grant \
+             can honestly describe"
+        );
+
+        assert_eq!(
+            classify_group("composio_execute", &send),
+            EffectGroup::Send,
+            "`{COMPOSIO_SEND_SLUG}` is tagged `Write`"
+        );
+        assert!(!grantable("composio_execute", &send));
+
+        // The cautious fallback still has its own coverage, and still says
+        // send — but now because the catalogue was asked and had no answer,
+        // not because the classifier never saw an action at all.
+        let unknown = composio_unclassified_args();
+        assert_eq!(
+            classify_group("composio_execute", &unknown),
+            EffectGroup::Send
+        );
+        assert!(!grantable("composio_execute", &unknown));
+
+        // And the split survives the round trip through the park queue: the
+        // group asserted above is the one the operator's card is built from.
+        let (p, queue) = queued_policy("supervised", &[]);
+        let _ = p.check(&request("composio_execute", send.clone())).await;
+        let queued = queue.drain(MAX_APPROVAL_REQUESTS_PER_TURN);
+        assert_eq!(queued.len(), 1);
+        assert_eq!(queued[0].effect.group, EffectGroup::Send);
+        assert_eq!(
+            queued[0].effect.payload, send,
+            "the card shows the arguments the agent actually sent, action key \
+             included"
         );
     }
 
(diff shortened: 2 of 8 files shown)


oxoxDev (commented):
This is the good kind of test PR: the fixtures named their action under `tool_slug`, which nothing on the classifier path reads, so every Composio test in these modules resolved through the unknown-is-a-send fallback and never reached the catalogue at all. Read and send came out identical, which means the read/send split had no coverage while appearing to have plenty. Building the fixtures from the same key the classifier reads, in one `test_support` module, is the right fix — and the new assertion that a read and a send classify *differently* is the one that would actually have caught it.

Two things.

**Asserting classification rather than the parking decision is the right choice, and worth stating in the PR body too.** You note it in the test comment — it keeps the assertion honest across #559, which changes whether a read parks but not what it is. That reasoning is the difference between a test that survives the next policy change and one that gets "fixed" by loosening it. It reads as an aside in a comment; it is really the design of the test.

**Question — is the dead key gone, or just unused?** If `tool_slug` still appears anywhere in fixtures or production arg-building, the same silent miss can be reintroduced by copy-paste from an older test. A grep-clean plus, ideally, the catalogue lookup logging or erroring when it is handed an arg shape with no readable action key, would close it properly. Right now the fallback is "unknown ⇒ send", which is safe but indistinguishable from "the classifier could not find the key at all" — the exact ambiguity that hid this for however long it has been there.

**Merge-order note:** this, #604 and #606 are all cut from the same parent and all three edit `src/policy/consequence.rs` and `src/harness/policy.rs`. The `DECLARED` table and the test modules will conflict pairwise. Since all three are yours, worth deciding a sequence up front rather than resolving the same table three times.

oxoxDev:
Pushed the review follow-up to this branch (beb4a55) rather than leaving it as a comment — shout if you'd rather own it.

**Dead key:** grep-clean in this repo. `tool_slug` survives only in prose and in the test_support assertion that it must not come back. (It is still live in openhuman's Composio *client* — `{"tool_slug": …}` in the request body it posts to the provider, plus the redaction allowlist — which is a different wire and correctly named there. Worth knowing, since that is presumably where the fixtures got the name.)

**The ambiguity you flagged:** `consequence_of` could not distinguish "the catalogue has never heard of this slug" from "these arguments carry no slug at all", and both silently produced a send. Added `composio_action_slug() -> Result<&str, ActionKeyMiss>` — `NotAnObject` / `KeyAbsent` / `NotAString` / `Empty` — logged at warn on the miss path. **The verdict is unchanged in every case**, so this can never loosen a decision; only the log differs. `standing_scope_of` reads through the same helper, so a call the classifier could not read a slug from cannot resolve a toolkit either.

Two tests: `a_missing_action_key_is_distinguishable_from_an_unknown_action` pins the reason per arg shape *and* that the verdict stays a send, and `an_unreadable_action_key_resolves_no_grant_scope` pins the scope side.

Cherry-picked onto #609 and #625 so the stack does not regress the parent on rebase.

Issue #562, opened 2026-08-10
approvals: the operator cannot change the autonomy tier from the console


oxoxDev:
## Symptom

An operator drowning in approval cards has no way to stop it. The autonomy tier is read from `[policy].mode` in the company manifest and nothing in the console reads or writes it.

There is no policy control anywhere in `frontend/src/views` — the only match for the word in the console is an unrelated comment in the workflow create dialog.

## The change

Surface the tier as a company setting the operator can see and change, with the consequence of each option stated in the operator's own terms rather than in tier names:

- **Read-only** — the agents can look at things but change nothing and spend nothing.
- **Auto** (once the tier exists) — the agents work on their own and stop before anything that leaves the company or spends money.
- **Supervised** — the agents ask before every change, including their own scratch files.
- **Full** — the agents act without asking, except for the few things on the always-ask list.

Show the always-ask list alongside it, and make it editable if that is cheap — it is the operator's real lever and today it is a manifest array with a default of `payment.send`, `filing.submit`, `external.publish`.

## Notes

- Writing the tier means writing the manifest, so this rides whatever path already persists company settings — worth checking whether that path exists before scoping.
- Changing the tier mid-flight should be safe by construction: the policy is read per call, so the next call picks up the new tier. Confirm rather than assume.

M3gA-Mind:
**Scoping pass — verified against `upstream/main` @ `843d5c2`.**

The absence claim is exact, and stronger than the body states. Searching the whole console — not just `views/` — for `policyMode`, `policy_mode`, `always_approve` and `alwaysApprove` returns **zero matches in `frontend/src/`**. The only hits for "policy" anywhere in `frontend/src/views/` are three unrelated comments:

- `WorkflowCreateDialog.tsx:90` (retry policy, in prose)
- `workflows/NodeDetailPanel.tsx:16` (error-handling policy)
- `workflows/graph.ts:90` (the default `stop` policy)

So there is nothing to extend — this is a new surface, not a broken one.

**Two dependencies worth naming before this is picked up:**

1. **The "Auto" option does not exist yet.** Three of the four tiers this issue proposes to show are real today; `auto` is #560. Either this ships with three options and gains the fourth later, or it waits. Shipping three is fine and probably better — the operator drowning in cards today gets `full` as an escape hatch immediately.

[...]

Issue #560, opened 2026-08-10
approvals: no tier between "ask before every write" and "ask before almost nothing"


oxoxDev:
## Symptom

An operator has two settings and needs a third. `supervised` parks every write and every outward read — unusable for real work. `full` parks nothing except the `always_approve` list (`payment.send`, `filing.submit`, `external.publish`) — more open than most operators want, because it also stops asking before a shell command, an arbitrary HTTP request, a git push, or an image generation that costs money.

So companies run supervised and drown, or run full and lose the gate that matters.

## The change

A fourth tier — the default the product should ship with. Working name `auto`, matching what Claude Code calls auto/accept-edits mode.

Its contract, stated as the operator would read it: **the agent works without interrupting me, and stops before anything that leaves the building or spends money.**

Runs unattended:
- Every `Reach::Nothing` tool (already true in all tiers)
- Every outward **read** — Composio reads (see sub-issue on that), `mcp_list_tools`, catalogue listings
- The agent's own sandbox writes: `file_write`, `edit`, `apply_patch`, `csv_export`, `memory_store`

Still parks:
- `shell`, `http_request`, `curl`, `web_fetch` — arbitrary code, arbitrary addresses
- `git_operations` — can push to a configured remote
- `workspace_write` — overwrites guidance the operator authored
- `publish_artifact`, `media_generate_image`, `media_generate_video` — outward or billed
- `mcp_call_tool`, `mcp_registry_tool_call` — can perform any effect the third-party server advertises
- `composio_execute` for any Write/Admin action, and `composio_authorize`
- `run_workflow` — performs whatever the workflow performs, which this layer cannot see
- Everything in `always_approve`

`Standing::Grantable` already marks almost exactly the sandbox-write set, so the split largely exists in the declaration table already — this tier is mostly reading a field that is already there.

## Also

Decide whether `auto` becomes the value `default_policy_mode()` returns, and whether the shipped presets should declare it explicitly rather than relying on the default. Changing the default changes behaviour for every existing company on next load, so it wants a deliberate call rather than a silent flip.

## Acceptance

- `[policy].mode = "auto"` parses, and an unknown mode string still falls back to the safe default.
- Under `auto`: a sandbox write and an outward read both run unattended.
- Under `auto`: `shell`, `git_operations`, `workspace_write`, a Composio write, and a media generation all still park.
- `always_approve` still wins over the tier, as it does over `full` today.
- `readonly` and `supervised` behaviour is unchanged.

[...]

PR #604, opened 2026-08-11
feat(policy): add the `auto` autonomy tier between supervised and full (#560)

M3gA-Mind:
## Summary

Adds `auto`, a fourth autonomy tier between `supervised` and `full` (#560).

An operator had two settings and needed a third. `supervised` parks every write — including the agent's own scratch files — so companies drown in cards. `full` parks nothing but the always-ask list, so it also stops asking before a shell command, an arbitrary HTTP request, or a git push. Companies ran one and suffered, or ran the other and lost the gate that mattered.

`auto`'s contract, as the operator reads it: **the agent works without interrupting me, and stops before anything that leaves the building or spends money.**

The tier line is one predicate, `Consequence::parks_under_auto`:

```rust
self.reach.parks_under_supervision() && !self.standing.is_grantable()
```

It reads the declaration table that already exists rather than adding a second list. `Standing::Grantable` already marks almost exactly the set `auto` wants to run — the agent's own sandbox writes (`file_write`, `edit`, `apply_patch`, `csv_export`, `memory_store`) and reads scoped to one connected account — and everything that can run arbitrary code, reach an arbitrary address, overwrite operator-authored guidance, spend on submit, or perform an effect this layer cannot see is `PerCall`, argued tool by tool in `DECLARED`. A fresh list would be exactly the hand-maintained carve-out `src/policy/consequence.rs` was written to delete.

### Named trap: the fifth site, which is not in the issue

The issue and its scoping comment name three exhaustive match sites. There is a **fourth consumer and a fifth site**, and the fifth is the one the feature turns on:

`POLICY_MODES` (`src/company/types.rs`) is the *manifest validator's* list, read at `manifest.rs:233`. It runs **before** `PolicyMode::parse` ever sees the string. Add the variant and the parse arm but not the list entry, and `[policy].mode = "auto"` is rejected at load with "must be one of readonly, supervised, full" — the tier is unreachable from the only place anybody sets it.

**And every test in `harness::policy` still passes**, because they all construct a `Policy` directly and never cross that boundary. Right assertion, input never reaches the code.

(The fourth consumer is `src/workflows/caps/mod.rs:125`, which parses the mode for the workflow capability bundle — see the `autonomy_for` section, where it matters.)

### Decision 1 — `autonomy_for(Auto)` maps to `Supervised`, not `Full`

This is a security decision, not a naming one, and the tempting answer is wrong.

`auto` is *more* permissive than `supervised` at opencompany's `ApprovalPolicy`. Matching that feel here and mapping to `AutonomyLevel::Full` would loosen a **different** gate, because this policy is not always sitting underneath that one:

- A workflow `tool_call` node runs through `WorkflowToolInvoker` (`src/workflows/caps/tools.rs`), which gates on the `[tools].allow` grant list and **this policy** — no `ApprovalPolicy` is installed on that path. (Workflow *agent* nodes go through the roster and do have one.) For those nodes this policy is the entire tier.
- Upstream `gate_decision` (`command_checks.rs:157`): `Supervised` prompts on `CommandClass::Write`; `Full` **allows** it. And `Write` is the fail-closed default class for any command not recognised as read/network/destructive — so this is the ordinary case, not an exotic one.

Mapping `auto` to `Full` would therefore let routine state-changing shell commands run unprompted in workflow nodes on an `auto` company, while the tier's stated contract is that `shell` parks. The inverse of what the operator selected.

It costs nothing in the other direction: every tool `exec_security` governs is `PerCall` and parks under `auto` at the authoritative layer anyway. `auto` buys its autonomy in tools this policy does not govern.

**Worth flagging for review:** the intuitive test case does not discriminate here. `Destructive` prompts under `Supervised` **and** under `Full`, so `rm -rf /` cannot tell the two mappings apart — a test written around it passes whichever mapping is chosen. I wrote that test first and it failed for this reason. `Write` is the whole of the difference, and that is what the test now asserts.

### Decision 2 — `require_approval_for_medium_risk` names `Auto` explicitly

`SecurityPolicy` set this from `mode == PolicyMode::Supervised` — exhaustive by accident, and it answers `false` for a new variant.

That is not cosmetic given Decision 1. The flag is **inert unless `autonomy == Supervised`**, so leaving it would have paired `Supervised` autonomy with the medium-risk gate switched off — the same loosening the mapping was chosen to prevent, arriving through the back door. Now `matches!(mode, Supervised | Auto)`.

### Public API removal — `PolicyMode::security_tier()` is **deleted**

Calling this out under its own heading so it is not discovered in the diff.

`security_tier()` was dead: repo-wide, its only references were its own test at `policy.rs:977-985`. It was deleted rather than given a fourth arm because **its documented premise is what this issue invalidates** — a 1:1 mapping onto OpenHuman's security-tier *words*, and `auto` has none. Both available arms were wrong: `"auto"` names a tier upstream does not have; `"supervised"` breaks the 1:1 the function exists to assert. Keeping a dead accessor alive by making it lie is worse than removing it.

Its load-bearing assertion — `parse("bogus") == Supervised`, the safe-fallback criterion — is preserved in the replacement test.

### Two boundaries `auto` deliberately does not draw

Stating these plainly rather than letting them read as oversights:

- **`Reach::Money` does not park.** `web_search` is billed but changes nothing, and it already runs unattended under `supervised` — for a reason that binds harder here: openhuman resolves a `RequireApproval` inline and never re-dispatches, so a parked search is a search that never happens and an agent with no search invents citations. `auto` must not be *stricter* than the tier it replaces. This is a pre-existing property of `supervised`, not something introduced here; the per-agent daily cap is the real spend boundary and it sits above the tier dispatch. Generation that spends on submit (`media_generate_image`, `media_generate_video`) is `Reach::Consequence` and parks.
- **`always_approve` is not consulted by the tier.** It is checked above the dispatch and wins over every tier, `full` included — asserted for `auto` by test.

### Sequencing: this composes with #559, it does not depend on it

The brief sequenced this after #559. That turns out to be unnecessary, and saying so here so nobody re-serialises them.

A Composio read is `Standing::Grantable` **today**, while still carrying `Reach::Consequence`. #559 reclassifies its *reach* and leaves its standing alone. `parks_under_auto` keys on both, and the `Grantable` half already decides the answer — so a Composio read runs unattended under `auto` in both worlds. **The two can land in either order and neither can silently invert the other.**

### Related

- #559 — the outward-read reclassification (composes; see above).
- #460 — fixes the workflow `tool_call` approval gap from the other side. That work and the `autonomy_for` reasoning here agree about the same hole; readable together.
- #562 — the console control for this tier.

### Not in this PR, by design

`default_policy_mode()` is **untouched** and still returns `"supervised"`. The issue asks whether `auto` should become the default; it is a one-line change, which is exactly why it wants a deliberate call rather than a rider — every company with no `[policy]` block changes behaviour on next load. Filed separately.

## API Or Behavior Changes

**Public API — removal:** `PolicyMode::security_tier()` is deleted (dead in-tree; see above).

**Public API — additions:** `PolicyMode::Auto` variant; `PolicyMode::ALL`; `Consequence::parks_under_auto()`. `PolicyMode` is a non-`#[non_exhaustive]` public enum, so the new variant is technically breaking for any external exhaustive match; there are none in-tree.

**Behaviour — new opt-in only.** `[policy].mode = "auto"` is now accepted and selects the new tier. No existing company changes behaviour: the default is unchanged, and `readonly` / `supervised` / `full` are untouched (asserted by test). An unknown mode string still falls back to `Supervised`.

**Manifest validation:** the error text for a bad `[policy].mode` now reads `readonly, supervised, auto, full`.

## Tests

Commands run locally, from a clone with `git submodule update --init --recursive`:

- [x] `cargo fmt --all -- --check` — **ran, exit 0.**
- [x] `cargo clippy --all-targets -- -D warnings` — **ran; fails, and the failures are pre-existing and not in this diff.** Both are `large size difference between variants` in `vendor/tinyagents/src/harness/agent_loop/stream.rs:48` and `:97`. This diff touches **zero** files under `vendor/` (`git diff --name-only upstream/main | grep -c '^vendor/'` → 0). `-p opencompany` does not avoid it — the vendored crate is a path dep and is linted either way. **No clippy finding in opencompany's own sources.** Flagging rather than claiming this box.
- [x] `cargo build --all-targets` — **not run as `build`; ran `cargo check --features openhuman,tinycortex --all-targets`, exit 0.** Narrower than the box (check, not build) and wider (the gated feature lane, which the plain command omits).
- [x] `cargo test` — **ran the gated lane, `cargo test --features openhuman,tinycortex`, exit 0: 2737 passed, 0 failed, 3 ignored** (plus the integration targets: 10, 1, 11, 2 passed). Ran the gated lane rather than bare `cargo test` because default features skip roughly a third of the code, including everything this change touches.

Exit codes captured from unpiped redirects, not from a pipeline's last stage.

### Revert-and-check

Every new test was proven to fail with its fix reverted, one revert at a time, verifying the edit applied before running:

| Revert | Test that must fail | Result |
| --- | --- | --- |
| Drop `"auto"` arm from `PolicyMode::parse` | `every_tier_is_reachable_from_a_manifest_and_parses_to_itself` | ✅ failed |
| Drop `"auto"` from `POLICY_MODES` | `every_tier_is_accepted_by_name_from_a_company_toml` | ✅ failed |
| `parks_under_auto` ignores `Standing` | `auto_runs_sandbox_writes_and_outward_reads_but_parks_anything_that_leaves` | ✅ failed |
| `autonomy_for` maps `Auto → Full` | `auto_borrows_supervised_exec_security_rather_than_full` | ✅ failed |
| `autonomy_for` maps `Auto → Full` | `auto_gates_write_class_commands_exactly_as_supervised_does` | ✅ failed |
| `require_approval_for_medium_risk` back to `mode == Supervised` | `auto_borrows_supervised_exec_security_rather_than_full` | ✅ failed |
| Remove the `always_approve` arm from `check()` | `auto_yields_to_always_approve_and_leaves_the_lower_tiers_alone` | ✅ failed |

**Two of these tests were vacuous on the first attempt, and revert-and-check is the only reason that is not still true.** Both walked `POLICY_MODES` to check `POLICY_MODES`, so deleting `"auto"` from it removed the *case* rather than failing the *assertion* — the identical shape to the trap they exist to catch. Fixed in `c653a60`: the policy-side cases now come from a third list (`PolicyMode::ALL`, kept honest by an exhaustive `match` the compiler will refuse to let drift), and the manifest-side cases are literals that the reverting edit cannot remove.

### Not verified

- **Not run in a live app.** No manual/end-to-end run of a company on `mode = "auto"`; the tier is proven at the policy and exec-security layers by unit test only.
- **The workflow `tool_call` path is reasoned about, not exercised.** The claim that no `ApprovalPolicy` sits above `WorkflowToolInvoker` comes from reading the construction path and confirming `consequence_of` has no call site there — not from a run.
- **No console/UI work** (#562).
- **Coverage percentage not measured.**

## Documentation

- `docs/spec/company-brain/approvals.md` — new **The `auto` tier** section (contract, why it reads `Standing`, the two boundaries it does not draw, the #559 composition, why it is not the default); tier list, mode dispatch and the "may it run unattended?" line updated; the deletion of `security_tier()` and the loss of the 1:1 mapping recorded.
- `docs/spec/runtime/manifest.md` — `[policy].mode` schema line, the prosumer validation-error example, and the `[policy]` description.
- Rustdoc carries the arguments at the sites that need them: `Consequence::parks_under_auto`, `autonomy_for` (the `Supervised`-not-`Full` argument), `exec_security`, `PolicyMode`, `PolicyMode::parse` (the validator/parse two-place trap), `POLICY_MODES`, and `Policy.mode`.

Both Markdown files remain under the 500-line limit (464 / 331).

Refs #560

### DIFF
--- a/docs/spec/company-brain/approvals.md
+++ b/docs/spec/company-brain/approvals.md
@@ -17,8 +17,17 @@ Effect kinds that MAY require sign-off, grouped by what they risk:
 | **Identity** | handle registration/renewal, key rotation, delegated signer mint/expand | always approval |
 
 `readonly` mode gates *every* effect; `full` mode auto-allows everything
-except `[policy].always_approve` entries. Modes mirror OpenHuman's security
-tiers so an OpenHuman-backed `ApprovalGate` maps 1:1.
+except `[policy].always_approve` entries. `auto` sits between `supervised` and
+`full`: the agent's own sandbox writes and its outward reads run unattended,
+and anything that leaves the company or spends on submit still parks — see
+[the tier line](#the-auto-tier) below.
+
+Three of the four names are OpenHuman's own security tiers. `auto` is not, so
+the mapping is no longer 1:1 and `PolicyMode::security_tier()` — the accessor
+that asserted it was — has been deleted rather than made to lie. Where the two
+vocabularies still have to meet, `harness::toolbelt::autonomy_for` borrows
+`Supervised` for `auto`; the argument is on that function and matters, because
+a workflow `tool_call` node has no `ApprovalPolicy` above it.
 
 ## Approval lifecycle
 
@@ -213,7 +222,8 @@ So the two questions are now separate answers from one declaration
 (`src/policy/consequence.rs`), one entry per tool:
 
 - **may it run unattended?** — `readonly` denies anything that mutates or
-  reaches outside; `supervised` parks it. This is what the approval card is for.
+  reaches outside; `supervised` parks it; `auto` parks only the part of it that
+  leaves the company. This is what the approval card is for.
 - **may an operator hand it over for a stretch of time?** — refused for anything
   that can execute arbitrary code (`shell`), reach an arbitrary address
   (`http_request`, `curl`, `web_fetch`, `git_operations`), act through a
@@ -226,6 +236,52 @@ So the two questions are now separate answers from one declaration
 What stays grantable is the low-consequence middle the feature exists for:
 writes confined to the agent's own sandboxed workspace (`file_write`, `edit`,
 `apply_patch`, `csv_export`, `memory_store`), and **Composio reads**.
+
+### The `auto` tier
+
+That low-consequence middle is also the whole of the `auto` tier (issue #560).
+An operator had two settings and needed a third: `supervised` parks every write
+including the agent's own scratch files, so companies drown in cards; `full`
+parks nothing but the always-ask list, so it stops asking before a shell
+command or a git push too. Companies ran one and suffered, or ran the other and
+lost the gate that mattered.
+
+`auto`'s contract, in the operator's words: **the agent works without
+interrupting me, and stops before anything that leaves the building or spends
+money.** The line is one predicate,
+`Consequence::parks_under_auto` — a call parks under `auto` when it would park
+under `supervised` **and** it is not `Standing::Grantable`.
+
+It reads the existing declaration rather than adding a list, because the split
+was already there and argued tool by tool. That reuse does widen what
+`Grantable` means — from "an operator may hand this to one teammate until a
+deadline" to "this runs unattended for everyone while the company sits in
+`auto`" — which is sound because `Standing` is decided by what a tool can reach
+rather than by what it is called, and because the widening is exactly what the
+operator chooses when they select the tier. It is written down so that a later
+edit loosening `Grantable` knows it is loosening two things.
+
+Two boundaries `auto` deliberately does not draw:
+
+- **`Reach::Money` does not park.** `web_search` is billed but changes nothing,
+  and it already runs unattended under `supervised` for a reason that binds
+  harder here — openhuman resolves a `RequireApproval` inline and never
+  re-dispatches, so a parked search is a search that never happens and an agent
+  with no search invents citations. `auto` must not be stricter than the tier it
+  replaces. The per-agent daily cap holds spend, and it sits above the tier
+  dispatch. Generation that spends on *submit* (`media_generate_image`,
+  `media_generate_video`) is a `Consequence` and parks.
+- **`always_approve` is not consulted by the tier at all.** It is checked above
+  the dispatch and wins over every tier, `full` included.
+
+The tier composes with the Composio read/send reclassification rather than
+depending on it: a Composio read is `Grantable` today while still carrying
+`Reach::Consequence`, and reclassifying its *reach* leaves its standing alone,
+so it runs unattended under `auto` either way.
+
+`auto` is **not** the default. Issue #560 argues it should become one, but
+`default_policy_mode()` returning it would change behaviour for every existing
+company with no `[policy]` block on its next load, so that is its own decision.
 
 `composio_execute` is the reason arguments are part of the question. Every
 Composio action arrives under that one tool name, so classifying the name
@@ -298,7 +354,7 @@ A tool call is decided in this order:
    and it re-checks the live arguments, so a grant minted on a Composio read
    cannot admit a Composio send.
 4. `[policy].always_approve` → park.
-5. mode dispatch (`readonly` / `supervised` / `full`).
+5. mode dispatch (`readonly` / `supervised` / `auto` / `full`).
 
 The grant sits **above** `always_approve` on purpose. A tool on that list still
 parks the first time, which is what the list is for; but once the operator has
--- a/docs/spec/runtime/manifest.md
+++ b/docs/spec/runtime/manifest.md
@@ -13,8 +13,8 @@ a safe default; the defaults produce a working company with only
 
 Parsing lives in `src/company/manifest.rs` (`CompanyManifest::from_path`,
 serde + validation). Validation errors MUST be actionable in prosumer
-language ("`[policy].mode` must be one of readonly, supervised, full — you
-wrote `supervized`"), never serde traces.
+language ("`[policy].mode` must be one of readonly, supervised, auto, full —
+you wrote `supervized`"), never serde traces.
 
 ## Full schema
 
@@ -70,7 +70,7 @@ allow = ["web.*", "docs.*", "search"]  # company-wide grant; agents intersect
 search_daily_calls = 200           # per-company daily web_search cap (0 = paused)
 
 [policy]                           # see company-brain/approvals.md
-mode = "supervised"                # readonly | supervised (default) | full
+mode = "supervised"                # readonly | supervised (default) | auto | full
 always_approve = ["payment.send", "filing.submit", "external.publish"]
 auto_approve_under_usd = 1.0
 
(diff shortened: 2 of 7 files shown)


oxoxDev (commented):
The tier's contract — *the agent works without interrupting me, and stops before anything that leaves the building or spends money* — is a sentence an operator can actually hold, and deriving the line from `Consequence::parks_under_auto` over the existing `Standing::Grantable` declaration rather than a new list is the right instinct: the split was already argued tool by tool. Deleting `PolicyMode::security_tier()` instead of making it lie about a 1:1 mapping that no longer exists is the correct call and the kind of thing that usually gets fudged.

**Major — the reuse of `Grantable` changes its meaning, and the write-up is more honest about that than the code is.**
You name it precisely: `Grantable` moves from "an operator may hand this to one teammate until a deadline" to "this runs unattended for everyone while the company sits in `auto`". That is a real widening, and the argument for it (standing is decided by reach, not by name) holds. But the note lives in a doc and a PR body, while the trap is in `consequence.rs` — someone loosening a single tool to `Grantable` for a delegation reason will not read `approvals.md` first.

Put it on the declaration site: a comment on `Standing::Grantable` itself saying it now has two consumers and that loosening it grants unattended execution under `auto`. One sentence where the edit happens is worth more than three paragraphs where it doesn't.

**Question — is `parks_under_auto` asserted per tool anywhere, or only as a predicate?**
The predicate is easy to verify in isolation; what an operator feels is *which specific tools* changed. A table test pinning a representative set — a sandbox write (runs), an outward read (runs), a send (parks), `shell` (parks), a spend-on-submit media generate (parks) — would fail loudly if a future `Grantable` edit silently moves a tool across the line. That is the regression the widening above makes possible, and the predicate test alone will not catch it.

**Good catch worth calling out:** the `POLICY_MODES` validator running *before* `PolicyMode::parse` means a tier added to the enum and parser but not that list is rejected at load while every `harness::policy` test still passes, because they construct `Policy` directly and never cross the boundary. Adding the test for exactly that trap, with literal mode words rather than derived ones, is the right shape.

**Not-a-default is the right call.** `default_policy_mode()` returning `auto` would change behaviour for every existing company with no `[policy]` block on its next load. Keeping that as its own decision is correct; #560 arguing otherwise doesn't make it this PR's to take.

**Merge-order note:** this, #603 and #606 are cut from the same parent and all three edit `src/policy/consequence.rs` and `src/harness/policy.rs` — pairwise conflicts in the `DECLARED` table and the test modules. All three are yours, so worth sequencing deliberately. This one is the largest, so it probably wants to go last.

M3gA-Mind:
@coderabbitai review

oxoxDev:
Cross-reference to #612 (`workflows: consult ApprovalPolicy for tool_call nodes`), opened about an hour after this one — flagging here because the interaction is invisible to git.

This PR's `autonomy_for` documents the `auto` → `AutonomyLevel::Supervised` borrow as a security decision, and the stated reason is the workflow path:

> **except** on the workflow `tool_call` path, where no such gate is installed and this policy is the whole tier.

#612 installs that gate — `apply_tool_call_gates` consults the company's `ApprovalPolicy` for `tool_call` nodes before compile. Once both land the sentence is false, and the argument for `Supervised` over `Full` loses its stated basis.

Worth settling explicitly rather than by merge order: with both in, a `tool_call` node under `auto` is asked about by `ApprovalPolicy` *and* carries `require_approval_for_medium_risk: true` from `matches!(mode, Supervised | Auto)` here. Either that is deliberate defence-in-depth, or `auto` does not deliver its own contract on the workflow path — the one place this PR chose extra strictness precisely because nothing else was watching. Whichever merges second needs to fix the other's doc comment; both files compile fine with the stale claim in place.

oxoxDev:
Pushed both review items to this branch (cd536b3) — revert if you'd rather write them yourself.

**The widening now lives at the declaration site.** `Standing` and the `Grantable` variant carry a note that this field decides two things since #560 — delegable to one teammate, *and* runs unattended for everyone under `auto` — and that loosening a tool for a delegation reason moves it across the tier line as a side effect.

**Per-tool pin:** `the_auto_tier_line_is_pinned_tool_by_tool` walks `declared_tools()` and asserts the exact set of tools that park under `supervised` and run under `auto` — today `apply_patch`, `csv_export`, `edit`, `file_write`, `memory_store`. It is derived from the whole table rather than a sample, so a tool joining or leaving that set fails here and has to be named deliberately; the predicate test alone would not notice. Plus the other direction (shell, http_request, git_operations, workspace_write, publish_artifact, both media generates, mcp_call_tool, run_workflow, undeclared) and the boundary you argued: `web_search` is billed and still does not park.

The argument-classified half is separate and feature-gated: `the_auto_line_reads_composio_arguments_not_the_tool_name` — catalogue read runs, send parks, unclassified action parks.

Your test through `check()` already covered the same tools; this one is the exhaustive complement, so a *new* tool cannot join the unattended set silently.

Issue #441, opened 2026-08-06
approvals: a Composio action can never be granted standing, because consequence is read from the tool name and not the action


oxoxDev:
## What happens

An operator approves a teammate's use of Composio, expecting that teammate to be able to keep using it. It asks again on the very next call, and on every call after that. There is no setting anywhere that changes this.

Reported from a staging tenant, 2026-08-06.

## Why it happens

Consequence is classified from the **tool name alone**. Every Composio action arrives as one tool, `composio_execute`, and that name is classified as a **send** effect — reaching a third-party account. The action being run lives in the call's arguments and the classifier never reads them.

Only effects classified as "other" may be granted standing. So no Composio action can ever hold a standing grant, and the scope control correctly does not appear.

The consequence of a Composio call is not a property of `composio_execute`. It is a property of the action: listing issues, reading a thread and searching a workspace are reads with no external effect at all; sending an email, posting a message and opening a pull request are the sends the classification was written for. Collapsing both into one verdict means the cautious answer has to win for everything, and the operator pays for it on every read.

## The lever the operator reached for does not exist

The manifest offers three knobs, and none of them is "this tool is fine for this teammate":

| Knob | What it actually does |
|---|---|
| `mode` | `readonly` / `supervised` / `full` — company-wide, all tools at once |
| `always_approve` | a list of kinds that **always park**. It means *always ask*, not *auto-approve* — the name reads as the opposite of its behaviour |
| `auto_approve_under_usd` | a money threshold, and a read costs nothing, so it never applies |

So the only way to stop the asking is to set the whole company to `full`, which turns the gate off for everything. That is precisely the escape the standing-grant work was built to remove, and an operator worn down by repeated prompts will take it.

`always_approve`'s name is worth fixing on its own. An operator reading the manifest will reasonably conclude it does the opposite of what it does.

## What the fix should do

Classify a Composio call by **what it is actually going to do**, not by the name of the tool that carries it:

- Read the action from the arguments and classify on that. A read is an "other" effect and can hold a standing grant; a send stays a send and keeps its per-call decision.
- Do not build a hand-maintained list of read-only slugs. That drifts the moment a provider adds an action, and the provider's own catalogue is the source of truth for what an action does.
- A slug that cannot be classified confidently must fall back to the **cautious** verdict, not the convenient one. An unknown action is a send until proven otherwise.
[...]

PR #606, opened 2026-08-11
fix(policy): classify read_workspace_state as a consequence (#459)

### LINKED ISSUE #459: approvals: read_workspace_state runs git in a directory the agent can write, so repository config decides what runs
## What this is

`read_workspace_state` is classified as a pure read: it never parks under `supervised` and is allowed under `readonly`. That classification describes its *intent* accurately and its *mechanism* not at all — it shells out to `git`, in a directory whose entire contents the agent controls.

`vendor/openhuman/src/openhuman/tools/impl/system/workspace_state.rs:202-207`:

```rust
async fn run_git(dir: &std::path::Path, args: &[&str]) -> anyhow::Result<String> {
    let output = tokio::process::Command::new("git")
        .args(args)
        .current_dir(dir)
        .output()
        .await?;
```

`execute` calls it twice (`:73`, `:88`) with `status --porcelain` and `log --oneline`. There is no `GIT_CONFIG_NOSYSTEM`, no `-c` hardening and no environment scrub, so git reads `.git/config` from that directory like any other repository. Several git config keys name a command to run — `core.fsmonitor` is invoked by `git status`, and there are others.

The directory is `{root}/{company}/{agent}/workspace`, which is exactly where the agent's own file tools write, and `file_write` will happily create a `.git/` and the files under it.

## Why this belongs on the approvals boundary

The two halves sit on opposite sides of the gate, which is what makes it a classification question rather than an upstream one:

| Tool | Verdict today |
|---|---|
| `file_write` | parks once, and **may be granted standing** — it is one of the low-consequence tools the standing grant exists for |
| `read_workspace_state` | **never parks**, allowed in every mode including `readonly` |

So the sequence "grant a scope on writing files, then read workspace state" is approved once and thereafter unattended, while the second step is not the read the classification says it is. `shell` itself is correctly per-call and non-grantable; this is a path around it rather than through it.

Predates the #444 boundary work — both classifications were the same before it. It was found by the sweep that lane ran over every wired tool, not introduced by it.

## What the fix should do

Pick one, in the classifier rather than by narrowing the feature:

- **Harden the invocation** so the tool means what its classification says: run git with `GIT_CONFIG_NOSYSTEM=1`, `GIT_CONFIG_GLOBAL=/dev/null` and `-c core.fsmonitor=` (plus the other command-valued keys), so repository config in an agent-writable directory cannot decide what executes. This keeps `read_workspace_state` a genuine read and is the option that preserves the behaviour operators want. It touches `vendor/openhuman`, so it needs an upstream change or a local wrapper.
- Or **reclassify** it as a consequence, so it parks under `supervised` and is denied under `readonly`. Honest, but it makes a routine orientation step cost an approval, which is the friction #441 and #443 are about.

The first is better. The second is the fallback if the vendored change cannot land quickly.

## How to verify

- With a crafted `.git/config` in an agent workspace naming a command, `read_workspace_state` produces the status output and runs nothing else.
- A real git repository in a workspace still reports status and log correctly.
- Whichever option lands, `src/policy/consequence.rs` and the tool agree — the coverage test already fails if a tool is undeclared, but not if a declaration is optimistic.

## Related

Found by the sibling sweep in the #441 / #443 / #444 lane. Same shape as those: a security-relevant decision keyed on what a tool is *called* rather than what it *does*.

M3gA-Mind:
## Summary

Closes #459 — but as a **stopgap**, and the PR should be read that way.

`read_workspace_state` is declared `Reach::Nothing`: it never parks under `supervised` and is allowed under `readonly`. That describes its intent and not its mechanism. It shells out to `git status --porcelain` and `git log` in `{root}/{company}/{agent}/workspace` — the directory the agent's own `file_write` writes into — and the vendored `run_git` sets no `GIT_CONFIG_NOSYSTEM`, no `GIT_CONFIG_GLOBAL`, no `-c` overrides and no environment scrub. Several git config keys name a command git then executes, so a `.git/config` the agent authored decides what runs.

This PR moves the declaration to `Reach::Consequence`, non-grantable, alongside `shell`. It parks under `supervised`, is denied under `readonly`, and can never be handed over on a standing grant.

**I confirmed the exploit rather than reasoning about it.** On git 2.50.1:

```
git init -q .
printf '#!/bin/sh\ntouch "$(dirname "$0")/PWNED"\nexit 1\n' > hook.sh && chmod +x hook.sh
git config core.fsmonitor "$PWD/hook.sh"
git status --porcelain     # -> PWNED exists
```

`git status` executes the command named by the repository's own config. Nothing about that is specific to a crafted repo — `.git/config` is an ordinary file that `file_write` can create.

### Why this is a stopgap, and why it is a blunt one

The real fix is upstream, in openhuman's `run_git` (`src/openhuman/tools/impl/system/workspace_state.rs`). It is **not merely unwritten — it is not straightforward**, which is the honest reason this reclassification is worth taking in the meantime:

- The exposure is the **repository** config, in a directory the agent controls. `GIT_CONFIG_NOSYSTEM` and `GIT_CONFIG_GLOBAL` — the hardening the issue body proposes — close the *system* and *global* files and do not reach `.git/config` at all.
- Clearing the command-valued keys with `-c` does work (command-line config outranks every file), but it is a denylist: correct only until git adds a key, and stale in the direction that reads as fixed.

So the upstream change has to *refuse* unrecognised repository config rather than scrub known-bad keys, which is a design question for openhuman's maintainers rather than a patch to copy in. I have opened that separately as a proposal.

**There is no pin to bump to.** I fetched the same path from `tinyhumansai/openhuman` `main` today and diffed it against the vendored pin (`vendor/openhuman` @ `d5ba6ab2c978ed24dd8ab74d38ae925f965a4682`): byte-identical, SHA-256 `5b7007f7334848b1631bb3256f72473838951af5c81b04b70413abd8052cebe6`. Upstream is not fixed either.

**Revert this when, and only when, a hardened `run_git` is vendored.** The declaration carries that instruction in a comment so the next person to read the table knows it is a placeholder and not a judgement.

### The cost, stated plainly

A routine orientation step now costs an approval under `supervised` and is refused under `readonly`. That is real friction and it is the friction #441/#443 exist to remove. I think it is the right trade while the mechanism does not match the label, but it is a product call and a reviewer may reasonably want it made explicitly.

## API Or Behavior Changes

Behaviour change, deliberate and user-visible:

- `read_workspace_state` under `supervised` — was: ran unattended. Now: parks for an operator.
- `read_workspace_state` under `readonly` — was: allowed. Now: denied.
- `read_workspace_state` standing grant — was: not grantable (unchanged; `d()` already meant `PerCall`). Explicitly asserted now so it cannot drift, since `file_write` *is* grantable and the pair together would reopen the hole for the length of a grant.

No API, wire, or schema change. The tool remains registered and in the same `shell` capability namespace (`namespace_of` already mapped it there — the classification now agrees with the gating).

## Tests

Commands actually run, from the repo root, on the gated lane this change lives in:

- [x] `cargo fmt --all -- --check` — run, exit 0. (It failed the first time — the new declaration needed wrapping — so this is a real pass, not an assumed one.)
- [x] `cargo clippy --all-targets -- -D warnings` — run as CI runs it on this lane: `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, exit 0. Worth noting for anyone repeating it: **without `--no-deps` this command fails**, on two pre-existing `large_enum_variant` lints inside `vendor/tinyagents`. Those are not from this change and CI does not see them, because the gated lane passes `--no-deps` (`.github/workflows/ci.yml:434`).
- [x] `cargo build --all-targets` — **not run as `build`**; ran `cargo check --features openhuman,tinycortex --all-targets` instead, exit 0. That is the gated lane and it covers `examples/`, which default features skip. Ticked because the box's intent (everything compiles, including examples) was verified, by a narrower command than the literal one.
- [x] `cargo test` — **not run bare**; ran `cargo test --features openhuman,tinycortex`. Default features skip roughly a third of this crate, so the bare command would not have exercised the gated policy surface this change touches.

Two new tests, plus two existing lists updated:

- `policy::consequence::tests::reading_workspace_state_is_classified_with_shell_because_it_runs_git` — pins reach, parking, readonly denial, and non-grantability, and pins the reach to `shell`'s so the two cannot drift apart.
- `harness::policy::tests::reading_workspace_state_parks_supervised_and_denies_readonly` — the end-to-end gate decision, which is what an operator actually experiences.
- Removed `read_workspace_state` from `a_workspace_read_never_parks_whatever_its_name_begins_with` and `a_workspace_read_runs_without_asking_whatever_its_name_begins_with`, each with a comment pointing at the new test. Those two lists asserted the old behaviour and are the reason the change is not silent.

**Revert-and-check**, run rather than reasoned about. With the declaration put back to `Reach::Nothing`:

```
test policy::consequence::tests::reading_workspace_state_is_classified_with_shell_because_it_runs_git ... FAILED
test harness::policy::tests::reading_workspace_state_parks_supervised_and_denies_readonly ... FAILED
test result: FAILED. 0 passed; 2 failed
```

and both pass again once restored. Being precise about what that does and does not prove: the two **new** tests carry the whole proof. The two **edited** lists still pass under the revert — removing an entry from a list cannot fail — so they are a necessary consequence of the change, not evidence for it. I mention it because "I edited four tests" would otherwise read as four tests' worth of assurance, and it is two.

## Documentation

No doc change needed — `grep -rn "read_workspace_state" docs/` returns nothing; the classification is documented at the declaration site, which is where this table's reasoning lives by convention. The stopgap status, the upstream dependency and the revert condition are all recorded in the comment above the entry.

### DIFF
--- a/src/harness/policy.rs
+++ b/src/harness/policy.rs
@@ -774,7 +774,8 @@ impl ToolPolicy for ApprovalPolicy {
         if self.mode == PolicyMode::Readonly && is_external_effect(tool, &request.arguments) {
             return ToolPolicyDecision::deny(format!(
                 "'{tool}' mutates or reaches outside; this desk is read-only, \
-                 so an earlier approval does not apply"
+                 so an earlier approval does not apply{}",
+                readonly_denial_suffix(tool)
             ));
         }
 
@@ -887,7 +888,8 @@ impl ToolPolicy for ApprovalPolicy {
             PolicyMode::Readonly => {
                 if reach.denied_under_readonly() {
                     ToolPolicyDecision::deny(format!(
-                        "'{tool}' mutates or reaches outside; this desk is read-only"
+                        "'{tool}' mutates or reaches outside; this desk is read-only{}",
+                        readonly_denial_suffix(tool)
                     ))
                 } else {
                     ToolPolicyDecision::Allow
@@ -895,6 +897,20 @@ impl ToolPolicy for ApprovalPolicy {
             }
         }
     }
+}
+
+/// The ", because …" a `readonly` denial carries when the tool's name argues
+/// against its classification (issue #459), or an empty string.
+///
+/// `readonly` denying `read_workspace_state` is the case this exists for: the
+/// operator sees a tier that promises reads still work refuse something called
+/// `read_*`, and "mutates or reaches outside" alone reads as a bug in the tier.
+/// The reason lives in [`crate::policy::consequence::denial_reason`], next to
+/// the classification it explains, so the two cannot drift.
+fn readonly_denial_suffix(tool: &str) -> String {
+    crate::policy::consequence::denial_reason(tool)
+        .map(|why| format!(" — {why}"))
+        .unwrap_or_default()
 }
 
 /// Does this tool call mutate state or reach an external counterparty?
@@ -2811,6 +2827,44 @@ mod tests {
             ),
             "`readonly` promises nothing runs; a git config key can name a command"
         );
+    }
+
+    /// …and the `readonly` denial says **why**, because this is the one an
+    /// operator ends up confused by.
+    ///
+    /// `read_workspace_state` used to run on a `readonly` desk, so a company
+    /// that sat there gets a refusal on what was a normal first move — and a
+    /// tier that promises reads still work refusing something called `read_*`
+    /// reads as a bug in the tier unless the message names git. The
+    /// `supervised` park explains itself by producing a card to approve; this
+    /// one has to carry its reason.
+    #[tokio::test]
+    async fn the_readonly_denial_of_a_read_shaped_tool_says_why() {
+        let ToolPolicyDecision::Deny { reason } = policy("readonly", &[], None)
+            .check(&request("read_workspace_state", serde_json::json!({})))
+            .await
+        else {
+            panic!("`readonly` denies it");
+        };
+        assert!(
+            reason.contains("git"),
+            "an operator must be able to tell this from a mis-classification: {reason}"
+        );
+        assert!(
+            reason.contains("workspace"),
+            "and where the config it obeys comes from: {reason}"
+        );
+
+        // A tool whose name already argues for the verdict carries no such
+        // clause — every denial ending in an explanation is an explanation
+        // nobody reads.
+        let ToolPolicyDecision::Deny { reason } = policy("readonly", &[], None)
+            .check(&request("shell", serde_json::json!({})))
+            .await
+        else {
+            panic!("`readonly` denies shell");
+        };
+        assert!(!reason.contains("git"), "{reason}");
     }
 
     /// A standing grant admits any arguments, which was a fair summary of a
--- a/src/harness/policy.rs
+++ b/src/harness/policy.rs
@@ -2763,6 +2763,9 @@ mod tests {
     /// read-only rule matched a *prefix* and none of these names begins with a
     /// read-only word. Nobody had reported them; they were found by asking the
     /// same question of every registered tool.
+    ///
+    /// `read_workspace_state` was in this list until issue #459 — see
+    /// [`reading_workspace_state_parks_supervised_and_denies_readonly`].
     #[tokio::test]
     async fn a_workspace_read_runs_without_asking_whatever_its_name_begins_with() {
         let p = policy("supervised", &[], None);
@@ -2772,7 +2775,6 @@ mod tests {
             "grep",
             "image_info",
             "list",
-            "read_workspace_state",
             "memory_recall",
         ] {
             assert_eq!(
@@ -2781,6 +2783,34 @@ mod tests {
                 "`{tool}` reads the agent's own workspace"
             );
         }
+    }
+
+    /// Issue #459: the sibling that turned out not to be a read at all. It
+    /// shells out to `git status` in the agent's own workspace, and the
+    /// vendored `run_git` lets that directory's `.git/config` — which
+    /// `file_write` can author — decide what git executes. So it goes through
+    /// the gate the way `shell` does, and this is the assertion an operator
+    /// actually feels.
+    #[tokio::test]
+    async fn reading_workspace_state_parks_supervised_and_denies_readonly() {
+        assert!(
+            matches!(
+                policy("supervised", &[], None)
+                    .check(&request("read_workspace_state", serde_json::json!({})))
+                    .await,
+                ToolPolicyDecision::RequireApproval { .. }
+            ),
+            "running git under agent-authored config must reach an operator"
+        );
+        assert!(
+            matches!(
+                policy("readonly", &[], None)
+                    .check(&request("read_workspace_state", serde_json::json!({})))
+                    .await,
+                ToolPolicyDecision::Deny { .. }
+            ),
+            "`readonly` promises nothing runs; a git config key can name a command"
+        );
     }
 
     /// A standing grant admits any arguments, which was a fair summary of a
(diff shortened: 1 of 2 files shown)


oxoxDev (commented):
Verified the premise rather than trusting the write-up, and it is exactly as stated. `run_git` in openhuman's `workspace_state.rs` is a bare `Command::new("git").args(args).current_dir(dir)` — no `GIT_CONFIG_NOSYSTEM`, no `-c` overrides, no env scrub — running in the directory `file_write` writes into. The contrast inside the same crate makes the case sharper than the description does: `shell.rs` calls `cmd.env_clear()`, and `node_exec.rs` and `python_exec.rs` do too. The codebase already treats "spawn a process for an agent" as needing a scrub; this path just never got one.

One more piece of evidence worth adding, because it turns this from a judgement call into a consistency fix: `git_operations` is **already** `Reach::Consequence` on `main` (`consequence.rs:204`), and its `run_git_command_in` is the same unscrubbed `Command::new("git")` shape. So `read_workspace_state` was the odd one out — the identical primitive gated in one tool and open in the other. Worth a line in the comment; it answers "why is this not overreach" before anyone asks.

**Question — where does the revert instruction live?**
The comment says revert to `Reach::Nothing` once a hardened `run_git` is vendored, "and not before". That is the right disposition, but it lives only in a code comment in this repo, while the fix belongs to a different repo entirely. Nothing will surface it when someone bumps the openhuman pin. Please open the openhuman-side issue for the hardened `run_git` and link it from that comment — a stopgap whose removal condition is recorded only in prose next to the stopgap is how these become permanent. Your own note that the real fix has to *refuse unknown config* rather than scrub known-bad keys is the most valuable thing here and deserves to be written down where the person doing it will find it.

**Question — blast radius of the `readonly` arm.**
`readonly` now denies `read_workspace_state` outright, where it previously ran. If an agent's orientation step calls it early, a `readonly` company gets a denial on what used to be a normal first move. Is that reachable in any default toolbelt/agent template today, and does the denial message tell the operator *why* (it runs git) rather than just "denied"? The approval-parking arm under `supervised` is self-explaining; the `readonly` one is where a confused operator ends up.

Otherwise this is the right call. The stopgap is honest about being blunt, the cost is named, and both directions are asserted.

oxoxDev:
Pushed all three review items to this branch (ef3fb39).

**Precedent line added** to the table comment: `git_operations` is already `Reach::Consequence` and its `run_git_command_in` is the same unscrubbed `Command::new("git")`, so this is a consistency fix rather than a judgement call — `read_workspace_state` was the odd one out.

**Revert condition now has a home:** tinyhumansai/openhuman#5494, linked from the comment. It states the exposure (repository config in an agent-writable dir, which `GIT_CONFIG_NOSYSTEM`/`GIT_CONFIG_GLOBAL` do not reach), that the fix has to *refuse unknown config* rather than scrub known-bad keys, that `shell.rs`/`node_exec.rs`/`python_exec.rs` already `env_clear()`, and that `run_git_command_in` needs the same. Assigned to you.

**Blast radius — yes, reachable.** `read_workspace_state` ships in the `shell` namespace and every bundled company grants `*`, so a `readonly` desk now gets a denial where the call used to run. No agent template forces it as a first move, so it is not guaranteed on an orientation step, but it is available on every default toolbelt.

So the denial had to say why: added `consequence::denial_reason`, appended to both `readonly` deny sites (the brake and the tier arm). The operator now reads "…this desk is read-only — it runs `git status` and `git log` in the agent's workspace, and git takes its configuration from that same directory — so it is gated like `shell` rather than like a read". Pinned by `the_readonly_denial_of_a_read_shaped_tool_says_why`, which also asserts `shell` carries no such clause — every denial ending in an explanation is an explanation nobody reads. Left the `supervised` park alone; it explains itself with a card.

PR #609, opened 2026-08-11
approvals: let a Composio read run under supervised

### LINKED ISSUE #559: approvals: a Composio read parks, so reading a mailbox costs the operator an approval

M3gA-Mind:
## Summary

Fixes #559.

A Composio **read** — checking a mailbox, listing a repo's pull requests, reading a
calendar — classified as `Reach::Consequence`, which is exactly the value `supervised`
parks on. Every page of every list raised its own approval card: it interrupted a person,
refused the call, and dead-ended the turn.

This adds **`Reach::ExternalRead`** and moves the Composio read branch onto it
(`src/policy/consequence.rs`). That is the whole behavioural change — one field on one
branch.

### Why `ExternalRead` and not `Reach::Money`

The issue offers both. Option 2 is correct, and the reason is one line of code:
`costs_money()` is `matches!(self, Self::Money)`, and that predicate feeds the **daily
spend cap**. Reusing `Money` would make every mailbox page count against the operator's
budget — failing this issue's own last acceptance criterion with the fix meant to satisfy
it. `ExternalRead` gives the same two tier answers as `Money` while keeping
`costs_money()` honest.

### Why nothing else had to change

All three tier questions are `matches!` predicates, and `Reach` is never matched
exhaustively outside `consequence.rs`. So the new variant answers correctly by
construction:

| predicate | `ExternalRead` | effect |
|---|---|---|
| `denied_under_readonly()` = `!matches!(self, Nothing)` | `true` | `readonly` still denies — criterion 4 |
| `parks_under_supervision()` = `matches!(self, Consequence)` | `false` | `supervised` runs it — criterion 1 |
| `costs_money()` = `matches!(self, Money)` | `false` | not counted as spend — criterion 5 |

Criteria 2 and 3 (sends, unrecognised slugs, missing/non-string `tool` still park) hold
because the `send` binding is shared by all those paths and only the read branch moved.
Asserted anyway in `a_composio_send_and_every_unclassifiable_call_still_park`, across all
six shapes — nothing structurally stops a later edit touching that shared binding.

`Standing::Grantable` is unchanged on the read — but see the finding below, because the
issue's stated reason for keeping it does not hold.

## API Or Behavior Changes

**Behaviour:** under `supervised`, a Composio action the vendored catalogue tags `Read`
now runs without raising an approval. Under `readonly` it is still denied. Sends, admin
actions and anything the catalogue cannot place are unchanged — they still park.

**API:** `Reach` gains a public variant, `ExternalRead`. Any downstream exhaustive `match`
on `Reach` would need a new arm; there are none in this crate.

## Tests

All exit-code gated via unpiped redirect (`cmd > log 2>&1; echo "EXIT:$?"`).

- [x] `cargo clippy --all-targets -- -D warnings` — EXIT:0 (default lane). Gated lane run
      as CI runs it, `--locked --no-deps --features openhuman,tinycortex --all-targets`
      — EXIT:0.
- [x] `cargo build --all-targets` — covered by the two clippy runs above, which build all
      coverage.
- [x] `cargo test` — `cargo test --lib` default lane and
      `cargo test --features openhuman,tinycortex --lib` gated lane (**2743 passed, 0
      failed**). Integration targets were run on this branch's parent (PR #603); not
      re-run here — the diff is library-only.

### Negative control

Revert the one-line change (`Reach::ExternalRead` → `Reach::Consequence` on the read
branch), change nothing else. Gated lane, EXIT:101, **4 failed**:

```
harness::policy::tests::a_composio_read_runs_under_supervision_without_parking
policy::consequence::tests::a_composio_read_runs_under_supervision_and_is_still_denied_under_readonly
policy::consequence::tests::a_composio_read_is_grantable_and_a_send_is_not
policy::test_support::tests::the_read_fixture_classifies_as_a_read
```

Restored; green again (2743 passed, 0 failed, EXIT:0).

This control is only possible because of PR #603 (issue #470): before it, the Composio
fixtures named their action under a key nothing reads, so no test reached the read/send
classification at all and this revert would have failed nothing.

**One assertion removed as vacuous.** I had written a `check(read) == Allow` into
`a_grant_scoped_to_one_provider_does_not_admit_another_providers_read`. It survived the
revert — that policy holds a standing grant admitting a GitHub read at step 2b, so the
tier never gets a say and the assertion proved nothing while looking like it proved the
change. Deleted, with a comment saying why, and the real coverage lives in
`a_composio_read_runs_under_supervision_without_parking`, which uses a policy with no
grant.

## A finding that contradicts the issue body — worth a maintainer's eye

> "Keep `Standing::Grantable` either way — it stops mattering for the supervised path but
> still governs `readonly` and any future tier."

**It does not govern `readonly`.** `ApprovalPolicy::check()` denies every external effect
under `readonly` at **step 1** (`src/harness/policy.rs:773`), *above* both the single-use
and standing grant checks — deliberately, and documented there as the emergency-brake
ordering from #243. So:

* `full` — everything allowed; scope irrelevant.
* `supervised` — after this change a read never parks, so the scope is never consulted.
* `readonly` — the brake denies before the grant checks run, so the scope is never
  consulted.

Net: **issue #457's per-toolkit standing-grant scoping is now unobservable for Composio
reads in every tier** — and `composio_execute` is, per `standing_scope_of`'s own doc
comment, "the exception the type exists for". The machinery is still correct; nothing
routes to it any more.

This surfaced as a real failure of
`a_grant_scoped_to_one_provider_does_not_admit_another_providers_read`, which used to
drive the scope check through `check()`. I did **not** relax it to green. It now pins
`standing_grant_allows` directly, so the scoping logic keeps its coverage, and the test
carries a comment explaining why it no longer goes through the gate.

Two things a maintainer may want to decide, neither of which I have done here:

1. Whether the `readonly` + standing-grant combination was *intended* to be reachable. If
   it was, the brake ordering — not this issue — is where the bug is.
2. Whether #457's scoping should now be considered dormant, and what that means for #560
   (the new `auto` tier), which reads this same classification and could route reads
   somewhere the scope *is* consulted.

## Documentation

`Reach::ExternalRead` carries its own doc comment explaining why it is separate from both
`Money` (the spend cap reads `costs_money()`) and `Nothing` (a `readonly` desk must not
reach a counterparty). The read branch in `composio_execute_consequence` records what the
old first-time park actually cost. No user-facing docs affected.

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -5620,7 +5620,7 @@ members = ["eng1", "eng2"]
             None,
         )
         .with_requests(requests.clone());
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL" });
+        let args = crate::policy::test_support::composio_send_args();
         let request = ToolPolicyRequest::new(
             "composio_execute",
             args.clone(),
@@ -5856,7 +5856,14 @@ members = ["eng1", "eng2"]
         let log: Arc<dyn crate::ports::EventLog> =
             Arc::new(crate::store::FsEventLog::new(dir.path().to_path_buf()));
         let requests = crate::harness::policy::ApprovalRequestQueue::default();
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL", "to": "a@b.test" });
+        // Issue #470: a real catalogued send, with the action's own parameters
+        // under `arguments` where the tool's schema puts them — so the
+        // re-dispatch path this test covers carries a call the classifier can
+        // actually read.
+        let args = crate::policy::test_support::composio_args_with(
+            crate::policy::test_support::COMPOSIO_SEND_SLUG,
+            serde_json::json!({ "to": "a@b.test" }),
+        );
         requests
             .grants()
             .grant(crate::runtime::grants::GrantedCall {
--- a/src/harness/policy.rs
+++ b/src/harness/policy.rs
@@ -1484,6 +1484,66 @@ mod tests {
         );
     }
 
+    /// Issue #559, at the gate an agent actually hits: a Composio read runs
+    /// under `supervised` instead of parking, and nothing is queued for a
+    /// human — while a send on the same tool still parks.
+    ///
+    /// This is the behaviour the issue is about. `a_composio_read_and_a_
+    /// composio_send_are_classified_differently` above pins what the two calls
+    /// *are*; this pins what the desk *does* with them, which is the part an
+    /// operator notices when every page of a mailbox raises a card.
+    #[tokio::test]
+    async fn a_composio_read_runs_under_supervision_without_parking() {
+        let (p, queue) = queued_policy("supervised", &[]);
+
+        assert_eq!(
+            p.check(&request("composio_execute", composio_read_args()))
+                .await,
+            ToolPolicyDecision::Allow,
+            "reading a connected account changes nothing and costs nothing"
+        );
+        assert_eq!(
+            queue.queued(),
+            0,
+            "no card was raised, so no human was interrupted"
+        );
+
+        // Paging the same list is not a second decision, because there was
+        // never a first one. This is the symptom the issue opens with.
+        for _ in 0..5 {
+            let _ = p
+                .check(&request("composio_execute", composio_read_args()))
+                .await;
+        }
+        assert_eq!(queue.queued(), 0);
+
+        // The send half is untouched: same tool, same desk, still parks.
+        assert!(matches!(
+            p.check(&request("composio_execute", composio_send_args()))
+                .await,
+            ToolPolicyDecision::RequireApproval { .. }
+        ));
+        assert_eq!(queue.queued(), 1);
+    }
+
+    /// A `readonly` desk still denies the read. That tier's contract is that
+    /// nothing outside the company is reached at all — #559 moves the read out
+    /// of the parking bucket, not out of the reaching-outward one.
+    #[tokio::test]
+    async fn a_readonly_desk_still_denies_a_composio_read() {
+        let p = policy("readonly", &[], None);
+        assert!(matches!(
+            p.check(&request("composio_execute", composio_read_args()))
+                .await,
+            ToolPolicyDecision::Deny { .. }
+        ));
+        assert!(matches!(
+            p.check(&request("composio_execute", composio_send_args()))
+                .await,
+            ToolPolicyDecision::Deny { .. }
+        ));
+    }
+
     /// `always_approve` parks regardless of tier — including under `full` — so
     /// that arm has to record its request too.
     #[tokio::test]
@@ -2896,11 +2956,25 @@ mod tests {
         );
     }
 
-    /// Issue #457, through the real admission path. Re-classifying the live
-    /// action (the test above) separates a read from a send; it cannot separate
-    /// one provider's read from another's, because both are reads under the same
-    /// tool name for the same teammate. The card said "read from GitHub" and the
-    /// grant has to mean that.
+    /// Issue #457's scope check, pinned **directly** rather than through
+    /// `check()`.
+    ///
+    /// It used to run through the real admission path, asserting that a grant
+    /// scoped to `github` admitted a GitHub read and re-parked a Gmail one.
+    /// Issue #559 made that unobservable from `check()`, and the honest thing
+    /// is to say so rather than relax the assertion until it passes:
+    ///
+    /// * under `supervised` a catalogue read no longer parks at all, so it is
+    ///   allowed by the tier long before the scope is consulted;
+    /// * under `readonly` the brake at step 1 denies every external effect
+    ///   *above* the grant checks, so the scope is not consulted there either;
+    /// * under `full` everything is allowed.
+    ///
+    /// So `standing_grant_allows` is still correct and still worth pinning —
+    /// this test does that — but no tier currently routes a Composio read to
+    /// it. See the note in the PR for #559; the issue's claim that
+    /// `Standing::Grantable` "still governs `readonly`" does not hold against
+    /// the ordering in `check()`.
     #[tokio::test]
     async fn a_grant_scoped_to_one_provider_does_not_admit_another_providers_read() {
         let queue = ApprovalRequestQueue::default();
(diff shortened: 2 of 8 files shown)


oxoxDev (commented):
Checked the two claims the whole design rests on, and both hold. `Reach` is genuinely never matched exhaustively outside `consequence.rs` — the only occurrences elsewhere (`harness/build.rs:461`, `harness/policy.rs:567/570/2708`) are prose in comments, not match arms — so a new variant really is inert until a predicate answers differently. And the three predicates are `matches!` one-liners, so the verdict table in the description is exactly what the code does.

`ExternalRead` over `Money` is the right call and for the right reason: `costs_money()` feeds the daily spend cap, so folding a mailbox read into `Money` would bill the operator per page and break the acceptance criterion the change exists to satisfy. The negative control — revert the one field, 4 targeted failures — is the evidence that makes this reviewable in minutes rather than an afternoon.

**Major — the comment kept on `Standing::Grantable` is wrong about `readonly`.**

> `Standing::Grantable` stays. It stops mattering for `supervised` now that nothing parks there, but it still governs `readonly` and any tier added later.

It does not govern `readonly`. The `readonly` brake denies *before* any grant is consulted — `harness/policy.rs:756` ("`readonly` outranks a grant — the brake wins (issue #243)") and `:813` ("`readonly` brake denies before either is consulted, leaving the grant"). The deny comes off `reach.denied_under_readonly()`, which reads `Reach`, never `Standing`.

So after this PR, `Grantable` on the Composio read is inert under `readonly` (brake outranks grants) *and* inert under `supervised` (nothing parks). Its one live consumer is the `auto` tier in #604, where `parks_under_auto` = parks-under-supervised **and** not-`Grantable`. That is a good reason to keep it — just not the reason written down. Worth correcting in place, because the next person deciding whether this field is dead code will read that comment and reach the wrong conclusion in either direction.

Your own summary already hints at this ("the issue's stated reason for keeping it does not hold"), so the body and the code comment currently disagree with each other.

**Stacking — this is cut on top of #603, not on `main`.**
Same eight files, and the test notes say integration targets were run on "this branch's parent (PR #603)". So this PR's diff carries #603's changes, and the two must land in order. Worth saying so in the description — a reviewer arriving cold sees `test_support.rs` and the `COMPOSIO_ACTION_KEY` visibility change here and cannot tell which PR owns them. Also: if #603 gets review changes, this branch needs re-cutting, and stacked branches in this repo have a habit of quietly regressing the parent's fix on rebase — re-grep the `tool` key after any rebase rather than trusting the merge.

**The pile on `consequence.rs` is now four deep.** #603 (parent of this), #604 (`auto` tier), #606 (`read_workspace_state`), and this one all edit `src/policy/consequence.rs` and `src/harness/policy.rs`. The semantics compose fine — I checked `auto` against `ExternalRead` and it lands correctly, since `parks_under_auto` derives from `parks_under_supervision`, which `ExternalRead` answers `false` — but the textual conflicts in the `DECLARED` table and the test modules are guaranteed. All four are yours; a stated order would save resolving the same table three times.

**Minor:** every check is still pending here, including both Rust lanes. Nothing to gate on yet.

oxoxDev:
Pushed the comment correction to this branch (f0ad3d2), plus #603's follow-up cherry-picked so the stack stays consistent.

The kept note now says what `Standing::Grantable` actually governs after this PR: **not** `readonly` (that brake denies off `Reach::denied_under_readonly` above any grant check and never reads `Standing`), and not `supervised` (nothing parks there for a catalogue read any more). What it does still govern is the **mint** side — `Effect::may_be_granted_standing` and the re-check in `standing_grant_allows`, which is what stops a Composio *send* arriving under the same tool name from being mintable — and it is the field any unattended tier has to read, which is exactly what #560's `parks_under_auto` derives. Removing it would make both unrepresentable.

Grepped for the same claim elsewhere: the only other occurrence is your own note in the `harness::policy` test module, which already says it does not hold.

Issue #607, opened 2026-08-11
workspace: the shared tree cannot be searched, so discovery costs one read per candidate


oxoxDev:
## Problem

The workspace is now genuinely shared: agents create nodes (#551), authorship is recorded (#326), published deliverables land under `Agents/<agent>/<task>/` in the tree (#552), and the console follows writes live (#327). Every agent can reach every other agent's output.

**But nothing can search it.** `WorkspaceStore` exposes `tree`, `read`, `write`, `create`, `rename_move`, `delete`, `is_empty` — seven methods, none of which answers "which notes mention X". The four agent-facing tools are `workspace_list`, `workspace_read`, `workspace_write`, `workspace_create`. The console has no search box either; the `filter` calls in `WorkspaceView.tsx` are folder/file kind splits, not text.

So the only way an agent finds anything it did not itself write is: `workspace_list` a folder, `workspace_read` each candidate, repeat per folder. That is one tool round-trip per hop, each pulling a full note body into context, with no way to know which hop is the right one before paying for it. The cost scales with how well the workspace is working — the more agents publish, the more expensive discovery gets, which is exactly backwards.

A tree you cannot search is a filing cabinet, not shared memory. Search is what turns the surface #551/#552/#570 built into something an agent will actually use.

## Proposal

A workspace search that returns *matching nodes with enough context to choose*, so the agent spends one call to locate and one to read, instead of N to hunt.

Sketch, not a specification — the design questions below are real and are the substance of this issue:

- Query over note **content and name**, scoped to one company (the port's existing per-company isolation invariant is non-negotiable: "Company A's files MUST be invisible to company B").
- Results carry the node id, name, path or parent, `updated_by`, and a short matching excerpt — enough for the caller to pick a node without reading it first.
- A bounded result count, with the zero-means-nothing lesson from #555 applied: whatever type expresses the limit must not have a value that silently means *unlimited*.
- A `workspace_search` agent tool alongside the existing four.
- A search box in the Workspace tab, using the same backend path, so the operator and the agents get the same answer.

### The design question this issue exists to decide

**Port method versus company-layer helper.** These are genuinely different bets and the plan should argue one:

[...]

Issue #570, opened 2026-08-10
workspace: an agent folder is provisioned for every agent at boot, even ones that never produce anything


oxoxDev:
## Summary

#551 provisions `Agents/<agent-id>/` eagerly — one folder per manifest agent, at boot and on every roster rebuild. So a company with twelve agents opens its Workspace tab on twelve empty folders, eleven of which may never hold anything.

The folder should be the record of an agent having produced something, not a promise that it might. This changes provisioning to scaffold the system roots eagerly and mint the per-agent folder on demand.

## Problem

The eager model has three costs, all visible on day one:

* **The tree opens noisy.** Twelve empty folders is the first thing an operator sees, and every one of them is indistinguishable from a folder whose contents were deleted.
* **An empty folder asserts something false.** `Agents/marketing-lead/` existing implies the marketing lead has a working area with something in it. Emptiness is the common case and reads as a bug.
* **It scales with the roster, not with the work.** Adding an agent to the manifest adds a folder whether or not that agent is ever dispatched.

`Desks/` does not exist at all today, so a desk that produces something has nowhere of its own to put it.

## Solution

1. **Two system roots, scaffolded eagerly on every company: `Agents/` and `Desks/`.** Empty, created at boot, independent of the roster — a company with no agents still gets both. These sit alongside the template-seeded top-level folders (`Playbooks/`, `Standards/`, and the per-company domain folder such as `Product/` or `Deals/`), which are untouched.

2. **`Agents/<agent-id>/` is minted on demand**, the first time that agent actually produces something — an artifact published by it, or its first `workspace_create` inside its own home. Idempotent: the second write finds the folder rather than duplicating it.

3. **`Desks/<desk-id>/` is minted on the same terms**, when a desk first produces an artifact.

4. **Existing companies pick up the roots on their next boot.** Provisioning already runs every boot and rebuild and is idempotent, so this needs no backfill migration.

5. **`workspace_create` auto-mints the calling agent's home.** Without this, an agent writing `Agents/<self>/notes.md` hits the missing-parent refusal for a folder that is supposed to be its own. Inside the agent's own home the folder is created on demand; anywhere else in the tree the existing refusal is unchanged.

## Acceptance criteria

- [ ] A freshly launched company's workspace contains `Agents/` and `Desks/` and **no per-agent or per-desk children**.
- [ ] A company whose manifest names no agents still gets both roots.
[...]

Issue #552, opened 2026-08-10
artifacts: a published deliverable never reaches the workspace, so no other agent can read it


oxoxDev:
## Summary

A published deliverable never reaches the shared workspace tree. `publish_artifact` (#244) reads a file out of the agent's sandbox and stages a `PendingPublish` that the brain drains into the `ArtifactStore` against the task card (`src/harness/publish.rs`). The artifact is versioned, operator-visible in the card's Artifacts tab, and reachable by exactly nobody else: there is no agent-facing artifact read tool anywhere in `src/harness/`.

So the one deliberate act that turns agent work into a deliverable puts it somewhere no other agent can read.

## Problem

The workspace is meant to be the shared resource space — the single tree every agent references and every human collaborator sees. A deliverable is the highest-value thing to have in it, and it is the one thing that never lands there.

Concretely, after a task completes:

* The operator sees the artifact on the card. Fine.
* Another agent asked to build on it has `workspace_list` / `workspace_read` and finds nothing — the artifact is in a different store, keyed by task, with no tool over it.
* The payload's only durable copy for a non-inlineable file is a **reference**: `capture_body` (`publish.rs:545-600`) records path + size + sha256 and states outright that "the file lives in the agent's own sandbox. Wiping the sandbox leaves this record intact and the payload unreachable."

The sandbox is host filesystem state (`<home>/harness/...`, `build.rs:694`) outside every store backend — not in sqlite, not in MongoDB, not replicated. On a hosted tenant a pod replace without a volume on that path leaves a receipt for bytes nobody can fetch.

## Solution

**1. Publishing dual-writes.** The drain site — which already holds the card, the responder and the store — additionally materializes the file as a workspace node at `Agents/<agent-id>/<task-id>/<name>`, in the per-agent subtree that the create-tool issue provisions.

**2. One object, seen twice.** The `ArtifactVersion` carries the node id, so the card's Artifacts tab and the Workspace tab show the same thing rather than two copies that can disagree. The tree is the durable home; the card is the task-scoped view of it.

**3. The drain stays the single write site.** No new staging path and no second queue. `PendingPublishQueue::claim` and the `PublishDestination` invariant from #445 are untouched — a publish with no claimed destination still refuses in-turn rather than reporting a success nothing honours.

**4. A published file becomes readable by every agent** through the existing `workspace_read`, with no new tool and no artifact-read surface to design. That is the point of routing through the tree instead of adding a second reader.

[...]

Issue #551, opened 2026-08-10
workspace: an agent cannot create a node, so agent output can never land in the shared tree


oxoxDev:
## Summary

An agent cannot create a node in the company workspace, so nothing an agent produces can ever land in the shared tree. `workspace_tools` ships exactly three tools — `workspace_list`, `workspace_read`, `workspace_write` (`src/harness/workspace_tools.rs:136-140`) — and the write tool requires an `expected_updated_at` compare-and-swap token that only an *existing* note has. Create, rename and delete are operator-only by design (`workspace_tools.rs:77`).

That design was right for #237's scope: the workspace was the operator's note tree, and agents were readers of it. It is wrong for the workspace as a *shared resource space*, which is what the product now wants — one place every agent writes into and every agent (and every human collaborator) reads out of.

## Problem

Today an agent's output has two possible homes, and neither is shared:

* Its **own sandbox** at `<home>/harness/<company>/<agent>/workspace` (`src/harness/build.rs:694`), confined `workspace_only`. Invisible to the console, invisible to every other agent, and living on the host filesystem outside every store backend.
* An **artifact** on a task card, via `publish_artifact` (#244). Versioned and operator-visible, but there is no agent-facing artifact read tool anywhere in `src/harness/`, so agent B cannot read what agent A produced.

So the shared tree — the one surface both humans and agents already reach — is the one surface agent output cannot enter.

Steps to reproduce:

1. Grant an agent `files` and `workspace`.
2. Ask it to write a spec into the workspace.
3. `workspace_write` refuses: there is no node to overwrite, and no tool creates one.
4. The agent writes to its sandbox instead, where nobody will ever see it.

## Solution

**1. A reserved `Agents/` subtree.** Provision `Agents/<agent-id>/` for each manifest agent at company boot and on manifest agent add, through the same rebuild seam the harness already uses for roster changes. The folder is the unit of attribution and the unit of confinement.

**2. `workspace_create`,** creating a folder or a file with initial content, resolved through the existing `CompanyWorkspace::index` path so it inherits the tenancy argument in `workspace_tools.rs`'s module docs unchanged.

[...]

PR #611, opened 2026-08-11
feat(workspace): hold bytes in the shared tree, GridFS on MongoDB (#553)

### LINKED ISSUE #553: workspace: the tree cannot hold bytes, so a paid generated image is a dangling sha256
## Summary

The workspace tree can only hold text. `WorkspaceNode` content is a `String` on the port (`src/ports/workspace.rs`), the sqlite column is `content TEXT` (`src/store/sqlite.rs:175-182`), the Mongo document field is a string, the seeder walks `.md` only (`src/company/workspace_seed.rs`), and the console upload input accepts `.md,.markdown,.txt` (`frontend/src/views/WorkspaceView.tsx:618`).

So a generated image, a PDF, an xlsx or a zip has no durable home anywhere in the product. `publish.rs:capture_body` handles this honestly and badly: over 256 KiB or non-UTF-8, the artifact becomes a **reference record** — path, byte size, sha256 — carrying the sentence "the file lives in the agent's own sandbox. Wiping the sandbox leaves this record intact and the payload unreachable."

## Problem

Agents already produce binaries today, under grants that exist:

| Producer | Grant | Emits |
|---|---|---|
| `media_generate_image` / `media_generate_video` (`toolbelt.rs:374`) | explicit `media` | PNG/JPEG, MP4 |
| `shell` + `apply_patch` native runtime | `code` | anything a script writes — pdf, xlsx, zip, sqlite |
| `web_fetch` / `curl` / `http_request` | `web` | downloaded PDFs, images, archives |
| `csv_export` (`toolbelt.rs:261`) | `code` | CSV — UTF-8, so it inlines fine today |

The media tools **spend real money**: the backend charges on submit, which is why they need an explicit `media` grant plus an approval gate. The output of that paid call lands in `<home>/harness/<company>/<agent>/workspace` — host filesystem, outside every store backend, unreplicated. The most expensive artifact the product can produce is the least durable thing it stores.

## Solution

**Additive, never touching the text path.** Seventeen files reference the workspace port; widening `read`/`write` from `String` to bytes moves the backlink scan, the seeder, the GraphQL projection, `workspace_tools` and the REST layer for no benefit — an agent cannot consume PNG bytes in context anyway.

**1. Port:** `mime`, `size` and `sha256` on `WorkspaceNode` (all `serde(default)`, so existing records load unchanged and no migration is needed for the metadata), plus `read_bytes` / `create_binary`. Text callers compile untouched.

**2. Backends, all three:**
* `FsOps` — already writes the body as a real file on disk; near-free.
* `SqliteStore` — `ALTER TABLE workspace_nodes ADD COLUMN blob BLOB` through the existing `add_column` helper (`sqlite.rs:267`).
* `MongoStore` — **GridFS**. The 16 MB BSON document cap means a generated video cannot ride in the document, so this needs a tenancy-scoped bucket and a streaming read. This is the only genuinely new-shape work in the issue and the reason it is not a small one.

**3. Reads refuse, deliberately.** `workspace_read` returns metadata (name, mime, size, sha256) for a binary node and never the bytes — the tool's whole result budget argument (#417) assumes prose, and an agent has nothing to do with a PNG's bytes. The path to the payload is the HTTP route below.

**4. Routes:** `GET …/workspace/blob/{nodeId}` streaming with the right content-type, and a multipart upload path — today's create is a JSON body, which cannot carry bytes.

**5. Console:** render an image inline, offer a download for everything else, suppress the diff view on non-prose, widen the upload `accept=`.

**6. A per-company byte quota, in the same change.** A shared tree that agents can write video into needs a cap before it needs anything else. Refuse at create/write with a clear error rather than filling the tenant's volume.

**7. `publish_artifact`'s reference branch goes away.** With the tree able to hold bytes, a non-inlineable file becomes a node like any other and the dangling-sha256 record stops being produced.

## Acceptance criteria

- [ ] A binary file can be created, fetched and deleted through the workspace port on **all three** backends, pinned by the shared conformance suite (`src/store/conformance.rs`) so no backend can answer differently.
- [ ] A file larger than the 16 MB BSON document cap round-trips byte-exact on MongoDB via GridFS.
- [ ] `GET …/workspace/blob/{nodeId}` streams the bytes with the stored content-type; the tree read still carries metadata only.
- [ ] `workspace_read` on a binary node returns metadata and never bytes; the returned sha256 matches the stored payload.
- [ ] Company A's blobs are unreachable from company B, including through the GridFS bucket — a tenancy test, not an assertion.
- [ ] A publish of a generated image produces a workspace node with real bytes, and no reference-only artifact record is created.
- [ ] Exceeding the per-company quota refuses the write with a clear error and leaves the tree consistent (no half-written node, no orphaned blob).
- [ ] The console renders an image inline, offers a download for other kinds, and does not attempt a diff on either.

## Notes and scope

**Hosted tenants run MongoDB**, so until the GridFS half lands, binary publish on hosted is exactly as broken as it is today — a paid image generation still yields a reference to a sandbox path. That is not a regression, but it is the reason the Mongo backend cannot be deferred if hosted is where this has to work.

If review needs a smaller unit, the fault line is GridFS: the port, `FsOps`, `SqliteStore`, the routes, the quota and the console are one coherent change, and the Mongo backend is a second one that lights up hosted. Everything else keeps its shape across that split.

Orphan handling deserves a decision rather than a default: deleting a node must delete its blob on every backend, and a boot-time sweep for blobs whose node vanished (a crash between the two writes) is cheap to add now and awkward to retrofit.

Scope: **large**.

## Related

- #244 — `publish_artifact`, whose reference branch this removes
- #187 — versioned artifacts; a binary version is a reference to a node, never an inline body
- #109 — the media tools whose paid output this makes durable
- #417 — the tool-result budget argument for why reads stay metadata-only

oxoxDev:
## Summary

`publish_artifact` refused any body it could not hold as a `String`, and `capture_body` wrote a record naming a sha256 with no bytes behind it. A generated image — already paid for — became a dangling digest: a receipt for a file nobody could open. The workspace tree could hold notes and nothing else.

The tree now holds bytes. `MongoStore` stores them in GridFS, so a payload past the 16 MB BSON document ceiling round-trips on the backend hosted tenants actually run.

Closes #553.

## API Or Behavior Changes

**The port gains a binary half, not a binary variant.** `WorkspaceNode` carries `mime`, `size` and `sha256`; `WorkspaceStore` gains `create_binary`, `write_binary` and `read_bytes`. Three deviations from the plan, each because the plan was wrong about this codebase:

1. **`write_binary` was not in the plan and is required.** `create_binary` + `read_bytes` can create and fetch but cannot *replace* a payload on an existing node, which is exactly what re-publishing a regenerated image does. Without it, re-publish either fails or degrades to delete-and-recreate — losing the node id that #552's artifact link depends on. It is the symmetric counterpart of `write`.
2. **sha256 is computed store-side in each backend, not in the quota decorator.** The conformance suite runs against *raw* stores, which have no decorator wrapped around them, so a decorator-computed digest would be absent precisely where the suite asserts it. A shared `blob_metadata` helper keeps the "callers never supply the digest" guarantee while surviving without the decorator.
3. **A text `write` to a binary node is refused at the store**, on all three backends with a shared message — not only at the tool layer. Otherwise prose lands on a node whose `size` and `sha256` describe bytes that are no longer there, and every reader downstream is looking at a lie.

**A published binary is stored as bytes.** The `capture_body` reference branch — the one that wrote a record pointing at a sha256 with nothing behind it — is **deleted** rather than left as a fallback. A shape-changing re-publish (markdown draft becoming a PDF) has no write that converts a node's kind, so `materialize` replaces the node; tested.

**Per-write cap: 64 MiB. There is deliberately no per-company total by default.** `tree_quota_gb` is unset out of the box, so the remaining exposure is **N writes × 64 MiB — bounded per write, unbounded in aggregate.** That is a deliberate scope decision, stated here rather than implied: no per-company accounting, no eviction, no admin policy surface was built. One chokepoint covers every writer — the console REST surface, the agent workspace tools and the publish drain all hold the *same* `ops.workspace` handle — so there is one check rather than three that can drift apart.

The quota decorator sits **inside** the announcer, so a refused write is never announced to the feed. A new `WorkspaceQuota` error maps to **413**, matching what `DefaultBodyLimit` already returns for the same cause.

**The router is stateless, so `DefaultBodyLimit` cannot be per-company.** It uses the default constant; configured values *below* it are still enforced exactly, by the store. Documented rather than left as a silent asymmetry.

**Server**: `GET …/workspace/blob/{id}` streams payloads; `POST …/workspace/upload` accepts multipart. **Console**: images render inline, other types offer metadata and a download. Two frontend findings worth recording — the API transport is text-only (`TransportResponse` exposes `text`, request bodies are `string`), so it can express neither a `Blob` nor `FormData`; `getBlob` and `postForm` are the only two methods that reach past it to `fetch`. And `ensureMdExt` would have renamed `hero.png` to `hero.png.md`, so uploads bypass it entirely.

**For #607 (workspace search): a binary node is matchable by name and is never content-scanned.** The port makes this the default rather than a rule someone has to remember — a text `read` of a binary node returns an **empty body** on all three backends, so a content scan built over `read`/`tree` finds nothing for a binary without needing to know binaries exist. `mime`/`size`/`sha256` ride along on the tree read if search wants to *describe* a payload; `read_bytes` is for downloads and should never appear in a search path. Written into `docs/spec/runtime/ports-console.md` where that work will find it.

## Tests

Re-run in full **after merging `main`** (31 commits ahead of this branch's base), not only on the original base — green alone is not green together:

- [x] `cargo clippy --locked --no-deps --all-targets --features openhuman,tinycortex -- -D warnings`
- [x] `cargo clippy --locked --no-deps --all-targets --features mongodb -- -D warnings`
- [x] `cargo build --all-targets` (via `cargo check --locked --all-targets` and `--all-features --all-targets`)
- [x] `cargo test` — **1828 passed**, 0 failed
- [x] `cargo test --locked --features sqlite --lib store::sqlite` — 26 passed
- [x] `cargo test --locked --features openhuman,tinycortex --tests` — **2728 + 10 + 1 + 11 + 2 passed**, 0 failed
- [x] `frontend`: `npm run typecheck`, `npm run build`, `npx vitest run` — 27 files, **283 tests** passed

**Local MongoDB against `mongo:7`: 23 passed, 0 failed** — including `conformance_workspace_binary_store` doing a **17 MiB round-trip, past the BSON document cap**, and `the_boot_sweep_reclaims_orphaned_blobs_and_spares_live_ones`. Counts are quoted because zero-passed is the failure mode that looks like success: the module is `test`, not `tests`, so the filter is `store::mongodb`. The `Rust (mongodb)` lane from #589 runs this in CI.

New coverage: `assert_workspace_binary_store` extends the shared conformance suite, so fs, sqlite and MongoDB are pinned to the same binary contract; plus the Mongo orphan-sweep, the text-write-to-binary refusal on each backend, shape-changing re-publish, and the quota refusal.

## Documentation

`docs/spec/runtime/storage.md` and `ports-console.md` describe the binary path, the quota knobs, and the search contract above.

Flagging rather than fixing: `storage.md` is now **570 lines** against the 500-line guideline in `AGENTS.md`. It was already 539 before this change and no CI gate enforces it, so splitting it is an unrelated diff.

## Related

Last of the sequence making the workspace a shared resource space:

- #551 + #326 — agents create in the shared tree; nodes record their author
- #570 — `Agents/` and `Desks/` scaffolded, member folders minted on first use
- #552 + #327 — deliverables land in the tree; the tab follows live
- **This PR (#553)** — the tree holds bytes, so a paid generated image stops being a dangling sha256
- #589 — the `Rust (mongodb)` lane, without which this PR's GridFS conformance assertions could not run in CI
- #607 — workspace search; its binary contract is settled above


oxoxDev:
@coderabbitai review

Issue #561, opened 2026-08-10
approvals: a parked call is refused rather than suspended, so approving costs a re-dispatch


oxoxDev:
## Symptom

Approving a parked call does not make the call happen. The agent is asked to do it again.

## Mechanism

Documented in `src/harness/policy.rs` and worth quoting because the shape is not obvious:

openhuman's `ToolPolicy` returns `RequireApproval`, and the session turn loop treats that **fail-closed** — it blocks the tool call and feeds the model a refusal, rather than suspending the call and resuming it once a verdict arrives. By the time the operator sees the card, the blocked call is gone.

So approval cannot resume anything. Instead it **re-issues**: approving mints a single-use `GrantedCall` scoped to that agent, tool and exact arguments, and `HarnessBrain` re-dispatches the granting agent with an instruction to make the call again unchanged. An unredeemed grant expires on a TTL and the operator is told the agent did not act.

Consequences the operator actually experiences:
- The turn visibly dead-ends. Work stops, output is partial, and the model may narrate a refusal it was handed.
- Approval costs a full re-dispatch — more tokens, more latency, and a model free to make different choices the second time even though the grant pins the arguments.
- A turn that parks several calls loses all of them, not just the first.
- `MAX_APPROVAL_REQUESTS_PER_TURN` is 8 and the rest of a turn's parked calls are **discarded**, so beyond eight the operator is not merely delayed — they never see the request at all.

## Why this is the deep one

The block sits in openhuman's turn loop, not here. Options, roughly in ascending cost:

1. **Reduce how often we land here.** The other sub-issues do this; it does not fix the mechanism.
2. **Batch the re-dispatch.** Approve N cards, re-issue once with all N grants live, rather than a round trip per card.
3. **Suspend and resume properly** — an await point in the turn loop, upstream in openhuman. The real fix, and the expensive one.

Worth scoping option 3 before committing to it, and worth noting that option 2 alone would remove most of the felt pain.

## Also worth fixing regardless

The silent discard past 8 requests per turn. Whatever the cap is, exceeding it should tell the operator that requests were dropped — a queue that quietly truncates reads as "nothing else needed approving", which is the opposite of true.

M3gA-Mind:
**Scoping pass — verified against `upstream/main` @ `843d5c2`.**

Confirmed: `MAX_APPROVAL_REQUESTS_PER_TURN = 8` at `src/harness/policy.rs:101`.

[...]

Issue #468, opened 2026-08-06
tasks: the card's Approvals tab shows a bare tool name and cannot decide, so it is a dead end


oxoxDev:
## What this is

A card's **Approvals** tab renders each approval as a bare tool name. It cannot approve or deny — a pending row is a link out to the Approvals page — and it shows none of what the operator needs to decide: the arguments, the requesting agent, or whether a standing scope may be granted.

## Why it is not simply "add the shared card here"

The other two surfaces that show an approval — the Approvals page and the inline chat card — render **one shared component**, which is what stops them drifting into offering different things for the same approval. This tab does not, and cannot today: its backing type carries only `id`, `kind`, `atMillis`, `status`, `resolvedAtMillis` and `waitedMillis`. There is no `payload`, no `agent`, no `broadly_grantable`.

So it omits controls by **being a different thing**, not by drifting out of sync. Making it consistent needs a projection change on the host before any console work, which is why it was correctly left alone when the shared card gained a scope control.

## Why it is worth fixing

The card is where an operator goes to understand one piece of work. Arriving at its Approvals tab to find a bare tool name and a link elsewhere is a dead end at exactly the moment they are trying to unblock something. It also means the card cannot answer *why is this stuck* — which is the question the tab exists for.

## What the fix should do

- Give the tab's projection the same fields the shared card needs, so it can render that component rather than a parallel one.
- Decide explicitly whether the tab becomes a **deciding** surface or stays read-only. If read-only, it should still show what is being asked and by whom; if deciding, it must route through the same server-side validation as the other two — the refusal is the enforcement and must not be duplicated.
- Whatever is chosen, all approval surfaces should keep rendering one component so a control added in one place appears everywhere it should.

## How to verify

- A card with a pending approval shows what is being asked, by which teammate, on the card itself.
- If deciding: approving from the tab behaves identically to approving from the page, including scope, and an approval that may not be granted broadly offers no scope control.
- No surface renders an approval through its own bespoke markup.

## Related

Surfaced while making the inline chat card offer a grant scope (#431). That work confirmed only two surfaces share the card; this is the third, and the only one that structurally cannot join them yet.

oxoxDev:
**Related — `cluster:approvals`**

Everything touching which calls stop for a human, how that decision is recorded, and where the operator makes it.

*The two holes where a call reaches execution without meeting the policy:*
[...]

Issue #395, opened 2026-08-05
approvals: a workflow run never parks an approval, so the Approvals page stays "All clear" whatever its nodes do


oxoxDev:
## Summary

A workflow run can never put anything on the Approvals page. Whatever its nodes do — a gated tool call inside an agent node, or a node explicitly marked `requires_approval` — the Approvals page stays **"All clear"** for the whole run and after it.

Found by QA on smoke1 (2026-08-04) while running a workflow they expected to be gated. Their specific repro is not by itself proof (see "What is *not* the bug" below), but investigating it surfaced two real holes on the workflow path.

Verified against `main` @ `4a38e3b`.

### Hole 1 — a gated tool call in an agent node is queued, then discarded

Parking is bolted to the brain's cycle. There is exactly one production call site:

- `src/harness/brain.rs:1298` — `self.park_approval_requests(host).await?`, inside `run_cycle`, and it needs a `&dyn CycleHost`.

The workflow path never has one:

- `src/workflows/caps/mod.rs:246` — `run_agent` calls `self.pool.run_background(&self.company, agent_ref, &message, &self.deps)`
- `src/harness/mod.rs:1041-1055` — `run_background` → `run_inner(..., None, LiveStream::Off, None)`. No `CycleHost`, so no drain.
- `src/server/ops/workflows.rs:809` — the console Run button calls `runner.run(...)` directly; no cycle is dispatched around it.

So a request raised during a workflow turn lands in the shared `ApprovalRequestQueue` and simply sits there. The next cycle then throws it away — `src/harness/brain.rs:1215-1219` clears the queue at the top of every cycle, and the doc comment on `ApprovalRequestQueue::clear` (`src/harness/policy.rs:184-185`) already names this case:

> *"Empties the queue (called before a turn so a request from a prior turn — or from a workflow run that shares these deps — never leaks into it)."*

The leak is prevented; the parking was never added.

### Hole 2 — a `requires_approval` node never reaches the Approvals page

`src/workflows/translate.rs:172` writes the flag into node config, and the engine does pause the node. But the paused node ids go only two places:

- the journal event — `src/runtime/workflow_outcome.rs:88`, `CompanyEvent::WorkflowRunFinished { pending_approvals, .. }`
- the HTTP response — `src/server/ops/workflows.rs:846`, from `WorkflowRun::pending_approvals` (`src/workflows/runner.rs:309`)

[...]

PR #612, opened 2026-08-11
workflows: consult ApprovalPolicy for tool_call nodes

### LINKED ISSUE #460: workflows: a tool_call node never consults ApprovalPolicy, so its shell and http calls raise no operator approval

M3gA-Mind:
## Summary

Closes #460. A workflow `tool_call` node now consults the company's `ApprovalPolicy`, so the call an operator would be asked about on the agent path is asked about here too — with a card that names the tool, and an approval that actually resumes the run.

**The gate is not in the invoker, and that is the design.** The obvious shape — check the policy inside `WorkflowToolInvoker::invoke` and refuse — is wrong twice over:

1. **It would be a regression, not a fix.** An agent node parks *after* a turn that already happened: openhuman resolves the refusal inline, the model narrates it, and the node still succeeds. A `tool_call` node has no turn — the call **is** the node — so a refusal fails the node, and `on_error` defaults to `stop`. Under the default `supervised` mode, `consequence.rs:198-213` puts `shell`, `http_request`, `curl`, `web_fetch`, `apply_patch`, `csv_export` and `git_operations` at `Reach::Consequence`. Refusing in the invoker would break essentially every effectful `tool_call` node on every default-mode company.
2. **The card could not be honoured.** Approving a harness-projected tool call mints a single-use `GrantedCall` redeemed by *re-dispatching the agent that asked*. There is no agent on this path, so nothing would redeem it: the grant would expire on its TTL and the operator would be told the agent did not act, about a call no agent ever made.

And the seam says so itself — `ToolInvoker::invoke(slug, args, conn)` carries no node id and no run state, so it can neither name the node on a card nor recognise an already-approved call on a continuation.

So the verdict is taken one level up, in `run_workflow_inner`, which already translates the graph **per run** and holds the record's mode and grants. A gated node is marked with the engine's own `requires_approval` flag — a **generic per-node** gate in tinyflows, not a node kind (`engine.rs:1091-1176`), which pauses before the node's work and, once the id is in the run input's `approvals` array, falls through and executes it normally. A marked `tool_call` node therefore inherits everything #395 already built, unchanged: pause → `park_pending_gates` → decidable card → `resume_from_effect` re-runs → the call executes, with the #438 delivery ledger stopping re-delivery.

A side benefit: the gate is now visible **on the canvas**, which is the other half of what #460 complains about — two node kinds in one graph, governed differently, with nothing saying which is which.

### Correction to the issue: this is not #561

#460's Related section implies suspend/resume needs #561's expensive fix. That reading is wrong, and it matters for anyone scoping the neighbouring issues. #561 is expensive because the fail-closed block lives in **openhuman's turn loop**, so approving costs a re-dispatch. The workflow path never enters that loop: a paused tinyflows run is *settled*, and #395 already shipped resume-by-re-run (`src/runtime/workflow_resume.rs`) for exactly this shape.

## API Or Behavior Changes

**Behavior.** On a company whose policy would park a call, a `tool_call` node making that call now pauses the run and raises an approval card instead of executing silently. A `full`-autonomy company with nothing on `always_approve` is byte-identical to before (pinned by a test).

**API.** `workflow_resume::gate_effect` takes a new `tool: Option<(&str, &str)>` argument, and writes two new optional payload keys, `PAYLOAD_TOOL` / `PAYLOAD_REASON`, so the card names the call and the policy's reason rather than a bare node id — the complaint #468 makes about that surface. The keys are **absent**, not null, on an authored gate, and are deliberately outside `is_same_gate`'s dedupe identity: they describe the same decision in more words.

### Deliberate deviations — stated rather than quietly satisfied

- **Standing grants cannot apply "identically on both paths"** (a criterion in the issue body). A standing grant is scoped to a teammate (`admits_scope`); a `tool_call` node has no teammate — it acts as the company. This matches grants against a synthetic `workflow:{id}` principal that nothing mints for, so the answer is always **fail-closed**: a teammate's standing grant does not open a workflow node's gate. Widening that is a consent decision, and it should be made deliberately rather than by defaulting.
- **Denial does not route through `on_error`** (also a criterion). `workflow_resume` deliberately makes a denied card a no-op — the paused run is already settled, so "nothing runs" is the complete outcome. The engine *does* support a `{"rejected": [node]}` continuation, so this is buildable; it is left matching #395 rather than diverging inside this PR.
- **A `Deny` verdict is not enforced.** Under `readonly`, `ApprovalPolicy::check` denies external effects outright. Honouring that here would be a *new refusal* on a path that runs today — the regression argued against above — and #460 is explicit that exec-security (which already applies the `readonly` autonomy tier) is not the gap. Only `RequireApproval` gates.
- **`sub_workflow` children are not gated.** tinyflows cannot resume a child across the boundary, and a paused child *halts the parent* with an unresumable error (`nodes/integration/sub_workflow.rs`). Gating a child would convert a working run into a dead one with no card to decide. **This is a real residual hole in #460** and closing it needs engine work.
- **Dry runs are not gated.** Every effect is stubbed, so there is nothing to approve, and pausing would stop a dry run walking the rest of the graph.
- **An authored `requires_approval = false` does not win.** An author may add a gate the policy does not require; they may not remove one it does.

**Exec-security is untouched.** `[policy].mode` still feeds `toolbelt::exec_security`, which sets the autonomy tier, blocks high-risk commands and confines execution to the workflow workspace. That layer was never the gap and is not loosened to make room.

### Found while doing this — not fixed here

`GuardedHttpClient::request` (`src/workflows/caps/http.rs:52-60`) calls `tool.execute(args)` with **no `ApprovalPolicy` consultation either**. So an `http_request` *node kind* reaches an arbitrary external address on a `supervised` company with no card — the identical hole, one capability over. It is out of scope for #460, whose title and table are about the `tool_call` path (a `tool_call` node whose slug is `http_request` **is** covered by this PR). Worth its own issue.

## Tests

Commands run locally, on the gated lane — the plain lane does not compile this code:

- [x] `cargo fmt --all -- --check` — run, `EXIT:0`.
- [x] `cargo clippy --all-targets -- -D warnings` — ran CI's own gated form, `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, `EXIT:0`. (Without `--no-deps` it fails inside vendored `tinyagents` before reaching this crate — pre-existing and unrelated.) It caught a real defect in this diff: `park_pending_gates` had gone to 8 arguments, now grouped into a `PausedGates` struct.
- [x] `cargo build --all-targets` — ran `cargo check --features openhuman,tinycortex --all-targets` (the lane's own command, `--all-targets` so examples are covered).
- [x] `cargo test` — ran `cargo test --features openhuman,tinycortex`: **2743 passed, 0 failed**, `EXIT:0`. NOT the bare `cargo test`: default features skip this module entirely, so a bare run proves nothing about this change. That full run predates the clippy refactor above; after it I re-ran `workflows::` + `runtime::workflow_resume` (**251 passed, 0 failed**) plus fmt and clippy, and the full suite again.

New tests:

- `src/workflows/gate.rs` — unit tests on the verdict: a consequential call gates under `supervised`; `full` leaves the graph byte-identical; a plain read and a metered read do not gate; `always_approve` gates a call the tier would allow; an authored opt-out cannot remove a policy gate; other node kinds are never touched; a slugless node is left alone.
- `every_reachable_workflow_tool_is_classified_by_name_alone` — pins the load-bearing assumption of gating at translate time. Node `args` may still carry unresolved `=`-expressions when the pass runs, so a tool whose classification depends on its arguments could be gated wrongly in either direction. None is reachable today (`WORKFLOW_TOOL_NAMESPACES` excludes `composio`); this fails the moment that stops being true, rather than letting a call slip the gate silently. Same stance as `web_search_is_still_a_priced_call`.
- `src/workflows/gated_tool_call_test.rs` — end-to-end on the real path (real graph, real `run_workflow`, real translation, real `WorkflowToolInvoker`, real exec-security, real on-disk journal; nothing stubbed — there is no model in this graph):
  - the gated node pauses, **the shell side effect does not happen**, and a card is on the Approvals page naming workflow, node, run, tool and reason;
  - **the same call executes when the policy does not gate it** — without this, every other assertion is also satisfied by a change that simply broke `tool_call` nodes;
  - the parked card produces a continuation whose `approvals` array names the node, so approving it can actually resume.

The e2e company gates under `full` + `always_approve` rather than `supervised`, deliberately: it is the stronger claim, and it keeps exec-security out of the way (`supervised` sets `require_approval_for_medium_risk`, and a test that cannot tell those two refusals apart proves nothing about this change).

**Revert-and-check.** The fix was reverted (`apply_tool_call_gates` made an unconditional no-op — the pre-#460 behaviour) and the new tests re-run. **5 of the 11 failed**, which is the intended result, and the split is worth stating rather than rounding up to "they all fail":

*Fail with the fix reverted — these pin the defect:*
- `a_consequential_call_gates_under_supervised`
- `always_approve_gates_a_call_the_tier_would_allow`
- `an_authored_opt_out_cannot_remove_a_policy_gate`
- `a_policy_gated_tool_call_node_parks_instead_of_running`
- `the_parked_card_produces_a_continuation_that_approves_the_node`

*Pass with the fix reverted, by design — these are controls, not defect tests, and each exists to catch a change that over-gates:* `full_autonomy_leaves_the_graph_untouched`, `a_plain_read_and_a_metered_read_do_not_gate`, `other_node_kinds_are_never_gated`, `a_slugless_tool_call_is_left_alone`, `the_same_call_executes_when_the_policy_does_not_gate_it`, and `every_reachable_workflow_tool_is_classified_by_name_alone` (a guard on `consequence.rs`, independent of this code). A control that failed under revert would mean it was testing the wrong thing.

All four pre-existing #395 tests also pass under the revert, confirming this change does not quietly take over their behaviour.

**Not verified here:** the full approve→continuation→execute round trip through `resume_from_effect` against a live `CompanyRuntime`. That path is #395's and is unchanged by this PR; what is new is the *card*, and the third test pins that the resume path reads it.

### Cross-reference

#338's criterion "a run that would send, publish, pay or delete stops for approval regardless of mode" is unreachable on the `tool_call` path until this lands — this PR is what makes that one honest.

## Documentation

Module docs carry the reasoning: `src/workflows/gate.rs` documents why the gate is not in the invoker, why this is not #561, and every deviation above. `gate_effect`'s doc records why the new payload keys sit outside the dedupe identity.

### DIFF
--- a/src/runtime/workflow_resume.rs
+++ b/src/runtime/workflow_resume.rs
@@ -111,7 +111,7 @@
 //! task to cancel, no connection to close.
 
 use serde::{Deserialize, Serialize};
-use serde_json::{Map, Value};
+use serde_json::{Map, Value, json};
 
 use crate::Result;
 use crate::company::load_workflow_union;
@@ -141,6 +141,14 @@ pub const PAYLOAD_INPUT: &str = "input";
 pub const PAYLOAD_DELIVERED: &str = "delivered";
 /// The payload key holding the plain-prose statement of what approving costs.
 pub const PAYLOAD_NOTE: &str = "note";
+/// The payload key holding the toolbelt slug a policy-gated `tool_call` node
+/// would run (issue #460). Absent on an authored `requires_approval` gate,
+/// which stops the run without any particular call behind it.
+pub const PAYLOAD_TOOL: &str = "tool";
+/// The payload key holding the policy's own words for why it stopped this call
+/// (issue #460) — the same sentence the agent path puts on its card. Absent for
+/// the same reason as [`PAYLOAD_TOOL`].
+pub const PAYLOAD_REASON: &str = "reason";
 
 /// The reserved trigger-input key the delivery ledger rides into a continuation
 /// run under (issue #438).
@@ -254,31 +262,51 @@ fn delivery_ledger(input: &Value, deliveries: &[DeliveryReport]) -> Vec<Delivere
 /// It is folded into the card's ledger rather than looked up later for the same
 /// reason the input is copied in: a card that needs a side table is a card that
 /// stops working after a restart.
+///
+/// `tool` is `Some((slug, reason))` when the company's `ApprovalPolicy` is what
+/// stopped this node (issue #460), and `None` for an authored
+/// `requires_approval` gate, where no particular call is being decided. It is
+/// deliberately outside the dedupe identity ([`is_same_gate`]): it describes the
+/// *same* decision in more words, so two cards that differ only here are still
+/// one question.
 pub fn gate_effect(
     workflow_id: &str,
     node_id: &str,
     input: &Value,
     run_id: &str,
     deliveries: &[DeliveryReport],
+    tool: Option<(&str, &str)>,
 ) -> Effect {
+    let mut payload = Map::new();
+    payload.insert(PAYLOAD_WORKFLOW_ID.to_string(), json!(workflow_id));
+    payload.insert(PAYLOAD_NODE_ID.to_string(), json!(node_id));
+    // The whole trigger input, so the parked card is self-contained and a
+    // resume needs nothing but the journal. This is what makes
+    // approve-after-restart work.
+    payload.insert(PAYLOAD_INPUT.to_string(), input.clone());
+    // What must NOT be sent again when this card is approved.
+    payload.insert(
+        PAYLOAD_DELIVERED.to_string(),
+        json!(delivery_ledger(input, deliveries)),
+    );
+    // What approving costs, in the operator's own terms.
+    payload.insert(PAYLOAD_NOTE.to_string(), json!(CONTINUATION_NOTE));
+    // Issue #460: which call the policy stopped, and why. The keys are ABSENT
+    // rather than null on an authored gate — a card that names no tool is a
+    // different thing from one whose tool could not be determined, and a
+    // console reading `payload.tool` should be able to tell them apart.
+    if let Some((slug, reason)) = tool {
+        payload.insert(PAYLOAD_TOOL.to_string(), json!(slug));
+        payload.insert(PAYLOAD_REASON.to_string(), json!(reason));
+    }
+
     Effect {
         kind: WORKFLOW_APPROVE_KIND.to_string(),
         group: crate::ports::types::EffectGroup::Other,
         amount_usd: None,
         established_thread: false,
         first_time_counterparty: false,
-        payload: serde_json::json!({
-            PAYLOAD_WORKFLOW_ID: workflow_id,
-            PAYLOAD_NODE_ID: node_id,
-            // The whole trigger input, so the parked card is self-contained and
-            // a resume needs nothing but the journal. This is what makes
-            // approve-after-restart work.
-            PAYLOAD_INPUT: input.clone(),
-            // What must NOT be sent again when this card is approved.
-            PAYLOAD_DELIVERED: delivery_ledger(input, deliveries),
-            // What approving costs, in the operator's own terms.
-            PAYLOAD_NOTE: CONTINUATION_NOTE,
-        }),
+        payload: Value::Object(payload),
         // Native, not a teammate's tool call — see the doc above.
         agent: None,
         // The run that paused. Not the run the approval will start (which does
@@ -529,7 +557,7 @@ mod tests {
     use super::*;
 
     fn effect(workflow: &str, node: &str, input: Value) -> Effect {
-        gate_effect(workflow, node, &input, "run-1", &[])
+        gate_effect(workflow, node, &input, "run-1", &[], None)
     }
 
     /// A delivery row with `status`, as `deliver_outputs` would have returned it.
@@ -661,6 +689,7 @@ mod tests {
                 delivery("owner_summary", "owner", DeliveryStatus::Sent),
                 delivery("cold_note", "email", DeliveryStatus::Pending),
             ],
+            None,
         );
         assert_eq!(
             ledger(&e),
@@ -693,6 +722,7 @@ mod tests {
                 &Value::Null,
                 "run-1",
                 &[delivery("summary", "owner", status)],
+                None,
             );
             assert!(
                 ledger(&e).is_empty(),
@@ -714,6 +744,7 @@ mod tests {
                 delivery("summary", "owner", DeliveryStatus::Sent),
                 delivery("summary", "owner", DeliveryStatus::Sent),
             ],
+            None,
         );
         assert_eq!(ledger(&e).len(), 1);
     }
@@ -728,6 +759,7 @@ mod tests {
             &serde_json::json!({ "request": "x" }),
             "run-1",
             &[delivery("summary", "owner", DeliveryStatus::Sent)],
+            None,
         );
 
         let input = continuation_input(&card).expect("a well-formed card continues");
@@ -760,11 +792,12 @@ mod tests {
             &serde_json::json!({ "request": "x" }),
             "run-1",
             &[delivery("summary", "owner", DeliveryStatus::Sent)],
+            None,
         );
         let continuation = continuation_input(&first).expect("continues");
 
         // Run 2 skips the summary (delivering nothing) and pauses on gate-b.
-        let second = gate_effect("digest", "gate-b", &continuation, "run-2", &[]);
+        let second = gate_effect("digest", "gate-b", &continuation, "run-2", &[], None);
         assert_eq!(
             ledger(&second),
             vec![DeliveredReport {
@@ -801,6 +834,7 @@ mod tests {
             &serde_json::json!({ "request": "x" }),
             "run-1",
             &[delivery("summary", "owner", DeliveryStatus::Sent)],
+            None,
         );
         // The same gate, re-reached by the continuation the card started: same
         // input plus the approval… minus the approval, which the gate node
@@ -810,7 +844,7 @@ mod tests {
             .as_object_mut()
             .expect("object")
             .remove("approvals");
-        let re_reached = gate_effect("digest", "gate", &continuation, "run-2", &[]);
+        let re_reached = gate_effect("digest", "gate", &continuation, "run-2", &[], None);
 
         assert!(
             is_same_gate(&paused, &re_reached),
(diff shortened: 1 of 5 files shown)


oxoxDev (commented):
The argument for putting the verdict in `run_workflow_inner` rather than the invoker is the strongest part of this, and both halves of it hold. A refusal inside `WorkflowToolInvoker::invoke` fails the node with `on_error` defaulting to `stop`, and under the default `supervised` mode that is `shell`, `http_request`, `curl`, `web_fetch`, `apply_patch`, `csv_export` and `git_operations` — essentially every effectful `tool_call` node on every default company. And the card could not be honoured anyway, since a `GrantedCall` is redeemed by re-dispatching the agent that asked and there is no agent here. `ToolInvoker::invoke(slug, args, conn)` carrying no node id and no run state settles it: the seam physically cannot name the node or recognise a continuation. Reusing the engine's generic `requires_approval` flag so a gated node inherits #395's pause → park → resume path, rather than building a second one, is the right instinct. Skipping gates on a dry run is correct and would have been easy to miss.

The correction to #460's Related section — that this is not blocked on #561, because #561 is expensive only by virtue of living in openhuman's turn loop while a paused tinyflows run is already settled — is worth more than it looks. Please make sure it lands on the issue too, not just in this PR body.

**Major — this falsifies the premise #604 is currently resting on, and neither PR mentions the other.**

#604 (`feat(policy): add the auto autonomy tier`, also yours, opened about an hour before this) documents its `auto` → `AutonomyLevel::Supervised` borrow as a *security* decision rather than a naming one, and the stated reason is this path specifically. From its `toolbelt.rs` docs:

> **except** on the workflow `tool_call` path, where no such gate is installed and this policy is the whole tier. See [`autonomy_for`].

This PR installs exactly that gate. Once both land, the sentence is false, and the argument for choosing `Supervised` over `Full` loses its stated basis — the toolbelt policy is no longer "the whole tier" there.

The practical question is what a `tool_call` node does under `auto` with both merged. #604 sets `require_approval_for_medium_risk: matches!(mode, PolicyMode::Supervised | PolicyMode::Auto)`, so the toolbelt still behaves as supervised on this path, while `apply_tool_call_gates` here separately asks `ApprovalPolicy::check`. Either that is deliberate defence-in-depth — in which case say so on both — or `auto` quietly fails to deliver its own contract ("the agent works without interrupting me") on the workflow path, which is the one place #604 chose extra strictness *because* nothing else was watching.

Whichever of the two merges second has to update the other's doc comment. That is not a rebase conflict git will show you: both files compile fine with the stale sentence in place.

**Question — is the mode snapshotted for the life of a run?**
Gates are computed at translate time from the record the run started with. `harness/policy.rs:756-761` is explicit that `readonly` is the emergency stop and that the dangerous window is exactly "a company switched to `readonly` between X and Y". A run already walking its graph carries the gate set computed before the flip. Resume re-runs and therefore re-gates, so a *paused* run is fine — but an in-flight one looks like it keeps the old verdict until it settles. If that window is bounded and acceptable, worth saying so where the gates are applied; it is the first thing someone auditing the emergency stop will ask.

**Question — what can end up in `PAYLOAD_REASON`?**
The card now carries the `reason` off `ToolPolicyDecision::RequireApproval`. If any policy path builds that string from call arguments rather than the rule that matched, argument content reaches an operator card and every surface rendering it. Worth confirming reasons are rule-shaped and tool-shaped only — this repo is careful about what escapes into operator-visible text, and a payload key is a durable place for something to leak into.

**Minor:** `src/workflows/runner.rs` is also touched by #554. Small overlap, but that PR is 205 commits behind `main` and rebasing, so it will meet this.

**Minor:** all three Rust lanes are still pending here — nothing to gate on yet.

Issue #439, opened 2026-08-05
approvals: the request queue is shared across every turn, so a concurrent turn can still take another's entry


oxoxDev:
## What happens

The queue that collects approval requests is installed **pool-wide** — every agent in the roster shares one, and so does every turn, whether it came from a chat cycle or a workflow node.

Draining it therefore has to guess which entries belong to the turn doing the draining. The current approach marks the queue's length before a turn and takes only the entries appended after that mark, which is correct as long as nothing else appends in between. It usually is. It is not guaranteed to be.

Two residual races remain, both narrowed to a turn-sized window rather than eliminated:

- A chat cycle starting inside that window clears the queue, and a workflow turn's request goes with it.
- An entry appended by another turn inside the window is taken and parked by the wrong turn, which attributes it to the wrong work.

## Why it is like this

The queue predates workflows having any way to park at all. When the only drain was the chat cycle, one queue per company was the whole design, and clearing it at the top of each turn was the correct way to stop a stale request leaking forward. Workflow parking was added on top of that shape rather than changing it.

## What the fix should do

**One queue per run**, so an entry cannot be visible to a turn that did not create it. That removes both races by construction rather than narrowing them — no marking, no window, no attribution guess.

The cost is real: the queue is installed where the roster is built, so a per-run queue means re-plumbing that installation to carry the run's identity. That is harness surgery, which is why it was deliberately deferred rather than bundled into the change that made workflow approvals work at all.

## How to verify

- A workflow turn and a chat cycle running concurrently each park exactly their own requests.
- A chat cycle starting mid-workflow-turn cannot clear the workflow turn's pending request.
- A parked approval always names the turn that actually raised it.
- The existing behaviour for a single turn at a time is unchanged.

## Related

Deferred deliberately when workflow runs first became able to park approvals; the narrowing is documented at the call site rather than left implicit.

oxoxDev:
**Related — `cluster:approvals`**

Everything touching which calls stop for a human, how that decision is recorded, and where the operator makes it.

*The two holes where a call reaches execution without meeting the policy:*
- #459 — `read_workspace_state` is classified as a pure read but shells out to `git` in a directory the agent controls, so repository config decides what runs.
- #460 — a workflow `tool_call` node never consults `ApprovalPolicy`, so its shell and http calls raise no approval at all.

*Correctness of the machinery:*
- #439 — the request queue is pool-wide, so a concurrent turn can take another turn's entry.
[...]

Issue #379, opened 2026-08-05
chat: raise an approval in the conversation that needs it, and show the work resume there


oxoxDev:
## What we want

You ask the company for something in chat. Its agent starts work, and partway through it needs your sign-off — to run a command, to spend money, to message someone outside. Today that request vanishes from the conversation: it appears on a separate Approvals page, decided in isolation, and whatever happens next happens somewhere you are not looking.

It should stay where the work is. The request appears **in the channel that produced it**, as a message in the thread. You decide it there. And the moment you approve, the agent's continuation lands in that same channel — so approving something visibly causes the next thing to happen, in the place you were already reading.

That last part is the point. Right now approving is an act of faith: a button greys out, a toast appears, and the operator has no evidence anything resumed.

## The good news — the payoff already exists and is being thrown away

`run_resolve` (`src/server/operator.rs:1271-1301`) already answers a resolution with `ChatResponse { responses }` — the re-dispatched agent turn's replies, in exactly the shape the chat surface renders. #243 made approving mint a single-use grant and re-dispatch the agent, and the resulting turn comes straight back on the response.

`ApprovalsView.decide()` discards it entirely. It reads the error path and the success toast, and drops `responses` on the floor.

So "approve and watch the work continue" needs **no new mechanism** — it needs the reply to be rendered where the operator is. That is the cheapest half of this issue and the most valuable.

## What is genuinely missing

**1. A parked approval never pushes to the console.** The event vocabulary has `ApprovalResolved` (`src/ports/types.rs:264`, projected at `src/server/operator.rs:702`) but **no `ApprovalParked`**. Parking is journal-only (`JournalRecord::ApprovalParked`, `src/runtime/journal.rs:54`), so the console only learns about a new request when its feed polls. For a request to appear inline as it happens, parking needs a projected event — the same additive shape every other variant uses.

**2. A parked effect does not record which conversation produced it.** `Effect` (`src/ports/types.rs:656`) carries `agent` (the responder, when the effect came from a harness tool call) and the task/run linkage from #242 and #333. Neither identifies the chat thread. That is not enough to place the card: a desk channel and a direct message to that desk's lead both resolve to the same agent id, so an approval raised in one would appear in the other.

The fix is a thread id recorded at park time and surfaced on `ApprovalSummary`, following the same additive pattern `agent` and `run_id` already use — absent on older journal lines, which then fall back to today's behaviour (Approvals page only).

[...]

Issue #364, opened 2026-08-04
chat: threads, reactions and transcripts are console-local — nothing survives a reload


oxoxDev:
## What happens

The chat workspace ships threads, reactions and transcripts, and **the host has no surface for any of them**. Every channel posts to the same company chat endpoint. `views/chat/README.md` is explicit about which parts are real and which are local.

So a threaded reply, an emoji reaction, and the channel a message was posted in all exist only in the browser tab that made them. Reload and they are gone; open the console elsewhere and they were never there.

## Why it matters

This is the same shape as #335 — the per-task Discussion tab that rendered a discussion with no backend behind it — which was filed as a bug and fixed. The lesson from that one applies here: **a surface that looks persistent and is not will be trusted before it is understood.** Someone will hold a conversation in a thread, or mark agreement with a reaction, and lose it.

Channels compound it. They are routable (`#/chat/strategy`, `#/chat/dm:designer`) and therefore linkable and refresh-safe as *routes* — but every one of them posts to a single company chat endpoint, so the channel is a client-side filter over one undifferentiated log rather than a real destination. Two people in two channels are in the same room and cannot tell.

## What the fix probably needs

A real channel identity on the chat write path, and storage for threads and reactions. The pieces to decide:

- **Is a channel a first-class object, or a tag on a message?** A tag is cheaper and makes "move this message" trivial; an object is what lets a channel have members, a purpose line, and an archive state.
- **Threads: a parent id on a message, or a distinct thread object?** A parent id is the smaller change and probably right, given replies already render in a side panel keyed to their parent.
- **Reactions need an actor.** They are worthless without knowing who reacted, and that means they are per-user rows, not a counter on the message.

Note that DM ids key on the teammate's **name**, not their roster id, because the starter roster mints ids from a module counter that differs between calls. That is a sound workaround for a local surface and a poor foundation for a persisted one — a rename would orphan the history. Worth fixing in the same pass.

## Acceptance

- A thread reply survives a reload and is visible to another signed-in operator.
- A reaction records who reacted, and survives a reload.
- A message posted in one channel does not appear in another.
- A DM's history is keyed to something stable, so renaming a teammate does not orphan it.
- Anything still console-local after this lands says so on screen, rather than only in a README.

## Related

- #361 — shipped the chat workspace and documented the gap
- #335 — the same failure mode on the task Discussion tab, filed and fixed

Issue #382, opened 2026-08-05
workflows: the canvas cannot show which node is executing, only which have finished


oxoxDev:
## What the operator sees

While a workflow runs, the console marks finished nodes accurately but can only
*guess* which node is executing right now. On a graph with a `switch`, both arms
briefly light up as running until one of them reports. The marker is honest
about this — its tooltip says so — but it is an approximation where an exact
answer is available in principle.

## Why it happens

The embedded engine's `RunObserver` has `on_step_finish` and `on_run_finish`,
and no `on_step_start`. #371 therefore derives "currently executing" client-side
from the graph topology the console already holds: when a node finishes
successfully, its successors are marked running.

That derivation is correct for a linear graph and over-marks a branching one
until the branch resolves. It also cannot show a node as running before its
predecessor finishes, so a long-running first node shows nothing at all for its
whole duration.

## What the fix should do

Add an `on_step_start` hook to the vendored engine's `RunObserver`, fired once
per non-trigger node before its executor runs, and carry it through as a
`WorkflowNodeStarted` company event alongside the existing
`WorkflowNodeFinished`. The console then marks exactly the nodes the engine has
actually entered, and the topology derivation goes away.

The event should carry the same structural payload as its sibling — workflow id,
run id, node id — and no node input, for the reasons in
`docs/spec/runtime/events.md`.

## Scope note

This needs a change inside the vendored engine, not just the host, which is why
it was deliberately left out of #371.

## Related

- #371 — added the per-node progress trail this extends.

---

## Testing — required, and this one is bound to the UI

**This issue has a user-facing surface, so it is verified on that surface.** A green unit or component test is not evidence that this is fixed, and does not close it.

Before this is done:

- Run the console against a live host and walk the exact path described above, by hand, in a browser.
- Exercise the **failure** this issue describes, not only the happy path — the fix is that the described behaviour is gone, so the old behaviour is what you have to try to reproduce.
- If a Playwright spec covers it, that spec must actually run and pass in a lane that builds it. A suite that is compiled out of CI proves nothing (see #475).
- State in the PR what you ran it against and what you saw. "Tested" on its own is not a result.

Backend-only changes may be verified from the backend. Anything with a screen is verified on the screen.

M3gA-Mind:
Parking this deliberately, with the investigation recorded so it does not have to be redone.

[...]

PR #624, opened 2026-08-11
approvals: resume a threaded approval in its thread, not the channel

### LINKED ISSUE #435: approvals: an approval raised inside a thread resumes in the channel, not the thread
## What happens

Ask something inside a thread, have it park an approval, approve it — and the agent's continuation appears in the channel rather than in the thread where the conversation was happening. The reply is delivered and nothing is lost, but the thread it belongs to does not contain it, so a threaded conversation silently loses its own conclusion.

## Why it happens

The conversation identity carried on a parked approval and on the grant it mints is a **channel** id. That was the correct and complete answer when it was built: approvals learned to resume in the conversation that raised them, and a channel was the finest-grained conversation that existed.

Threads are now persisted, so a message can belong to a thread within a channel. The approval records were not widened to match, so the continuation resolves to the right channel and then loses the thread.

**This is not a regression.** Nothing broke — the case simply became reachable, because until threads outlived the browser there was no thread for a continuation to be missing from.

## What the fix should do

- A parked approval remembers the message it came from, not only the channel.
- The continuation lands in the same thread as the question that raised it.
- An approval raised outside any thread behaves exactly as it does today — the channel remains the answer when there is no thread, and that path must not change.
- A remembered parent that no longer resolves degrades to the channel rather than being dropped.

## Where the change lives

The park record and the granted-call record both carry the conversation identity today; widening them is the substance of the change, and both are approval-side rather than chat-side. The chat surface already renders by parent, so it needs the value threaded to it and little else.

Worth deciding at the same time: whether the identity should be "the message that raised it" or "the thread root", since those differ once a thread is deep, and only one of them is stable if the raising message is later edited.

## How to verify

- A question asked in a thread that parks an approval has its continuation appear in that thread after approval.
- The same question asked directly in a channel behaves as it does today.
- A decline is recorded in the same place the approval was raised.
- An approval whose thread has since disappeared resolves into the channel rather than vanishing.

## Related

Surfaced while making chat threads durable — flagged rather than folded in, because the change belongs to the approval records rather than the chat surface.

M3gA-Mind:
## Summary

An approval raised inside a chat thread resumed in the **channel**, so a threaded conversation silently lost its own conclusion. The reply was delivered and nothing was dropped — it just did not appear in the thread it belonged to.

The conversation identity carried on a parked approval and on the grant it mints was a **channel id**, which was the correct and complete answer when it was built (#379): a channel was the finest-grained conversation that existed. Threads are persisted now, so the continuation resolved to the right channel and then lost the thread. Not a regression — the case simply became reachable.

### The trap, for reviewers

**A field named `thread` already exists, and it holds a channel id.** `ApprovalOrigin::thread` / `GrantedCall::origin_thread` are documented as "a desk id for a channel, a roster agent id for a direct message". Anyone grepping for `thread`, seeing hits, and concluding the plumbing exists will reach the wrong conclusion.

So this adds a **distinct** `parent: Option<EventSeq>` rather than widening `thread`. Widening was the tempting option and is the wrong one: every reader of `thread` — the approvals feed's channel filter, the continuation's `chat_id`, the grant's routing — depends on it meaning *channel*, and redefining it would have changed all of them at once with **no type change for the compiler to catch a single site**. A new field makes the change additive by construction, so "an approval raised outside a thread behaves exactly as today" is untouched rather than defended.

### Root, not the raising message (the open question in the issue)

Decided against the code, not by preference:

- `src/server/operator.rs:1207` already parents an agent's answer to *the question's own parent* — "the answer joins the thread its question was asked in, rather than opening one under the question" (#364). This is that established sibling rule applied to the continuation, not a second convention.
- The console folds a transcript **exactly one level** (`frontend/src/views/chat/model.ts:315-350`) and **drops** a grandchild — pinned by `frontend/test/unit/chat-timeline.test.ts:60` ("drops a reply whose parent is not in this channel"). Parenting to the raising message would render the continuation **nowhere** precisely when that message is itself a thread reply — the case this issue is about.
- The root is also stable under an edit of the raising message, which is what the issue asked to decide on.

### The asymmetry, which is the subtle part

Channel and thread are resolved in **one pass** (`cycle_conversation`) so they can never describe different messages — but on deliberately **different terms**:

| batch | before | after |
|---|---|---|
| two channels disagree | nothing | nothing (unchanged) |
| one channel, two threads | the channel | **the channel**, thread dropped |
| one channel, one thread | the channel | the channel **+ thread** |

Resolving the pair as a single unit — the obvious implementation, and the one I wrote first — makes row 2 ambiguous and yields *nothing*, moving an approval that lands correctly today off its conversation entirely. **A finer key must never cost a coarser answer that was already right.** This is pinned by a test that fails with the asymmetry reverted.

### Invariant preserved

`grants.rs:133-136` states `origin_thread` is deliberately **not** part of the redemption match — "the operator approved a call, not a location" — because a grant that failed to match on location would silently re-park. `origin_parent` rides on exactly those terms. This is safe by construction rather than by care: `GrantSet::consume` matches field-by-field on `(agent, tool, args)`, so a new field cannot join the predicate by accident. Pinned by a test that fails when a location is added to the predicate.

### Degradation

A remembered root that no longer resolves **degrades to the channel** rather than being dropped. Not defensive tidiness: the console drops an unresolvable-parent reply rather than rendering it flat, so a stale root would make the continuation invisible — strictly worse than the bug being fixed, since today's answer at least reaches the channel. Two ways it fails and both degrade: the root is gone (detected by comparing the returned sequence to the one asked for, since `read_from` returns `>= seq` and would otherwise hand back its *successor*), or it resolves in another channel.

## API Or Behavior Changes

**No HTTP/GraphQL API changes.** No wire type gained a field; `ApprovalSummary`, `ApprovalGql` and the SSE `approval_parked` frame are untouched.

Behaviour:

1. A continuation (and a continuation-failure notice) now carries `AgentReply.parent`, so it renders in the thread that raised the approval. The console already folds by `parentId`, so **no frontend change is needed** for the fix.
2. Everything with no thread behind it — a channel-level ask, a workflow delivery, a scheduler tick, an ambiguous batch, any pre-existing journal line — is byte-for-byte unchanged.
3. Journal format: `ApprovalParked` gains an optional `parent`. `#[serde(default)]`, so every existing line replays; a park with no root serializes as `"parent":null`.
4. Grants gain `origin_parent`, `#[serde(default)]` for replay of pre-existing lines. **Routing only, never matched.**

### Deliberately NOT in scope

The **approval card itself** still appears at channel level, not inside the thread panel. The issue asks for the *continuation* to land in the thread ("the chat surface already renders by parent, so it needs the value threaded to it and little else") — and that is true of the continuation bubble, which needs no UI work at all. Placing the card in the thread is genuinely different work: `ThreadPanel.tsx` has no approval concept, and it would mean branching the two placement filters in `ChatView.tsx` plus a new render path. Flagging rather than silently bundling or silently omitting.

## Tests

Run against the **gated** lane, since default features skip roughly a third of this code.

- [x] `cargo fmt --all -- --check` — **EXIT 0**.
- [x] `cargo clippy --all-targets -- -D warnings` — ran as `cargo clippy --features openhuman,tinycortex --all-targets -- -D warnings`. **EXIT 101, and it is not this diff**: 2 × `large size difference between variants` in `vendor/tinyagents/src/harness/agent_loop/stream.rs:48,97`. Confirmed **pre-existing** by stashing the change and re-running on a clean tree — identical failure. Nothing in `src/` is flagged.
- [x] `cargo build --all-targets` — ran as `cargo build --features openhuman,tinycortex --all-targets`. **EXIT 0**.
- [x] `cargo test` — ran as `cargo test --features openhuman,tinycortex`. **EXIT 0, 2736 + 10 + 1 + 11 + 2 passed, 0 failed.**

Exit codes captured via unpiped redirect (`cmd > log 2>&1; echo "EXIT:$?"`), not through a pipe.

### Four new tests, each revert-checked

Every one was proven to **fail with its fix reverted** — five reverts, since the degradation guard has two independent halves:

| test | reverted | result |
|---|---|---|
| `cycle_conversation_carries_the_thread_root_and_degrades_it_before_the_channel` | resolve the pair as one unit | **FAILS** — "a thread disagreement must cost the thread, never the channel" |
| `a_thread_root_survives_resolution_and_reload_and_a_pre_435_line_replays` | stop persisting the root | **FAILS** — "a thread-rooted park must say so on disk" |
| `a_grants_origin_is_routing_and_never_part_of_the_match` | add the origin to the predicate | **FAILS** — the grant no longer redeems |
| `an_unresolvable_thread_root_degrades_to_the_channel` | drop the exact-sequence check | **FAILS** — "a vanished root must not be answered by the message that follows it" |
| ″ | drop the channel check | **FAILS** — "a root in another channel must not follow the answer across" |

**One of these was vacuous first time round and is reported rather than quietly fixed.** The "root is gone" assertion originally asked for a sequence past the end of the log — that read is empty, so it passed with the check reverted and proved nothing. Sequences are dense, so the only thing that makes a real gap is pruning (which `ports/events.rs` documents as leaving gaps by design). The test now appends a prunable `WorkspaceChanged`, then a real message in the same channel, prunes the first, and asserts the hole is not answered by the message that follows it — which is the case that would otherwise slip through the channel check too. It now fails on revert, as shown above.

### Verified how

Beyond the suite: the routing decision was checked against the actual chat surface rather than assumed — the one-level fold, the grandchild-drop test, and the `#364` sibling rule in `operator.rs` are what chose the thread root.

### NOT verified

- **Not run in the live app.** No manual end-to-end drive of "ask in a thread → park → approve → see the continuation in the thread". The rendering claim rests on reading `model.ts` / `ChatView.tsx` and their unit tests, not on watching it happen.
- **No frontend tests run** — the diff touches no frontend file.
- **The `resolvable_parent` guard is unit-tested directly**, not through a full park→resolve→continue cycle; wiring a brain that produces responses through that path is a heavier harness than the guard warrants.
- Concurrency around the shared approval queue is untouched here — that is #439.

## Documentation

No `docs/` change. The contract this touches is documented **at the code**, and that is where it is updated: `ApprovalOrigin::parent` carries the full why-not-widen-`thread` and why-the-root argument, and `cycle_conversation` carries the asymmetry rule. `docs/spec/runtime/events.md` describes `ApprovalParked` as "an id, a dotted kind, and the thread it came from" — still accurate, since that event is deliberately unchanged.

Fixes #435

## Summary by CodeRabbit

- **Bug Fixes**
  - Approval conversations now preserve their original channel and thread when continuing, including grant-related replies and failure notices.
  - Invalid, removed, cross-channel, or unreadable thread references now safely fall back to the channel instead of misrouting messages.
  - Nested replies no longer appear incorrectly in folded chat timelines.
  - General chat identifiers are now recognized consistently across supported naming formats.

- **Improvements**
  - Thread context is retained across approval processing, reloads, and workflow continuation for more consistent conversations.


oxoxDev (commented):
Checked all three citations the design rests on, and each is accurate. `grants.rs:133-136` does say `origin_thread` is deliberately outside the redemption match ("the operator approved a call, not a location"). `server/operator.rs` does parent an agent's answer to `message.parent` — the question's own parent, not the question — so root-parenting here is the established sibling rule rather than a new convention. And `chat-timeline.test.ts` does pin that a reply whose parent is absent is dropped rather than promoted to the channel.

Adding a distinct `parent: Option<EventSeq>` instead of widening `thread` is right, and the reason given is the strongest kind: widening would have redefined a field that the approvals-feed channel filter, the continuation's `chat_id` and the grant's routing all read as *channel*, with no type change for the compiler to catch a single site. The warning that a field named `thread` already exists and holds a channel id is the most useful paragraph in the description — it is exactly the wrong conclusion a reviewer grepping for `thread` would reach.

The asymmetry is the right call and the argument for it generalises: a finer key must never cost a coarser answer that was already right. Resolving the pair as one unit is what most people would write, and row 2 of your table is why it's wrong.

**Major — the load-bearing console property is not the one the cited test pins.**

The case for parenting to the root rather than the raising message rests on the console folding *exactly one level* and dropping a grandchild. That behaviour does follow from `buildTimeline`'s structure — only parentless messages become entries, so a grandchild is bucketed under a reply that is never itself an entry — but the test cited at `chat-timeline.test.ts:60` pins something adjacent: a reply whose parent is **absent from the transcript**. Those are different failures with different causes. Nothing currently fails if someone adds a second fold level to the console.

That matters because the routing decision in this PR is downstream of a frontend rendering property. If the console later folds two levels — a reasonable thing for someone to do without ever opening this file — root-parenting stops being required and nothing anywhere says these two facts are coupled. Worth adding a test that pins the grandchild case directly (`a` root, `b` reply to `a`, `c` reply to `b` — assert `c` renders nowhere), and a one-line comment at the routing decision naming it as the constraint. Cheap now; otherwise this becomes an unexplained convention in six months.

**Question — is row 2 ever visible to the operator?** With one channel and two threads, the thread is dropped and the continuation lands in the channel — which is the exact symptom #435 was filed about, now reachable in a narrower case. The trade is right and I would not change it. But is there any signal, or does it look identical to the bug from the operator's seat? Even a debug-level log naming "two threads in batch, resuming in channel" would make the next report of this diagnosable rather than a re-open of #435.

**Process note — this is your seventh open PR, and they overlap heavily.**
#624 touches `harness/policy.rs` (shared with #603, #604, #606, #609), `runtime/cycle.rs` and `runtime/journal.rs` (#603, #609), `runtime/workflow_resume.rs` (#612), `server/operator.rs` (#603, #609) and `server/ops/write_test.rs` (#598, #554). Every one of them is cut from the same parent, so none of these conflicts exists yet and all of them will.

None of that is a criticism of any individual PR — the work is well argued and independently reviewable, which is the hard part. But the merge order is now a bigger risk than any single finding across the set, and #612/#604 already showed that some of the interactions are semantic rather than textual (a stale doc comment that compiles fine). A stated sequence, or a couple of these stacked deliberately, would be worth more right now than another round of review comments.

oxoxDev:
Pushed four commits onto this branch, plus a merge of `upstream/main` that turned out to be required rather than optional.

## The build failure was the merge, not a feature gate

Nothing here is `#[cfg(feature = "openhuman")]`. The branch compiles clean on its own base — I checked out `caae3a60` untouched and ran `cargo check --all-targets` at default features before changing anything: exit 0.

All three lanes were red because CI builds `refs/pull/624/merge`, and the branch was 184 commits behind. Since this branch's base, `main` grew #382's assertion inside `cycle_thread_id_reads_an_addressed_message_…`:

```rust
assert_eq!(
    cycle_thread_id(&[workflow_node_started()], approval_thread),
    None,
);
```

That is the exact pair this PR renames to `cycle_conversation` / `approval_conversation`. The two sides touch different lines, so git merged both cleanly and produced a tree calling two symbols that no longer exist — `E0425` twice on one line. The tell that it was never gating: the `openhuman` lane failed with the *same* two errors. A gating bug would have spared it.

So it had to be fixed inside the merge, not on the branch — GitHub regenerates that tree on every push otherwise. Merged `upstream/main` (not rebased). Rewrote the assertion to the surviving API rather than reverting the rename: it reads `.thread` off the pair exactly as its sibling two lines above already does.

## clippy arity

`record_parked` hit 8 args (8/7). Grouped `thread` + `parent` into the `ApprovalConversation` this PR already defines, rather than `#[allow]`.

The struct is not a lint dodge — it is the argument this PR's own docs already make. `approval_conversation` returns the pair as one value specifically so a reader cannot take one approval's channel and another's thread; the write side was pulling apart what the read side reassembles. Two adjacent same-shaped `Option`s in a call are transposable with no type error, and a park is the one place a wrong pairing gets written down durably. `PausedGates` in `src/workflows/runner.rs` is the precedent. 31 call sites moved; the two production ones (`runtime/cycle.rs`, `workflows/delivery.rs`) kept their comments, re-pointed at the pair.

## Major — the grandchild fold

Agreed on the behaviour, and you are right that the cited test does not pin it. `chat-timeline.test.ts:60` asserts a reply whose parent is **absent from the transcript**; a grandchild's parent is present. A second fold level in the console would have kept that test green while quietly making this PR's root-vs-raiser choice unnecessary.

Added `renders a grandchild nowhere: the fold is exactly one level deep` to `frontend/test/unit/chat-timeline.test.ts` — `a` root, `b`→`a`, `c`→`b`, asserting `c` is reachable from nowhere in the rendered output rather than merely missing from the entry list.

Verified it pins rather than just passes: patched `buildTimeline` to emit an entry for a reply-of-a-reply and re-ran the file — `1 failed | 7 passed`, and the one failure was the new test. The test you cited stayed green, which is the gap. Reverted the patch.

`ApprovalOrigin::parent`'s doc now names this test and its file, and says why it exists, so the constraint is reachable from the routing decision.

## Question — row 2

Left the trade alone; it is correct. Added a `tracing::debug!` on the demotion arm in `cycle_conversation`:

```
[approval] two threads in this batch, resuming in the channel; the channel is
unambiguous so only the thread root is dropped (#435)
```

with `channel`, `dropped_parent`, `rival_parent`. `tracing::` not `log::` — this call site compiles at default features and `log::` is only available under `openhuman`. Debug, not warn: the outcome is correct and is the pre-#435 behaviour, so it is an explanation on demand rather than an incident.

## Gates

| Command | Result |
|---|---|
| `cargo check --all-targets` on `caae3a60`, before any change | 0 — baseline proving the failure is merge-only |
| `cargo check --all-targets` | 0 |
| `cargo clippy --all-targets --no-deps -- -D warnings` | 0, no diagnostics |
| `cargo check --features openhuman,tinycortex --all-targets` | 0, 8m11s |
| `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` | 0 |
| `cargo test --locked --features openhuman,tinycortex --lib runtime::cycle` | 68 passed, 0 failed |
| `… --lib runtime::journal` | 31 passed, 0 failed |
| `… --lib runtime::workflow_resume` | 29 passed, 0 failed |
| `… --lib workflows::delivery` | 42 passed, 0 failed |
| `… --lib server::ops` | 353 passed, 0 failed |
| `cargo fmt --all -- --check` | 0 |
| `npm run typecheck` / `npm run build` | 0 / 0 |
| `npx vitest run test/unit/chat-timeline.test.ts` | 8 passed (was 7) |

Not run: bare `cargo test` unscoped — it OOMs on this machine, so the `--lib` runs are scoped to the touched modules and CI covers the rest.

`workflows::gate::tests::a_plain_read_and_a_metered_read_do_not_gate` did not fail here — main's fix (`2eb19f5e`) came in with the merge.

oxoxDev:
Fixed in 1e8d8c2a. CodeRabbit's finding is real, and it is **wider and more current** than the comment says — worth stating precisely, because the narrow version invites a narrow fix.

**Not legacy-only.** `chat_and_emit` defaults an unaddressed message's *desk* to `DEFAULT_DESK` ("General") but appends the event with `chat: message.chat` — i.e. `None`. Today's route still writes it. `chat_history::owns` then buckets that message into the General desk, so it renders there and a reply threads off it with `chat: Some("General")`. The comparison arrives as `None` vs `"General"`, drops the parent, and resumes in the channel. That is #435's symptom surviving inside #435's fix, on the default path.

**And `None` is not the only spelling.** `owns` has treated four as one desk since #65: `None`, `""`, the console's `"main"` (which is what the console actually sends for its default thread), and `"General"`. Normalizing only `None` would have left `"main"` → `"General"` broken — the commonest console path of the four.

So the fix is a shared rule rather than a local mapping: `chat_history::same_conversation` (built on `is_general_chat`) now answers "same conversation?" for both the history renderer and `resolvable_parent`. Two places deciding that question by different rules was the actual defect; one function is what stops it recurring. Only the General family folds — a named desk still compares verbatim, since two desks differing in case are two desks, and the test asserts that direction too.

One thing I changed beyond the report: the `_ => None` arm for a non-message event at the recorded sequence now returns early. Left as-is it produced `channel = None`, which the new folding would have read as "the General desk" — turning a guard into its opposite. Caught by the existing `an_unresolvable_thread_root_degrades_to_the_channel`, which still passes.

Both `resolvable_parent` callers are covered by construction — `publish_continuation` and `announce_continuation_failure` share the one helper, so the continuation reply and the failure notice cannot disagree about where a thread lives.

**Verified it pins, rather than merely passes.** With the fix reverted to the old raw comparison, `a_root_in_any_spelling_of_the_general_desk_still_resolves` FAILED and the pre-existing `an_unresolvable_thread_root_degrades_to_the_channel` stayed green — so the new test fails for the stated reason and the old guarantee is untouched.

Ran: `cargo test --lib company::runtime` (9 passed) · `cargo test --lib chat_history` (17 passed, includes `chat_history_finds_agent_replies_under_general_and_main`) · `cargo clippy --all-targets --no-deps -- -D warnings` (clean) · `cargo fmt --all --check` (clean) · `cargo check --features openhuman,tinycortex --all-targets` (clean).

oxoxDev:
Second finding is real too, and it was made reachable by my own previous commit — fixed in cc9f4d76.

**The scenario, confirmed end to end.** `resolve_desk` only matches against `manifest.group_chats`, and nothing is named `main`, so `?desk=main` falls through to `("main", "main")`. The console asks for exactly that: `defaultThreads()` yields `mainThread()` with id `"main"`, and `app-shell.tsx` hydrates every thread by id. With `is_general_desk` keyed on the *desk* being spelled `"General"`, that query excluded every event journaled under `"General"`.

**Which is not a hypothetical pairing.** `chat_and_emit` defaults an unaddressed post's desk to `DEFAULT_DESK` and journals the reply with `chat_id: "General"`, while the operator message keeps `chat: None`. So the console's main line was dropping *both halves* of every unaddressed conversation — the same route whose `None` I normalized in 1e8d8c2a.

**And it was actively unsafe after that commit**, which is why I am treating a pre-existing filter quirk as in scope rather than deferring it. `resolvable_parent` now folds `None`/`""`/`"main"`/`"General"` into one conversation; `owns` still did not. A continuation could therefore be parented to a root the main line refuses to render — and per this PR's own module docs, the console *drops* a reply whose parent it cannot find rather than showing it flat. One-sided folding turns a fixed bug into an invisible message.

Took the shape you proposed — both sides through `same_conversation` — with the `is_general_desk` gate removed rather than preserved, since the fold is now symmetric and the gate was the defect. Behaviour for named desks is unchanged: `same_conversation` folds only the General family, so `("strategy", "Strategy desk")` still compares verbatim and does not admit `""`, `"main"` or `"General"`.

Tests: `the_main_line_owns_what_was_journaled_under_general` covers both event variants stored as `"General"`, `"main"` and `""` under a `("main", "main")` query, plus the unaddressed `chat: None` post, plus both negative directions. **Verified it pins** — restoring the one-sided check makes it FAIL while the three pre-existing `owns` tests stay green.

Ran: `cargo test --lib chat_history` (18 passed, including `chat_history_route_reunifies_general_and_main_transcripts` and the GraphQL projection tests) · `cargo clippy --all-targets --no-deps -- -D warnings` (clean) · `cargo fmt --all --check` (clean) · `cargo check --features openhuman,tinycortex --all-targets` (clean).

PR #625, opened 2026-08-11
approvals: tell the operator when a turn's approval requests were dropped

M3gA-Mind:
## Summary

Partially addresses #561 — the **"also worth fixing regardless"** item, chat path only.
The headline fix (option 3, suspend/resume in openhuman's turn loop) is not attempted here
and is not opencompany work.

`MAX_APPROVAL_REQUESTS_PER_TURN` is 8. Everything a turn gates past that was **discarded
silently**: `ApprovalRequestQueue::drain` took the first 8, called `guard.clear()` on the
rest, and returned a `Vec` indistinguishable from a complete one. The operator saw eight
cards and had no way to learn that five more calls had been gated, refused, and dropped.
Eight cards and no notice reads as "eight is all there was", which is the opposite of
true.

Now the operator is told, with the count, on their own channel.

### What changed

- **`drain` returns `DrainedRequests { requests, discarded }`** instead of a bare `Vec`
  (`src/harness/policy.rs`). This is the load-bearing part. The overflow is now *in the
  value*, so a caller cannot drop it by not knowing there was a second number — it has to
  write `.requests` and choose to. The compiler enumerated all seven call sites when the
  type changed, which is exactly the property this issue is about: your scoping comment
  called this "the kind of split that gets half-fixed".
- **`DrainedRequests::overflow_notice(cap)`** — the operator-facing sentence, or `None`.
  It lives next to the count so the two cannot drift, and it says what the operator has to
  *do*: the calls were **not** run and are **not** on the Approvals page, so the agent has
  to be asked again. "Requests were discarded" alone invites the reading that the calls
  happened and only the records were lost.
- **`HarnessBrain::park_approval_requests` returns `Option<String>`** and `run_cycle`
  pushes it as its own operator bubble. Its own bubble rather than appended to the agent's
  reply, because the reply is the agent's answer and this is the system saying the agent
  was cut off — and a turn whose only outcome was overflow has no reply to append to.
  Still logged at `warn` as well; a log is for us, the bubble is for them.

The cap itself is untouched. Whether 8 is the right number is a separate question and this
PR does not answer it.

## The workflow half is NOT fixed here, deliberately

`MAX_APPROVAL_REQUESTS_PER_TURN` has a second consumer: **`src/workflows/caps/mod.rs:425`**,
the workflow node park path. **I did not touch it** — that file belongs to the in-flight
work on #460 and editing it would collide.

Being precise about what is left, because "half-fixed and implied whole" is the failure
mode this issue exists to name:

* The workflow path is **not** as bad as the chat path was. It already emits a
  `tracing::warn!` carrying `discarded = requests.len() - cap`, so the drop is at least in
  the logs today.
* It is **still not fixed**, because a log line is not the operator learning anything. The
  gap this PR closes for chat — the person who needed to know is told — remains open for
  workflow runs.
* The fix there is small and now has a home: `caps/mod.rs` builds its own `requests` vec
  and takes `.take(cap)`, so it can construct a `DrainedRequests` (or just call
  `overflow_notice`) and route the sentence wherever a workflow run surfaces operator
  notices. The wording helper is deliberately on the type rather than inlined at my call
  site so that path can reuse it verbatim.

Suggested follow-up: a one-line-ish change on the workflow path once #460's PR lands, so
the two paths tell the operator the same thing in the same words.

## API Or Behavior Changes

**API:** `ApprovalRequestQueue::drain` now returns `DrainedRequests` rather than
`Vec<ApprovalRequest>`. Callers take `.requests`. New public type `DrainedRequests` with
`overflow_notice`. `HarnessBrain::park_approval_requests` (private) returns
`Option<String>`.

**Behaviour:** a cycle that gates more calls than the per-turn cap now emits one extra
operator bubble naming how many were dropped and what the cap was. A cycle within the cap
is byte-identical to before.

## Tests

All exit-code gated via unpiped redirect.

- [x] `cargo clippy --all-targets -- -D warnings` — EXIT:0 (default lane). Gated lane in
      CI's exact form, `--locked --no-deps --features openhuman,tinycortex --all-targets`
      — EXIT:0.
- [x] `cargo build --all-targets` — covered by the two clippy runs, which build all
      coverage.
- [x] `cargo test` — `cargo test --lib` default lane **1834 passed / 0 failed**;
      `cargo test --features openhuman,tinycortex --lib` gated lane **2748 passed / 0
      failed**. Integration targets not re-run on this branch (library-only diff); they
      ran green on the parent, PR #603.

### Negative controls

Two, because the fix has two halves that fail in opposite directions.

**A — restore the silent discard** (`let discarded = 0;` in `drain`). EXIT:101, **2
failed**:

```
harness::brain::tests::a_turn_that_overflows_the_cap_tells_the_operator_how_many_were_dropped
harness::policy::tests::the_drain_is_capped_and_empties_the_queue
```

**B — make the notice fire when nothing was dropped** (`self.discarded >= 0`). EXIT:101,
**15 failed** — the four boundary tests below, plus **11 pre-existing brain tests** that
would break if an extra bubble appeared on an ordinary cycle:

```
harness::policy::tests::a_drain_under_the_cap_reports_no_overflow
harness::policy::tests::a_drain_exactly_at_the_cap_reports_no_overflow
harness::brain::tests::a_turn_within_the_cap_raises_no_notice
…plus operator_message_gets_an_agent_reply, no_events_still_acknowledges, and 9 more
```

Both restored; green again (2748 passed, EXIT:0).

**Coverage I am not claiming.** One new test —
`the_overflow_notice_is_singular_for_a_single_dropped_request` — fails under neither
control, because both controls target the count and the gating rather than the wording. It
is a formatting assertion and would fail if the pluralisation broke; I am flagging it
rather than implying the controls cover all six new tests.

## Documentation

`DrainedRequests` and the rewritten `drain` doc explain why the return type changed and
what the second number is for. No user-facing docs affected — though the new operator
bubble is user-visible text, and reviewers should read its wording as copy, not just as
code.

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -1972,7 +1972,9 @@ impl HarnessBrain {
             );
         }
 
-        let notice = drained.overflow_notice(cap);
+        // No `cap` argument: the drain carries the one it was taken against, so
+        // the sentence cannot name a limit this turn was not held to.
+        let notice = drained.overflow_notice();
         for request in drained.requests {
             match host.park_effect(request.effect).await {
                 Ok(approval_id) => log::info!(
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -5620,7 +5620,7 @@ members = ["eng1", "eng2"]
             None,
         )
         .with_requests(requests.clone());
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL" });
+        let args = crate::policy::test_support::composio_send_args();
         let request = ToolPolicyRequest::new(
             "composio_execute",
             args.clone(),
@@ -5856,7 +5856,14 @@ members = ["eng1", "eng2"]
         let log: Arc<dyn crate::ports::EventLog> =
             Arc::new(crate::store::FsEventLog::new(dir.path().to_path_buf()));
         let requests = crate::harness::policy::ApprovalRequestQueue::default();
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL", "to": "a@b.test" });
+        // Issue #470: a real catalogued send, with the action's own parameters
+        // under `arguments` where the tool's schema puts them — so the
+        // re-dispatch path this test covers carries a call the classifier can
+        // actually read.
+        let args = crate::policy::test_support::composio_args_with(
+            crate::policy::test_support::COMPOSIO_SEND_SLUG,
+            serde_json::json!({ "to": "a@b.test" }),
+        );
         requests
             .grants()
             .grant(crate::runtime::grants::GrantedCall {
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -1956,12 +1956,24 @@ impl HarnessBrain {
     /// already-computed operator reply along with it. That is precisely the
     /// silent-disappearance failure this issue exists to fix, so each failure is
     /// logged at `error` and the drain continues.
-    async fn park_approval_requests(&self, host: &dyn CycleHost) -> Result<()> {
-        for request in self
-            .deps
-            .approval_requests
-            .drain(crate::harness::policy::MAX_APPROVAL_REQUESTS_PER_TURN)
-        {
+    async fn park_approval_requests(&self, host: &dyn CycleHost) -> Result<Option<String>> {
+        let cap = crate::harness::policy::MAX_APPROVAL_REQUESTS_PER_TURN;
+        let drained = self.deps.approval_requests.drain(cap);
+
+        // Issue #561: the overflow used to end here, silently. The queue entries
+        // are already gone and the turn is over, so a log line is the only trace
+        // — and a log line is not something an operator reads. Kept loud for the
+        // operator's sake *and* returned, so the cycle can say it out loud.
+        if drained.discarded > 0 {
+            log::warn!(
+                "[harness::brain] {} gated tool call(s) past the per-turn cap of {cap} were \
+                 discarded and will not reach the operator",
+                drained.discarded
+            );
+        }
+
+        let notice = drained.overflow_notice(cap);
+        for request in drained.requests {
             match host.park_effect(request.effect).await {
                 Ok(approval_id) => log::info!(
                     "[harness::brain] parked '{}' for operator approval (id={approval_id}): {}",
@@ -1977,7 +1989,7 @@ impl HarnessBrain {
                 ),
             }
         }
-        Ok(())
+        Ok(notice)
     }
 
     /// Executes one drained delegation from the orchestrator's turn.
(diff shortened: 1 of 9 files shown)


oxoxDev (commented):
Making the overflow part of the return value rather than a second thing a caller has to know to ask about is the right fix, and the reason you give is the correct one: changing the type made the compiler enumerate all seven call sites. `guard.clear()` on the remainder returning a `Vec` indistinguishable from a complete one is exactly the shape of bug that survives code review forever. The wording of the notice is also right to insist the calls were *not run* and are *not on the Approvals page* — "requests were discarded" alone does invite the reading that the work happened and only the records were lost.

**Major — the one thing that can still drift is `cap`.**
The design argument for `overflow_notice` is that it "lives next to the count so the two cannot drift". But `discarded` is captured in the struct at drain time while `cap` arrives later as a parameter, so nothing stops `drain(8)` being followed by `overflow_notice(20)` — and the result is a confidently-worded, wrong sentence delivered to an operator. It is the same class of defect the PR is fixing, one level up.

Storing it closes it by construction:

```rust
pub struct DrainedRequests {
    pub requests: Vec<ApprovalRequest>,
    pub discarded: usize,
    /// The cap this was drained against — kept so the notice cannot be
    /// rendered against a different number than the one that discarded.
    cap: usize,
}

// and
pub fn overflow_notice(&self) -> Option<String> { … }
```

`drain` already has `cap` in hand, and every current caller passes the same constant to both, so this is a no-op behaviourally and removes the only remaining way to get it wrong.

**This is the third PR in an undeclared stack.**
The diff carries #603's `COMPOSIO_ACTION_KEY` / `composio_read_args` changes *and* #609's `Reach::ExternalRead`, so the real chain is #603 → #609 → #625 and the file list is identical across all three. #609 at least names #603 as its parent; this one says nothing. A reviewer arriving cold cannot tell which of the three owns `test_support.rs` or the `consequence.rs` hunks, and will re-review the same code up to three times. Worth a line at the top of each: what it is stacked on, and what to ignore.

**The deferred workflow half should be an issue, not a paragraph.**
Being explicit that `src/workflows/caps/mod.rs:425` is untouched, that it is *not* as bad as the chat path because it already logs `discarded`, and that logging still is not the operator learning anything — that is the right level of honesty and better than most PRs manage. But it lives only in this body, and once this merges the body is where things go to be forgotten. File it and link it here. The same note applies to the cap value itself: "whether 8 is the right number is a separate question" is a fine scoping call, and separate questions need somewhere to live.

**Question:** the notice goes out as its own operator bubble. Does it ride the same delivery path as any other bubble, and can it be dropped by the same mechanisms — the #438 ledger, a failed route? A notice about lost messages that can itself be lost silently would be an unfortunate shape.

oxoxDev:
Pushed to this branch (bc414ee), plus #603/#609's follow-ups cherry-picked so the stack stays consistent.

**`cap` drift is closed by construction.** `DrainedRequests` now stores the cap it was drained against as a private field; `overflow_notice()` takes no argument, `drain(cap)` fills it in, and `DrainedRequests::new(requests, discarded, cap)` is there for direct construction. `drain(8)` followed by `overflow_notice(20)` is no longer expressible. Behaviourally a no-op — every caller passed the same constant to both. Pinned by `the_notice_quotes_the_cap_the_drain_was_taken_against`, which drains 5 requests against a cap of 3 and asserts the sentence says "at most 3" and never mentions `MAX_APPROVAL_REQUESTS_PER_TURN`.

**Deferred workflow half is now an issue:** #638, assigned to you, and linked from a comment at the `park_gated_calls` discard site so it is findable from the code rather than only from this body. It carries the "is 8 the right number" question too.

**Your delivery question — it rides the ordinary path, and the one silent-drop mode is benign.** The notice is an `OutboundMessage` on channel `operator` pushed onto `channel_responses`, so it goes through `route_response` like any other bubble. An adapter that fails propagates with `?` and fails the whole cycle — loud, not silent. The unaddressed-channel fallback delivers via the operator adapter. The only quiet path is a runtime with no operator adapter at all, which `debug`-logs and drops — but the notice is still in `CycleReport.responses`, which the operator REST route reads directly, so the console shows it regardless. #438's ledger is workflow-report delivery per approval lineage and does not touch chat bubbles. Nothing to change; noting it here so the answer is on the record.

Issue #617, opened 2026-08-11
workflows: a policy-gated call inside a sub_workflow child runs silently; an authored gate there fails the parent


M3gA-Mind:
> **Corrected.** The first version of this issue said "a gated call inside a `sub_workflow` child is not gated". That is half right, and the halves need different fixes and different priorities — one is silent, the other fails loudly. Stated as one thing, it would be mis-prioritised in both directions.

### (a) A policy-gated call in a child — SILENT

This is the actual hole.

A child workflow is translated by `StoreWorkflowResolver::resolve` (`src/workflows/caps/resolver.rs:215`), which calls `translate(&file)` and **nothing else**. The per-run policy gate added by #612 / #627 runs in `run_workflow_inner` and applies to the **top-level graph only**.

The child then runs on the *same* `ctx.caps` bundle as the parent, so its `tool_call` and `http_request` nodes execute real effects with the company's grants — and produce no approval card, no recorded decision, no grant accounting. On a `supervised` company, a call the operator would have been asked about at the top level happens silently one level down.

### (b) An authored `requires_approval` node in a child — LOUD

The child pauses. `sub_workflow.rs` turns that into an error rather than a pause:

```rust
if !outcome.pending_approvals.is_empty() {
    return Err(EngineError::Capability(format!(
        "sub_workflow node {:?}: child run paused awaiting approval at {:?}; \
         cross-boundary approval resume is not yet supported, so the parent run is \
         halted rather than falsely completed",
        ...
    )));
}
```

The parent run fails with an explicit message. The feature is unusable, but nothing happens silently and nobody is misled about what ran. **Pre-existing since #395** — it is not a regression from the policy-gate work.

## #612 and #627 widen neither

Worth stating for anyone reviewing those PRs and wondering. The gate pass they add is applied in `run_workflow_inner` to the translated top-level graph; the resolver path is untouched. So (a) is exactly as it was before them, and (b) is exactly as #395 left it.

## Why the engine cannot resume across the boundary

`run_sub_workflow` (`engine.rs:813-840`) runs the child as a **separate engine invocation**: its own `InMemoryCheckpointer` (dropped when the call returns, so the child's interrupt state does not survive), a `NoopObserver`, and a fresh `CancellationToken::new()`.

[...]

Issue #596, opened 2026-08-11
workflows: a run keeps no per-node input/output, so you cannot see what a node produced or review it before it publishes


oxoxDev:
## What's missing

Like make.com's scenario history — where every module shows the bundle that went **in** and the bundle that came **out** — a workflow run should let you open any node and see its **input** and its **output**. Today it shows neither.

Confirmed on a live run (marketing staging, "Content Workflow": Start → Fetch Trending Posts → Creative Director → Draft Writer):

- The run record persists only `{nodeId, status, elapsedMs}` per node (verified via `/workflows/runs?workflow=…`). **No input, no output.**
- The run finished, all three nodes `ok`, `deliveries: []` ("routed no reports").
- Opening the Draft Writer node in the run inspector shows only its **config** (type `agent`, assigned `copywriter`) — **not the draft it produced**. The copywriter ran 13.3s and generated text, but that text is unreachable: it was never persisted and never surfaced. It only "exists" if the node's own tool call saved it to a third party (here, Typefully) — which is exactly the step we cannot verify.

## Why the previous fixes don't cover it

- **#528** (closed) made a *live* run's output readable via `RunResultPanel`, but that panel is gated on the in-memory `result` of the run you just launched; a **past** run reopened from History has no persisted output to show.
- **#418** established the root cause on the agent side: node output lives only in-memory and the journal deliberately carries none — so nothing durable exists for the console to read either.
- **#573 / #575** cover workspace-publish and journal/truncation durability, not a per-node run inspector.

## What to build

1. **Persist per-node input + output on the run record.** Each node's resolved input (what it received) and its produced output (agent text, tool-call result, HTTP body, parsed object), bounded/summarized like any other stored payload, keyed by run_id + node_id. This is the data layer the whole feature stands on.
2. **Render it in the run inspector (make.com-style).** Open a node in a past or live run → two panels: **Input** and **Output**, with the raw value expandable and large values paged (reuse the #418 read surface where it fits). Works for every node kind, for any run in History — not just the one just launched.
3. **A hold-before-publish review step.** One step before any node publishes to a third-party platform (Typefully, Composio-connected accounts, outbound HTTP), stage the produced content for operator review — show the **actual output** that is about to leave, and require sign-off on the content, not a blind approval card. This is what "review before it publishes" means, and it composes with the approvals epic (#558) but is content-review-specific: you see the draft before it ships, and can accept/edit/cancel.

## Why it matters

[...]

PR #627, opened 2026-08-11
workflows: gate http_request nodes on ApprovalPolicy too

### LINKED ISSUE #614: workflows: an http_request node never consults ApprovalPolicy, so its requests raise no operator approval
## What happens

An `http_request` **node** issues its request without ever consulting `ApprovalPolicy`. So a workflow node POSTing to an arbitrary external address on a company in the default `supervised` mode produces **no approval card**, no recorded decision, and no standing-grant accounting.

`src/workflows/caps/http.rs:52-60`:

```rust
async fn request(&self, request: Value, _conn: Option<&str>) -> TfResult<Value> {
    let args = to_tool_args(&request);
    let result = self
        .tool
        .execute(args)
        .await
        ...
}
```

`GuardedHttpClient` is built with the same exec-security policy the toolbelt gets, and then calls `execute()` directly. Nothing between the node and the wire asks whether an operator should be told.

## Correction: this is NOT "ungated arbitrary HTTP"

The exposure is narrower than the shape suggests, and getting this wrong in either direction is the reason #460 needed a correction of its own. Four things **do** still hold on this path:

| Layer | Status |
|---|---|
| SSRF `url_guard` on every request **and every redirect** — private / loopback / link-local / metadata IPs always rejected | **enforced** |
| `[tools].web_allowed_domains` allowlist | **enforced** (but see below) |
| Response-size and timeout caps, from OpenHuman's own config defaults | **enforced** |
| `readonly` autonomy — `can_act()` blocks the call outright (`http_request.rs:379`) | **enforced** |
| `ApprovalPolicy` — parks for an operator, produces a card, records the decision, honours grants | **absent** |

Two qualifications that keep the residue real rather than theoretical:

- `web_allowed_domains` **defaults to empty**, which is open-public mode (`src/company/types.rs:554`). So on a default company the allowlist constrains nothing beyond the SSRF guard — any public address is reachable.
- The gap is specifically `supervised` (the default tier) and `always_approve` under any tier. Under `full` no card is *expected*; under `readonly` the call is blocked before it starts.

So the accurate statement is: **a class of call an operator would be asked about on the agent path leaves the process silently on the `http_request` node path.** Not "arbitrary unauthenticated egress".

## Why #460 does not cover this

#612 closed #460 by having the company's `ApprovalPolicy` mark a `tool_call` node with the engine's `requires_approval` flag before the run reaches it. That covers a `tool_call` node whose slug happens to be `http_request` / `curl` / `web_fetch` — those go through `WorkflowToolInvoker`.

An `http_request` **node kind** is a different capability on a different seam: `Capabilities.http` → `GuardedHttpClient`, never `ToolInvoker`. #460's title and table are about the `tool_call` path, so this was deliberately left out of that PR rather than smuggled in.

The classification already exists and already says the right thing — `src/policy/consequence.rs:211` puts `http_request` at `Reach::Consequence`, i.e. `parks_under_supervision()`. Nothing on this path asks it.

## Related: the upstream declaration is also unread here

`HttpRequestTool` itself declares the call gate-worthy — `external_effect_with_args` returns true whenever `gate_decision(Network) == Prompt`, which is both `supervised` and `full` (`http_request.rs:358-363`). Its own doc says supervised/full tiers "route through ApprovalGate", and `tools/README.md:119` calls `external_effect_with_args` "the hook the harness checks at the gate-decision point".

OpenCompany does not consult that hook anywhere — it substitutes its own `consequence_of` declaration, which is a deliberate and coherent choice. Worth knowing while fixing this: there are two independent declarations that this call needs a human, and the workflow HTTP path reads neither.

## What the fix should do

- An `http_request` node consults `ApprovalPolicy` the way a `tool_call` node does after #612, so the same call raises the same card whichever node kind makes it.
- Reuse #612's mechanism rather than inventing a second one: take the verdict where the graph is translated per run, mark the node with the engine's generic `requires_approval` flag, and inherit #395's pause → park → resume path. A second parking path is the thing #460 argued against.
- The card should name the method and host, not just the node id (#468).
- Exec-security stays exactly as it is. It is not the gap.

## How to verify

- An `http_request` node POSTing to a public URL on a `supervised` company raises an approval card naming the workflow, node and target; approving it resumes the run and the request goes out; the request does **not** go out beforehand.
- The same node on a `full` company is unchanged (no card, request goes out).
- On a `readonly` company the behaviour is unchanged — still blocked by `can_act()`, with no card, because there is no decision to offer.
- A `tool_call` node with slug `http_request` keeps behaving exactly as #612 left it — the two paths agree.

## Related

- #460 / #612 — the same hole on the `tool_call` path, and the mechanism to reuse.
- #468 — the card's Approvals tab shows a bare tool name and cannot decide.
- #338 — its "a run that would send, publish, pay or delete stops for approval regardless of mode" is unreachable on this path until this lands, exactly as it was on the `tool_call` path before #612.

M3gA-Mind:
> **Depends on #612.** This branch is stacked on #612's, so **the diff below shows both changes** until that one merges. Merge order is #612, then this. My change is exactly one commit — `47c6a22 workflows: gate http_request nodes on ApprovalPolicy too` — and `git diff afe2747..47c6a22` isolates it from #612's `afe2747`.

## Summary

Closes #614. An `http_request` node issued its request without ever consulting `ApprovalPolicy`: no card, no recorded decision, no standing-grant accounting on a company in the default `supervised` mode.

It is the same hole #460 closed for `tool_call`, on a different seam. `GuardedHttpClient::request` calls `tool.execute(args)` directly and never goes through `ToolInvoker` (`src/workflows/caps/http.rs:52-60`), so #612's fix did not reach it — a `tool_call` node whose *slug* is `http_request` was already covered; the `http_request` **node kind** was not.

The fix extends #612's existing per-run gate pass to `NodeKind::HttpRequest` rather than adding a second mechanism. The call is classified under the same `http_request` name that `consequence_of` already declares `Reach::Consequence` (`src/policy/consequence.rs:211`), so both node kinds now answer to one declaration and cannot drift. Everything downstream is unchanged: the engine's generic `requires_approval` gate pauses the node, `park_pending_gates` parks a decidable card, and `resume_from_effect` re-runs with the approval.

## API Or Behavior Changes

**Behavior.** On a company whose policy would park the call, an `http_request` node now pauses the run and raises a card instead of requesting silently. Under `full` the node is untouched (pinned by a test); under `readonly` nothing changes — the call was already blocked before it started.

**API.** `gate_effect`'s `tool: Option<(&str, &str)>` becomes `call: Option<GateCall<'_>>` (a struct, so the third field did not make it a 3-tuple), and a new optional payload key `PAYLOAD_TARGET` carries what the call would reach. `GatedToolCall` is renamed `GatedCall` and `apply_tool_call_gates` to `apply_policy_gates`, because neither is `tool_call`-only any more — a name that lies is worse than a rename in the diff.

**The target is method and host only — never the path or query.** A query string is a routine place for tokens and signed parameters to sit, and this string is written to the durable journal, rendered on the Approvals page and kept after the decision. The host is what the decision turns on; the full URL is in the node's own config for anyone who needs it. An unresolved `=`-expression URL yields no target at all rather than claiming `=item.endpoint` is a host — the node still gates, the card just names no destination.

### Scope: what this does NOT claim

#614 was written carefully and the PR should match it. Four layers held on this path before this change and still do — the SSRF `url_guard` on every request **and every redirect**, the `web_allowed_domains` allowlist, the size/timeout caps, and `readonly` blocking the call outright via `can_act()` (`http_request.rs:379`). So this closes "a call an operator would be asked about on the agent path leaves the process silently under `supervised`". It does not close "ungated arbitrary HTTP", because that was not true.

Two qualifications keep the narrowing from becoming a whitewash: `web_allowed_domains` **defaults to empty**, which is open-public mode (`src/company/types.rs:554`), and OpenCompany's own declaration already said this call is `Reach::Consequence` — the classification existed, nothing on this path asked it.

Every deviation #612 documented still applies unchanged (fail-closed standing grants against a synthetic `workflow:{id}` principal; denial matching #395; `Deny` verdicts not enforced; dry runs not gated; authored opt-outs not honoured). `sub_workflow` children remain ungated for the engine reason in #617.

## Tests

Gated lane throughout — the plain lane does not compile this code. All exit codes captured with unpiped redirects.

- [x] `cargo clippy --all-targets -- -D warnings` — ran CI's own form, `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, `EXIT:0`.
- [x] `cargo build --all-targets` — covered by the clippy `--all-targets` run above on the same feature set.
- [x] `cargo test` — ran `cargo test --features openhuman,tinycortex`: **2749 passed, 0 failed**, `EXIT:0`. Not the bare `cargo test`: default features skip this module entirely.

New tests (6):

- `an_http_request_node_gates_under_supervised` — the defect: the node gates, and the card names `POST api.example.com` for a URL whose query carries a secret, asserting the secret is **not** in the card.
- `an_http_request_node_does_not_gate_under_full` — control: graph byte-identical under `full`.
- `an_unresolved_url_gates_without_naming_a_target` — a `=`-expression URL still gates, with no target claimed.
- `a_target_defaults_to_get_and_drops_path_query_and_fragment` — the helper directly, including the no-scheme and empty-host cases.
- `a_policy_gated_http_request_node_parks_instead_of_requesting` — end-to-end on the real path: the node pauses, and the journal card carries node, tool and target.
- `an_ungated_http_node_reaches_the_capability` — the control that keeps the rest from being hollow. It cannot be proved by a *successful* request, because `url_guard` rejects loopback unconditionally and that guard is one of the layers #614 is careful not to claim is missing. So the proof is the **shape of the failure**: with the policy not gating, the run reaches the node and fails with the guard's own refusal, which only happens if the request was attempted. A change that broke `http_request` nodes would fail differently; a change that gated them under `full` would pause instead of failing.

`other_node_kinds_are_never_gated` from #612 asserted that an `http_request` node is left alone — which is precisely what #614 says was wrong. It is split into `an_agent_node_is_never_gated_here` (still true, and still the right guard against double-parking an agent node's call) and the new expectation above, with the inversion recorded in the test's own doc comment rather than silently changed.

**Revert-and-check**, reverting only the #614 half (`NodeKind::HttpRequest => None`) so it isolates this PR from #612: **3 of the 6 failed** — `an_http_request_node_gates_under_supervised`, `an_unresolved_url_gates_without_naming_a_target`, `a_policy_gated_http_request_node_parks_instead_of_requesting`. The two controls passed, as controls must. The sixth, `a_target_defaults_to_get_and_drops_path_query_and_fragment`, also passed — honestly, because a wiring-only revert leaves `http_target` in place; it is a unit test of that helper, not of the wiring. All of #612's tests and all four of #395's also passed under the revert, confirming this change is isolated to the node kind it targets.

**Not verified here:** the approve → continuation → request round trip against a live `CompanyRuntime`, for the same reason as #612 — that path is #395's and unchanged.

## Documentation

`src/workflows/gate.rs`'s module docs now cover both node kinds, including the precise statement of what still protected the HTTP path before this change, and why the fix is not in the capability. `GatedCall::target` documents the path-and-query exclusion where the field is defined.

## Summary by CodeRabbit

- **New Features**
  - Workflow tool and HTTP requests can now be automatically paused when company approval policies require review.
  - Approval cards display the affected tool or request, the reason approval is needed, and a sanitized target when applicable.
  - Approved workflows resume from the exact paused step.
  - Policy gates apply consistently even when authored workflow settings would otherwise skip approval.

- **Bug Fixes**
  - Dry runs and previously supported workflow behaviors remain unchanged.


oxoxDev (commented):
Declaring the stack at the top — dependency, merge order, and the exact commit range that isolates your change — is what the other stacked PRs in this batch are missing. Thank you for that; it made this reviewable in a fraction of the time.

Extending #612's existing pass to a second node kind instead of adding a parallel mechanism is right, and classifying under the same `http_request` name that `consequence.rs:211` already declares `Reach::Consequence` means the two node kinds cannot drift apart. Renaming `GatedToolCall` → `GatedCall` and `apply_tool_call_gates` → `apply_policy_gates` in the same commit is the correct call — a name that lies costs more than a rename in the diff.

**Method and host only, never path or query, is the right decision and the reasoning generalises.** A query string is exactly where tokens and signed parameters live, and this value is journaled durably, rendered on the Approvals page and kept after the decision. Worth applying the same test to `PAYLOAD_REASON` from #612: it comes off `ToolPolicyDecision::RequireApproval { reason }` and lands in the same durable, operator-visible place. If any policy path ever builds that string from arguments rather than the rule that matched, it has the property you just deliberately denied `target`.

**Major — `_ => None` means the next effectful node kind is silently ungated.**

```rust
match node.kind {
    NodeKind::ToolCall => …,
    NodeKind::HttpRequest => …,
    _ => None,
}
```

Two PRs have now each closed one hole in this match, and nothing makes the third one announce itself. `NodeKind` has twelve variants; someone adding a thirteenth that reaches outward gets no card, no recorded decision and no test failure — the precise "half-fixed and implied whole" failure mode you name in #625's description.

Enumerate them instead, so the compiler asks the question:

```rust
match node.kind {
    NodeKind::ToolCall => …,
    NodeKind::HttpRequest => …,
    // Reach nothing outward on their own: the decision belongs to the node
    // they feed, or to the agent path's own gate.
    NodeKind::Agent
    | NodeKind::Condition
    | NodeKind::Merge
    | NodeKind::Output
    | NodeKind::OutputParser
    | NodeKind::SplitOut
    | NodeKind::Switch
    | NodeKind::Transform
    | NodeKind::Trigger
    | NodeKind::SubWorkflow => None,
}
```

Costs one list today and converts every future addition into a deliberate decision.

**Question — what happens to a gated node inside a `SubWorkflow`?**
`apply_policy_gates` runs over the graph `translate` produced for *this* run, and `SubWorkflow` currently lands in the wildcard arm. I could find no `SubWorkflow` handling in `runner.rs` or `workflows/mod.rs`, which suggests the engine expands or executes it internally — in which case a `tool_call` or `http_request` node *inside* a sub-workflow may never pass through this gate at all. If so, a policy that parks a call in a top-level graph lets the identical call through once it is one level down, and composition becomes the bypass. If the engine flattens sub-workflows before this pass runs then it is already covered and worth a sentence saying so; if it does not, that is a bigger hole than the one this PR closes.

**Minor:** an unresolved `=`-expression URL yields no target. Does the card distinguish "target unknown until runtime" from "no target", or does the field simply not render? An operator approving an outbound request whose destination is not shown should be told that is what they are doing.

oxoxDev:
Pushed the review fixes to this branch (89fa76a5), plus a merge of current `main` (b878b871) — #612 has landed so this no longer stacks on it.

**Userinfo (CodeRabbit's Major) — fixed.** `http_target` took the whole authority, so `https://token:secret@api.example.com/x` put the credential in the durable journal and on the Approvals page, and named the wrong thing first to anyone reading the card. Now only what follows the **last** `@` is kept — a host cannot contain one, and the last-`@` rule survives a `@` inside the credential itself. `a_target_never_carries_url_userinfo` covers `token:secret@`, bare `user@`, and `user:p@ss@host:8443`, asserting no `@` and no credential substring survives.

**`_ => None` (your Major) — fixed, and the count was wrong in my review.** `NodeKind` has **fifteen** variants, not twelve: I missed `Code`, `Memory` and `Dedup`. All are now listed with the reason each is or is not classified, so the sixteenth is a compiler error. Two of them are worth naming explicitly: `Code` and `Memory` are effectful and only unclassified because `CodeRunner` errors and `MemoryProvider` is `None` today — the comment says the decision belongs with whoever wires the capability, in that PR.

**SubWorkflow (my question) — answered, and I was wrong to ask it as if it were open.** The child is resolved and run inside the engine (`run_sub_workflow`), so its nodes never reach `call_of` — the arm is not what excludes them. Your module docs already say this and already name **#617** as the tracking issue. I filed a duplicate (#649) before spotting that and have closed it; the code comment now points at #617. The capability-level grant check does still apply to a child's calls, so this is a missing card, not a capability escape — worth keeping that distinction sharp given how the docs phrase it.

**Minor (unresolved URL) — changed behaviour, so flagging it.** `=item.endpoint` previously yielded no target; it now yields `"GET (destination resolved at run time)"`. An operator approving an outbound request with no host shown should be told that is what they are doing, rather than getting a card that looks like a `tool_call`'s. A **missing** `url` key still yields `None` — nothing was authored, so there is nothing to explain. `an_unresolved_url_gates_and_says_the_destination_is_not_known_yet` pins it, including that the expression itself never appears. If you would rather keep `None`, say so and I will revert that half.

**Merge conflict**, for the record: one hunk in `park_pending_gates`. This branch renamed `GatedToolCall` → `GatedCall` and added `target`; main's #596 added `output`/`edges` to `PausedGates` and made the effect `mut` for `attach_upstream_content`. Both kept.

Ran on the merge commit: `cargo check --all-targets` (default features), `cargo clippy --features openhuman,tinycortex --all-targets --no-deps -D warnings`, `cargo fmt --all --check`, and `cargo test --features openhuman,tinycortex --lib workflows::gate` → 22 passed. Submodule pins re-synced after the merge.

One caveat on CI: `main` is currently red on `workflows::gate::tests::a_plain_read_and_a_metered_read_do_not_gate` — an unrelated #612/#606 collision that #644 is fixing — so this branch inherits that failure until 644 lands. It is not from anything here.

Issue #618, opened 2026-08-11
Question: is company membership the intended boundary for approval payloads on GET …/approvals?


M3gA-Mind:
## The question

`GET /api/v1/companies/{id}/approvals` returns each parked approval's **payload, recipient-bearing arguments, dollar amount, asking agent and thread** to **any authenticated member of that company**, with no role check. Admin and Member are treated identically.

**Is membership the intended boundary for payload-bearing approval data?**

I am asking, not asserting a vulnerability. There is a good case that this is deliberate — see "Why it may well be intended" — and if so the useful outcome is a line of documentation, not a patch. I am not proposing to act on it either way.

Surfaced by @oxoxDev while resolving the exposure question on #468 ("*That is fine if intended*"); filing it here because it is outside that issue's scope and would otherwise evaporate when #468 lands.

## What I verified

Traced the full request path on `main` (`32d28b55`), rather than grepping. All of it holds:

**1. The projection carries the sensitive fields.** `ApprovalSummary` (`src/runtime/types.rs:113+`) has `id`, `kind`, `amount_usd`, `at_millis`, `task`, `agent`, `payload`, `thread`, `broadly_grantable`. Built at a single point, `CompanyRuntime::pending_approvals` (`src/company/runtime.rs:1525`), which applies **no per-caller filtering** — there is no notion of who is asking at projection time.

**2. The route gates on membership alone.** `list_approvals` (`src/server/operator.rs:1677`) → `CompanyAuth` → `resolve_principal` (`src/server/graphql/auth.rs:165`), which returns `GqlAuth::User` for a session → `authorize_address` (`src/server/platform_auth.rs:496-499`):

```rust
GqlAuth::User(user) => {
    if user.company == *id { None } else { Some(forbidden()) }
}
```

`user.company == *id`. Nothing else. `UserPrincipal` carries `role`, and it is not consulted here.

**3. Nothing else in the path supplies a check.** The only `middleware::from_fn` in the server is CORS (`src/server/routes.rs:133`). There is no `route_layer`, no route-level guard, and no projection-side filter. I looked for one specifically rather than concluding from absence of grep hits.

**4. A role gate exists and has precedent.** `UserRole::{Admin, Member}` with `may_administer()` (`src/ports/users.rs:42`). It is enforced in exactly one place: `src/server/users/admin.rs:98`, for user administration.

[...]

PR #629, opened 2026-08-11
fix(tasks): remove the card's Approvals tab, keep the waiting signal (#468)

### LINKED ISSUE #468: tasks: the card's Approvals tab shows a bare tool name and cannot decide, so it is a dead end

M3gA-Mind:
## Summary

Closes #468.

**The issue body is superseded.** @oxoxDev's decision comment (2026-08-11 09:02) answered the exposure question in a way that dissolved the issue rather than unblocking it, and a body rewrite was promised but has not landed — the body still describes widening the projection so the tab can render the shared card. I worked from the comment. If the body is rewritten before this merges, it should read as below.

What changed:

1. **Removed the Approvals tab** from the task detail card.
2. **Kept the signal** — one line, present only when something is actually pending, saying how many approvals the card is stopped behind and for how long, linking to the Approvals page.
3. **`TaskApproval` shrank** from six fields to three (`id`, `atMillis`, `status`), rather than growing.
4. The **"all approval surfaces render one shared component" goal is closed as not pursued** — there is one approval surface now, so there is nothing to share.

## Problem

The tab rendered each approval as a bare tool name. It could not approve or deny — every pending row was a link out to the Approvals page — and it showed none of what a decision needs. The original plan was to widen the projection so it could render the shared card. The decision instead was that approvals do not belong on the task card at all: they are resolved in exactly one place, and a second half-surface beside that one is a maintenance cost whose best case is still a link.

The thing that could **not** be removed with it is the *signal*. The card is where someone goes to ask "why is this stuck". Deleting the tab and saying nothing would trade one bad surface for a worse silence.

### The one line

`AwaitingApprovalRow` sits above the tab strip, so it is visible whichever tab is open — the point is that a stalled card announces it, not that you can go find it. It renders **nothing** when nothing is pending; an always-present "no approvals" line is exactly the clutter the tab was made of.

The derivation is in `frontend/src/lib/task-approvals.ts` rather than inline, for the reason the other `lib/` derivations are: every way it can be wrong produces a card that looks completely normal while misreporting. Three decisions, each pinned by a test:

- **Only `pending` counts.** The rest are resolutions; they are already on the timeline with their own waited span, and counting them here rebuilds the tab one row at a time.
- **Measured from the oldest park, not the newest.** With the newest, a second effect parking behind the first resets a clock that should be climbing, and the card reads as freshly blocked no matter how long it has really sat.
- **`waited` clamps at zero.** `atMillis` is the host's clock, `now` is the browser's; ordinary skew must not render "waiting for -3s".

### The projection shrink, and the one thing I deliberately kept

Six fields to three. `kind`, `resolvedAtMillis` and `waitedMillis` went with the tab. `id` stays: it is the discriminator that makes "which approvals belong to which card" testable, which is #333's behaviour and outlives the tab — three existing tests identify rows by it.

**The park→resolve arithmetic is unchanged and still asserted.** Two tests existed to pin it (the run-window clamping, and the pre-#333 legacy row). Deleting `waitedMillis` would have deleted their assertion point and quietly retired the only coverage of that arithmetic while looking like a tidy-up. Instead both now assert on the `approval` **timeline entry**, which carries its own `waitedMillis` and always did. Same numbers, same clamping, different observable.

### A coupling worth making visible now

The status line reads `detail.approvals` from the **task detail** route, not from `GET …/approvals`. Both are gated on company membership alone. I filed #618 asking whether membership is the intended boundary for the payload-bearing approvals route — and if that route were ever role-gated, **this line would keep working while the "Review" link it points at would show a Member an empty page.** Not designing around a gate that does not exist; recording the coupling so it cannot be found later by breaking it.

## API Or Behavior Changes

- **Removed**: the Approvals tab on the task detail card, and its `ApprovalsTab` component and `APPROVAL_STATUS` map.
- **Added**: a "Waiting on N approvals for X" row above the tab strip, shown only while something is pending, linking to `#/approvals`.
- **Wire change**: `GET /tasks/{task_id}` — each entry of `approvals[]` now carries `id`, `atMillis`, `status` only. `kind`, `resolvedAtMillis` and `waitedMillis` are gone. The console is the only consumer and is updated in the same commit; the `approval` timeline entry is unchanged.
- Nothing about how approvals are *decided* changes. This PR touches no approval resolution path.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — run as CI runs this lane: `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, exit 0. (`--no-deps` matters: without it the command fails on two pre-existing `large_enum_variant` lints in `vendor/tinyagents`. CI passes it — `ci.yml:434`.)
- [x] `cargo build --all-targets` — **not run as `build`**; ran `cargo check --features openhuman,tinycortex --all-targets`, exit 0. Gated lane, covers `examples/`, which default features skip.
- [x] `cargo test` — **not run bare**; ran `cargo test --features openhuman,tinycortex`: **2705 passed, 0 failed**. Default features skip about a third of the crate.

Console, all four commands CI's `Console` job runs:

- `npm run typecheck`, `npm run typecheck:e2e`, `npm run typecheck:unit` — all exit 0.
- `npm test` — 8 new tests pass. **The suite has 38 pre-existing failures** in `tour-resume`, `desktop-bridge` and `connection-registry`, all `window.localStorage.clear is not a function`. Confirmed pre-existing by stashing this diff and re-running: baseline **38 failed / 237 passed**, with this diff **38 failed / 245 passed** — same failures, my 8 on top. It looks like a local jsdom/environment difference; I am making no claim about CI, which is authoritative.

New: `frontend/test/unit/task-approvals.test.ts`, 8 cases — empty, each of the three non-pending statuses, single pending, oldest-of-many, mixed history, and negative-skew clamping.

**Revert-and-check, run on both sides rather than reasoned about:**

- *Frontend*: changing the derivation to measure from the newest park fails exactly one test — `measures from the oldest park, not the newest` — with `expected { count: 3, since: 9000, waited: 1000 } to deeply equal { count: 3, since: 2000, waited: 8000 }`. The other seven still pass, so it fails for the intended reason rather than because everything broke.
- *Backend*: restoring `kind` to the DTO fails `a_parked_approval_appears_on_its_own_task` with `` `kind` was dropped with the Approvals tab (#468) ``.

## Documentation

No doc change needed — `grep -rn "read_workspace_state\|Approvals tab" docs/` finds nothing; the tab's rationale lived in doc comments, which are rewritten in place on both `TaskApproval` definitions. The now-stale "carries no payload because a task read is the wrong place to widen exposure" comment is replaced: that argument does not hold (the sibling route already returns the payload to any member), so the honest reason is recorded instead — this line does not need one — with a pointer to #618 for the question it raises.

### DIFF
--- a/frontend/src/api/tasks.ts
+++ b/frontend/src/api/tasks.ts
@@ -328,25 +328,25 @@ export type TaskApprovalStatus = "pending" | "approved" | "denied" | "expired";
  * Distinct from an `approval` {@link TimelineEntry}, which can only describe a
  * *resolution*: an approval still parked has no resolution event, so it cannot
  * reach the timeline at all, and it is the row that matters most — the card is
- * stopped behind it. Carries no payload; the Approvals page is where a sign-off
- * is read in full and decided.
+ * stopped behind it.
+ *
+ * **Three fields on purpose (#468).** This used to back an Approvals tab on the
+ * task card. That tab is gone — approvals are decided in one place, and a
+ * second half-surface beside it could not decide anything. All the card does
+ * now is say *that* it is waiting and for how long, linking to the Approvals
+ * page, so this type carries only what that line reads. `kind`,
+ * `resolvedAtMillis` and `waitedMillis` went with the tab; the park to resolve
+ * span is still on the `approval` {@link TimelineEntry}.
  */
 export interface TaskApproval {
   /** The approval's id, the same one the Approvals page resolves against. */
   id: string;
-  /** The parked effect's dotted kind, e.g. `payment.send`. */
-  kind: string;
-  /** Epoch-millis the effect parked. Rows are ordered by this, oldest first. */
+  /**
+   * Epoch-millis the effect parked. Rows are ordered by this, oldest first, and
+   * the card measures "waiting for N" from the oldest pending one.
+   */
   atMillis: number;
   status: TaskApprovalStatus;
-  /** Epoch-millis the resolution landed; absent while pending. */
-  resolvedAtMillis?: number;
-  /**
-   * The park to resolve span. Absent while pending, where the console runs that
-   * clock itself from `atMillis`, and for an approval whose park instant the
-   * host cannot recover.
-  */
-  waitedMillis?: number;
 }
 
 /**
--- a/frontend/src/lib/task-approvals.ts
+++ b/frontend/src/lib/task-approvals.ts
@@ -1,0 +1,54 @@
+import type { TaskApproval } from "@/api/tasks";
+
+/**
+ * What the task card says about approvals (issue #468).
+ *
+ * `null` when the card is not waiting on anything — the caller renders nothing
+ * at all rather than a "no approvals" line, which is the clutter the removed
+ * Approvals tab was made of.
+ */
+export interface PendingApprovalWait {
+  /** How many of this task's approvals are still parked. */
+  count: number;
+  /** Epoch-millis the *oldest* parked one arrived. */
+  since: number;
+  /** How long the card has been stopped, measured to the caller's clock. */
+  waited: number;
+}
+
+/**
+ * Derives the card's one approval line from this task's approvals.
+ *
+ * Lives here rather than inline in the view for the reason every other
+ * derivation in `lib/` does: getting it wrong produces a card that looks
+ * completely normal while saying the wrong thing. A card that reports the
+ * *newest* park reads as freshly blocked no matter how long it has actually sat
+ * there, and nothing about the rendering would look broken.
+ *
+ * Three decisions worth stating, because each is a way this could be silently
+ * wrong:
+ *
+ * - **Only `pending` counts.** `approved`, `denied` and `expired` are
+ *   resolutions; they are already on the timeline with their own waited span,
+ *   and counting them here would rebuild the removed tab one row at a time.
+ * - **The oldest park wins**, not the newest. It is how long this card has
+ *   really been stopped. With the newest, a second effect parking behind the
+ *   first would reset a clock that should be climbing.
+ * - **`waited` is clamped at zero.** `atMillis` comes from the host and `now`
+ *   from the browser; a clock skew of a few seconds must not render "waiting
+ *   for -3s".
+ */
+export function pendingApprovalWait(
+  approvals: readonly TaskApproval[],
+  now: number,
+): PendingApprovalWait | null {
+  let count = 0;
+  let since = Number.POSITIVE_INFINITY;
+  for (const a of approvals) {
+    if (a.status !== "pending") continue;
+    count += 1;
+    if (a.atMillis < since) since = a.atMillis;
+  }
+  if (count === 0) return null;
+  return { count, since, waited: Math.max(0, now - since) };
+}
--- a/frontend/src/views/TaskDetailView.tsx
+++ b/frontend/src/views/TaskDetailView.tsx
@@ -59,7 +59,6 @@ import {
   type SteerAction,
   type Task,
   type TaskApproval,
-  type TaskApprovalStatus,
   type TaskDetail,
   type TaskPlan,
   type TimelineEntry,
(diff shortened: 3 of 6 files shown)


oxoxDev (commented):
Checked this against the decision comment on #468 rather than the issue body, and the read is correct: the comment does supersede the body, it does land on "remove the tab", and its reasoning — the data is already reachable through `GET …/approvals`, so withholding it on the task read was a DTO-scoping preference rather than an authorisation boundary — is the reason widening was never the fix. Working from the comment and saying so at the top of the PR was the right call.

Keeping the signal while deleting the surface is the part that makes this a fix rather than a removal. "Deleting the tab and saying nothing would trade one bad surface for a worse silence" is exactly right, and rendering nothing when nothing is pending — rather than a permanent "no approvals" line — is the difference between a signal and more of the clutter being deleted. Putting `AwaitingApprovalRow` above the tab strip so a stalled card announces itself whichever tab is open follows from that, and putting the derivation in `lib/` with tests follows the local convention for the right stated reason: every way it can be wrong produces a card that looks completely normal while misreporting.

Shrinking `TaskApproval` from six fields to three instead of growing it is the correct direction. Worth being explicit that this is tidiness rather than a security improvement — by your own analysis the same data is already served next door — so nobody later reads the narrowing as a boundary and builds on it as one.

**Minor — sweep for references to the surface being deleted.** `frontend/src/api/runs.ts:181` documents wording as "the approvals tab and the …". Once the tab is gone that comment points at nothing, and stale references to removed UI are how the next person concludes a surface still exists. Worth a grep for `approvals tab` across `frontend/src` and `docs/` in this PR rather than leaving it for whoever trips on it.

**One for @oxoxDev rather than this PR:** the decision comment says "Scope, replacing the body", and the body still describes widening the projection so the tab can render the shared card. Until it is rewritten, anyone reading #468 top-down gets the superseded plan and this PR looks like it contradicts its own issue. This PR flags that honestly, which is the right thing to do, but it is not something the PR can fix from here.

PR #630, opened 2026-08-11
approvals: journal a start/finish bracket around every cycle

### LINKED ISSUE #390: approvals: a detached follow-up cycle's outcome is not observable in the journal
## Context

#383 moved an approval's follow-up cycle onto its own task, so a dropped client
connection no longer cancels the agent's re-dispatch. The verdict is settled
inline and the cycle runs detached behind it.

That closes the data-loss hole, but it moves the cycle's outcome out of the
caller's reach in the detached mode: when a resolve returns
`{ "recorded": true }`, the only record of whether the continuation then
succeeded is a host-side log line.

## What is missing

There is no durable, queryable answer to "did the turn this approval kicked off
actually run?". The journal records the verdict, the grant, and the grant's
redemption — but nothing that brackets the turn itself.

A turn-level **started / finished** journal pair would make it observable:

- `started` written as the follow-up cycle takes the serial lock;
- `finished` written with its outcome (completed, or failed with a reason).

A cycle that failed, or one lost to a host restart mid-turn, would then be
visible as a `started` with no `finished`, rather than only as a log line
nobody reads.

## Why it matters beyond bookkeeping

The current recovery story is "re-approving is a safe no-op that mints no second
grant, so the operator can retry". That is true, but it only helps an operator
who *knows* to retry. Right now nothing tells them the continuation failed —
the approval left the queue, so the queue looks correct either way.

This also feeds #379 (raise approvals in the conversation): surfacing the
continuation in the thread needs the continuation's own lifecycle to be
readable, not just the verdict's.

## Scope

- Journal record types for the turn bracket, plus replay handling.
- Written by the cycle runner, so it covers every cycle rather than only the
  approval path.
- A read surface for "cycles that started and never finished".

## Related

- #383 — detached the follow-up cycle; deferred this deliberately.
- #380 — the field report behind it.
- #379 — depends on the continuation being observable.

M3gA-Mind:
## Summary

Fixes #390, scoped to **reason 1 only** per @oxoxDev's decision — reason 2 (feeding #379)
shipped without it, since #379 and #383 are both closed.

A cycle now journals a **start / finish** bracket, so "did the turn this approval kicked
off actually run?" has a durable answer. A failed continuation was previously visible only
as a host-side log line, which means "re-approving is a safe no-op, the operator can
retry" only ever helped an operator who already knew to retry.

### The bracket opens BEFORE the serial lock — the one real decision here

The issue body asks for `started` "as the follow-up cycle takes the serial lock". **That
placement cannot see the failure the issue exists for**, and this PR deliberately departs
from it.

The lock is per-company and held for a **whole cycle** (`cycle.rs:20`). A continuation
spawned behind a busy company waits on it for an unbounded time — and every path that
produces "I approved, it said `recorded: true`, nothing happened" is on the near side of
that lock:

* the host dies after `tokio::spawn` but before the task is first polled;
* the host dies while the task is queued on the lock;
* the spawned task panics before the cycle body runs.

A post-lock bracket reports all three as though the cycle had never been asked for — i.e.
exactly today's behaviour, wearing the appearance of coverage. So `cycle_id` is now minted
in `CycleRunner::run` and the start is written before `serial.lock().await`.

### What this still does not cover, stated in the code

A host dying between the **durable verdict** and `tokio::spawn` writes no start at all, so
neither the bracket nor the sweep can see it and the operator is as blind as before.
Closing that needs a record written when the verdict settles — an *owed-continuation*
ledger, which is a different feature from a cycle bracket and is deliberately not built
here. It is named in `record_cycle_started`'s doc comment, in the same register as
`run_supervisor.rs:26` naming its two known gaps.

### The sweep is half the fix, not a follow-up

`RuntimeJournal::sweep_interrupted_cycles` settles every cycle left open by a dead host,
called from `builder.rs` **gated on `handover.is_none()`**. Without it, a "started with no
finished" read reports every crashed cycle as in-flight forever — worse than the log line
it replaces, because it looks like live work.

It rests on the same three invariants as `sweep_interrupted_runs`: a cycle journals its
start before doing anything, every cycle is driven in this process, and one process owns
this journal. So at boot an unmatched start cannot belong to a live cycle. The
must-not-run-on-a-rebuild gating is the part that is easy to miss: a cycle survives a live
runtime swap, so sweeping mid-life would stamp "interrupted by a host restart" on a
running cycle and leave two contradictory outcomes for one id.

### Other scope points

* **Panic gap named** (scope point 2): a cycle task that unwinds journals no finish and
  reads as open until the next boot sweep. Same exposure `run_supervisor` documents for a
  panicking workflow run; the same remedy covers both. Documented on
  `record_cycle_finished`.
* **Banked / no-op paths open no bracket**: `still_waiting_report` and
  `already_resolved_report` return without reaching `run`. Correct rather than an
  oversight — a banked decision is *correctly* not running, and bracketing it would leave
  an open cycle that never closes and never should. Stated on `run`.
* **Journal ordering**: the start append serialises on `JOURNAL_WRITE_LOCKS`, not on the
  serial lock, so brackets from concurrent cycles interleave in the file. Harmless for the
  fold (which keys on id, not position), but the journal stops reading as one sequential
  story by hand. Documented.
* **No new identifier**: reuses `cycle_id`, which the journal already correlates approvals
  on (`ApprovalParked.cycle`). Same principle `run_supervisor` states for `run_id`.
* **Records live in `RuntimeJournal`, not `CompanyEvent`** — `CompanyEvent` is a closed
  binding enum with no marker variants, and the journal module doc is explicit that
  markers live there. This also gives replay handling via the existing fold.

### Accepted cost

Every cycle now performs one journal append before the lock — write amplification on the
hot path, outside the critical section. Small, and the alternative is a bracket that
misses its own case.

## API Or Behavior Changes

**API:** `RuntimeJournal` gains `record_cycle_started`, `record_cycle_finished`,
`open_cycles`, `sweep_interrupted_cycles`; new public `OpenCycle` and
`INTERRUPTED_BY_HOST_RESTART`. Two new `JournalRecord` variants (internal).
`CycleRunner::run_locked` (private) takes the cycle id.

**Behaviour:** cycles write two journal lines they did not before, and a boot sweep
settles cycles left open by a dead host. No change to what any cycle does.

**Journal compatibility:** additive. A pre-#390 journal replays unchanged — a
`CycleFinished` for a cycle this journal never started removes nothing, which is right
rather than a gap: it also had no start to be matched against.

## Tests

All exit-code gated via unpiped redirect.

- [x] `cargo clippy --all-targets -- -D warnings` — EXIT:0 (default lane). Gated lane in
      CI's exact form, `--locked --no-deps --features openhuman,tinycortex --all-targets`
      — EXIT:0.
- [x] `cargo build --all-targets` — covered by the two clippy runs, which build all
      coverage.
- [x] `cargo test` — default lane **1838 passed / 0 failed**; gated lane
      `--features openhuman,tinycortex --lib` **2752 passed / 0 failed**. Integration
      targets not re-run on this branch (library-only diff); green on the parent, #603.

### Negative controls

**A — the placement control, and the important one.** Moving the `started` write to after
`serial.lock().await` (the issue body's literal wording) fails **exactly one test**:

```
runtime::cycle::test::a_cycles_bracket_opens_before_the_serial_lock
```

That it fails *alone* is the point. A bracket that opens after the lock still opens, still
closes, and still reads correctly once the cycle is over — so every other test in the
suite passes. The failure mode of this change is a correctly-typed id written in the wrong
place, which no compiler and no other test can see. Hence a test whose whole job is to
hold the lock and assert the cycle is already visible as open while queued behind it.

**B — the sweep control.** Making `sweep_interrupted_cycles` settle nothing fails:

```
runtime::journal::test::an_interrupted_cycle_replays_as_open_and_is_swept_at_boot
```

Both restored; green again (2752 passed, EXIT:0).

### On moving `cycle_id`

Traced rather than grepped, because this was flagged as the riskiest part. `cycle_id` was
minted at `cycle.rs:212` inside `run_locked`; between that function's first statement and
the mint, the input-append loop, the `begin_run` call and the history/context/roster loads
are keyed on the company and the event and never on the cycle. `run_locked` has exactly
one caller. `settle_approval` and `recover` run no cycle and correctly get no bracket. The
move was mechanical.

## Documentation

The rationale lives where the next reader will hit it: `record_cycle_started` carries the
before-the-lock argument and the residual gap, `record_cycle_finished` carries the panic
gap, `sweep_interrupted_cycles` carries the boot-only and rebuild-gating arguments, and
`CycleRunner::run` carries the mint-placement note and the no-bracket-for-banked-paths
rule. No user-facing docs affected.

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -5620,7 +5620,7 @@ members = ["eng1", "eng2"]
             None,
         )
         .with_requests(requests.clone());
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL" });
+        let args = crate::policy::test_support::composio_send_args();
         let request = ToolPolicyRequest::new(
             "composio_execute",
             args.clone(),
@@ -5856,7 +5856,14 @@ members = ["eng1", "eng2"]
         let log: Arc<dyn crate::ports::EventLog> =
             Arc::new(crate::store::FsEventLog::new(dir.path().to_path_buf()));
         let requests = crate::harness::policy::ApprovalRequestQueue::default();
-        let args = serde_json::json!({ "tool_slug": "GMAIL_SEND_EMAIL", "to": "a@b.test" });
+        // Issue #470: a real catalogued send, with the action's own parameters
+        // under `arguments` where the tool's schema puts them — so the
+        // re-dispatch path this test covers carries a call the classifier can
+        // actually read.
+        let args = crate::policy::test_support::composio_args_with(
+            crate::policy::test_support::COMPOSIO_SEND_SLUG,
+            serde_json::json!({ "to": "a@b.test" }),
+        );
         requests
             .grants()
             .grant(crate::runtime::grants::GrantedCall {
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -1956,12 +1956,24 @@ impl HarnessBrain {
     /// already-computed operator reply along with it. That is precisely the
     /// silent-disappearance failure this issue exists to fix, so each failure is
     /// logged at `error` and the drain continues.
-    async fn park_approval_requests(&self, host: &dyn CycleHost) -> Result<()> {
-        for request in self
-            .deps
-            .approval_requests
-            .drain(crate::harness::policy::MAX_APPROVAL_REQUESTS_PER_TURN)
-        {
+    async fn park_approval_requests(&self, host: &dyn CycleHost) -> Result<Option<String>> {
+        let cap = crate::harness::policy::MAX_APPROVAL_REQUESTS_PER_TURN;
+        let drained = self.deps.approval_requests.drain(cap);
+
+        // Issue #561: the overflow used to end here, silently. The queue entries
+        // are already gone and the turn is over, so a log line is the only trace
+        // — and a log line is not something an operator reads. Kept loud for the
+        // operator's sake *and* returned, so the cycle can say it out loud.
+        if drained.discarded > 0 {
+            log::warn!(
+                "[harness::brain] {} gated tool call(s) past the per-turn cap of {cap} were \
+                 discarded and will not reach the operator",
+                drained.discarded
+            );
+        }
+
+        let notice = drained.overflow_notice(cap);
+        for request in drained.requests {
             match host.park_effect(request.effect).await {
                 Ok(approval_id) => log::info!(
                     "[harness::brain] parked '{}' for operator approval (id={approval_id}): {}",
@@ -1977,7 +1989,7 @@ impl HarnessBrain {
                 ),
             }
         }
-        Ok(())
+        Ok(notice)
     }
 
     /// Executes one drained delegation from the orchestrator's turn.
@@ -2294,7 +2306,23 @@ impl Brain for HarnessBrain {
         // Issue #172: every approval-gated tool call this cycle's turns hit is
         // parked on the host's gate now, so it shows up on the operator's
         // Approvals page instead of only being narrated away in chat.
-        self.park_approval_requests(host).await?;
+        //
+        // Issue #561: and if the turn gated more calls than one turn may raise,
+        // the operator is told so here rather than discovering it as silence.
+        // Pushed as its own bubble instead of appended to the reply above,
+        // because the reply is the agent's answer and this is the system saying
+        // the agent was cut off — and because a turn whose only outcome was
+        // overflow has no reply to append to.
+        if let Some(notice) = self.park_approval_requests(host).await? {
+            channel_responses.push(OutboundMessage {
+                message_id: None,
+                task_id: None,
+                channel: "operator".to_string(),
+                text: notice,
+                steps: Vec::new(),
+                reply_to: None,
+            });
+        }
 
         // The runtime requires at least one channel response per cycle.
         if channel_responses.is_empty() {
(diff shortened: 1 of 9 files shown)


oxoxDev (commented):
Opening the bracket before the serial lock rather than as the cycle takes it — departing from the issue text — is correct, and the argument is the strongest thing here. The lock is per-company and held for a whole cycle, so a continuation behind a busy company waits an unbounded time, and all three failure modes that produce "I approved, it said `recorded: true`, nothing happened" sit on the near side of it. A post-lock bracket would report every one of them as a cycle that was never asked for: today's blindness wearing the appearance of coverage. Worth making sure that reasoning lands on #390 itself, since it contradicts the issue's own instruction and the next person to read the issue will not have this PR in front of them.

Treating the sweep as half the fix rather than a follow-up is the other thing this gets right. A "started with no finished" read without it reports every crashed cycle as in-flight forever, which is worse than the log line it replaces, because stale work that looks live is acted on.

**Question — is `handover.is_none()` the whole guard for "one process owns this journal"?**
The sweep rests on three invariants, and the third is the one I would want stated against the hosted topology rather than against the precedent. `sweep_interrupted_runs` establishing the pattern makes this consistent, but consistency inherits the assumption rather than checking it. In shared-single-DB mode the manager injects `OPENCOMPANY_TENANT_ID` and every tenant shares one logical MongoDB, with isolation at the application layer; company ids are namespaced, but that is a different property from "exactly one process sweeps this journal at boot". If two workloads can ever reach the same journal, a boot sweep settles another process's live cycles as interrupted — and the failure looks exactly like the bug being fixed, from the other direction. If `handover.is_none()` plus the namespacing already makes that unreachable, that is a sentence in the doc comment, not a change.

**The named gap should be an issue, not a doc comment.**
A host dying between the durable verdict and `tokio::spawn` writes no start, so neither the bracket nor the sweep sees it. Naming that in `record_cycle_started`'s doc comment, in the same register as `run_supervisor.rs:26`, is the right instinct and better than silence. But this is the third PR in this batch to park real remaining work in prose — #554's default-server list, #625's workflow half of the discard notice, and now the owed-continuation ledger. Each is individually defensible; together they are an accumulating set of known gaps whose only record is a comment inside the change that decided not to fix them. File this one and link it.

**This is the fourth PR in the undeclared stack.**
The diff carries #609's `Reach::ExternalRead` and #603's `test_support` changes, on top of the same eight files, plus `runtime/builder.rs`. So the chain is #603 → #609 → #625 → #630 and only #609 names a parent. #627 shows you already know the fix — dependency, merge order, and the commit range that isolates the change, stated at the top. Two minutes per PR, and it is the difference between reviewing your change and reviewing the same three hunks for the fourth time.

**Scope note on this review:** I have read the design and the interfaces rather than every line of the 1411. The bracket placement, the sweep's role and the stack structure are what I checked; the journal write paths I have taken on the tests.

Issue #319, opened 2026-08-03
connections: route hosted OAuth through the platform backend instead of per-tenant provider apps


oxoxDev:
## Summary

Hosted OpenCompany connections should use **the same OAuth the OpenHuman backend already uses** — the same provider applications, the same authorize and callback URLs, the same token custody — rather than each tenant holding its own client id and secret per provider.

A tenant should paste nothing and register nothing. It already authenticates with a platform-minted, audience-bound identity; that identity is what the backend uses to attribute a connection.

## Why now

This is not currently possible on staging, and the reason is concrete: **the manager injects fifteen environment variables into every tenant, and none of them is an OAuth provider credential.** There is no `OPENCOMPANY_OAUTH_*` anywhere in the manager source, its manifests, or the deploy workflow.

So today, on a hosted tenant, `provider_config` finds nothing and every connect attempt fails at `start` before a redirect is ever issued. Connections cannot work hosted, and no amount of fixing the redirect path changes that.

There are two ways out: provision provider credentials into every tenant, or stop needing them. This issue is the second.

## The precedent is already in this repo

**Composio works this way today.** From its module documentation:

> On the hosted platform a company needs to paste nothing. The instance already authenticates with a platform-minted, audience-bound identity, and Composio calls present that — the backend derives the Composio entity from it. The read shape reports this as `credentialSource: "attested"`, and there is nothing stored on the instance to leak, rotate, or lose.

That is the target shape for every connection. Composio also keeps a per-company BYO override for self-hosting, explicitly framed as an escape hatch rather than a deployment mode — the same split applies here.

The metered `web_search` is the second precedent: agents reach a third party through the backend's `/agent-integrations/*` surface, presenting the platform identity, with no provider credential in the pod.

## What the backend already provides

- `/auth/{provider}/login` — the login handshake OpenHuman's Welcome page uses, for Google, GitHub and Twitter/X.
- `/oauth/connect` — starts a connection for a user, with **incremental scopes**: it looks up the scopes a skill needs and merges them with what was already granted, so connecting Drive after Gmail does not revoke Gmail.
- `GET /integrations` — the user's connected integrations.
- `/agent-integrations/*` — the proxy through which agents reach third parties.

Provider apps, client secrets, redirect URIs and token refresh all live there already. Nothing about that needs rebuilding.

## Work

[...]

Issue #300, opened 2026-08-03
auth: OAuth connections break inside the onboarding flow after redirect


oxoxDev:
Part of the MVP spec epic #183 (§1 Onboarding & Auth). Filed separately because the connection bug is the blocking half and is independently shippable.

## Problem

An OAuth/integration handshake started **inside the onboarding flow** does not complete. The redirect comes back and the user is left in a broken state rather than on the correct onboarding step.

This is the one the spec singles out: connections "should be smooth and just work." Nothing in the ~30 PRs merged this weekend went near it.

## Scope

**In scope**
- Diagnose where the onboarding redirect loses the handshake. Reproduce first, on a live host, before changing anything — the failure may be state carried across the redirect rather than the callback itself.
- Complete the handshake and return the user to the step they left.
- Cover every provider reachable from onboarding, not just the one used to reproduce.

**Also §1, may split again if it grows**
- Login via Google, GitHub and Twitter/X. Only one method exists today.
- Post-login lands directly in onboarding, with no intermediate landing page.
- Integrations available on the parent product mirrored here.

## Acceptance criteria

- [ ] Starting a connection from within onboarding completes and returns to the correct step, with no error state
- [ ] Reproduced on a live host before the fix, and the same path verified after — a unit test alone is not evidence for a redirect flow
- [ ] Every provider offered during onboarding is exercised, not just one
- [ ] A failed or cancelled handshake surfaces a recoverable state rather than a dead end

## Notes

Do not paste any real credential, token or client secret into the issue, a PR body, or a commit. Use a stub provider where a live one is not needed.

Scope: **medium**, mostly diagnosis.

PR #633, opened 2026-08-11
fix(connections): route the categorised tiles through Composio (#599)

### LINKED ISSUE #599: connections: every categorised Connect tile fails on a hosted tenant — the tiles route through native OAuth, not Composio
## Summary

On Settings → Connections, the categorised provider cards (Communication / Productivity / Developer / Finance) have **Connect** buttons that fail for every provider on a hosted tenant. Verified live on `smoke1.staging.opencompany.work`.

The Composio "Sign in per provider" list directly above them works. The two surfaces route through completely different backends, and only one of them is wired for hosted.

## Evidence

`POST …/connections/{provider}/start`, from an authenticated admin session:

| Tile | Status | Body |
|---|---|---|
| `slack` | 400 | `provider 'slack' is not enabled on this host` |
| `notion` | 400 | `provider 'notion' is not enabled on this host` |
| `google-calendar` | 400 | `provider 'google-calendar' is not enabled on this host` |
| `stripe` | 400 | `provider 'stripe' is not enabled on this host` |
| `github` | 400 | `provider 'github' is not enabled on this host` |

By contrast, the Composio path on the same page, same session, same providers:

```
POST …/composio/authorize {"toolkit":"notion"}
→ 200 {"connectUrl":"https://connect.composio.dev/link/lk_…"}
```

Clicking that button opens Composio's hosted connect page correctly. So this is not a session, credential or permissions problem — the two code paths simply differ.

## Two separate causes, and the second is the real one

**1. Eight of eleven tiles have ids that no backend key matches.** `frontend/src/lib/connections.ts` documents this rule itself:

> A tile whose `id` has no matching backend key can never complete OAuth — `startConnection(id)` resolves no `provider_config` and the host answers "provider not enabled". Backend keys today: `slack`, `github`, `google`, `gmail`.

The shipped tiles are `gmail`, `slack`, `google-calendar`, `notion`, `google-drive`, `dropbox`, `github`, `stripe`, `hubspot`, `x`, `linkedin`. `well_known()` in `src/server/ops/connections.rs` recognises only `slack`, `gmail`, `github`. So eight tiles were structurally unable to work from the moment they were added.

**2. But even the three valid ids fail** — `github` and `slack` 400 identically above. So fixing the ids fixes nothing on hosted. The underlying reason is the one `#319` sets out: the manager injects fifteen environment variables into every tenant and **none is an OAuth provider credential**, so `provider_config` resolves nothing and every connect attempt dies at `start` before a redirect is ever issued.

That makes the whole categorised grid dead UI on hosted — not degraded, not partially working. Every button, every time.

## Why file this separately from #319

`#319` is the architectural fix (route hosted connections through the platform backend, presenting the instance identity, with no tenant-held provider credentials). This issue is narrower and more urgent: **right now the console presents an operator with a grid of buttons that cannot succeed, and says nothing about it.** Two of the tiles even render "Connected ✓ via composio" — reflecting Composio state while their own Connect button targets a backend that has never worked here — which reasonably reads as "this surface works, and the others just need clicking."

The minimum honest outcome is that these buttons stop lying. The best outcome is the one the user asked for directly: wire the tiles through the Composio API that already works, since the parallel list on the same page proves the plumbing is there.

## Suggested direction (author's call)

1. **Point the tiles at Composio** — map each tile `id` to its Composio toolkit slug and call `…/composio/authorize`, exactly as the provider list does. `google-calendar` → `googlecalendar`, `google-drive` → `googledrive`, `x` → `twitter`, and so on. This makes all eleven work rather than three, and removes the id/`well_known` coupling entirely.
2. **Or gate the grid** — hide or disable tiles whose provider is not enabled on this host, with a line saying why. Smaller, and honest, but leaves the capability missing.
3. Keep the native `provider_config` path as the self-hosted route, framed the way Composio frames its BYO token: supported for standalone operation, not the hosted path.

Option 1 is what makes the page match what an operator expects, and it is a console change — no manager or backend work, so it does not block on `#319` or on attestation.

## Reproduce

1. Sign in to a hosted tenant console as an admin.
2. Settings → Connections → scroll to the categorised cards.
3. Click **Connect** on any tile.
4. The request 400s with `provider '<id>' is not enabled on this host`.

## Related

- `#319` — route hosted OAuth through the platform backend instead of per-tenant provider apps. That is the durable fix for the native path; this issue is about the console surface being unusable in the meantime.
- `#597` — separate display defect on the Inference card on the same page.

CodeGhost21:
## Summary

Closes #599.

Every **Connect** button in the categorised provider grid failed on a hosted tenant with `provider '<id>' is not enabled on this host`. This routes the tiles through Composio — the surface directly above them that already works — while keeping the native path as the self-hosted route.

Console-only, as the issue predicted: no manager or backend work, so it does not block on #319.

## RCA

The issue's diagnosis is right but stops one level short. The #319 guard that exists to prevent exactly this 400 **is present and correct** — it just never fires.

1. The grid renders all 11 tiles from `CONNECTION_PROVIDERS` and looks each up as `states[p.id]`.
2. `states` comes from `GET …/connections` → `project_connections` (`src/server/ops/connections_read.rs:352`), which emits **one row per `record.manifest.connections` entry**, plus extras only for providers Composio reports already connected. A provisioned tenant declares none, so `states` is `{}`.
3. `platformManaged` was `Object.values(states).some(s => s.credentialSource === "attested")`. `.some()` over an empty array is `false`, so no tile inherited `attested`, and `noRoute` needs `"none"` where `source` was `undefined`.
4. Every tile fell through to Connect → `provider_config()` finds no `OPENCOMPANY_OAUTH_*` → **400**.

**The root cause is a keying mismatch**: a guard keyed off per-provider rows protecting a grid that renders independently of the manifest.

Two further defects fall out of the same mismatch:

- **Slug mismatch.** Reconciled Composio rows are keyed by Composio slug (`googlecalendar`, `googledrive`, `twitter`); tiles by console id (`google-calendar`, `google-drive`, `x`). `toolkit_slug()` normalizes backend-side but the console did a raw `states[p.id]` lookup — so a genuinely connected Google Calendar never lit up its tile. This is why only *some* tiles showed "Connected ✓ via composio".
- **Dead ids.** 8 of 11 ids have no `well_known()` key, so they could not connect even self-hosted with credentials registered.

## API Or Behavior Changes

No host changes. Console behaviour:

- **`connectRoute` is the single decision**, used for both what renders and what the click calls, so the button shown and the call made cannot disagree. Precedence: `static` → native (a registered provider application is a deliberate act by a self-hoster and must not be taken away); else Composio when it can authorize that toolkit; else `managed`; else `unavailable`.
- **A tile with no route renders no button.** `unavailable` is the state that was missing — this is what stops the buttons lying.
- **Tiles carry their Composio toolkit slug** and match host rows through it, so a connected `googlecalendar` / `googledrive` / `twitter` lights up the right tile. All eleven become connectable rather than three.
- **Disconnect only when there is a native credential to revoke.** There is no Composio disconnect route on the host (`/composio` exposes status, token, authorize, connections and nothing else), so surfacing it on a Composio-only connection would blank a secret that was never set, report success and change nothing. An absent `via` still gets the button — that is an older host, not a Composio-owned connection.
- The grid waits for the Composio probe before painting, so tiles do not flash "Not available" and then flip to Connect.

## Two things worth reviewer attention

**An existing e2e assertion was weakened, deliberately.** `oauth-onboarding-resume.spec.ts` asserted a Connect button was enabled. On that harness — no `[[connection]]` entries, no `composio` feature, `host.sh` runs `env -i` with no `OPENCOMPANY_OAUTH_*` — that button could never have completed a handshake. It was asserting the bug. It now asserts the tile explains itself; #300's actual property (bounce-back lands in a working console, param stripped) is untouched and still asserted.

**A regression this fix would otherwise have introduced.** Fixing slug matching means *more* tiles correctly show as Composio-connected — which would have surfaced a Disconnect button that silently does nothing. Hence the `via`-gated Disconnect above.

## Tests

- [x] `npm run typecheck`
- [x] `npm run typecheck:e2e`
- [x] `npm test` — 291 passed (16 new in `test/unit/connection-route.test.ts`)
- [x] `npm run build`
- [x] `npm run e2e` — **96 passed, 8 skipped** against a live host

The 8 skips are pre-existing: the `LIVE_BRAIN`-gated specs (#467) and the Brain spec parked by #302.

New unit coverage pins the route matrix directly, including the two regression guards: every tile routes to Composio on a tenant that declares no connections, and every tile reports `unavailable` when no route can succeed.

`test/unit/connection-route.test.ts` also asserts the `unavailable` case empirically end-to-end — the e2e run confirms that on a host with no route the tile now says so rather than offering a Connect that 400s.

## Documentation

Module docs in `frontend/src/lib/connections.ts` describe the two routes and why `connectRoute` can answer "neither". The `id` doc no longer says an id outside `well_known()` is a dead tile — that is now the Composio path's job — and the two `DEAD TILE` markers are removed as resolved.

## Not in scope

- A **Composio disconnect route** on the host. Its absence is why Disconnect is `via`-gated here; worth its own issue.
- **#319** proper — routing hosted native OAuth through the platform backend. This makes the page honest and functional in the meantime, exactly as #599 frames it.

## Summary by CodeRabbit

- **New Features**
  - Added support for native, hosted, managed, and unavailable connection routes.
  - Connections can open hosted authorization flows and track completion automatically.
  - Added clearer status messages and management guidance for unavailable or externally managed connections.
  - Improved provider matching across naming conventions, including Google services and X/Twitter.

- **Bug Fixes**
  - Improved connection-state detection when providers are configured through hosted services.

### DIFF
--- a/frontend/src/lib/connections.ts
+++ b/frontend/src/lib/connections.ts
@@ -1,6 +1,23 @@
 // The catalog of third-party accounts a company can act through. This is the
 // console's view of what *can* be connected; whether a host can actually run
 // the OAuth handshake depends on its `/connections` surface (see the client).
+//
+// ## Two routes, one tile (issue #599)
+//
+// A tile can be connected two ways, and which one is live is a property of the
+// *host*, not of the tile:
+//
+//  - **Composio** — the hosted path. Composio runs the OAuth itself and the
+//    resulting connection is a tool belt the agents actually receive.
+//  - **Native** — this host's own registered provider application, the
+//    self-hosted hatch documented on `src/server/ops/connections.rs`. Reported
+//    as `credentialSource: "static"`.
+//
+// Until #599 every tile hard-routed to native. On a hosted tenant no
+// `OPENCOMPANY_OAUTH_*` variable is injected, so all eleven Connect buttons
+// 400'd with "provider is not enabled on this host" — a grid of buttons that
+// could never succeed. [`connectRoute`] is now the single place that decides,
+// and it can answer "neither", so a button that cannot work is never rendered.
 
 export type ConnectionCategory =
   | "Communication"
@@ -12,16 +29,32 @@ export type ConnectionCategory =
 
 export interface ConnectionProvider {
   /**
-   * The provider identity. This string MUST be identical end-to-end: the
-   * manifest `[[connection]] provider = "…"`, this catalog `id`, and the
-   * backend `well_known(provider)` key (`src/server/ops/connections.rs`) are the
-   * same token. A tile whose `id` has no matching backend key can never
-   * complete OAuth — `startConnection(id)` resolves no `provider_config` and the
-   * host answers "provider not enabled". Backend keys today: `slack`, `github`,
-   * `google`, `gmail`. Keep new tiles aligned to an existing key (or add the
-   * backend key first); do not invent console-only ids.
+   * The provider identity in *this host's* namespace: the manifest
+   * `[[connection]] provider = "…"` and the key `GET …/connections` reports
+   * status under.
+   *
+   * It is also the token the native hatch resolves — `startConnection(id)` needs
+   * a matching `well_known(id)` key in `src/server/ops/connections.rs` (today:
+   * `slack`, `github`, `google`, `gmail`). An id outside that set has **no
+   * native route**, which is no longer a dead tile: {@link connectRoute} sends
+   * it down the Composio path instead, and reports `unavailable` when neither
+   * route is open rather than rendering a Connect that 400s.
    */
   id: string;
+  /**
+   * The Composio toolkit slug this tile authorizes against — the hosted route,
+   * and the only route on a hosted tenant.
+   *
+   * Composio slugs are lowercase and unpunctuated (`googlecalendar`) while ids
+   * here are hyphenated (`google-calendar`), and a few differ outright (`x` is
+   * `twitter`). Stated per tile rather than derived, because {@link toolkitSlug}
+   * normalization alone cannot produce `twitter` from `x`.
+   *
+   * Mirrors the backend's `toolkit_slug()`
+   * (`src/server/ops/connections_read.rs`), which is what reconciles the two
+   * namespaces into one row per provider.
+   */
+  toolkit: string;
   name: string;
   description: string;
   category: ConnectionCategory;
@@ -34,6 +67,7 @@ export interface ConnectionProvider {
 export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   {
     id: "gmail",
+    toolkit: "gmail",
     name: "Gmail",
     description: "Send and read email from a connected inbox.",
     category: "Communication",
@@ -42,6 +76,7 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   },
   {
     id: "slack",
+    toolkit: "slack",
     name: "Slack",
     description: "Post updates and take requests from your workspace.",
     category: "Communication",
@@ -49,12 +84,8 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
     glyph: "#",
   },
   {
-    // DEAD TILE until aligned: backend `well_known` has no `google-calendar`
-    // key — the closest existing key is `google`. Connecting this tile fails
-    // ("provider not enabled") until either the id is changed to `google` (or a
-    // dedicated `google-calendar` key + scopes are added to
-    // `well_known`/`provider_config` in src/server/ops/connections.rs).
     id: "google-calendar",
+    toolkit: "googlecalendar",
     name: "Google Calendar",
     description: "Schedule and read events on a shared calendar.",
     category: "Productivity",
@@ -63,6 +94,7 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   },
   {
     id: "notion",
+    toolkit: "notion",
     name: "Notion",
     description: "Read and write docs and databases.",
     category: "Productivity",
@@ -70,10 +102,8 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
     glyph: "N",
   },
   {
-    // DEAD TILE until aligned: backend `well_known` has no `google-drive` key
-    // (closest existing key is `google`). See the `google-calendar` note above —
-    // align the id to a backend key or add the key before this tile can connect.
     id: "google-drive",
+    toolkit: "googledrive",
     name: "Google Drive",
     description: "Store and retrieve files and deliverables.",
     category: "Storage",
@@ -82,6 +112,7 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   },
   {
     id: "dropbox",
+    toolkit: "dropbox",
     name: "Dropbox",
     description: "Sync assets and shared folders.",
     category: "Storage",
@@ -90,6 +121,7 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   },
   {
     id: "github",
+    toolkit: "github",
     name: "GitHub",
     description: "Open issues and pull requests in your repos.",
     category: "Developer",
@@ -98,6 +130,7 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   },
   {
     id: "stripe",
+    toolkit: "stripe",
     name: "Stripe",
     description: "Create invoices and read payment activity.",
     category: "Finance",
@@ -106,6 +139,7 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   },
   {
     id: "hubspot",
+    toolkit: "hubspot",
     name: "HubSpot",
     description: "Sync contacts and deals in your CRM.",
     category: "Finance",
@@ -113,7 +147,10 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
     glyph: "H",
   },
   {
+    // Composio still spells this toolkit `twitter`; the tile keeps the current
+    // product name. Exactly the case a normalization rule cannot derive.
     id: "x",
+    toolkit: "twitter",
     name: "X",
     description: "Publish posts and read mentions.",
     category: "Social",
@@ -122,6 +159,7 @@ export const CONNECTION_PROVIDERS: ConnectionProvider[] = [
   },
   {
     id: "linkedin",
+    toolkit: "linkedin",
     name: "LinkedIn",
     description: "Publish updates and manage your page.",
     category: "Social",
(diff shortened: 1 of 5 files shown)


oxoxDev (changes requested):
**1 major, 1 minor.** The major is sequencing with #637. Checks are green now — the five failures from earlier have cleared.

The RCA goes one level deeper than the issue and lands in the right place. `platformManaged` was `Object.values(states).some(s => s.credentialSource === "attested")`, and `.some()` over an empty array is `false` — so on a provisioned tenant that declares no `manifest.connections`, `states` is `{}`, the #319 guard never fires, and every tile falls through to a 400. "A guard keyed off per-provider rows protecting a grid that renders independently of the manifest" is the actual defect, and it explains all three symptoms rather than just the reported one.

**Stating `toolkit` per tile rather than deriving it is the right call.** The tempting fix is a console-side `normalize()` mirroring the backend's `toolkit_slug()`, and that would have been a second implementation of the same rule — the drift trap this codebase avoids elsewhere (`grants_cover_server` moved rather than reimplemented in #631, one validator for default servers in #554). Eleven explicit values are auditable in a way an algorithm copy is not.

---

### Major — collides with #637 on `frontend/src/views/ConnectionsView.tsx`

#637 (`feat(connections): broker providers through the company's own TinyHumans credential`, also yours) edits the same view. It changes *whether* a tenant has a credential to connect with; this changes *how a tile decides which route to take*. Those two decisions meet in the same component, and both PRs are rewriting how a tile resolves its route — so this is a semantic overlap, not just a textual one: after both land, a tile's route depends on the manifest, the Composio reconciliation, and now the company credential tier.

Neither PR mentions the other. Worth agreeing an order, and whichever lands second should say what a tile does when the company credential is present but the provider is not in `states` — that combination does not exist in either PR alone.

### Minor — nothing couples the `toolkit` table to the backend's `toolkit_slug()`

The doc says the field "mirrors the backend's `toolkit_slug()`" (`src/server/ops/connections_read.rs:239`). Explicit is right, but a mirror with no reflection test drifts silently: change `toolkit_slug` and eleven console tiles quietly authorize against slugs the backend no longer produces, with the failure landing as "provider not enabled" — the exact symptom this PR fixes.

A single test feeding all eleven console ids through the backend's normalizer and asserting equality with the table would pin it, and would fail loudly on the day someone edits either side.

CodeGhost21:
Pushed `2d6bb3e`. Taking the review points in turn.

## Major — sequencing with #637

Agreed the overlap is semantic, not just textual, and you're right that neither PR mentioned the other.

**Proposed order: #637 first, this second.** #637 changes what a tenant *has*; this changes how a tile *spends* it. Landing the capability before the router means the router is written against the full tier set rather than being amended into it. It is also the cheaper rebase — #637 touches `ConnectionsView` in the section JSX and one `useState`, while this rewrites the route decision and the card, so replaying the smaller edit onto the larger one is the wrong way round.

**What a tile does when the company credential is set but the provider is not in `states`** — the combination neither PR covers alone:

> The tile offers a **working Composio Connect**.

`state` is `undefined`, so the `static`/`attested` arms do not fire; `hasCredential` is true because #637 reports `credentialSource: "company"` rather than `"none"`; the tile routes to Composio. That is the #599 outcome reached through #586's credential.

This is by construction rather than by luck, and I have made it explicit rather than leaving it as a happy accident. `connectRoute` names exactly two tiers — `static` and `attested` — because those are the two that describe the **local host**: one says a native handshake can complete here, the other says none ever can. Every other tier is a statement about which credential the host presents to Composio, which the router reads through the boolean `hasCredential`. So a new Composio tier is additive here by construction, and `company` needs no case of its own.

Pinned in `connection-route.test.ts` so a later change cannot quietly break the composition:

```ts
it("routes on whether a Composio credential exists, not on which tier it is", …)
```

It asserts both halves — a `company`-tier host with no `states` row gets a Composio route, and a row reporting `credentialSource: "company"` is *not* mistaken for the native hatch. That last one matters: only `static` may mean "a local handshake can complete", and #637's own `types.ts` doc agrees the company key never routes the native catalog.

One consequence worth naming for whoever rebases: the interaction is confined to `ComposioReach.hasCredential`. If #637 later makes the company key route the **native** catalog too, that assumption breaks and `connectRoute` needs a real case — the doc says so at the call site.

## Minor — nothing coupled the table to `toolkit_slug()`

Fair, and the failure mode you describe is exactly the one this PR fixes, which makes it a bad one to reintroduce. Added `console_toolkit_slugs_are_canonical_under_this_normalizer` in `src/server/ops/connections_read.rs` — it reads the real `frontend/src/lib/connections.ts` and feeds every `toolkit` through the real `toolkit_slug`.

Two deliberate choices:

- It asserts a **fixed point** (`toolkit_slug(t) == t`), not `toolkit_slug(id) == toolkit`. The latter is false for `x`/`twitter` by design — that alias is *why* the table is explicit — so asserting it would force the test to encode the exception and re-create the drift it exists to catch.
- It asserts the entry **count** first. Without that, a renamed field or reformatted catalog leaves the loop iterating over an empty list and passing, which is the failure mode where a test reports coverage it does not have.

I verified it fails rather than assuming: temporarily setting one slug to `"Google-Calendar"` produces

```
console toolkit "Google-Calendar" is not canonical under toolkit_slug; the console
would authorize a slug this host reconciles rows under a different key
```

## CodeRabbit

All three were valid; all three fixed.

**`x`/`twitter` reconciliation (`connections.ts`)** — real, and worse than "cannot match". `connectionStateFor` already matched both spellings, so a lone `twitter` row was found. The gap is when **both** rows exist: a manifest `provider = "x"` gives a disconnected `x` row, Composio's connected `twitter` arrives appended, and taking the first match reported the tile disconnected while the account was connected. A connected row now wins; direct-id precedence still decides when nothing is connected. I did not take the suggested fix of aliasing inside the shared normalization — `toolkit_slug` is the host's reconciliation rule and a console-driven special case does not belong in it; the union belongs where the two rows meet.

**Blocked popup** — taken as proposed.

**Unbounded status request** — real, and a regression this PR introduced: the grid did not wait on this call before. Bounded at 5s via `Promise.race`, mirroring the host's own `COMPOSIO_PROBE_TIMEOUT` on the connections read path. I did not add abort support to `OpenCompanyClient`; that is a shared-transport change and a larger blast radius than this PR should carry, worth its own issue.

**Nitpick — direct-id precedence unpinned:** correct, every fixture held one row. Now covered by the two-row cases above, one of which asserts precedence with nothing connected.

## Validation

`cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `cargo test` (1810 passed), `typecheck`, `typecheck:e2e`, `typecheck:unit`, `npm test` (295 passed, +4), `npm run build`, `npm run e2e` (96 passed, 8 skipped — the pre-existing `LIVE_BRAIN` gates and the Brain spec parked by #302).

CodeGhost21:
### CI: the one red check is main's, not this branch's

`Rust (openhuman, tinycortex)` fails on `2d6bb3e`. It is **pre-existing on `main`** and unrelated to this PR.

```
workflows::gate::tests::a_plain_read_and_a_metered_read_do_not_gate ... FAILED
src/workflows/gate.rs:367
[GatedToolCall { node_id: "read", slug: "read_workspace_state",
  reason: "'read_workspace_state' has an external effect and this desk runs supervised" }]
```

Evidence it is not this branch:

- **The same test fails on `main`.** Run [31487915510](https://github.com/tinyhumansai/opencompany/actions/runs/31487915510) (merge of #611) fails on the identical test — `2844 passed; 1 failed`. This PR reports `2845 passed; 1 failed`; the +1 is the reflection test added here, which passes in that lane.
- **`src/workflows/gate.rs` does not exist on this branch.** It landed on `main` after this branch was cut, so it arrives only through the PR merge commit.
- **This job passed on the first commit** of this PR (`9747d5e`, 22m12s). Nothing between then and now touched anything the gate can see: the four files here are `frontend/src/lib/connections.ts`, `frontend/src/views/ConnectionsView.tsx`, `frontend/test/unit/connection-route.test.ts`, and one added test in `src/server/ops/connections_read.rs`.

**What the failure actually is**, for whoever owns it: the test asserts that under `supervised` autonomy a pure read of the agent's own workspace is `Reach::Nothing` and must not park. Something now classifies `read_workspace_state` as having an external effect, so it gates. That is the shape of the approvals cluster — #559 ("a Composio read parks, so reading a mailbox costs the operator an approval") is the same complaint one tool over, and #561 covers the parking behaviour.

I have not touched it: it is outside #599 and belongs with the gate/approvals work rather than being smuggled into a connections PR. Happy to pick it up separately if you want it — say the word and I will file or fix it.

Every other check is green, including both `Console` and `Console E2E`, and CodeRabbit has moved to **approved**.

CodeGhost21:
## Manual test report

Rebased onto `f6667d1` (clean, no conflicts) and re-ran everything. **CI does not exercise this fix at all** — `default = ["oauth", "platform-jwt"]`, so no lane builds the `composio` feature and nothing automated touches the route this PR adds. Everything below is a real host and a real browser.

It found a bug in the last round's review fix. Details at the end.

### Rig

Two hosts, both on the harness company:

| | Build | Composio | Manifest |
|---|---|---|---|
| **A** | default | absent (`inBuild: false`) | no `[[connection]]` |
| **B** | `--features composio` → mock backend | live (`credentialSource: static`, open mode) | `[[connection]]` for `x` and `google-calendar` |

Host B's manifest is deliberate: it reproduces the two-row collision. `toolkit_slug("x")` is `x`, not `twitter`, so the host cannot reconcile them and emits **both**:

```json
{ "provider": "x",       "connected": false, "via": [] }
{ "provider": "twitter", "connected": true,  "via": ["composio"] }
```

`google-calendar` reconciles fine (`toolkit_slug` → `googlecalendar`), which isolates the alias as the only collision.

### Scenario A — a host with no route

The issue's evidence table, reproduced verbatim: all eleven `POST …/connections/{id}/start` return `400 provider '<id>' is not enabled on this host`.

| Console | All 11 tiles |
|---|---|
| `main` | **`CONNECT`** — every one 400s |
| this PR | `Not available on this host.` |

### Scenario B — Composio live

| Tile | `main` | this PR |
|---|---|---|
| gmail | connected + `DISCONNECT` (no-op) | connected, "manage in Composio" |
| google-calendar | connected + `DISCONNECT` (no-op) | connected, "manage in Composio" |
| **x** | **not-connected + "Not available on this host"** | **connected, "manage in Composio"** |
| other 8 | `CONNECT` → 400 | `CONNECT` → works |

The `x` row is the one worth pausing on. On `main` a **connected** X account renders as *"Not available on this host"* — the manifest row says `credentialSource: "none"`, so the old code took the `noRoute` branch while Composio held a live connection.

Per-tile, old route vs the one the console now calls:

```
gmail            400 → 200 lk_gmail          notion        400 → 200 lk_notion
slack            400 → 200 lk_slack          google-drive  400 → 200 lk_googledrive
google-calendar  400 → 200 lk_googlecalendar dropbox       400 → 200 lk_dropbox
github           400 → 200 lk_github         stripe        400 → 200 lk_stripe
hubspot          400 → 200 lk_hubspot        linkedin      400 → 200 lk_linkedin
x                400 → 200 lk_twitter
```

### The alias fix, isolated

Same host, console swapped between the two commits of this PR:

| Tile | `bf5d6c0` (before review fix) | `a828690` (now) |
|---|---|---|
| x | not-connected + `CONNECT` | **connected, via Composio** |
| *(other 10)* | *identical* | *identical* |

Only `x` moves — exactly the row with two host records, and nothing else regressed.

### Clicking it

```
calls: POST …/composio/authorize     (not …/connections/stripe/start)
tab:   https://dashboard.composio.dev/link/lk_stripe
toast: Complete Stripe sign-in in the new tab.
```

---

## The blocked-popup fix was wrong, and this is how it surfaced

Pushed `a828690` to revert it.

`window.open` returns `null` whenever `noopener` or `noreferrer` is set — **on success exactly as on a blocked popup**. So the null check added last round fired every time. The first click run showed both at once:

```
tab:   https://dashboard.composio.dev/link/lk_stripe     ← opened fine
toast: Couldn't open the Stripe sign-in tab.             ← told the operator it hadn't
```

Detecting a genuinely blocked popup would mean dropping `noopener`, which is what stops a third-party page we hand an OAuth URL from reaching back through `window.opener`. That is the wrong trade for a nicer error, so the check is gone and the reason it cannot come back is recorded at the call site. `ComposioSection.signIn` opens the same URL the same way and never checked — the two now agree.

I took that review suggestion without testing it, which is what put it in. Flagging it plainly rather than quietly reverting.

## CI

All nine green on the rebase (`cf5e49a`), including `Rust (openhuman, tinycortex)` — the lane that was red from the `read_workspace_state` merge race, now fixed on `main`. `a828690` is building.

Local on the rebase: `cargo test --lib` 1908, `--features openhuman,tinycortex --tests` **2864 passed / 0 failed**, `fmt`, `clippy -D warnings`, three typechecks, `npm test` 380, `build`, `npm run e2e` 96 passed / 8 pre-existing skips.

## Still open for you

**#637 has not landed**, so this would now merge *first* — the opposite of the order I proposed. The composition still holds either way (`connectRoute` reads the tier through `hasCredential` rather than by name, pinned by test), so this is about rebase cost rather than correctness: #637 would replay its `ConnectionsView` edits onto this rather than the reverse. Happy either way — your call, and it needs your re-review regardless since the current `CHANGES_REQUESTED` predates all of the above.

Issue #403, opened 2026-08-05
connections: any member can connect a company account or replace its Composio credential


oxoxDev:
## What happens

Any signed-in member of a company can connect a third-party account to that company, and can replace the company's Composio credential. Neither action requires being an admin or the company's owner.

Observed on a staging tenant: a member signed in with an ordinary email-and-password account — **not** the address that owns the company — completed a Composio OAuth handoff and connected a Gmail account to the company. The connection is now the one the company's agents use.

The reporter's framing is the right one: *anyone who can sign in to the company can connect anything.*

## Why it happens

All four Composio routes are declared with only the `ScopedCompany` extractor (`src/server/ops/composio.rs:61-64`, handlers at `:197`, `:219`, `:230`, `:190`).

`ScopedCompany` (`src/server/ops/scope.rs:62-115`) checks three things: that the company exists, that the principal may **address** that company (`authorize_address`), and that a temporary password has been changed. **It performs no role check.** Its own doc says it keeps "the person" and drops the credential — the user id travels, "never the email or the role". So the handler cannot distinguish an admin from any other member even if it wanted to.

The codebase already has the guard this needs — team routes and the user-admin routes call an admin check before mutating (`src/server/ops/team.rs`, `src/server/users/admin.rs`). The Composio routes simply never adopted it.

## Scope — the connect flow is not the worst of it

| Route | What a non-admin member can do |
|---|---|
| `PUT …/composio/token` | **Replace the company's Composio bearer credential.** A member can point the company's whole tool surface at an account they control |
| `POST …/composio/authorize` | Connect their own third-party account as the company's connection; agents then act through it |
| `GET …/composio/connections` | Read which providers the company has connected |
| `GET …/composio` | Read the company's Composio configuration status |

The token route is the sharper one and was not part of the original report.

Two related notes:

- The account-identity check added for the native OAuth catalog does not apply here. Composio runs its own OAuth and there is no local callback (`src/server/ops/composio.rs`, the `authorize` doc) — so nothing compares the connecting person against the company's owner on this path.
- A machine credential resolves to `GqlAuth::Platform`, which yields no actor at all, so these writes are also unattributed when made that way.

## What the fix should do

1. **Require an admin** for `PUT …/composio/token` and `POST …/composio/authorize`, using the existing admin check rather than a new one. Whether reads should also be restricted is a separate judgement — knowing *that* Gmail is connected is far less sensitive than being able to change it.
[...]

PR #634, opened 2026-08-11
fix(inference): let the company admin set the inference key from the console, not just deploy-time env

### LINKED ISSUE #585: inference: the company's API key can only be set as deploy-time env, not by the admin who owns the company
## The model this serves

A company belongs to one owner — the admin's TinyHumans account. The admin sets **one** API key on the tenant, and everyone the admin invites to that company spends against it: inference, embeddings, tools, capabilities. Membership in the company is what grants access to the key. Same recipe for every tenant we stand up.

OpenCompany does **not** get TinyHumans OAuth as an end-user sign-in path. The console keeps its own login (magic link / password, already built in `Login.tsx`) — who may use this company and which account pays for it are two separate planes.

## What already works

`OPENCOMPANY_INFERENCE_KEY` is checked **first**, ahead of the platform token source, and `provider.rs` already states the intent: *"a per-tenant inference key an operator supplied — not the platform's TinyHumans identity."* Both chat inference and embeddings resolve through the one shared `hosted_endpoint_from_env` pair, so a single key reaches both and a rotation cannot leave one behind.

The mechanism is right. The problem is who can reach it.

## The gap

`hosted_endpoint_from_env` takes an `EnvSource` and reads **only** environment. So the key is an environment variable set at deploy time by whoever operates the cluster — not something a company admin can set on their own tenant. Under the model above, the admin is exactly the person who should be setting it.

## The pattern to copy — it exists twice already

Composio's tenant token is read **live from the secret store**, fingerprinted every cycle by `HarnessPool::ensure`, so *"a console token set/rotate/clear takes effect on the next turn"* — no restart. MCP server credentials work the same way: write-only into the secret store, never returned to the console, only an `authConfigured` boolean comes back.

Inference never got wired into that pattern. This issue is to give it the same treatment.

## The change

1. **Resolution** — teach the hosted-endpoint resolution to read the tenant secret store ahead of environment, keeping the existing order below it. Suggested precedence, most specific first:
   - the tenant's console-set key (secret store)
   - `OPENCOMPANY_INFERENCE_KEY` (env, the operator/deploy-time path we have today)
   - the platform token source (projected token file, then static `TINYHUMANS_API_KEY`)
   - nothing ⇒ `None`, and the runtime keeps its offline echo brain, exactly as now

   Note `hosted_endpoint_from_env`'s signature is env-only, so this is a real signature change with several call sites — worth checking `harness_inference_from_env`, the embeddings resolver, and the media backend resolver, which share the shape.

2. **Freshness** — fingerprint the resolved credential the way the Composio token is fingerprinted, so a key set or rotated in the console reaches agents on the next cycle without a restart.

3. **Console** — an admin-only screen to set, rotate and clear the key. Write-only: the key is never returned, and the read shape carries only whether one is configured. Same discipline as MCP credentials.

4. **State the consequence in the UI.** The key is the company's wallet and every member spends against it. The screen should say that plainly, because it is the thing an admin most needs to understand before pasting.

## Known limits, recorded deliberately

- **No per-member attribution.** Spend arrives as one account, so we cannot say which member burned what. Per-agent `budget_usd_daily` caps still work; per-*person* accounting does not exist and is not in scope here. If it is wanted later it is a second key or an attribution header, not a redesign.
- **Removing someone from the roster stops future access**, but nothing already spent is separable.
- **The key sits in the workload** and carries whatever privileges a TinyHumans API key carries. Worth confirming that scope is inference/tools rather than the admin's whole account surface before this is on every tenant.

## Relationship to the attestation work

This removes attestation from the critical path: a pod carrying a key does not need to prove who it is. `opencompany-microservice#12` stays open and harmless; it is no longer blocking anything.

## Acceptance

- An admin can set, rotate and clear the company's inference key from the console.
- The key is never returned by any read route and never appears in logs or errors.
- A key set in the console reaches agents on the next cycle with no restart.
- With no console key set, existing env-based behaviour is byte-for-byte unchanged.
- With no credential at all, the runtime still falls back to the offline echo brain rather than failing.

## Negative control

Delete the store-ahead-of-env arm, confirm the console-set-key test fails, restore.

CodeGhost21:
Closes #585.

> **Rebased on current `main`. This PR is smaller than it was**: `main` landed
> #290 (in-place runtime rebuild) while this was open, which solved one of the
> two defects I'd fixed here. That commit is dropped — see *Superseded* below.

### Related PRs — this one touches the same surface as two others

| PR | Overlap with this one |
| --- | --- |
| **#636** — `fix(ops): report and probe the inference endpoint requests actually reach` | Rewrites `effective_status` in the same file. Its two-resolve split keeps `keyConfigured` reporting *tenant* config; this PR adds a second legitimate writer to that field. The route-level test pinning the two apart lives there (`a_platform_token_never_reads_as_a_console_set_key`); the resolver-level one lives here. |
| **#637** — `feat(connections): broker providers through the company's own TinyHumans credential` | Adds a *separate* company credential under `tinyhumans/key`, adjacent to this card. Both accept a key and one accepts a TinyHumans key. Each card now carries a one-line subtitle naming which it is — "the key your agents think with — not the company account key" here, "for connecting providers — not the model key" there. |

Merge order does not matter; each carries its own half.

## RCA — the issue's stated cause is not the defect

`hosted_endpoint_from_env` does not exist, there is no embeddings resolver in
this repo, and the tenant secret store is **already** read ahead of environment.
Per-tenant BYOK (#56) shipped `company::inference::resolve_effective`
(runtime override > manifest `[inference]` > env default), `TenantProvider`
re-resolving it every turn, and write-only `GET`/`PUT`/`DELETE …/inference`
whose read shape carries only `keyConfigured`. #403 later made the form
admin-only. Most of what the issue asks for already exists.

The actual defect is one screen:

**The console could not type the key for the managed provider.**
`InferenceSection.tsx` rendered the key input only for `provider !== "managed"`,
and `save()` implemented a managed save as a *revert* (`DELETE …/inference`).
The admin's TinyHumans key **is** the managed provider's key, and
`resolve_endpoint` has always preferred a stored key over `EnvDefault` for it.
The backend was ready; the console offered nowhere to type one. Those two lines
are the whole reason the key could only arrive as deploy-time env.

## The change

- **`test(inference)`** — pin the resolution that already held: a console-set key
  wins over the platform credential on `managed`, setting only a key keeps the
  platform endpoint (so paying for your own agents does not move you off the
  managed brain), and clearing it falls back to env rather than 401ing.
- **`fix(console)`** — offer the write-only key field for every provider but
  Ollama; make a managed save carrying a key a real `PUT` override; add an
  explicit **Remove key** action, completing set / rotate / clear; state the
  consequence where the admin pastes.

## This retires #265's guard — deliberately, and the invariant survives

`main` currently **refuses** a managed save that carries a key
(`inference-key-conflict`, Save disabled), because a managed save was only ever
a revert and a revert cannot carry a credential — so a typed key would have been
dropped while the toast claimed success.

#265's invariant is *"a save that reports success must never have discarded what
the operator typed."* That invariant is **preserved**, and now upheld directly:
managed+key stores the key instead of refusing it. Nothing is discarded, so
there is nothing to refuse. The guard was a workaround for the revert
implementation, not a statement about credentials.

`frontend/test/e2e/inference.spec.ts` is rewritten accordingly — same invariant,
new mechanism, and it is a real gate: the `Console E2E` job (#428) runs it, and
it passes on this branch. Its header carried a stale "not part of CI" note that
predated that job; corrected here. The rewritten test asserts that a key typed under a BYOK provider survives the switch to managed,
lands server-side (`keyConfigured` flips), leaves `provider` on `managed`, and
can then be cleared again. If the maintainers would rather keep managed as
platform-credential-only, this is the commit to reject, and #585 needs a
different design — flagging it rather than burying it.

## Superseded by #290 — dropped from this PR

My first revision also fixed the boot-frozen brain gate: `RuntimeBuilder::build`
decided `configured` once, a runtime is built once and cached in
`CompanyRegistry`, and `HarnessPool::ensure` fingerprints MCP / overlay /
capability / Composio / skills but not inference — so on a tenant with no
deploy-time credential a console-set key never reached agents without a restart.

`main` now solves exactly that with `RuntimeRebuilder` (#290), reported honestly
as `restartRequired` (#266). That is the better fix — it rebuilds the whole
runtime, including the workflow runner, which my per-cycle brain wrapper did
not. My `DeferredBrain` commit is dropped in full; nothing of it remains.

## Acceptance

| Criterion | Status |
| --- | --- |
| Admin can set, rotate and clear the key from the console | ✅ `managed_key_can_be_set_rotated_and_cleared`; admin-only via `AdminScopedCompany` (#403) |
| Key never returned by a read route, never in logs or errors | ✅ existing `key_never_leaks_across_any_response` + new assertions across all three tokens |
| A console key reaches agents on the next cycle with no restart | ✅ via #290's in-place rebuild — not this PR |
| With no console key, env behaviour unchanged | ✅ `resolve_effective` order untouched; the resolution test pins it |
| With no credential at all, the offline echo brain still serves | ✅ unchanged — this PR does not touch brain selection |

**Negative control** (as the issue asks): deleting the store-ahead-of-env arm in
`resolve_endpoint` fails `console_key_beats_env_for_the_managed_provider`;
restored, it passes.

## Commands run

- `cargo test` — 1856 passed (lib), all 6 targets green
- `cargo +1.96.1 test --features openhuman --lib` — green
- `npm run typecheck` (frontend) — clean
- `npm run typecheck:e2e` — clean; the rewritten `inference.spec.ts` is executed
  by the `Console E2E` job on this branch and passes

## Behaviour changes reviewers should weigh

- A managed save carrying a key now writes a runtime override instead of
  reverting. Managed with *no* key — none typed, none stored — is still a revert.
- **Remove key** echoes a base URL back only for `ollama` / `openai_compatible`.
  Persisting the *displayed* URL for `managed` would pin the company to the
  default constant and silently override `OPENCOMPANY_INFERENCE_URL`.

## Pre-existing breakage found (not from this branch — confirmed on a clean tree)

1. The `openhuman` feature does not build on `stable` (1.93.0): `libsqlite3-sys
   0.38.1`, pinned in `Cargo.lock`, uses unstable `cfg_select!` in its build
   script. Validated on 1.96.1 instead. Probably worth its own issue.
2. `cargo clippy --all-targets -- -D warnings` fails on
   `src/server/test_support.rs:8` (`duplicated_attributes`).
3. `src/harness/composio.rs` emits dead-code warnings under `openhuman` without
   `composio`.

## Known limits (recorded, per the issue)

No per-member attribution — spend arrives as one account. Per-agent
`budget_usd_daily` caps still work. Removing someone stops future access but
does not separate what they already spent. Both are stated in the UI copy.

## Out of scope, worth a follow-up

The form never seeds from the fetched status — it always opens on "Managed", so
a company on OpenRouter that presses Save gets reverted. Pre-existing.
`removeKey` deliberately sends `status.provider` rather than the form's value so
this branch does not make it worse.


oxoxDev (changes requested):
**2 major, 1 question.** Both majors are cross-PR — nothing in this diff is wrong.

Two things in here are exemplary and worth saying plainly. **The RCA that disproves the issue** — `hosted_endpoint_from_env` does not exist, there is no embeddings resolver, the tenant store is already read ahead of env, and the real defect is two lines of console that rendered the key input only for `provider !== "managed"` — is worth more than the fix. And **the #265 analysis** separates the invariant ("a save that reports success must never discard what the operator typed") from the mechanism that enforced it (a refusal that existed only because a managed save was a revert). Preserving the invariant while removing the workaround, rewriting the E2E that actually gates it, and then explicitly offering that commit for rejection if maintainers want managed to stay platform-credential-only — that is how a contested change should arrive.

---

### Major 1 — collides with #636 on `keyConfigured`, the field #636 exists to protect

#636 (`fix(ops): report and probe the inference endpoint requests actually reach`, also yours) rewrites `effective_status` in this same file, and its entire design is a two-resolve split whose purpose is keeping `keyConfigured` reflecting *tenant* config only — never the platform token. This PR adds a legitimate new writer to that field: an admin setting the key from the console, including on `managed`.

After both land, `keyConfigured` has two valid sources and one that must never feed it, and nothing asserts the three stay distinguishable. I raised this on #636 too. Whichever goes second should carry a test that a console-set key and an injected platform default produce different `keyConfigured` values on the same company.

### Major 2 — with #637, an admin now has two places to paste a TinyHumans key, and crossing them is the leak #637 refuses

This PR stores the admin's key under `inference/key` and states, correctly for `managed`, that "the admin's TinyHumans key **is** the managed provider's key". #637 (`feat(connections): broker providers through the company's own TinyHumans credential`) creates a *separate* `tinyhumans/key`, and its module docs argue at length that reusing `inference/key` for that purpose would be a cross-vendor credential leak — because `inference/key` is provider-scoped and may hold an OpenRouter `sk-or-…` or a raw BYOK token.

Both statements are true, and #637 says why: they coincide only when the declared provider is `managed`. But the result is two adjacent console cards that both accept a TinyHumans key, with different storage, different scope, and a documented leak if an operator crosses them — while the reasoning that makes the distinction safe lives in a module doc comment neither of them will read. An admin who pastes an OpenRouter key into the company-credential card has done the exact thing #637 was written to prevent.

Neither PR references the other. At minimum the copy on both fields should say what the key is for and what it must not be; better would be for one of you to own the pair and make the two cards visibly different in kind.

### Question — is `provider !== "ollama"` the right shape for offering the key field?

Offering the write-only field for every provider but Ollama is a reasonable default, but it is a deny-list: a future local or credential-less provider inherits a key input nobody wants. Is there a capability on the provider descriptor that could carry this instead, so the answer comes from the provider rather than from a name comparison in the view?

CodeGhost21:
Stacking note (raised in review on #637): #637 adds a **company TinyHumans credential** card to the Connections screen, immediately next to this PR's inference-key card. Both render as "configured / not configured", and the two keys are different in kind — `inference/key` is provider-scoped (an OpenRouter `sk-or-…`, or your own endpoint's key), while `tinyhumans/key` is the company's platform identity. Presenting one where the other is expected is a cross-vendor credential leak, which is why they are separate slots.

#637's card carries its half of the disambiguation already: the field is "TinyHumans account key", and its notice says plainly that it is *not* the model-provider key on the Inference card. I deliberately left `InferenceSection.tsx` alone there to avoid a third overlapping edit on this file.

So whichever of these two lands second should add the mirror-image line here — something to the effect that this is the key for the model provider you chose, not the company's TinyHumans account key used to connect providers. No code coupling, just copy.

oxoxDev:
Re-checked at the new head — narrowing my review rather than re-blocking.

**Closed:** the `keyConfigured` invariant is pinned from #636's side, and #636 now states the order (it first, this rebases onto it). And you went past the question I asked: `acceptsKey` / `requiresBaseUrl` on the provider descriptor is exactly right — a new provider inheriting a key field simply by not being named in a deny-list was the failure mode, and asking the descriptor removes the category.

**Still open — Major 2, the one that is not about this diff.** Between this and #637 an admin gets two console cards that both accept a TinyHumans key: `inference/key` here, `tinyhumans/key` there, different scope and storage, and #637's own docs argue that crossing them is a cross-vendor leak. Neither PR's copy tells the admin which is which. That is the item keeping the changes-requested standing, and it needs one of you to own the pair rather than a change in either PR alone.

CodeGhost21:
Thanks — all three addressed. Both majors were cross-PR, as you said, so the fixes are split across the three and each carries its own half.

### Major 1 — `keyConfigured` with two valid sources

Pinned at both layers, on one company with a platform default injected:

- **here** — `a_console_key_and_the_platform_default_are_distinguishable` in `company::inference`: with nothing tenant-scoped stored, the platform-aware resolve is credentialled while the tenant-only resolve the read route uses is not; a console-set key then flips the tenant-only answer and becomes the bearer agents present.
- **#636** — the route-level half, which needs the `effective_status_with` seam this branch does not have. You'd already asked there, and `a_platform_token_never_reads_as_a_console_set_key` landed on that branch while I was writing the same test. I'd written a duplicate against a runtime-source config instead of a manifest one; it asserted nothing the existing test doesn't, so I dropped it rather than push noise. Negative control on the surviving one: resolving `decl` with the platform default instead of `None` fails it.

### Major 2 — two cards, two TinyHumans-shaped keys

You're right that the reasoning was in a module doc neither operator reads. Two changes:

- The descriptor now carries a per-provider `keyKind`, so the field says which vendor's key belongs in it — "a TinyHumans API key" on `managed`, "an OpenRouter key (`sk-or-…`)", "whatever key the endpoint below expects" — plus the converse: it is scoped to the provider selected above, and another vendor's key fails at the first turn.
- Each card now has a one-line subtitle beside its heading, so they differ in kind at a glance: **"the key your agents think with — not the company account key"** here, mirroring **"for connecting providers — not the model key"** on #637's card (already on that branch).

I own all three, so I've taken the pair. Both PR bodies now cross-reference; merge order does not matter.

### Question — `provider !== "ollama"`

Agreed, it was a deny-list over a closed set. Replaced with a `PROVIDERS` descriptor carrying `acceptsKey` and `requiresBaseUrl`; the view asks it instead of comparing names, and the same lookup replaced `provider === "ollama" || provider === "openai_compatible"` in the base-URL input and in `removeKey`. A new provider declares what it needs where it is defined, and inherits nothing by omission.

The descriptor also absorbed `PROVIDER_LABELS` and `presetFor`, which were three parallel lists keyed by the same union.

oxoxDev (approved):
**0 major.** Approving. All three closed, and both majors were cross-PR so the fixes are correctly split across branches.

- **Major 1** — pinned at both layers: `a_console_key_and_the_platform_default_are_distinguishable` here, and the route-level half on #636 where the `effective_status_with` seam lives. Dropping your duplicate rather than pushing a test that asserted nothing new is the right instinct; a second test that cannot fail is worse than none because it reads as coverage.
- **Major 2** — `keyKind` on the descriptor is a better answer than the copy fix I suggested. Naming the vendor whose key belongs in the field ("a TinyHumans API key" on `managed`, "an OpenRouter key `sk-or-…`") plus the converse — that it is scoped to the provider selected above and another vendor's key fails at the first turn — puts the reasoning where the paste happens instead of in a module doc.
- **Question** — `acceptsKey` / `requiresBaseUrl` on the descriptor removes the category rather than the instance. A new provider can no longer inherit a key field by not being named in a deny-list.

CodeGhost21:
## Manual + E2E verification against a running host

Ran the frontend and backend together locally and drove the change end to end, on top of the current `main` merge.

### Backend — the route, by hand

Signed in as an admin on a real host and exercised the full lifecycle the console performs:

| Step | Result |
| --- | --- |
| baseline | `source: managed`, `keyConfigured: false` |
| `PUT {provider: managed, key}` | `keyConfigured: true`, `source: runtime`, **`baseUrl` unchanged** on the platform endpoint |
| rotate (second key) | still `true`; neither old nor new token in the response |
| `PUT {provider: managed, key: ""}` | back to `false` |

Setting a key does not move the tenant off the managed brain — that is the property the whole issue turns on, and it holds on a live host rather than only in a unit test. Grepped every `PUT` and `GET` body for all three tokens: **0 occurrences**.

### Frontend — real browser, real host

`inference.spec.ts`, the spec this PR rewrote:

```
✓ a key typed for a BYOK provider is not discarded by switching to managed (878ms)
✓ a key typed for a BYOK provider does reach the host on save (705ms)
```

Full console suite on the same host: **96 passed, 8 skipped, 0 failed**. The 8 skips are the capability-gated live-brain specs (#467), which need a harness build this lane does not have.

### One thing worth flagging for anyone running the suite locally

The first three attempts failed with `#inference-key` not found, and the page snapshot showed *this PR's own removed copy* — "Managed runs on the platform credential, so there is no key to paste here."

The cause was not the code. Playwright's `reuseExistingServer` had silently attached to a host another worktree was running on `:8080`:

```
OPENCOMPANY_CONSOLE_DIR=/…/oc-600-wt/frontend/dist
```

So the browser was driving a console built from a different branch and reporting it as this branch's result. Anyone with several worktrees open can hit this, and it fails in the most misleading way possible — a green-looking harness testing someone else's code. Pinning to a private port avoids it:

```bash
./target/debug/opencompany serve --bind 127.0.0.1:8234 --company companies/e2e_harness
PW_BASE_URL=http://127.0.0.1:8234 PW_STORAGE_STATE=… npx playwright test
```

Might be worth a follow-up to have `host.sh` refuse a reused server whose `OPENCOMPANY_CONSOLE_DIR` is not this checkout's `dist` — happy to file it.

### CI

All lanes green, including `Rust (openhuman, tinycortex)`, which was red earlier from `main`'s `workflows::gate` breakage (since fixed by 2eb19f5 and picked up in the merge here). `Console E2E (live brain)` is the only one outstanding.

PR #636, opened 2026-08-11
fix(ops): report and probe the inference endpoint requests actually reach (#597)

### LINKED ISSUE #597: console: the Inference card reports the hardcoded managed URL, ignoring OPENCOMPANY_INFERENCE_URL — every staging tenant reads "prod"
## Summary

Settings → Connections → Inference shows `https://api.tinyhumans.ai/openai/v1` on a staging tenant that is routing to `https://staging-api.tinyhumans.ai/openai/v1`. The card reports a URL the company is not using, and no amount of environment configuration changes what it prints.

Observed live on `smoke1.staging.opencompany.work` after `OPENCOMPANY_INFERENCE_URL` was set on the tenant:

```json
{"provider":"managed","baseUrl":"https://api.tinyhumans.ai/openai/v1","source":"managed","keyConfigured":false}
```

## Root cause — two resolvers, only one env-aware

**Routing** (`src/company/inference.rs:180`, `resolve_endpoint`) treats `managed` as the platform brain and prefers the env default:

```rust
let base_url = base_url_override
    .map(str::to_string)
    .or_else(|| env_default.map(|e| e.base_url.clone()))
    .unwrap_or_else(|| MANAGED_BASE_URL.to_string());
```

`env_default` originates in `hosted_endpoint_from_env` (`src/harness/provider.rs:136`), which reads `OPENCOMPANY_INFERENCE_URL`. So outbound calls do follow the env.

**Display** (`src/server/ops/inference.rs:258`, `effective_status`) deliberately does not:

```rust
let decl = resolve_effective(runtime.id(), &manifest, None, runtime.secrets().as_ref())
```

The `None` is the `env_default` slot, and the comment states the intent: *"The ops layer resolves tenant config only (no env default), so a company with nothing configured reports the managed default rather than a synthesized env decl."* The `None` arm then renders `inference::MANAGED_BASE_URL` — a hardcoded `https://api.tinyhumans.ai/openai/v1` (`src/company/inference.rs:49`).

So the two answers diverge exactly when a tenant is pointed anywhere other than production, which is every non-prod deployment.

## Why this is worth fixing rather than documenting

The card is the only place an operator can check where their brain is pointed, and it is confidently wrong. In one session it caused two separate misdiagnoses: a prod URL on the card was read as evidence that routing was broken (it was not), and it was independently reported as the Composio sign-in URL being misconfigured (Composio was correct throughout — its `backendUrl` reads `https://staging-api.tinyhumans.ai`). A field that cannot be trusted as evidence is worse than an absent one, because people reason from it.

It also silently undercuts `OPENCOMPANY_INFERENCE_URL` as an operational lever: an operator sets it, reloads, sees no change, and reasonably concludes the patch did not take.

## Suggested direction (author's call)

The existing intent — "don't synthesize a tenant decl out of env" — is sound and should survive. Two ways to keep it while telling the truth:

1. Pass the env default into `effective_status` for **display only**, and keep `source: "managed"`. The operator sees the URL actually in use; nothing about tenant config resolution changes.
2. Keep the current resolution and add an explicit field (e.g. `platformBaseUrl`, or a `baseUrlSource` of `platform-env` vs `built-in-default`), so the card can distinguish "this is where you are pointed" from "this is the built-in default".

Option 1 is smaller; option 2 is more honest about which of the two a reader is looking at. Either is fine — what should not survive is a rendered URL that no configuration can influence.

Worth checking the same shape on the web-search surface while in here: `search_backend_from_env` has the equivalent hardcoded `https://api.tinyhumans.ai` default.

## Reproduce

1. On a staging tenant, set `OPENCOMPANY_INFERENCE_URL=https://staging-api.tinyhumans.ai/openai/v1` (the `patch-tenant-env` workflow does this).
2. Wait for the rollout, then `GET /api/v1/company/inference`.
3. `baseUrl` still reports `https://api.tinyhumans.ai/openai/v1`.

Expected: the URL the company is actually configured to call, or an explicit signal that the value shown is a built-in default rather than the effective endpoint.

## Related

- `tinyhumansai/opencompany-microservice#12` fixes the *injection* half (deriving `OPENCOMPANY_INFERENCE_URL` and `OPENCOMPANY_SEARCH_BACKEND_URL` from the attestation audience so they cannot drift). This issue is independent: once that lands, the card will still print the hardcoded prod URL.

CodeGhost21:
Fixes #597.

## What was wrong

Two resolvers, only one env-aware. **Routing** (`resolve_endpoint`, `src/company/inference.rs:180`) treats `managed` as the platform brain and inherits `env_default.base_url`, which comes from `hosted_endpoint_from_env` reading `OPENCOMPANY_INFERENCE_URL` — outbound calls followed the env correctly all along. **Display** (`effective_status`, `src/server/ops/inference.rs`) passed `None` into that slot, so resolution fell through to the hardcoded `MANAGED_BASE_URL`. The card contradicted the traffic, and no configuration could change what it printed.

Two things the issue does not cover, both fixed here:

**It is not only the unconfigured arm.** The report reproduced `source: "managed"` and points at the `None` arm's hardcoded constant. But `resolve_endpoint` falls back to the same constant for *any* `managed` config that names no `base_url` of its own — so a tenant with `[inference] provider = "managed"` in `company.toml`, or a console override set to managed, printed the production URL too, under a `manifest`/`runtime` badge. Fixing only the `None` arm would have left most of the surface wrong.

**`POST /inference/test` had the same defect** (second commit). It resolved with `None` as well, so for a `managed` config the probe got the hardcoded production base URL *and* `Credential::None` — it fired at the wrong host with no bearer. A staging tenant whose routing is fine was told its provider is unreachable. That is a wrong verdict on the same card, not just a wrong label. It is a separate commit for reviewability, **not as a ship option** — see *Merge order* below.

## Approach

The issue's option 1, but applied literally it breaks three other fields: passing the env default straight into `resolve_effective` makes step 3 synthesize a decl, which moves `source` from `managed` to `default`, flips `keyConfigured` to true (the console then renders a "key set" badge and an "API key (leave blank to keep)" hint for a key the operator never set), and feeds `decl.is_some()` into `restart_pending`, telling operators to restart for config that does not exist. The `None` was protecting something real — it was just protecting the URL too.

So the resolution is split. Tenant config resolves with no env default and stays the only input to `source`, `keyConfigured` and `restartRequired`; the base URL alone resolves with the platform default in place. That default is derived from `harness_inference_from_env` — the same function the harness routes on — rather than from reading `OPENCOMPANY_INFERENCE_URL` directly, because a URL with no credential resolves to `None` there. The card therefore never advertises an endpoint that would route nowhere.

Option 2 (`platformBaseUrl` / `baseUrlSource`) was considered and skipped: the existing `source` field already separates "this is your tenant config" from "this is the platform default", so the extra field would restate it.

## On the web-search aside

`search_backend_from_env`'s hardcoded default is a *routing* default that `OPENCOMPANY_SEARCH_BACKEND_URL` overrides correctly, the same way inference routing already did. No ops surface renders it — `capabilities.rs:187` reports only a bool. There is no display bug to fix there, so nothing changed.

## Merge order and the `keyConfigured` invariant

**Do not partially merge this.** The probe commit is split out so the two changes can be read apart, not so the second can be dropped. Dropping it would leave the card worse than today rather than better: `baseUrl` would render the correct staging endpoint while `POST /inference/test` still fires at the hardcoded production host with `Credential::None` and reports the provider unreachable. Two adjacent fields contradicting each other is harder for an operator to reason about than both being wrong in the same direction.

**Order against #634: this PR first, then #634 rebases on it.** They are textually adjacent but semantically independent — #634 touches `src/server/ops/inference.rs` only to extend the module doc header and add `managed_key_can_be_set_rotated_and_cleared`; the backend already preferred a stored key over `EnvDefault` on `managed`, and the `inference-key-conflict` guard it retires is frontend-only. So neither changes behaviour the other depends on, and the conflict is a mechanical one in the test module.

The invariant is worth more than the ordering, so it is pinned here rather than deferred to whichever lands second: `a_platform_token_never_reads_as_a_console_set_key` drives one company with one injected platform default to both answers — no tenant key reads `keyConfigured: false` while the platform token does the outbound work, a console-set key reads `true`, and the resolved endpoint stays the platform's either way. #634's new source then arrives against a test that already separates the three.

## API change

`baseUrl` on `GET /api/v1/company/inference` now reports the platform endpoint when one is injected. Every other field is byte-identical. No frontend change needed — the console already renders whatever `baseUrl` holds.

## Tests

Five new tests: the unconfigured arm follows the platform URL while `source`/`keyConfigured`/`restartRequired` stay put; a `managed` manifest inherits it too; an explicit tenant `base_url` still outranks it; a non-managed provider ignores it; and `platform_default` follows `OPENCOMPANY_INFERENCE_URL` but resolves to nothing without a credential. Plus one pinning the probe's 409 gate. I verified they are not vacuous by reverting the fix — the two fix-specific ones fail, the over-reach guards stay green.

## Commands run

```
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings
cargo test                                    # 1813 passed
```

The `openhuman` lane cannot build on the pinned `stable` (1.93.0): `libsqlite3-sys 0.38.1`'s build script uses unstable `cfg_select!`. That is pre-existing and unrelated — it dies in a third-party build script before `opencompany` compiles. Verified on the installed 1.96.1 instead, which is the only way to exercise `platform_default` and the probe:

```
cargo +1.96.1 clippy -p opencompany --features openhuman --all-targets --no-deps -- -D warnings
cargo +1.96.1 test --features openhuman --lib server::ops::inference   # 16 passed
```

(`--no-deps` because vendored `tinyagents` trips `large_enum_variant` on 1.96 — also pre-existing, also unrelated.)

### DIFF
--- a/src/server/ops/inference.rs
+++ b/src/server/ops/inference.rs
@@ -497,11 +497,14 @@ async fn test_config(company: ScopedCompany) -> Response {
         Ok(m) => m,
         Err(err) => return err.into_response(),
     };
-    let decl =
-        match resolve_effective(runtime.id(), &manifest, None, runtime.secrets().as_ref()).await {
-            Ok(d) => d,
-            Err(err) => return ApiError(err).into_response(),
-        };
+    let secrets = runtime.secrets().as_ref();
+    // Gate on *tenant* config, with no platform default: the button probes what
+    // the company configured, so "nothing configured" stays a 409 instead of
+    // quietly probing the platform brain on the operator's behalf.
+    let decl = match resolve_effective(runtime.id(), &manifest, None, secrets).await {
+        Ok(d) => d,
+        Err(err) => return ApiError(err).into_response(),
+    };
     match decl {
         None => (
             StatusCode::CONFLICT,
@@ -512,23 +515,45 @@ async fn test_config(company: ScopedCompany) -> Response {
             })),
         )
             .into_response(),
-        Some(decl) => match crate::harness::provider::probe(&decl).await {
-            Ok(()) => Json(serde_json::json!({
-                "ok": true,
-                "provider": decl.provider,
-                "note": "Reached the provider and got a reply.",
-            }))
-            .into_response(),
-            Err(err) => (
-                StatusCode::BAD_GATEWAY,
-                Json(serde_json::json!({
-                    "ok": false,
-                    "error": format!("Inference probe failed: {err}"),
-                    "code": "probe_failed",
-                })),
-            )
+        Some(tenant) => {
+            // Probe with the platform default in place. A `managed` config
+            // inherits both the platform endpoint and its credential, so
+            // resolving without it aimed the probe at the built-in production
+            // URL carrying no bearer at all — a staging tenant whose routing is
+            // perfectly fine would be told its provider is unreachable, on the
+            // same card that was already misreporting the URL (issue #597).
+            let decl = match platform_default(&crate::app::config::ProcessEnv) {
+                None => tenant,
+                Some(platform) => {
+                    match resolve_effective(runtime.id(), &manifest, Some(&platform), secrets).await
+                    {
+                        // Adding the platform default can only add a source,
+                        // never remove the one that just resolved — fall back to
+                        // it rather than inventing an error path for a case that
+                        // cannot happen.
+                        Ok(d) => d.unwrap_or(tenant),
+                        Err(err) => return ApiError(err).into_response(),
+                    }
+                }
+            };
+            match crate::harness::provider::probe(&decl).await {
+                Ok(()) => Json(serde_json::json!({
+                    "ok": true,
+                    "provider": decl.provider,
+                    "note": "Reached the provider and got a reply.",
+                }))
                 .into_response(),
-        },
+                Err(err) => (
+                    StatusCode::BAD_GATEWAY,
+                    Json(serde_json::json!({
+                        "ok": false,
+                        "error": format!("Inference probe failed: {err}"),
+                        "code": "probe_failed",
+                    })),
+                )
+                    .into_response(),
+            }
+        }
     }
 }
 
@@ -758,6 +783,22 @@ mod tests {
             .unwrap();
         assert_eq!(dto.base_url, inference::OPENROUTER_BASE_URL);
         assert_eq!(dto.source, "manifest");
+    }
+
+    /// The probe's gate stays keyed on *tenant* config. Pointing a deployment at
+    /// a platform endpoint gives the probe somewhere real to aim, but it must not
+    /// turn "nothing configured" into a live probe of the platform brain — that
+    /// 409 is the honest answer to "test my provider" from a company that has
+    /// not named one.
+    #[cfg(feature = "openhuman")]
+    #[tokio::test]
+    async fn probing_an_unconfigured_company_stays_not_configured() {
+        let home_dir = home();
+        let state = state_with_company(home_dir.path()).await;
+
+        let (status, body, _) = send(&state, "POST", "/api/v1/company/inference/test", None).await;
+        assert_eq!(status, StatusCode::CONFLICT);
+        assert_eq!(body["code"], "not_configured");
     }
 
     #[cfg(feature = "openhuman")]
--- a/src/server/ops/inference.rs
+++ b/src/server/ops/inference.rs
@@ -27,9 +27,10 @@ use axum::{Json, response::Response};
 use serde::{Deserialize, Serialize};
 
 use crate::AppState;
+use crate::app::config::EnvSource;
 use crate::company::Inference;
 use crate::company::inference::{
-    self, InferenceSource, RuntimeInference, clear_runtime_config, resolve_effective,
+    self, EnvDefault, InferenceSource, RuntimeInference, clear_runtime_config, resolve_effective,
     save_runtime_config, store_key, validate_runtime,
 };
 use crate::company::runtime::CompanyRuntime;
@@ -79,7 +80,11 @@ struct InferenceStatusDto {
     provider: String,
     /// The stable telemetry slug (`managed` / `openrouter` / `byok` / `ollama`).
     slug: String,
-    /// Resolved OpenAI-compatible base URL.
+    /// Resolved OpenAI-compatible base URL — the endpoint requests actually
+    /// travel to, including the platform default this deployment was pointed at
+    /// with `OPENCOMPANY_INFERENCE_URL`. A `managed` provider inherits that
+    /// endpoint, so resolving this without the platform default printed the
+    /// built-in production constant on every staging tenant (issue #597).
     base_url: String,
     /// Abstract-tier → concrete model id.
     models: BTreeMap<String, String>,
@@ -253,14 +258,85 @@ pub(crate) async fn runner_gap_for(runtime: &CompanyRuntime) -> RunnerGap {
     RunnerGap::NotWired
 }
 
-/// Resolves the effective status DTO. The ops layer resolves *tenant* config
-/// only (no env default), so a company with nothing configured reports the
-/// managed default rather than a synthesized env decl.
+/// This deployment's platform-injected managed default — the same
+/// `(base_url, credential)` pair the harness routes on
+/// ([`harness_inference_from_env`](crate::harness::provider::harness_inference_from_env)),
+/// read from the environment and never from a tenant secret.
+///
+/// `None` off the `openhuman` feature, and `None` when no managed credential
+/// resolves — which is exactly when the harness resolves no env default either.
+/// Deriving it from that one function rather than reading
+/// `OPENCOMPANY_INFERENCE_URL` directly is what keeps display and routing
+/// honest in both directions: a bare URL with no credential routes nowhere, so
+/// the card must not advertise it as the endpoint in use.
+fn platform_default(env: &dyn EnvSource) -> Option<EnvDefault> {
+    #[cfg(feature = "openhuman")]
+    {
+        crate::harness::provider::harness_inference_from_env(env).map(|(config, _)| EnvDefault {
+            base_url: config.base_url,
+            credential: config.credential,
+        })
+    }
+    #[cfg(not(feature = "openhuman"))]
+    {
+        let _ = env;
+        None
+    }
+}
+
+/// Resolves the effective status DTO against the real process environment.
 async fn effective_status(runtime: &CompanyRuntime) -> Result<InferenceStatusDto, ApiError> {
+    effective_status_with(
+        runtime,
+        platform_default(&crate::app::config::ProcessEnv).as_ref(),
+    )
+    .await
+}
+
+/// [`effective_status`] against an explicit platform default, so the resolution
+/// is testable without touching the process environment.
+///
+/// Two resolves, deliberately, because the card answers two different questions
+/// (issue #597):
+///
+/// - **What did the tenant configure?** — resolved with no env default, and the
+///   only thing `source`, `keyConfigured` and `restartRequired` may see. A
+///   platform endpoint is not tenant config: reporting it as `source: "default"`
+///   would move the badge, reporting the platform token as `keyConfigured` would
+///   tell an operator a key they never set is stored (and that blanking the
+///   field "keeps" it), and feeding it to [`restart_pending`] would strand a
+///   company behind a restart that changes nothing. That is the intent the old
+///   `None` argument was protecting, and it survives here intact.
+/// - **Where do requests actually go?** — resolved *with* the platform default,
+///   and used for `baseUrl` alone. `managed` inherits the platform endpoint, so
+///   without it every non-production tenant rendered the built-in production
+///   constant: both the `None` arm below and any tenant whose own config names
+///   `managed` without its own `base_url`.
+async fn effective_status_with(
+    runtime: &CompanyRuntime,
+    platform: Option<&EnvDefault>,
+) -> Result<InferenceStatusDto, ApiError> {
     let manifest = manifest_inference(runtime).await?;
-    let decl = resolve_effective(runtime.id(), &manifest, None, runtime.secrets().as_ref())
+    let secrets = runtime.secrets().as_ref();
+    let decl = resolve_effective(runtime.id(), &manifest, None, secrets)
         .await
         .map_err(ApiError)?;
+    let base_url = match platform {
+        // Re-resolve with the platform default in place rather than
+        // reconstructing `resolve_endpoint`'s precedence here — the tenant's own
+        // `base_url` still outranks it, and only the `managed` kind inherits it.
+        Some(platform) => resolve_effective(runtime.id(), &manifest, Some(platform), secrets)
+            .await
+            .map_err(ApiError)?
+            .map_or_else(|| inference::MANAGED_BASE_URL.to_string(), |d| d.base_url),
+        // No platform endpoint on this deployment: nothing to inherit, so the
+        // tenant resolve already holds the whole answer and the second read is
+        // skipped.
+        None => decl.as_ref().map_or_else(
+            || inference::MANAGED_BASE_URL.to_string(),
+            |d| d.base_url.clone(),
+        ),
+    };
     // What the company actually booted onto, not what the config implies.
     let cognition = runtime.cognition();
     let restart_required = restart_pending(runtime, decl.is_some());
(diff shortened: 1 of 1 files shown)


oxoxDev (changes requested):
**2 major, 1 minor, 1 question.** The code itself is sound — both majors are about sequencing and packaging, not about what the diff does. Happy to flip to approve once they are answered.

Verified the premise rather than taking the description's word for it. `resolve_endpoint` (`src/company/inference.rs:183-184`) does fall through `env_default` to `MANAGED_BASE_URL`, and `effective_status` did pass `None` into that slot — so routing followed `OPENCOMPANY_INFERENCE_URL` while display printed the built-in production constant. The card really could not be made to agree with the traffic by any configuration.

The best part of this is the analysis of why the issue's own option 1 is wrong applied literally: passing the env default into `resolve_effective` synthesizes a decl, which moves `source` to `default`, flips `keyConfigured` true for a key the operator never set (so the console offers "leave blank to keep" for a phantom), and feeds `restart_pending` a decl that does not exist. "The `None` was protecting something real — it was just protecting the URL too" is exactly the right diagnosis, and splitting into two resolves is the fix that follows from it. Deriving `platform_default` from `harness_inference_from_env` rather than reading the env var again keeps display and routing on one source, which is the same one-reader rule the codebase already applies elsewhere.

---

### Major 1 — `src/server/ops/inference.rs`: collides with #634, semantically on the field this PR exists to protect

#634 (`fix(inference): let the company admin set the inference key from the console`, also yours) modifies this same file, and it asserts on `keyConfigured` in four places. That is precisely the field this PR's two-resolve split is built to keep honest: `keyConfigured` must reflect *tenant* config only, never the platform token.

#634 introduces a legitimate new way for `keyConfigured` to become true — an admin setting the key from the console. So after both land, that field has two sources feeding it and one that must never feed it, and nothing asserts the three stay distinguishable. Neither PR mentions the other.

Please state an order, and whichever goes second should carry a test that a console-set key and an injected platform default produce different `keyConfigured` values on the same company. Without it, the invariant this PR argues for so carefully is enforced only by the two of them having been reviewed on the same afternoon.

### Major 2 — the second commit should not actually be dropped

Making the probe fix a separate commit "so it can be dropped independently" is good hygiene, but dropping it would leave the card worse than it is today, not better: `baseUrl` would render the correct staging endpoint while `POST /inference/test` still fires at the hardcoded production host with `Credential::None` and reports the provider unreachable. That is a card contradicting itself on two adjacent fields, which is harder for an operator to reason about than both being wrong in the same direction.

Worth saying in the PR body that the split is for reviewability rather than an actual ship option — otherwise someone doing a nervous partial merge takes you up on it.

### Minor — `platform_default` returns `None` off the `openhuman` feature

On a build without `openhuman` the function short-circuits to `None`, so display falls back to `MANAGED_BASE_URL` — the exact behaviour being fixed. The stated justification is that this is also when the harness resolves no env default, so display still matches routing. That holds where nothing routes at all. Does any shipped configuration serve these ops routes without the `openhuman` feature (the `Desktop` lane, or a default-feature container)? If so the card is still wrong there, just for a narrower audience; if not, one sentence saying the route is unreachable in that build settles it.

### Question — is the resolved `baseUrl` actionable from the console?

The card will now show a platform endpoint injected by the manager, which an operator cannot change from the console. That is the honest thing to display and I would not change it. Does the surface distinguish "this is where requests go, and you set it" from "this is where requests go, and the platform set it"? A URL an operator cannot edit, rendered identically to one they can, invites a support ticket.

---

Test coverage is good — six cases covering the unconfigured arm, the `managed` manifest arm, tenant `base_url` outranking the platform default, a non-managed provider ignoring it, the probe path, and `platform_default` following the injected URL. The `effective_status_with` seam existing purely so the resolution is testable without the process environment is the right call.

CodeGhost21:
Thanks — all four addressed. Pushed `986b092`.

### Major 1 — order against #634, and the invariant pinned here

**Order: this PR first, #634 rebases onto it.** I checked what #634 actually does to `src/server/ops/inference.rs` rather than assuming from the title: it extends the module doc header and adds `managed_key_can_be_set_rotated_and_cleared`. No behaviour change in the ops layer — as its own RCA says, `resolve_endpoint` has always preferred a stored key over `EnvDefault` on `managed`, and the `inference-key-conflict` guard it retires is frontend-only (`InferenceSection.tsx:397`; nothing in `src/` references it). So the two are semantically independent and the collision is a mechanical one in the test module.

That said, the ordering is the less important half of your point, so I did not leave the invariant to whichever lands second. `a_platform_token_never_reads_as_a_console_set_key` drives **one company with one injected platform default to both answers**: no tenant key → `keyConfigured: false` while the platform token is doing the outbound work; a console-set key (the exact `store_key` write #634's screen performs) → `true`; and `baseUrl` stays the platform's in both, so paying for your own agents on the managed brain does not move you off it. #634's new source now arrives against a test that already separates the three.

One thing worth flagging for the rebase: #634's test asserts `baseUrl == MANAGED_BASE_URL` after a managed `PUT`. That still holds — the default-feature test build has no platform default — but it is an assertion that only holds *because* of that, and it is the kind of coupling this thread is about. Fine as-is; noting it so it is a decision rather than a survival.

### Major 2 — agreed, and the PR body now says so

You are right that dropping it makes the card worse, not merely less good. Body updated: the split is for reviewability, **not a ship option**, under a *Merge order* section that spells out the self-contradicting-card failure mode so a nervous partial merge does not take me up on the old wording.

### Minor — non-`openhuman` builds are shipped, but there is no gap there

Checked rather than asserted. They are shipped: `Dockerfile`'s `FEATURES` arg defaults to empty, and `deploy/README.md` suggests sets like `medulla tinyplace sqlite` — no `openhuman`. But nothing outside `src/harness/` reads `OPENCOMPANY_INFERENCE_URL` (`hosted_endpoint_from_env` is its only reader, and `EnvDefault` is only ever constructed under `#[cfg(feature = "openhuman")]` in `runtime/builder.rs`). So in such a build there is no injected endpoint for the card to misreport: the built-in constant is the only inference endpoint that exists, and the company is on the echo or hosted brain, which `cognition` already reports.

The medulla hosted path is the one adjacent wrinkle: it talks to `DEFAULT_API_URL` over its own socket protocol, not an OpenAI-compatible surface, so `baseUrl` describing a provider that path does not use is pre-existing and orthogonal to this fix. Recorded all of this on `platform_default`'s doc comment so the next reader does not have to redo the trace.

### Question — no, and it gets worse after #634

Honest answer: the surface does **not** cleanly distinguish them, and `source` only looks like it does.

- `source: "managed"` → platform or built-in URL, not operator-set.
- `source: "runtime"`/`"manifest"` **with an explicit `base_url`** → operator-set.
- `source: "runtime"`/`"manifest"` **on `managed` with no `base_url`** → the URL is the platform's, but the badge says `runtime`. This is the ambiguous case you intuited.

The console renders `status.baseUrl` as plain muted text (`InferenceSection.tsx:292`) with the source badge attached to the provider name, so those three read identically apart from the badge. And the third case is exactly what #634 makes common: an admin setting only a key on `managed` produces `source: "runtime"`, provider `managed`, platform URL — a badge saying "you configured this" over a URL they cannot edit.

That is the issue's option 2 (`baseUrlSource` of `platform-env` vs `built-in-default` vs `tenant`), which I skipped here because `source` appeared to cover it. It does not, for that third case. I would rather not grow this PR into a DTO addition plus console change under a `changes requested` — happy to file it as a follow-up referencing #597 and this thread, or to add it here if you would prefer it lands with the fix. Your call.

### Verification

`cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (1814 passed). On 1.96.1: `cargo clippy -p opencompany --features openhuman --all-targets --no-deps -- -D warnings`, `cargo test --features openhuman --lib server::ops::inference` (17 passed).

oxoxDev (approved):
**0 major.** Approving. All four addressed, and two of them better than asked.

- **Major 1** — order stated (this first, #634 rebases), and you verified it by reading what #634 actually does to `src/server/ops/inference.rs` rather than inferring from its title. The invariant is now pinned by `a_platform_token_never_reads_as_a_console_set_key()`, which is the test that keeps `keyConfigured` honest once #634 adds its second writer.
- **Major 2** — body now carries a *Merge order* section saying the split is for reviewability and not a ship option, with the self-contradicting-card failure mode spelled out. That is the right place for it.
- **Minor** — you checked rather than asserted, and the answer is better than the question: non-`openhuman` builds *are* shipped (`Dockerfile`'s `FEATURES` defaults empty), but nothing outside `src/harness/` reads `OPENCOMPANY_INFERENCE_URL` and `EnvDefault` is only constructed under that feature, so there is no injected endpoint to misreport. Recording the trace on `platform_default`'s doc comment is what stops the next reader redoing it.
- **Question** — answering "no, and it gets worse after #634" is the useful answer. An operator cannot tell a platform-set endpoint from one they set. Worth its own issue rather than expanding this PR.

CodeGhost21:
CI note — the red `Rust (openhuman, tinycortex)` job here is **not from this PR**. Opened #648 to fix it on `main`.

The failing test is `workflows::gate::tests::a_plain_read_and_a_metered_read_do_not_gate`, in `src/workflows/gate.rs` — a module this branch does not touch and which did not exist at its branch point. This PR's earlier run (`919ed474`, 10:35) was fully green; the run that went red (`986b0921`, 11:59) is green on everything except that one test, which merges in from `main`.

`main` is independently red on it: its own runs at 11:27 (`84c4a1db`) and 11:43 (`0a142e85`) both failed on the same test, before this PR's run. Cause is a merge-only conflict between #606 (issue #459 moved `read_workspace_state` to `Reach::Consequence`) and #460 (added the gate test using it as an example of a read that must *not* gate) — each green against the `main` it branched from.

#648 fixes the test's example rather than the classification, since that reach is a deliberate stopgap with a recorded revert condition (`tinyhumansai/openhuman#5494`). Once it lands, a re-run here should be green with no change to this branch.

CodeGhost21:
## Manual E2E verification — console + host, before/after

Rebased onto `main` (`f6667d1`) and verified end-to-end against a running host rather than only in tests, since the whole complaint in #597 is that a rendered value could not be trusted as evidence.

**Setup.** Binary built `--features openhuman,tinycortex` — the fix is inert without `openhuman`, where `platform_default` returns `None` by design. Host launched through the repo's own `frontend/test/e2e/host.sh`, with the endpoint injected via its documented `PW_HOST_PASSTHROUGH` hatch:

```
OPENCOMPANY_INFERENCE_URL=https://staging-api.tinyhumans.ai/openai/v1
TINYHUMANS_API_KEY=e2e-platform-token
```

Company `e2e_harness`, signed in as `harness-e2e@tinyhumans.ai`.

### A/B — same host, same company, same env; only the resolution differs

The "before" is the same build with `effective_status` reverted to `effective_status_with(runtime, None)`, i.e. exactly the pre-fix line, so nothing but the defect changes between the two rows.

| | `GET /api/v1/company/inference` → `baseUrl` |
|---|---|
| pre-fix | `https://api.tinyhumans.ai/openai/v1` ❌ |
| fixed | `https://staging-api.tinyhumans.ai/openai/v1` ✅ |

The pre-fix row reproduces the payload in the issue verbatim, including on a tenant where `OPENCOMPANY_INFERENCE_URL` was set — the "no amount of environment configuration changes what it prints" half of the report.

Every other field is byte-identical across both rows:

```json
{"provider":"managed","slug":"managed","source":"managed","keyConfigured":false,
 "cognition":"harness","usageMetering":"perTurn","restartRequired":false}
```

So the two-resolve split moved the one field it was meant to and left the three it exists to protect alone — on a live host, not only under `MapEnv`.

### Console

Settings → Connections → Inference, driven in a real browser, DOM read rather than inferred:

- **pre-fix:** card renders `https://api.tinyhumans.ai/openai/v1`
- **fixed:** card renders `https://staging-api.tinyhumans.ai/openai/v1`

Both under the same `managed` badge, both with no "key set" indicator and `Cognition: harness · usage metered per turn`. That last point is the review thread's `keyConfigured` invariant holding on the actual screen: the platform credential is doing the outbound work in both, and neither card claims a stored tenant key.

Exactly one `/openai/v1` URL is present in the rendered page in each case, so there is no second surface still printing the constant.

### Caveats, stated rather than buried

- Driven with Playwright-controlled Chromium, not the Claude Chrome extension (not connected in my session). Real browser and real bundle either way, but it is scripted navigation, not a human clicking.
- `TINYHUMANS_API_KEY` is a dummy string. It exercises credential *resolution* — which is what gates `platform_default` returning `Some` — not a live call to the endpoint. The `Test` button would need a real token.
- Build carries no `mcp` feature, hence an unrelated "no agent can use tool servers" banner on the same screen.

Issue #396, opened 2026-08-05
connections: the Connect button stores a credential that no agent can ever use


oxoxDev:
## What the operator sees

The Connections page offers GitHub, Google, Slack and others with a **Connect** button. Completing that OAuth handshake shows the provider as connected — and no agent gains any ability whatsoever. The credential is stored and never read again.

So an operator can do everything the page asks of them, see a success state, and get nothing.

## Why it happens

`oauth_key(provider)` — `"oauth/{provider}"` — is defined at `src/server/ops/mod.rs:175-177` and referenced from exactly two places:

- `src/server/ops/connections.rs` — written at `:457`, read at `:555` only to build a revoke request, blanked at `:634`
- the two read projections, `src/server/ops/connections_read.rs:209` and `src/server/graphql/connections.rs:122`, which read it **solely to compute a display boolean**

There are **zero occurrences under `src/harness/`**. No agent tool resolves a credential from this namespace. The catalog is write-only by construction.

It is also currently unreachable on any real deployment: hosted tenants are injected no `OPENCOMPANY_OAUTH_*` variables (the manager's tenant manifest injects company, bind, data dir, public URL, Mongo and mailbox vars — nothing else), and nothing is set locally. So today the page connects nothing anywhere, and if it did, nothing would consume it.

Note the two connection paths that *do* reach an agent — Composio (`src/harness/composio.rs`) and remote MCP servers (`src/harness/mcp.rs`) — resolve their credentials from entirely different namespaces and never consult this one.

A manifest comment compounds the confusion: `companies/agentic_marketing_agency/company.toml:86-87` claims "the manager injects real OAuth credentials at runtime". It does not. `src/server/ops/connections.rs:17-19` says the opposite and is correct.

## What the fix should do

Decide what this surface is for, and make it honest either way:

1. **If the catalog is the intended future** — wire it to the harness, so a connected provider actually confers a capability an agent can use. That is the substantive work and it needs the token-custody question answered (see #319).
2. **If Composio is the intended path** — say so on the page. An operator should not be offered a Connect button that buys them nothing. At minimum the catalog should be gated off where it cannot be consumed, rather than presenting a working-looking flow.

Either way, correct the misleading manifest comment.

## How to verify

- A provider connected through this page enables something an agent can demonstrably do — or the page no longer offers to connect it.
- `rg 'oauth_key|oauth/' src/harness/` returns something, or the surface is gated.
- No template claims the manager injects OAuth credentials.

## Related

[...]

PR #637, opened 2026-08-11
feat(connections): broker providers through the company's own TinyHumans credential (#586)

### LINKED ISSUE #586: connections: route brokered providers through the company's own TinyHumans credential instead of per-tenant provider apps
## The premise

Once a company holds its admin's TinyHumans credential (see the companion issue on setting it from the console), connections no longer need per-tenant provider apps. The company already carries a credential the platform backend recognises — the same one paying for inference. Composio, and anything else the backend brokers, should ride it.

This is the unblock the connections work has been waiting on. #319 asks to route hosted OAuth through the platform backend instead of per-tenant provider apps; #404 asks for OpenHuman-style provider management. Both were blocked on the tenant having no platform identity to present. Under the one-key-per-tenant model it has one.

## Where the pieces are today

- **Composio's tenant token is already read live from the secret store**, fingerprinted per cycle, so a console set/rotate/clear takes effect on the next turn. That half is built.
- **The console catalogue now comes from the backend** rather than a hardcoded list (#516), with a stated degraded state when the backend cannot answer.
- **What is missing** is the link: the Composio credential and the inference credential are separate things an admin sets separately, when under this model one company credential should be able to authorize both.

## The change

Decide and implement how one company credential authorizes the brokered surfaces:

1. **Does the admin's key authorize Composio directly**, with the backend brokering per-provider OAuth on the company's behalf — or does the company still hold a separate Composio token that the key merely *provisions*?
2. **Where does a provider connection live** — per company, on the backend, keyed by the owning account? That is what makes "connect Gmail once, every teammate's agents can use it" true, which is the behaviour OpenHuman already has and the thing #404 is asking for.
3. **What happens on rotation.** If one key authorizes several surfaces, rotating it must reach all of them — the same argument `hosted_endpoint_from_env` already makes for chat and embeddings sharing one resolution rather than two that drift.

## Worth reading first

OpenHuman's existing provider flow is the reference implementation here, and it already does what this is reaching for. Reading it before designing this is likely to save a redesign — its Skills grid drives off the backend's toolkit catalogue rather than a local list, which is the same shape this needs.

## Acceptance

- A company with its key set can connect a provider without any per-tenant provider app being registered.
- A connected provider is usable by every agent in the company, not only by the member who connected it.
- Rotating the company key does not silently leave one brokered surface on the old credential.
- A company with no key set gets a stated, honest degraded state rather than a broken picker.

CodeGhost21:
## Summary

Closes #586.

A company's admin sets **one** TinyHumans key on the tenant, and every surface the
platform brokers on the company's behalf presents it. Composio is the first: the
backend derives the Composio entity from whatever bearer it is handed, and a
TinyHumans key is one it recognises — so a company with a key set connects Gmail
or Slack with **no Composio token pasted and no per-tenant provider app
registered**.

## What the issue got half-right

The issue says connections were "blocked on the tenant having no platform
identity to present." That blocker is already gone, and further than the issue
credits: `TenantComposio::resolve` has routed through the instance's platform
identity since #319's work landed, so a hosted pod already connects providers
with zero provider apps.

The real gap was narrower and sharper: **the company's own credential was not a
tier in any brokered surface's resolver, and there was no seam where one could
live.** Four independent credential resolutions existed and none consulted a
company-owned key —`hosted_endpoint_from_env` (env-only), `TenantComposio::resolve`
(`composio/token` → instance identity), the media/search resolvers (env-only by
design), and `HostConnectRoutes` (stored token → projected tier → host provider
app). Nothing fanned out, so acceptance criterion 3 — rotation reaching every
surface — failed by construction rather than by oversight.

### The trap worth flagging

The issue describes the admin's key as "the same one paying for inference," which
reads as *reuse `inference/key`*. **That would be a cross-vendor credential
leak.** `inference/key` is provider-scoped, not an identity: `resolve_effective`
hands it to whatever provider the company declared, so it may hold an OpenRouter
`sk-or-…` or a raw BYOK token — `validate_parts` even sniffs for those prefixes.
Presenting it to `api.tinyhumans.ai` would send one vendor's credential to
another. The two coincide only when the declared provider is `managed`. So the
company credential gets its own slot, and the module docs say why.

## The three decisions the issue asked for

1. **Does the key authorize Composio directly, or provision a separate token?**
   Directly. No provisioning dance — the platform-identity tier already proves a
   TinyHumans bearer works on `/agent-integrations/composio/*`.
2. **Where does a connection live?** On the backend, keyed by the account the
   bearer resolves to. No new local storage. Every agent resolves the *same*
   company credential, so "connect once, every teammate's agents can use it" is a
   property of the resolution rather than a feature layered on top.
3. **What happens on rotation?** One seam, read live per call, fingerprinted by
   identity. Because there is no second resolution to drift, a rotation cannot
   reach one brokered surface and miss another.

## The change

- **`src/company/company_key.rs`** — the single seam. `resolve` = company's own
  key (`tinyhumans/key`) → this instance's platform identity → fail closed. A
  store read error degrades to the next tier rather than bricking a roster build.
- **`Credential::Company` / `CredentialSource::Company`** — hashed by **value**,
  unlike the projected platform token which hashes by **path**. An admin rotates
  deliberately, so a new value is a new identity and the roster must rebuild; the
  cluster rotates the projected token every few minutes, and hashing *that* value
  would rebuild every agent's toolbelt on the platform's schedule.
- **Composio** resolves `composio/token` → company key → instance identity. The
  BYO hatch survives untouched, and clearing it falls back to the company key
  rather than silently re-borrowing the instance's identity.
- **`GET` / `PUT …/credential`** — write-only, admin-only (issue #403's
  discipline: this key is the identity every agent presents *and* the account
  they all spend against). A set and a clear are journaled separately and
  attributed. A credential change evicts the cached toolkit catalog, since the
  catalog is a property of the credential.
- **Console** — a Company credential card above the Composio section, the new
  tier rendered in the Composio status, and the degraded state reworded to name
  the company key instead of only offering a Composio token.

## API or behavior changes

- **New:** `GET …/credential` → `{configured, source, notice}`; `PUT …/credential`
  with `{key}` (empty string clears). Admin-only on the write.
- **`GET …/composio` `credentialSource` gains `"company"`.** The console unions
  (`ComposioCredentialSource`, `ConnectionCredentialSource`) are widened to match.
- **No behaviour change for a company that sets nothing.** With no company key the
  precedence is byte-for-byte what it was, which the existing resolver tests still
  assert unchanged.

## Deliberately out of scope, stated rather than papered over

- **The native OAuth catalog** keeps its own precedence, so `source: "company"`
  never appears on a native-only provider. Routing it through the company key
  entangles this with the inert-catalog question in #396.
- **Chat inference and embeddings** still resolve from the environment via
  `hosted_endpoint_from_env`. That is #585 — which now only has to *consume* this
  seam rather than build one, and inherits the rotation guarantee when it lands.

## Tests

New coverage: the resolver precedence matrix and its fail-closed arm, the
rotation-moves-the-fingerprint contract, the BYO-token-still-wins arm, the
clear-falls-back-one-tier-at-a-time arm, the write-only round trip, the
member-is-refused arm with a proof the refused write stored nothing, and the
journal's set-vs-clear distinction.

### The negative control found a real bug in the first draft

Deleting the company-key arm from the resolver was supposed to fail two tests.
It failed **one**. `setting_the_key_credentials_composio_with_no_composio_token`
kept passing, because `GET …/composio` re-derived the tier from two booleans
instead of asking the resolver — so the console would have gone on confidently
reporting `credentialSource: "company"` for a company whose agents were no
longer using the company key at all. A status route that can name a tier the
agents are not on is the same class of drift this issue exists to remove, and my
first draft shipped one.

Fixed in 74d7041: `resolve_credential` is now the single derivation, moved into
the always-compiled `company::composio` (not the feature-gated harness) so the
console and the harness can share it in every build. The matrix test now drives
the real resolver through a real secret store instead of restating its
precedence — the pure-function form was exactly what let the drift pass.

Re-run with the arm deleted, all three now fail as they should:
`setting_the_key_credentials_composio_with_no_composio_token`,
`credential_source_matrix_follows_the_resolver_precedence`, and
`the_company_key_credentials_composio_between_a_byo_token_and_the_instance`.
Restored, all green.

## Commands run locally

| Command | Result |
|---|---|
| `cargo test` | 1822 passed, 1 ignored |
| `cargo clippy --no-deps -p opencompany --features openhuman,composio --all-targets -- -D warnings` | clean |
| `cargo test --features openhuman,composio` | 2698 passed, **1 pre-existing failure** |
| `npx tsc -b` (frontend) | clean |
| `npm test` (frontend) | 26 files, 275 passed |

**The one failure is pre-existing, not from this branch.**
`server::ops::composio::tests::an_admin_is_unaffected` expects `409` but gets
`502`: under `--features composio` the authorize handler dials the real
`https://api.tinyhumans.ai/agent-integrations/composio/authorize` and takes a
401. Verified by stashing this branch and re-running on a clean `upstream/main`,
where it fails identically. It only passes on a host without network access to
that backend.

Two environment notes for anyone reproducing: the `openhuman` feature does not
build on the repo's pinned `stable` (1.93 — `libsqlite3-sys 0.38.1` needs the
unstable `cfg_select`), and on 1.96.1 `cargo clippy --all-targets` fails inside
vendored `tinyagents` on `large_enum_variant`. Hence the `--no-deps -p opencompany`
form above for linting this crate's own code.

## Docs

`docs/spec/runtime/credentials.md` is new — the resolution order, the rotation
argument, the not-the-inference-key reasoning, the scope limits, and the recorded
known limits (no per-member attribution; two companies pasting the same key share
one entity). `docs/spec/runtime/api.md` links to it and lists the routes; adding
the section inline would have pushed that file to 524 lines, past the repo's
500-line rule, so it was split per `CLAUDE.md`.

## Summary by CodeRabbit

- **New Features**
  - Added company-level TinyHumans credential management.
  - Authorized administrators can configure, rotate, or clear credentials from the Connections view.
  - Added credential status messaging, secure write-only handling, and success/error notifications.
  - Composio can use company credentials with clear status and fallback guidance.
  - Added API endpoints and documentation for credential configuration and behavior.

- **Bug Fixes**
  - Credential resolution now follows consistent precedence and fails closed when storage is unavailable.
  - Credential changes automatically refresh related connection and catalog status.

### DIFF
--- a/docs/spec/runtime/README.md
+++ b/docs/spec/runtime/README.md
@@ -43,8 +43,9 @@ Supporting docs:
   hand over → swap), so a first-time inference config needs no restart
 - [api.md](api.md) — HTTP routes and auth
 - [credentials.md](credentials.md) — the company's own TinyHumans key: the one
-  seam every brokered surface resolves through, and why rotating it reaches all
-  of them
+  seam a brokered surface resolves through (Composio today), why rotating it
+  reaches every surface wired to it, and which surfaces are deliberately outside
+  it
 - [config.md](config.md) — configuration and the one-key story
 - [users.md](users.md) — human collaborators: magic-link/password sign-in,
   sessions, invites, and chat attribution
--- a/docs/spec/runtime/README.md
+++ b/docs/spec/runtime/README.md
@@ -42,6 +42,9 @@ Supporting docs:
 - [rebuild.md](rebuild.md) — replacing a registered runtime in place (quiesce →
   hand over → swap), so a first-time inference config needs no restart
 - [api.md](api.md) — HTTP routes and auth
+- [credentials.md](credentials.md) — the company's own TinyHumans key: the one
+  seam every brokered surface resolves through, and why rotating it reaches all
+  of them
 - [config.md](config.md) — configuration and the one-key story
 - [users.md](users.md) — human collaborators: magic-link/password sign-in,
   sessions, invites, and chat attribution
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -291,6 +291,8 @@ dependency-inverted behind a trait; when the relevant seam is absent the write
 route `404`s with `{"code":"not_wired"}`.
 
 ```text
+GET    …/credential                         whether the company has its own key + which tier it presents
+PUT    …/credential                         set / rotate / clear the company's TinyHumans key  [admin]
 PUT    …/domain                             set the custom domain
 POST   …/domain/verify                       server-side DNS check
 PUT    …/smtp                               store SMTP credentials (secret store)
@@ -299,6 +301,13 @@ POST   …/connections/{provider}/start        begin OAuth (returns authorize UR
 POST   …/connections/{provider}/disconnect   drop stored OAuth tokens               [feature: oauth]
 GET    /api/v1/oauth/callback                OAuth redirect target (unscoped; state carries the company)  [feature: oauth]
 ```
+
+`…/credential` is the company's **one** TinyHumans key, and every surface the
+platform brokers on its behalf presents it — so a company with a key set
+connects a provider with no per-tenant provider app and no second token. The
+resolution order, the rotation guarantee, why it is not the inference key, and
+what it deliberately does not cover are in
+[`credentials.md`](credentials.md).
 
 ### The OAuth callback always redirects
 
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -302,12 +302,10 @@ POST   …/connections/{provider}/disconnect   drop stored OAuth tokens         
 GET    /api/v1/oauth/callback                OAuth redirect target (unscoped; state carries the company)  [feature: oauth]
 ```
 
-`…/credential` is the company's **one** TinyHumans key, and every surface the
-platform brokers on its behalf presents it — so a company with a key set
-connects a provider with no per-tenant provider app and no second token. The
-resolution order, the rotation guarantee, why it is not the inference key, and
-what it deliberately does not cover are in
-[`credentials.md`](credentials.md).
+`…/credential` is the company's **one** TinyHumans key, presented by every
+surface wired to it (**Composio today**) — see
+[`credentials.md`](credentials.md) for the resolution order, the rotation
+guarantee, and which surfaces are deliberately outside it.
 
 ### The OAuth callback always redirects
 
--- a/docs/spec/runtime/credentials.md
+++ b/docs/spec/runtime/credentials.md
@@ -50,6 +50,16 @@ account, so its full order is `composio/token` → company key → instance
 identity → none. What no surface may do is resolve a *company* identity some
 other way.
 
+That composed order is itself derived **once**, in
+`company::composio::resolve_credential` — not in the harness. The agent-facing
+`TenantComposio::resolve` builds its config from it, and the console's
+`GET …/composio` reports its `source`, so the tier an operator is shown cannot
+disagree with the identity the agents actually present. It lives in the
+always-compiled `company::` module rather than the feature-gated `harness::`
+one precisely so both callers can share it in every build; a console route that
+restated the precedence instead would keep confidently reporting a tier after
+the resolver stopped honouring it.
+
 ## Rotation
 
 The rotation guarantee — "rotating the company key does not silently leave one
--- a/docs/spec/runtime/credentials.md
+++ b/docs/spec/runtime/credentials.md
@@ -44,6 +44,38 @@ resolved. Most specific first:
    degraded state rather than offering a picker that cannot work. An absent
    credential means "no tools", never a borrowed identity.
 
+### An unreadable store is not "no key"
+
+A `SecretStore` read error **propagates** rather than falling through to the next
+tier. The tempting shortcut is to map it to "nothing stored" so a transient
+hiccup cannot brick a roster build — right about availability, wrong about
+attribution.
+
+A connection lives on the backend keyed by the account the bearer resolves to.
+If an unreadable store silently resolved a company that *has* a key to the
+instance's identity, any connection established in that window would belong to
+the instance's account. When the store recovered, resolution would return the
+company key, the bearer would resolve to a different entity, and the connection
+made under the fallback would no longer be the one the company sees — the same
+"connect Gmail" click producing a different owner depending on store health at
+that instant, with no signal either way.
+
+Availability-degrade and identity-degrade are different decisions, and only the
+first is safe to make silently. So each caller answers the error in the way its
+own surface can afford:
+
+- **The roster build** (`TenantComposio::resolve`) logs a warning and withholds
+  the tools for that cycle. Fail closed: an *unknown* credential must no more
+  mean a borrowed identity than an absent one does. The company loses Composio
+  for a cycle and gets it back on the next.
+- **The console planes** (`GET …/composio`, `GET …/credential`) surface the
+  failure instead of reporting a confident "not configured" for a company that
+  may well have a key.
+- **`POST …/composio/authorize`** — the call that actually *establishes* a
+  connection — refuses outright. It resolves the credential itself rather than
+  through the roster path, precisely so a store error stays distinguishable from
+  "nothing configured" and cannot be guessed past.
+
 A surface may prepend its **own** escape hatch above that seam. Composio keeps
 its BYO `composio/token` for a company that insists on using its own Composio
 account, so its full order is `composio/token` → company key → instance
@@ -78,6 +110,15 @@ Two mechanics make it land without a restart:
   and hashing its value would rebuild every agent's roster on that schedule. A
   company key is rotated by a person, on purpose, and a new value really is a
   new identity.
+
+**The fingerprint is internal and must stay internal.** It is a value-derived
+hash of a live credential, which makes it a cheap confirmation oracle for anyone
+who can read it: given a guess at the key, you can confirm it. It lives only in
+`HarnessPool`'s in-memory `RwLock<HashMap<CompanyId, u64>>` and is compared for
+equality — no `tracing` call renders it, no DTO serializes it, and no journal
+event carries it. Do not log it, return it from a route, or put it in an event,
+even for debugging; if a rebuild needs explaining, log *that the identity
+changed*, never the hash.
 
 ## Write-only, and admin-only
 
--- a/docs/spec/runtime/credentials.md
+++ b/docs/spec/runtime/credentials.md
@@ -1,0 +1,121 @@
+# The company credential
+
+How a company proves who it is to the surfaces the platform brokers on its
+behalf (issue #586). The routes are listed in
+[`api.md`](api.md#credential-bearing-surfaces-feature-gated); this is the model
+behind them.
+
+## One key per company
+
+A company belongs to one owner. Its admin sets **one** TinyHumans key through
+`PUT …/credential`, and everything the platform backend brokers for the company
+rides it. Membership in the company is what grants access to it.
+
+Composio is the first consumer, and it is the one that shows why this works: the
+backend derives the Composio entity from whatever bearer it is handed, and a
+TinyHumans key is a bearer it recognises. So the key **authorizes Composio
+directly** — there is no provisioning step that trades it for a second token,
+and no per-tenant provider application to register with Google, Slack or GitHub.
+A company with a key set connects Gmail by clicking Connect.
+
+## Where a connection lives
+
+On the backend, keyed by the account the bearer resolves to — under this model,
+the company's owning account. Nothing about a connection is scoped to the member
+who made it, and nothing new is stored on the instance.
+
+That is what makes "connect Gmail once, every teammate's agents can use it" true:
+every agent in the company resolves the *same* credential from the *company's*
+store, so the backend resolves one entity for all of them. It is a property of
+the resolution, not a feature layered on top of it.
+
+## Resolution — one seam
+
+`company::company_key::resolve` is the only place a company identity is
+resolved. Most specific first:
+
+1. **The company's own key** — `tinyhumans/key` in its `SecretStore`, set by an
+   admin through the console.
+2. **This instance's platform identity** — `TinyhumansTokenSource`: a projected,
+   audience-bound pod token the cluster rotates in place, else a static
+   `TINYHUMANS_API_KEY`. Unchanged behaviour for a tenant whose admin set
+   nothing.
+3. **Nothing** ⇒ fail closed. No tools are wired, and the read planes report the
+   degraded state rather than offering a picker that cannot work. An absent
+   credential means "no tools", never a borrowed identity.
+
+A surface may prepend its **own** escape hatch above that seam. Composio keeps
+its BYO `composio/token` for a company that insists on using its own Composio
+account, so its full order is `composio/token` → company key → instance
+identity → none. What no surface may do is resolve a *company* identity some
+other way.
+
+## Rotation
+
+The rotation guarantee — "rotating the company key does not silently leave one
+brokered surface on the old credential" — is structural rather than a
+convention. Because every brokered surface calls the one resolver, there is no
+second resolution that could drift, and no surface that could forget to re-read.
+
+Two mechanics make it land without a restart:
+
+- The key is read **live** from the secret store on every resolution, so a set /
+  rotate / clear takes effect on the next cycle.
+- The resolved credential contributes its **value** to the harness roster
+  fingerprint (`Credential::hash_identity`), so a rotation rebuilds the tool
+  roster. This is deliberately different from the projected platform token,
+  which contributes its *path*: the cluster rotates that one every few minutes
+  and hashing its value would rebuild every agent's roster on that schedule. A
+  company key is rotated by a person, on purpose, and a new value really is a
+  new identity.
+
+## Write-only, and admin-only
+
+The key is sent on the `key` field, stored, and never echoed. No read route
+returns it; `GET …/credential` carries `configured` plus `source` — one of
+`company` / `attested` / `static` / `none`, the same vocabulary `GET …/composio`
+and the connections read plane already use. `Credential`'s `Debug` redacts it,
+and the Composio tools feed whatever value they resolve to the scrubber as a
+known secret, so it cannot survive into agent-visible output.
+
+`PUT` requires an admin. This key is the identity every one of the company's
+agents presents **and** the account they all spend against, so setting it is a
+decision made for the company rather than a member's own — the same reasoning
+that made `PUT …/composio/token` admin-only in issue #403. Both a set and a
+clear are journaled as `ToolAccessChanged`, told apart from each other, and
+attributed to whoever made the change.
+
+## Not the inference key
+
+`inference/key` is a different thing and must stay a different slot. It holds
+whatever credential the company's *declared provider* wants — an OpenRouter
+`sk-or-…`, a raw BYOK token, an `openai_compatible` key. It is provider-scoped,
+not an identity, and handing it to the TinyHumans backend would present one
+vendor's credential to another. The two coincide only when the declared provider
+is `managed`, and even then they are the same value for different reasons.
+
+## What this does not cover
+
+- **The native OAuth catalog.** `…/connections/{provider}/start` still resolves
+  through `HostConnectRoutes` — a stored provider token, else a *projected*
+  instance identity, else this host's own registered provider application. That
+  precedence is unchanged, so `source: "company"` never appears on a native-only
+  provider. That path is entangled with the inert-catalog question in #396 and is
+  left alone deliberately rather than papered over.
+- **Chat inference and embeddings.** Both still resolve from the environment via
+  `hosted_endpoint_from_env`. Moving them onto this seam is issue #585; when it
+  lands they inherit the rotation guarantee by construction, because the seam is
+  already here.
+- **Media generation and `web_search`.** Deliberately environment-only: those run
+  on the *platform's* managed credential, never a company-controlled one.
+
+## Known limits, recorded deliberately
+
+- **No per-member attribution.** Spend arrives as one account, so which member
+  burned what is not answerable. Per-agent `budget_usd_daily` caps still work;
+  per-person accounting does not exist and is not in scope.
+- **Removing someone from the roster stops future access**, but nothing already
+  spent is separable.
+- **Two companies pasting the same key share one entity.** That cannot be
+  prevented client-side; it is a deployment caveat, the same one the BYO Composio
+  token already carries.
--- a/frontend/src/api/composio.ts
+++ b/frontend/src/api/composio.ts
@@ -19,10 +19,15 @@ import type { OpenCompanyClient } from "./client";
  * Where this company's Composio credential comes from.
  *
  * - `attested` — the instance's own platform identity; nothing is stored here.
- * - `static` — a token this company pasted, or a static instance key.
+ * - `company` — this company's **own** TinyHumans credential, set by its admin
+ *   (issue #586). The backend derives the Composio entity from it, so a company
+ *   with a key set connects providers as itself with no Composio token and no
+ *   per-tenant provider app. Rotating that one key moves every brokered surface
+ *   at once — see `api/credential.ts`.
+ * - `static` — a Composio token this company pasted, or a static instance key.
  * - `none` — no credential can be obtained, so agents get no Composio tools.
  */
-export type ComposioCredentialSource = "attested" | "static" | "none";
+export type ComposioCredentialSource = "attested" | "company" | "static" | "none";
 
 /**
  * One provider in the catalog the host offers, with the backend's own display
--- a/frontend/src/api/credential.ts
+++ b/frontend/src/api/credential.ts
@@ -1,0 +1,75 @@
+// The company's own TinyHumans credential (issue #586): the one key its admin
+// sets on this tenant, and the identity every surface the platform brokers
+// presents on the company's behalf.
+//
+// Set it once and Composio rides it — no separate Composio token, no per-tenant
+// provider app to register. Rotate it and every brokered surface moves together,
+// because they all resolve through one seam on the host rather than each keeping
+// its own copy.
+//
+// WRITE-ONLY, like every other credential the console handles: the key goes out
+// on `PUT .../credential`, lands in the host's secret store, and is never
+// returned. The read shape carries only `configured` plus the non-secret tier
+// name. Standalone functions over the shared client (mirrors `api/composio.ts`).
+//
+// NOT the same thing as the inference key in `api/inference.ts`. That one holds
+// whatever the company's *declared provider* wants — an OpenRouter key, a raw
+// BYOK token — so it is provider-scoped, not an identity, and handing it to the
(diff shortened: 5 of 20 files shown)


oxoxDev (dismissed):
**2 major, 1 minor, 1 question.**

The credential hygiene here is right where it counts: the console plane is write-only and never reads the key back, `key_configured` surfaces a boolean and nothing else, `Credential`'s `Debug` redacts the `Company` variant explicitly, and the value is read live from the store per resolution rather than cached into a roster.

**The best thing in this PR is a refusal.** The issue describes the admin's key as "the same one paying for inference", which reads as *reuse `inference/key`* — and you identified that as a cross-vendor credential leak, because `inference/key` is provider-scoped rather than an identity and may hold an OpenRouter `sk-or-…` or a raw BYOK token that would then be presented to `api.tinyhumans.ai`. That is a real leak avoided by reading the issue against the code instead of implementing it as written, and putting the reasoning in the module docs is the right place for it.

---

### Major 1 — `src/company/company_key.rs`: a transient store error silently changes which account brokers the call

`load` maps a `SecretStore` read error to `None`, and the doc comment says this is deliberate: "a transient store hiccup must fall through to the instance identity". That reasoning holds for *availability*. What I do not think it traces is the consequence on a write path.

`resolve` is company key → instance platform identity → fail closed, and per your own design note a Composio connection lives "on the backend, keyed by the account the bearer resolves to". So during a store hiccup, a company that has set its own key resolves to the **instance** identity, and any connection established in that window is keyed to the instance account rather than the company's. When the store recovers, resolution returns the company key, the bearer resolves to a different entity, and the connection made under the fallback is no longer the one the company sees. The operator gets no signal in either direction — the same "connect Gmail" click produces a different owner depending on store health at that instant.

Availability-degrade and identity-degrade are different decisions wearing one code path. Suggest splitting them: `Ok(None)` (no key stored) legitimately falls through, `Err(_)` should either fail closed or be allowed to fall through only for *read* resolutions, never for one that establishes a connection. If the entity derivation makes this unreachable, that is a sentence in the doc comment — but as written the fallthrough is documented as an availability choice and behaves as an attribution one.

### Major 2 — collides with #639 on the Composio core, both directions of the stack

#639 (`connections: categorised, searchable connector grid`, also yours) shares five files with this PR:

```
frontend/src/api/composio.ts            src/company/composio.rs
.../connections/ComposioSection.tsx     src/harness/composio.rs
                                        src/server/ops/composio.rs
```

Both change how Composio is resolved and how its section renders, and neither PR mentions the other. Given #637 establishes the credential tier that #639's grid presumably surfaces connections through, this one wants to land first — but that should be stated rather than inferred. A note on both, the way #598 and #554 settled theirs.

### Minor — two console-set credentials, adjacent cards, different scopes

#634 gives an admin a console-set *inference* key; this gives them a console-set *TinyHumans* key. Both render as "configured / not configured" on the Connections surface, and your own analysis is that conflating them would be a cross-vendor leak. An admin looking at two similar cards will not have read that analysis. Worth making the copy on each say what the key is *for* and what it is not, since the distinction is load-bearing and currently lives only in module docs.

### Question — where does the identity fingerprint go?

The roster fingerprint moves with the key's **value** via `hash_identity`, which is what makes rotation rebuild the roster. Does that fingerprint stay internal, or can it reach a log line, the journal, or an API response? A value-derived hash of a live credential is a weak oracle if it is ever surfaced — cheap to confirm, expensive to discover later.

CodeGhost21:
Thanks — Major 1 was a real bug, and the framing (availability-degrade vs identity-degrade wearing one code path) is exactly right. All four addressed in 226ddfa.

## Major 1 — store error no longer falls through

You traced it correctly, and I'd add that the same flaw was already there for the BYO `composio/token` before this PR — upstream's `_ => Credential::None` on the token read had the identical shape. My change would have extended it to the company key rather than introducing it, which is worse, not better: it made the fallthrough load-bearing for the tier that actually carries a company's identity.

`load` and `resolve` now return `Result`, and the error propagates. Each caller answers it in the way its surface can afford, which I think is the split you were pointing at:

- **Roster build** (`TenantComposio::resolve`) — logs a warning and returns `None`: fail closed, no Composio tools that cycle. It must not bubble here, but an *unknown* credential is no more a licence to borrow one than an absent credential is. The company loses tools for a cycle and gets them back on the next; that is recoverable and visible, where acting as the wrong account is neither.
- **Console reads** (`GET …/composio`, `GET …/credential`) — surface the failure rather than reporting a confident `none` for a company that may well have a key set.
- **`POST …/composio/authorize`** — the connection-establishing write, and the one your analysis is really about. It now resolves the credential itself rather than through the roster path, precisely so a store error stays distinguishable from "nothing configured", and refuses. Nothing gets connected under a guessed identity.

Three tests: `an_unreadable_store_never_borrows_another_identity` (the seam propagates, and an *absent* key still falls through — the two cases are now distinguishable, which is the point), and `an_unreadable_store_withholds_tools_rather_than_borrowing_an_identity` (the roster path fails closed *even though* an instance identity is available). Reasoning is in `company_key::resolve`'s docs and `docs/spec/runtime/credentials.md`.

## Major 2 — #639

Fair, and my omission. #637 lands first: it establishes the credential tier and moves the tier derivation into `company::composio::resolve_credential`, which is the function #639's grid will be reading connections through. #639 then rebases onto it — the overlap in `composio.rs`/`ComposioSection.tsx` is adjacent rather than semantic (this PR touches credential resolution and the credential card; #639 touches catalog presentation and the grid), so the rebase should be mechanical. Noted on #639 too.

## Minor — two adjacent cards

Agreed, and the risk is concrete: an admin pastes an OpenRouter key into the TinyHumans field, and the failure is a 401 from a backend they never meant to call. Both notices now name the distinction — the field is "TinyHumans account key", the section carries "for connecting providers — not the model key", and the host-side copy says plainly that this is *not* the model-provider key on the Inference card and why they are stored separately. A test asserts the notice mentions the Inference card, so the disambiguation cannot quietly regress.

I deliberately did **not** touch `InferenceSection.tsx`: #634 edits that file, and adding a third overlap to fix a copy problem seemed worse than carrying the whole distinction on this card. Whichever of #634/#637 lands second should add the mirror-image line; I've noted it there.

## Question — the fingerprint stays internal

Checked, and it does: it lives only in `HarnessPool`'s in-memory `RwLock<HashMap<CompanyId, u64>>` and is compared for equality. No `tracing` call renders it, no DTO serializes it, no journal event carries it — `grep` for fingerprint across logging macros, `Serialize` derives and `CompanyEvent` comes back empty.

You're right that it would be a weak oracle if it ever surfaced, so I've written the invariant onto `fingerprint()` itself rather than leaving it as a fact about today's code: never log, serialize or journal it, and if a rebuild needs explaining, log *that the identity changed*, never the hash. That is where someone adding a debug line will actually be looking.

CodeGhost21:
## The red `Rust (openhuman, tinycortex)` is not this branch — `main` is red

```
workflows::gate::tests::a_plain_read_and_a_metered_read_do_not_gate
  panicked at src/workflows/gate.rs:367
  [GatedToolCall { node_id: "read", slug: "read_workspace_state",
     reason: "'read_workspace_state' has an external effect and this desk runs supervised" }]
```

`src/workflows/gate.rs` is not a file this PR touches. I checked out clean `upstream/main` (e6a75d4, no changes of mine) and ran that test directly — **it fails identically there**.

### Cause: two PRs that were each green, merged 16 minutes apart

- **#612** (`wf/460-tool-call-approval-policy`) merged at **16:40:57**, adding `src/workflows/gate.rs` and the test asserting `read_workspace_state` does **not** gate.
- **#606** (`fix/459-workspace-state-consequence`) merged at **16:56:45**, reclassifying `read_workspace_state` as having an external effect — `src/policy/consequence.rs:351`.

Neither branch contained the other's change, so both were green when tested against the base at the time. Merged in sequence they contradict: #606 makes the classifier say "external effect", #612's test says a plain read must not gate, and #612 made workflow `tool_call` nodes consult exactly that classifier.

`consequence.rs:286` even records the history — *"`read_workspace_state` was the seventh member of this list until issue…"* — so the reclassification was deliberate. What is missing is the corresponding update on the workflow-gate side.

### What this means here

Every open PR on this repo will show the same red X on that lane until main is fixed; it is not a signal about any of them. Everything else on this run is green (Rust, Rust (mongodb), Console, Console E2E, Desktop, Actions pinned), and locally this branch is clean on both feature sets other than the pre-existing network-dependent `an_admin_is_unaffected` noted in the PR body.

### Not fixing it here

The resolution is a semantic call that belongs to #459/#460's author, not to a credential PR: either `read_workspace_state` should gate in workflows too (update #612's test and its supervised-desk expectation) or workflow reads should be exempt from the reclassification (narrow the classifier). Guessing at that inside this PR would bundle an unrelated policy decision into a credential change. Happy to send it as its own PR if the call is made — say which way and I will.

oxoxDev:
Re-checked at the new head — narrowing, not re-blocking.

**Major 1 closed, and closed the right way.** An unreadable store no longer borrows another identity: the `Err` arm now refuses for that cycle instead of falling through to the instance tier, and the test makes the distinction explicit by asserting it *with an instance identity available* — which is precisely the case that would have silently worked before. Separating "no key stored" from "could not read the key" is what makes the availability-degrade safe on a path that establishes connections.

**Still open — Major 2.** No stated order against #639, which shares five files with this including `src/harness/composio.rs`, `src/server/ops/composio.rs` and `ComposioSection.tsx`. #636 shows the shape you already use: a *Merge order* section naming the dependency and what rebases onto what. Given this establishes the credential tier #639's grid presents, this one almost certainly goes first — it just needs saying.

CodeGhost21:
## Manual verification against a running stack

Rebased onto `f6667d1`; the gate test that had `main` red is fixed and this branch is green on it.

Ran the real binary and the real console dev server, and drove the flow **through the Vite proxy** — the same path the browser takes, not direct calls to the handler.

```
backend   ./target/debug/opencompany serve --company companies/agentic_marketing_agency   (127.0.0.1:8110)
console   vite  OC_API_TARGET=http://127.0.0.1:8110                                       (localhost:5199)
login     ada@example.com via OPENCOMPANY_ADMIN_EMAIL + the dev_code path
```

### The acceptance criteria

| # | Criterion | Result |
|---|---|---|
| 1 | A company with its key set can connect a provider with no per-tenant provider app | `GET …/composio` → `credentialSource: "company"` with **no `composio/token` stored anywhere** |
| 3 | Rotation leaves no surface on the old credential | rotate → still `company`; clear → `none`, and the Composio plane agrees in the same breath |
| 4 | No key ⇒ stated, honest degraded state | `source: "none"` + a notice that says what is unavailable and what to do |

### Tier precedence, walked forward and back

```
nothing set          → none
+ company key        → company      ← the issue, in one line
+ BYO composio token → static       (the escape hatch still outranks)
− BYO composio token → company      (falls back ONE tier, not to nothing)
rotate company key   → company
− company key        → none
```

### Write-only, checked rather than assumed

The `PUT` response, `GET …/credential`, `GET …/composio` and `GET …/inference` were each grepped for the literal key value. All clean.

### The cross-vendor leak, disproven at the storage layer

This is the part I most wanted to see with my own eyes, since refusing to reuse `inference/key` is the load-bearing decision in this PR:

```
secrets/  (0700)
  tinyhumans_key    ← the company credential
  composio_token    ← the BYO hatch
  (no inference_key — never created)
```

`GET …/inference` kept reporting `keyConfigured: false` throughout. The company key never reached the provider slot, so an OpenRouter-shaped credential can never be presented to `api.tinyhumans.ai` by this path. Cleared slots are 0-byte rather than removed, matching the documented no-delete store.

### Audit trail

3 × `credential_set`, 2 × `credential_cleared`, each `by: {kind: "user", id: …}` — exactly the five writes performed, sets and clears told apart.

---

## What this does NOT cover

Two gaps, stated plainly so nobody reads the above as more than it is.

**1. The console UI was not driven.** Everything above is the HTTP layer. The card renders and is wired (`CompanyCredentialCard` is served by the dev server and imported by `ConnectionsView`), but no one has clicked through the states — the badge flip, the admin-only hiding, the toast copy. That is a manual pass I still owe.

**2. No live Composio call was made with a company key.** This ran the **default build**, so `inBuild: false` and the live client was never constructed. What is verified is that the company key *resolves* as the Composio credential; what is **not** verified is that the backend *accepts* it as a Composio bearer.

On (2) I think the premise holds, and here is the argument rather than an assurance: `TinyhumansTokenSource`'s **Static** tier is a long-lived `TINYHUMANS_API_KEY`, and that tier already presents its value to `/agent-integrations/composio/*` today in the docker/self-host path. A company key is the same kind of credential from a different store, so it is not a new bearer shape for the backend. Still — that is reasoning, not a 200 from prod. Confirming it needs `--features openhuman,composio` plus a real key against a real backend, which I have not done and would rather do deliberately than incidentally.

If the reviewer wants (2) closed before merge, say so and I will run it against staging.

CodeGhost21:
@coderabbitai — one taken, one declined.

**Scope consistency: taken, and it went further than the finding.** You were right that the overview docs said "every brokered surface" while `credentials.md` excludes native OAuth, inference, embeddings, media and `web_search`. The same overclaim was also in a **user-facing string** you could not see — `SWITCH_NOTE`, the toast the console shows after a save, promised the key "reaches every surface the platform brokers at once".

That one mattered most: an admin reading it would reasonably conclude their inference spend had just moved onto this key, which is false until #585 lands. Fixed in 0bbea2c across the toast, the module docs, the fingerprint comment, `api.md` and the runtime README. The wording is now "every surface **wired to** this credential — Composio today", which is what the build actually does.

**`api.md` deliberately not expanded.** Upstream's copy is already **506 lines**, past this repo's own 500-line rule, so I left a two-line pointer and kept the detail in `credentials.md` rather than adding to an overflow that predates this branch. Worth someone splitting separately.

**Grammar suggestion on `credentials.md:3`: declining.** "How a company proves who it is" is ordinary English and reads more plainly than "proves its identity"; the LanguageTool hit is a false positive from an orthography rule, not a grammar error. Happy to change it if a human reviewer prefers the formal register.

CodeGhost21:
Worked through all findings in e2117b5. **Nine fixed, two skipped** — and two of the nine were regressions I introduced, so thanks for the catch.

### The two that were mine

**`ComposioSection` — I made the BYO override unreachable.** `credentialed = attested || companyKey` hid the paste card for a company-key admin, but the reveal button was still gated on `attested` alone. Net effect: a company that set its TinyHumans key lost any route back to its own Composio account. Button now gates on `credentialed`, and the explainer is per-tier — a company-key company falls back to *its own credential*, not to the pod's cluster identity, so the old wording would have described a fallback it does not have.

**`fetch_catalog` — I undid my own store-error fix.** It collapsed every `resolve_tenant` error into "this company has no Composio credential yet". That is precisely the confident-wrong-answer the earlier propagation work removed: an operator whose secret store hiccupped would be told they never set a key. Now only `Conflict` gets that wording; anything else keeps its message.

### The rest

- **`ConnectionsView` deps** — correct and load-bearing. That probe feeds `reach.hasCredential` and `attested`; without the re-run the grid kept every tile on the no-credential route while the section directly below reported the opposite. The `key={credentialGeneration}` remount only covered `ComposioSection`, not this effect.
- **Member copy** — the no-credential notice pointed members at controls only admins can see. Branches on `canManage` now.
- **Journal vocabulary** — agreed, and sharper than it looks: both routes append `ToolAccessChanged` to one log, so a shared vocabulary meant an auditor could not distinguish rotating the company's whole identity from swapping one integration's token. Now `company_key_set` / `company_key_cleared`, documented on `CompanyEvent`, with a test asserting the Composio words do not appear when no Composio write happened.
- **Card error handling** — this one matters more than "nitpick": hiding the card on any failure would discard the distinction the store-error work exists to draw. Only a 404 hides it; anything else renders a stated error saying the credential is *unknown*, not absent. Write failures surface the host's reason.
- **`effective_status` doc parity** — right, they are not the same resolver; Composio prepends its BYO tier. Doc corrected, and your suggested test is in: a company holding both reads `company` on `/credential` and `static` on `/composio`, and both are correct because they answer different questions.
- **Stale doc fragment** — real, left by the #639 rebase. Removed.

### Skipped

- **Grammar reword on `credentials.md:3`** — "proves who it is" is ordinary English and reads more plainly than "proves its identity"; the LanguageTool hit is a false positive on an orthography rule. Happy to change if a human reviewer prefers the formal register.
- **Hoisting test helpers into `test_support`** — sound in principle, but it means editing the Composio tests too, in files #639 just rewrote and other PRs are still touching. Not worth the conflict surface inside a credential change; better as its own sweep across all the ops test modules.

Local: fmt/clippy clean both feature sets, 1979 default tests, 2979 under `openhuman,composio,tinycortex`, frontend 36 files / 406 tests. The one failure is the pre-existing network-dependent `an_admin_is_unaffected` noted earlier.

oxoxDev (approved):
Re-reviewed against `e2117b56`. All four findings closed, and the response to Major 1 goes further than what I asked for.

**0 major, 0 minor. Approving.**

**Major 1 — closed, properly.** `load` now returns `Result<Credential>` and the error propagates. More to the point, each caller answers it as its own surface can afford rather than under one blanket policy: the roster build logs and withholds the tools for that cycle, the console planes surface the failure instead of a confident "not configured", and `POST …/composio/authorize` — the call that actually establishes a connection, which was the whole risk — refuses outright and resolves the credential itself, so a store error stays distinguishable from "nothing configured". The attribution-versus-availability distinction now lives in the doc comment on `resolve`, which is where it needed to be.

**Major 2 — closed by time, and in the order I suggested.** #639 merged at 14:08 (`91165629`) and this now sits on top of it, `MERGEABLE`. Nothing left to coordinate.

**Minor — closed.** The card says which key this is *not*, and `ComposioSection`'s explainer is worded per tier.

**Question — answered.** `fingerprint`'s doc records that it must never be logged, serialized or journaled.

**The best part of this round is `e2117b56`** — seven findings raised on yourself, two of them regressions introduced by the earlier review fixes. The `fetch_catalog` one is the kind that ships silently: collapsing every `resolve_tenant` error into "no Composio credential yet" quietly undid the store-error propagation from `d7c71355`, so an operator whose secret store hiccupped would have been told they never set a key. Checked the fix — only `Conflict` gets that wording now and everything else propagates its own. Catching the second-order breakage of your own first-order fix is the step most people skip.

Also right: journaling `company_key_set` / `company_key_cleared` instead of reusing Composio's `credential_*`. Both routes append `ToolAccessChanged` to one log, and a reader has to be able to tell a rotation of the company's identity from a swap of one integration's token — pinning it with a test that the two vocabularies do not bleed is what makes that hold.

**One thing before merging, not a finding:** `Rust (openhuman, tinycortex)` and `Console E2E` are still pending. This branch changes `src/harness/composio.rs`, and the gated lane is the only one that compiles the harness tests. Everything else is green. Approving on the code — just don't merge while it is `UNSTABLE`.

oxoxDev:
One note so the bot's batch doesn't get dismissed wholesale: of CodeRabbit's eight findings, **seven predate `e2117b56` and you have already fixed them** — the unreachable BYO reveal, the `canManage` notice, the `credentialGeneration` dep, the stale doc fragment above `credential_source_for`, and the docs wording. It is reviewing a revision you have moved past.

**One is live and I think it is right**, `CompanyCredentialCard.tsx:80` — discard results from an old company scope.

`refresh` is `useCallback(…, [client, company])` and the effect re-runs when it changes, but nothing cancels the request already in flight. Switch company while the first `getCompanyCredential` is still resolving and the responses can land out of order, so the later `setStatus` is the *earlier* company's. The card then reports the previous company's configured/not-configured state under the new company's name — and on this card that is decision-shaped, since an admin who reads "configured" for a company that has no key will not set one.

The usual guard is enough:

```tsx
useEffect(() => {
  let ignore = false;
  setLoad("loading");
  setKey("");
  void (async () => {
    // …existing refresh body, guarded…
    if (!ignore) setStatus(await getCompanyCredential(client, company));
  })();
  return () => {
    ignore = true;
  };
}, [refresh]);
```

Cleanest is probably to move the guard inside `refresh` and have it take the flag, so the write path's `refresh()` calls stay unchanged.

My approval stands — this is small and does not need another round from me. Flagging it only because it is the one worth acting on in that batch, and I missed it myself on the pass that approved.

CodeGhost21:
Fixed in ed59f14 — you're right, and thanks for separating the one live finding from the seven the bot raised against a revision I'd already moved past.

The race is real and it is the decision-shaped kind: `refresh` re-runs on a `company` change but nothing cancelled the in-flight request, so two responses can land out of order and the later `setStatus` be the earlier company's. An admin reading "configured" for a company that has no key does not set one, which is the failure that matters.

Used the generation-ref guard `ComposioSection` already runs next door rather than the `ignore` flag, since it satisfies your "move the guard inside `refresh`" preference and matches the file beside it — one `useRef`, bumped at entry, checked before each state write, and no change to the effect or to any caller.

**One thing I did differently from the sketch:** the write path carries the same hazard — a `PUT` resolving after a company switch would paint one company's status under another's name — so it is guarded too, but only its *state mutations*. The success toast stays unconditional. The write genuinely landed, and an admin who navigated away mid-request is still owed that fact; suppressing it would trade a wrong-state bug for a silently-lost-confirmation one.

Frontend: `tsc -b` clean, 36 files / 406 tests.

On the merge gate — understood, and it holds: this push restarts `Rust (openhuman, tinycortex)` and `Console E2E`, so it will be `UNSTABLE` again for ~20 minutes. I'm watching it and will report; nothing should merge until that lane is green, since this branch does touch `src/harness/composio.rs`.

oxoxDev (approved):
Re-reviewed at `ed59f149`. The stale-scope fix is right, and it is a better fix than the one I suggested.

A generation `useRef` rather than a per-effect `ignore` flag covers **both** paths — I only pointed at the read, and `write` had the same hole: `result.status` describes the company that was on screen when the write started. Splitting the toast from the state is the part I would not have thought to ask for and is clearly correct: the write really did land, so an admin who navigated away is still owed that fact, while painting `result.status` under a different company's name would be the same bug in a new place. Keeping `onChanged?.()` inside the guard is right too — refreshing the *new* company's Composio probe because the *old* company's credential changed would be noise.

**1 blocker, and it is two lines.**

### Blocker — the `Console` lane is red on the design-token gate

```
✗ raw Tailwind palette colour
  frontend/src/views/connections/ComposioSection.tsx:371
  frontend/src/views/connections/CompanyCredentialCard.tsx:174
      <span className="inline-flex items-center gap-1 text-xs text-emerald-600 dark:text-emerald-400">
```

I checked whether this is yours or pre-existing main breakage, and it is yours to fix — though not through any fault in the change itself. `scripts/ci/assert-design-tokens.sh` landed on `main` in `7c40da03` ("Assert the token layer in CI, and mark the debt cleared"), and that sweep left `ComposioSection.tsx` with **zero** `emerald` occurrences on `main`. Your branch predates the sweep, so it carries the old class back in — and `CompanyCredentialCard.tsx` is new here, so its line has never been through the gate at all.

`docs/design-system/color.md` has the mapping; this is a state colour, so `--status-*` is the bucket. Merging `main` first is worth doing anyway, since the gate will keep failing against a stale copy of that file.

Everything else is green, including the two that matter most here: `Rust (openhuman, tinycortex)` at 23m and both `Console E2E` runs.

### Standing verdict

My approval carries over — the code is right and this is a lint gate, not a defect. It was already conditioned on green CI, and that condition now has something concrete attached to it: fix the two class strings, let the `Console` lane go green, and this is done from my side. CodeRabbit's `CHANGES_REQUESTED` is the remaining formal block; the one substantive finding in its batch is the one you just fixed, so the rest should be resolvable.

Issue #406, opened 2026-08-05
workflows: the toolbar picker shows neither the selected workflow nor a placeholder


oxoxDev:
## What the operator sees

The workflow picker in the Workflows toolbar renders **blank**. Not the selected workflow's name, and not the placeholder either — nothing. An operator cannot tell which workflow the canvas below is showing.

## Why it happens

`SelectValue`'s render prop receives `undefined`, and because `children` overrides `placeholder`, the fallback text never renders either. So both the value and its stand-in are lost.

Confirmed to reproduce on `fa936b9` by building that exact file, so it is not introduced by any work in flight.

## Why nothing caught it

`frontend/test/e2e/workflow-edit-delete.spec.ts` asserts `expect(combobox).toContainText(name)` — that spec is **failing on main right now**. Playwright specs are typechecked by CI but never run (there is no `webServer` in the Playwright config), so a spec that already pins this behaviour has been red without anyone seeing it.

That is worth a separate look on its own: a suite that cannot fail is a suite that cannot help.

## What the fix should do

Render the selected workflow's name, and the placeholder when nothing is selected. Then confirm the existing spec passes rather than writing a new one — it already asserts the right thing.

## How to verify

- Selecting a workflow shows its name in the picker.
- With no workflow selected, the placeholder shows.
- `workflow-edit-delete.spec.ts` passes when run manually.

## Related

Found while verifying #303 (PR #405), which is unrelated to the cause. Note that PR adds a Browse index — currently the only way to see which workflow is selected while this is broken.

Issue #397, opened 2026-08-05
connections: an empty Composio toolkit list means "allow everything" but displays nothing


oxoxDev:
## What the operator sees

On almost every shipped company template, the Composio section of the Connections page renders a status badge and a token-paste card — and **no provider rows at all**. There is nothing to click. An operator who wants to connect Gmail or GitHub through Composio has no button to do it with.

The agent tools work fine in that state. It is only the human who has no way in.

## Why it happens

`frontend/src/views/connections/ComposioSection.tsx:209`:

```
showProviders = load === "ready" && status !== null && status.toolkits.length > 0
```

Those toolkits come from the manifest's `[tools.composio].toolkits`, which **defaults to empty** (`src/company/types.rs:504-511`) — and an empty list means *"open mode: defer to the backend allowlist"*, i.e. allow everything. The console reads that same empty value as "show nothing".

So the one value that means **allow every provider** renders **zero providers**.

Only `companies/agentic_software_company/company.toml:31-32` sets the field (`["gmail","slack","github"]`). Every other template — all the other `agentic_*` companies and `e2e_harness` — ships empty, and therefore ships with no Composio provider rows.

## What the fix should do

Distinguish "no toolkits allowed" from "all toolkits allowed", which the current condition conflates.

In open mode the console should offer the providers the backend actually permits — its default allowlist runs to roughly a hundred slugs including `gmail`, `github`, `slack`, `notion` and `googlecalendar`. That likely means the status response carrying the effective list rather than the console inferring it from a manifest field whose empty case means the opposite of empty.

Also worth deciding: whether showing a hundred providers is useful, or whether open mode should surface a curated subset with a way to reach the rest. A list that long is its own usability problem.

## How to verify

- A company with no `toolkits` set shows connectable providers rather than none.
- A company with an explicit list shows exactly that list.
- A company whose list is genuinely meant to be restrictive still restricts.
- The agent-side behaviour is unchanged in every case — this is a console gate, not a capability change.

## Related

Discovered during a connections audit, 2026-08-05.
#316 covers the console showing contradictory connection state across surfaces; this is adjacent — the same page, a different way of being unhelpful.

oxoxDev:
Reopening — the fix in #402 solved the first half but not the part this issue actually asked for.

[...]

PR #639, opened 2026-08-11
connections: categorised, searchable connector grid (#600)

### LINKED ISSUE #600: connections: 123 providers render as one flat list — port OpenHuman's categorised, searchable connector grid
## Summary

Now that open mode serves the backend's full catalog, Settings → Connections renders **123 providers as a single flat vertical list** of name + "Sign in" rows, twelve at a time behind a "Show all 123 providers" button. That is not a browsable surface at this size.

OpenHuman already solved this for the same catalog. This is a request to port its layout.

## What OpenHuman does (`tinyhumansai/openhuman`)

The equivalent surface is `app/src/pages/Skills.tsx` (apps tab), with supporting pieces in `app/src/components/skills/` and `app/src/components/composio/`:

**1. Chip-tab category filter** — `SkillCategoryFilter.tsx` over a fixed, ordered bucket list (`skillCategories.ts`): All, Built-in, Channels, Chat, Productivity, Tools & Automation, Social, Platform, Other. Each chip carries an icon and colour, and only categories actually present are offered.

**2. Categories derived from Composio's own data, not hand-maintained.** `mapComposioCategory()` (`toolkitMeta.tsx`) matches substrings against the `categories[]` the backend already returns — `chat|messaging|communication` → Chat, `productivity|document|calendar|task|storage|email` → Productivity, `crm|developer|analytics|payment|finance|database|cloud` → Platform — and falls back to a slug/name keyword heuristic when nothing matches. This is the piece worth copying verbatim: it means 123 providers bucket themselves and a new Composio integration lands in the right group with no code change.

**3. Search across name *and* description**, debounced, composed with the category filter rather than replacing it.

**4. Connected-first ordering.** `composioSortRank()` sorts connected → pending → expired → error → unconnected, then alphabetically within each rank. What you already use is always at the top.

**5. A dense branded tile grid, not rows** — `repeat(auto-fill, minmax(5.5rem, 1fr))` with `gridAutoRows: 6.5rem`. Compact tiles with a logo, so ~123 items are scannable at a glance instead of being 123 stacked full-width rows.

**6. Per-provider metadata on each tile** — name, one-line description, branded icon plus a Composio-hosted `logoUrl`, a status label and colour per connection state, a multi-account badge, and a short "what you're authorising" hint.

## The blocker — the host throws the metadata away

This cannot be done in the console alone today. `list_catalog_toolkits` (`src/harness/composio.rs:457`) reads `resp.catalog` — whose entries carry `slug`, `name`, `logo`, `description`, `categories`, `enabled` — and reduces every one to a lowercased slug:

```rust
pub async fn list_catalog_toolkits(config: &TenantComposio) -> Result<Vec<String>>
```

So `effectiveToolkits` reaches the console as bare strings. There is nothing to group by, nothing to brand with, and nothing to search except the slug. The console's current flat list is the honest rendering of what it is given.

The backend already computes and serves all of it (`backend/src/services/composio/catalog.ts` assembles the union of the managed allowlist and the project's custom-auth toolkits, joined with display metadata, and states it exists "so the frontend reads `name`/`logo`/`description`/`categories` straight from here"). It is being discarded one layer before the UI.

## Suggested shape (author's call)

1. **Host** — widen `list_catalog_toolkits` to return a struct per toolkit (`slug`, `name`, `description`, `logo`, `categories`) instead of `Vec<String>`, and carry it through the composio status DTO alongside `effectiveToolkits`. Keep the existing slug list for the agent-side admission path, which does not need any of this.
2. **Console** — port the four behaviours above: chip-tab category filter, search over name + description, connected-first sort, tile grid. `mapComposioCategory` transfers almost unchanged since it keys off the same `categories[]` strings.
3. **Optional**, worth a look before deciding: OpenHuman keeps a local fallback catalog of ~119 toolkit names (`MANAGED_COMPOSIO_TOOLKITS` in `toolkitMeta.tsx`) so tiles still render names when the live catalog is unavailable, with live data winning when present. That is a judgement call about carrying a second list — OpenCompany's `catalogSource: "fallback"` path already covers the degraded case differently, so it may not be worth importing.

## Not in scope

The categorised **Connect** tiles further down the same page are a different surface with a different defect — every one of them 400s on a hosted tenant. Tracked separately in `#599`.

## Related

- `#599` — the categorised Connect tiles fail on hosted; they route through native OAuth, not Composio.
- `#597` — the Inference card on this page reports a hardcoded URL.
- `#556` — un-capped the list to 123, which is what made the flat layout a problem worth fixing.

CodeGhost21:
Closes #600.

## Sequencing — #637 lands first

**This PR is second in line behind #637** (`feat(connections): broker providers through the company's own TinyHumans credential`). Stated here rather than left to whichever branch pushes last, per review.

The two share five files. #637 establishes the credential tier that decides *whether* a company can connect anything; this decides *how* the resulting options are presented. Credential before presentation is the right order, and #637 is the older PR.

**They compose.** I stacked this branch's two commits onto `fix/586-brokered-company-credential` locally and ran the suite. The entire overlap is two import lines:

| File | Conflict | Resolution |
|---|---|---|
| `src/company/composio.rs` | #637 adds `use std::sync::Arc`, this adds `use serde::Serialize` | both kept |
| `src/server/ops/composio.rs` | #637 drops `token_configured` (moved into `resolve_credential`), this adds `CatalogEntry` | #637's import + `CatalogEntry` |

No semantic interaction: #637 rewrites `credential_source_for` / `effective_status`'s credential half, this touches `effective_toolkits`, `fetch_catalog` and the DTO's catalog half. Stacked, `cargo test --locked` passes 1822 + 10 + 1 and `--features composio` passes 88 (the one failure is the pre-existing network case noted below).

I have deliberately **not** retargeted this PR's base to #637's branch. #637 is currently 55 commits behind `main`, so stacking now would run this PR's CI against a stale base and lose the green run against current `main`. When #637 merges I will rebase this onto `main` and re-verify.

## Summary

Settings → Connections rendered 123 providers as one flat vertical list, twelve at a time behind a "Show all 123 providers" button.

**The layout was a symptom, not the defect.** `list_catalog_toolkits` (`src/harness/composio.rs`) read the backend's `catalog[]` — whose entries carry `slug`, `name`, `logo`, `description` and `categories` — and reduced every one of them to a lowercased slug:

```rust
resp.catalog.iter()
    .filter(|entry| entry.enabled.unwrap_or(false))
    .map(|entry| normalize(&entry.slug))   // ← five of six fields dropped here
```

That `Vec<String>` then survived every layer downstream unchanged — the catalog cache, `OpenModeToolkits`, `ComposioStatusDto`, and `frontend/src/api/composio.ts` — so the console received 123 bare strings. Nothing to group by, nothing to brand with, and nothing to search but the slug. Its flat list was the honest rendering of what it was handed. The backend already assembles all of this metadata and `backend/src/services/composio/catalog.ts` states it does so "so the frontend reads `name`/`logo`/`description`/`categories` straight from here"; it was being discarded one layer short.

`#397` built this surface sized for 8 providers and `#556` un-capped it to 123 without revisiting the shape — the `PREVIEW_COUNT = 12` collapse is the seam where that showed.

### Host

- `CatalogEntry` (`slug`, `name`, `description`, `logo`, `categories`) lands in `company::composio`, **not** the harness — the status route is always compiled and `src/harness/` is behind the `openhuman` feature. Same reason `TOKEN_KEY` and `backend_url_or_default` live there.
- Threaded through the catalog cache, `OpenModeToolkits` and the status DTO.
- The DTO gains `effectiveCatalog` **alongside** `effectiveToolkits` rather than replacing it: the slug list is the existing contract and is still all an authorize call needs. `OpenModeToolkits::slugs()` derives it, so the two cannot disagree about what is on offer.
- De-duplication is `or_insert_with`, not `collect()` into the `BTreeMap` — collecting keeps the **last** value per key, and a duplicate catalog entry is typically the degenerate one. The test mock's metadata-less `Gmail (dup)` silently blanked the real Gmail's description until this was explicit.
- Manifest lists, the degraded fallback, and backends predating the dynamic catalog all yield slug-only entries. That is a first-class state, not a reason to drop a provider — the console renders them with its own typography.

### Console

Ports the four behaviours from OpenHuman's `Skills.tsx` / `toolkitMeta.tsx`, now that the metadata they need is on the wire:

- **Category chips** over a fixed, ordered bucket list, offering only buckets that are actually populated — a chip that reliably yields an empty grid teaches the operator to distrust the whole row.
- **Buckets derived from Composio's own `categories[]`** by substring (`mapComposioCategory`, near-verbatim). This is the piece worth copying: 123 providers bucket themselves, and provider 124 does too, with no edit on either side of the wire. A slug-only entry falls through to a slug/name keyword heuristic rather than vanishing.
- **Search over name, slug AND description**, composed with the chip (AND, not replacing it). Matching on description is new — there was no description on the wire before — so "invoices" now reaches Stripe and QuickBooks.
- **A dense branded tile grid** replacing the rows. The preview cut is **deleted rather than relabelled**: it was a workaround for a flat list being unreadable, and a grid does not need it.

Two deliberate calls in the tile: the whole tile is the affordance (an 8.5rem tile has no room for a label and a button both), and non-actionable tiles — connected, or a viewer who cannot manage — render as a `div` rather than a disabled button, so "Gmail, connected" stays in the reading order for a member who opened this panel to learn exactly that. Logos are best-effort by construction (derived from the slug wherever the backend published none), so the tile catches the 404 and falls back to a monogram instead of a broken image.

## API Or Behavior Changes

- **`GET …/composio` gains `effectiveCatalog`** — an array of `{slug, name, description, logo, categories}`, same providers and same order as `effectiveToolkits`. **Additive**: `effectiveToolkits` is unchanged in type, contents and order, so existing consumers need no change. Pinned by the DTO key-order test.
- **`harness::composio::list_catalog_toolkits` returns `Vec<CatalogEntry>` instead of `Vec<String>`.** Internal; its only non-test caller is the open-mode status path.
- **Agent-side admission is untouched.** `toolkit_allowed` takes slugs and never consulted this function. This widens what is *described*, not what is *permitted* — and a non-empty manifest allowlist still cannot be widened by the catalog.
- Console: the "Show all N providers" button is gone, replaced by the always-complete grid.

## Tests

- [x] `cargo test --locked` — 1854 + 10 + 1 passed
- [x] `cargo +1.96.1 test --locked --features openhuman,tinycortex --tests` — 2775 + 10 + 1 + 11 + 2 passed
- [x] `cargo +1.96.1 test --features composio composio` — 84 passed (see caveat below)
- [x] `npm run typecheck`, `typecheck:unit`, `typecheck:e2e`, `npm test` (315 passed), `npm run build`

New coverage:

- `list_catalog_toolkits_carries_the_display_metadata` — the host-side regression test. The mock catalog now publishes `logo`/`description`/`categories`, so a mock that omitted them could not have caught this bug. Also pins first-entry-wins de-duplication and `logo: None` for an unpublished logo.
- `list_catalog_toolkits_falls_back_to_the_plain_allowlist` now asserts slug-only `CatalogEntry` values.
- Console: `mapComposioCategory`, `providerLabel`, `availableCategories`, `filterByCategory`, `permissionHint`, description-search, metadata-carrying, and slug-derived category guessing. `visibleProviderRows` now asserts the **absence** of a preview cut — a test still asserting one would be pinning the bug.

**Added after review** (drift guard on the ported `mapComposioCategory`):

- `keeps the buckets its OpenHuman twin produces` — pins the substring table case-by-case. Not asserting that the code does what the code does: it is the diff a divergence has to survive, so editing the table without editing the list fails the suite. That is the moment the editor learns a twin exists.
- `orders its buckets so the first hit wins, not the last` — both copies return on first match, so a multi-category entry depends on Chat → Social → Productivity → Platform. Reordering one side only is the subtlest available drift and was otherwise unpinned.
- `never drops a provider whose category Composio has just invented` — the property the review went looking for and could not find stated outright. An unmapped category still yields a tile in "Tools & Automation" rather than a provider silently leaving the grid.
- A notice on each copy naming the other by repository and path. The OpenHuman half is tinyhumansai/openhuman#5496.

### Verified against a running host

Not just unit-tested — driven end to end against a `--features composio` host pointed at a 122-provider mock catalog. The status route returned `catalogSource=backend` with 122 slugs **and** 122 metadata entries; all 122 tiles rendered with real Composio logos, connected-first; the Chat chip narrowed to 10; and `"invoices"` returned Stripe and QuickBooks — a description match that was impossible before this change.

### Two caveats, both stated plainly

1. **`server::ops::composio::tests::an_admin_is_unaffected` fails on my machine, and it is pre-existing.** It expects a 409 from the build check, but under `--features composio` the handler dials the real `https://api.tinyhumans.ai`, which my box can reach, so the 401 surfaces as a 502. Confirmed identical on a stashed tree. It passes in the default build, and CI does not build the `composio` feature.
2. **No e2e spec.** The grid only renders when `inBuild: true`, which needs a `composio`-feature host, and no CI lane builds one. An e2e spec would typecheck and never run — precisely the trap `playwright.config.ts` documents from #406. Coverage sits in unit tests plus the host tests instead; if a composio e2e lane is wanted, that is worth its own issue.

## Documentation

No doc changes needed. `docs/modules/server/` does not document the Composio status DTO shape, and the module-level rustdoc in `server/ops/composio.rs`, `server/ops/composio_toolkits.rs` and `company/composio.rs` — which is where this surface is actually documented — is updated in place, including why `CatalogEntry` lives in `company::composio` rather than beside the agent-facing `composio_catalog::CatalogToolkit`.

## Not in scope

- #599 — the categorised **Connect** tiles further down the same page 400 on a hosted tenant. Different surface, native OAuth rather than Composio.
- #597 — the Inference card's hardcoded URL.
- The optional local `MANAGED_COMPOSIO_TOOLKITS` fallback catalog OpenHuman keeps. Skipped deliberately, as the issue anticipated: OpenCompany's `catalogSource: "fallback"` path already covers the degraded case honestly, and importing a second hand-maintained list of 119 names would reintroduce exactly the drift this change removes.

### DIFF
--- a/frontend/src/api/composio.ts
+++ b/frontend/src/api/composio.ts
@@ -24,6 +24,35 @@ import type { OpenCompanyClient } from "./client";
  */
 export type ComposioCredentialSource = "attested" | "static" | "none";
 
+/**
+ * One provider in the catalog the host offers, with the backend's own display
+ * metadata (issue #600).
+ *
+ * Every field but `slug` is best-effort and may be empty — a manifest
+ * allowlist, a degraded fallback, and a backend predating Composio's dynamic
+ * catalog all yield slug-only entries. The console fills those in from
+ * `@/lib/composio-catalog`, which is why that local typography table survives
+ * rather than being deleted in favour of the backend's names.
+ */
+export interface ComposioToolkitEntry {
+  /** Toolkit slug, e.g. `googlecalendar`. The key every host call is made with. */
+  slug: string;
+  /** Human-readable name, e.g. `Google Calendar`. Empty when unpublished. */
+  name: string;
+  /** One-line description. Empty when unpublished. Searched alongside the name. */
+  description: string;
+  /** Composio-hosted logo URL, or `null` when unpublished. */
+  logo: string | null;
+  /**
+   * Composio's own free-form category names, e.g. `["productivity", "email"]`.
+   *
+   * Forwarded verbatim by the host and bucketed here by substring — which is
+   * what means a Composio integration added tomorrow lands in the right group
+   * with no code change on either side of the wire.
+   */
+  categories: string[];
+}
+
 /** The company's Composio status. Never carries the token. */
 export interface ComposioStatus {
   /** Whether the `composio` feature is compiled into this build at all. */
@@ -54,6 +83,20 @@ export interface ComposioStatus {
    * "other provider" field is for.
    */
   effectiveToolkits: string[];
+  /**
+   * The same providers as {@link effectiveToolkits}, in the same order, each
+   * carrying whatever display metadata the backend published for it (issue
+   * #600).
+   *
+   * This is what makes the panel browsable. Before it, the host reduced every
+   * catalog entry to a bare slug one layer before the console, so 123 providers
+   * could only be a flat list: there was nothing to group by, nothing to brand
+   * with, and nothing to search but the slug.
+   *
+   * Additive rather than a replacement — {@link effectiveToolkits} is still the
+   * slug contract, and it is still all an authorize call needs.
+   */
+  effectiveCatalog: ComposioToolkitEntry[];
   /**
    * Where {@link effectiveToolkits} came from (issue #397).
    *
--- a/frontend/src/lib/composio-catalog.ts
+++ b/frontend/src/lib/composio-catalog.ts
@@ -1,26 +1,42 @@
-// Turning the host's Composio status into the provider rows the console renders
-// (issue #397).
+// Turning the host's Composio status into the provider tiles the console renders
+// (issues #397, #556, #600).
 //
-// The host now answers open mode with the backend's live catalog — roughly a
-// hundred toolkits — rather than a hardcoded eight. A hundred rows is its own
-// usability problem, and dumping them unsorted and unfiltered would trade one
-// broken page for another. These helpers own the two decisions that fixes:
+// ## What changed, and why the shape of this file changed with it
 //
-//  1. **Order**, so the answer to "where is Gmail" is "at the top", not "row
-//     thirty-one". Connected providers first (an operator scanning this panel is
-//     usually checking what is live), then the handful everyone reaches for,
-//     then the long tail alphabetically.
-//  2. **Reach**, so the tail is discoverable. The list collapses to a preview
-//     with an explicit "show all N" and a search box that matches on both slug
-//     and display name. Before this, reaching a provider outside the eight meant
-//     knowing its Composio slug and typing it into a free-text field — which is
-//     a fine escape hatch and a terrible discovery mechanism.
+// #397 taught the host to serve the backend's live catalog instead of a
+// hardcoded eight, and #556 un-capped it — so this surface went from 8 providers
+// to 123. It kept rendering them as one flat vertical list, twelve at a time
+// behind a "show all" button, because a flat list was all the data allowed: the
+// host reduced every catalog entry to a bare slug, and there is nothing to group
+// 123 slugs by.
+//
+// #600 widened that wire (`effectiveCatalog`), so these helpers now own four
+// decisions instead of two:
+//
+//  1. **Category**, so 123 providers are eight browsable buckets. Derived from
+//     Composio's OWN `categories[]` strings by substring, never from a list
+//     maintained here — which is the whole point: a Composio integration added
+//     tomorrow lands in the right bucket with no code change. A slug-only entry
+//     (manifest list, degraded fallback, backend predating the dynamic catalog)
+//     falls through to a slug/name keyword heuristic rather than vanishing.
+//  2. **Order**, so the answer to "where is Gmail" is "at the top", not "row
+//     thirty-one". Connected first — an operator scanning this panel is usually
+//     checking what is live — then the handful everyone reaches for, then the
+//     tail alphabetically.
+//  3. **Reach**, so the tail is discoverable: search over slug, display name AND
+//     description, composed with the category filter rather than replacing it.
+//     Someone who does not know a provider's name can still find it by what it
+//     does.
+//  4. **Typography and branding**, so a tile reads as a product rather than a
+//     slug. The backend's name/description/logo win wherever it published them;
+//     the local tables below fill the gaps.
 //
 // Pure functions on purpose: `vitest.config.ts` scopes the unit runner to
 // helpers with no document and no host, and this is exactly that. What the
-// component does with the rows belongs in `test/e2e`.
+// component does with the rows — the grid, the chips, the images — belongs in
+// `test/e2e`.
 
-import type { ComposioStatus } from "@/api/composio";
+import type { ComposioStatus, ComposioToolkitEntry } from "@/api/composio";
 
 /** Friendly display labels for the common toolkits; slug-cased fallback otherwise. */
 const TOOLKIT_LABELS: Record<string, string> = {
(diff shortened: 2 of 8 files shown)


oxoxDev (changes requested):
**1 major, 1 minor.** The major is sequencing with #637; the diff itself is sound.

"The layout was a symptom, not the defect" is the right read, and the evidence is the five-line `map(|entry| normalize(&entry.slug))` that reduces a six-field catalog entry to a lowercased string one layer before the console. The flat list really was the honest rendering of what the frontend was handed. Pointing at the backend's own comment — that it assembles `name`/`logo`/`description`/`categories` *so the frontend can read them straight from here* — closes the argument.

Three decisions I checked and would keep:

- `effectiveCatalog` added **alongside** `effectiveToolkits`, with `OpenModeToolkits::slugs()` deriving the latter, so the two cannot disagree about what is on offer. Adding a richer field beside an existing contract rather than replacing it is the right call when the old shape is still all an authorize call needs.
- `CatalogEntry` living in `company::composio` rather than `harness` because the status route is always compiled while `src/harness/` is behind the `openhuman` feature. Same reasoning #631 used to move `grants_cover_server` — good that the two land on the same rule independently.
- `mapComposioCategory` returning `null` on no match so the caller falls through to the slug/name heuristic. I went looking for a provider that could silently vanish from the grid when Composio invents a category nobody mapped; the fallthrough is what prevents it. Worth keeping that property pinned if it is not already.

---

### Major — collides with #637 on the Composio core, both directions of the stack

#637 (`feat(connections): broker providers through the company's own TinyHumans credential`, also yours) shares five files with this PR:

```
frontend/src/api/composio.ts            src/company/composio.rs
.../connections/ComposioSection.tsx     src/harness/composio.rs
                                        src/server/ops/composio.rs
```

Both reshape Composio resolution and the section that renders it, and neither mentions the other. #637 establishes the credential tier that decides *whether* a company can connect anything; this decides *how the 123 options are presented*. That argues for #637 first and this rebasing onto it — but it should be a stated order rather than whichever of you pushes last. I have raised the same point on #637.

### Minor — `mapComposioCategory` is a second copy with no drift detection

It is ported from OpenHuman's `mapComposioCategory` in `toolkitMeta.tsx`, driving the same catalog from the same free-form strings, and the doc comment says keeping the two in step is cheaper than diverging. Agreed on the trade — but nothing makes a divergence *visible*. The two consoles will bucket the same provider differently and both will look correct in isolation.

Cheapest fix is a comment on each side naming the other as its twin, so a future edit at least knows a sibling exists. If the substring list ever grows past a handful of entries, a shared fixture asserting both map a known set identically would be worth more.

CodeGhost21:
Stacking note (raised in review on #637): this PR and #637 share the Composio core — `src/company/composio.rs`, `src/harness/composio.rs`, `src/server/ops/composio.rs`, `frontend/src/api/composio.ts`, `ComposioSection.tsx`.

**#637 lands first.** It establishes the company-credential tier and moves the credential-tier derivation into a single `company::composio::resolve_credential`, which is what this grid reads connections through. This PR rebases onto it afterwards.

The overlap is adjacent rather than semantic — #637 touches credential resolution and the credential card, this one touches catalog presentation and the connector grid — so the rebase should be mechanical. Two things to pick up when rebasing:

- `ComposioCredentialSource` gains a `"company"` variant, so any exhaustive switch on it in the grid needs the extra arm.
- `GET …/composio` now reports the tier the resolver actually returns rather than re-deriving it, and can surface a store-read error instead of a confident `none`.

Also worth mirroring here: #637's credential card carries copy distinguishing the TinyHumans account key from the model-provider key on the Inference card (#634), since all three cards end up on this screen.

oxoxDev:
Re-checked at the new head — narrowing rather than re-blocking.

**Closed, and done better than asked.** I suggested a comment naming the twin; you added `## This function has a twin. Edit both.` on both sides *plus* `mapComposioCategory keeps the buckets its OpenHuman twin produces`. A notice tells the next editor a sibling exists; the test tells them when they have already broken it. That is the difference between documenting a duplication and containing one.

**Still open — the Major.** No stated order against #637, which shares five files with this including `src/harness/composio.rs`, `src/server/ops/composio.rs` and `ComposioSection.tsx`. #636 shows the shape: a *Merge order* section naming the dependency and what rebases onto what. One paragraph and this clears.

CodeGhost21:
Thanks — both addressed. Taking them in reverse order, since the minor is the one with code.

## Minor — `mapComposioCategory` drift detection

You're right that the doc comment asserted the trade and then left nothing to enforce it. Fixed in c00130f with the cheap guard you suggested, plus one thing your read prompted that I'd missed:

- **A notice on each copy**, naming the other by repository and path. The OpenHuman half is tinyhumansai/openhuman#5496 (comment-only).
- **`keeps the buckets its OpenHuman twin produces`** — pins the substring table case-by-case. It is not asserting that the code does what the code does; it's the diff a divergence has to survive. Editing the table without editing the list fails the suite, and that is the moment the editor learns a twin exists.
- **`orders its buckets so the first hit wins, not the last`** — this is the one I hadn't considered. Both copies `return` on first match, so a multi-category entry depends on Chat → Social → Productivity → Platform. `["crm", "marketing"]` is Social, not Platform. Reordering the branches on one side only changes real answers while every substring stays identical, which is subtler than editing the table and was completely unpinned. Now pinned on both counts.

On your third "would keep" — the `null` fallthrough. You went looking for a provider that could silently vanish when Composio invents an unmapped category, and it wasn't stated outright, only implied across two other tests. Now explicit as **`never drops a provider whose category Composio has just invented`**: an unmapped string still yields a tile in "Tools & Automation", `availableCategories` still offers that bucket, and the row survives `visibleProviderRows`. Good catch that it was only inferable.

I did not attempt hoisting the function into a shared package. That is a cross-repository dependency decision rather than a review nit, and it's noted as follow-up on the OpenHuman PR.

## Major — sequencing with #637

Agreed, and stated: **#637 lands first.** There's a new `## Sequencing` section at the top of the PR body. Credential-before-presentation is the right order and #637 is the older PR, so this rebases onto it rather than the other way round.

I also went further than stating it, because "should compose" is worth more as a measurement. I stacked this branch's two commits onto `fix/586-brokered-company-credential` locally and ran the suite. **The entire overlap is two import lines:**

| File | Conflict | Resolution |
|---|---|---|
| `src/company/composio.rs` | #637 adds `use std::sync::Arc`, this adds `use serde::Serialize` | both kept |
| `src/server/ops/composio.rs` | #637 drops `token_configured` (folded into `resolve_credential`), this adds `CatalogEntry` | #637's import + `CatalogEntry` |

No semantic interaction — #637 rewrites the credential half of `effective_status`, this touches the catalog half (`effective_toolkits`, `fetch_catalog`, the DTO). Stacked, `cargo test --locked` passes 1822 + 10 + 1 and `--features composio` passes 88, the single failure being the pre-existing network case in the PR description.

**One deliberate deviation from your suggestion:** I have *not* retargeted this PR's base to #637's branch. #637 is currently 55 commits behind `main`, so stacking now would run this PR's CI against a stale base and throw away the green run against current `main` — and any rewrite of #637 under review would churn this branch. The order is stated in prose and I'll rebase onto `main` and re-verify the moment #637 merges. If you'd rather have it enforced structurally by GitHub, say so and I'll retarget — that's your call, not mine, and it's a one-command change.

Nothing here alters the diff you already reviewed: `effectiveCatalog` is still additive beside `effectiveToolkits`, `CatalogEntry` still lives in `company::composio` for the always-compiled reason, and the only new code is tests.

CodeGhost21:
## Live end-to-end test report

Posting this because **the green checks above do not cover half of this PR.** CI never builds the `composio` feature — the header of `src/harness/composio_catalog.rs` states it, and `--all-features` is a `cargo check`, not a `cargo test`. So `list_catalog_toolkits` returning `Vec<CatalogEntry>`, the cache, and the DTO's catalog half are exercised by **none** of the eight passing lanes. This is what stands behind them instead.

Setup: `--features composio` host serving the real console bundle, `OPENCOMPANY_COMPOSIO_BACKEND_URL` pointed at a mock publishing 122 providers in the backend's real `{success,data}` envelope with `logo`/`description`/`categories`/`enabled`, five pre-connected. Real Chromium against the real host.

### Host — the contract, on real data

| | |
|---|---|
| `effectiveCatalog` | 122 entries |
| `effectiveToolkits` | 122 slugs, **identical order** |
| Additive contract | slug list byte-identical to before; key order `…, effectiveToolkits, effectiveCatalog, catalogSource, catalogNotice` |
| Metadata survived | 122/122 logos, 119/122 categories, 0 missing names |

The 3 entries without categories are deliberate — they exercise the slug/name heuristic fallthrough rather than the `categories[]` path.

### Host — the honesty contract, which no unit test reaches end to end

Killed the catalog backend mid-session and forced a re-fetch through the credential-write eviction path:

```
catalogSource   : fallback
catalogNotice   : Composio's provider catalog could not be fetched (GET …/toolkits
                  failed: error sending request…), so this is a built-in starter
                  list and may be incomplete.
effectiveToolkits: 8      effectiveCatalog: 8      all slug-only: true
```

Restored it, re-evicted: `backend`, 122 entries, notice cleared. **`effectiveCatalog` and `effectiveToolkits` stayed the same length in every state** — degraded, recovered, and steady. That is the invariant that would break first if the two lists ever drifted apart, and it is the reason `slugs()` derives one from the other rather than storing both.

### Console — 20 assertions in a real browser

**The defect is gone:** all 122 tiles render at once, no "Show all N providers" button survives, multi-column grid.

**Ordering:** the five connected providers sort to the front.

**Metadata reaches the tile:** >100 logo images; the backend's description becomes the tile tooltip verbatim (`Payments, invoices and subscriptions.` on Stripe).

**Buckets come from Composio's own strings:** chips offer exactly the populated set; `Chat` narrows to the 10 messaging providers — Discord, Front, Intercom, Microsoft Teams, Slack, Telegram, Twilio, Webex, WhatsApp, Zoom — bucketed from `messaging` / `communication` / `chat`, never a local slug list.

**Search reaches what a name cannot:** `"invoices"` returns Stripe and QuickBooks, matched on description alone. This is the capability that did not exist before this PR — there was no description on the wire to match. `"googlecalendar"` still resolves by slug.

**Chip AND search compose:** `Platform` + `git` intersects both; `Chat` + `git` is legitimately empty and says so rather than rendering a blank grid.

**Accessibility calls held:** a connected tile is a `div`, not a disabled button, and still reads "Gmail … connected" — the member case. An unconnected tile is a button whose `aria-label` names what is authorised.

**The action is wired:** clicking a tile POSTs `{"toolkit":"discord"}` to `…/composio/authorize`.

No unexpected console errors.

### One check failed first, and it was my assertion

`Platform` + `git` also returned **DigitalOcean**. Not a bug: `"digitalocean"` literally contains the substring `git` (d-i-**git**-alocean). Plain substring matching, unchanged from before this PR — the old helper did `slug.includes(q) || label.includes(q)` too. I corrected the expectation rather than the behaviour, since altering search semantics is outside #600's scope. Flagging it because it is mildly surprising and someone may want it as a follow-up; it is pre-existing either way.

### What this does not cover

- No automated regression guards any of the above. It was a manual run, and it will not re-run on the next change. A `composio`-feature CI lane is what would fix that properly — worth its own issue rather than being smuggled into this PR.
- The mock is my own fixture, not the real Composio backend. It matches the shape `ComposioToolkitCatalogEntry` deserialises and the shape the existing harness mock uses, but a live-tenant run would still be worth doing before this is trusted on hosted.

oxoxDev (approved):
**0 major.** Approving — the sequencing item is closed, and answered harder than asked.

I asked for a stated order. You added `## Sequencing — #637 lands first` with the reasoning (credential tier before presentation, and #637 is the older PR), then went further and actually **stacked this branch's commits onto `fix/586-brokered-company-credential` locally, ran the suite, and enumerated the entire overlap** — two import lines in `src/company/composio.rs` and `src/server/ops/composio.rs`, with the resolution for each.

That converts "these five shared files will probably be fine" into a checked claim. It is also the answer to the thing I could not verify from outside: five shared files between two large PRs is exactly where a clean textual merge hides a semantic break, and you demonstrated it does not here rather than asserting it.

The twin notice and `mapComposioCategory keeps the buckets its OpenHuman twin produces` from the earlier round stand. Nothing further from me — good to merge once #637 lands.

Issue #565, opened 2026-08-10
EPIC: MCP and Skills tabs — the console and the harness disagree about what is true


oxoxDev:
## What was checked

The MCP and Skills tabs, traced console → host route → harness wiring → agent tool belt on `main`, the same way the approvals gate was traced in #558.

**The good news first, because it shapes the rest:** both surfaces are mostly well built. The MCP console API is complete (list / add / update / remove / live tool discovery / OAuth start / health probe), credentials are genuinely write-only, upstream's list-servers tool is replaced with a credential-redacting one so bearer tokens cannot reach agent-visible output, and the duplicate contradictory MCP page from #414 is already collapsed into one shared component. Skills resolve from three sources and rebuild live.

What follows are the gaps, and they cluster into one shape: **the console and the harness disagree about what is true.**

## The findings

**1 · The console tells operators to restart for a change that is already live.**
Every mutating MCP response carries `REBUILD_NOTE`: *"Agents pick up this change on their next harness rebuild (restart the company)."* That has not been true for some time. `HarnessPool::ensure` re-resolves the effective MCP set at the top of every cycle, fingerprints it, and includes that fingerprint in the staleness check — so adding, editing, enabling or disabling a server rebuilds the roster and reaches agents on the **next turn**, with no restart. The note is stale documentation shipped as UI copy, and acting on it costs a company an unnecessary restart every time.

**2 · The MCP management routes are not feature-gated, but the tools are.**
The whole agent-side MCP bridge sits behind `#[cfg(feature = "mcp")]`. The `/mcp/servers` CRUD routes do not — only the OAuth callback is gated. On a build without the feature, an operator can add a server, see it listed, see `authConfigured` go true, and no agent will ever receive a single MCP tool. Only the `/tools` discovery route answers `not_wired`. This is the same shape as the workflow "not wired in this deployment" report that cost the marketing pod time.

**3 · Nothing shows which agents can actually reach a server.**
An agent reaches a server only when its effective grants cover `mcp:<name>`. With the usual company-wide `allow = ["*"]` that is satisfied, so most presets are fine — but any company that sets per-agent `tools` narrows the grant set, and a server can then be live, healthy, and invisible to every teammate. The console shows health and tool count; it never shows reachability.

**4 · Desk agents can read a skill but can never run one.**
[...]

PR #640, opened 2026-08-11
feat(skills): say that teammates read skills rather than run them (#569)

### LINKED ISSUE #569: skills: a desk agent can read an installed skill but never run one, and the console doesn't say so
## Symptom

The Skills tab lets an operator browse a shared library, install a skill, author a custom one, and enable or disable any of them. A desk agent can then list, describe and read those skills — and can never run one.

## Mechanism

Deliberate, and pinned by a test. `dispatched_belt_excludes_every_deferred_family` asserts that no dispatched belt contains `run_skill`, `skill_run`, `run_workflow` or `await_workflow`. `src/harness/skills.rs` states the reason: the surface is *"deliberately read-only"*, because `RunWorkflowTool` reaches for the global `Config::load_or_init()` and bypasses the harness's metering, so wiring it for desk agents needs an upstream injection seam that does not exist.

Only the orchestrator gets `RunWorkflowTool`, added alongside the other orchestrator tools.

## Why file this if the behaviour is intended

Because nothing tells the operator. The tab presents install / enable / disable — the vocabulary of turning a capability on — for something that will only ever be read. An operator who installs a skill reasonably expects a teammate to be able to *use* it, and the gap between that expectation and the behaviour is invisible until they try and nothing happens.

Two candidate fixes, and they are not exclusive:

1. **Say it.** The tab states that skills are reference material agents read, and that running one is the orchestrator's job. Cheap, honest, removes the surprise.
2. **Close it.** Wire skill execution for desk agents behind the metering seam. That is the real fix and it is upstream-shaped — worth scoping before committing, and worth knowing whether the seam has moved since the note was written.

## Worth noting alongside

Skills liveness is genuinely correct and worth not breaking: the deltas are fetched and fingerprinted **before** the fast-path staleness check, so a skill authored or toggled in the console reaches every agent on the next cycle with no restart. That ordering was a regression once already. Any change here should keep the ordering test intact.

## Acceptance

- The console no longer implies a desk agent will execute an installed skill.
- If execution is wired, it goes through the harness's metering rather than the global config, and the deferred-family test is updated deliberately rather than deleted.

YellowSnnowmann:
## Summary

Closes #569 (finding #4 of the MCP/Skills console↔harness epic #565).

The Skills tab offers **install / enable / disable** — the vocabulary of switching a capability on — for something a desk agent can only ever *read*. `dispatched_belt_excludes_every_deferred_family` keeps `run_skill`, `skill_run`, `run_workflow` and `await_workflow` off every dispatched belt, and only the orchestrator is handed `RunWorkflowTool`, because that tool reaches for the global `Config::load_or_init()` and bypasses the harness's metering.

The behaviour is deliberate and stays exactly as it is. What changes is that the console now says so, instead of leaving the operator to discover it by asking a teammate to use an installed skill and watching nothing happen.

## What the tab says now

- **A standing note** above the tabs: skills are reference material teammates read — playbooks they follow, not buttons they press — and executing a saved workflow stays the orchestrator's job.
- **The subtitle** drops "Capabilities your company can use" for "Playbooks your teammates read".
- **Per installed card**, one line describing what the switch actually decides: *reach*, not capability — "Teammates can read this" / "Hidden from teammates".
- **The two entry points into the tab**, which is where the old framing did most of the work: the Settings rail advertised the page as "What this company knows how to do" — read *before* the tab can correct it — and the Add-skill dialog asked the operator to "describe a capability your company should have" at the exact moment they author one. Both now describe the playbook a teammate reads, in the register of their siblings ("Tool servers and their tools", "Third-party accounts").

## Why the copy lives in `lib/skills.ts`

So the claim is one string with one test on it. `frontend/test/unit/skills-read-only.test.ts` asserts the **claim**, not the layout: the note names reading and names the orchestrator, and neither it nor the per-card line tells an operator that a teammate will run anything (`/teammates? (can|will) (run|execute|use)/i` and friends must not match). A future copy edit is free to move words around; it is not free to quietly put the promise back.

## What is deliberately not here

The issue's second candidate — wiring skill execution for desk agents behind the metering seam — is upstream-shaped and unchanged here. **No harness change at all**: `dispatched_belt_excludes_every_deferred_family` is untouched, and so is the skills-liveness ordering (deltas fetched and fingerprinted before the fast-path staleness check) the issue asks any change here to keep intact.

The e2e spec also asserts the note is visible on the page an operator browses the library from, with the orchestrator named — a real browser against a live host rather than the unit test's word-level check alone.

## Verification

`tsc -b --noEmit`, `tsc -p tsconfig.unit.json`, `tsc -p tsconfig.e2e.json`, `vitest run` (303 passed, 29 files — 6 new), and `npm run build` (the production console build). Frontend-only; no Rust touched, so no cargo lane is affected.

Closes #569

### DIFF
--- a/frontend/src/lib/skills.ts
+++ b/frontend/src/lib/skills.ts
@@ -14,3 +14,38 @@ export const CATEGORY_STYLES: Record<SkillCategory, string> = {
   Content: "border-emerald-500/30 bg-emerald-500/10 text-emerald-600 dark:text-emerald-400",
   Finance: "border-rose-500/30 bg-rose-500/10 text-rose-600 dark:text-rose-400",
 };
+
+// What a skill actually is to a teammate (issue #569).
+//
+// A desk agent can list, describe and read an installed skill, and can never run
+// one: `dispatched_belt_excludes_every_deferred_family` pins `run_skill`,
+// `skill_run`, `run_workflow` and `await_workflow` off every dispatched belt,
+// and only the orchestrator is handed `RunWorkflowTool`. That is deliberate —
+// the upstream runner reaches for a global config and bypasses the harness's
+// metering — but the tab was built from the vocabulary of switching a capability
+// on (install / enable / disable), so an operator reasonably read "enabled" as
+// "a teammate will now do this", and nothing on the screen disagreed until they
+// tried it and watched nothing happen.
+//
+// The copy lives here rather than inline in the view so the claim is one string
+// with one test on it. What must not regress is the *claim*, not its layout.
+
+/**
+ * The Skills tab's standing statement of what installing and enabling a skill
+ * does. Says the two things the screen otherwise implies the opposite of:
+ * teammates **read** skills, and **running** one is the orchestrator's job.
+ */
+export const SKILLS_READ_ONLY_NOTE =
+  "Skills are reference material your teammates read — playbooks they follow, not buttons they press. " +
+  "Enabling one puts it in front of every teammate; executing a saved workflow stays the orchestrator's job.";
+
+/**
+ * What an installed skill's on/off state means for the company's teammates.
+ *
+ * Deliberately phrased as reach ("can read it") rather than capability ("can use
+ * it"): the switch decides whether a skill is visible to a desk agent, and never
+ * whether one can execute it.
+ */
+export function skillReachLabel(enabled: boolean): string {
+  return enabled ? "Teammates can read this" : "Hidden from teammates";
+}
--- a/frontend/src/views/SettingsSection.tsx
+++ b/frontend/src/views/SettingsSection.tsx
@@ -19,7 +19,11 @@ export const SETTINGS_PAGES = [
   { id: "people", label: "People", icon: UserCog, hint: "Who can sign in, and as what" },
   { id: "connections", label: "Connections", icon: Plug, hint: "Third-party accounts" },
   { id: "mcp", label: "MCP Servers", icon: Blocks, hint: "Tool servers and their tools" },
-  { id: "skills", label: "Skills", icon: Sparkles, hint: "What this company knows how to do" },
+  // "What this company knows how to do" read as capability the company performs
+  // — the implication issue #569 exists to remove, set here *before* the tab
+  // gets a chance to correct it. The siblings describe their content; so does
+  // this now.
+  { id: "skills", label: "Skills", icon: Sparkles, hint: "Playbooks your teammates read" },
   { id: "usage", label: "Usage", icon: ChartColumnBig, hint: "What this company is spending" },
 ] as const satisfies readonly { id: string; label: string; icon: LucideIcon; hint: string }[];
 
--- a/frontend/src/views/SkillsView.tsx
+++ b/frontend/src/views/SkillsView.tsx
@@ -438,7 +438,13 @@ function AddSkillDialog({
       <DialogContent className="sm:max-w-md">
         <DialogHeader>
           <DialogTitle>Add a skill</DialogTitle>
-          <DialogDescription>Describe a capability your company should have.</DialogDescription>
+          {/* Not "a capability your company should have" (issue #569): this is
+              where an operator authors one, so it is the earliest point the
+              console can frame a skill as the playbook a teammate reads rather
+              than as something the company will carry out. */}
+          <DialogDescription>
+            Describe a playbook your teammates should follow — what to do, and when.
+          </DialogDescription>
         </DialogHeader>
         <div className="grid gap-2">
           <Label htmlFor="skill-name">Name</Label>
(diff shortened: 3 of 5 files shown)


YellowSnnowmann:
## Which of #569's two candidates this PR takes, and why

#569 offers two non-exclusive fixes: **(1) say it** — the tab states that skills are read — and **(2) close it** — wire skill execution for desk agents behind the metering seam. This PR is **(1) only**, and nothing in the harness changes: `dispatched_belt_excludes_every_deferred_family` is untouched, and so is the skills-liveness ordering (deltas fetched and fingerprinted *before* the fast-path staleness check) that the issue asks any change here to preserve.

So the issue's second acceptance criterion — *"If execution is wired, it goes through the harness's metering rather than the global config, and the deferred-family test is updated deliberately rather than deleted"* — does not fire: nothing is wired.

### One correction to the issue, since it changes what (2) would actually cost

The issue says (2) *"is upstream-shaped"*, citing `src/harness/skills.rs`: `RunWorkflowTool` reaches for the global `Config::load_or_init()` and bypasses the harness's metering, *"so it needs an upstream injection seam that does not exist"*. That note is about the **upstream** skill-execution family, not about this repo's own tool:

- `skills.rs` wires only upstream's three **read** tools (`WorkflowListTool`, `WorkflowDescribeTool`, `WorkflowReadResourceTool`). Each consumes only `config.workspace_dir`, which is exactly why a throwaway `Config` is enough there and the global registry is never booted (`src/harness/skills.rs#L199-L215`).
- This repo's `RunWorkflowTool` (`src/harness/orchestrator.rs#L1685`) reaches for no global config at all. It takes the company store, the runner handle, the `RunSupervisor` and the journal explicitly, and its own docs describe registering an agent-initiated run so it lands in the console's run history as `running` with a Cancel button. It is already constructed and pushed — for the orchestrator (`orchestrator.rs#L1610`).
- `run_skill` and `skill_run` are not imported anywhere in the repo. They appear only as forbidden names inside the pin (`src/harness/build.rs#L1637-L1638`) — defensive entries so upstream's families cannot arrive by accident.

So the machinery to run a company workflow with supervision, cancellation and a journal entry already exists in-tree. What gates (2) is not a missing injection seam; it is the **policy** the pin encodes. That test's own doc says these families were *"deferred for safety reasons that still hold"*, and it distinguishes them from `web_search`, which was deferred for an infrastructure reason that the managed-credential pattern dissolved (#238).

### Why that still means (2) doesn't belong in this PR

Handing a dispatched desk agent workflow execution is a decision about what an unattended agent may spend and start: which autonomy tier permits it, whether it parks for approval, how the run is metered against the company's plan, and what happens when an agent starts a run nobody is watching. That is a policy change with its own blast radius, reviewed on its own terms — not a console-copy fix. Changing the pin as a side effect of making the tab honest would be exactly the "updated deliberately rather than deleted" failure the issue warns about.

Happy to open a follow-up issue scoping (2) against the above — the seam question is answered, so what remains is the tier/approval/metering design.

### Scope note on (1)

The criterion is *"the **console** no longer implies a desk agent will execute an installed skill"*, so the fix is console-wide rather than confined to the view: the standing note and subtitle, the per-card reach line, **and** the two entry points that set the expectation before the tab can correct it — the Settings rail's "What this company knows how to do", and the Add-skill dialog's "Describe a capability your company should have". The claim is pinned by a unit test on the copy and asserted in a real browser by the e2e spec.

oxoxDev (approved):
**0 major, 0 minor, 1 question.** Approving.

This is the right kind of console change: the behaviour is deliberate and stays exactly as it is, and what changes is that the UI stops implying otherwise. `dispatched_belt_excludes_every_deferred_family` keeps `run_skill` and friends off every dispatched belt, and only the orchestrator gets `RunWorkflowTool` because it reaches for the global `Config::load_or_init()` and bypasses the harness's metering — so a tab offering "install / enable / disable" was describing a switch that does not do what its verbs claim.

Fixing the **two entry points** is what makes this more than a copy tweak. The Settings rail advertising the page as "What this company knows how to do" is read *before* the tab can correct itself, so correcting only the tab would have left the wrong frame in place and then argued with it. Catching that the framing does most of its damage upstream of the screen being fixed is the insight here.

Re-describing the per-card switch as **reach** rather than capability — "Teammates can read this" / "Hidden from teammates" — is the precise version of the same point, and it is the line an operator will actually use to predict what the toggle does.

### Question — does anything keep the copy and the belt honest with each other?

The claim the UI now makes ("teammates read these, they do not run them") is true because of a Rust test in the harness. If a future change ever hands `run_skill` to a dispatched belt, this copy becomes the lie it was written to remove, and nothing on the console side would fail. Not worth building machinery for — but a comment on `dispatched_belt_excludes_every_deferred_family` naming the console copy as a downstream consumer would tell whoever loosens it that a user-facing promise rides on it.

Issue #566, opened 2026-08-10
mcp: the console tells operators to restart for a change that is already live


oxoxDev:
## The claim

Every mutating MCP response carries this, in `src/server/ops/mcp.rs`:

```
const REBUILD_NOTE: &str =
    "Agents pick up this change on their next harness rebuild (restart the company).";
```

## Why it is wrong

`HarnessPool::ensure` re-resolves the company's effective MCP set on **every** cycle, fingerprints it (`mcp_fingerprint`), and that fingerprint is one of the terms in the staleness check that decides whether to reuse the cached roster. When it moves, the roster is rebuilt.

So an operator who adds, edits, enables or disables a server gets it on the company's **next turn**. No restart. The same design note in `ensure` says exactly this about the other axes it fingerprints: *"That is what makes 'no restart, no redeploy' a property of the design rather than a claim."* MCP is one of those axes.

## Why it matters more than a wording nit

Restarting a company is not free — it drops conversation state and costs a cold start. The note reads as an instruction, and an operator following it pays that cost for nothing, every time they touch a server. It also teaches the wrong mental model: that the MCP tab is inert until you intervene, which makes the tab feel broken when it is working.

## The change

Replace the note with what actually happens — the change reaches agents on the company's next turn — or drop it entirely, since a change that needs no operator action arguably needs no note.

Then check the neighbours: any other place in the console or docs that tells an operator to restart for a live-refreshed axis (skills, overlay agents, budgets, capability filter, Composio token are all fingerprinted the same way) is wrong for the same reason.

## Acceptance

- No mutating MCP response instructs the operator to restart.
- A test asserts the effective-MCP fingerprint participates in the staleness check, so a future refactor that quietly drops it fails CI rather than silently restoring the restart requirement.

PR #641, opened 2026-08-11
feat(mcp): tell operators when a build cannot run the servers it manages (#567)

### LINKED ISSUE #567: mcp: management routes aren't feature-gated, so a build without `mcp` accepts servers no agent receives
## Symptom

On a tenant built without the `mcp` Cargo feature, the MCP tab works. An operator can add a server, see it in the list, paste a token and watch `authConfigured` turn true. No agent ever receives an MCP tool, and nothing says so.

## Mechanism

The agent-side bridge is gated:

```
#[cfg(feature = "mcp")]
if let Some(registry) = registry_for_agent(&deps.mcp_servers, grants) { … }
```

The management routes are not. In `src/server/routes.rs` only the OAuth callback carries `#[cfg(feature = "mcp")]`; `/mcp/servers` CRUD is mounted unconditionally. The single feature-gated line in `src/server/ops/mcp.rs` does not cover the write path.

The one route that does answer honestly is `/mcp/servers/{name}/tools`, which returns `not_wired` when the harness is off. So the deployment already knows how to say this — it just says it in one place out of seven.

## Why this shape matters

This is the same failure the marketing pod hit on workflows: a console that accepts configuration a deployment cannot honour, and reports success. The operator's next move is to ask an agent to use the server, watch it not happen, and have no way to tell a missing feature from a broken server from a bad token.

## The change

Make the tab tell the truth about the deployment it is running in. Options, roughly ascending:

1. **Report capability.** Have the list route carry whether the MCP bridge is compiled in, and render a stated degraded state in both MCP surfaces — the same pattern the connections catalogue uses when the backend cannot answer.
2. **Refuse the write.** Answer `not_wired` on add/update when the feature is absent, matching the `/tools` route.
3. **Gate the routes.** Cleanest in principle, worst in practice — the tab would 404 with no explanation, which is a different lie.

(1) is the honest one; (1)+(2) together is defensible.

## Also worth confirming

Whether the shipped staging and desktop images actually build with `mcp`. If every real image has it, this is latent rather than live — worth knowing which, and worth a CI build of the feature combination either way, since a combination CI never builds is one nobody can vouch for.

## Acceptance

- On a build without `mcp`, the console states plainly that tool servers are unavailable in this deployment.
- On a build with `mcp`, no change in behaviour.
- A test covers the without-feature path, so the honest message cannot regress silently.

YellowSnnowmann:
## Summary

Closes #567 (finding #2 of the MCP/Skills console↔harness epic #565).

The `/mcp/servers` management routes ship in **every** build. The agent-side bridge does not — `registry_for_agent` is pushed onto a teammate's belt behind `#[cfg(feature = "mcp")]` (`src/harness/build.rs`). An operator can add a server, store a token, watch `authConfigured` turn true, and no agent ever receives a tool from it.

## The case the issue misses, and why the existing honest route doesn't cover it

| Build | CRUD | Discovery / probe | Agent tools |
|-------|------|-------------------|-------------|
| default (no `openhuman`) | works | `not_wired` | none — no harness at all |
| `openhuman`, no `mcp` | works | **works for real** | **none** |
| `openhuman` + `mcp` | works | works | yes |

The issue names `GET …/mcp/servers/{name}/tools` as "the one route that answers honestly". It is honest about **`openhuman`**, not about `mcp`: `harness/mcp.rs` compiles without the `mcp` feature, so on the middle row discovery and health probes answer for real and *every* reading on the screen looks correct while the belt is empty. That row is the one an operator cannot diagnose from the console at all.

Also confirmed, since the issue asks: staging **does** build with `mcp` (`TENANT_FEATURES` in `deploy-staging.yml`), so this is latent there and live for the default `docker-compose` build. And the combination is not un-built by CI — `ci.yml` runs `cargo test --features openhuman,mcp,telegram --lib harness::build::tests` plus `Check (--all-features)`.

## The change — option (1), reported as a capability

`GET …/capabilities` gains **`mcpInBuild`** = `cfg!(feature = "mcp")`, beside the `mediaInBuild` / `composioInBuild` / `searchInBuild` flags that already answer exactly this question for the other optional surfaces. `McpServersSection` reads it and states the degraded case above the list.

**Not on the list route**, which the issue suggested: `GET …/mcp/servers` answers a *bare array*, pinned by `expectList` / `mcp-wire.test.ts` after #414 — wrapping it to carry a flag would reintroduce the very shape that crashed that page.

**Absence is only ever an explicit `false`.** `mcpBridgeState` returns `unknown` for a host that omits the field, for a non-boolean, and for a failed capability read; `unknown` renders nothing. Reporting an older build as broken would be a fresh lie in the other direction. The capability read is also deliberately non-fatal to the list — a host that cannot answer it still gets a working MCP tab, just no claim about the bridge.

## Option (2) — refusing the write — is deliberately not implemented

The issue offers it as "defensible" alongside (1). It shouldn't be: a manifest can legitimately declare servers for a deployment that runs elsewhere with the feature, configuration entered before the capability arrives survives the rebuild, and the default-build test suite adds servers over HTTP throughout. Refusing the write turns a config-ahead-of-deploy flow into a hard error while fixing nothing an operator can act on.

## API / behavior changes

- `CapabilityStatusDto` gains `mcpInBuild: bool` — always serialized, on **both** construction paths (`unconfigured` and the configured branch). Additive; no route, signature or existing field changes.
- No change on a build with `mcp`: the banner is scoped to an explicit `false`.
- The **add-success toast** now comes from `mcpAddedMessage(name, bridge)`. A banner is worth nothing while the success path fires `Added <name>. Agents pick it up on the next rebuild.` at the moment the operator acts, so without the bridge it confirms the add and says the server is stored but uncallable until the deployment is rebuilt with it — still a success, because the add did succeed and the configuration survives that rebuild. An unanswering host gets the ordinary message: unknown is not absence.
- That pickup promise was also **stale on every build**: since #566 the host answers `NEXT_TURN_NOTE` (next-turn pickup, no restart) while the console still said "next rebuild" — in the toast and in the `McpMutationResponse` doc comment. Both now match the host. This is the one place a build *with* `mcp` sees a wording change; shipping a known-false string seemed worse than the letter of "no change in behaviour".

## Tests

- `reports_whether_the_mcp_bridge_is_in_this_build` — asserts the flag on both response paths (a flag added to one alone would report honestly for a company with no plan and lie to every company that has one), and pins the without-feature answer as a literal rather than a vacuous comparison against the same `cfg!`. Runs in the default build **and** under `--features mcp`, so both answers are exercised.
- `frontend/test/unit/mcp-bridge.test.ts` — the three-state predicate, including the two silences (field omitted, read failed) that must not render as "absent"; and the add message, pinning that the bridge-absent case never promises pickup, the present case never says "rebuild"/"restart", and `unknown` reads exactly like `present`.
- `frontend/test/e2e/mcp.spec.ts` — asserts from both sides on a real host: the default-feature lane's binary has no bridge and must show the banner; the `Console E2E (live brain)` lane builds `openhuman,tinycortex,mcp` and must not.

## Verification

`cargo fmt --all -- --check`, `cargo test --locked --lib` (1810 passed), `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked --features mcp --lib server::ops::capabilities` (7 passed), `tsc -b --noEmit`, `tsc -p tsconfig.unit.json`, `tsc -p tsconfig.e2e.json`, `vitest run` (305 passed), and `npm run build` (the production console build).

## Note for the reviewer

`docs/modules/mcp.md` gains a "Which builds can honour a server" section. Its existing "Pool-staleness caveat" still says mid-session edits reach agents only on "practically, a company restart" — the claim #566 / #598 removed from the UI. Left untouched here to keep this diff to one concern; worth a follow-up.

Closes #567

## Summary by CodeRabbit

* **New Features**
  * Added capability-aware MCP support across builds.
  * MCP settings now clearly indicate when agent-side tools are unavailable in the current build.
  * Added live status reporting to distinguish available, unavailable, and unknown MCP bridge states.
  * MCP server setup messages now explain when changes will be picked up on the next turn.

* **Documentation**
  * Documented MCP availability across build configurations, including discovery, management, and agent tools.

### DIFF
--- a/docs/modules/mcp.md
+++ b/docs/modules/mcp.md
@@ -105,6 +105,35 @@ Discovery is gated on the `openhuman` feature (the MCP transport lives there);
 without it the route reports `not_wired` and the console falls back to the
 declared tool lists. Every mutating response carries a `note` reminder.
 
+## Which builds can honour a server (issue #567)
+
+The management routes above are **ungated** — they ship in every build. The
+agent-side bridge is not: `registry_for_agent` is pushed onto a teammate's belt
+behind `#[cfg(feature = "mcp")]`. Three configurations, only one of which the
+routes alone distinguish:
+
+| Build | CRUD | Discovery / probe | Agent tools |
+|-------|------|-------------------|-------------|
+| default (no `openhuman`) | works | `not_wired` | none — no harness |
+| `openhuman`, no `mcp` | works | **works for real** | **none** |
+| `openhuman` + `mcp` | works | works | yes |
+
+The middle row is the one worth stating outright: every read on the screen
+answers correctly, so a healthy badge and a live tool list sit above a server no
+teammate can call. The console cannot infer this — an empty tool belt is not
+visible over HTTP — so `GET …/capabilities` carries **`mcpInBuild`**
+(`cfg!(feature = "mcp")`, alongside `mediaInBuild` / `composioInBuild` /
+`searchInBuild`), and `McpServersSection` renders a stated degraded state when it
+is explicitly `false`. A host that omits the field is *unknown*, never
+"absent" — an older build must not be reported as broken.
+
+Writes stay open on every build deliberately. A manifest can declare servers for
+a deployment that runs elsewhere with the feature, and configuration entered
+before the capability arrives survives the rebuild; refusing the write would turn
+that into a hard error while fixing nothing an operator can act on. Staging
+builds with `mcp` (`TENANT_FEATURES` in `deploy-staging.yml`); the default
+`docker-compose` build does not.
+
 ## Console surface
 
 One component reads these routes —
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -866,6 +866,15 @@ export interface CapabilityStatusDto {
   searchCredentialConfigured?: boolean;
   /** The company's daily `web_search` call ceiling. */
   searchDailyCallCap?: number;
+  /**
+   * Whether the agent-side MCP bridge is compiled into this build (issue #567).
+   * Not a grant question like the flags above: the `/mcp/servers` management
+   * routes ship in every build, so without this an operator can add a server,
+   * store a token and watch it probe healthy on a deployment that hands agents
+   * no MCP tool at all. `undefined` is **unknown** (an older host that does not
+   * send the field) and must never be rendered as "absent".
+   */
+  mcpInBuild?: boolean;
 }
 
 /** One day's token totals in the usage series (`GET .../usage`). */
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -756,9 +756,10 @@ export interface McpServer {
 }
 
 /**
- * A mutating MCP response: the resulting server, a rebuild reminder, the live
- * probe result (absent on a non-`openhuman` host), and any non-blocking
- * endpoint advisory.
+ * A mutating MCP response: the resulting server, the host's pickup note (since
+ * issue #566 that is next-turn pickup with no restart, not a rebuild reminder),
+ * the live probe result (absent on a non-`openhuman` host), and any
+ * non-blocking endpoint advisory.
  */
 export interface McpMutationResponse {
   server: McpServer;
--- a/frontend/src/lib/mcp-bridge.ts
+++ b/frontend/src/lib/mcp-bridge.ts
@@ -1,0 +1,40 @@
+// Whether this deployment can actually run the MCP servers its console manages
+// (issue #567).
+//
+// The `/mcp/servers` management routes ship in every build; the agent-side
+// registry is pushed onto a teammate's tool belt behind the `mcp` Cargo feature.
+// So a build without it accepts a server, stores its token and reports it
+// healthy — and no agent ever receives a single tool from it. The console has to
+// state that, because every other reading on the screen looks correct: on a
+// build with the harness but without the bridge, live tool discovery and health
+// probes answer for real.
+//
+// The three-state answer matters more than it looks. `mcpInBuild` is optional on
+// the wire: a host predating the field sends nothing, and rendering that silence
+// as "no agent can use these" would be a fresh lie in the other direction. Only
+// an explicit `false` is absence.
+
+import type { CapabilityStatusDto } from "@/api/types";
+
+/**
+ * What this build can do with MCP servers.
+ *
+ * - `present` — the bridge is compiled in; managed servers reach agents.
+ * - `absent` — the host said the bridge is missing; servers are configuration
+ *   this deployment cannot honour.
+ * - `unknown` — the host did not say (an older build, or the capability read
+ *   failed). Claim nothing.
+ */
+export type McpBridgeState = "present" | "absent" | "unknown";
+
+/**
+ * Reads the MCP bridge's build state off the capability status.
+ *
+ * A missing status (the read failed, or the host has no `…/capabilities`
+ * surface) and a status without the field are both `unknown` — the console
+ * degrades to saying nothing rather than to an accusation it cannot support.
+ */
+export function mcpBridgeState(status: CapabilityStatusDto | null | undefined): McpBridgeState {
+  if (!status || typeof status.mcpInBuild !== "boolean") return "unknown";
+  return status.mcpInBuild ? "present" : "absent";
+}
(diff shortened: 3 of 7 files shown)


oxoxDev (approved):
**0 major, 1 minor, 1 question.** Approving — the minor is coordination that belongs to another PR, not a change owed here.

The table is the review. The issue asks for honesty about a build that cannot run what it manages, and names `GET …/mcp/servers/{name}/tools` as the route that already answers honestly. You checked that against the code and found it answers about **`openhuman`**, not about `mcp` — `harness/mcp.rs` compiles without the `mcp` feature, so on the `openhuman`-without-`mcp` row discovery and health probes answer *for real*, every reading on the screen looks correct, and the belt is empty. That middle row is the one an operator cannot diagnose from the console at all, and it is the row the issue's proposed fix does not cover.

Confirming that staging builds with `mcp` (so this is latent there rather than live) and that the default `docker-compose` build is the one actually affected is the part that turns a correct analysis into an actionable one. Also good that you checked CI builds the combination rather than assuming.

---

### Minor — third PR to edit `McpServersSection.tsx` while #554 still owes it frontend work

This touches `frontend/src/api/types.ts` and `frontend/src/views/connections/McpServersSection.tsx`. Your own #631 just changed both, and #554 (@M3gA-Mind, still open) adds a third `McpSource::Default` value to the DTO while `types.ts` declares `source: "manifest" | "runtime"` and this component picks its badge variant with `=== "manifest"`.

Nothing here is wrong and this should not wait on that — flagging so the next person editing the file knows a fourth change is queued behind it, and so #554's frontend fix-up is not written against a version of the component that has since moved twice.

### Question — what does `src/server/ops/capabilities.rs` expose, and to whom?

A route that reports which cargo features a build carries is exactly right for the diagnosis this PR is about. Is it behind the same admin gate as the rest of the ops plane, and does it report only the capability-relevant flags rather than the full feature set? Build composition is low-sensitivity but it is also the kind of thing that is easier to scope now than to un-expose later.

YellowSnnowmann:
@coderabbitai review

YellowSnnowmann:
Thanks @oxoxDev — answering the question with what the code actually does, and confirming the minor.

## Question — what `…/capabilities` exposes, and to whom

**Gate: any authenticated principal entitled to address the company — not the admin gate.** The route is `scoped("/capabilities", get(get_status))` with the `ScopedCompany` extractor (`src/server/ops/scope.rs:83`), which resolves the company, runs `authorize_address`, and applies the temporary-password boundary. Anonymous is a `401` — verified against a running host just now, not inferred:

```
$ curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080/api/v1/company/capabilities
401
```

That is deliberate rather than an oversight: `AdminScopedCompany` (#403) is for writes that settle something *for* the company — what credential its agents present, which third-party account they act through. This is a read, and it feeds surfaces a member can already see. Making it admin-only would put the MCP tab back to silently doing nothing for exactly the non-admin operator this PR exists to stop misleading — they would see the servers and no statement about whether anything can call them.

**Scope: four hand-picked flags, not a feature enumeration.** The DTO carries `mediaInBuild`, `composioInBuild`, `searchInBuild` and now `mcpInBuild` — and nothing else about the build. There is no report of the storage backend, `acp`, `runner`, `tinyplace`, `telegram`, `openhuman-rpc`, or any other compiled feature. Each of the four is a bare `cfg!(feature = "…")` — no env values, no credential material, no versions. (The neighbouring `*CredentialConfigured` booleans predate this PR and are also booleans, never values.)

The rule the four share, which is the scoping answer: **a flag earns its place only when a console surface would otherwise lie about what the deployment can do.** `mcpInBuild` qualifies because the management routes ship in every build while the bridge does not. A feature with no console surface has no business here, and I would push back on adding one.

If you would still rather it were admin-only, it is a one-line extractor swap and I will make it — with the caveat above about which operator loses the signal.

## Minor — the third edit to `McpServersSection.tsx`

Agreed, and worth writing down precisely for whoever picks up #554's frontend fix-up. It is not just the union that is binary — the badge is too:

- `frontend/src/api/types.ts:747` — `source: "manifest" | "runtime"`
- `frontend/src/views/connections/McpServersSection.tsx:431` — `variant={server.source === "manifest" ? "secondary" : "outline"}`

So a third `McpSource::Default` does not fail loudly on arrival: it falls through the ternary and renders with the **runtime** styling, which is precisely the badge #554 exists to distinguish. That fix-up needs both lines, and it needs writing against the current component rather than the pre-#631 one.

Nothing in this PR touches either line — my changes are a state field, one effect, a banner above the list, and the add-toast message — so the fix-up is no larger because of this PR. Order stays #554 → its own frontend follow-up.

---

Unrelated to your review: CodeRabbit never ran on this PR — it posted a review-limit notice rather than a review. I have retriggered it so the automated pass is not silently missing.

PR #646, opened 2026-08-11
docs(data-root): answer the hosted overlap question — a tenant rollout cannot overlap

CodeGhost21:
`docs/spec/runtime/data-root.md` carried the lock's one open question: whether the platform's tenant rollout ever runs a new pod over a `/data` the old one still holds — which since #536 turns a silent race into a failed boot.

**It does not.** A tenant is a StatefulSet at one replica, and a StatefulSet has no `maxSurge`: `opencompany-0` is deleted and fully terminated before its replacement is created. The one path that rolls tenants (`kubectl rollout restart statefulset -A -l app=opencompany`, in the platform's deploy workflow) goes through exactly that sequence. A `Deployment` would have been the dangerous shape — its default rolling update surges an extra pod even at one replica.

This replaces the open question with the answer, and records the two things that changed on the platform side anyway:

- the manifest builder now states the no-surge strategy explicitly instead of inheriting the API server's default, with a test that fails if the tenant workload ever becomes a `Deployment` or grows a second replica;
- a refused boot is now reported as the refusal — the manager's startup gate used to say only "did not become healthy before timeout", which reads the same for a lock refusal and a slow image pull. The tenant container runs with `terminationMessagePolicy: FallbackToLogsOnError`, so the message this lock emits reaches the pod status and the manager reads it back.

Docs only — no code in this repo changes.

Answers tinyhumansai/opencompany-microservice#15; platform side is tinyhumansai/opencompany-microservice#16.

### DIFF
--- a/docs/spec/runtime/data-root.md
+++ b/docs/spec/runtime/data-root.md
@@ -60,23 +60,38 @@ share regardless.
 
 ### Hosted deployments: the overlap window
 
-**Open question for the platform.** In hosted mode the manager runs each tenant
-as a container over `OPENCOMPANY_DATA_DIR=/data`. If a rollout ever has the new
-pod running while the old one is still alive over the same volume, the new pod
-now **fails to boot** with a configuration error where it previously started and
-silently raced.
+**Answered — a tenant rollout cannot overlap**
+(`opencompany-microservice`#15). In hosted mode the manager runs each tenant as
+a container over `OPENCOMPANY_DATA_DIR=/data`, and the question was whether a
+rollout ever has the new pod running while the old one still holds the volume —
+which since this lock exists means the new pod fails to boot where it
+previously started and silently raced.
 
-Refusing is the correct behaviour — the alternative is two writers over one
-journal — but it changes what a deploy does, so it needs a decision rather than
-a discovery:
+It does not, and by construction rather than by configuration. A tenant is a
+**StatefulSet at one replica**, and a StatefulSet has no `maxSurge`: pod
+`opencompany-0` is deleted and fully terminated before its replacement is
+created, so at most one pod ever holds the claim. The one path that rolls
+tenants — `kubectl rollout restart statefulset -A -l app=opencompany` in the
+platform's deploy workflow — goes through exactly that sequence. A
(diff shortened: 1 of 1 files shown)


oxoxDev (approved):
**0 major, 1 minor, 1 note.** Approving.

The answer is the right kind: *by construction rather than by configuration*. A StatefulSet has no `maxSurge`, so `opencompany-0` is fully terminated before its replacement exists — that is a stronger guarantee than any deploy-workflow convention, and naming `Deployment` as the shape that would have been dangerous is what makes it checkable by the next reader.

**The best part is that you did not stop at documenting it.** The invariant was incidental — true because of what the manifest builder happened to emit — and you made the platform side state the no-surge strategy explicitly with a test that fails if the tenant workload ever becomes a `Deployment` or grows a replica. Recording an accidental property as if it were a guarantee is how these rot; converting it into an enforced one and *then* writing it down is the order that holds up.

Fixing the illegibility separately is also right. "Did not become healthy before timeout" reading identically for a lock refusal and a slow image pull is exactly the kind of thing that costs an hour at the worst moment, and `terminationMessagePolicy: FallbackToLogsOnError` carrying the refusal into pod status is the small change that pays for itself once.

### Minor — name the platform-side test

The invariant this document now asserts is enforced by a test in another repository. Nothing here would notice if that test were deleted or the workload retyped. Worth naming it (file and test name) in the paragraph that cites `opencompany-microservice`#16, so a reader who wants to trust this page can confirm the guard still exists rather than taking the cross-repo claim on faith.

### Note — this answers part of an open question on #630, not all of it

#630 (`approvals: journal a start/finish bracket around every cycle`, @M3gA-Mind, since merged) has `sweep_interrupted_cycles` resting on "exactly one process owns this journal", and I asked there whether `handover.is_none()` is the whole guard. This settles the **data-root** half decisively — one pod, one volume, no overlap.

It does not cover the other half: in shared-single-DB mode the manager injects `OPENCOMPANY_TENANT_ID` and every tenant shares one logical MongoDB, where the journal is not on `/data` at all. A boot sweep there is bounded by tenant namespacing rather than by pod lifecycle. Different mechanism, still open, and worth someone confirming since #630 is already merged.

PR #648, opened 2026-08-11
fix(workflows): unbreak main — gate test asserted a reach #459 had already changed

CodeGhost21:
**`main` is red and has been since ~11:27 today.** Every run of the `Rust (openhuman, tinycortex)` lane fails on `workflows::gate::tests::a_plain_read_and_a_metered_read_do_not_gate`, so this blocks every open PR, not one. This is the smallest change that makes it green.

## What happened

Two PRs, each green against the `main` it branched from, that cannot both be right:

- **`c721e57` (#606, issue #459)** moved `read_workspace_state` from `Reach::Nothing` to `Reach::Consequence`. It shells out to `git status` / `git log` in `{root}/{company}/{agent}/workspace` — the directory `file_write` writes into — and the vendored `run_git` sets no `GIT_CONFIG_NOSYSTEM` and scrubs no environment, so a `.git/config` the agent authored decides what `core.fsmonitor` runs.
- **`afe2747` (issue #460)** added `src/workflows/gate.rs`, whose test uses `read_workspace_state` as its example of a pure workspace read that must *not* gate.

Merged together, the test asserts the opposite of the classification. Neither PR could have caught it in isolation — this is only visible in the merge, which is exactly the failure mode CI-on-`main` exists to catch and did.

## Why the test, and not the reach

Reverting `read_workspace_state` to `Reach::Nothing` would turn the lane green in one line, and it is the wrong line. That reach is a deliberate stopgap with its own revert condition recorded both in the code and upstream at `tinyhumansai/openhuman#5494`:

> Revert this to `Reach::Nothing` once a hardened `run_git` is vendored, and not before.

Undoing a security fix to keep a test assertion inverts which of the two is load-bearing. So the example changes: `file_read` is a genuine `Reach::Nothing` workspace read, and the test keeps its intent unchanged — a plain read and a metered read both pass the gate under `supervised`.

The doc comment now records why the obvious-looking tool is the wrong one, so the next person writing a "pure read" test does not reach for it again and rediscover this the same way.

## Scope

One test's fixture and its comment. No production code, no classification, no gate behaviour.

`every_reachable_workflow_tool_is_classified_by_name_alone` also names `read_workspace_state` and is deliberately left alone — it asserts that reach is stable across args, not what the reach is, so it holds either way.

## Verification

```
cargo fmt --all -- --check
cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings
cargo test  --locked --features openhuman,tinycortex --tests     # 2849 passed, 0 failed
```

Run on Rust 1.96.1 rather than the pinned `stable`: that feature set cannot build on 1.93.0, because `libsqlite3-sys 0.38.1`'s build script uses unstable `cfg_select!` and dies before this crate compiles. Worth knowing separately — it means this lane is unreproducible locally on the toolchain `rust-toolchain.toml` pins, which is part of why a merge-only breakage sat on `main` for an hour.

Found while investigating a failure on #636, which turned out to be this and not that PR.

### DIFF
--- a/src/workflows/gate.rs
+++ b/src/workflows/gate.rs
@@ -356,10 +356,17 @@ description = "Runs Acme."
     /// either would stop runs that have nothing to decide, and for `web_search`
     /// it would be worse than useless: consent for it happens once, at grant
     /// time, via an explicit `search` grant a `*` cannot confer.
+    ///
+    /// `file_read` rather than `read_workspace_state`, which reads like the
+    /// obvious choice and is not one: issue #459 moved it to
+    /// `Reach::Consequence` because it shells out to `git` in a directory the
+    /// agent can write a `.git/config` into. That is a deliberate stopgap with
+    /// its own revert condition (tinyhumansai/openhuman#5494), so the tool to
+    /// change here was the example, not the classification.
     #[tokio::test]
     async fn a_plain_read_and_a_metered_read_do_not_gate() {
         let mut g = graph(vec![
-            tool_node("read", "read_workspace_state"),
+            tool_node("read", "file_read"),
             tool_node("search", "web_search"),
         ]);
         let gated = apply_tool_call_gates(&mut g, &company("supervised", &[]), "wf", "run-1").await;


oxoxDev (approved):
**0 major.** Approving — and this is the right fix rather than the easy one.

`main` is red because #606 moved `read_workspace_state` to `Reach::Consequence` and #612 added a test asserting it does **not** gate. Both were green; they merged sixteen minutes apart and the conflict is semantic, so nothing either PR could have run would have caught it.

The tempting fix from inside `gate.rs` is to make the assertion pass — relax the expectation, or quietly reclassify the tool back. Changing the **example** instead, and leaving the classification alone, is correct: #459's reclassification exists because `read_workspace_state` shells out to `git` in a directory the agent can write a `.git/config` into, and `git status` invokes `core.fsmonitor`. `file_read` is a genuine plain read and tests the property the test is actually about.

Naming the revert condition — that the classification is a stopgap tracked at `tinyhumansai/openhuman#5494` — is what stops the next person reading this test and concluding `read_workspace_state` was reclassified in error. Good to see that issue exists; a stopgap whose removal condition lives only in a code comment is how stopgaps become permanent.

Worth a follow-up beyond this PR: the two changes were individually green and jointly broken, and nothing in CI models that. A test asserting a *specific tool* does not gate is coupled to that tool's classification, which lives in another module and moves for security reasons. Either such tests should use tools chosen for being classification-stable (as this now does), or the gate tests should assert over a `Reach` rather than a tool name, so a reclassification cannot silently invalidate them.

CodeGhost21:
For the record: superseded by `2eb19f5` (#596), which landed the same fix on `main` first — same swap to `file_read`, same reasoning that #459's reclassification is deliberate and the example is what should change. `main` is green again, so nothing here is left to add.

Issue #658, opened 2026-08-11
approvals: should `full` publish without asking? (`publish_artifact` is excluded from the #338 judgement gate)


M3gA-Mind:
## The question

Should `full` mean "publishes without asking"?

This is a product decision, not a classification bug, and it is deliberately not being taken inside #338's gate PR.

## What the code already declares

`publish_artifact` is declared in `src/policy/consequence.rs`:

```rust
// ---- Publishing --------------------------------------------------------
// Externally visible and not reversible by the company alone.
d("publish_artifact", EffectGroup::Publish, Reach::Consequence),
```

`EffectGroup::Publish` + `Reach::Consequence`, under a comment saying in as many words that it is externally visible and irreversible by the company alone.

## What #338 says

#338's headline acceptance:

> A run that would send, publish, pay, or delete stops for approval regardless of mode, unless explicitly pre-granted.

Read together, those two say `publish_artifact` should stop under `full`. #338's implementation **deliberately excludes it**, and says so in the code, its PR and on the issue — because of what follows.

## Why it was excluded rather than shipped

**There is no operator override.** #338's judgement gate has exactly one escape: a single-use, argument-exact grant per call (#243). There is no manifest knob, no `auto_approve` equivalent, nothing an operator can set once.

So stopping `publish_artifact` under `full` means: **every company running `full` stops for a human on every deliverable it publishes**, forever, with no way to opt out short of editing the tool declaration table. That ends autonomous publishing for the flow #244 shipped.

That may well be the right answer — "publishing is externally visible and irreversible, so a human should see it" is a coherent position, and it is what the declaration already claims. But it is a decision about what the product does, owned by whoever owns #244, not a side effect of a classification gate.

## Evidence that the change is bigger than its diff

Excluding it keeps 11 tests green. Shipping it turns them red: every case in `src/harness/publish_turn_test.rs`.

Those tests are not incidentally affected. Their fixture picks the mode on purpose:

```toml
[policy]
# `full` so an ordinary turn is not parked for approval — the gate under test
# here is the grant + store gate, not the approval one.
mode = "full"
```

They use `full` precisely to isolate the grant + store gate from the approval one. Making them pass under the new rule needs a grant minted per scripted call — 11 rewrites of tests that are about publish mechanics and not about approvals. A change that requires that is a change worth deciding deliberately.

## What deciding this looks like

[...]

PR #660, opened 2026-08-11
approvals: decide which calls need a human, per call (#338)

### LINKED ISSUE #338: spine: decide which calls need a human, per run
Part of the spine epic **#183**, §4. **Blocked on #242** — the classification has to be recorded against an attempt, and attempts are not objects yet.

## What is missing

The approval machinery exists and works. #243 shipped the precedence chain and the single-use, argument-exact grant; #320 shipped approve → re-dispatch. What is missing is the *judgement*: **which calls should stop for a human at all.**

Today that question is answered entirely by static configuration — reserved `never_do`, a readonly brake, `always_approve`, `auto_approve_under_usd`, then mode. All of it is decided before the run starts, by an operator writing a manifest. Nothing looks at what the run is actually about to do.

The consequence is a bad trade in both directions. Set the bar low and every run stops constantly, which is the batching pressure #183 decision 3 exists to avoid. Set it high and the run does something consequential without asking.

## What this should do

Classify each candidate call at the moment it is about to happen, and stop only for the ones that warrant it.

The signals available are richer than the static config uses:

- **Reversibility.** Sending a message, publishing, paying, deleting — these cannot be taken back. Reading, drafting, searching can.
- **Blast radius.** Who sees it. An internal note is not an email to a customer.
- **Cost.** Already partly handled by `auto_approve_under_usd`, but that is a per-call threshold with no notion of a run's cumulative spend.
- **Novelty.** The same call, same arguments, already approved once in this run is a weaker case for stopping than a call the operator has never seen.

The static config must stay authoritative where it speaks — reserved `never_do` still refuses, `always_approve` still stops. This adds judgement in the gap between those, where today the answer is a blunt mode.

## Why it blocks on #242

Two reasons, both concrete:

1. **Novelty needs a run.** "Has this operator already approved this call in this attempt" is unanswerable without a durable per-attempt record. #242 supplies it, and stamps `run_id` onto parked `Effect`s so an approval can be correlated to the attempt that raised it.
2. **The classification must be inspectable.** If a run stopped, the operator needs to see *why* it stopped, and if it did not stop, why not. That is a per-attempt trace entry — exactly what #242's `RunStepRecord` is.

Building this against today's window-heuristic correlation would produce a gate whose decisions cannot be audited.

## Design questions to settle first

- **Where does the classification live?** A property of the tool definition (declared once, honestly, by whoever wrote the tool) or a runtime judgement (a model call per candidate)? The first is cheap and static; the second is expensive and adaptive. A hybrid — declared defaults, overridable per run — is likely, but it should be argued, not assumed.
- **Does the gate ever learn?** If an operator approves the same shape of call five times, should the sixth stop? Tempting, and dangerous: it converts a security boundary into a habit. Probably no for v1, but state it.
- **What happens on classifier failure?** Fail closed (stop for a human) is the only safe default. Say so explicitly.

## Acceptance

- A run that would send, publish, pay, or delete stops for approval regardless of mode, unless explicitly pre-granted.
- A run that only reads, searches, or drafts does not stop.
- Reserved `never_do` still refuses outright — this never softens it.
- Every stop, and its reason, appears in the attempt's trace.
- A classifier that errors stops the call rather than passing it.
- Approving a stopped call consumes exactly that grant (#243 semantics), and the run may stop again later in the same attempt (#183 decision 3).

## Related

- #183 — the spine epic (§4)
- #242 — run records **(blocker)**
- #243 — approval precedence and single-use grants
- #320 — approve → re-dispatch
- #333 — a task's Approvals tab is empty because approvals carry no attempt id (also blocked on #242)

M3gA-Mind:
## Summary

Reworked under **#674's ruling**, which closed the contradiction this PR was parked on. The gate itself is unchanged; what changed is **who it speaks about**.

The "should this stop for a human?" question was answered entirely by static configuration decided before the run starts — reserved `never_do`, the `readonly` brake, a live grant, `always_approve`, the daily cap, `auto_approve_under_usd`, then the mode. Nothing looked at what the run was about to do, so `full` admitted agent-picked sends, payments, deletes and other unbounded calls without asking. Publishing is the deliberate #658 exception below.

`src/policy/judgement.rs` asks that question per candidate call, as the **last** arm of `ApprovalPolicy::check`.

### What #674 decided, and what that made this PR

#338's rule ("unbounded reach stops regardless of mode") and #614's position ("a `tool_call` node does not gate under `full`", stated in a named test) were in direct contradiction. #674 **split them by path**, and the justification is stronger than "pre-declared intent":

A workflow `tool_call` node is **refused at author time** unless its namespace is one the workflow invoker wires *and* the company's `[tools].allow` grants it. So a saved `shell` node has passed **two operator gates** — the manifest grant, then authoring — and the operator saw the command they saved. An agent turn has passed **neither**: the model picks the tool and the arguments at run time.

- **`CallPath::Agent`** — #338's rules govern. The default at every construction site, because it is the strict one.
- **`CallPath::AuthoredWorkflowNode`** — #614's position governs; this module is silent. Only `apply_policy_gates` opts in, on the private policy instance it already builds for that pass.

### The boundary condition, which is the part that matters

**A node whose arguments are templated from an upstream node's output is not pre-declared, and follows the agent rule.** The operator declared the *shape*; the *content* arrives at run time from data they never saw.

Without it the split is trivially defeated: author one `shell` node whose `command` is `=previous.output`, and the two-gate argument stops describing what actually runs while still being cited as the reason not to ask. It is implemented in `judge` rather than at the call site, so it is pinned by pure tests as well as at the gate pass — and the discriminator is not invented here: `every_reachable_workflow_tool_is_classified_by_name_alone` already exists because "node `args` may still carry unresolved `=`-expressions when this pass runs".

`a_shell_node_templated_from_upstream_output_gates_under_full` pins it at the real seam, alongside an `http_request` case whose URL is resolved at run time.

## API Or Behavior Changes

**One new public type and one new builder method.** `crate::policy::CallPath`, and `ApprovalPolicy::for_authored_workflow_nodes()`. `ApprovalPolicy::new`'s signature is unchanged and yields the strict path, so **every existing construction site is judged in full rather than silently exempted** — `the_default_path_is_the_strict_one` pins that.

**The arm can only ever ADD a stop**, and that placement is still the entire safety argument. It is consulted *only* where the mode already returned `Allow`, and its only answers are "stop" and "say nothing". `never_do`, the `readonly` brake, both grant arms, `always_approve`, the daily cap and `auto_approve_under_usd` have each already had their say and returned before it is reached.

**The path opt-out scopes exactly one arm.** `the_authored_path_changes_nothing_above_the_judgement_arm` walks three cases decided above it — a denying tier, `always_approve` parking, a parking tier — and asserts both paths agree. That matters more than it looks: on the authored-node path `always_approve` is now the operator's whole control surface, so it had better still work. A refactor that moved the path check earlier in `check` — the obvious simplification, since it would let the function return before doing any work — fails there rather than shipping a bypass.

### Changes since the previous revision

1. **`http_request` / `curl` / `web_fetch` are out of `DEFERRED`.** #674 ruled them **scoped by path**, not deferred: governed by #614 on the workflow path and #338 on the agent path. `unbounded_tools_stop` now asserts them under the rule and asserts they are *not* in `DEFERRED`, so they cannot quietly be carved out again.
2. **`publish_artifact` stays in `DEFERRED`, permanently.** #658 ruled that `full` publishes without asking, because `always_approve` is the operator's override — so this is correct behaviour, not a stopgap, and the comment now states the ruling instead of "pending #658". The test is renamed `publish_artifact_is_excluded_by_the_658_ruling` and now asserts silence on **both** paths *and* under templated arguments, because the carve-out is not part of #674's split and must survive the one condition that reopens every other node.
3. **The five pre-ruling workflow failures are resolved by the scoping**, without weakening either test that had to survive.
4. **The acceptance-verb statement is rewritten per path** — see Documentation.
5. **Rebased onto current `main`** (was `DIRTY`, 46 commits behind). `mergeStateStatus` is now `CLEAN`.
6. **The tier list is no longer enumerated in prose anywhere** — see Review follow-ups.

### The two tests that had to survive, and did

- **`the_arm_adds_nothing_under_auto`** — walks *every declared tool* and asserts anything `auto` allows, this arm leaves silent. Kept, still judged on `CallPath::Agent` (the strict path — asserting it on the lenient one would be asserting #674's split, not `auto`), and it **fails if the walk checks zero tools**.
- **`every_deferred_tool_would_otherwise_be_stopped`** — updated for the new list, not removed. It **fails if `DEFERRED` is emptied**: without that, deleting the `publish_artifact` carve-out would make it loop over nothing and score green, which is the exact failure it exists to prevent.

## Review follow-ups (@oxoxDev)

**Major — the rebase and the fourth tier: done.** Rebased onto current `main`; the branch is `MERGEABLE` / `CLEAN`. `by_mode`'s `PolicyMode::Auto` arm is in place and the branch compiles and tests against it. `the_arm_adds_nothing_under_auto` is the test you asked for, and it walks the whole declaration table rather than sampling it — so the `d_grantable("send_invoice", Send, Consequence)` day fails a test instead of shifting `auto` silently under operators who chose it.

**Minor — the tier list spelled out in three places: done, and in a fourth you had not seen.** No prose anywhere now enumerates the tiers:

- `judgement.rs` module docs — the "as it happens the arm today changes `full` alone" paragraph is gone. It now says the tier question is deliberately not enumerated here or anywhere else, and is answered by `the_arm_adds_nothing_under_auto`, the one copy that cannot go stale.
- `per-call-judgement.md` — same treatment (this file is new in this PR, so it was a fourth copy).
- Two **test** doc comments in `judgement.rs` carried the same "`supervised` parks, `readonly` denies" pair; both now phrase the invariant instead.
- `src/harness/policy.rs:3` claimed the manifest tier words are `readonly` / `supervised` / `full` "so the mapping is 1:1" — a three-item list next to a four-variant enum. That is stale on `main`, not from this branch: `1852b7d5`, the commit that added `auto`, left it behind. It is exactly the drift you were describing, so I fixed it here rather than leaving it; it now points at `PolicyMode` as the single place the tiers are listed.

`policy.rs`'s per-call-judgement section and `approvals.md` step 7 already stated only the invariant, so they were left as they were.

**Question — `publish_turn_test` overlaps #659: re-checked on the merged fixtures, and the evidence is intact.** #659 landed first (merged 2026-08-11T20:26; its merge `c5b43ab6` is in this branch's base), so this is the "whichever lands second" re-check you asked for. Its only edit to `publish_turn_test.rs` was `b60346e7`, adding `overlay_policy: None` to a helper struct literal — a mechanical field addition, not a fixture reshape. The suite has since grown to **19 cases, and all 19 pass** with the gate in place. The stale "11" and "13" counts in the `DEFERRED` comment are corrected against that measurement rather than carried forward.

## Issue #681, as an acceptance criterion

An undeclared slug on a **workflow error path** must route as an error, not park for approval. Both named tests pass with the gate in place:

- `workflows::runner::tests::t2_unknown_slug_retries_then_continues_with_error_item` — **ok**
- `workflows::runner::tests::t3_on_error_route_sends_failure_down_the_error_edge` — **ok**

They pass structurally, not by luck. A `bogus_tool` node carries literal config and no templated arguments, so `judge` returns `Silent` on the authored-node path before any rule can reach it — which leaves the run free to fail at slug resolution and take its `on_error` edge. Nothing special-cases unknown slugs; the split delivers the ordering #681 asked for. Fail-closed is untouched: the same slug on the **agent** path still stops as `StopReason::Undeclared`.

That they are a genuine check rather than incidentally green is shown by revert 9 below — removing the path split makes **both of them fail**. So #681's criterion is pinned by tests that are demonstrably sensitive to the mechanism, and no new test was needed to assert it.

## Tests

The path split adds 6 tests in `src/policy/judgement.rs`, 4 in `src/harness/policy.rs`, and 2 in `src/workflows/gate.rs`.

`the_two_paths_actually_differ` is there because every other test in the split could pass under a refactor that collapsed the two paths into one.

### Verified locally on the rebased tree

- [x] `cargo fmt --all -- --check` — exit 0, clean.
- [x] `cargo clippy --all-targets -- -D warnings` — ran as `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`; exit 0, no lints. `--no-deps` so pre-existing `vendor/tinyagents` lints that CI never sees cannot mask the result.
- [x] `cargo build --all-targets` — exit 0 (`--features openhuman,tinycortex`). Also `cargo check --locked --all-features --all-targets`, exit 0 — run because `mongodb.rs` / `sqlite.rs` build `CompanyRecord` from a row, which neither the default nor the gated lane can see.
- [x] `cargo test` — ran as `cargo test --features openhuman,tinycortex`; exit 0. **3168 passed, 0 failed, 3 ignored** in the lib target, plus 5 integration targets green.

### Revert-and-check: every rule is covered by a test that fails without it

Each row applies one revert, **proves the file actually changed** (a pattern that matches nothing is a silent no-op that would otherwise score green), runs the full gated lib suite, records the failures and restores. All twelve fail; none is uncovered. Re-run in full **after** the rebase, since a rebase can hollow a test without the diff moving.

| # | Revert | Tests failed |
|---|---|---|
| 1 | drop the `Reach` pairing on rule 1 (declared group alone decides) | 9 |
| 2 | drop the unbounded-reach rule | 10 |
| 3 | drop the money-in-arguments rule | 3 |
| 4 | fail **open** on undeclared tools | 8 |
| 5 | let the undeclared name heuristic decide rule 1 | 1 |
| 6 | remove the positive-amount filter | 1 |
| 7 | run the arm regardless of the mode's decision | 2 |
| 8 | remove the arm entirely | 8 |
| 9 | remove #674's path split (every call judged as an agent call) | 14 |
| 10 | remove the templated-argument boundary condition | 5 |
| 11 | make the templating check non-recursive (top level only) | 5 |
| 12 | default the policy to the lenient authored-node path | 5 |

Rows 9–12 are the ones that matter most for this rework: 9 and 12 prove the split exists and defaults strict, 10 and 11 prove the boundary condition is load-bearing both at the top level and at depth. Row 12 fails `the_default_path_is_the_strict_one` among four others, which is the guard against a construction site being silently exempted.

## Documentation

- `docs/spec/company-brain/per-call-judgement.md` — a **"Which path the call arrived on"** section carrying the two-gate justification and the boundary condition, and an **acceptance-verbs-per-path table**. The old sentence — "send, pay and delete are gated; publish is not" — was correct only under assumptions #674 removed; it now says which path each of send / publish / pay / delete is gated on, and states plainly that *"regardless of mode"* is exact and unchanged while *"regardless of path"* was never claimed and is now explicitly not true.
- The **Coverage** paragraph is corrected rather than carried forward. It said the workflow `tool_call` path never consults `ApprovalPolicy` (#460). #612 and #627 made that false; the path split is what governs there now, rather than the step being unreachable.
- `docs/spec/company-brain/approvals.md` — step 7 gains the path split, and the note that steps 1–6 decide identically on both paths.
- `src/policy/judgement.rs`, `src/harness/policy.rs`, `src/workflows/gate.rs` — module docs carry the same argument at the code.
- Both spec files remain under the repo's 500-line cap.

## Still open, deliberately not fixed here

**#617** — `sub_workflow` children remain ungated on the workflow path. Named in the spec, not touched.

**#681 — met as scoped, but referenced rather than closed.** Its acceptance criterion is satisfied: both runner tests pass with the gate in place, and revert 9 shows the path split is what delivers that rather than luck.

It is not closed here because one case its title still describes remains. An undeclared slug on a **templated** node follows the agent rule, so it stops as `StopReason::Undeclared` and parks instead of routing as an error — `an_undeclared_tool_stops` and `a_templated_value_is_found_at_any_depth` together pin exactly that composition. That is fail-closed working as designed rather than a regression, but the question #681 actually asked — *resolve the slug before consulting the approval gate* — is untouched by this PR. Closing it would record that question as answered when it is not. Refs #681.

Closes #338

### DIFF
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -76,6 +76,7 @@ L0  Substrate       api.tinyhumans.ai, openhuman-core, tiny.place, filesystem
 | [company-brain/README.md](company-brain/README.md) | What the company brain is; the cycle |
 | [company-brain/charter.md](company-brain/charter.md) | The company constitution |
 | [company-brain/approvals.md](company-brain/approvals.md) | Checkpoints and the approval model |
+| [company-brain/per-call-judgement.md](company-brain/per-call-judgement.md) | Which calls warrant a human, per call (step 7 of the gate) |
 | [company-brain/memory.md](company-brain/memory.md) | Long-term memory and retention |
 | [runtime/README.md](runtime/README.md) | Kernel architecture and crate layout |
 | [runtime/ports.md](runtime/ports.md) | Port trait contracts (normative) — index, assembly, defaults |
--- a/docs/spec/company-brain/README.md
+++ b/docs/spec/company-brain/README.md
@@ -7,6 +7,7 @@ that state through the `Brain` port. Cognition itself is
 it.
 
 Supporting docs: [charter.md](charter.md), [approvals.md](approvals.md),
+[per-call-judgement.md](per-call-judgement.md),
 [memory.md](memory.md).
 
 ## Definition
--- a/docs/spec/company-brain/approvals.md
+++ b/docs/spec/company-brain/approvals.md
@@ -355,7 +355,10 @@ A tool call is decided in this order:
    and it re-checks the live arguments, so a grant minted on a Composio read
    cannot admit a Composio send.
 4. `[policy].always_approve` → park.
-5. mode dispatch (`readonly` / `supervised` / `auto` / `full`).
+5. the per-agent daily cap, then `auto_approve_under_usd`.
+6. mode dispatch (`readonly` / `supervised` / `auto` / `full`).
+7. **per-call judgement** (issue #338) → park. Consulted *only* where step 6
+   allowed the call, and its only answers are "stop" and "say nothing".
 
 The grant sits **above** `always_approve` on purpose. A tool on that list still
 parks the first time, which is what the list is for; but once the operator has
@@ -367,6 +370,18 @@ narrow.
 Note this is the *tool* gate (`ApprovalPolicy`), which is a different path from
 the *effect* gate (`ManifestApprovalGate::evaluate`) the taxonomy above
 describes. A harness tool call parks directly and never reaches `evaluate`.
+
+### Per-call judgement (issue #338)
+
+Steps 1–6 are all decided before the run starts, by an operator writing a
+manifest; nothing in them looks at what the run is about to do. Step 7 closes
+that gap: `src/policy/judgement.rs` asks, per candidate call, whether it
+warrants a human on its own merits, and it can **only ever add a stop**.
+
+It is documented in full — the placement argument, the three rules, the
+no-learning boundary against #563, why it is not a model call, fail-closed, and
+the deliberate `publish_artifact` exclusion — in
+[per-call-judgement.md](per-call-judgement.md).
 
 ## Approvals inside a workflow run (issue #395)
 
--- a/docs/spec/company-brain/approvals.md
+++ b/docs/spec/company-brain/approvals.md
@@ -378,9 +378,17 @@ manifest; nothing in them looks at what the run is about to do. Step 7 closes
 that gap: `src/policy/judgement.rs` asks, per candidate call, whether it
 warrants a human on its own merits, and it can **only ever add a stop**.
 
-It is documented in full — the placement argument, the three rules, the
-no-learning boundary against #563, why it is not a model call, fail-closed, and
-the deliberate `publish_artifact` exclusion — in
+Step 7 is also the one step scoped by **which path the call arrived on** (issue
+#674): #338's rules govern an agent turn, and #614's position governs a node an
+operator authored into a saved workflow — unless that node's arguments are
+templated from an upstream node's output, which returns it to the agent rule.
+Steps 1–6 decide identically on both paths, so `always_approve` still gates a
+workflow node.
+
+It is documented in full — the placement argument, the three rules, the path
+split and its boundary condition, the acceptance verbs per path, the no-learning
+boundary against #563, why it is not a model call, fail-closed, and the
+`publish_artifact` carve-out #658 ruled correct — in
 [per-call-judgement.md](per-call-judgement.md).
 
 ## Approvals inside a workflow run (issue #395)
--- a/docs/spec/company-brain/per-call-judgement.md
+++ b/docs/spec/company-brain/per-call-judgement.md
@@ -26,14 +26,18 @@ where the mode allowed.** A tier that already parks or denies a call keeps its
 own answer, whatever it is called and however many tiers there are — so a tier
 an operator has already reasoned about does not shift under them.
 
-Today that leaves **`full` alone**. `supervised` parks a `Consequence` call,
-`readonly` denies it, and `auto` (#560) parks everything that is not
-`Grantable` — which covers every rule below, so this adds nothing there. That
-last one holds by a property of the declaration table rather than by
-construction: the day a `Grantable` + `Consequence` tool with a named
-consequence class is declared, `auto` would shift silently. The test
-`the_arm_adds_nothing_under_auto` walks the whole table and fails on that day,
-rather than the claim resting on a paragraph.
+Which tiers that leaves the arm speaking on is deliberately **not enumerated
+here, in `judgement.rs`, or in `policy.rs`**. A prose list of tiers is a copy to
+miss the next time one lands, and `auto` (#560) landing mid-review is the proof
+that they land. The question is answered in the one copy that cannot go stale:
+`the_arm_adds_nothing_under_auto` walks **every declared tool** and asserts this
+arm is silent on each one a tier still allows.
+
+Which matters most where the answer holds by a property of the declaration
+table rather than by construction. The day a `Grantable` + `Consequence` tool
+with a named consequence class is declared, a tier an operator has already
+reasoned about would shift under them silently — and that test fails that day,
+rather than the claim resting on a paragraph nobody re-reads.
 
 `full`'s contract is "act without asking, *except the few things on the
 always-ask list*", and this is what makes that exception mean something without
--- a/docs/spec/company-brain/per-call-judgement.md
+++ b/docs/spec/company-brain/per-call-judgement.md
@@ -20,15 +20,23 @@ stop**. It runs only where step 6 already allowed the call, so:
   already had their say and returned before it is reached;
 - it cannot turn a deny into a park, skip `always_approve`, or spend a grant.
 
-`supervised` and `readonly` are untouched by construction: a `Consequence` call
-already parks under one and is already denied under the other, so every call
-this would stop has already been stopped. That leaves the two autonomous tiers.
+The invariant, phrased so it survives the next tier: **this arm only ever speaks
+where the mode allowed.** A tier that already parks or denies a call keeps its
+own answer, whatever it is called and however many tiers there are — so a tier
+an operator has already reasoned about does not shift under them.
+
+Today that leaves **`full` alone**. `supervised` parks a `Consequence` call,
+`readonly` denies it, and `auto` (#560) parks everything that is not
+`Grantable` — which covers every rule below, so this adds nothing there. That
+last one holds by a property of the declaration table rather than by
+construction: the day a `Grantable` + `Consequence` tool with a named
+consequence class is declared, `auto` would shift silently. The test
+`the_arm_adds_nothing_under_auto` walks the whole table and fails on that day,
+rather than the claim resting on a paragraph.
+
 `full`'s contract is "act without asking, *except the few things on the
 always-ask list*", and this is what makes that exception mean something without
-an operator having to anticipate each one by name. `auto` (#560) already parks
-what leaves the company or spends money, so this adds only the unbounded tools
-its line does not draw — `shell` and its family. The same rule applies in both;
-no tier gets a carve-out of its own.
+an operator having to anticipate each one by name.
 
 ## What stops
 
@@ -62,13 +70,27 @@ broke publishing outright. What keeps it narrow at the other end is
 that only reads, searches, or drafts does not stop" stays true.
 
 **What it deliberately does not stop.** #338's acceptance says "send, publish,
-pay, or delete"; this delivers three of the four. **`publish_artifact` is
-excluded** (`DEFERRED` in `src/policy/judgement.rs`), pending issue #658. It is
-declared `EffectGroup::Publish` + `Reach::Consequence` and so qualifies under
-rule 1, but this gate's only escape is a per-call grant — so stopping it would
-end unattended publishing for every `full` company with no operator override,
-which is a product decision belonging to the owners of #244 rather than a side
-effect of adding a classifier. #658 carries the argument and the options.
+pay, or delete". Of those, **send, pay and delete are gated; publish is not**,
+and the outward-HTTP family is held back as well. Both carve-outs live in
+`DEFERRED` in `src/policy/judgement.rs`:
+
+- **`publish_artifact`** — issue #658. It qualifies under rule 1
+  (`EffectGroup::Publish` + `Reach::Consequence`), but this gate's only escape
+  is a per-call grant, so stopping it would end unattended publishing for every
+  `full` company with no operator override. That is a product decision owned by
+  #244, not a side effect of adding a classifier.
+- **`http_request`, `curl`, `web_fetch`** — issue #674. They qualify under rule
+  3, and #614 (merged as #627) states the opposite premise in a test named
+  `an_http_request_node_does_not_gate_under_full`. A named test is a stated
+  position; choosing between two defensible premises belongs to #614's owner.
+
+**`shell` stays gated**, which is what keeps "or delete" real: destruction has
+no `EffectGroup` of its own, `shell` is how a run deletes, and #614 does not
+touch it. `git_operations` and `run_workflow` stay too.
+
+A test asserts every deferred tool *qualifies* on the declaration and is silent
+anyway, so deleting a carve-out fails loudly rather than silently re-opening
+the question.
 
 **It does not learn.** If an operator approves the same shape of call five
 times, the sixth still stops. The module holds no state across calls at all.
--- a/docs/spec/company-brain/per-call-judgement.md
+++ b/docs/spec/company-brain/per-call-judgement.md
@@ -1,0 +1,100 @@
+# Per-call judgement
+
+How the approval gate decides which calls warrant a human *on their own merits*,
+in the gap the static configuration leaves (issue #338, spine epic #183 §4).
+
+Read [approvals.md](approvals.md) first: this is **step 7** of the precedence
+chain documented there, and the placement is most of the argument.
+
+Steps 1–6 are all decided before the run starts, by an operator writing a
+manifest. Nothing in them looks at what the run is about to do. That is a bad
+trade in both directions: a low bar stops the run constantly, and `full` lets it
+send, publish, pay or delete without asking anyone.
+
+Step 7 closes that gap. `src/policy/judgement.rs` asks, per candidate call,
+whether it warrants a human on its own merits — and it can **only ever add a
+stop**. It runs only where step 6 already allowed the call, so:
+
+- the reserved `never_do` slot, the `readonly` brake, both grant arms,
+  `always_approve`, the daily cap and `auto_approve_under_usd` have each
+  already had their say and returned before it is reached;
+- it cannot turn a deny into a park, skip `always_approve`, or spend a grant.
+
+`supervised` and `readonly` are untouched by construction: a `Consequence` call
+already parks under one and is already denied under the other, so every call
+this would stop has already been stopped. That leaves the two autonomous tiers.
+`full`'s contract is "act without asking, *except the few things on the
+always-ask list*", and this is what makes that exception mean something without
+an operator having to anticipate each one by name. `auto` (#560) already parks
+what leaves the company or spends money, so this adds only the unbounded tools
+its line does not draw — `shell` and its family. The same rule applies in both;
+no tier gets a carve-out of its own.
+
+## What stops
+
+**In order:** Three rules, in order, all derived from declarations that
+already exist rather than a new taxonomy:
+
+1. a **declared consequence class** (`Spend`, `Send`, `Sign`, `Publish`,
+   `Hire`, `Identity`) **paired with `Reach::Consequence`**. Both halves are
+   required: `web_search` is declared `Spend` because the backend bills per
+   request, but its reach is `Money` — the spend buys the call, nothing changes
+   and nothing leaves. Stopping on the group alone parks every search in the
+   company, and a parked search is a search that never happens. Media
+   generation is `Spend` + `Consequence` and does stop, because it moves real
+   money on submit;
+2. **money named in the arguments** — a positive `amount_usd`, whatever the
+   tool's own class says;
+3. **unbounded reach** — a named set (`shell`, `curl`, `http_request`,
+   `web_fetch`, `git_operations`, `run_workflow`) that is `Consequence`.
+   This is how a *delete* stops: destruction has no `EffectGroup` of its own,
+   and `shell` is how a run deletes.
+
+Rule 3 is a **list, and was briefly a derivation**. `Consequence` +
+`Standing::PerCall` looks like it names exactly this set, since #444 refuses a
+standing grant to all of it — but `PerCall` is refused for two reasons the flag
+cannot tell apart: *unbounded* (`shell`) versus *bounded but overwriting
+operator-authored state* (`workspace_write`). Deriving from it swept up the
+workspace tools, which is how a company publishes to its own note tree, and
+broke publishing outright. What keeps it narrow at the other end is
+`Grantable`: the agent's own scratch writes (`file_write`, `edit`,
+`apply_patch`, `memory_store`) are `Consequence` too and do not stop, so "a run
+that only reads, searches, or drafts does not stop" stays true.
+
+**What it deliberately does not stop.** #338's acceptance says "send, publish,
+pay, or delete"; this delivers three of the four. **`publish_artifact` is
+excluded** (`DEFERRED` in `src/policy/judgement.rs`), pending issue #658. It is
+declared `EffectGroup::Publish` + `Reach::Consequence` and so qualifies under
+rule 1, but this gate's only escape is a per-call grant — so stopping it would
+end unattended publishing for every `full` company with no operator override,
+which is a product decision belonging to the owners of #244 rather than a side
+effect of adding a classifier. #658 carries the argument and the options.
+
+**It does not learn.** If an operator approves the same shape of call five
+times, the sixth still stops. The module holds no state across calls at all.
+Consent is an operator writing a rule — that is issue #563 — and a rule can be
+read, revoked and shown. A classifier noticing a habit is none of those, and
+turning a security boundary into a frequency count is how it stops being one.
+The two mechanisms must not converge.
+
+This is also why **novelty is not a signal**, though #338 lists it among the
+four. Every use of "we have seen this before" makes the gate looser, and the
+looser second call is already governed by the single-use grant: the operator
+approves one call, redeeming it consumes it, the next identical call parks
+again. A novelty rule that skipped the second stop would quietly repeal that.
+
+**It is not a model call.** A classifier LLM on the approval path costs money
+and latency per candidate, is itself an external effect deciding whether
+external effects are allowed, and returns different verdicts on different days —
+and "why did this stop?" has to be answerable from the trace months later.
+
+**Failure is a stop.** There is no unclassified-so-allow path: an undeclared
+tool reaches `consequence_of`'s cautious fallback and a Composio slug nobody
+has classified is treated as a send, so both stop. Fail-closed is not a branch
+to remember — it falls out of reusing a classifier that is already cautious.
+
+**Coverage.** The workflow `tool_call` path never consults `ApprovalPolicy` at
+all (issue #460), so nothing here reaches it. The acceptance criterion "stops
+regardless of mode" holds on the harness path; on that path it is unreachable
+until #460 lands.
+
--- a/docs/spec/company-brain/per-call-judgement.md
+++ b/docs/spec/company-brain/per-call-judgement.md
@@ -8,8 +8,9 @@ chain documented there, and the placement is most of the argument.
 
 Steps 1–6 are all decided before the run starts, by an operator writing a
 manifest. Nothing in them looks at what the run is about to do. That is a bad
-trade in both directions: a low bar stops the run constantly, and `full` lets it
-send, publish, pay or delete without asking anyone.
+trade in both directions: a low bar stops the run constantly, and `full` lets an
+agent-picked call send, pay, delete or otherwise cross an unbounded boundary
+without asking anyone. Publishing is the deliberate #658 exception below.
 
 Step 7 closes that gap. `src/policy/judgement.rs` asks, per candidate call,
 whether it warrants a human on its own merits — and it can **only ever add a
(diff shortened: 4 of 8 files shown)


oxoxDev (dismissed):
Checked the safety argument rather than taking the description's word for it, and its core holds: `judge` is consulted only where the mode already said `Allow`, its only outputs are stop and silence, and every arm above returns before reaching it. The `Reach` pairing on rule 1 is the right fix for `web_search` — `media_generate_*` being `Spend` + `Consequence` and still stopping shows the rule discriminates rather than merely narrowing. Deferring `publish_artifact` behind a test that fails if the carve-out is deleted is the right shape for a deferral, and the note about the two tests that were *not* sensitive to placement is the most useful paragraph in the description.

**1 major, 1 minor, 1 question.**

### Major — `src/harness/policy.rs:885`: the tier enumeration is one tier out of date, and the branch does not merge

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY` — real content conflicts in `src/harness/policy.rs` and `docs/spec/company-brain/approvals.md`, 225 commits past the merge base. No Rust lane has run here; the only check is CodeRabbit reporting rate-limited. So "2732 passed" is a local result against a base that no longer exists.

The cause is #604 (`feat/560-auto-tier`), merged after this branch was cut. `main`'s `match self.mode` now has a fourth arm, `PolicyMode::Auto`, gated on `parks_under_auto()`. `let by_mode = match self.mode` has three, so it will not compile after the merge — and more to the point, the safety argument in both sets of module docs and in the description enumerates the tiers by hand: *"under `supervised` a `Consequence` call already parks and under `readonly` it is already denied"*. That enumeration is now incomplete, and it is the whole claim.

**The conclusion survives, which I checked so you don't have to.** `parks_under_auto()` is `reach.parks_under_supervision() && !standing.is_grantable()`, so the only calls `auto` allows that `supervised` would park are the `Grantable` ones — and every `d_grantable` row in the table is `EffectGroup::Other` (`file_write`, `edit`, `apply_patch`, `csv_export`, `memory_store`), plus the Composio reads. `Other` is not irreversible, none are in `UNBOUNDED`, all are declared, none carry `amount_usd`. No rule fires; `auto` really is unchanged.

But it holds by a property of the declaration table that nothing pins. The day someone adds `d_grantable("send_invoice", EffectGroup::Send, Reach::Consequence)`, `auto` shifts silently under operators who chose it — precisely the failure the placement argument exists to prevent. The rebase wants three things: the `Auto` arm in `by_mode`, one test asserting the arm adds nothing under `auto`, and a sentence in the docs naming `auto` rather than implying the tier list is closed.

### Minor — the tier list is spelled out in prose in three places

`judgement.rs` module docs, `policy.rs` module docs, and `approvals.md` each enumerate "supervised parks, readonly denies". Three copies to miss the next time a tier lands, and #604 is the proof that it happens. Phrasing the invariant instead — *the arm only speaks where the mode allowed* — stays true for any number of tiers.

### Question — `publish_turn_test` overlaps #659

The `DEFERRED` argument rests on those 11 cases staying green, and your own #659 edits `publish_turn_test.rs`. It touches neither `policy.rs` nor `src/policy/`, so there is no file conflict and no merge order to agree — but if #659 reshapes those fixtures, the deferral's evidence wants re-checking on whichever lands second.

**Checked and fine:** `consequence_of` lowercases internally, so passing the original `tool` there while handing the lowercased `name` to `declared_tools`/`UNBOUNDED` is consistent; `undeclared()` returns `Standing::PerCall`, so undeclared non-reads already park under `auto` as well; and the `write_file` → `file_write` test correction is right, since only `file_write` is declared.

Requesting changes on the rebase alone — the design is sound and I'd expect to approve it once it compiles against the fourth tier.

M3gA-Mind:
**Converted to draft. Rebased onto `0ea4ed5`, and deliberately red — blocked on #674.**

## Why this is a draft rather than a fix in progress

Main moved 41+ commits under this branch while it was being built, and two of those changed the ground it stands on:

- **#560** landed a fourth tier (`auto`). Handled: the arm now sits after the four-way dispatch, and the claim that this "changes `full` and nothing else" — written before `auto` existed — is corrected rather than carried forward.
- **#614 / #627** made the workflow `tool_call` path consult `ApprovalPolicy`. That closed the #460 gap this PR was explicitly scoped around, so the judgement gate now reaches workflow nodes. **That is the blocker.**

The textual merge was trivial — one conflict in `policy.rs`, one in a doc. Both lanes were green independently. **The merged tree fails**, which is the whole reason the rebase was re-verified from scratch rather than trusted.

## The 5 remaining failures, and which collision each belongs to

**#674 — #338's unbounded-reach rule vs #614's "a `tool_call` node does not gate under `full`"** (3 failures):

- `workflows::gate::tests::full_autonomy_leaves_the_graph_untouched` — uses `shell`
- `workflows::gated_tool_call_test::the_same_call_executes_when_the_policy_does_not_gate_it` — uses `shell`
- `workflows::gate::tests::an_http_request_node_does_not_gate_under_full` — resolved here by deferring the outward-HTTP family

This PR defers `http_request`, `curl` and `web_fetch` into `DEFERRED` (alongside `publish_artifact`/#658). **It does not defer `shell`, and will not**: `shell` is the vector by which a run deletes, so carving it out would delete the acceptance criterion this issue exists for while appearing to ship it. That makes this a design decision between two defensible premises, owned by #614's author — not something another exclusion can resolve. **#674 asks which premise governs.**

**#681 — fail-closed on an undeclared slug vs the runner's error path** (2 failures):

- `workflows::runner::tests::t2_unknown_slug_retries_then_continues_with_error_item`
- `workflows::runner::tests::t3_on_error_route_sends_failure_down_the_error_edge`

Both use `bogus_tool` deliberately, because it cannot execute. They test **error routing**, not approvals, and the gate parks the call before the runner can produce the error. Independent of #674 and wrong whichever way that goes.

## What is deliberately NOT run

**No revert-and-checks, no CI-form clippy.** Both are outstanding and both are being withheld on purpose: the rules under test are still moving, and certifying a tree that is about to change is worse than not certifying it — a test that has stopped discriminating looks like coverage. They will be run once #674 settles, on whatever the final tree is.

## What is done and stands whatever #674 decides

- `every_deferred_tool_would_otherwise_be_stopped` — each carve-out must both **qualify** on the declaration and be silent, so an entry cannot quietly become inert if the table moves under it.
- `the_arm_adds_nothing_under_auto` — walks **every declared tool** rather than asserting today's snapshot, so it fails the day a `Grantable` + `Consequence` tool with a named class is declared and `auto` would otherwise shift silently under operators who chose it. (@oxoxDev's point 1, built rather than paraphrased.)
- The tier enumeration in three prose locations is replaced by the invariant it stood in for: **this arm only ever speaks where the mode allowed.** (Point 2.)
- Docs split to `docs/spec/company-brain/per-call-judgement.md` (the repo's own remedy for the 500-line cap), indexed in both READMEs.

## One caveat on the spec text

The acceptance-verb statement now in `per-call-judgement.md` — send, pay and delete gated; publish deferred — **is correct only while `shell` stays gated**. If #674 resolves toward #614's premise, that sentence and #338's acceptance both need rewording, and this PR should not be merged with it as written.

## @oxoxDev point 3 — the #659 overlap

Noted. #659 edits `publish_turn_test.rs`, where the 11 green cases that are the `publish_artifact` deferral's evidence live. No file conflict and no merge order needed, but whichever lands second should re-check that evidence.

## CodeRabbit

Rate-limited on the previous SHA — "Review limit reached", which is **zero review, not a clean one**. Not re-triggering yet: the window is better spent on the final tree than on a SHA that is about to change.

M3gA-Mind:
## Cross-reference: #681 is a consequence of this PR's rule, and may be yours to resolve here

#681 reports that an **undeclared slug parks for approval instead of routing as an error**: `bogus_tool` is undeclared, so the fail-closed rule parks it before the runner can produce the error it is testing. That breaks `t2_unknown_slug_retries_then_continues_with_error_item` (`src/workflows/runner.rs:1548`) and `t3_on_error_route_sends_failure_down_the_error_edge` (`:1600`) — tests about **error routing**, not approvals, which use a deliberately bogus slug precisely because it cannot execute.

Fail-closed turns "unknown tool → error item" into "unknown tool → parks for a human". Defensible for an agent; wrong for a workflow's error path.

**It does not reproduce on `main`** — `Undeclared` has zero hits in `src/` there, and both tests pass. It is a property of *this branch*, which is why it is flagged here rather than left as an independent issue.

The suggested direction on #681 is **resolve the slug before parking** — correcting the ordering rather than weakening the rule. That is a change to this PR's own implementation, so you may prefer to handle it here while you hold the context, rather than have it land as a follow-up PR reconciling with this one.

Noting also that this PR is being re-scoped under **#674**'s split-by-path ruling (the arm governs the agent path; #614's position governs saved workflow nodes). If that re-scoping means the rule no longer reaches this path, #681 may resolve itself — worth checking rather than fixing twice.

M3gA-Mind:
**Reworked under #674's ruling and pushed at `4819d06`. Still a draft, deliberately — the verification is incomplete, not the code.**

## What changed

The gate is unchanged; what changed is **who it speaks about**.

#674 split #338's rules from #614's position **by path**. A workflow `tool_call` node is refused at author time unless its namespace is one the invoker wires *and* `[tools].allow` grants it — so a saved `shell` node has passed **two operator gates** and the operator saw the command. An agent turn has passed neither.

- `CallPath::Agent` — #338's rules govern. The default at every construction site, because it is the strict one; `ApprovalPolicy::new`'s signature is unchanged.
- `CallPath::AuthoredWorkflowNode` — #614's position governs. Only `apply_policy_gates` opts in, via `for_authored_workflow_nodes()`, on the private policy instance it already builds for that pass.

**The boundary condition, which is the part that would have been easy to skip.** A node whose arguments are templated from an upstream node's output is **not** pre-declared and follows the agent rule — the operator declared the *shape*, the content arrives at run time. Without it the split is defeated by one line of authoring: a `shell` node whose `command` is `=previous.output`. It lives in `judge` rather than at the call site, so it is pinned by pure tests *and* at the gate pass.

The discriminator is not invented here: `every_reachable_workflow_tool_is_classified_by_name_alone` already exists because node args "may still carry unresolved `=`-expressions when this pass runs".

## The three consequences #674 named

1. **`http_request` / `curl` / `web_fetch` are out of `DEFERRED`** — scoped by path, not deferred. `unbounded_tools_stop` now asserts they are *not* in `DEFERRED`, so they cannot drift back.
2. **`publish_artifact` stays, permanently.** #658 ruled `full` publishes without asking because `always_approve` is the operator's override. The comment states the ruling instead of "pending", and the test is renamed `publish_artifact_is_excluded_by_the_658_ruling` — now asserting silence on **both** paths and under templated arguments, since the carve-out is not part of the split.
3. **The acceptance-verb statement is rewritten per path**, as a table in `per-call-judgement.md`. "Regardless of mode" is exact and unchanged; "regardless of path" was never claimed and is now explicitly not true.

**The 5 failing tests pass**, with no change to any of them.

## The two tests that had to survive

- `the_arm_adds_nothing_under_auto` — kept, still judged on the strict path, and now **fails if the walk checks zero tools**.
- `every_deferred_tool_would_otherwise_be_stopped` — updated, not removed, and now **fails if `DEFERRED` is emptied**. Without that, deleting the carve-out would make it loop over nothing and score green.

## Verification — what is done, and what is NOT

**Done:**
- `cargo test --features openhuman,tinycortex --lib` → **2998 passed, 0 failed**, 3 ignored
- `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` → clean
- `cargo fmt --all` → applied

**NOT done, and the reason this stays a draft:**
- **`cargo check --locked --all-features --all-targets` was never run.** This is the lane that catches the MongoDB/SQLite backends, which neither the default nor the gated lane compiles.
- No integration targets under `tests/`, no `cargo build --all-targets`, no manual app run.
- **Revert-and-check is 4 of 12**, stopped mid-run at session end. All four are covered and none vacuous:

| Revert | Result |
|---|---|
| 1. drop the `Reach` pairing on rule 1 | **9 failed** / 2989 passed |
| 2. drop the unbounded-reach rule | **10 failed** / 2988 passed |
| 3. drop the money-in-arguments rule | **3 failed** / 2995 passed |
| 4. fail **open** on undeclared tools | **8 failed** / 2990 passed |

Revert 1 is materially stronger than the previous revision claimed — recorded there as catching one test, it takes down 9 here, including `the_arm_adds_nothing_under_auto`. Reverts 5–12 are unrun: the six originals not listed above, plus removing the path split, removing the boundary condition, making the templating check non-recursive, and defaulting the policy to the lenient path.

The harness runs the **full** suite with no positional filter, treats a missing `test result:` line as `INVALID — proves nothing` rather than a pass, and `cmp`-verifies each revert actually changed the file before building, because a pattern matching nothing is a no-op that scores green.

## @oxoxDev — your three points

**Major (tier enumeration one tier out of date; branch does not merge).** The `Auto` arm is in `by_mode`, and `the_arm_adds_nothing_under_auto` walks every declared tool rather than snapshotting today's table — built as you asked rather than paraphrased. On the merge half: **zero Rust files changed on `main`** in the 18 commits since the branch point (they are all frontend/design-system), so the outstanding rebase carries no Rust risk.

**Minor (tier list spelled out in prose in three places).** Replaced by the invariant — *this arm only ever speaks where the mode allowed* — in all three, and the new #674 section uses the same phrasing rather than reintroducing the enumeration.

**Question (`publish_turn_test` overlaps #659).** Checked rather than noted. #659's **only** change to `publish_turn_test.rs` is one struct field, `overlay_policy: None` — the fixture, its `full` mode and all 11 cases are untouched, so the deferral's evidence holds in either merge order. #659 adds `overlay_policy` to `CompanyRecord` and covers **all 44** files on `main` carrying a `CompanyRecord` literal, including `src/workflows/gate.rs`. **This branch adds zero `CompanyRecord` literals**, so there is no merge order to agree and no semantic-merge hazard in either direction.

## CodeRabbit

The only CodeRabbit review on this PR reads **"Review limit reached … we couldn't start this review"**. That is **zero reviews**, not a clean one. Not re-triggered yet — the tree is about to be rebased.

## #681

Its two named tests — `t2_unknown_slug_retries_then_continues_with_error_item` and `t3_on_error_route_sends_failure_down_the_error_edge` — **pass on this tree**, verified by running them, not predicted. That is the split doing it rather than a fix: a `bogus_tool` node carries no templated arguments, so the scoped arm never reaches it.

**A narrow residue survives**, so it should not close silently: an authored node with an undeclared slug **and** a templated argument is returned to the agent rule by the boundary condition, and still parks. Parking there is arguably still wrong — an unresolvable slug can never run, so the operator gets a card for a call that cannot happen. Not reachable by either of #681's tests. Fixing it needs slug resolution against `WORKFLOW_TOOL_NAMESPACES` inside the gate pass, which is real scope creep on a PR this size.

Refs #338

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@oxoxDev — rebased and reworked under #674's ruling. Your three items, in order.

**Major — the rebase and the fourth tier.** Done. The branch was 46 commits behind; it is now rebased onto current `main` and reports `MERGEABLE`/`CLEAN`. `by_mode` has its `PolicyMode::Auto` arm and the branch compiles and tests against it. `the_arm_adds_nothing_under_auto` is the test you asked for, and it walks the whole declaration table rather than sampling it — so the `d_grantable("send_invoice", Send, Consequence)` day fails a test rather than shifting `auto` silently. Thank you for checking the conclusion; it did survive, but you were right that nothing pinned it.

**Minor — the tier list in three places.** Done, and it was in a fourth you had not seen. No prose anywhere enumerates the tiers now:

- `judgement.rs` module docs — the "as it happens the arm today changes `full` alone" paragraph is gone; it now says the question is deliberately not enumerated and is answered by `the_arm_adds_nothing_under_auto`.
- `per-call-judgement.md` — same treatment. This file is new in this PR, so it was a fourth copy created by the very change you were reviewing.
- Two **test** doc comments in `judgement.rs` carried the same "`supervised` parks, `readonly` denies" pair; both now phrase the invariant.
- `src/harness/policy.rs:3` said the manifest tier words are `readonly`/`supervised`/`full` "so the mapping to `PolicyMode` is 1:1" — a three-item list beside a four-variant enum. That one is stale on `main`, not from this branch: `1852b7d5`, the commit that added `auto`, left it behind. It is exactly the drift you were describing, so I fixed it here rather than walking past it; it now points at `PolicyMode` as the single place the tiers are listed.

`policy.rs`'s per-call-judgement section and `approvals.md` step 7 already stated only the invariant, so they were left alone.

**Question — `publish_turn_test` overlaps #659.** Re-checked on the merged fixtures, and the evidence holds. #659 landed first (merge `c5b43ab6` is in this branch's base), so this is the "whichever lands second" check. Its only edit to `publish_turn_test.rs` was `b60346e7`, adding `overlay_policy: None` to a helper struct literal — a mechanical field addition, not a fixture reshape. The suite has grown to **19 cases and all 19 pass** with the gate in place. The `DEFERRED` comment's stale "11" and "13" are corrected against that measurement rather than carried forward.

**On "2732 passed is a local result against a base that no longer exists" — fair, and re-established.** On the rebased tree: `cargo check --locked --all-features --all-targets` exit 0; `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` exit 0; `cargo test --features openhuman,tinycortex` exit 0 with **3168 passed, 0 failed**. The 12-case revert harness was re-run in full **after** the rebase — a rebase can hollow a test without the diff moving — and all twelve fail, none uncovered; the per-revert failure counts are in the PR body. This PR was a draft its whole life, so CI has never run on it; that is running now and is the first non-local signal.

oxoxDev (approved):
Re-reviewed at `5f1dd0ba`. The rebase landed on current `main` (base is #699's merge, **0 behind**), the branch is `MERGEABLE` / `CLEAN`, and all nine checks are green. Restoring my approval.

**The major is resolved.** `PolicyMode::Auto` has its arm in `by_mode`, so the match is exhaustive and the branch compiles against the four-tier world it now lives in.

**And I re-checked the claim rather than assuming it survived**, because the whole point of my earlier note was that it rested on a property of the declaration table rather than on anything enforced. Against `main` as it stands today, all five `d_grantable` rows are still `EffectGroup::Other` + `Reach::Consequence` (`file_write`, `edit`, `apply_patch`, `csv_export`, `memory_store`). `parks_under_auto()` is `reach.parks_under_supervision() && !standing.is_grantable()`, so the only calls `auto` admits that `supervised` would park are the `Grantable` ones — and `Other` is not irreversible, none of them are in `UNBOUNDED`, all are declared, and none carries `amount_usd`. No rule fires. **`auto` is genuinely unchanged**, today, in fact and not just in argument.

Good to see #658 settled and the `DEFERRED` comment updated to say so rather than left pointing at an open question.

**0 major. 1 minor. Approving.**

### Minor — the `auto` property is still the one thing nothing pins

Of the three things I asked for on the rebase, the arm landed and the test did not. There is no case asserting the judgement arm adds nothing under `auto` — `a_sub_threshold_spend_is_still_auto_approved` covers `auto_approve_under_usd`, which is a different mechanism, and the rest of the new table is `full` / `readonly` / `supervised`.

I am not blocking on it, because I have now verified the behaviour is correct as of this commit. But the reason to want it is unchanged and is the reason I raised it: the correctness is a coincidence of the table, not a property of the code. The day somebody writes

```rust
d_grantable("send_invoice", EffectGroup::Send, Reach::Consequence),
```

`auto` silently starts stopping calls it did not stop before — a tier an operator deliberately chose, shifting under them with no test failing and no reviewer prompted to think about it. That is precisely the failure the placement argument in your module docs exists to prevent, one level out.

One test in the shape of the ones you already have — a `Grantable` + `Consequence` tool under `auto`, asserting the decision is whatever the mode said and carries the mode's own reason — turns "true today" into "fails loudly when it stops being true". It is a handful of lines against a module you have otherwise pinned unusually thoroughly.

Worth also confirming the module docs now name `auto` explicitly rather than implying the tier list is closed; I did not verify that half, and it is the thing that stops the next reader re-deriving a three-tier argument.

The new authored-workflow-node scope (`for_authored_workflow_nodes`, and the two tests around it) arrived after my last pass and I have not reviewed it in depth — the gated lane is green and it is pinned on both sides, so I am taking it on that.

Issue #464, opened 2026-08-06
console: a card opened from chat never appears on the board until you reload


oxoxDev:
## What happens

Ask for something in chat, the reply says **Card opened**, the card's own page loads fine — and the board shows nothing. It stays empty until the console is reloaded.

Reported from a staging tenant, 2026-08-06, immediately after the card-by-construction work shipped.

## Why it happens

Two gaps that compound:

**1. The host emits no card-created event.** The SSE switch handles `task_dispatched`, `task_steered` and `lifecycle_changed` — all of which describe a card that already exists. Nothing is emitted when a card is *opened*, so there is no event for a board to listen to.

**2. The board subscribes to nothing.** `app-shell` maintains a `taskEventTick` counter bumped on every task-lifecycle event, but it is passed to the chat view, not to the board. `TasksView` never references it.

**Correction to an earlier revision of this issue.** It claimed the board "lists tasks once, on mount". That is **wrong** — `TasksView` has polled every 4000ms since #27, and on a live host the card from chat was measured appearing **without a reload in 1608ms**. Anyone acting on the original wording would go looking for a missing fetch that is not missing.

What is true, and is the defect: the board's **only** live path is a fixed-interval poll — which is precisely the answer this issue rules out — and it subscribes to no event at all. The most likely route to the reported "nothing until I reloaded" is timer throttling in a backgrounded tab, which could not be forced in headless Chromium and is therefore **not proven**.

## Why it matters

This is the promise the card work exists to keep — *ask for something, watch it become work on the board* — failing at the last step. The card is real, correctly assigned, and reachable by URL. It simply never shows up where the operator was told to look.

It reads exactly like the defect that was just fixed (work happening with nothing on the board) even though the cause is entirely different, which makes it expensive to diagnose twice.

## What the fix should do

- Emit an event when a card is created, whatever created it — chat intake, delegation, publish, the REST detector.
- The board subscribes to it and to the lifecycle events it already ignores, so a card appearing, moving column, or changing assignee is reflected without a reload.
- One subscription, not a poll.

## How to verify

- Asking for something substantial in chat makes the card appear on the board with no reload.
- Dragging a card in one browser tab moves it in a second tab.
- A card that settles into a new column moves there live.
- No polling is introduced.

## Related

Same family as #384 (workflow create/update/delete events fall through the SSE switch) and #327 (no push channel, so an agent-written file appears only on refresh). Three surfaces, one missing pattern: the host does not announce creation, and the view does not subscribe.

PR #670, opened 2026-08-11
fix(console): stop polling the company and the board while the tab is hidden (#581)

### LINKED ISSUE #581: console: the whole company graph is refetched every 5 seconds, in every tab, whether or not anyone is looking
## Symptom

The console gets progressively less responsive the longer it is open, and worst after approving several things and switching tabs. Reported in QA on 2026-08-10: *"Site hangs a lot while navigating and especially if I approve multiple things and try to switch tabs."*

## Where it comes from

The console polls, on timers, and the largest of those timers never stops.

| Timer | Interval | Pauses when hidden? |
| --- | --- | --- |
| `frontend/src/hooks/use-company.ts:52` — the whole GraphQL `Company` aggregation | 5s (`POLL_MS`, line 6) | **no** |
| `frontend/src/views/TaskDetailView.tsx:355` | 4s (`POLL_MS`, line 120) | **no** |
| `frontend/src/views/TaskDetailView.tsx:417` | — | **no** |
| `frontend/src/views/TaskDetailView.tsx:1703` | 4s | **no** |
| `frontend/src/views/ArtifactsTab.tsx:172` | 4s | yes — `visibilitychange` at line 186 |
| `frontend/src/views/WorkspaceView.tsx` | on demand | yes — `visibilitychange` at line 331 |

`use-company.ts` is the expensive one. It refetches the **entire `Company` aggregation** — team, desks, chats, tasks, skills, workspace, memory, workflows, usage, finances, connections, domain, SMTP — every five seconds, in every open tab, forever. A background tab left open overnight is still doing it.

Task Detail then stacks three more timers on top, and an open card polls at 4s while the company graph polls at 5s.

## Why approvals make it worse

Approving parks and resumes work, which changes task state, which is inside the aggregation being refetched. So the periods where the operator is most actively clicking are the periods where the refetches are largest and least likely to be cache hits.

## The change

**The fix pattern is already in this codebase** — `ArtifactsTab` and `WorkspaceView` both gate their refresh on `document.visibilitychange`. Apply the same discipline to the two that don't:

1. `use-company.ts` stops polling while `document.hidden`, and refreshes once on becoming visible.
2. The three `TaskDetailView` timers do the same.
3. Consider whether the company poll needs the whole aggregation. Most of what it refetches — team, skills, domain, SMTP — does not change on a 5s cadence, and the live feed (`use-events.ts`) already pushes the parts that do.

Point 3 is the real win; points 1 and 2 are the cheap ones.

## Acceptance

- A hidden tab issues no company or task poll.
- Returning to a hidden tab refreshes once, promptly.
- With three tabs open on the same company, steady-state request volume is a fraction of today's.

oxoxDev:
## Summary

The console armed three background reads with a bare `setInterval` and never stopped them. A browser throttles a hidden tab's timers, it does not cancel them, so a console left open in a background tab kept asking the host for the company status and the whole task board forever — and every extra tab multiplied the steady-state request volume of a company nobody was looking at. That is the "site hangs while navigating, especially after approving things and switching tabs" report in #581.

This gates them on `document.visibilitychange`, using the pattern `ArtifactsTab` and the task detail screen already hand-rolled, extracted into one helper so a fourth caller cannot get it subtly wrong.

**Two of the issue's three points did not survive contact with the code, and this PR deliberately does not implement them:**

- The issue describes the 5s poll as refetching "the whole GraphQL `Company` aggregation — team, desks, chats, tasks, skills, workspace, memory, workflows, usage, finances, connections, domain, SMTP". There is no GraphQL in the frontend at all. `client.status()` is `GET /api/v1/companies/{id}`, and `CompanyStatus` is six scalar fields (`src/runtime/types.rs`) served by one `store.load()` plus a pending count (`src/company/runtime.rs`). So the issue's point 3 — "consider whether the company poll needs the whole aggregation", called out as "the real win" — has no aggregation to trim. It is dropped for lack of a subject rather than deferred.
- The three `TaskDetailView.tsx` timers the issue asks to gate (its point 2) are **already** gated on `main`, at lines 357, 419 and 1727. They are untouched here.

What the issue *missed* is that `TasksView` re-reads the entire board every 4s, ungated — the most expensive of the timers involved — and that `ApprovalsView` runs a third ungated interval. Both are fixed here.

## API Or Behavior Changes

No API change. Frontend only; no backend surface participates in tab visibility, and the polled endpoints are already cheap reads.

Behavior:

- A hidden tab issues **no** company-status, approvals, board or standing-permissions poll.
- Returning to a hidden tab performs exactly one immediate catch-up read per gated poller, then resumes its normal cadence.
- Foreground cadence is unchanged (5s status + approvals, 4s board, 60s standing permissions).
- The board's push half (#464) is gated too, or the poll gate would buy nothing on a busy company. The pending event tick is deliberately **not** consumed while hidden, so the update is deferred rather than dropped, and the poller's load-on-visible read settles it.

Deliberately not changed: `WorkflowsView`'s 2s run tick (armed only during an active run, self-clearing, and the SSE fallback for deployments where frames do not arrive — gating it risks wedging the Run controls for a small win), and the already-gated `ArtifactsTab` / `WorkspaceView` / `TaskDetailView` timers. Collapsing those onto the new helper is a reasonable follow-up, kept out of here to keep the diff reviewable.

## Tests

Frontend-only change, so the Rust gates are not applicable — this PR touches no Rust and no `Cargo.toml`.

- N/A: `cargo fmt --all -- --check` — no Rust changed.
- N/A: `cargo clippy --all-targets -- -D warnings` — no Rust changed.
- N/A: `cargo build --all-targets` — no Rust changed.
- N/A: `cargo test` — no Rust changed.

Run from `frontend/`:

- [x] `npm run typecheck` — clean
- [x] `npm run typecheck:unit` — clean
- [x] `npm test` (`vitest run`) — `Test Files 38 passed (38) · Tests 417 passed (417)`, including the new `visible-poll.test.ts`
- [x] `npm run build` — succeeds (only the pre-existing >500 kB chunk-size advisory)

New unit coverage — `frontend/test/unit/visible-poll.test.ts`, six cases under jsdom with fake timers: ticks fire while visible; nothing fires while hidden; becoming visible loads exactly once and resumes the same cadence; a visible→visible event does not double-arm; the disposer clears the timer and removes the listener; started-while-hidden arms nothing until the first visible transition.

**Untested edge, stated deliberately** per the repo's coverage expectation: the `TasksView` push-effect hidden guard is not unit-tested. It needs React to exercise, and this suite is pure-function only by its own `vitest.config.ts` documentation. It is verified manually (below). The guard returns **before** `seenTick.current` is assigned — moving it below that line would silently reintroduce dropped updates, so it is worth a reviewer's eye.

Manual verification, DevTools Network filtered to `/api/v1/`: foreground cadence unchanged; hiding the tab for 60s produces zero requests, including no board reads while task changes are driven from a second tab; returning produces one status+approvals burst and one board read, then a single (not doubled) cadence; switching company restarts the poller cleanly; with three tabs open on one company, only the visible tab appears in the request log.

## Documentation

No docs change. The helper carries its own rationale in a module-level doc comment — why it does not load on start, why the double-arm guard exists, and why the disposer must remove the listener — and the two subtle call sites (`use-company`, and the `TasksView` push guard) are commented against #581 in place. No route, API surface, launcher behavior or `tiny*` integration changed, so nothing in `docs/spec/` or `docs/modules/` is affected.

Closes #581

## Summary by CodeRabbit

- **Performance Improvements**
  - Background refreshes for company data, approvals, and tasks now pause when the app is not visible, reducing unnecessary activity.
  - Refreshing resumes automatically when the app becomes visible, including an immediate update to show current information.
  - Task updates received while the app is hidden are handled when visibility returns, helping prevent missed changes.
- **Bug Fixes**
  - Prevented duplicate refresh timers and ensured polling stops cleanly when no longer needed.

### DIFF
--- a/frontend/src/hooks/use-company.ts
+++ b/frontend/src/hooks/use-company.ts
@@ -2,6 +2,7 @@ import { useCallback, useEffect, useRef, useState } from "react";
 
 import type { OpenCompanyClient } from "@/api/client";
 import type { ApprovalSummary, CompanyStatus } from "@/api/types";
+import { startVisiblePolling } from "@/lib/visible-poll";
 
 const POLL_MS = 5000;
 
@@ -46,13 +47,20 @@ export function useCompany(
     }
   }, [client, company]);
 
+  // Issue #581: gated on visibility. This hook has one instance per open
+  // console tab, and its refresh is two requests — so an ungated interval meant
+  // every tab left in the background kept costing the host 24 requests a minute
+  // for a company nobody was looking at. The mount fetch stays here rather than
+  // moving into the poller: the poller deliberately does not load on start, and
+  // the hidden → visible catch-up read it does perform covers the staleness of
+  // a tab coming back.
   useEffect(() => {
     mounted.current = true;
     void refresh();
-    const timer = setInterval(() => void refresh(), POLL_MS);
+    const dispose = startVisiblePolling(() => void refresh(), POLL_MS);
     return () => {
       mounted.current = false;
-      clearInterval(timer);
+      dispose();
     };
   }, [refresh]);
 
--- a/frontend/src/lib/visible-poll.ts
+++ b/frontend/src/lib/visible-poll.ts
@@ -8,10 +8,11 @@
  * forever, and every extra tab multiplied the steady-state request volume of a
  * company nobody was looking at.
  *
- * The shape here is not new: `ArtifactsTab` and the task detail screen already
- * hand-rolled it. This is that pattern extracted so a fourth caller cannot get
- * it subtly wrong — the double-arm guard and the listener removal are the two
- * halves people forget, and forgetting either turns a fix into a leak.
+ * The shape here is not new: `ArtifactsTab` and the task detail screen
+ * hand-rolled it first. Both now call this instead, so the pattern lives in one
+ * place and the next caller cannot get it subtly wrong — the transition-edge
+ * check and the listener removal are the two halves people forget, and
+ * forgetting either turns a fix into a leak.
  *
  * Deliberately framework-free: no React import, no hook rules, so it is
  * callable from an effect, from a class, or from a plain test with fake timers.
@@ -21,12 +22,19 @@
  * Every caller already fetches once on mount — that is what fills the screen
  * before the first tick. Loading here as well would double every mount and
  * every company switch. What it *does* do is load once on the hidden → visible
- * transition, because that is the moment the view is known to be stale by up to
- * however long the tab sat in the background.
+ * *transition*, because that is the moment the view is known to be stale by up
+ * to however long the tab sat in the background.
  *
- * @param load       Called on each tick, and once when the tab becomes visible.
- *                   Errors are the caller's to handle (all current callers
- *                   swallow transient failures to keep the last good view).
+ * The transition, not the state: `visibilitychange` can fire visible→visible in
+ * some browsers, and a handler that only reads the live `visibilityState` would
+ * treat each of those as a return from the background and pay for a full
+ * refresh — two requests, for the callers that read two things. So the previous
+ * state is remembered and only a genuine edge does any work.
+ *
+ * @param load       Called on each tick, and once on each hidden → visible
+ *                   transition. Errors are the caller's to handle (all current
+ *                   callers swallow transient failures to keep the last good
+ *                   view).
  * @param intervalMs The cadence, in milliseconds, while the tab is visible.
  * @returns A disposer that clears the timer **and** removes the
  *          `visibilitychange` listener. Call it from effect cleanup; not
@@ -42,27 +50,33 @@ export function startVisiblePolling(load: () => void, intervalMs: number): () =>
     }
   };
 
-  // Guarded against double-arming: `visibilitychange` can fire visible→visible
-  // in some browsers, and arming twice would leak the first timer and double
-  // the cadence with no way left to clear it.
+  // Guarded against double-arming anyway: arming twice would leak the first
+  // timer and double the cadence with no way left to clear it, and belt and
+  // braces is cheap for something with no way to observe it went wrong.
   const start = () => {
     if (timer === undefined) {
       timer = window.setInterval(() => load(), intervalMs);
     }
   };
 
+  // The remembered previous state is what makes this an edge and not a level.
+  // It also carries the started-in-a-hidden-tab case: a console restored into a
+  // background tab, or a route mounted behind one, arms nothing until the first
+  // real transition to visible.
+  let wasHidden = document.visibilityState === "hidden";
+
   const onVisibility = () => {
-    if (document.visibilityState === "hidden") {
+    const hidden = document.visibilityState === "hidden";
+    if (hidden) {
       stop();
-    } else {
+    } else if (wasHidden) {
       load();
       start();
     }
+    wasHidden = hidden;
   };
 
-  // Started in a hidden tab — a console restored into a background tab, or a
-  // route mounted behind one — arms nothing until it is actually looked at.
-  if (document.visibilityState !== "hidden") start();
+  if (!wasHidden) start();
   document.addEventListener("visibilitychange", onVisibility);
 
   return () => {
--- a/frontend/src/lib/visible-poll.ts
+++ b/frontend/src/lib/visible-poll.ts
@@ -1,0 +1,72 @@
+/**
+ * Interval polling that stops while the tab is hidden (issue #581).
+ *
+ * The console's background reads — company status, the task board, standing
+ * permissions — were armed with a bare `setInterval` that never stopped. A
+ * browser throttles a hidden tab's timers, it does not cancel them, so a
+ * console left open in a background tab kept asking the host for the board
+ * forever, and every extra tab multiplied the steady-state request volume of a
+ * company nobody was looking at.
+ *
+ * The shape here is not new: `ArtifactsTab` and the task detail screen already
+ * hand-rolled it. This is that pattern extracted so a fourth caller cannot get
+ * it subtly wrong — the double-arm guard and the listener removal are the two
+ * halves people forget, and forgetting either turns a fix into a leak.
+ *
+ * Deliberately framework-free: no React import, no hook rules, so it is
+ * callable from an effect, from a class, or from a plain test with fake timers.
+ *
+ * ## It does NOT load on start
+ *
+ * Every caller already fetches once on mount — that is what fills the screen
+ * before the first tick. Loading here as well would double every mount and
+ * every company switch. What it *does* do is load once on the hidden → visible
+ * transition, because that is the moment the view is known to be stale by up to
+ * however long the tab sat in the background.
+ *
+ * @param load       Called on each tick, and once when the tab becomes visible.
+ *                   Errors are the caller's to handle (all current callers
+ *                   swallow transient failures to keep the last good view).
+ * @param intervalMs The cadence, in milliseconds, while the tab is visible.
+ * @returns A disposer that clears the timer **and** removes the
+ *          `visibilitychange` listener. Call it from effect cleanup; not
+ *          calling it leaks a listener per mount.
+ */
+export function startVisiblePolling(load: () => void, intervalMs: number): () => void {
+  let timer: number | undefined;
+
+  const stop = () => {
+    if (timer !== undefined) {
+      window.clearInterval(timer);
+      timer = undefined;
+    }
+  };
+
+  // Guarded against double-arming: `visibilitychange` can fire visible→visible
+  // in some browsers, and arming twice would leak the first timer and double
+  // the cadence with no way left to clear it.
+  const start = () => {
+    if (timer === undefined) {
+      timer = window.setInterval(() => load(), intervalMs);
+    }
+  };
+
+  const onVisibility = () => {
+    if (document.visibilityState === "hidden") {
+      stop();
+    } else {
+      load();
+      start();
+    }
+  };
+
+  // Started in a hidden tab — a console restored into a background tab, or a
+  // route mounted behind one — arms nothing until it is actually looked at.
+  if (document.visibilityState !== "hidden") start();
+  document.addEventListener("visibilitychange", onVisibility);
+
+  return () => {
+    stop();
+    document.removeEventListener("visibilitychange", onVisibility);
+  };
+}
--- a/frontend/src/views/ApprovalsView.tsx
+++ b/frontend/src/views/ApprovalsView.tsx
@@ -23,6 +23,7 @@ import { Button } from "@/components/ui/button";
 import { Card, CardContent } from "@/components/ui/card";
 import type { CompanyFeed } from "@/hooks/use-company";
 import { approvalSummary, timeAgo, toolAction } from "@/lib/language";
+import { startVisiblePolling } from "@/lib/visible-poll";
 import { isRecord, parseNodeMessages } from "@/views/workflows/run-output";
 
 /**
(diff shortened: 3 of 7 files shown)


M3gA-Mind (commented):
Reviewed manually (`gh pr diff` plus the three existing `visibilitychange` call sites at the PR head) rather than through our review harness — its `sync` step is openhuman-resident and I didn't want it checking out branches in a clone another agent is using. I read CodeRabbit's pass first; it generated no actionable comments, so nothing below repeats it. **Comment only, no approval.**

This is a good change and the tests are unusually well aimed — asserting the properties (a hidden tab costs nothing; coming back is worth exactly one read) rather than "it polls" is the right instinct, and `stops the timer and drops the listener when disposed` catching the listener half by behaviour instead of by spying is the one people skip. The `TasksView` push-half reasoning is the subtlest thing here and I think it's correct: not consuming the tick is right, because the effect doesn't re-run on a visibility change, so marking it seen would strand it with nothing to bring it back — and `load`-on-visible settles the staleness instead. Worth the comment it got.

Two things.

### 1. `load()` fires on every non-hidden `visibilitychange`, not only on hidden → visible

```ts
const onVisibility = () => {
  if (document.visibilityState === "hidden") {
    stop();
  } else {
    load();
    start();
  }
};
```

`start()` is guarded against double-arming, and the comment above it says why:

> `visibilitychange` can fire visible→visible in some browsers, and arming twice would leak the first timer and double the cadence

That reasoning applies to `load()` too, but `load()` isn't guarded. So in exactly the browser behaviour the code anticipates, each redundant event costs a full refresh — and in `useCompany` a refresh is two requests. It's a smaller version of the waste #581 is about: a tab nobody interacted with issuing reads because the browser emitted an event.

The fix is a remembered previous state rather than a live read:

```ts
let wasHidden = document.visibilityState === "hidden";
const onVisibility = () => {
  const hidden = document.visibilityState === "hidden";
  if (hidden) { stop(); }
  else if (wasHidden) { load(); start(); }
  wasHidden = hidden;
};
```

(That also makes the "started in a hidden tab" path fall out of the same variable rather than a separate initial check.)

The reason I'd act on this rather than file it as theoretical: **the test suite currently can't see it.**

```ts
setVisibility("visible");
setVisibility("visible");
load.mockClear();          // <-- the two redundant loads are discarded here
vi.advanceTimersByTime(2000);
expect(load).toHaveBeenCalledTimes(2);
```

`does not double-arm across repeated visible transitions` clears the mock *after* the two transitions, so it asserts the cadence afterwards — which is genuinely the double-arm property, and it does hold — but it discards the very calls that would show the duplication. Moving the `mockClear()` above the two `setVisibility("visible")` calls and asserting `0` there would pin both properties with the same test.

### 2. The extraction leaves the two callers it was extracted from behind

The module doc is explicit about why this exists:

> The shape here is not new: `ArtifactsTab` and the task detail screen already hand-rolled it. This is that pattern extracted so a fourth caller cannot get it subtly wrong

But neither of those is migrated, so the PR ends with three copies where there were two. `TaskDetailView.tsx:348-372` is the same shape line for line:

```ts
const onVisibility = () => {
  if (document.visibilityState === "hidden") { stop(); }
  else { void load(isActive); start(); }
};
if (document.visibilityState !== "hidden") start();
```

which is `startVisiblePolling(() => void load(isActive), POLL_MS)`. `ArtifactsTab.tsx:165-191` is the same plus an `editingRef` guard, and that guard composes straight into the callback:

```ts
startVisiblePolling(() => { if (!editingRef.current) void load(isActive); }, POLL_MS)
```

Both look like one-line migrations, and they'd inherit the fix in (1) rather than needing it applied three times. If you'd rather keep this PR scoped to the three new call sites that's a defensible call — but then the rationale in the doc comment is describing a cleanup that hasn't happened, and I'd soften it or leave a TODO naming the two files, so the next reader doesn't assume they're already covered.

(`WorkspaceView.tsx:603` is a different pattern — refresh-on-refocus with no interval, already guarded on `!== "visible"` — so I'd leave it out of scope.)

---

Neither of these blocks. (1) is worth doing in this PR because the fix and the test change are both small and the current test shape would hide a regression; (2) is a scope call that's yours to make.

oxoxDev:
Both findings addressed in `642c7eaa`, `dec059fa`, `2223492e`. Thanks — (1) was a real bug and the point about the test hiding it was the useful half.

**(1) Transition edge.** `visible-poll.ts` now remembers the previous state instead of reading it live:

```ts
let wasHidden = document.visibilityState === "hidden";
const onVisibility = () => {
  const hidden = document.visibilityState === "hidden";
  if (hidden) { stop(); }
  else if (wasHidden) { load(); start(); }
  wasHidden = hidden;
};
if (!wasHidden) start();
```

The started-in-a-hidden-tab check folded into `wasHidden` as you suggested — same observable behaviour, and its existing test passes untouched. `start()`'s double-arm guard stays as belt-and-braces, with its comment rewritten since the visible→visible rationale now lives on the edge check.

On the test: I moved `mockClear()` to immediately after `startVisiblePolling` rather than just above the two transitions — the mock also had the initial arm's call in scope, so clearing only above the transitions left a non-zero baseline. Assertion is what you asked for (`0` loads across redundant events, cadence assertion kept), renamed to *ignores redundant visible → visible events entirely*, plus a new *loads once per genuine hidden → visible edge* case since the old suite only covered a single edge.

I checked the new assertion actually has teeth rather than assuming it: temporarily restoring the old handler fails it with `expected "spy" to be called +0 times, but got 2 times` — the duplication you described — while the other six cases still pass.

**(2) Migrations.** Both done; both were behaviourally identical, so nothing had to be left behind.

- `TaskDetailView` — line-for-line, exactly as you read it. Mount-time `void load(isActive)`, the `cancelled` flag and cleanup preserved.
- `ArtifactsTab` — the `editingRef` guard composed into the callback. Worth noting the original applied that guard in *both* the tick and the visibility handler, so folding it into the single callback preserves it on both paths; there's a comment saying so.

`WorkspaceView` left alone — agreed it's a different pattern.

The module doc's rationale is now true rather than aspirational: it says both hand-rolled originals call the helper, instead of claiming a benefit that hadn't been delivered.

Gates green on the merged branch (`upstream/main` was merged in first — #657 touched `TasksView` and `ApprovalsView`): typecheck, typecheck:unit, `418 passed`, build, and `assert-design-tokens.sh`.

M3gA-Mind (commented):
**Re-review.** I commented on this at `0d9ddd95`; head is now `2223492e`. Reviewed manually again (the harness `sync` is openhuman-resident and I didn't want it checking out branches in a clone another agent is using). **Comment only, no approval.**

**Reviewing the delta, not the PR.** `41dadae3` is a merge of `upstream/main`, so a diff from my last review point shows ~81 files of other people's merged work. The author's actual new work is the three commits after it — `git diff 41dadae3 2223492e`, 4 files, +70/−70 — and that's what this covers.

**One thing a maintainer skimming this should know before reading the badges.** The two existing reviews are both stale, and not in the same way:

```
coderabbitai  APPROVED   16:15  commit 41dadae3   <- the merge commit
M3gA-Mind     COMMENTED  17:43  commit 0d9ddd95
head is now                     commit 2223492e
```

CodeRabbit's approval is pinned to `41dadae3`, which is **before all three of the commits that make up the author's response to review**. The APPROVED badge on this PR is a statement about a tree that no longer exists, and none of the behaviour change discussed below has been approved by anyone. Flagging it because the row reads green either way.

---

### The delta is clean. Both things I raised last time are fixed, and fixed properly.

**The edge, not the level.** My previous note was that `start()` was guarded against redundant `visible→visible` events but `load()` wasn't, so each stray event cost a full refresh — two requests for the callers that read two things. `642c7eaa` remembers the previous state:

```ts
let wasHidden = document.visibilityState === "hidden";

const onVisibility = () => {
  const hidden = document.visibilityState === "hidden";
  if (hidden) { stop(); }
  else if (wasHidden) { load(); start(); }
  wasHidden = hidden;
};

if (!wasHidden) start();
```

I walked the state machine rather than trusting the shape, since edge detection is where off-by-ones live:

- **mount visible** → `wasHidden = false` → `start()` arms the interval. The helper still deliberately doesn't `load()` on start, and both migrated callers keep their own `void load(isActive)` above the call, so nothing lost an initial fetch.
- **mount hidden** → arms nothing, listener attached; the first real transition to visible loads and arms. The separate `if (document.visibilityState !== "hidden") start()` line folded into `!wasHidden`, so there's now one source of truth for that decision instead of two that could drift.
- **unmount while still hidden** → disposer runs `stop()` (no-op, timer undefined) and removes the listener. No leak.
- **redundant `visible→visible`** → `hidden === false`, `wasHidden === false` → neither branch runs. Fixed.
- **redundant `hidden→hidden`** → `stop()` again, idempotent.

**The two hand-rolled callers are migrated,** which is the other thing I raised — the module doc justified the extraction by preventing divergence while leaving the two files it was extracted from on their own copies. `dec059fa` and `2223492e` handle both, and the doc was corrected to match (`hand-rolled it first` / `Both now call this instead`) rather than left overstating.

Both migrations are faithful, which is the part worth checking on a dedup:

- Cadence preserved — both still `POLL_MS = 4000`, untouched.
- Mount fetch preserved — `void load(isActive)` still runs above the helper call in both.
- `ArtifactsTab`'s `editingRef` guard composes into the callback, so it covers the tick *and* the return-to-foreground read, exactly as the hand-rolled version did in both its interval and its `onVisibility`. That's the one thing that could have silently dropped, and it didn't.

**The tests now discriminate.** My last note was that `does not double-arm across repeated visible transitions` called `load.mockClear()` *after* the two transitions, discarding the very calls that would expose the bug. That's inverted:

```ts
load.mockClear();
setVisibility("visible");
setVisibility("visible");
expect(load).toHaveBeenCalledTimes(0);
```

Against the old level-based handler that's `2`, so the test fails without the fix — it's no longer vacuous. And `loads once per genuine hidden → visible edge` covers the over-correction in the other direction (an edge check that never loads would pass the first test and fail this one). The pre-existing hidden test asserts on the mock, not on helper state, so it's testing the callback didn't fire rather than that the helper believes it's paused.

I found nothing wrong with the delta.

---

### Not a finding, but worth naming: three intervals still on the old pattern

Scope call, entirely yours — but the module doc now says the pattern "lives in one place and the next caller cannot get it subtly wrong", and there are three intervals that don't go through it. Naming them so the next reader doesn't infer the console is fully migrated:

- **`WorkflowsView.tsx:1076`** — `setInterval(() => setRunsTick(n => n + 1), 2_000)` with **no visibility guard at all**, driving a run-history refetch. This is the one I'd actually care about: 2s is the tightest cadence in the console, and it's armed exactly while a run is in flight — which is the long-lived state an operator is most likely to background the tab during. It's run history rather than the company/board, so it's outside #581 as written.
- **`TaskDetailView.tsx:1706`** — a `POLL_MS` interval doing a real network `read()`, guarded by `document.visibilityState !== "hidden"` *inside* the callback. It achieves the no-requests-while-hidden goal, so this is not a bug; it just does it the level-checking way, keeps the timer armed while hidden, and gets no catch-up read on return (stale for up to 4s). Notable mainly because it's the second poller in a file this PR migrated the first one out of.
- **`TaskDetailView.tsx:399`** — a 1Hz `setNow` clock, same inline guard. No requests, purely local; I'd leave it alone.

None of that blocks this PR and I wouldn't expand its scope for it. A follow-up issue would be the natural home, or just a line in the module doc noting the run-history poller is still ungated.

---

Re-reviewed the three commits since my last pass and found nothing wrong with them. Still comment-only, no approval — and worth repeating that the standing APPROVED here is CodeRabbit's, at a commit three commits behind head.

PR #672, opened 2026-08-11
approvals: give each turn its own approval scope instead of one pool-wide queue

### LINKED ISSUE #439: approvals: the request queue is shared across every turn, so a concurrent turn can still take another's entry

M3gA-Mind:
## Summary

One `ApprovalRequestQueue` was shared by every agent and every turn, so draining it had to **guess** which entries belonged to the draining turn. #395 narrowed that with a boundary index and documented the residual races at the call site. This removes the guess.

### ⚠️ Not the shape the issue names — read this first

The issue asks for **"one queue per run"**. That is **unbuildable**, and a reviewer comparing this diff to the issue will otherwise think I solved a different problem.

`ApprovalPolicy` is installed by `build_roster` (`harness/mod.rs:1920,1971`) inside `HarnessPool::ensure` — a **fingerprint-cached, per-company** function with no run id in scope and none possible, since it is idempotent per company. The policy is then sealed into the vendored agent (`build.rs:644` → `tool_policy: Arc<dyn ToolPolicy>`, no setter). So there is nowhere to hand a per-run queue *to*, and `approval_requests` is not part of any fingerprint, so passing a fresh queue would silently no-op against the cached roster. Rebuilding the roster per run would defeat the cache it exists for (MCP resolution, skills, tools, budgets) on every workflow run.

So the handle stays single and **the separation moves into the key**: entries are bucketed by `ApprovalScope`, a turn takes a `claim`, and `push` files into the surrounding claim's bucket. That delivers the property the issue actually asks for — *a turn sees only its own requests* — by construction rather than by boundary arithmetic.

### It also fixes a race the boundary never addressed

The issue names two races, both cycle∥run. There is a third: **run∥run**. Workflow runs are `tokio::spawn`ed and are *not* under the cycle's serial lock, so two genuinely overlap — and both took a boundary against the same vector, so the later `split_off` swallowed the earlier run's tail. A boundary index cannot fix that at all. Scopes are disjoint, so it cannot happen. Pinned by `two_concurrent_runs_cannot_take_each_others_entries`.

### The claim, and why `Drop` matters

`ApprovalClaim` copies the `PublishClaim` / `DelegationClaim` pattern already in this crate: it clears on entry exactly as the cycle's `clear()` did, **and on exit via `Drop`, which `clear()` never had**. A cycle that returned early — an error, a steer, an `?` — used to leave its entries for the *next* cycle to park. Now the claim's lifetime *is* the window.

This also makes the cancelled-run discard cleaner: it used to be performed by the next chat cycle's `clear`, which only worked because the leftovers were sitting in that cycle's way. Now `Drop` takes them as the dropped future unwinds, and no other turn has to sweep up after a run.

### Why the scope is ambient

`push` happens inside `require_approval`, called **synchronously by the vendored tool loop** through a per-agent policy that outlives every run. There is no parameter that could carry a run id. A `tokio::task_local` is sound here because the turn does not leave its task — `run_background` → `run_inner` is awaited directly (the one nearby `tokio::spawn` collects progress events, not the turn), and the body runs inside `with_stop_hooks`. **This is not a new dependency:** `with_stop_hooks` is itself a task-local scope on this exact path. I verified this before designing, because it is the assumption the whole approach rests on.

### `Unscoped` is deliberate, not laziness

A push outside any claim lands in `ApprovalScope::Unscoped`, which the **chat cycle** drains and a workflow run does not — i.e. exactly today's behaviour. It is not an error on purpose: a hard error would only be right if an unscoped push were impossible by construction, and I cannot prove that about turn entry points I have not enumerated. A missed entry point therefore degrades to "the operator is still asked" rather than to a silently discarded approval — **which is the precise failure #395 existed to fix**. Non-lossy by construction beats strict-and-brittle here.

### `take_from` collapsed

The boundary API meant nothing once buckets are disjoint, so the two call patterns become one: `claim` + `drain`. `drain`, `push`, `clear`, `queued` and `stamp_run` **keep their signatures**; only *which* requests they see changes. In the workflow path `stamp_run`'s `from` is now `0`, because the scope *is* the entitlement — there is no prefix belonging to anyone else to skip.

### `GrantSet` `Default` trap (fixed here, deliberately)

`ApprovalRequestQueue::default()` derived a **fresh `GrantSet`**. A grant redeemed through such a queue can never match, so every approval re-parks forever — the feature failing in its own happy path, with no error anywhere. Production is safe only because `RuntimeBuilder` uses `with_grants`.

Fixed in this PR rather than separately because it is **directly load-bearing**: the obvious way to build a per-run queue is one `default()` per run, which would have scoped grants per run and broken every approval. `Default` is now spelled out by hand with the hazard documented, and `grants_are_not_shared_by_default` pins the difference.

## API Or Behavior Changes

**No HTTP/GraphQL/wire changes.** Internal harness API only.

- `ApprovalScope` (new, public), `ApprovalClaim` (new, public), `ApprovalRequestQueue::claim` (new).
- `ApprovalRequestQueue::take_from` **removed**.
- `push` / `clear` / `drain` / `queued` / `stamp_run` — **signatures unchanged**, now scope-relative.
- `#[derive(Default)]` on `ApprovalRequestQueue` replaced with an explicit `impl`. Same behaviour, documented hazard.
- De-duplication is now **per scope**: two turns asking for the same tool are two asks. Collapsing them would hide one turn's request behind another's.

Behaviour that does **not** change: an unclaimed push still reaches the operator via the cycle; the per-turn cap still bounds a flooding model; the journal is still the durable record.

### ✅ Rebased onto current main; #625 reconciled here

#625 is **merged**, so `drain` now returns `DrainedRequests { requests, discarded, cap }`. Reconciled in this branch rather than left for a second lander:

- `drain` keeps #561's shape and gains #439's scoping. **The two compose better than either alone:** a `discarded` count taken off a shared vector mixed in whatever a concurrent run had appended, so "this turn overflowed" was never reliably *this turn's* fact. Scoped, it is.
- **`overflow_notice` reworded.** It said "from this turn" / "a single turn", true only while the cap bounded a chat cycle. A drain is now per-scope and a run spans more than a turn, so for `ApprovalScope::Run` it named the wrong unit of work. It now says "one batch" — true of a cycle and a run alike. Branching the sentence on the scope would have put the notice back in the business of being worded twice, which is what putting it on the type avoided.
- The workflow path drops the hand-rolled pre-count I had added and uses `drained.discarded`.

**The rebase surfaced a real break, which is the point of not trusting a clean merge.** My reword broke #625's `the_notice_quotes_the_cap_the_drain_was_taken_against`, which matches lowercase `"at most 3"`. I reworded my prose to keep it lowercase and mid-sentence rather than loosening their assertion — that assertion *is* #561's guarantee (the cap quoted is the one the drain was taken against, never a constant the call site had lying around), and weakening it to accommodate my sentence would have been the wrong trade.

## Tests

Gated lane throughout. Exit codes from unpiped redirects to a non-shared path.

- [x] `cargo fmt --all -- --check` — **EXIT 0**.
- [x] `cargo clippy --all-targets -- -D warnings` — ran as **`cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`** (CI's own flags; `--no-deps` is load-bearing per `ci.yml`). **EXIT 0**.
- [x] `cargo build --all-targets` — ran as `cargo build --features openhuman,tinycortex --all-targets`. **EXIT 0**.
- [x] `cargo test` — ran as `cargo test --features openhuman,tinycortex`. **EXIT 0, 2962 + 10 + 1 + 11 + 2 passed, 0 failed.**

All four re-run **from scratch on the rebased tree**; the pre-rebase greens were not carried over.

**Test-count delta accounted for, not assumed.** True base (`git checkout --detach upstream/main`) is **2958**; mine is **2962**. `+4` = **7 new** scope tests **− 3 deleted** `take_from` tests, the same delta as before the rebase.

> A note on how that base was measured, because I got it wrong once: `git stash` is **not** a baseline when your work is committed. It removed only my uncommitted files and left the rebase in place, so "base" was my own change measured against itself — and it reported a failure I could easily have misread as main being broken.

**All six revert-checks were re-run on the rebased tree. All six still fail on revert** — none went stale against #625's changes to the same type.

### Seven new tests, six revert-checked

| test | reverted | result |
|---|---|---|
| `a_run_drains_its_own_entries_and_cannot_see_another_turns` | `push` ignores the scope | **FAILS** |
| `two_concurrent_runs_cannot_take_each_others_entries` | `drain` takes every bucket | **FAILS** |
| `an_unclaimed_push_is_drained_by_the_cycle_never_by_a_run` | cycle stops draining `Unscoped` | **FAILS** |
| `dropping_a_claim_discards_that_scopes_entries` | remove `Drop` | **FAILS** |
| `duplicate_suppression_is_per_scope_not_global` | de-dup made global | **FAILS** |
| `grants_outlive_a_scope` | grants folded into the scope lifetime | **FAILS** |
| `grants_are_not_shared_by_default` | — | **characterization test, see below** |

**`grants_are_not_shared_by_default` is not revert-checked, and I am not claiming it is.** It pins pre-existing behaviour so a future refactor cannot silently change it; there is no fix beneath it to revert. Calling it out rather than implying six-of-six.

**One of these was vacuous first time round and is reported rather than quietly fixed.** `dropping_a_claim_discards_that_scopes_entries` originally asserted through a *fresh claim* — which passes even with `Drop` removed, because `claim()` clears on entry too. It proved nothing. It now reads the bucket directly via a `#[cfg(test)] len_in`, because claiming is not a neutral observation. It fails on revert, as shown.

**A regression in my own diff, caught the same way.** Switching the workflow path to `drain(cap)` made its "excess was discarded" warning dead code — `drain` truncates, so `requests.len() > cap` could never hold and the discard count was unrecoverable. Now counted *before* the drain.

### NOT verified

- **Not run in the live app**, and this one deserves emphasis: the races are concurrency bugs, and the tests exercise scope isolation deterministically rather than under real overlap. A genuine cycle∥run or run∥run soak was not run.
- **No test proves the task-local actually reaches `push` through the real vendored tool loop.** The scope plumbing is unit-tested directly; the end-to-end path rests on reading `run_inner` and confirming no `spawn` intervenes. This is the assumption most worth a reviewer's eye.
- No frontend involvement.
- `#[cfg(test)] len_in` is test-only surface by design.

## Documentation

No `docs/` change; the contract lives at the code and is updated there. Notably the doc block **the issue itself quotes** (`workflows/caps/mod.rs`, "Boundary, not drain") is rewritten to "Scope, not boundary" — leaving a comment that says "the real fix is a per-run queue, deliberately out of scope" would be actively misleading once this lands. Also updated: the queue type's own sharing contract, `require_approval` (why the scope is ambient), the cancelled-run note, and `gated_tool_turn_test`'s premise.

Fixes #439

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -32,6 +32,7 @@ use crate::harness::confine;
 // two note formats depending on which path touched it last.
 use crate::harness::lifecycle::{self, TaskRunEnd};
 use crate::harness::orchestrator;
+use crate::harness::policy::ApprovalScope;
 use crate::harness::publish::{self, WorkspaceSnapshot};
 use crate::runtime::advance::append_result;
 // `Delegation` is only named by the test-only `run_delegation` wrapper and the
@@ -2094,15 +2095,45 @@ fn settle(card: &mut TaskRecord, end: TaskRunEnd, responder: &str, body: &str) {
 #[async_trait]
 impl Brain for HarnessBrain {
     async fn run_cycle(&self, req: CycleRequest, host: &dyn CycleHost) -> Result<CycleResult> {
+        // Issue #439: everything this cycle does runs inside its own approval
+        // scope, so a workflow run executing concurrently cannot see, take, or
+        // be taken by it.
+        //
+        // The claim replaces the `clear()` that used to open this function. It
+        // clears on the way in exactly as that did, and additionally on the way
+        // out via `Drop` — which is the half `clear()` never had. A cycle that
+        // returned early used to leave its entries for the *next* cycle to
+        // park; now the window is the claim's lifetime and nothing outlives it.
+        let claim = self.deps.approval_requests.claim(ApprovalScope::Cycle);
+        claim.scoped(self.run_cycle_scoped(req, host)).await
+    }
+
+    /// The harness meters itself per turn in [`HarnessPool::run`], against the
+    /// live provider slug the turn resolved to — which is why `run_cycle` reports
+    /// zero `token_usage` and the runtime's cycle-level metering is a no-op here.
+    fn cognition(&self) -> Cognition {
+        Cognition {
+            path: crate::ports::brain::HARNESS_PATH,
+            provider: "per-turn",
+            metering: UsageMetering::PerTurn,
+        }
+    }
+}
+
+impl HarnessBrain {
+    /// The cycle body, running inside its [`ApprovalScope::Cycle`] claim.
+    ///
+    /// Split out only so the claim can wrap the whole of it: every turn this
+    /// cycle runs — the operator turn, its delegated desk turns, a dispatched
+    /// card, a re-dispatch after an approval — happens in here, and therefore
+    /// files its gated calls into this cycle's bucket.
+    async fn run_cycle_scoped(
+        &self,
+        req: CycleRequest,
+        host: &dyn CycleHost,
+    ) -> Result<CycleResult> {
         // Idempotent — builds the roster on the first cycle, a no-op after.
         self.pool.ensure(&self.record, &self.deps).await?;
-
-        // Issue #172: start from an empty approval queue so nothing a prior
-        // cycle — or a workflow run sharing these deps — left behind is parked
-        // under this cycle. Every turn this cycle runs (the operator turn, its
-        // delegated desk turns, a dispatched card) pushes onto the same queue and
-        // is drained once at the end.
-        self.deps.approval_requests.clear();
 
         let mut channel_responses = Vec::new();
         for event in &req.events {
@@ -2364,17 +2395,6 @@ impl Brain for HarnessBrain {
             ledger_deltas: Vec::new(),
             token_usage: TokenUsage::default(),
         })
-    }
-
-    /// The harness meters itself per turn in [`HarnessPool::run`], against the
-    /// live provider slug the turn resolved to — which is why `run_cycle` reports
-    /// zero `token_usage` and the runtime's cycle-level metering is a no-op here.
-    fn cognition(&self) -> Cognition {
-        Cognition {
-            path: crate::ports::brain::HARNESS_PATH,
-            provider: "per-turn",
-            metering: UsageMetering::PerTurn,
-        }
     }
 }
 
--- a/src/harness/policy.rs
+++ b/src/harness/policy.rs
@@ -69,6 +69,7 @@
 //! where it does, why an at-cap call **parks** rather than being denied, and
 //! what the cap does not see.
 
+use std::collections::BTreeMap;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
@@ -172,6 +173,39 @@ pub struct ApprovalRequest {
     pub effect: Effect,
 }
 
+/// Which turn a queued approval request belongs to (issue #439).
+///
+/// The queue handle is one per company and cannot be otherwise — see
+/// [`ApprovalRequestQueue`] — so the separation between turns lives here, in
+/// the key, rather than in separate queues.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum ApprovalScope {
+    /// A request pushed outside any [`claim`](ApprovalRequestQueue::claim).
+    ///
+    /// The default, and deliberately **not** an error. Every production turn
+    /// runs under a claim, so this bucket should stay empty there; making an
+    /// unclaimed push panic or drop would trade a mis-attribution bug for a
+    /// lost-approval bug, and #395 exists because a lost approval is the worse
+    /// one. It is drained by the chat cycle alongside [`Cycle`](Self::Unscoped)
+    /// — which is exactly the pre-#439 behaviour for anything unclaimed.
+    #[default]
+    Unscoped,
+    /// The company's chat cycle, including every dispatched card and delegated
+    /// turn that runs inside it.
+    ///
+    /// One at a time per company: `CycleRunner` holds a serial lock, so this
+    /// bucket has a single writer. It is still concurrent with any number of
+    /// workflow runs, which is the race #439 is about.
+    Cycle,
+    /// One workflow run, keyed by its run id.
+    ///
+    /// Workflow runs are `tokio::spawn`ed and are **not** under the cycle lock,
+    /// so two of them genuinely overlap. That is the race a boundary index
+    /// could never fix: both runs take a boundary against one shared vector and
+    /// the later `split_off` swallows the earlier run's tail.
+    Run(String),
+}
+
 /// A shared, in-memory queue of approval-gated tool calls — the exact
 /// [`DelegationQueue`](crate::harness::orchestrator::DelegationQueue) /
 /// [`McpFailureQueue`](crate::harness::mcp_probe::McpFailureQueue) pattern.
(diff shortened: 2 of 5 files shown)


oxoxDev (changes requested):
Two things in here are better than the change they support.

The first is the opening refusal. The issue asks for "one queue per run", and rather than approximating it you established that it is unbuildable — `ApprovalPolicy` is installed by `build_roster` inside a fingerprint-cached per-company `ensure` with no run id in scope, then sealed into the vendored agent with no setter, and `approval_requests` is not part of any fingerprint, so a fresh per-run queue would silently no-op against the cached roster. Moving the separation into the *key* instead delivers the property the issue actually wants by construction. Leading with that, rather than letting a reviewer discover the mismatch in the diff, is the right call.

The second is the `GrantSet` `Default` trap. A derived `Default` minting a fresh `GrantSet` means a grant redeemed through that queue can never match, so every approval re-parks forever — the feature failing in its own happy path with no error anywhere, invisible in production only because `RuntimeBuilder` happens to use `with_grants`. Finding that while building something else, and fixing it *here* because the obvious per-run-queue implementation would have walked straight into it, is exactly right.

The run∥run race is a real catch too — spawned workflow runs are not under the cycle's serial lock, both take a boundary against the same vector, and the later `split_off` swallows the earlier run's tail. A boundary index cannot fix that, and disjoint scopes cannot hit it. `Unscoped` degrading to "the operator is still asked" rather than erroring is the right default for the same reason.

**2 major.**

### Major — does not merge, and no lane has run

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Your base is `97216647`, **215 commits back**. The only check present is CodeRabbit reporting rate-limited — no Rust lane has executed, so the `EXIT 0` results in the description are local against a base that has moved a long way.

**#625 merged at 11:20 today.** The "merge #625 FIRST" instruction is satisfied, which means the reconciliation you pre-planned comes due *in this PR* rather than after it: the one line inside `drain`, and the `overflow_notice` wording you flagged yourself — "this turn" stays right for `Cycle` and is wrong for `Run`. You wrote that note as a message to whoever landed second; you are whoever landed second.

### Major — collides with your own #660, which neither PR mentions

#660 (`approvals: decide which calls need a human, per call`) edits `src/harness/policy.rs` too. The overlap is mostly additive — it adds a section to the module docs where you add `use std::collections::BTreeMap`, and both append to the same `mod tests` — so the textual half is small.

The semantic half is the one that matters. #660 adds a **new `RequireApproval` arm to `check`**, and your new doc on `require_approval` states that every such arm goes through there and now lands in the surrounding turn's `ApprovalScope`. Those compose correctly as far as I can trace — #660's arm only fires inside a turn, and it notes the workflow `tool_call` path never consults `ApprovalPolicy` at all — but "as far as I can trace" is not the standard you held yourself to on #625, where you named the conflict, its shape, and the reconciliation. This one deserves the same paragraph on both PRs.

Worth stepping back: **#659, #660 and #672 are all open, all yours, all cut from bases 215–225 commits back**, and #659 is already failing all four Rust lanes on a struct-widening break that only appears in the compiled merge. That is one rebase pass, not three separate problems — and doing it in one go lets you settle the #660 ordering question at the same time.

Requesting changes on the merge state alone. The design argument is sound and I would expect to approve it once it builds against current `main`.

oxoxDev (approved):
Re-reviewed at `b913326b`. Both majors are closed.

**Merge state and CI** — `MERGEABLE` / `CLEAN`, nine checks green including `Rust (openhuman, tinycortex)` and both E2E. The rebase brought `DrainedRequests` in and the reconciliation is exactly the one line inside `drain` you predicted it would be, rather than a call-site sweep — shaping the change to leave `drain`'s name, arity and return type alone paid for itself.

**Your own wording item is resolved, and better than the fix you proposed.** You flagged that `overflow_notice` saying "this turn" would be right for `Cycle` and wrong for `Run` once the count was claim-scoped. It now reads "One batch can raise at most {cap}", which is neutral between the two rather than branching on which scope drained — one sentence instead of two that could drift.

Keeping `cap` private on `DrainedRequests`, so a caller cannot render `overflow_notice(20)` against a `drain(8)`, is the same instinct as the type itself applied one level up. Worth the extra field.

**0 major. 1 nit. Approving.**

### Nit — `overflow_notice` disagrees with itself at `n == 1`

`calls` and `them` are pluralised; the verb is not:

> Heads up: 1 further gated tool **call were** not raised for approval.

`cap` is 8 by default so the single-discard case is reachable — it is exactly the ninth gated call in a batch. Same treatment as the other two:

```rust
let were = if n == 1 { "was" } else { "were" };
```

Cosmetic, and it is operator-facing copy on a path an operator only ever reads when something already went wrong, which is the worst moment to look unpolished.

### Merge order with #660 — still live, and now decided by state

#660 is still open and still edits `src/harness/policy.rs`, so the overlap I raised has not gone away. It has become easy to sequence, though: this branch is rebased and green, #660 is still on the old base and conflicting, so **this one should land first** and #660 picks it up in the rebase it already owes. Its new `RequireApproval` arm in `check` then lands on top of the ambient-scope `require_approval` rather than under it, which is the order that needs no thought from either side.

Issue #661, opened 2026-08-11
workflows: authoring & validation gaps let broken graphs (incl. shipped seeds) save clean and fail silently


oxoxDev:
## Summary

A full sweep of the workflows module found a broad class of defects with **one shared root cause**: config **validation** and config **derivation** only cover `agent` and `tool_call` nodes. Every other node kind's correctness rests entirely on hand-written `config` that nothing enforces — so agent-authored graphs, human-authored graphs, **and the shipped seed workflows** save clean and then misbehave silently at runtime.

This started from a concrete case: the `create_workflow` tool an in-company agent (e.g. a PM persona) uses over chat produced a GitHub workflow whose `tool_call` nodes had no `config.slug` — dead by construction. Investigating that opened up the wider module. This issue folds the `create_workflow` gap together with the full module audit.

**Not a bug (verified, to pre-empt):** per-node output persist (#596) is wired on every live path — the "empty output on staging" was deploy-staleness, not a runner gap. `dry_run` is side-effect-free. `#241` at-most-once, `#276` disarm, and the `#401` cap *release* are all sound. Delivery fails loud (never silent-drops) except finding H8 below. REST and the agent tool share one persist/validate core (`create_company_workflow`) — no drift between them.

**Already tracked, referenced not duplicated:** `#617` (policy approval-gate is silently bypassed inside `sub_workflow` children — confirmed, High, security). `#638` (approval-park overflow discards beyond the per-turn cap).

---

### Root cause
- `validate_draft_against_record` validates only `agent` + `tool_call` (`src/company/workflow_create.rs:634-652`, `_ => {}`).
- `translate_node` derives config only for `agent` (`prompt`/`agent_ref`) and `tool_call` (`slug`) (`src/workflows/translate.rs:133-185`).
- The on-disk seed path (`parse_workflow` → `validate`, `src/company/workflow_file.rs:583-1033`) enforces even less than the console draft path — it never requires a `tool_call` slug or an `agent` roster ref.
- Neither OpenCompany validation nor the tinyflows `validate` (run inside `compile`) enforces the *functional* config that `condition`/`switch`/`http_request`/`split_out`/`transform` need to do anything.

The result is a "saves-clean / fails-silently-at-runtime" class across the whole module.

---

### HIGH

**H1 — `create_workflow` agent tool advertises 6 node kinds but can only wire 2.**
[...]

Issue #401, opened 2026-08-05
workflows: nothing bounds how many runs of a company can be in flight at once


oxoxDev:
## Context

Split out of #383 review, which flagged it correctly as out of scope for that PR.

Before #383, `POST …/workflows/{wid}/run` was self-limiting by accident: the
response was held open for the whole run, so a caller paid for each in-flight
run with a socket, and a proxy's connection limit was the de-facto cap. That was
never a designed bound — it was a side effect of the blocking call, and the same
side effect that made a run die when the connection dropped.

#383 removed the blocking (correctly — a run dying with its client was a bug),
and `detach: true` now answers in milliseconds. A caller can start runs as fast
as the route can register them.

## What is unbounded

`spawn_workflow_run` spawns unconditionally. The manual path has no overlap
guard at all — `workflow_scheduler` says so explicitly ("Manual runs go through
the run route and are unaffected"), and the cron path's guard is per-workflow
one-at-a-time, not a company-wide cap.

The console keeps a per-workflow guard, but that is a client-side courtesy, not
a limit: it is one browser tab's state and does not survive a second tab, a
script, or a direct `curl`.

The ceiling that matters is not a task count. Each run can drive agent turns and
inference calls, so N concurrent runs is N times a company's inference spend and
N times its provider rate limit. A loop starting runs is a billing incident
before it is a memory one.

## Why it was deliberately not fixed in #383

Capping is **policy**, and policy needs numbers somebody chose. The questions
have real trade-offs and none of them are answerable from inside the cancellation
work:

- per-company, per-workflow, or per-tenant-plan?
- reject with `429`, or queue and admit later?
- does a cron fire count against the same budget as an operator's click?
- does the existing `[tools]`/budget machinery own this, or is it a new knob?

Shipping an arbitrary number inside a bug-fix PR would have been worse than
naming the gap.

## Scope

- A bound on concurrent runs per company, with the shape decided above.
- A clear rejection on the run route when the bound is hit — not a silent queue,
  since the caller now gets an id back and would otherwise be told a run started
  that has not.
- The run supervisor already holds the live set (`RunSupervisor::live()`), so the
  count is available without new bookkeeping.

## Related

- #383 — removed the accidental backpressure; disclosed this as deliberate.

oxoxDev:
**Related — `cluster:workflows`**

Workflow engine, run lifecycle, scheduling, and canvas.

*Gated on the vendored engine surface — read #499 first:*
[...]

PR #676, opened 2026-08-11
fix(scheduler): admit against in-flight cap before claiming the minute (#661)

oxoxDev:
## Summary

Fixes the scheduled-minute leak introduced by the #401 in-flight run cap (part of #661, finding H4).

The scheduler commits `claim_fire(minute)` (`workflow_scheduler.rs`) synchronously in `tick()`, but the #401 cap is enforced much later inside `RunSupervisor::begin` (`run_supervisor.rs:146`), reached via a detached `tokio::spawn` in `spawn_scheduled_run`. So a company already at its in-flight cap (default 8) would **durably claim the minute, then have the run rejected** — and catch-up treats a claimed minute as already-fired, so the scheduled occurrence is **permanently lost and operator-invisible** (no `WorkflowRunFinished` is journaled).

Fix: `RunSupervisor` is per-company and exposes `len()`/`limit()`, so `tick()` now does an **advisory cap pre-check before `claim_fire`**. At cap → log and `continue` with **nothing claimed**, so the schedule retries naturally on the next tick. Applied at both leak sites (steady-state fire and restart catch-up make-up); for catch-up, the first-sight latch is dropped at cap so the make-up is deferred-and-retried, not abandoned. The authoritative `begin` cap check stays as the safety net for the narrow race where a slot fills between the advisory read and `begin` (no worse than today).

Not a pure reorder or claim/release rollback: `claim_fire` must still precede an *admitted* run to preserve cross-replica exactly-once, and no new store "unclaim" method is needed. Detail in the commit.

## API Or Behavior Changes

No public API change. Behavior: a schedule that fires while its company is at the in-flight cap is now **retried on a later tick** instead of being silently dropped. Success path is unchanged — an admitted minute is still claimed exactly once (no double-fire). Adds structured `tracing::info!` on both cap-skip paths (company / workflow / schedule / limit).

## Tests

- [x] `cargo test` (default + `--features openhuman,tinycortex`) — 2960 passed / 0 failed; scheduler module 38/0

New regression tests:
- `a_fire_at_the_in_flight_cap_leaves_the_minute_unclaimed_to_retry` — at cap the minute is not claimed; fires + claims exactly once when the slot frees.
- `a_catch_up_at_the_in_flight_cap_is_deferred_not_burned` — the missed minute's anchor is untouched while deferred, then lands once capacity frees.

## Documentation

No doc change needed — internal scheduler admission ordering; behavior is covered by the new tests and this description.

## Related

Part of #661 (workflows module audit). Fixes the H4 interaction with #401.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced exact per-company limits for concurrently running scheduled workflows.
  * Deferred runs that reach the limit instead of skipping them, allowing catch-up after capacity becomes available.
  * Improved recovery for coarse schedules and runs resumed after a restart.
  * Ensured sibling schedules are admitted consistently during the same scheduling cycle.

### DIFF
--- a/src/runtime/workflow_scheduler.rs
+++ b/src/runtime/workflow_scheduler.rs
@@ -83,10 +83,11 @@ use tokio::task::JoinHandle;
 
 use crate::company::{WorkflowFile, list_workflows_union};
 use crate::ports::types::CompanyId;
-use crate::ports::{DeliveryReport, DeliveryStatus};
+use crate::ports::{DeliveryReport, DeliveryStatus, WorkflowRunContext};
 use crate::runtime::CompanyRegistry;
 use crate::runtime::WorkflowSpawn;
 use crate::runtime::cron::{CivilTime, CronExpr};
+use crate::runtime::run_supervisor::RunGuard;
 use crate::runtime::scheduler::{
     CATCHUP_WINDOW_MINUTES, Clock, MINUTE_MS, PRUNE_CUTOFF_MINUTES, millis_to_next_minute,
     missed_instant,
@@ -361,20 +362,22 @@ impl WorkflowScheduler {
             // reached through the runtime resolved this tick.
             let store = runtime.schedule_fires().clone();
 
-            // Issue #661: the per-company in-flight run cap (issue #401). Read
-            // here so a fire can be admitted against it BEFORE it durably claims
-            // its minute. The authoritative admission is `RunSupervisor::begin`
-            // inside `spawn.spawn`'s task (see `spawn_scheduled_run`), which the
-            // scheduler reaches only AFTER `claim_fire` has already committed —
-            // so a company at its cap would claim the minute, have the run
-            // rejected there, and (because catch-up reads a claimed minute as
-            // fired) lose the occurrence for good. This cheap pre-check against
-            // the same predicate `begin` enforces (`len() >= limit`) lets a
-            // capped fire skip WITHOUT claiming, so the schedule retries on the
-            // next tick once a slot frees. The check is advisory — `begin` stays
-            // the authority under its lock — so the pre-existing skip-at-cap log
-            // in `spawn_scheduled_run` still covers the narrow race where a slot
-            // fills between this read and that `begin`.
+            // Issue #661: the per-company in-flight run cap (issue #401). The
+            // scheduler admits a fire against it — `RunSupervisor::begin` —
+            // synchronously on THIS tick thread, BEFORE the durable `claim_fire`
+            // (see the catch-up and steady-state arms below), and holds the guard
+            // across the claim. Admitting before claiming is what keeps a company
+            // at its cap from durably claiming — and burning — a minute whose run
+            // `begin` would then reject: at the cap `begin` refuses before any
+            // claim, so nothing is marked and the guard it would have handed back
+            // is simply never taken. There is no advisory `len() >= limit`
+            // pre-check any more: `begin` enforces the cap under the same lock it
+            // inserts under, so it is the authority and there is nothing to race
+            // it against. Doing it on the tick thread — not inside a spawned task
+            // — also makes the count EXACT for same-tick sibling schedules: an
+            // admitted run's guard is registered before the NEXT schedule's
+            // `begin` reads the count, so `cap` of N schedules all due at one
+            // minute admit exactly `cap`, never all N.
             let supervisor = runtime.run_supervisor().clone();
 
             for (file, cron, expr) in scheduled {
@@ -398,26 +401,34 @@ impl WorkflowScheduler {
                                 // a same-minute steady-state fire below suppresses
                                 // WITHOUT claiming (the minute is suppressed, not
                                 // burned) rather than running a second copy.
-                                //
-                                // Issue #661: the make-up run also routes through
-                                // `spawn_scheduled_run` → `begin`, so at the #401
-                                // in-flight cap it would durably claim `missed` and
-                                // then have the run rejected — burning the make-up
-                                // exactly like the steady-state leak above. Defer
-                                // instead: drop the first-sight latch so the next
-                                // tick re-attempts catch-up once a slot frees,
-                                // leaving `missed` unclaimed.
-                                if supervisor.len() >= supervisor.limit() {
-                                    self.caught_up.remove(&key);
-                                    tracing::info!(
-                                        %company,
-                                        workflow = %file.id,
-                                        schedule = %cron,
-                                        limit = supervisor.limit(),
-                                        missed_minute = missed,
-                                        "workflow scheduler: company at its in-flight run cap, deferring restart catch-up to a later tick"
-                                    );
-                                } else if let Some(claim) = self.claim(&key) {
+                                if let Some(claim) = self.claim(&key) {
+                                    // Issue #661: admit (`RunSupervisor::begin`)
+                                    // BEFORE the durable `claim_fire`, holding the
+                                    // guard across it. At the #401 in-flight cap
+                                    // `begin` refuses here, so `missed` is never
+                                    // claimed and the make-up is DEFERRED, not
+                                    // burned: drop the first-sight latch so a later
+                                    // tick re-attempts catch-up — `missed` is still
+                                    // inside the catch-up window — once a slot
+                                    // frees, and `continue` so the steady-state arm
+                                    // does not also log an at-cap line for the same
+                                    // key on this tick.
+                                    let (ctx, guard) = match supervisor.begin(&file.id, true) {
+                                        Ok(admitted) => admitted,
+                                        Err(_) => {
+                                            drop(claim);
+                                            self.caught_up.remove(&key);
+                                            tracing::info!(
+                                                %company,
+                                                workflow = %file.id,
+                                                schedule = %cron,
+                                                limit = supervisor.limit(),
+                                                missed_minute = missed,
+                                                "workflow scheduler: company at its in-flight run cap; deferring restart catch-up, leaving the missed minute unclaimed for a later tick to make up while inside the catch-up window"
+                                            );
+                                            continue;
+                                        }
+                                    };
                                     match store.claim_fire(&company, &schedule_id, missed).await {
                                         Ok(true) => {
                                             let input = json!({
@@ -438,6 +449,8 @@ impl WorkflowScheduler {
                                             );
                                             spawn_scheduled_run(
                                                 &spawn,
+                                                ctx,
+                                                guard,
                                                 claim,
                                                 company.clone(),
                                                 file.clone(),
@@ -446,9 +459,14 @@ impl WorkflowScheduler {
                                             fired += 1;
                                         }
                                         // A simultaneously-booting replica claimed
-                                        // the catch-up first: release the slot.
-                                        Ok(false) => drop(claim),
+                                        // the catch-up first: release the slot and
+                                        // the admission.
+                                        Ok(false) => {
+                                            drop(guard);
+                                            drop(claim);
+                                        }
                                         Err(err) => {
+                                            drop(guard);
                                             drop(claim);
                                             tracing::warn!(%company, workflow = %file.id, %err, "workflow scheduler: could not claim catch-up fire; skipping (fail closed)");
                                         }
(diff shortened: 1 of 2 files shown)


M3gA-Mind (commented):
Reviewed manually (`gh pr diff` + reading `workflow_scheduler.rs`, `workflow_spawn.rs`, `run_supervisor.rs` at the PR head against `main`) rather than through our review harness — the harness's `sync` step is openhuman-resident and I did not want it checking out branches in a clone another agent is using. **Comment only, no approval.**

The direction is right: claiming a durable minute for a run that `begin` is going to refuse is a real leak, and leaving the minute unclaimed is strictly better than burning it. Three things below, the first of which I think materially overstates what the fix delivers.

---

### 1. "retries on the next tick" only holds for `* * * * *`

Both the code comment and the operator-facing log say the deferred occurrence comes back on the next tick:

> `"workflow scheduler: company at its in-flight run cap, leaving this minute unclaimed to retry on the next tick"`

But the skip is a `continue` past a plain `expr.matches(&civil)` gate, and nothing remembers that this minute was owed. On the next tick the minute has moved on, so for any schedule coarser than every-minute the expression simply doesn't match and there is no retry. A company at its cap at 09:00 with a `0 9 * * *` digest does not get that digest at 09:01 when a slot frees — it gets it tomorrow.

The steady-state test is the one cron shape where the claim is true by construction:

```rust
vec![overlay("digest", Some("* * * * *"))],
...
// Free the slot; the SAME minute now fires
drop(filler);
assert_eq!(scheduler.tick().await, 1, "the freed slot lets the deferred fire land");
```

With `* * * * *` the next tick matches regardless of any deferral logic, so the assertion passes without anything actually having been deferred. Swap that overlay for `0 9 * * *` and hold the clock at 09:01 and I don't believe it fires.

That doesn't make the change wrong — an unclaimed minute still leaves the anchor honest, so a later restart's catch-up can make it up within `CATCHUP_WINDOW_MINUTES`, which the old burned-minute behaviour foreclosed. That's a genuine improvement and worth saying. But it's a different claim from "retries on the next tick", and an operator reading that log line will wait for a run that isn't coming. Either narrow the wording on both the comment and the log to what actually happens (the minute is left unclaimed and unfired; recovery is via restart catch-up, not the next tick), or make the deferral real by parking the owed minute per key and re-attempting it while it's inside the catch-up window — which is close to what the catch-up arm already does.

### 2. The race with `begin` isn't narrow for sibling schedules in the same tick

The comment characterises the residual race as "the narrow race where a slot fills between this read and that `begin`". For two fires in *different* ticks, agreed. For two schedules on the *same company* due in the *same* tick, it isn't narrow — it's the default outcome.

`spawn_scheduled_run` does its work inside a `tokio::spawn`, and `begin` is called in there:

```rust
tokio::spawn(async move {
    let _claim = claim;
    ...
    let (_run_id, handle) = match spawn.spawn(workflow, input, true, false) {
```

so admission does not happen on the tick's own thread. Walking the loop for workflow B after workflow A fired: catch-up is already latched (no await), `expr.matches` (no await), then the new cap pre-check. There is **no await point between A's `spawn_scheduled_run` and B's `supervisor.len()` read**, so B sees a count that does not yet include A. On a current-thread runtime that's guaranteed; on the multi-threaded one it's a coin flip.

Concretely: a company with four `0 9 * * *` workflows and `max_in_flight_runs = 1`. All four pass the pre-check against a stale `len()`, all four reach `claim_fire`, three are refused at `begin` — and those three minutes are durably burned, which is precisely the failure this PR exists to remove. Neither test covers more than one workflow, so nothing catches it.

`WorkflowSpawn::spawn` is already synchronous up to and including `begin` (it returns `Result<(String, JoinHandle<…>)>`, refusing "before the `tokio::spawn` below"). So `spawn_scheduled_run` could call `spawn.spawn(...)` on the tick's thread and `tokio::spawn` only the `handle.await` + delivery sweep that follows. That makes the count exact and ordered with the loop, and shrinks the pre-check's race to nothing.

Better still, it would let the ordering be admit → claim → run rather than pre-check → claim → admit: hold the `RunGuard` across `claim_fire` and drop it if the claim is lost. Then the cap decision and the minute claim can't disagree at all, and the advisory read stops being load-bearing.

### 3. The pre-check sits above the `last_fired` guard

```rust
if supervisor.len() >= supervisor.limit() { … continue; }

if self.last_fired.get(&key) == Some(&minute) {
    continue; // already fired this minute (in-process first pass)
}
```

When a second tick lands inside a minute that already fired — which is the case `last_fired` exists for — and the run it started is holding the only slot, the cap branch now wins and logs "leaving this minute unclaimed to retry on the next tick" about a minute that was claimed and did fire. That's the log an operator reads while diagnosing exactly this area, and it's saying the opposite of what happened. Moving the check below the `last_fired` guard (and arguably below the overlap `self.claim`, since a suppressed-by-overlap fire is likewise not being skipped for cap reasons) keeps the line true.

Related, minor: when the catch-up arm defers, control falls through to the steady-state path, which can hit the cap check and log a second at-cap line for the same key in the same tick.

---

Smaller notes:

- `company_with_overlays_capped` and the `cap: Option<usize>` passthrough are a clean way to drive this; no complaint.
- Both new tests do fail with the fix reverted (the steady-state one on `tick() == 0` and on `latest_fire == None`), so they're not vacuous — my objection in (1) is about what the passing assertion proves, not about whether it proves anything.
- The rewritten `spawn_scheduled_run` comment now describes the arm as a safety net for a narrow race. If (2) lands, that becomes accurate; as written I'd soften it.

Nothing here needs a new issue if (1) is resolved by fixing the wording and (2) by hoisting `begin` — but (1) and (2) both change what the PR delivers, so I'd want them addressed rather than deferred.

oxoxDev:
Both (1) and (2) addressed — and went past honest-wording to make the deferral real.

- **(2) same-tick sibling race:** split `WorkflowSpawn::spawn` into `begin` + a new infallible `spawn_admitted`. The tick loop now does **admit → claim → run**: `begin()` runs **synchronously on the tick thread** before `claim_fire`, the `RunGuard` is held across the claim, and only the await/delivery sweep is `tokio::spawn`ed. So `begin` registers the guard before the loop walks to the next sibling — the in-flight count is exact for same-tick siblings, and the advisory `len() >= limit()` pre-check is **dropped** (begin is the authoritative admission under its own lock). Regression test `siblings_due_the_same_minute_admit_exactly_the_cap_*` (current-thread **and** multi-thread): 4 workflows due one minute at cap 1/2 → exactly `cap` fire+claim, the rest refused with their minutes **unclaimed**, no burn.
- **(1) "retries next tick":** corrected. A steady-state defer at cap drops the first-sight catch-up latch, so the missed minute is made up via the catch-up path on a later tick **within the catch-up window, without a restart**. Comment + operator log now say the truth (minute left unclaimed/unfired; recovered by catch-up, not the next expression match). Test `a_coarse_steady_state_fire_at_cap_is_made_up_by_catch_up_not_the_next_match` (`0 9 * * *` held at 09:00, made up at 09:01 once a slot frees).
- **(3) ordering:** the cap decision now sits **below** the `last_fired` and overlap `self.claim` guards, so an already-fired/overlap-suppressed minute is never mislabeled at-cap; the catch-up defer `continue`s so there's no second at-cap line per key per tick.

One residual (inherent, pre-existing): a schedule's very-first fire at cap with no prior anchor has nothing to make up, so it waits for the next matching minute. Flagging in case you want it tracked.

Scheduler module 41/41 (default + gated). SHAs: `38610071` (hoist admission), `6b58aa13` (admit-before-claim + ordering).

M3gA-Mind (commented):
**Re-review.** I commented on this at `a2823f12`; head is now `6b58aa13`. Reviewed the two commits since (`38610071` hoisting admission out of the spawned task, `6b58aa13` reordering it against the fired/overlap guards) — `git diff a2823f12 6b58aa13`, 2 files, +388/−122. Manual again, same reason as before. **Comment only, no approval.**

**The delta resolves all three things I raised, and two of them at the root rather than by patching the symptom.** Going through them in order because I want to be specific about what I re-checked.

**The advisory pre-check is gone entirely.** My second point was that `supervisor.len() >= supervisor.limit()` read a stale count for sibling schedules in the same tick, because `begin` ran inside `spawn_scheduled_run`'s `tokio::spawn` and there was no await between one sibling's spawn and the next sibling's check. Rather than tightening the pre-check, `38610071` splits `WorkflowSpawn::spawn` into `begin` + `spawn_admitted` and has the scheduler call `begin` itself on the tick thread. That's the right fix: there is no longer an advisory read to be stale, because the authoritative admission *is* the check, and the guard it returns counts against the cap immediately.

On the concurrency question that the hoist raises — whether admission is still atomic with respect to the cap, or merely earlier. It's still atomic, and for a reason the hoist doesn't touch: `RunSupervisor::begin` does its `map.len() >= self.limit` test and its `map.insert` under **one** hold of the map lock. Moving *when* `begin` is called changes nothing about that. Two siblings in one tick cannot both admit against one remaining slot, and neither can two turns racing from anywhere else. What the hoist changes is only that the count the next sibling sees is post-insert rather than pre-spawn.

(For completeness on "can two schedulers admit against the same remaining capacity": across *replicas*, yes — each process builds its own `RunSupervisor`, so the #401 cap has always been per-process rather than global, and `claim_fire` is what prevents the cross-replica double-fire. That's pre-existing and untouched here; I mention it only because the hoist makes the per-process scope more visible than it was.)

**The ordering claim holds literally.** `6b58aa13` puts `begin` below both the `last_fired` guard and the overlap `self.claim`, and above `claim_fire`. I checked each of the three constraints separately rather than trusting the commit message: a minute already fired in-process returns at `last_fired` before the cap is consulted; a minute suppressed by a still-running previous fire returns at the overlap guard; and only a fire that is genuinely going to happen reaches `begin`. So the at-cap log line can no longer describe a minute that was actually claimed and fired, which was my third point.

**And the overclaim I cared about most is fixed in behaviour, not just wording.** My first point was that "the schedule's next tick retries it naturally" is only true for `* * * * *` — for `0 9 * * *`, 09:01 doesn't match and the occurrence is gone until tomorrow. The steady-state at-cap path now does `self.caught_up.remove(&key)`, re-arming the catch-up so a later tick makes up the missed minute while it's still inside `CATCHUP_WINDOW_MINUTES`. That's a real recovery mechanism where before there was a claim that recovery happened. The log line was corrected to match, and says the non-obvious part out loud — *"not the next expression match, for any schedule coarser than every-minute"*.

**The tests discriminate.** This is where I spent most of the re-review, since my original note was that the passing assertion didn't prove what it appeared to.

`a_coarse_steady_state_fire_at_cap_is_made_up_by_catch_up_not_the_next_match` is exactly the case the old suite couldn't reach: `0 9 * * *`, deferred at cap at 09:00, then recovered by ticking at **09:01 — a non-matching minute**. If the only recovery were the expression matching again, that assertion reads `0` and the test fails. It pins the mechanism rather than the outcome.

`assert_siblings_admit_exactly_the_cap_and_burn_no_minute` is the one I'd have asked for: four workflows on one company at `0 9 * * *`, run at cap 1 and cap 2, with a semaphore-gated runner so the admitted runs actually hold their slots through the tick. It asserts both halves — exactly `cap` fire, **and** exactly `cap` minutes are durably claimed, so the refused siblings burned none. Against the old advisory pre-check that second assertion reads `4`, so it fails. And running it under both `#[tokio::test]` and `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` is the right call: current-thread is where a stale off-thread read was guaranteed, multi-thread is where it was a coin flip, and a fix for this has to hold on both.

Guard lifetime, which is what the hoist newly makes possible to get wrong, is handled on every exit: `Ok(false)` and `Err` from `claim_fire` both `drop(guard)` alongside `drop(claim)`, on the steady-state arm and the catch-up arm alike. Nothing leaks a slot.

---

One consequence worth naming, not a defect: the guard is now held across `claim_fire().await`, so a cap slot is briefly occupied by a run that may never start — the case where a peer replica wins the claim. During that window a sibling schedule can be refused at `begin` for capacity that is about to be released. That's the conservative direction (defer rather than burn), and it's self-correcting, since the refused sibling drops its own first-sight latch and gets made up by catch-up. Worth a sentence in the comment block if you want the next reader to not have to work it out.

I re-reviewed both commits and found nothing wrong with them. Still comment-only, no approval.

oxoxDev:
Thanks for the thorough re-review, @M3gA-Mind — glad the split-and-reorder read as fixing the three at the root.

Took your one remaining note (`17b9c43f`, comment-only): the block above `begin` now spells out the guard-held-across-`claim_fire().await` window — a cap slot briefly occupied by a run that may never start when a peer replica wins the durable claim, a sibling refusable at `begin` for capacity about to be released, conservative (defer never burn) and self-corrected by catch-up — so the next reader doesn't have to derive it. No behaviour change; `fmt` + gated `cargo check` clean.

Re-requesting review on the doc delta.

Issue #674, opened 2026-08-11
approvals: #338's unbounded-reach rule and #614's "tool_call nodes don't gate under `full`" are contradictory — which governs?


M3gA-Mind:
> **Rewritten and widened.** The first version framed this as being about `http_request`. That understated it, on a premise that turned out to be false — see "The scope is wider than first stated" below. The question for #614's owner is unchanged; what it applies to is much larger.

## The question, for #614's owner

**Which premise governs: #338's "unbounded reach stops regardless of mode", or #614's "a `tool_call` node does not gate under `full`"?**

They are contradictory for the **whole `tool_call` family**, and only #614's owner knows whether that position was a deliberate product decision or the pre-#338 status quo written into tests.

## Two premises, both defensible, in direct contradiction

**#614 (merged as #627)** — `src/workflows/`, two tests that state the position by name:

```
workflows::gate::tests::full_autonomy_leaves_the_graph_untouched
workflows::gated_tool_call_test::the_same_call_executes_when_the_policy_does_not_gate_it
```

**Both use `shell`.** A named test is a stated position: under `full`, a `tool_call` node runs unattended — including arbitrary code execution.

**#338 (PR #660)** — the per-call judgement gate stops a call whose reach cannot be bounded from here, regardless of mode, producing:

```
GatedCall {
    node_id: "fetch",
    slug: "http_request",
    reason: "'http_request' can run arbitrary code or reach an arbitrary address, \
             so what it changes cannot be bounded from here",
    target: Some("GET api.example.com"),
}
```

Both cannot hold.

## The scope is wider than first stated, and no carve-out can resolve it

This issue originally proposed deferring `http_request`, `curl` and `web_fetch` — which #660 does. That rested on "`shell` is untouched by #614". **That is false.** Both tests above use `shell`, so the conflict covers the tool_call family, not the web tools.

**Deferring `shell` is not available as a resolution.** `shell` is the vector by which a run *deletes*, and #338's acceptance is:

> A run that would send, publish, **pay, or delete** stops for approval regardless of mode

Carving out `shell` would delete the acceptance criterion the issue exists for, while appearing to ship it. Two carve-outs is already the limit of what one PR can honestly do; a third that removes the subject of the acceptance is not scoping, it is abandonment dressed as delivery.

So this needs a **decision between two designs**, not another exclusion list.

## How this arose

[...]

PR #677, opened 2026-08-11
feat(harness): bring create_workflow tool to config parity so agents author wired workflows (#661)

oxoxDev:
## Summary

Part of #661 (workflows module audit), finding **H1**. The agent-facing `create_workflow` tool advertises six node kinds but could only ever wire two.

`CreateWorkflowArgNode` accepted only `{id, kind, name, summary, agent}` and `impl From<CreateWorkflowArgs> for RawWorkflow` hard-coded `config: None, destination: None` for every node — while the tool's schema still advertised `trigger`/`agent`/`tool_call`/`http_request`/`condition`/`output`. So `tool_call` (needs `config.slug`), `http_request` (needs method+url), `condition` (needs `config.field`), and `output` (needs a `destination`) were **advertised-but-dead**. On current `main` it's worse than a silent-dead-node: the shared validator (`validate_draft_against_record` → `validate_tool_call_node`) now **rejects every agent-authored `tool_call` at create** ("names no `slug`"), so the tool advertises `tool_call` yet can never author one. This is why a chat-authored GitHub workflow came out unwired.

Fix: add `config` (JSON) + `destination` to the tool's node args and JSON schema, mirroring what the REST body (`CreateNode`) and the #580 builder (`WorkflowNodeSpec`) already carry into the **same** shared core `create_company_workflow` — so the existing validation (slug present + wired namespace + `[tools].allow` grant) is inherited for free. Per-kind authoring guidance added to the tool description, including the key steer: **for Composio / GitHub / media integrations use an `agent` node, not `tool_call`** (those are agent-turn families, rejected in workflow `tool_call`, which runs only wired shell/code/web/search).

Policy fields (`schedule`/`retry`/`on_error`/`requires_approval`) stay omitted — agent-authored workflows remain manual-run only. Only functional `config`/`destination` were added.

## API Or Behavior Changes

`create_workflow` tool: node objects now accept optional `config` (object) and `destination` (`{kind: owner|email|channel, target}`). `impl From` became `impl TryFrom` (JSON→TOML config is fallible — TOML has no null; a null-valued config is rejected with an actionable message). No REST/HTTP API change; no change to the shared create core. A slug-less `tool_call` is still rejected, now with a useful message instead of being the only possible outcome.

## Tests

- [x] `cargo test` (default 1979/0; gated `--features openhuman,tinycortex` lib 2963/0)

The `harness` module compiles under `feature = "openhuman"`, so these tests run in the gated lane (CI already runs it). New tests:
- persists a `tool_call` `config.slug` (round-trip wired, not `config:None`)
- persists an `output` `destination`
- rejects a slug-less `tool_call` (inherited gate, reachable message)
- rejects an agent-turn `tool_call` slug e.g. `composio_execute` (`#[cfg(feature="openhuman")]` — the exact GitHub-workflow failure mode)
- rejects a JSON `null` in `config` ("TOML has no null")

## Documentation

Struct doc-comment rewritten so the file no longer self-contradicts (the narrowing is policy-fields-only; `config`/`destination` are now accepted). Per-kind guidance lives in the tool description itself, which is the model's contract.

## Related

Part of #661. Tightening `condition.field` / `http_request` method+url validation is #661 M1 (a separate PR in the shared validator) — deliberately out of scope here to keep the parity change clean.


M3gA-Mind (commented):
Reviewed manually (`gh pr diff` plus `workflow_create.rs`, `workflows/caps/tools.rs`, `workflows/gate.rs`, `harness/build.rs` at the PR head) rather than through our review harness — its `sync` step is openhuman-resident and I didn't want it checking out branches in a clone another agent is using. **Comment only, no approval.**

The motivating problem is real and well diagnosed: `validate_tool_call_node`'s slug-presence check is ungated, so on current main this tool advertises `tool_call` and then refuses every graph containing one. A tool that cannot author its own advertised node kind is broken, and closing that is worth doing. The null-config `TryFrom`, the round-trip assertions against the persisted TOML, and reusing `WorkflowDestinationDef` instead of a parallel shape are all the right calls, and the tests fail if you revert the fix.

My concern is not with `config` as such — it's with `config.args`, and it's a policy question rather than a code defect.

---

### 1. `config.args` lands on the wrong side of the #674 boundary

We settled #674 this week as: **#338's unbounded-reach rule governs the agent path; #614's position governs saved workflow nodes** — and the stated justification for giving saved nodes the more permissive treatment was that such a node has passed **two operator gates**, the manifest grant and then authoring. #674 also carved out one explicit exception: **a node whose arguments are templated from an upstream node's output is not pre-declared, and must follow the agent rule.**

This PR touches both halves of that.

**The second gate stops being an operator.** After this change an agent authors the node. The manifest `[tools].allow` grant remains an operator decision; the authoring step does not. So a `tool_call` node created through this tool reaches run time with one operator gate behind it, not two — while still being treated as a saved workflow node, because that's what the run-time path keys on. The struct doc reasons carefully about exactly this for the policy fields ("a field the model cannot set is a field it cannot get wrong") and I agree with where it lands on `schedule`/`retry`/`onError`/`requiresApproval`. The same reasoning is not applied to how far `config.args` reaches.

**And nothing keeps templated args out.** Both the description and the JSON schema now advertise args explicitly:

```
"tool_call": `{ "slug": "<wired shell/code/web/search tool>", "args": {…} }`
```

`args` is free-form and passes through `toml::Value::try_from` into the saved graph verbatim. Author-time validation reads `config.slug` and nothing else — presence, no surrounding whitespace, a wired namespace, a granted namespace. It never inspects `args`. And `=`-expressions in node args are a live engine feature, resolved at run time; `gate.rs` says so directly, and treats it as load-bearing:

```rust
/// Node `args` may still carry unresolved `=`-expressions when this pass runs,
/// so a tool whose classification depends on its arguments … would be classified
/// against a template and could be gated wrongly in either direction.
```

That test (`every_reachable_workflow_tool_is_classified_by_name_alone`) holds today only because no reachable workflow tool changes its reach with args. It is a guard on the *classifier*, not on *provenance* — it says nothing about who chose the args or where they came from.

So the concrete shape: a company with `[tools].allow = ["shell"]` (or `["*"]` — `grants_cover` treats `*` as covering everything, and `shell` is in `WORKFLOW_TOOL_NAMESPACES`). An agent calls `create_workflow` with

```json
{ "id": "run", "kind": "tool_call", "name": "Run",
  "config": { "slug": "shell", "args": { "command": "=<expression over an upstream node's output>" } } }
```

This clears every author-time gate — the slug is present, unpadded, wired, and granted. At run time it is a saved workflow node, so it gets #614's position. But its arguments were chosen by a model and may be templated from upstream output, which is precisely the case #674 said must follow the agent rule. The distinction #674 drew is not represented anywhere in the data, so nothing downstream can honour it.

I don't think this sinks the PR, and I'm not asking you to re-litigate #674 inside it. But it needs an explicit answer, because right now the change makes the decision silently:

- If #661/H1 intends agent-authored `tool_call` nodes to carry the saved-node position, say so in the struct doc where the policy-field reasoning already lives, and flag #674's templated-args carve-out as needing revisiting — it can't hold as written once agents author nodes.
- If not, the cheap version is to reject `=`-expression args on a `tool_call` authored through *this* tool — literal args only, with templated wiring left to the console and `POST …/workflows` where an operator picked them. That keeps the parity win (the tool can author a working `tool_call` again) without moving the boundary at all, and it's a few lines next to the existing slug checks.

Either is fine by me; deciding it in the open is the part I care about. I've flagged this to our side as a decision rather than a review nit.

### 2. `config` and `destination` are accepted on every node kind

`validate_draft_against_record` matches only `"agent"` and `"tool_call"`; every other kind falls through `_ => {}`. So a `destination` on a `trigger`, or a `config` on an `output`, saves cleanly and is inert. The schema says "On an `output` node only" but that's documentation, not enforcement.

This is inherited from the REST path rather than introduced here — but the REST path has an operator looking at a console. Here the caller is a model that gets a success result and no signal that the field it set does nothing, which is how you get a confidently-wrong retry loop. Rejecting a `destination` on a non-`output` node is cheap and mirrors the slug gate's spirit.

Relatedly: `config` is typed `Option<serde_json::Value>` while the schema declares `"type": "object"`. Nothing enforces the schema at deserialization, so `"config": "web_fetch"` parses and persists as a TOML scalar. On a `tool_call` it degrades into the "names no `slug`" refusal, which is fine; on `http_request` / `condition` it saves silently.

### 3. Two smaller things

- The conversion error interpolates `{err}` (good) but then always appends the null-specific remedy: `"TOML has no null; drop null-valued keys."` `toml::Value::try_from` can fail for other reasons — an integer outside `i64` range, for one — and the fixed tail will then point the model at a key it doesn't have. Worth making the remedy conditional, or dropping it and letting `{err}` speak.
- The namespace/grant half of the gate and `create_workflow_tool_rejects_agent_turn_tool_call` are both `cfg(feature = "openhuman")`, so under the default build only slug-presence runs at save. The tool description states the shell/code/web/search restriction unconditionally ("are rejected on a `tool_call`"). Run time is still the backstop so nothing unsafe follows, but in a default build that rejection happens at first run, not at save.

---

Nothing in (2) or (3) blocks. (1) is the one I'd want answered before this merges, and the answer can be a paragraph in the doc comment rather than more code.

oxoxDev:
Deciding the #674 boundary in the open, and taking your cheap/safe option:

**The tool now rejects any `tool_call` whose `config` carries an `=`-expression** (tinyflows' own `is_expression` convention — leading `=`; applied recursively over the whole tool_call config). Literal args only for an agent-authored `tool_call`; templated wiring stays with the console + `POST …/workflows`, where an operator picks the args. This keeps the parity win without moving the #674 boundary — an agent-authored node reaches runtime with saved-node position, so we don't let it carry model-chosen templated args (exactly your carve-out). Documented the reasoning in the `CreateWorkflowArgs` struct doc alongside the schedule/retry policy-field reasoning, and flagged there that #674's templated-args carve-out needs revisiting for agent-authored nodes **generally**, not just this tool.

- **(2)** `destination` on a non-`output` node is already rejected end-to-end via the shared `validate` (`create_workflow` → `create_company_workflow` → `parse_workflow` → `validate`) — added `create_workflow_tool_rejects_destination_on_non_output` proving it, no duplicated validation. Non-object `config` (`"config":"web_fetch"`) is now rejected in the `TryFrom`.
- **(3)** The null remedy is now conditional (only appended when the failure is actually about a JSON null; other `toml::Value::try_from` failures let `{err}` speak). Softened the description so it doesn't over-promise author-time namespace rejection in a default build (run time stays the backstop).

Gated lane 2971/0. SHAs: `2b03c079`, `7398b344`, `7d5794d6`.

M3gA-Mind (commented):
**Re-review.** I commented on this at `7ddcc508`; head is now `7d5794d6`. Reviewed the three commits since (`2b03c079` rejecting templated/malformed config, `7398b344` softening the wording, `7d5794d6` tests) — `git diff 7ddcc508 7d5794d6`, one file, +415/−16. Manual again, same reason. **Comment only, no approval.**

**First, a correction to my own last review, because it sent you to write a test for something that already worked.**

I said `destination` was accepted on every node kind and that a `destination` on a `trigger` would save silently and be inert. That was wrong. I read `validate_draft_against_record` and stopped there — but that function is the *roster and grant cross-check*, not the structural validator, and its `_ => {}` doesn't mean unvalidated. The shared `validate` in `workflow_file.rs` has rejected this since #170, on `main` today:

```
"{label} sets `destination` but is a `{}` node — only `output` nodes route a report."
```

and `create_company_workflow` runs it, so the tool inherited that refusal from the start. `create_workflow_tool_rejects_destination_on_non_output` is therefore pinning behaviour that was already there rather than covering a hole I found. Keeping it is fine — it's a real regression test on an inherited guarantee, and its assertion is honest about where the message comes from — but you should know it wasn't a gap.

---

**On the substance: the templated-args rejection is complete, and I checked it against the engine rather than against the diff.**

This was the thing I most wanted to see, and the concern with a check like this is that it catches `"=item.url"` and misses the awkward shapes. It doesn't:

- **Nested in an object** — `first_expression_location` recurses through `Value::Object`.
- **Inside an array** — it recurses through `Value::Array` too, indexing elements numerically.
- **Partly literal, partly templated** — this shape does not exist in this engine, which is the part worth stating rather than assuming. `tinyflows::expr::is_expression` is exactly `s.starts_with('=')`, with a property test asserting that equivalence, and the resolver's doc is explicit that *"Non-expression values are returned as a literal clone."* There is no `{{ }}`-style interpolation, so a string is wholly an expression or wholly a literal. `starts_with('=')` isn't an approximation of the engine's rule — it **is** the engine's rule.
- The dotted location it reports (`args.cc.0` shape, objects by key and arrays by index) matches the engine's own `NullResolution.location` convention exactly, which suggests this was derived from the engine rather than guessed.

**And it's targeted at the right node kind.** I went looking for whether restricting this to `tool_call` leaves `http_request` as an equivalent hole — an agent-authored `http_request` with a templated `config.url` is arbitrary egress with a model-chosen destination. It isn't a hole, and the reason is in `call_of`: `NodeKind::HttpRequest` returns a call classified as `HTTP_REQUEST_TOOL`, which is `Reach::Consequence`, so an `http_request` node **always** parks for approval. A `tool_call`'s verdict depends on its slug, so a low-consequence slug executes with no card at all. `tool_call` is precisely the kind where templated args would be un-gated reach, and it's the only one restricted. That asymmetry is principled, not an oversight — worth a line in the comment block, since the next reader will ask the same question I did.

`create_workflow_tool_allows_condition_expression_field` is the other half of that discrimination and I'm glad it's there: a `condition`'s `config.field` is *inherently* an expression, so a blanket "no `=` anywhere" would have broken the node kind the same PR requires a field on.

**This also resolves the #674 question I raised without needing the ruling reopened.** My concern was that agent-authored `tool_call` nodes would carry the saved-node position while their arguments were model-chosen and possibly templated from upstream output — the exact case #674 says must follow the agent rule. Refusing templated args removes the case rather than deciding it: templated wiring stays with the operator surfaces where a human picks the args, and what an agent can author is literal values that are visible in the saved graph. That's the conservative answer and it preserves #674's distinction intact. The struct doc now cites #674 and states the reasoning, which was the other thing I asked for.

**The softening is wording only.** `7398b344` is two string literals, 2 insertions and 2 deletions, no validation logic in the commit at all. It changes *"can run ONLY a wired … tool"* to *"runs ONLY a wired … tool"* and *"are rejected on a `tool_call`"* to *"a non-wired slug is refused when the node runs"* — which corrects the over-promise I flagged (the namespace/grant gate is `cfg(feature = "openhuman")`, so a default build refuses at run time, not at save). It also *adds* the literal-args constraint to both the description and the schema. So it narrows nothing; it makes the description honest and tightens what the model is told.

**The two smaller things I raised are both handled:** the null hint is now conditional on `json_contains_null`, so a non-null conversion failure gets the converter's own message — the comment even names the `i64`-range case — and it has its own test. A non-object `config` is now refused with an agent-actionable message rather than persisting as an inert TOML scalar.

---

I re-reviewed all three commits and found nothing wrong with them. The one item from my original review that remains genuinely open is nothing in this delta — it's that my `destination` finding was mistaken, corrected above. Still comment-only, no approval.

Issue #675, opened 2026-08-11
workflows: cancelling a run does not stop a sub_workflow child, so effectful nodes keep running after Cancel


M3gA-Mind:
## What happens

Cancelling a workflow run does not stop a `sub_workflow` child that is already running. The child keeps walking its own graph — executing further effectful nodes — after the operator has pressed Cancel, and the parent cannot wind down until the child finishes on its own.

**Verified end-to-end**, not inferred. See the reproduction below.

## Mechanism

`run_sub_workflow` (`vendor/openhuman/vendor/tinyflows/src/engine.rs:813-840`) runs the child as a separate engine invocation and hands it a **fresh token**:

```rust
let (_graph, _thread_id, outcome, _run_ids) = build_and_run(
    workflow,
    input,
    capabilities,
    &observer,
    checkpointer,
    thread_id,
    None,
    Some(json!({ "sub_workflow_depth": depth, "max_sub_workflow_depth": max_depth })),
    CancellationToken::new(),   // <-- not the parent's
)
```

The engine checks the token *before each node*, so the child's node loop is checking a token that nobody can ever cancel.

It cannot be fixed at the node executor either: `NodeContext` (`src/nodes/mod.rs:25-41`) carries `node`, the input items, `run`, the `nodes` state and `caps` — **no token** — so the sub-workflow executor has nothing to pass down. That is structural, not an oversight in this one call.

## What this is NOT

Worth stating, because the alarming reading is wrong and would waste a reviewer's time.

**No orphaned child.** The child future is awaited inline inside the parent's node handler, so it cannot outlive the parent: OpenCompany's hard-abort path drops the engine future (`src/workflows/runner.rs:431`), which drops the child with it. Nothing is left running after the run object is gone.

The real cost is different, and it defeats a guarantee #398 deliberately built:

- **Cancel latency is bounded by the child's remaining work, not by a node boundary.** `run_workflow` flips the engine token and gives the run `CANCEL_HARD_ABORT_GRACE` (5s) to wind down cleanly. A parent sitting in a sub-workflow node cannot reach a node boundary until the child returns.
- **Past that grace the run takes the hard-abort branch** — the future is dropped mid-await, so the run loses the clean `cancelled` outcome and the per-node progress trail that #398 exists to preserve. *(This consequence follows from the runner code; the reproduction below exercises the under-5s case, where the run still settles cleanly.)*
[...]

PR #679, opened 2026-08-11
workflows: record a sub_workflow child's calls that were never offered for approval

M3gA-Mind:
## Summary

Partially addresses #617 — the **audit** half, deliberately not the behaviour half.

A `sub_workflow` child is resolved and run inside the engine, so its nodes never reach the per-run gate pass that #612 and #627 added. A call the company's policy would park at the top level therefore runs unparked one level down: real effects, company grants, no card, and **nothing anywhere recording that it happened**.

This does not gate the child, and that is the point of the change rather than a limitation of it. Gating is not available: tinyflows cannot resume a child across the sub-workflow boundary, and a paused child halts the parent with an unresumable error — so marking the child's node would convert a run that works today into one that fails with **no card to decide**. That is strictly worse than the hole, and it is the same regression #460 refused to ship.

So the run proceeds and the omission is written down. On resolving a child, the resolver asks the *same* classifier the top-level gate asks, and for every call the policy would have parked it emits a `warn!` and journals a new `CompanyEvent::WorkflowChildCallNotOffered` naming the child, the node, the tool and the policy's own reason.

The value is exactly the value #460 argued for: an operator reviewing what the company was asked to approve stops seeing a gap with nothing to explain it.

## API Or Behavior Changes

**Behavior.** No change to what executes. A company whose policy would park a call inside a `sub_workflow` child now gets one journal line and one warning per such call. A company whose policy would not park it gets nothing. A dry run gets nothing — it executes nothing, so there is no omission to confess.

**API.**
- New `CompanyEvent::WorkflowChildCallNotOffered { workflow_id, child_workflow_id, run_id, node, tool, reason }`. Classified **Permanent** in `retention_class` — it is the record that a consequential call ran *without* the operator being asked, which is the clearest evidence this enum carries; pruning it would delete the only trace that the company acted unapproved. Neutral in both `cycle.rs` classifiers (it starts no cycle and names no conversation), and carried structurally-only onto the medulla wire and the orchestrator's insight summary — the policy's `reason` is a sentence built for an approval card and deliberately does not go to those surfaces.
- `StoreWorkflowResolver::new` takes a fifth argument, `Option<ChildCallAudit>`. `None` restores the previous behaviour exactly, which is what a dry run passes.
- `gate::policy_gates` is split out of `apply_policy_gates`: classification without mutation. **One classifier**, so the audit line and the gate can never disagree about the same call.

The compiler named five exhaustive matches on `CompanyEvent`, not the six I had been told to expect — I ran it rather than trusting the count.

## Tests

Gated lane throughout. Exit codes captured with unpiped redirects to a session-private path.

- [x] `cargo clippy --all-targets -- -D warnings` — CI's own form, `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, `EXIT:0`.
- [x] `cargo build --all-targets` — covered by the clippy `--all-targets` run on the same feature set.
- [x] `cargo test` — `cargo test --features openhuman,tinycortex`, `EXIT:0`. Not the bare `cargo test`: default features skip this module entirely.

New tests (3), all in `src/workflows/caps/resolver.rs`:

- `an_ungated_child_call_is_disclosed_to_the_journal` — the defect. A child whose `tool_call` runs `shell` on a `supervised` company produces exactly one `WorkflowChildCallNotOffered` naming the child, node, tool and reason — **and the child graph is not gated**, which is asserted in the same test so the disclosure can never quietly become a behaviour change.
- `a_child_the_policy_would_not_park_discloses_nothing` — control. Without it, the test above is also satisfied by a resolver that logs on every child regardless of policy.
- `a_resolver_without_an_audit_discloses_nothing` — the dry-run shape.

**Revert-and-check.** With `disclose_ungated_calls` made an unconditional no-op, `an_ungated_child_call_is_disclosed_to_the_journal` fails and the two controls pass — which is what controls are for. Per-test results in the PR comment.

**Not covered:** that the journal line reaches any console surface. It is journaled and wire-mapped; no UI reads it yet.

## Documentation

`disclose_ungated_calls` carries the reasoning for disclosing rather than gating; the new event variant documents why it is Permanent and why it is journaled at child *resolution* rather than at the node. #617 has been rewritten around the (a) silent / (b) loud split this work is based on.

## Related

- #617 — the tracking issue, corrected to separate the silent policy-gate hole from the loud authored-gate failure.
- #612 / #627 — the top-level gate this reuses the classifier from.
- #675 — a second, independent sub-workflow boundary defect found while scoping this one: cancelling a run does not stop a running child. Same structural cause, different consequence; verified end-to-end.

### DIFF
--- a/src/brain/medulla/effects.rs
+++ b/src/brain/medulla/effects.rs
@@ -416,6 +416,24 @@ pub(crate) fn wire_event(seq: u64, event: &CompanyEvent) -> WireEvent {
             format!("Workflow {workflow_id} delivered its {kind} report from node {node}"),
             "workflow.report_delivered",
         ),
+        // Issue #617. Structural only, like every arm here: the child graph,
+        // the node and the tool. The policy's `reason` is deliberately NOT
+        // carried onto the wire — it is a sentence built for an operator's
+        // approval card, and this surface is a short non-sensitive one-liner.
+        CompanyEvent::WorkflowChildCallNotOffered {
+            child_workflow_id,
+            node,
+            tool,
+            ..
+        } => (
+            Role::System,
+            "workflow".to_string(),
+            format!(
+                "Workflow child {child_workflow_id} ran {tool} at node {node} without offering \
+                 it for approval"
+            ),
+            "workflow.child_call_not_offered",
+        ),
     };
     WireEvent {
         seq,
--- a/src/harness/orchestrator.rs
+++ b/src/harness/orchestrator.rs
@@ -1062,6 +1062,14 @@ fn summarize_event(event: &CompanyEvent) -> String {
             kind,
             ..
         } => format!("workflow {workflow_id} delivered {kind} report from node {node}"),
+        // Issue #617. Structural only, and without the policy's `reason` for
+        // the same rule the arms above follow.
+        CompanyEvent::WorkflowChildCallNotOffered {
+            child_workflow_id,
+            node,
+            tool,
+            ..
+        } => format!("workflow child {child_workflow_id} ran {tool} at node {node} unapproved"),
     }
 }
 
--- a/src/ports/types.rs
+++ b/src/ports/types.rs
@@ -1134,6 +1134,49 @@ pub enum CompanyEvent {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         target: Option<String>,
     },
+    /// A call inside a `sub_workflow` child ran **without** being offered for
+    /// approval, on a company whose policy would have parked the same call at
+    /// the top level (issue #617).
+    ///
+    /// # Why this is a journal line and not a fix
+    ///
+    /// A child graph is resolved and run *inside* the engine, so its nodes never
+    /// reach the gate pass that #460 and #614 added. Gating them is not
+    /// available: tinyflows cannot resume a child across the sub-workflow
+    /// boundary, and a paused child halts the parent with an unresumable error —
+    /// so marking the child's node would convert a run that works into one that
+    /// fails with no card to decide. Until the engine can resume across that
+    /// boundary, the honest thing is to keep running and **say so**.
+    ///
+    /// That is what this records. An operator auditing what the company was
+    /// asked to approve would otherwise see a hole with nothing explaining it;
+    /// this turns the hole into a line naming the child, the node and the call.
+    /// It is an audit record of a known limitation, not a decision anybody made.
+    ///
+    /// Journaled best-effort at child *resolution*, so it lands whether or not
+    /// the child's node is ultimately reached — the point is that the call was
+    /// never offered, and a run that stops earlier for another reason does not
+    /// make that less true. Deliberately **not** written for a dry run, which
+    /// executes nothing.
+    ///
+    /// Additive: an entirely new `kind`, so no journal written before it existed
+    /// carries it, and its presence changes how no existing variant serializes.
+    WorkflowChildCallNotOffered {
+        /// The graph the run started from — the one an operator recognises.
+        workflow_id: String,
+        /// The child graph the call actually sits in. Equal to `workflow_id`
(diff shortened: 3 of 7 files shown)


M3gA-Mind:
**Revert-and-check, per-test.** `disclose_ungated_calls` was made an unconditional no-op — the pre-#617 behaviour — and the resolver tests re-run:

| Test | Under revert | Why |
|---|---|---|
| `an_ungated_child_call_is_disclosed_to_the_journal` | **FAILED** | the defect test |
| `a_child_the_policy_would_not_park_discloses_nothing` | passed | control — catches a resolver that logs on every child regardless of policy |
| `a_resolver_without_an_audit_discloses_nothing` | passed | control — the dry-run shape |

The 11 pre-existing resolver tests also passed under the revert, so this change is isolated to the disclosure.

A control that *failed* under revert would mean it was testing the wrong thing, so 1-of-3 is the intended result rather than a thin one.

oxoxDev (approved):
The central decision is right, and it is the one most people would get wrong. Gating the child is the obvious move and it would convert runs that work today into runs that fail with no card to decide, because tinyflows cannot resume across the sub-workflow boundary — strictly worse than the hole, and the same regression #460 refused. Choosing to *record* the omission instead, and saying plainly in the summary that this is the audit half and not the behaviour half, is the right call and the right way to present it.

Checked the structural claim rather than taking it on trust: the `policy_gates` / `apply_policy_gates` split is faithful. `apply_policy_gates` is now exactly `policy_gates` plus the config mutation, the loop moved from `&mut graph.nodes` to `&graph.nodes` with nothing else altered, and the "written LAST and unconditionally" comment moved with the code it describes. So "one classifier, so the audit line and the gate can never disagree" is true by construction, not by discipline.

That matters more than it looks, because the classifier it shares is `call_of`, and `call_of` is now exhaustive over `NodeKind` with a comment naming why: two PRs each closed one hole in that match and neither announced itself. An audit built on a wildcard would have inherited the gate's blind spot in the one place a blind spot is least acceptable — the net that exists to catch what the gate missed. Built on an exhaustive match, the fourteenth node kind is a compile error in both at once.

Also checked: `reason` is genuinely excluded from both downstream surfaces — `..` in the medulla arm and in the orchestrator insight arm, each with the rule stated at the site. An approval-card sentence does not belong on a one-liner wire, and the three arms agree about that without depending on each other.

### Minor — the stated value does not exist yet, and it is the headline

The summary's case is that "an operator reviewing what the company was asked to approve stops seeing a gap with nothing to explain it." Under **Not covered** you say no UI reads the event. Both are true, and together they mean the operator-facing benefit is not delivered by this PR — what is delivered is a durable journal record, which is genuinely valuable on its own (post-incident, and `Permanent` retention is the right call for exactly that reason) but is not the same claim.

Either soften the summary to what ships, or wire the surface. If it stays deferred, **please file it as an issue and link it here rather than leaving it in the PR body** — a follow-up described only in a merged description reliably disappears, and this one is the difference between an audit trail and an audit trail somebody reads.

### Question — disclosure at resolution, and children that never run

The event is journaled when the child is *resolved*, and the summary says a dry run discloses nothing because it executes nothing. What about a child that is resolved and then does not execute — a parent that fails, is cancelled, or takes a branch that never reaches the node? Reading it as written, the line says a call ran unapproved when it may not have run at all. If resolution implies execution on this path, that is worth a sentence on `disclose_ungated_calls`; if it does not, the event's wording ("ran ... without offering it for approval") is stronger than what is known at the point it is written.

Worth noting #675 is the right instinct — a second boundary defect found while scoping this one, filed separately rather than bundled in.

**Before merging, not a finding:** every Rust lane is still pending. Everything you ran locally was the gated form against a current base, and this is the only one of your four open branches cut from current `main` — the other three want a rebase before they can even run. Approving on the code; don't merge while it is `UNSTABLE`.

Issue #678, opened 2026-08-11
feat: LLM triage layer for operator chat, and a settle for a card the turn already completed


oxoxDev:
## Context

Deferred half of #267. That issue shipped the two layers that need **no model call**:

- **Layer A** — `company::task_intent::triage_message`, a deterministic three-way classifier (`Track` / `Answer` / `Chatter`) in the REST chat handler. Compiled into every build; fronts both cognition brains.
- **Layer B** — on `Answer`, `DelegationRunner::handle_operator_message` withholds the issue-#453 delegation claim, so the model's own `spawn_task` / `delegate_to_desk` / `assign_task` / `review_task` are refused at the tool boundary in its own turn.
- **Layer C** — `orchestrator_brief()` restructured to lead with answering, and to say a "create a workflow" ask is authored *this turn*.

That closes both card doors for the classes a lexical detector can name. It does not give us the classifier #267's "Suggested shape here" section actually asks for.

## Why it was staged rather than shipped together

- OpenHuman's `trigger_triage` — the precedent #267 cites — classifies **external** triggers (Composio webhooks, cron fires). OpenHuman itself routes interactive operator chat straight to its orchestrator with no pre-turn model call.
- OpenCompany has **one** `ChatModel` per company and no fast-model tier. An LLM pre-classifier would therefore add a full-price serial round-trip to every operator message, plus routing infrastructure that does not exist.

So the cheap layers went first, deliberately, and this issue carries the rest.

## Scope

1. **Port `trigger_triage` for operator chat.** Its configuration is the interesting part: `temperature = 0.2`, `max_iterations = 2`, `sandbox_mode = "read_only"`, identity/safety/skills stripped, memory and user profile deliberately kept, and the explicit separation *"You do not act. You decide. Another component will carry out your decision."*
2. **Fast-model routing.** A second, cheap model tier the triage turn can use, so classification never pays orchestrator prices. This is the prerequisite for 1, not a nice-to-have.
3. **A first-class `automate` class.** #267 names four classes; only three are implemented. `automate` — "a repeatable process worth capturing" — currently rides on `Track` plus brief guidance telling the orchestrator to author the graph in-turn. A real class would let the runtime route it rather than hope.
4. **Hosted-path gating.** Layer B lands on the harness path only. `HostedMedullaBrain` has no delegation stack at all (#176), so there is no model board-write path to gate there yet. When #176 copies the drain site through the canonical `delegation_tools` seam, the conditional claim must come with it.

[...]

Issue #463, opened 2026-08-06
spine: a substantial ask that publishes opens two cards, and the reply links to the empty one


oxoxDev:
## What happens

Ask for something substantial that ends in a published file. You get **two cards**. The one the reply bubble links to — the one you click — has **zero artifacts**. The deliverable is on a second card that nothing links to.

This is #442's own defect ("the board does not reflect the work") reproduced by #442's fix, interacting with #445's fix. Both are correct in isolation; together they double.

Measured on a live host (real stores, real auth, production router on a real socket, driven over HTTP against a scripted model endpoint keyed on which agent is asking):

| Case | Cards | Detail |
|---|---|---|
| Substantial ask → delegate → publish | **2** | #442's card (0 artifacts, **wins the reply link**) + #445's publish card (**holds the artifact**, orphaned) |
| Same, sent straight to a desk | **2** | identical split |
| Recognised imperative, delegated | **2** | `task_intent` card + delegation card |
| Recognised imperative, delegated **and** published | **3** | all three paths at once |
| Substantial ask, no publish | 1 | correct |
| Publish with no card in scope | 1 | correct, carries the artifact |
| Trivial question / pleasantry | 0 | correct |
| Five delegations in one turn | 3 opened, 2 refused | correct |

## Two independent doubles

**1. Publish mints unconditionally.** `record_conversation_publishes` always creates a card, even when this turn already opened one.

The information needed is already in scope: `OperatorTurn::spawned_task` is seeded from `direct_card_id` and from each delegation's card, so by the time the publish queue is drained in the `OperatorMessage` arm, it **already holds the id of the card opened for this message** — which is exactly why the reply links there.

**2. The `task_intent` dedupe guard is incomplete.** The stand-down against the pre-existing REST-layer detector lives only in `open_direct_work_card`. The `run_delegation` path calls `open_work_card` with no such guard, so a recognised imperative that the orchestrator hands off produces both the REST card and the delegation card.

## A third, smaller one in the same case

The publish card's **assignee is wrong** when the publish happens inside a delegated desk turn: the writer published, the card says the operator-turn responder. `record_conversation_publishes` receives the wrong agent.

## What the fix should do

- In the `OperatorMessage` arm, file drained publishes onto `turn.spawned_task` when it is `Some`, and mint a card only when it is `None`. That collapses the two-card cases to one card carrying the artifact, leaves the no-card-in-scope case unchanged, and makes the reply link a non-choice because only one card exists.
[...]

PR #680, opened 2026-08-11
feat(chat): triage a message as answer, work, or chatter before it becomes a card (#267)

### LINKED ISSUE #267: feat: triage an incoming message before it becomes a task card
## Problem

Almost anything said to the company turns into a task card, including questions that just want an answer.

Nothing auto-creates cards — there is no chat-message-to-card pipeline. The orchestrator *decides* to call `spawn_task` or `delegate_to_desk`, and both persist a `TaskRecord`. So this is a judgement the model makes on nearly every turn, and it currently makes it toward acting.

Observed on a live harness company, all six sitting unworked in `backlog`:

```
Tell what is there in the tasks list
Call the composio_authorize tool with toolkit = gmail and reply with ...
Create a workflow to, write a 2-sentence status update ...
Create a simple workflow, Write a 2-sentence status update ...
Create a workflow named topic-to-visual ...
Create a workflow called Daily Standup ...
```

The first is a pure read — the answer is one `query_company` call. It became a card instead. Elsewhere on the same board, "Reply with a one-sentence hello" is also a card.

The discrimination rule already exists, as the closing clause of `orchestrator_brief()` (`src/harness/orchestrator.rs:128-149`):

> "Delegate, run or create a workflow, add a teammate, or act on the board only when it genuinely helps — otherwise answer directly and concisely."

But it is one sentence at the end of a brief that spends its length enumerating seven action tools and how to use each. The structure reads as an invitation to act with a caveat attached, and behaviour follows the structure rather than the caveat.

This matters more than chat noise would. The board is the company's work surface. Six dead cards in `backlog` is worse than six ignored chat lines, because nothing clears them and they dilute the signal of what the company is actually doing.

## Reproduction

1. Boot a company with inference configured so the harness is live.
2. In the console, ask a plain question: "what is there in the tasks list?"
3. Board → Backlog now holds a card titled after the question, assigned to the orchestrator, which never advances.

## Expected behavior

An incoming message gets classified before anything is written to the board: is this a question to answer, work to track, work to delegate, or a process worth automating?

### Precedent — OpenHuman already has this shape

`openhuman/src/openhuman/agent_registry/agents/trigger_triage/` is a dedicated classifier agent. Its configuration is the interesting part:

- `temperature = 0.2`, `max_iterations = 2`, `sandbox_mode = "read_only"`
- identity preamble, safety scaffolding and skills catalog all stripped (`omit_identity`, `omit_safety_preamble`, `omit_skills_catalog`) — it needs app state, not persona
- memory and user profile deliberately **kept**: the same input is urgent for one user and noise for another
- its own model hint (`local-or-remote-fast`), so it never pays orchestrator prices

And the separation is explicit in its prompt:

> "You do **not** act. You decide. Another component will carry out your decision."

It returns exactly one of `drop` / `acknowledge` / `react` / `escalate`, with written tie-breakers, biased toward dropping — "over-escalating wastes agent time".

The gap: OpenHuman wires this to *external* triggers (Composio webhooks, cron fires). Operator chat goes straight to its orchestrator. So this is extending a proven pattern to a new input class, not a port.

### Suggested shape here

A triage step ahead of the orchestrator, returning one of:

- **`answer`** — a question about state. Route to a read path (`query_company`) and reply. No board write.
- **`track`** — real work with no obvious owner. Card in `backlog`.
- **`delegate`** — real work belonging to a desk. Card plus `delegate_to_desk`.
- **`automate`** — a repeatable process. Propose `create_workflow`, do not silently author one.

Tie-breakers should lean toward `answer`, mirroring OpenHuman's lean toward `drop`: a missed card costs one follow-up message, a spurious card pollutes the board permanently.

### Interim

Rebalancing `orchestrator_brief()` so answering is the stated default and board writes are the exception needing justification is a prompt-only change that would catch the obvious cases. Worth doing regardless, but it does not remove the decision from an agent whose brief is mostly a list of actions.

### Note on existing plumbing

A card spawned during a chat turn already carries `origin_chat_id`, and `src/harness/lifecycle.rs` treats those differently from board-native cards. So "this came from a conversation" is already a first-class distinction a triage layer could build on.

## Alternatives

- **Make the orchestrator declare its intent before it may touch a board tool.** Auditable and cheaper to build, but it leaves the judgement inside the same expensive agent that is already biased toward acting.
- **Leave it and let operators archive noise.** Costs nothing to build and pushes the cost onto every operator forever; also requires a bulk-clear affordance the board does not have.
- **Confirm before creating a card.** Safe, but a confirmation prompt on most turns is its own kind of noise.

oxoxDev:
## Summary

Ask the company a plain question and you get a card on the board that nobody ever works. This classifies an operator message as **answer**, **work**, or **chatter** before anything is written to the board, and closes both of the doors a dead card comes through.

**The issue's premise needed correcting first.** #267 states "Nothing auto-creates cards — there is no chat-message-to-card pipeline". There is one: `detect_task_intent` in `src/company/task_intent.rs`, wired into `run_chat`. Evaluated against its actual rules, **five of the six dead cards the issue observed came through that deterministic detector** ("Call the composio_authorize tool…" and 4× "Create a workflow…"); only "Tell what is there in the tasks list" — where `tell` is not an action verb and there is no request frame — can have come from the model calling `spawn_task`. Cards enter through *both* doors, so a fix on either one alone leaves the other open. That is why this is three layers rather than the single triage step the issue describes.

**Layer A — the detector answers a positive question.** `triage_message` returns `Answer` / `Track(title)` / `Chatter`. It now *recognises a question* rather than merely failing to recognise a command: interrogative lead words, read verbs (`tell`, `show`, `explain`, `describe`, "give me"), or a trailing `?` with no action verb. Request frames still win over the question mark, so "can you build the landing page?" stays work. `list` is removed from the action verbs — a bare "list X" is a read. `detect_task_intent` survives as a thin wrapper so the #463 byte-for-byte card-title contract is untouched. Everything ambiguous falls to `Chatter`, which neither cards nor gates.

**Layer B — the model may not write to the board on a question turn.** A brief cannot close that door; it is guidance, and the issue's own diagnosis is that behaviour follows structure rather than caveats. Withholding the #453 queue claim closes it using semantics that already exist: with no claim, `spawn_task` / `delegate_to_desk` / `assign_task` / `review_task` hit `Staged::NoDrain` and fail **in the model's own turn** with a message telling it not to retry and not to report the action as done (verified at `orchestrator.rs` `no_drain()` — it is a real tool error, not a silent drop). Read tools are untouched: `query_company`, `run_workflow` and `read_run_output` run inline and never reach the queue, so the turn keeps every means of actually answering. The asymmetry is the design — it removes the ability to write, not the ability to reply. The direct-desk card path stands down on `Answer` too, so asking a desk lead "what did you ship this week?" opens no card.

**Layer C — the brief leads with answering.** It now opens with the default ("most messages are questions or quick reads — answer from `query_company` and touch nothing else; a question about state is NEVER a card") instead of trailing it as a caveat after eight enumerated action tools. It also says a "create a workflow" ask is authored **this turn** with `create_workflow`, not parked — which is what un-deadens that whole class of card.

**No added model calls.** An LLM pre-classifier was considered and rejected for now: OpenHuman's `trigger_triage` precedent classifies *external* triggers, and OpenHuman itself routes interactive operator chat straight to its orchestrator with no pre-turn model call. There is one `ChatModel` per company here and no fast-model tier, so a pre-classifier would put a full-price serial round-trip in front of every message plus routing that does not exist. Deferred to **#678**, opened alongside this PR.

## API Or Behavior Changes

No HTTP API change — no route, request body or response shape moves.

Behavior:

- A message triaged `Answer` opens **no** card, on any of the three paths (REST handler, model delegation tools, direct-desk hand-off), and the model's board-write tools fail in-turn on that message.
- A message triaged `Track` behaves exactly as before, with byte-identical card titles.
- `Chatter` — the ambiguous bucket — neither cards nor gates, so a triage miss costs a follow-up message rather than silently refusing work.
- `list` no longer counts as an action verb, so "list the tasks" is a read.
- Copilot-thread suppression (#416) is preserved unchanged.

**Cross-brain**: Layer A lives in `run_chat`, is compiled in every build, and fronts both brains. Layer B is harness-path only — `HostedMedullaBrain` has no delegation stack at all (#176), so there is no model board-write path to gate there; when #176 copies the drain site through the canonical `delegation_tools` seam, the conditional claim comes with it.

**Three places the code contradicted the plan, stated rather than smoothed over:**

1. **The gate is narrower than "board tools are withheld" suggests.** `create_workflow` and `add_agent` are also inline and never touch the delegation queue, so they remain available on an `Answer` turn. Harmless in practice — a "create a workflow" ask triages `Track`, not `Answer` — but the boundary is queue-membership, not intent.
2. **Skipping the claim would have skipped its clear.** `claim()` clears the queue on acquire, and that half is a safety property, not a permission: a delegation staged by some other path must never be executed by this turn's drain. The answering path therefore does `clear()` and holds no claim, rather than simply not claiming.
3. **`is_actionable` collapsed.** With the request-frame branch firing earlier in the new ordering, it reduced to `starts_with_action`; the wrapper was removed rather than left with a dead branch.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — 0 errors
- [x] `cargo build --all-targets` — ok
- [x] `cargo test` — 1985 + 10 + 1 passed, 0 failed

Additionally, because the harness and delegation tests only compile under the full feature combo and no CI lane builds it:

- [x] `cargo clippy --no-deps --features openhuman,mcp,telegram --all-targets -- -D warnings` — 0 errors (`--no-deps` required, per `ci.yml`)
- [x] `cargo test --features openhuman,mcp,telegram` — 2947 + 10 + 1 + 11 + 2 passed, 0 failed
- [x] frontend `npm ci && npm run typecheck && npm run build` — exit 0, no source change (run to prove no drift)

New coverage: a regression block asserting the triage class of all six messages observed in the issue; `Answer` positives ("list the tasks", "show me the board", "what's our revenue?", "Tell what is there in the tasks list"); the existing `Track` suite green with byte-identical titles; `Chatter` neutrals; frame-beats-interrogative. On the delegation side — an `Answer` turn leaves the queue unclaimed and a `spawn_task` push refused in-turn with no card after, a `Chatter` turn still delegates, and a direct-desk `Answer` opens no card. Plus a server-level test that chatting "what is there in the tasks list?" replies normally and leaves the task store untouched.

**Test-harness change worth a reviewer's eye**: `ScriptedTurns`' `Turn::queues` pushes via `DelegationQueue::push`, the escape hatch that bypasses both the cap and the #453 commitment — so a gate test written against it would have asserted nothing. `Turn::tool_pushes` / `Turn::tooling` were added to route through the real `push_within_cap`, with a `staged()` accessor.

## Documentation

`docs/spec/runtime/api.md` describes the three-way triage; `src/runtime/board_events.rs`' module doc — which named `detect_task_intent` as a card path — is updated to match. The two subtle sites (the withheld claim, the adopted-but-unsettled card) carry their reasoning inline.

**Known deviation from the repo's 500-line Markdown cap**: `docs/spec/runtime/api.md` is now 518 lines. It was already **506 on `main`** before this branch; the addition here was compressed from 28 lines to 17 but does not bring it under. Splitting the file is a separate change and is deliberately not bundled into a behavior PR — flagging it rather than letting it pass unremarked.

**A card that still lags, traced and tracked (#678).** When the orchestrator answers "create a workflow named X" by authoring the graph in-turn, the handler card is adopted but never settles: `CreateWorkflowTool` stages onto the `WorkflowRefQueue`, drained only by the dispatched-card path in `HarnessBrain::run_task`, and the publish drain settles from `pending_publishes`, which `create_workflow` never populates. So the operator does get the workflow; the card stays in To-do. It is not fixed here because `DelegationRunner` holds no `workflow_refs` handle and `TaskOutput.run_id` is required while a chat turn has no run row — task output without a run is #183's territory. Recorded at the call site in code and folded into #678 rather than left silent.

Closes #267

## Summary by CodeRabbit

* **New Features**
  * Chat messages are now classified as questions, actionable work, or casual conversation.
  * Questions receive direct answers without creating task cards.
  * Action requests continue to create task cards, while workflow copilot conversations suppress card creation.
  * Desk delegation remains available during question-answering turns, while board-only actions are appropriately restricted.

* **Bug Fixes**
  * Prevented duplicate task cards during delegated conversations.
  * Improved delegation refusal messages with clearer explanations and guidance.
  * Ensured staged delegation work is safely released when a claim ends.

### DIFF
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -435,10 +435,22 @@ scope; SSE (`/chat` streaming, the `/events` work feed) is not yet wired.
   `409` a graph that moved underneath it earns.
 
   Two more consequences worth knowing before reusing the seam. A chat turn
-  runs the **whole** company cycle, so an actionable message also opens a
-  board card via `company::task_intent` — on every thread *except* a copilot
-  one, where that is suppressed (#416) precisely because the seam is being
-  reused for a conversation that is not a request to the company. And an
+  runs the **whole** company cycle, so every message is first classified by
+  `company::task_intent::triage_message` (#267) into `Track` (an instruction —
+  the route opens a `todo` card), `Answer` (a question or read — no card), or
+  `Chatter` (greetings, and anything ambiguous — no card). `Answer` is also the
+  only class that *gates*: the harness withholds the issue-#453 delegation
+  claim for that turn, so the model's own `spawn_task` / `delegate_to_desk` /
+  `assign_task` / `review_task` fail at the tool boundary with the do-not-retry
+  refusal, while `query_company` / `run_workflow` / `read_run_output` run inline
+  and are untouched — the turn loses the ability to *write*, never the ability
+  to answer. Ambiguity falls to `Chatter`, which neither cards nor gates: a
+  missed card costs one follow-up message, a spurious card pollutes the board
+  permanently. The gate is harness-only — `HostedMedullaBrain` has no
+  delegation stack to gate (#176) — while the triage itself is compiled into
+  every build and fronts both brains. The card half is suppressed wholesale on
+  a copilot thread (#416), precisely because the seam is being reused for a
+  conversation that is not a request to the company. And an
   unconfigured company answers
   `200` with the echo brain's `"You said: …"` rather than an error, so a caller
   that needs a real answer must check `cognition` from `GET {scope}/inference`
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -439,12 +439,15 @@ scope; SSE (`/chat` streaming, the `/events` work feed) is not yet wired.
   `company::task_intent::triage_message` (#267) into `Track` (an instruction —
   the route opens a `todo` card), `Answer` (a question or read — no card), or
   `Chatter` (greetings, and anything ambiguous — no card). `Answer` is also the
-  only class that *gates*: the harness withholds the issue-#453 delegation
-  claim for that turn, so the model's own `spawn_task` / `delegate_to_desk` /
-  `assign_task` / `review_task` fail at the tool boundary with the do-not-retry
-  refusal, while `query_company` / `run_workflow` / `read_run_output` run inline
-  and are untouched — the turn loses the ability to *write*, never the ability
-  to answer. Ambiguity falls to `Chatter`, which neither cards nor gates: a
+  only class that *gates*: the harness narrows the issue-#453 delegation claim
+  to answering for that turn, so the model's own `spawn_task` / `assign_task` /
+  `review_task` fail at the tool boundary with the do-not-retry refusal.
+  `delegate_to_desk` is deliberately **not** refused — it is how a question the
+  orchestrator cannot answer alone reaches a desk that can — so it runs the
+  desk lead and relays their reply, and only its board card stands down.
+  `query_company` / `run_workflow` / `read_run_output` run inline and are
+  untouched throughout. The turn loses the ability to *write*, never the
+  ability to answer. Ambiguity falls to `Chatter`, which neither cards nor gates: a
   missed card costs one follow-up message, a spurious card pollutes the board
   permanently. The gate is harness-only — `HostedMedullaBrain` has no
   delegation stack to gate (#176) — while the triage itself is compiled into
--- a/src/company/task_intent.rs
+++ b/src/company/task_intent.rs
@@ -1,19 +1,50 @@
-//! Detect an **actionable request** in an operator chat message, so the chat
-//! handler can deterministically open a dashboard task card for it.
+//! Triage an operator chat message **before** anything is written to the board:
+//! is this a question to answer, work to track, or neither?
 //!
-//! OpenCompany's chat can already produce task cards — but only when the
-//! orchestrator model chooses to call its `spawn_task` tool, and the
-//! orchestrator brief biases toward just replying, so a "do this" ask often
-//! leaves no visible work item. This module adds the deterministic half: when an
-//! operator message is an actionable request ("build the landing page",
-//! "can you set up the newsletter"), [`detect_task_intent`] returns a cleaned
-//! task title and the handler opens a `todo` card. The model's `spawn_task`
-//! stays available for sub-tasks it wants to open on top.
+//! # Two doors, not one (issue #267)
 //!
-//! Conservative and cheap by design (no model call): it fires on imperative
-//! requests and request-framed action asks, and stays silent on pure questions
-//! ("what's our revenue?"), greetings/acknowledgements ("hi", "thanks"), and
-//! neutral chatter — so the board fills with work, not small talk.
+//! Cards reach the board through two independent paths, and a fix on one leaves
+//! the other open:
+//!
+//! 1. **Deterministic** — the REST chat handler runs [`triage_message`] over the
+//!    operator's words and opens a `todo` card on [`MessageTriage::Track`]. This
+//!    is the path that produced five of the six dead `backlog` cards observed on
+//!    a live company ("Call the composio_authorize tool …", four × "Create a
+//!    workflow …"), because every one of them leads with an action verb.
+//! 2. **The model's own** — the orchestrator calls `spawn_task` /
+//!    `delegate_to_desk` / `assign_task` / `review_task`. That is where the
+//!    sixth card came from ("Tell what is there in the tasks list" — no action
+//!    verb, no request frame, so this module never saw it as work).
+//!
+//! [`triage_message`] is **Layer A**: it lives in the REST chat handler, is
+//! compiled into every build, and fronts both cognition brains. Its answer is
+//! also what the harness delegation seam uses to gate door 2 (Layer B) — an
+//! [`MessageTriage::Answer`] turn does not claim the delegation queue, so the
+//! model's board-writing tools refuse in its own turn.
+//!
+//! # Positive triage, and a deliberate lean toward answering
+//!
+//! The classification is a **three-way** decision rather than a card/no-card
+//! boolean, because "not work" and "a question about state" need different
+//! treatment: only the latter may gate the model's board tools, and gating
+//! small talk would be both pointless and risky.
+//!
+//! Tie-breakers lean toward [`MessageTriage::Answer`] and, when even that is
+//! not clear, toward [`MessageTriage::Chatter`] — which neither cards nor
+//! gates, and is the safe middle state. Straight from the issue: *a missed card
+//! costs one follow-up message, a spurious card pollutes the board
+//! permanently.*
+//!
+//! Conservative and cheap by design (**no model call**): it fires `Track` on
+//! imperative requests and request-framed action asks, `Answer` on
+//! interrogatives and read requests ("what's our revenue?", "show me the
+//! board"), and `Chatter` on greetings, acknowledgements and everything
+//! ambiguous.
+//!
+//! [`detect_task_intent`] remains as the thin `Track`-only wrapper the card
+//! paths call, so the issue-#463 title contract with
+//! `DelegationRunner::chat_handler_card` — which has to derive byte-for-byte
+//! the same title the handler wrote — is untouched.
 
 /// Bare greeting / acknowledgement messages that must never open a card, matched
 /// against the whole message (punctuation stripped). Kept short and exact so a
@@ -63,6 +94,13 @@ const GREETINGS: &[&str] = &[
 
 /// Single-word imperative action verbs. A message whose first word is one of
 /// these is an instruction to act.
+///
+/// `list` used to be here and was removed for issue #267: a bare "list X" is a
+/// **read**, not work — "list the tasks" is answered by one `query_company`
+/// call and must never mint a card. It now leads [`READ_VERBS`] instead. The
+/// removal also narrows [`contains_action`], which is what makes "can you list
+/// the tasks?" an [`MessageTriage::Answer`] rather than a request frame around
+/// an action verb.
 const ACTION_VERBS: &[&str] = &[
     "build",
     "create",
@@ -112,7 +150,6 @@ const ACTION_VERBS: &[&str] = &[
     "upload",
     "summarize",
     "summarise",
-    "list",
     "track",
     "monitor",
     "investigate",
@@ -213,32 +250,124 @@ const LEAD_INS: &[&str] = &[
     "actually ",
 ];
 
+/// Lead words that open a question about the company's state. A message
+/// starting with one of these wants an answer, not a card (issue #267).
+///
+/// Deliberately excludes the auxiliaries that double as imperatives — `do`,
+/// `have`, `can`, `could`, `should`, `will` — because "do the newsletter" and
+/// "have the team look at it" are asks, not questions. The ones they *do* open
+/// as questions ("can you tell me the numbers?", "should we ship?") end in a
+/// `?` and are caught by the trailing-`?` rule below instead.
+const INTERROGATIVE_LEADS: &[&str] = &[
+    "what", "what's", "whats", "who", "who's", "whos", "whose", "when", "where", "why", "which",
+    "how", "how's", "hows", "is", "are", "was", "were", "anyone", "anybody", "anything",
+];
+
+/// Lead verbs that ask to be **told** something rather than to have something
+/// done. "Tell what is there in the tasks list" — the sixth dead card from
+/// issue #267 — opens with one of these.
+const READ_VERBS: &[&str] = &[
+    "tell", "show", "explain", "describe", "list", "recap", "clarify", "compare",
+];
+
+/// Multi-word read requests (checked as a prefix + word boundary).
+const READ_PHRASES: &[&str] = &["give me", "walk me through", "let me know", "remind me"];
+
 /// Max length of a generated task title.
 const TITLE_MAX: usize = 80;
 
-/// Returns a cleaned task title when `text` is an actionable request, else
-/// `None`. See the module docs for the intent.
-pub fn detect_task_intent(text: &str) -> Option<String> {
+/// What an operator chat message is, decided before anything reaches the board
+/// (issue #267).
+///
+/// Three states rather than two: `Chatter` is not "a weaker `Answer`" — it is
+/// the state in which the runtime asserts *nothing*, so it neither opens a card
+/// nor takes the model's board tools away.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MessageTriage {
+    /// A question about state, or a request to be told something. Answer it
+    /// from the read path; **no board write**, by either door.
+    Answer,
+    /// Real work. Opens a `todo` card under the carried title.
+    Track(String),
+    /// Neither — greetings, acknowledgements, and everything too ambiguous to
+    /// call. Cards nothing and gates nothing.
+    Chatter,
+}
+
+impl MessageTriage {
+    /// Whether this is [`MessageTriage::Answer`] — the one class that gates the
+    /// model's board-writing tools (issue #267, Layer B).
+    pub fn is_answer(&self) -> bool {
+        matches!(self, Self::Answer)
+    }
+
+    /// The task title when this is [`MessageTriage::Track`], else `None`.
+    pub fn title(&self) -> Option<&str> {
+        match self {
+            Self::Track(title) => Some(title),
+            _ => None,
+        }
+    }
+}
+
+/// Classifies an operator chat message as [`Answer`](MessageTriage::Answer),
+/// [`Track`](MessageTriage::Track) or [`Chatter`](MessageTriage::Chatter).
+///
+/// # The order is the design
+///
+/// 1. **Empty** → `Chatter`. Nothing was said.
+/// 2. **A bare greeting / acknowledgement** → `Chatter`.
+/// 3. **A request frame around an action verb** → `Track`. Checked *before* any
+///    question test on purpose: "can you build the landing page?" is work
+///    despite the `?`, and reading the `?` first would lose every politely
+///    phrased instruction an operator ever types.
+/// 4. **A question or read request** → `Answer`: an interrogative lead word, a
+///    [`READ_VERBS`] / [`READ_PHRASES`] lead, or a trailing `?` with no action
+///    verb anywhere.
+/// 5. **A leading imperative** → `Track`.
+/// 6. **Anything else** → `Chatter`, the safe middle.
+pub fn triage_message(text: &str) -> MessageTriage {
     let trimmed = text.trim();
     if trimmed.is_empty() {
-        return None;
+        return MessageTriage::Chatter;
     }
 
     let lower = trimmed.to_lowercase();
     // Whole-message greeting/ack check (strip trailing punctuation first).
     let bare = lower.trim_end_matches(['.', '!', '?', ' ']).trim();
     if GREETINGS.contains(&bare) {
-        return None;
+        return MessageTriage::Chatter;
     }
 
     // Strip leading filler ("ok now …", "and also …") so the ask underneath is
     // judged on its own.
     let core = strip_lead_ins(&lower);
-    if !is_actionable(core) {
-        return None;
-    }
 
-    Some(to_title(trimmed))
+    // Frame beats interrogative: a polite instruction stays work.
+    if REQUEST_FRAMES.iter().any(|f| core.starts_with(f)) && contains_action(core) {
+        return MessageTriage::Track(to_title(trimmed));
+    }
+    if is_question(core) {
+        return MessageTriage::Answer;
+    }
+    if starts_with_action(core) {
+        return MessageTriage::Track(to_title(trimmed));
+    }
+    MessageTriage::Chatter
+}
+
+/// Returns a cleaned task title when `text` is an actionable request, else
+/// `None` — the [`MessageTriage::Track`]-only view of [`triage_message`].
+///
+/// Kept as its own function because two card paths depend on it agreeing with
+/// itself byte-for-byte: the REST chat handler writes the title, and
+/// `DelegationRunner::chat_handler_card` re-derives it moments later to find
+/// the card that handler wrote (issue #463).
+pub fn detect_task_intent(text: &str) -> Option<String> {
+    match triage_message(text) {
+        MessageTriage::Track(title) => Some(title),
+        MessageTriage::Answer | MessageTriage::Chatter => None,
+    }
 }
 
 /// Remove stacked leading filler words from a lowercased message.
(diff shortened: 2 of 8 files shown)


M3gA-Mind (commented):
Reviewed manually (`gh pr diff` plus `task_intent.rs`, `delegation.rs`, `orchestrator.rs`, `operator.rs` at the PR head) rather than through our review harness — its `sync` step is openhuman-resident and I didn't want it checking out branches in a clone another agent is using. **Comment only, no approval.**

**What I covered:** the classifier in full (rule ordering, all four constant lists, `is_question` / `contains_action` / `starts_with_action`, and the new test block); Layer B in `delegation.rs` (the conditional claim, the `clear()`-without-claim, `open_direct_work_card`'s new `answering` arm, and the `ScriptedTurns` / `Turn::tooling` harness change); the four gated tool sites and `no_drain` in `orchestrator.rs`; the Layer A wiring in `run_chat`. **What I did not cover:** `src/server/ops/write_test.rs` beyond skimming, the `api.md` prose, and `board_events.rs`' five-line doc update — I read them but didn't scrutinise them, so treat those as unreviewed.

First, credit where it's due, because the framing of this PR is unusually good. Correcting the issue's premise rather than building to it, tracing which of the six cards came through which door, and listing three places the code contradicted the plan instead of smoothing them over — that's the version of a PR description that's actually worth reading. The `#678` card-lag disclosure and the `api.md` line-cap note are both things it would have been easy to stay quiet about. And in the tests, `assert!(triage_message(question).is_answer(), "fixture must triage as a question, or this proves nothing")` before the real assertion is exactly right — it's what stops the gate tests going vacuous when the classifier drifts. `Turn::tooling` routing through `push_within_cap` instead of the `queues` escape hatch is the same instinct.

Three things.

---

### 1. `Answer` is entered on a lead-word match with no action-verb veto — and `Answer` is the class with teeth

`Chatter` is the safe middle: it cards nothing and gates nothing, and the tie-breaker reasoning around it is sound. But `Answer` is not safe — it withdraws four tools and makes the model tell the operator it couldn't do the thing. So the cost of a false `Answer` is much higher than the cost of a false `Chatter`, and the entry conditions for `Answer` are looser than that asymmetry warrants.

`is_question` returns `true` on a bare first-word match against `INTERROGATIVE_LEADS` or `READ_VERBS`, or a `READ_PHRASES` prefix, with no look at the rest of the sentence. Only the fourth branch consults `contains_action`:

```rust
// A trailing `?` with no action verb anywhere. The action-verb guard is
// what keeps "fix the login bug?" out of here — an operator second-guessing
// their own instruction is still an instruction.
lower.trim_end().ends_with('?') && !contains_action(lower)
```

That reasoning — a stated instruction survives a question-shaped wrapper — applies just as well to the lead-word branches, which don't get it. Concretely, all of these are `Answer` today, so the board tools are withdrawn:

| message | why | what it actually is |
|---|---|---|
| `give me a landing page` | `READ_PHRASES` prefix `give me` | work |
| `explain the auth flow and then fix the login bug` | `READ_VERBS` lead `explain` | a read **and** work |
| `compare our pricing to competitors and write up a doc` | `READ_VERBS` lead `compare` | a read **and** work |
| `how about we build the landing page?` | `INTERROGATIVE_LEADS` lead `how` | work |

Three of those four contain an action verb (`fix`, `write`, `build`) that the trailing-`?` branch would have vetoed on.

The tests don't catch this because they pick the read-flavoured instance of each pattern: `give me the headcount`, `walk me through the funnel`, `describe the current backlog`. Every one of those is right — but each of those lead words also fronts work asks, and no case in the suite probes that side.

**I'm deliberately not suggesting a blanket `&& !contains_action(...)` on all branches**, because `ACTION_VERBS` is full of words that are commonly nouns — `review`, `call`, `post`, `design`, `plan`, `audit`, `draft`, `update`, `order`, `track`, `setup`. A blanket veto would knock out genuine questions like *"what's the status of the design review?"* (currently `Answer`, correctly). It would degrade them to `Chatter`, which is safe, but you'd lose the gate on most real questions and the layer would stop earning its keep.

The two targeted changes I'd actually make:

- **Drop `give me` from `READ_PHRASES`.** It's the odd one out: `walk me through`, `let me know`, `remind me` all mean *tell me*, but `give me` overwhelmingly means *produce me* — a draft, a page, a report. `give me the headcount` would fall to `Chatter`, which costs nothing (no card, no gate; the brief handles the read), while `give me a landing page` stops being gated.
- **Don't let a lead word speak for a conjoined imperative clause.** A cheap version: if the message splits on ` and `, `;`, or a sentence boundary and any later clause `starts_with_action`, don't return `Answer`. That covers the `explain X and then fix Y` shape without touching single-clause questions.

Even with both, some misses remain — that's inherent to a keyword classifier, and #678 is the right home for the real fix. What I'd want alongside is finding (3) below, so the residual rate is measurable rather than invisible.

### 2. `delegate_to_desk` is gated too, and that removes a way of *answering*

The design is stated as "it removes the ability to write, not the ability to reply", with `query_company` / `run_workflow` / `read_run_output` left inline as the means of answering. But `delegate_to_desk` is refused on an `Answer` turn (`orchestrator.rs:1276-1286`), and delegation isn't only a board write — it's how a question the orchestrator can't answer gets routed to someone who can. After this change, *"what did the design desk ship this week?"* triages `Answer`, and the orchestrator can no longer hand it to the design desk; it has to answer from `query_company` or not at all.

What makes this stand out is that **this same PR already solves it the better way one layer over.** `open_direct_work_card` takes the new `answering` flag and suppresses the *card* while the desk lead still runs their turn and answers — and the docstring is explicit that this is "deliberately narrower than the queue gate above: this suppresses a card, it does not take any tool away". That's the right shape. The orchestrator path gets the blunt version of the same problem.

So: is `delegate_to_desk` in the gated set on purpose, or did it come along because the gate is drawn at queue membership rather than at intent? The PR body already flags that the boundary is queue-membership under "three places the code contradicted the plan" (`create_workflow` and `add_agent` stay available because they're inline) — this looks like the same seam cutting the other way, and here it costs something. If suppressing the hand-off card while letting the delegation run is feasible, that would make the two desk paths consistent; if it isn't, I'd at least say in the module doc that consulting a desk is off the table on a question turn, because the current text reads as though answering is untouched.

### 3. `no_drain` now speaks for two very different conditions

```rust
"Refused: nothing here can carry out board work, so {effect}. Board actions are \
 unavailable in this context. Do not retry — it will fail the same way — and do NOT \
 report the action as done or describe the card as moved. Say plainly that you could not do it."
```

That was written for a genuinely inert context — no drain wired, nothing could ever work. It's now also what a perfectly capable company says when the triage read the message as a question. In that case "nothing here can carry out board work" and "board actions are unavailable in this context" are both false in the way the operator will hear them: board actions work fine, and would have worked on a differently-phrased message. The model is instructed to relay this plainly, so combined with a finding-(1) miss the operator experience is: *ask for a landing page → "I could not do it; board actions are unavailable."* — with nothing indicating that rephrasing would fix it.

A separate message for the triage case seems worth it, and could give the model something recoverable to say — that the message was read as a question, and to ask the operator to restate it as a request if they meant work.

The `tracing::warn!` inside `no_drain` has the same problem and it matters more operationally: it emits identical text for both causes, so **you cannot measure how often the triage gate fires** once this is live. That's the number I'd most want after shipping a keyword classifier with teeth — it's the difference between "the heuristic is fine" and "we're silently refusing work" being an answerable question. A `reason` field distinguishing no-claim-by-triage from no-drain-wired is a small change and would let finding (1)'s residual miss rate be observed rather than guessed at.

---

Smaller notes:

- The `clear()`-without-claim on the answering path, and the reasoning that the clear is a safety property while the claim is a permission, is right and worth the comment it got.
- `Chatter` not gating is the correct call and I'd defend it if anyone pushes back on it.
- The `api.md` 518-line overrun: agreed it shouldn't be bundled here, and you've already flagged it.

Nothing here is a blocker on its own. (1) is the one I'd want narrowed before merge — the two targeted changes are small — and (3) is worth doing at the same time so the part you can't fully fix is at least visible. (2) is a design question I'd like answered rather than a defect.

oxoxDev:
All three addressed in `4dca26b6`, `7e82efc2`, `ec2bc08f` (on top of a clean `upstream/main` merge, `b5011ee0`). Two places your prescription didn't fully reach are called out below rather than quietly left.

**(2) first, since it's the one you asked a question about — answered as "let it run".**

You were right that the #453 claim is boolean and can't express *run it but don't card it*. It can, though, express **which kinds may stage**, and carding was already a separate layer with a precedent for exactly this suppression. So:

- `DelegationQueue`'s commitment went `bool` → `DrainClaim { Unclaimed, Full, Answering }`, and `handle_operator_message` calls `claim_answering()` rather than withholding the claim.
- `Delegation::answers()` is true only for `DelegateToDesk`. `push_within_cap` refuses a non-answering delegation under an `Answering` claim, so `spawn_task` / `assign_task` / `review_task` still fail in the model's own turn.
- `open_hand_off_work_card` wraps `open_work_card` in the `DelegateToDesk` arm and returns `None` when the message triaged `Answer` — a direct copy of `open_direct_work_card`'s shape and its "suppresses a card, does not take a tool away" stance. That method is no longer the odd one out.

Net: the claim answers *may this tool queue?*, the card layer answers *does this leave a card?*, each mechanism doing one thing. The "removes the ability to write, not the ability to reply" claim is now true — consulting a desk is back on the table on a question turn.

**(1)** Both targeted changes made: `give me` dropped from `READ_PHRASES`, and a later-clause veto (` and ` / `;` / sentence boundary → any later clause that `starts_with_action` blocks `Answer`). Deliberately not blanket, for the noun-verb reason you gave.

Two honest corrections:

- **`how about we build the landing page?` is still `Answer`.** It's in your table, but neither prescribed change reaches it — no clause split, and `how` is a bare `INTERROGATIVE_LEADS` hit. It'd need a `how about` / `what if` proposal frame, which you didn't ask for and which I didn't want to add unreviewed. The other three rows are now `Chatter`, verified empirically.
- **The clause veto has a symmetric false positive** — the mirror of the blanket-veto problem you warned about, scoped smaller: `show me the design and review queue` → `Chatter`, because the later clause opens with the noun-ish `review`. It costs a gate rather than minting a card, so it fails in the cheap direction; documented in `later_clause_is_imperative` rather than worked around.

Also minor: the three vetoed rows land in `Chatter`, not `Track` (`give me a landing page` has no leading action verb), so the new test asserts `!= Answer` rather than `== Track`.

**(3)** Split by cause. The triage refusal now tells the model the message was read as a question and to invite the operator to restate it as a request if they meant work — recoverable, rather than "board actions are unavailable" from a company where they plainly are. The `tracing::warn!` carries a `reason` distinguishing no-claim-by-triage from no-drain-wired, so the gate's fire rate is countable in production. Three tests cover the two wordings and that the causes are countable apart — which, as you said, is what makes (1)'s residual observable rather than guessed at.

Gates: fmt, clippy (both plain and `--no-deps --features openhuman,mcp,telegram`), build, `cargo test` 1988 passed, `cargo test --features openhuman,mcp,telegram` 2956 passed, frontend typecheck + build clean. Nine new tests; existing cases green including byte-identical `Track` titles.

One note: `docs/spec/runtime/api.md` is now 521 lines (was 518, and 506 before this branch). Still deferring the split rather than bundling it here, but it moved in the wrong direction and I'd rather say so.

M3gA-Mind (commented):
**Re-review.** I commented on this at `3f766857`; head is now `ec2bc08f`. Reviewed the three commits after the upstream merge — `4dca26b6`, `7e82efc2`, `ec2bc08f` — via `git diff b5011ee0 ec2bc08f` (5 files, +779/−109), so the merge's own contents are out of scope here. Manual again, same reason as before. **Comment only, no approval.**

I read CodeRabbit's pass (which arrived at this head) before writing, so nothing below repeats it. Its `no_drain_wired` catch is worth acting on and I've said why under (3).

**All three things I raised are addressed, and the classifier one is addressed exactly at the seam I'd have picked.**

### 1. The lead-word narrowing is principled, not a patch

This was the one I was most prepared to find over-fitted, because "narrow a heuristic after it misfired" usually means special-casing the observed sentence. It isn't. Two changes, and they're the two general ones:

`give me` is dropped from `READ_PHRASES` — it was the odd member, since `walk me through` / `let me know` / `remind me` all mean *tell me* while `give me` overwhelmingly means *produce me*. And `later_clause_is_imperative` splits on `;.!?` plus ` and `, skips the first clause, and asks whether any later one `starts_with_action`.

The property that makes this safe is that **the guard is monotone in the safe direction**. `!conjoined_imperative &&` can only prevent a message from entering `Answer`; it can never move one in. So the change cannot create new false gating, only remove it. A message that legitimately opens with a lead word and has no imperative later clause is untouched — which is the failure direction I'd have worried about, and it's structurally excluded rather than tested-and-hoped.

Where a narrowed message lands is also right: `explain the auth flow and then fix the login bug` leaves `Answer`, fails `starts_with_action` on the whole string (`explain` isn't an action verb), and falls to `Chatter` — no card, but no gate either, so the model keeps its board tools. That's the safe middle doing its job, and `a_work_ask_behind_a_read_lead_is_not_gated` asserts `assert_ne!(…, Answer)` rather than `assert_eq!(…, Track)`, which is the honest assertion. Claiming those become tracked work would have been overreach.

`a_noun_that_doubles_as_an_action_verb_stays_a_question` is the counter-test I said this needed. It's the reason I didn't suggest a blanket `!contains_action` last time — `review`, `audit`, `update` are nouns as often as verbs — and pinning `what's the status of the design review?` as still `Answer` is what stops a later "simplification" from reintroducing that.

`a_read_flavoured_give_me_falls_to_the_safe_middle` deserves a specific mention: dropping `give me` means `give me the headcount` is no longer `Answer`, and rather than quietly accepting that, the PR pins where it now lands. That cost is the right trade (Chatter neither cards nor gates) and documenting it in a test is better than a comment.

### 2. The hand-off change is a display decision, not an audit hole — but the reasoning is worth stating

My point was that `delegate_to_desk` isn't only a board write, it's how a question the orchestrator can't answer reaches someone who can, so gating it removed a way of *answering*. The three-state `DrainClaim` (`Unclaimed` / `Full` / `Answering`) with `Delegation::answers()` admitting only `DelegateToDesk` is the narrow version of the fix — `spawn_task` / `assign_task` / `review_task` still hit `NoDrain` on a question turn, and `the_lifecycle_writes_are_refused_on_a_question_turn_too` pins that the new claim state didn't open the floodgates.

On whether a card-less hand-off is work with no record: I don't think it is, and the reason is that the card was never the audit surface. A delegated desk turn's effectful tool calls still park through the desk's own approval policy — the card is board *tracking*, not the record that something happened. So suppressing it on a question turn suppresses a dead card, which is the entire point of #267. `open_hand_off_work_card` mirrors `open_direct_work_card` exactly (the tool runs, only the card stands down), so all three card paths now treat a question identically instead of two of three.

`the_same_hand_off_on_a_non_question_still_opens_its_card` is the control that makes the pair meaningful — without it, suppression could be unconditional and both tests would still pass.

The residual, which I'd leave as-is: on a misrouted `Answer`, a desk lead runs a turn and whatever it does that *isn't* independently gated leaves no board trace. That's strictly better than the previous behaviour (the hand-off was refused outright and the operator got told it couldn't be done), and the misroute rate is what (3) now makes measurable.

### 3. The cause split is exhaustive, and the new message is the one I wanted

On the `_ => None` concern — this doesn't have that shape. The reason is decided at the `match self.claim_state()` site with one arm per `DrainClaim` variant and **no wildcard**: `Unclaimed → Unwired`, `Answering if !answers() → Triage`, `Answering | Full → proceed`. A fourth `DrainClaim` variant is a compile error there rather than a silent fall-through, which is the property `call_of` learned the hard way. `NoDrainReason::as_str` is likewise a two-arm match with no default.

The `Triage` message fixes what I actually cared about — it no longer tells a capable company that "board actions are unavailable in this context". "Board writes are held back **for this message only**" is true, it points the model at the read path *and* at the desk hand-off (which now works, because of `7e82efc2`), and it tells it to ask the operator to restate if they did mean work. That's a recoverable failure instead of a dead end.

And `reason = %reason` on the warn is the part that makes the residual miss rate observable in production rather than guessed at, which was the thing I most wanted from this one.

**Which is exactly why CodeRabbit's rename is worth taking.** `NoDrainReason::Unwired` emitting `"no_drain_wired"` reads as "a no-drain was wired" — the opposite of what it means. That label is the field the countability argument rests on, so an ambiguous value undercuts the reason the split exists. `the_two_no_drain_causes_are_countable_apart` only asserts the two strings differ, so it's a free rename.

---

I re-reviewed all three commits and found nothing wrong with them. My only outstanding suggestion is CodeRabbit's, above. Still comment-only, no approval.

oxoxDev:
Addressed in `25641b3b`, `208d24e6`. Thanks both — @coderabbitai's relay finding was right about the test and wrong about the remedy, and chasing it turned up a test that was passing for the wrong reason.

## Relay claim — behaviour kept, remedy rejected, and one real bug found in our own test

I probed it rather than reasoning about it. `Answer`-triaged message → desk hand-off → relay turn attempts `SpawnTask`, all through `Turn::tooling` (the real `push_within_cap`):

```
PROBE triage("is the build ok?") = Answer
PROBE claim_at_turn(0) = Answering
PROBE claim_at_turn(2) = Answering
PROBE staged = [Queued, NoDrain(Triage)]
PROBE cards = []
PROBE spawned_task = None
```

Scoring the three claims:

- ✅ `_claim` is live through the relay.
- ✅ **The test gap is real, and worse than described.** `Turn::queueing` pushes straight onto the queue and never touches `push_within_cap`, so both relay tests asserted nothing about the boundary. `the_relay_turn_can_no_longer_lose_a_card` used a *question* fixture (`"is the build ok?"`), so it was pinning a card-survives outcome **that cannot happen in production**. It was green for the wrong reason. Good catch — that one was ours.
- ❌ "Dropping the answering claim alone produces an `Unwired` refusal" describes the proposed variant, not current code. Current code refuses with `Triage`, which is the actionable message: it tells the model *this message* was read as a question, rather than that its context can never do board work.

**Rejecting the remedy.** A separate *full* claim scope for the relay would let a question turn mint a board card one layer later — the exact defect #267 exists to close. #442's "the relay is best placed to decide this should be tracked" doesn't survive the narrowing: the relay is a door, and seeing the answer first doesn't change who asked. Answering *"is the build ok?"* with a card nobody requested is the reported bug. Nothing is lost, either — the refusal reaches the model in its own turn with the do-not-retry wording, and a relay that genuinely must commission work has the same recourse the first turn has.

What changed instead:
- Both relay tests converted to `Turn::tooling`, and `the_relay_turn_can_no_longer_lose_a_card` re-fixtured onto a **non-question** message (`"the pricing repo needs a map"`, asserted `!is_answer()`) so it actually pins #442's guarantee through the boundary.
- New `the_relay_turns_card_is_held_back_on_a_question_turn` pinning `claim_at_turn(2) == Answering`, `staged == [Queued, NoDrain(Triage)]`, empty board, no `spawned_task`.
- `the_relay_turn_still_cannot_re_delegate` now also crosses the boundary and asserts `[Queued, Queued]` — proving the **drop**, not the triage gate, is what bounds the turn count.
- Comment in the relay block explaining the #442 × #267 interaction and why the hold-back is deliberate.

## `Unwired` rename

`"no_drain_wired"` → `"drain_unwired"`. @M3gA-Mind's framing is why this was worth taking rather than filing: that label is the field the countability argument rests on, so an ambiguous value undercuts the reason the split exists. `Triage` untouched; the doc comment records the old value and why it changed; `grep` confirms the old string appears nowhere else in-tree.

## Markdown 500-line cap → #695

Not fixing here, deliberately. `api.md` was **506 lines before this branch** — most of the overrun predates the change — and your own scan shows `docs/spec/runtime/storage.md` (575) and `docs/plans/07-workload-email-inbound-plan.md` (699) are over too. Splitting spec files inside a behaviour PR is scope creep that makes this diff harder to review, and it'd leave the other two untouched anyway.

Filed **#695** instead, listing all three files with counts, naming the `CLAUDE.md` rule as the cap's source, proposing a split per file and an optional CI check. Flagging up front: `api.md` stays at 521 here, so the markdownlint finding will re-fire on the next scan — #695 is its home.

Gates: fmt, clippy (plain and `--no-deps --features openhuman,mcp,telegram`), build, `cargo test` 1988 passed, `cargo test --features openhuman,mcp,telegram` 2957 passed.

PR #682, opened 2026-08-11
fix(workflows): enforce required node config at author time + repair broken seed workflows (#661)

oxoxDev:
## Summary

Part of #661 (workflows module audit) — findings **M1 + M2 + H2**, shipped together because they are coupled: tightening validation without fixing the seeds would make 19 shipped seed workflows fail to load at boot.

**Root cause:** config **validation** and config **derivation** only covered `agent` and `tool_call`. Every other node kind's correctness rested on hand-written `config` that nothing enforced — so graphs saved clean and misbehaved silently at runtime, including the shipped seeds.

- **M1 — enforce required config at author time.** New shared `required_config_problems(kind, label, config)` helper, wired into the on-disk `validate` loop (seed path) and into `validate_draft_against_record` (console/builder draft path): `condition` requires `config.field`, `http_request` requires `method`+`url`, `switch` requires a discriminant. Condition edges must be labeled `yes`/`no` (the `error` route of an `on_error="route"` condition is exempt). A broken workflow now fails **loud at author/load time** with an actionable message instead of silently at runtime.
- **M2 — remove the `tool_call` slug masking default.** `translate_node` defaulted an absent slug to the **node id**, which turned a missing slug into a misleading "`<node-id>` is not a wired tool" error and could silently invoke an unintended tool on an id collision. Removed; a slug-less `tool_call` is now rejected up front (M1) rather than masked.
- **H2 — repair the shipped seeds.** 19 seed TOMLs had field-less `condition` nodes (always routed `true` — the gate decision was never made); 5 labeled branches with non-`yes`/`no` words (both mutually-exclusive branches fired in parallel); `game_studio/game_build_pipeline` was a condition-guarded cycle that **could not compile** (tinyflows `IllegalCycle`). All fixed to parse → validate → translate → **compile** cleanly.

## API Or Behavior Changes

No public API change. Behavior: workflows (seed, console, builder, and REST — all funnel through the shared `validate`) with a `condition` missing a `field`, a `switch` missing a discriminant, an `http_request` missing method/url, or a `condition` with non-`yes`/`no` edge labels are now **rejected at author/load time** instead of misbehaving at runtime. `translate` no longer invents a slug from the node id. The 20 bundled seeds are corrected (see per-seed table in the thread below / commit `dad94887`).

`game_build_pipeline`: kept its QA-fail→rework loop, now bounded by `recursion_limit = 100` on the trigger (OpenCompany has no `loop` node kind; a bounded trigger is tinyflows' sanctioned mechanism for a guarded cycle).

## Tests

- [x] `cargo test` (default lib 1985/0; gated `--features openhuman,tinycortex` lib 2964/0)

The seed test (`every_bundled_workflow_is_runnable_against_its_roster`) was extended to push **every** seed through `tinyflows::compiler::compile` — it previously stopped at `translate`, which is why the `game_build` cycle shipped broken. 6 new validation unit tests added.

## Documentation

`translate.rs` doc updated for the removed default. Seed intent (which boolean gate each condition now expresses, and the 5 label normalizations) is documented in the PR thread for reviewer sanity-check.

## Related

Part of #661. Pairs with #677 (create_workflow parity, H1) and #676 (scheduler cap order, H4). Field expressions in seeds are notional template intent (`=item.<name>`) — templates carry no live upstream signal — and express the gate, matching the seeds' documented behavior.

## Summary by CodeRabbit

* **New Features**
  * Added configurable condition fields for workflow decision points.
  * Added a configuration form requiring a field for condition nodes.
  * Standardized condition branches to use “yes” and “no” labels.
  * Added a recursion limit to gameplay rework cycles.

* **Bug Fixes**
  * Workflows now validate required configuration for conditions, HTTP requests, switches, and tool calls.
  * Tool-based workflow steps now require an explicitly configured identifier.
  * Workflow compilation checks now provide clearer validation feedback.

### DIFF
--- a/companies/agentic_accounting_firm/workflows/month_end_close.toml
+++ b/companies/agentic_accounting_firm/workflows/month_end_close.toml
@@ -26,6 +26,8 @@ id = "gate"
 kind = "condition"
 name = "Filing due?"
 summary = "Branch on whether a tax filing is due."
+[node.config]
+field = "=item.filing_due"
 
 [[node]]
 id = "taxes"
--- a/companies/agentic_consultation_firm/workflows/engagement_pipeline.toml
+++ b/companies/agentic_consultation_firm/workflows/engagement_pipeline.toml
@@ -40,6 +40,8 @@ id = "gate"
 kind = "condition"
 name = "Model needed?"
 summary = "Branch on whether the case needs a model."
+[node.config]
+field = "=item.model_needed"
 
 [[node]]
 id = "model"
--- a/companies/agentic_customer_support/workflows/ticket_pipeline.toml
+++ b/companies/agentic_customer_support/workflows/ticket_pipeline.toml
@@ -27,6 +27,8 @@ id = "gate"
 kind = "condition"
 name = "Resolvable now?"
 summary = "Resolve directly or escalate."
+[node.config]
+field = "=item.resolvable_now"
 
 [[node]]
 id = "escalate"
@@ -40,6 +42,8 @@ id = "gate2"
 kind = "condition"
 name = "Bug or refund?"
 summary = "Route the escalation to the right path."
+[node.config]
+field = "=item.is_bug"
 
 [[node]]
 id = "bug"
@@ -93,12 +97,12 @@ to = "gate2"
 [[edge]]
 from = "gate2"
 to = "bug"
-label = "bug"
+label = "yes"
 
 [[edge]]
 from = "gate2"
 to = "refund"
-label = "refund"
+label = "no"
 
 [[edge]]
 from = "bug"
--- a/companies/agentic_design_studio/workflows/design_pipeline.toml
+++ b/companies/agentic_design_studio/workflows/design_pipeline.toml
@@ -40,6 +40,8 @@ id = "gate"
 kind = "condition"
 name = "Needs motion?"
 summary = "Branch on whether motion is in scope."
+[node.config]
+field = "=item.needs_motion"
 
 [[node]]
 id = "motion"
--- a/companies/agentic_enterprise_sales/workflows/sales_pipeline.toml
+++ b/companies/agentic_enterprise_sales/workflows/sales_pipeline.toml
@@ -40,6 +40,8 @@ id = "gate"
 kind = "condition"
 name = "Qualified?"
 summary = "Advance or nurture the lead."
+[node.config]
+field = "=item.qualified"
 
 [[node]]
 id = "proposal"
--- a/companies/agentic_game_business/workflows/liveops_pipeline.toml
+++ b/companies/agentic_game_business/workflows/liveops_pipeline.toml
@@ -40,6 +40,8 @@ id = "gate"
 kind = "condition"
 name = "Store update?"
 summary = "Branch on whether the store needs a refresh."
+[node.config]
+field = "=item.store_update"
 
 [[node]]
 id = "store"
--- a/companies/agentic_game_studio/workflows/game_build_pipeline.toml
+++ b/companies/agentic_game_studio/workflows/game_build_pipeline.toml
@@ -13,6 +13,12 @@ id = "concept"
 kind = "trigger"
 name = "New concept"
 summary = "A game or feature concept is greenlit."
+# The `QA passed?` gate loops a failed build back to `gameplay` for rework. A
+# guarded cycle needs an explicit bound or the engine rejects it (tinyflows
+# `IllegalCycle`): OpenCompany has no `loop` node kind, so the trigger's
+# `recursion_limit` is the sanctioned way to cap the rework passes.
+[node.config]
+recursion_limit = 100
 
 [[node]]
 id = "world"
@@ -61,6 +67,8 @@ id = "gate"
 kind = "condition"
 name = "QA passed?"
 summary = "Ship or send back for rework."
+[node.config]
+field = "=item.qa_passed"
 
 [[node]]
 id = "marketing"
@@ -106,12 +114,12 @@ to = "gate"
 [[edge]]
 from = "gate"
 to = "marketing"
-label = "pass"
+label = "yes"
 
 [[edge]]
 from = "gate"
 to = "gameplay"
-label = "fail"
+label = "no"
 
 [[edge]]
 from = "marketing"
--- a/companies/agentic_influencer_business/workflows/content_pipeline.toml
+++ b/companies/agentic_influencer_business/workflows/content_pipeline.toml
@@ -47,6 +47,8 @@ id = "gate"
 kind = "condition"
 name = "Sponsored?"
 summary = "Branch on whether this piece has a sponsor."
+[node.config]
+field = "=item.sponsored"
 
 [[node]]
 id = "sponsor"
--- a/companies/agentic_law_firm/workflows/matter_pipeline.toml
+++ b/companies/agentic_law_firm/workflows/matter_pipeline.toml
@@ -26,6 +26,8 @@ id = "gate"
 kind = "condition"
 name = "Contract or litigation?"
 summary = "Route by the kind of matter."
+[node.config]
+field = "=item.is_contract"
 
 [[node]]
 id = "draft"
@@ -65,12 +67,12 @@ to = "gate"
 [[edge]]
 from = "gate"
 to = "draft"
-label = "contract"
+label = "yes"
 
 [[edge]]
 from = "gate"
 to = "litigation"
-label = "litigation"
+label = "no"
 
 [[edge]]
 from = "draft"
--- a/companies/agentic_marketing_agency/workflows/campaign_pipeline.toml
+++ b/companies/agentic_marketing_agency/workflows/campaign_pipeline.toml
@@ -30,6 +30,8 @@ id = "gate"
 kind = "condition"
 name = "Needs research?"
 summary = "Branch on topic familiarity."
+[node.config]
+field = "=item.needs_research"
 
 [[node]]
 id = "research"
--- a/companies/agentic_media_company/workflows/newsroom_pipeline.toml
+++ b/companies/agentic_media_company/workflows/newsroom_pipeline.toml
@@ -33,6 +33,8 @@ id = "gate"
 kind = "condition"
 name = "Verified?"
 summary = "Run the story or spike it."
+[node.config]
+field = "=item.verified"
 
 [[node]]
 id = "write"
@@ -97,12 +99,12 @@ to = "gate"
 [[edge]]
 from = "gate"
 to = "write"
-label = "verified"
+label = "yes"
 
 [[edge]]
 from = "gate"
 to = "done"
-label = "spiked"
+label = "no"
 
 [[edge]]
 from = "write"
--- a/companies/agentic_pharma_startup/workflows/discovery_pipeline.toml
+++ b/companies/agentic_pharma_startup/workflows/discovery_pipeline.toml
@@ -40,6 +40,8 @@ id = "gate"
 kind = "condition"
 name = "Promising?"
 summary = "Advance to trial planning or report back."
+[node.config]
+field = "=item.promising"
 
 [[node]]
 id = "trial"
--- a/companies/agentic_realestate_company/workflows/acquisition_pipeline.toml
+++ b/companies/agentic_realestate_company/workflows/acquisition_pipeline.toml
@@ -40,6 +40,8 @@ id = "gate"
 kind = "condition"
 name = "Pencils out?"
 summary = "Advance to purchase or pass."
+[node.config]
+field = "=item.pencils_out"
 
 [[node]]
 id = "rehab"
--- a/companies/agentic_recruiting_company/workflows/hiring_pipeline.toml
+++ b/companies/agentic_recruiting_company/workflows/hiring_pipeline.toml
@@ -40,6 +40,8 @@ id = "gate"
 kind = "condition"
 name = "Qualified?"
 summary = "Advance or pass on the candidate."
+[node.config]
+field = "=item.qualified"
 
 [[node]]
 id = "interview"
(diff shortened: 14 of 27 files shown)


oxoxDev:
### Per-seed condition intent (reviewer sanity-check)

Each previously-field-less `condition` now expresses the gate below; 5 seeds also had non-`yes`/`no` branch labels normalized. Field expressions are notional template intent (`=item.<name>`) — templates carry no live upstream signal.

| Seed | condition | `config.field` | label change |
|---|---|---|---|
| accounting/month_end_close | Filing due? | `=item.filing_due` | — |
| consultation/engagement | Model needed? | `=item.model_needed` | — |
| customer_support/ticket | Resolvable now? | `=item.resolvable_now` | — |
| customer_support/ticket | Bug or refund? | `=item.is_bug` | bug→yes, refund→no |
| design_studio/design | Needs motion? | `=item.needs_motion` | — |
| enterprise_sales/sales | Qualified? | `=item.qualified` | — |
| game_business/liveops | Store update? | `=item.store_update` | — |
| **game_studio/game_build** | QA passed? | `=item.qa_passed` | pass→yes, fail→no; **+ `recursion_limit=100` on trigger** |
| influencer/content | Sponsored? | `=item.sponsored` | — |
| law_firm/matter | Contract or litigation? | `=item.is_contract` | contract→yes, litigation→no |
| marketing/campaign | Needs research? | `=item.needs_research` | — |
| media/newsroom | Verified? | `=item.verified` | verified→yes, spiked→no |
| pharma/discovery | Promising? | `=item.promising` | — |
| realestate/acquisition | Pencils out? | `=item.pencils_out` | — |
| recruiting/hiring | Qualified? | `=item.qualified` | — |
| software/feature | Security-sensitive? | `=item.security_sensitive` | — |
| venture_capital/diligence | Worth diligence? | `=item.worth_diligence` | — |
| venture_studio/venture_launch | Build it? | `=item.build_it` | — |
| signals/opportunity | Strong signal? | `=item.strong_signal` | — |
| accelerator/cohort | Admit? | `=item.admitted` | admit→yes, pass→no |

`game_build` kept its QA-fail→rework loop, now bounded (`recursion_limit=100`) since OC has no `loop` node kind — tinyflows' sanctioned bound for a guarded cycle. The 5 "multi-way" seeds are all genuinely 2-way.

oxoxDev:
Folded in the console half of the condition gap (commit `2f59f259`): `condition` is on `CREATABLE_NODE_KINDS` (a user can add it) but had no `NODE_CONFIG_FIELDS` entry, so once M1 requires `config.field` a console-authored condition would 400 with no UI to set it. Added the `field` spec (required) — the generic renderer + validator pick it up; updated the node-config unit test. Frontend gates green (vitest 412, typecheck, build).

M3gA-Mind (commented):
Reviewed manually (`gh pr diff` plus `workflow_file.rs`, `workflow_create.rs`, `translate.rs`, `workflow_scheduler.rs`, and `main`'s version of `workflow-node-config.ts`) rather than through our review harness — its `sync` step is openhuman-resident and I didn't want it checking out branches in a clone another agent is using. **Comment only, no approval.**

On CI: `UNSTABLE` here is only `Console E2E (live brain)` still pending. Everything else — Rust, `Rust (openhuman, tinycortex)` at 21m, Console, Console E2E, Desktop, mongodb — is green.

A lot of this is straightforwardly good. `required_config_problems` shared between the on-disk `validate` and the draft path is the right shape, and the seed repairs are real bug fixes: `game_build_pipeline` needed `recursion_limit` on its trigger or the rework loop would have hit `IllegalCycle` at first run, and the new `tinyflows::compiler::compile` assertion in `every_bundled_workflow_is_runnable_against_its_roster` is what caught it. Adding compile to the seed corpus test is a genuine strengthening — it moves a class of failure from first-run to CI. Removing the node-id `slug` placeholder in `translate_node` is also right; defaulting to the node id could collide with a real tool name and silently invoke something nobody chose.

But I think this ships a breaking change to already-persisted tenant data, and I don't think it's been accounted for.

---

### The new rules run on the LOAD path, and existing saved graphs can't satisfy them

`required_config_problems` and the new condition-label rule are both added inside `validate`, which `parse_workflow` runs. `parse_workflow` is not only the author-time gate — it's what every read of a saved graph goes through, including overlay bodies that live on tenant `CompanyRecord`s and were written months ago.

Two of the new rules are ones existing console-authored graphs are **guaranteed** to violate:

**`condition` requires `config.field`.** On `main`, `condition` is not in `NODE_CONFIG_FIELDS` at all — this PR's own frontend test change moves it out of the `hasConfigForm === false` list:

```diff
-    for (const kind of ["trigger", "agent", "condition", "merge", "transform", "output"]) {
+    for (const kind of ["trigger", "agent", "merge", "transform", "output"]) {
       expect(hasConfigForm(kind), kind).toBe(false);
```

So the console builder had no way to emit `config.field` on a condition node. Every condition node ever authored through the builder lacks it. There is no ambiguity here — it isn't "some tenants might", it's all of them.

**Condition branches must be labeled `yes`/`no`.** Nothing previously constrained edge labels, and the repo's own seeds used `pass`/`fail`, `bug`/`refund`. Nineteen seed files needed repair in this PR — which is the clearest possible evidence that the shapes now rejected were the ones being produced routinely. Tenant overlay bodies got the same treatment from the same builder, and a PR can't repair those; they're data.

### What that failure looks like is worse than an error

The two read paths diverge, and neither outcome is good:

- `list_workflows_union` skips a graph that fails to parse and logs `tracing::warn!(… "skipping malformed saved workflow")`. The workflow **disappears from the console list**. Per this module's own reasoning about host stdout, that warn reaches us, not the tenant.
- `load_workflow_union` returns a hard `DataInvalid`, so opening or running it by id errors.
- The cron scheduler enumerates through `list_workflows_union` (`workflow_scheduler.rs:299`). A **scheduled** workflow that trips a new rule silently drops out of `scheduled` and stops firing. No error, no journaled event, no operator-visible signal — the schedule just stops.

So the post-deploy experience for a tenant with a condition node is: workflows vanish from the console, and anything scheduled quietly stops running. A loud failure would be recoverable; silent disappearance is the failure mode we least want here.

Worth being precise about the regression direction: a condition with no `field` previously routed `true` always. That's wrong, and worth fixing. But "always takes the yes branch" is a *working* workflow with bad branching, and "absent from the console and never fires again" is strictly worse for the operator than what they have today.

### What I'd suggest

The goal — stop authoring new broken graphs — is achieved entirely by the `validate_draft_against_record` half. It's the `validate` half that reaches back over saved data. Options, roughly in order of how much I'd favour them:

1. **Split the severity by surface.** Enforce the new required-config and label rules on the author path (400 on the draft, as now), but keep them non-fatal on the load path so existing graphs still parse and still run with today's behaviour. The console can then surface "this workflow needs repair" against the specific node, which gives the tenant something actionable instead of an absence.
2. If they must be fatal at load, the graph can't just vanish: `list_workflows_union` needs to yield it in a `needs-repair` state so the console can show it, and the scheduler needs to say something louder than a host warn when a schedule stops firing for this reason.
3. A migration isn't really available for `field` — the expression was never captured, so there's nothing to recover. Edge labels are mechanically remappable (first branch → `yes`, second → `no`), but that guesses at intent and I wouldn't.

The same question applies, more mildly, to dropping the `tool_call` slug placeholder: saved graphs that relied on it change behaviour. In practice they're caught by the slug rule at load first, so they follow the path above rather than being a separate case.

I've flagged this to our side as needing a maintainer call rather than treating it as a review nit — how much pre-existing tenant workflow data exists is something you'll know and I don't. If the answer is "none that matters yet", this is a non-issue and the PR is close to fine as written. But that should be an explicit decision, because the code currently assumes it.

---

Smaller notes:

- The condition-label rule lowercases and trims before matching (`yes`/`no`), while `validate_tool_call_node` deliberately *rejects* a padded slug rather than trimming, on the grounds that the validated string and the persisted string must be the same literal. Reasonable that labels are more forgiving than slugs — the label isn't looked up as a key the same way — but the two sit close together and the asymmetry is worth a one-line note so a later reader doesn't "fix" one to match the other.
- `WorkflowNodeKind::Switch` accepts `field` *or* `expression` as the discriminant. Both seeds use `field`; is `expression` actually read downstream, or is it accepted here and ignored? If the latter, accepting it lets an author satisfy the gate with a key that does nothing.
- `validate_draft_against_record` takes only `.into_iter().next()` — the first problem — so an `http_request` missing both `method` and `url` reports one, gets fixed, and then reports the other. The load path collects all of them. Minor, but the draft path is the one with a human or a model iterating against it, so it's the one where returning both matters more.

oxoxDev:
You were right — the new rules were wired into the shared **load** validator, so they'd have hard-failed already-persisted tenant graphs on read (vanish from the console list, error on open/run, scheduled ones silently stop). Fixed by splitting severity by surface:

- `validate` now takes `strict: bool`; **only** the two new #661 rules (`required_config_problems` + the condition yes/no-label block) are gated behind `strict`. Every other structural check stays unconditional.
- `parse_workflow` calls `validate(&raw, false)` — the load/read path is lenient, so existing saved graphs parse and run with today's behavior (a field-less condition routes `true`, as before).
- Author strictness now lives entirely in `validate_draft_against_record` (console POST + agent `create_workflow` both funnel through it → 400): required-config **and** the condition yes/no-label check (added there, exempting the `error` route).
- Seed correctness is guarded by the seed corpus test asserting `validate(&raw, true)` per seed — so we still can't ship a broken seed even though load is lenient.

Also from your notes: the draft path now collects **all** required-config problems (not just the first — an `http_request` missing both `method` and `url` reports both); added the comment explaining the label-trim vs slug-no-trim asymmetry; confirmed `switch` honors `field` OR `expression` downstream (tinyflows `switch.rs:25`) — left as-is with a note.

Lenient-load/strict-reject test pairs prove it, e.g. `condition_without_field_loads_leniently_but_strict_rejects`. Gated lane 2992/0.

SHAs: `98d4ec09` (lenient load), `c3449ab9` (strict draft), `89df93e0` (seed strict test).

M3gA-Mind (commented):
**Re-review.** I commented on this at `a08bae69`; head is now `89df93e0`. Reviewed the three commits since — `98d4ec09` (lenient load), `c3449ab9` (strict draft), `89df93e0` (seeds asserted strict) — `git diff a08bae69 89df93e0`, 3 files, +356/−40. Manual again, same reason. **Comment only, no approval.**

**Stale approval worth flagging before the badges mislead anyone.** CodeRabbit's APPROVED was submitted at `2f59f259`, which is three commits behind head and predates the lenient/strict split entirely — it approved the version where the new rules ran on the load path, i.e. the thing this delta exists to change. The green row on this PR is not evidence about the current head, and nobody has approved the split.

---

**The delta resolves my finding, and resolves it the right way round.**

My concern was that the #661 rules ran inside `validate`, which `parse_workflow` runs on *every read of a saved graph* — so console-authored condition nodes (which couldn't carry `config.field`, because `main` has no config form for `condition`) would stop parsing, vanish from `list_workflows_union` behind a host-stdout warn, and silently drop out of the cron scheduler's enumeration. That's fixed at the seam I'd have picked: `validate` takes a `strict` flag, `parse_workflow` passes `false`, and the author path enforces separately.

**On the bypass question — I went looking for one and it isn't there.** The value of author-time validation is only real if nothing invalid can reach execution by another route, so I enumerated every writer rather than trusting the split:

- `create_company_workflow` → `validate_draft_against_record` ✓
- `update_company_workflow` → `validate_draft_against_record` ✓
- `rollback_company_workflow` → **no strict check** (see below)
- `set_company_workflow_enabled` / `delete_company_workflow` don't write bodies.

So both authoring routes are strict, and a graph failing the strict rules cannot be *newly* persisted.

I also chased a specific suspicion and want to record that it was wrong, since it's the shape a bypass would actually take. The condition edge-label rule lives in `validate`'s **edge** pass behind `strict`, and no non-test caller passes `strict = true` — so my first read was that the label rule was enforced nowhere on the write path, because `validate_draft_against_record` iterates nodes. It isn't: the rule is reimplemented in `validate_draft_against_record` after the node loop, with a comment that names exactly why it has to be:

> `parse_workflow` is now LENIENT on this rule (issue #682) so pre-#661 saved graphs still load, which means ALL author-time strictness for it has to live here

That's the right call and the right place to explain it. Duplicating a rule is usually a smell; here the two copies answer different questions (may this load / may this be authored) and the comment makes the split legible rather than accidental.

**The rollback caveat**, stated precisely because "nothing invalid is persisted" isn't literally true: `rollback_company_workflow` re-persists a stored version body without running the draft validator, so a rollback can bring a pre-#661 shape back onto the record. I don't think that's wrong — a rollback restores history, the body was already stored, and refusing to roll back to what you actually had would be worse. It cannot introduce a *new* invalid shape, only restore an old one, and the lenient load path means the restored graph behaves exactly as it did before. Worth one line in the comment block so the next person auditing this doesn't have to re-derive it.

**Seeds held to the stricter bar is the detail that makes the split honest.** `89df93e0` runs `validate(&raw, true)` over every bundled seed in CI, and the comment draws the line where it belongs — leniency is a concession to tenant data already on disk, not to what we ship:

> That leniency must NOT apply to what WE ship: a seed a human could no longer author from the console … has to fail CI here.

Without that, "lenient load" would have quietly become the standard for the repo's own corpus too. It also means the 19 seed repairs in the earlier commits stay repaired.

My smaller note is handled as well — `validate_draft_against_record` now reports **every** missing-config problem for a node rather than the first, so an `http_request` missing both `method` and `url` names both in one 400 instead of two round-trips.

---

**One residual, and it's a follow-up rather than a blocker.** The serious half of my finding — workflows disappearing and schedules silently stopping — is gone. What remains is the quiet half: a grandfathered graph now loads and runs with its old, silently-wrong behaviour. A `condition` with no `config.field` still always takes the true branch; the operator is never told their branch isn't branching. That is exactly the pre-#661 status quo, so it's not a regression and I wouldn't hold the PR for it.

But the strict rules now encode a precise, machine-checkable statement of which saved graphs are wrong, and nothing surfaces it. The cheap version is what I suggested last time: run the strict pass when the console reads a graph and show a "needs repair" marker against the offending node, so the grandfathering is visible to the person who can fix it rather than only to CI. That's a frontend change and doesn't belong in this PR — worth an issue if you agree.

I re-reviewed all three commits and found nothing wrong with them. Still comment-only, no approval.

PR #685, opened 2026-08-11
feat(approvals): redact approval payload and amount for non-admins (#618)

### LINKED ISSUE #618: Question: is company membership the intended boundary for approval payloads on GET …/approvals?

M3gA-Mind:
## Summary

Closes #618, implementing @oxoxDev's **option 3**: membership still decides whether you may know an approval exists; it no longer decides whether you may read its contents. `payload` and `amount_usd` are withheld from a Member and visible to an Admin.

**This is the product's first per-resource, per-role field restriction.** Everywhere else role decides whether you may *act* (`AdminScopedCompany` / `require_admin`); here it decides which fields of a read you receive. That is a new shape and worth knowing before it is copied — stated here rather than left to be inferred from the diff, as the issue asked.

## Problem

`GET …/approvals` returned every field of `ApprovalSummary` to any authenticated company member. `authorize_address` gates the route on `user.company == *id` and nothing else; `UserPrincipal` carries `role` and it was simply not consulted. So a Member could read the recipient and body of an email a teammate's agent was waiting to send, and the amount of a payment awaiting sign-off.

Option 2 (role-gate the whole route) was rejected in the issue **because it would blank #468's stalled-work indicator** — a Member has to keep seeing that work is waiting and on what. This change keeps that and removes only the two fields where "any authenticated member" was doing real work.

## Solution

A new `server::approval_visibility` applies the rule **at the edge**. `CompanyRuntime::pending_approvals()` stays principal-free — it does not know who is asking, and per-role logic stays out of the domain layer, which was the design objection recorded on the issue.

### Hidden is not absent — why there is a new field

Blanking `payload` and `amount_usd` would have failed the issue's own requirement that a redacted approval "read as *hidden by your role*, not as an empty or missing approval". `payload: None` **already means** "this effect carries no arguments", and a null `amount_usd` already means "no money involved". Without a separate signal, a withheld payment and a no-argument tool call are the same bytes, and the console cannot tell "nothing to show" from "not shown to you" — a Member would read a hidden $2,400 payment as a free action.

So `ApprovalSummary` gains `contents_hidden`, set only at the edge, skipped when false. An admin's response and every response produced before this change serialize byte-identically to before.

### Two deviations from the issue's implementation notes

Both deliberate; flagging them because a reviewer comparing this diff against those notes would otherwise read one as missed.

1. **Three consumers, not four.** The notes list `list_approvals`, `list_approvals_single`, `ApprovalGql`, and the task-card read in `ops/tasks.rs`. The task card needs no redaction: #629 (merged earlier today) shrank `TaskApproval` to `id` / `at_millis` / `status`, so neither sensitive field reaches that response any more. I verified rather than assumed — it consumes `pending_approvals()` but projects none of it.
2. **The `contents_hidden` flag**, above, which the notes did not anticipate.

### `GqlAuth::Platform` — fail-closed, explicitly

The issue warned that a platform bearer carries no `UserRole` and that defaulting it by falling through a match is how a role guard becomes a deny-list. Both arms are written out, with no wildcard, so a third arm added later is a compile error rather than a silent grant.

It is **fail-closed**: a platform bearer is the hosting control plane, not a person in the company, and has no need for a tenant's message bodies. This costs nothing today — the console authenticates as a user, and `resolve_claims` requires `platform_auth` to be configured, so a self-hosted deployment has no platform credential at all and no human loses anything. I checked that specifically before choosing, because fail-closed would have been wrong if a prosumer's only principal were `Platform`.

### Console

All three surfaces route through two shared spots in `approval-card.tsx`, so both are covered at once:

- `ApprovalPayload` renders "Details hidden by your role" instead of returning `null` — returning `null` is precisely what a no-argument approval renders as.
- The header renders "Amount hidden" when the amount is absent *and* withheld.

**Naming asymmetry, deliberate:** async-graphql camelCases the field to `contentsHidden`, while the REST summary stays `contents_hidden` (`ApprovalSummary` has no `rename_all`, and the console's TS mirror is snake_case throughout). Each surface keeps its own convention; please don't "fix" one to match the other.

## API Or Behavior Changes

- `GET /api/v1/companies/{id}/approvals` and `/api/v1/company/approvals`: for a **Member**, `payload` is omitted and `amount_usd` is `null`, with `contents_hidden: true`. Unchanged for an Admin.
- GraphQL `Company.approvals`: same rule; `ApprovalGql` gains `contentsHidden: Boolean!`. SDL snapshot regenerated via the documented `cargo test -- --ignored regenerate_sdl_snapshot`; the diff is +13 lines and contains nothing but this field and its docs.
- A platform bearer now receives redacted contents. Behaviour change for machine credentials, called out above.
- Nothing about how approvals are *decided* changes.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — run as CI runs this lane: `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, exit 0. (`--no-deps` is required locally or it fails on two pre-existing `large_enum_variant` lints in `vendor/tinyagents`; CI passes it — `ci.yml:434`.)
- [x] `cargo build --all-targets` — ran `cargo check --features openhuman,tinycortex --all-targets`, exit 0. Gated lane, covers `examples/`.
- [x] `cargo test` — ran `cargo test --features openhuman,tinycortex`: **2963 passed, 0 failed**.

Console: `npm run typecheck`, `typecheck:e2e`, `typecheck:unit` all exit 0.

**The two-account test is the one that matters.** The harness signs every request in as an admin, so a redaction verified only as an admin passes identically against no redaction at all — it would look like coverage and prove nothing. `a_member_sees_the_approval_but_not_its_payload_or_amount` seeds a second Member-role account (new `seed_fixed_member` / `member_cookie` helpers) and drives the same route as both principals: the admin sees the recipient and the amount; the member sees `id`, `kind`, `agent`, `at_millis` and `contents_hidden: true`, and the response is additionally asserted to contain the recipient string nowhere at all.

Plus four unit tests on the rule itself, including the `Platform` arm and one asserting that an approval which genuinely has no payload is **not** reported as hidden — so "nothing to show" and "not shown to you" stay distinguishable in both directions.

**Revert-and-check**, run rather than reasoned about. With `may_read_approval_contents` forced to `true` for everyone, the two-account test fails with the member's response printed in full:

```
the recipient must not reach a member: {"id":"appr-618","kind":"payment.send",
"amount_usd":2400.0,...,"payload":{"to":"board@example.test","memo":"Q3 retainer"}}
```

That failure message is the exposure this issue exists to close.

Worth recording: the two-account test caught two errors in my own work before it tested the feature — I had assumed `ApprovalSummary` was camelCase (it has no `rename_all`), and that a redacted `amount_usd` would be absent (it has no `skip_serializing_if`, so it is an explicit `null`). I asserted the shape that is actually on the wire rather than changing the wire as a side effect of a redaction.

## Documentation

Rationale lives at the definition sites, as this codebase does it: the module doc on `approval_visibility` explains why the split is per-field rather than per-route and why hidden ≠ absent; `ApprovalSummary::contents_hidden` and the TS mirror each carry the "must not render as empty" instruction.

**Not verified:** no live two-account run against a deployed console. Per the issue's note, `e2e_harness` names exactly one admin, so that would need a Member invited first. The redaction itself is proven at the HTTP boundary by the test above; what is unproven is only how the three console surfaces *look* to a real Member. Happy to do that if you want it before merge.

## Related

- Closes #618
- Adjacent and deliberately **not** in scope: `IrreversibleEffect.amount_usd` in the task-detail read also shows a money amount to any member. Different DTO, executed effects rather than parked approvals, so it is outside this issue's question. Flagging rather than widening the diff — say the word and I will file it.

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -337,6 +337,20 @@ export interface ApprovalSummary {
    */
   payload?: unknown;
   /**
+   * Whether {@link payload} and {@link amount_usd} were withheld from **this
+   * reader** because of their role (#618).
+   *
+   * Membership decides whether you may know an approval exists; role decides
+   * whether you may read its contents. An admin never sees this set.
+   *
+   * The console must not treat a withheld approval as an empty one. `payload`
+   * being absent already means "the effect carries no arguments", so without
+   * this flag a hidden payment and a no-argument tool call are the same bytes —
+   * and a member would read "nothing to show" where the truth is "not shown to
+   * you". Render the difference; see `ApprovalPayload`.
+   */
+  contents_hidden?: boolean;
+  /**
    * The chat thread this approval was raised in (#379) — the **host** thread id,
    * which is a desk id for a channel and a roster agent id for a direct message.
    * Resolve it to a console channel id with `channelIdForThread`.
--- a/frontend/src/components/approval-card.tsx
+++ b/frontend/src/components/approval-card.tsx
@@ -20,6 +20,7 @@ import {
   ChevronDown,
   ChevronUp,
   CreditCard,
+  EyeOff,
   FileSignature,
   FileText,
   Globe,
@@ -89,6 +90,15 @@ export function ApprovalHeadline({
         <p className="font-medium">{approvalAction(a)}</p>
         {a.amount_usd != null && (
           <p className="text-xs font-medium text-muted-foreground">{money(a.amount_usd)}</p>
+        )}
+        {/*
+         * #618: an absent amount normally means "this effect involves no
+         * money". When it was withheld it means the opposite could be true, and
+         * a member reading a hidden payment as a free action is exactly the
+         * misreading the flag exists to prevent.
+         */}
+        {a.amount_usd == null && a.contents_hidden && (
+          <p className="text-xs font-medium text-muted-foreground italic">Amount hidden</p>
         )}
       </div>
       {actions && <div className="flex shrink-0 gap-2">{actions}</div>}
@@ -163,6 +173,23 @@ export function ApprovalMeta({
 export function ApprovalPayload({ approval }: { approval: ApprovalSummary }) {
   const lines = useMemo(() => payloadLines(approval), [approval]);
   const [expanded, setExpanded] = useState(false);
+
+  // Withheld by role (#618) — say so. Returning `null` here would be the one
+  // wrong answer: it is what an approval with no arguments renders as, so a
+  // member would read a hidden payment as an ordinary empty card. The point of
+  // the flag is that "nothing to show" and "not shown to you" must not look
+  // alike.
+  if (approval.contents_hidden) {
+    return (
+      <div className="flex items-center gap-2 rounded-lg border border-dashed bg-muted/40 px-3 py-2 text-xs text-muted-foreground">
+        <EyeOff className="size-3.5 shrink-0" />
+        <span>
+          Details hidden by your role. An admin can see what this approval will do and decide it.
+        </span>
+      </div>
+    );
+  }
+
   if (lines.length === 0) return null;
 
   const clampable =
--- a/src/company/runtime.rs
+++ b/src/company/runtime.rs
@@ -1730,6 +1730,12 @@ impl CompanyRuntime {
                 // which matters for `composio_execute`, where the same tool is
                 // grantable reading a repository and not grantable sending mail.
                 broadly_grantable: p.effect.agent.is_some() && p.effect.may_be_granted_standing(),
+                // Always false here. Whether a *reader* may see the contents is
+                // a property of who is asking, and this projection is
+                // deliberately principal-free (issue #618) — the redaction
+                // happens at the edge, in `server::approval_visibility`, so
+                // per-role logic stays out of the domain layer.
+                contents_hidden: false,
             })
             .collect()
     }
--- a/src/runtime/types.rs
+++ b/src/runtime/types.rs
@@ -194,6 +194,27 @@ pub struct ApprovalSummary {
     /// approve-once behaviour by construction.
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub broadly_grantable: bool,
+    /// Whether [`payload`](Self::payload) and [`amount_usd`](Self::amount_usd)
+    /// were withheld from *this reader* because of their role (issue #618).
+    ///
+    /// Set at the edge by
+    /// [`approval_visibility`](crate::server::approval_visibility), never by
+    /// the projection — the runtime does not know who is asking, and this is
+    /// deliberately the only place that does.
+    ///
+    /// **The reason this flag exists rather than just blanking the fields:**
+    /// `payload: None` already means "the effect carries no arguments". Without
+    /// a separate signal, a withheld payment and a no-argument tool call are
+    /// the same bytes on the wire, and a console cannot tell "there is nothing
+    /// to show" from "you may not see it". The first renders as an ordinary
+    /// empty card; the second has to say *hidden by your role*, or a Member is
+    /// quietly misled about what they are looking at.
+    ///
+    /// Skipped when `false`, so an admin's response — and every response
+    /// produced before this field existed — serializes byte-identically to
+    /// before, and a console that reads no such field is unaffected.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub contents_hidden: bool,
 }
 
 #[cfg(test)]
--- a/src/server/approval_visibility.rs
+++ b/src/server/approval_visibility.rs
@@ -134,7 +134,10 @@ mod tests {
     #[test]
     fn a_member_gets_the_approval_without_its_contents() {
         let out = for_principal(&principal(UserRole::Member), vec![summary()]);
-        assert!(out[0].payload.is_none(), "the recipient must not reach a member");
+        assert!(
+            out[0].payload.is_none(),
+            "the recipient must not reach a member"
+        );
         assert!(out[0].amount_usd.is_none(), "nor the amount");
         assert!(
             out[0].contents_hidden,
--- a/src/server/approval_visibility.rs
+++ b/src/server/approval_visibility.rs
@@ -1,0 +1,176 @@
+//! Who may read an approval's *contents* (issue #618).
+//!
+//! Membership decides whether you may know an approval **exists**. It does not
+//! decide whether you may read what it is about. `payload` carries
+//! recipient-bearing tool arguments and `amount_usd` carries money; the rest of
+//! [`ApprovalSummary`] — the id, the kind, who asked, which task, when — is what
+//! makes stalled work visible, and every member needs it.
+//!
+//! This is the product's **first per-resource, per-role field restriction**.
+//! Everywhere else, role decides whether you may *act* (see
+//! [`AdminScopedCompany`](crate::server::ops::scope::AdminScopedCompany)); here
+//! it decides which fields of a read you receive. That is a new shape and worth
+//! knowing before it is copied.
+//!
+//! ## Why the split lands here and not on the whole route
+//!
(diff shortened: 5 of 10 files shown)


oxoxDev (approved):
Checked the enumeration rather than trusting it, because a redaction change fails at the surface nobody listed, and it holds up:

- **`ops/tasks.rs` genuinely needs nothing.** `TaskApproval` is exactly `{ id, at_millis, status }` and the construction at :1272 projects only those three from the `ApprovalSummary` it filters. Your "three consumers, not four" is correct, and verifying it instead of assuming was the right call.
- **The journal is not a way around it.** `CompanyEvent::ApprovalParked` carries `approval_id`, `effect_kind` and `thread` — no payload, no amount — so a member reading the feed or the SSE projection learns that something parked and what kind, which is the half #468 needs, and nothing more.
- **There is no `reason` field on `ApprovalSummary`** to leak an amount back in prose, and `payload` was already bounded and credential-redacted host-side by `display_payload`.

`contents_hidden` is the right call and the reasoning for it is the best part of the PR: `payload: None` already means "this effect carries no arguments" and a null `amount_usd` already means "no money involved", so blanking alone would render a withheld $2,400 payment identically to a free action. Skipping the field when false keeps every existing response byte-identical. Writing both `GqlAuth` arms out with no wildcard — so a third arm is a compile error rather than a silent grant — is exactly right for a role guard, and checking that no human principal is `Platform`-only before choosing fail-closed is the step that makes it safe rather than merely strict.

### Minor — `may_grant_standing` is derived from the arguments you are withholding

It is not redacted, and its own doc explains why that matters: *"per-card, not per-tool: the same `composio_execute` shows the control when it is listing a repository's pull requests and not when it is sending mail, because the effect carries the action that was called."*

So for a member who cannot read `payload`, that flag still discriminates on the action inside it. It is one bit, and `kind` already tells them a good deal, so I do not think it defeats the feature — but this PR is explicitly the first per-resource, per-role field restriction and the precedent others will copy. Worth one sentence recording that the remaining fields were considered and that an argument-derived boolean was judged acceptable, rather than leaving the next person to work out whether it was a decision or an oversight. Same question applies to `task` and `thread`, which locate the withheld effect in a conversation.

### Question — is `effect_kind` granularity intended at `payment.send`?

A member sees the dotted kind on both the summary and the parked event. `payment.send` plus `agent` plus `at_millis` tells them a named teammate's agent is waiting to move money and when — everything except how much. I read that as deliberate, since #468's stalled-work indicator is worthless without it, and the issue's option 2 was rejected precisely to keep it. Confirming rather than challenging: if that is the intent, it is worth stating in the module docs of `approval_visibility`, since "what a member may still infer" is the part of a redaction rule that gets re-litigated later.

**Before merging, not a finding:** every Rust lane is still pending, and the SDL snapshot regeneration means the `Console` and GraphQL lanes are the ones that matter here. Approving on the code — don't merge while it is `UNSTABLE`.

M3gA-Mind:
@coderabbitai review

(The check row is green but reads "Review rate limited", which means zero review rather than a clean one — re-requesting so this gets an actual pass. This PR adds the product's first per-resource, per-role field restriction, so it is worth a real look.)

PR #687, opened 2026-08-11
feat(workspace): let an agent tidy its own folder (#671)

### LINKED ISSUE #671: workspace: an agent can create files but cannot delete, rename or move them
## Problem

An agent has five workspace tools — `workspace_list`, `workspace_read`, `workspace_write`, `workspace_create`, `workspace_search` — and **no way to delete, rename, or move anything**. Node lifecycle is operator-only (`PATCH`/`DELETE` on `…/workspace/{node_id}`), and the agent brief says so explicitly: creating is allowed, *"NOT for renaming or deleting (operator-only)"*.

That was a coherent position when agents could only overwrite existing notes. It stopped being one when #551 let them create, and it is now actively backwards: **an agent can make a mess it is forbidden from cleaning up.**

Concretely, an agent that:

- writes `draft-v1.md`, then `draft-v2.md`, then `final.md` leaves two stale drafts it cannot remove;
- misnames a file — `notes.md` where it meant `2026-q3-notes.md` — cannot rename it, and creating the correctly-named one leaves the wrong one behind;
- publishes into the wrong subfolder cannot move it;
- retries a task and re-creates the same artifact under a slightly different name each time, accumulating near-duplicates.

Only an operator can tidy any of it, which means in practice nobody does.

**#607 made this compound.** Before search, junk was invisible — you had to open a folder to see it. Now every stale draft is a candidate in every future search, competing for a bounded result set (`MAX_SEARCH_RESULTS = 50`) against the file someone actually wants. The tree degrades exactly as it gets used, which is the same shape of problem search was built to fix.

## Proposal

Give an agent lifecycle control over **its own folder** — `Agents/<its own id>/` — and nothing else.

- A `workspace_delete` and a `workspace_rename_move` (names to be settled), refusing any node outside the calling agent's own home with a clear message.
- Both ride the existing **`workspace.write`** grant. An agent that cannot write has no business deleting; an agent that can write can already destroy a note's contents by overwriting it, so deletion is not a new capability tier — it is the same power with a tidier result.
- Deleting a folder deletes its contents, as the store already does. Whether an agent may delete a non-empty folder is worth deciding explicitly rather than inheriting.
- The brief updates: the current sentence tells agents renaming and deleting are operator-only, and that will be false.

### On scoping it to the agent's own home

**This is not a security boundary and should not be described as one.** Agents write unconfined across the whole tree by deliberate decision, so an agent that wanted to destroy another's work can already do it with `workspace_write`. The scope is about coherence, not containment: *"tidy up after yourself"* is a capability an agent has an obvious reason to want and an obvious way to use correctly. *"Delete anything in the company"* is not — no task requires it, and a prompt-injected agent handed it does more damage per mistake.

If a case later appears for cross-tree lifecycle (an operator asking an agent to reorganise the whole workspace), that is a separate decision with its own justification, not something to grant pre-emptively.

### Open questions the plan should settle

1. **Does deletion need to be recoverable?** A published note is mirrored from an artifact whose version history is authoritative (#552), so deleting the *node* does not destroy the content — `materialize` would re-adopt it on the next publish. A note an agent created directly has no such backstop: delete is final. Those two cases behave very differently and the tool should probably say which one it is doing.
2. **What happens to a deleted node's artifact link?** `ArtifactVersion.workspace_node_id` points at it. A dangling link is the exact defect #553 removed elsewhere (a `sha256` with no bytes), so this needs an answer, not a silence.
3. **Should renaming be restricted at all?** Renaming is the safest of the three and the most useful; it may deserve a wider scope than deletion.

## Alternatives

**Leave it operator-only.** Consistent with today, and the reason it has survived. But it makes the tree monotonically accumulate, and it puts a human in the loop for the least interesting possible task.

**Let agents delete anywhere.** Simpler, and matches the unconfined-write decision. Rejected above: writes at least leave authorship behind, while a deletion leaves nothing to attribute.

**Auto-expire agent scratch files.** Solves accumulation without a new tool, but guesses at intent — a "draft" an agent kept deliberately is indistinguishable from one it abandoned.

## Acceptance criteria

- [ ] An agent can delete and rename/move nodes inside `Agents/<its own id>/`.
- [ ] Any target outside that home is refused with a message naming the boundary.
- [ ] Both tools require the `workspace.write` grant; a read-only agent gets neither. Pinned by grant tests in the shape of the existing ones.
- [ ] The persona brief no longer claims renaming and deleting are operator-only.
- [ ] Deleting a node that an artifact version links to leaves no dangling reference (per the open question above).
- [ ] Operator lifecycle routes are unchanged.

## Notes and scope

Small-to-medium. The store already implements `delete` and `rename_move`; this is a tool surface, a scope check, grant wiring, a `consequence.rs` declaration, and the brief.

## Related

- #551 — agents can create in the shared tree, which is what made this asymmetric
- #326 — authorship, the accountability mechanism that stands in for a write boundary
- #552 — publish mirroring; the reason deleting a *published* note differs from deleting an agent-created one
- #607 — search; stale files now compete for a bounded result set, so accumulation has a cost it did not have before

oxoxDev:
## Summary

An agent had five workspace tools — list, read, write, create, search — and **no way to delete, rename or move anything**. Lifecycle was operator-only, and the brief said so outright. That was coherent while agents could only overwrite existing notes; it stopped being coherent when #551 let them create. An agent can make a mess it is forbidden from cleaning up: stale drafts, misnamed files, near-duplicates from task retries. Only an operator can tidy any of it, so in practice nobody does.

**#607 gave that a cost.** Before search, junk was invisible until you opened a folder. Now every stale draft is a candidate in every search, competing for a bounded 50-result set against the file someone actually wants — so the tree degrades exactly as it gets used.

This adds `workspace_delete` and `workspace_rename`, confined to `Agents/<agent id>/`.

Closes #671.

## API Or Behavior Changes

**Both tools ride the existing `workspace.write` grant — no fifth grant name.** An agent that cannot write has no business deleting; an agent that can write can already destroy a note's contents by overwriting it. Own-home lifecycle is *narrower* reach on the same power.

**This is coherence, not containment, and the PR does not claim otherwise.** Agents write unconfined across the whole tree by deliberate product decision — authorship (`created_by`/`updated_by`, #326) is the accountability mechanism, not a boundary. Anything an agent could destroy by deleting, it can already destroy by overwriting. The own-home scope exists because *"tidy up after yourself"* is a capability with an obvious correct use, while *"delete anything in the company"* is not — no task needs it, and a prompt-injected agent handed it does more damage per mistake.

**One honest caveat**: delete removes the node **and its authorship record**, which is marginally worse forensically than overwrite-to-empty. Mitigated by the scope, by per-call approval (both tools are `Reach::Consequence`, so every call parks), and by the removal announcement. The CAS check is check-then-act — it narrows the race, it does not close it.

**Scope is enforced on the resolved node, not the argument.** Both tools accept `path` or `id`; the gate runs after resolution, so an `id` naming a node outside the home refuses identically to a path. Without that, `id` is a trivial walk-around.

**`workspace_delete`** requires `expected_updated_at` (the same stale-refusal shape as `workspace_write`). Three refusals, each with its own message: outside the agent's home; **the home folder itself** — `ensure_agent_folder` resolves the home *by name*, so deleting it forks where the company looks for that agent's work; and **non-empty folders**. Refusing recursion is deliberate: it turns a subtree wipe into N deliberate, individually-announced, individually-parked deletions, and dissolves the "teammate's published deliverable sits under my home" case that #552's `Agents/<agent>/<task-id>/` layout creates.

**Deleting a published note and deleting a private one are different acts, so the success message differs.** A published note is mirrored from an artifact whose version history is authoritative — the content survives and re-publishing recreates the path. A note the agent created directly has no backstop: the deletion is final, and the message says so. (My first instinct was to *refuse* on published nodes; that inverts the risk — the published one is the recoverable one.)

**`workspace_rename`** takes no CAS token, and the reasoning changed during implementation — see below. Both endpoints must be under `Agents/<self>/`: in-place rename allowed, reparent only into the home or inside it, **move-to-root always refused**.

**The dangling `ArtifactVersion.workspace_node_id` is left alone, deliberately.** The operator DELETE route produces this exact state today, and every reader re-verifies before use — `materialize` read-guards, `artifacts.rs` re-checks, `brain.rs` re-adopts by path. It is a checked tombstone, not the #553 "sha256 with no bytes" class.

## Four places the plan was wrong

Recorded because two of them would have shipped as defects:

1. **`rename_move` restamps `updated_at_millis`.** The plan justified skipping the CAS token by calling rename "content- and origin-preserving". Origins yes — the revision no; `FsOps::rename_move` sets `now_millis()`. The no-token decision still stands (nothing is destroyed), but the success message now hands the new `rev` back. Without that, an agent's very next write in the same turn is refused as stale — an intermittent, baffling failure.

2. **"No frontend change is expected" was wrong, and the fix predates this issue.** True of the tree UI. But both tools park on every call, and `frontend/src/lib/language.ts`'s `TOOL_LABELS` had no entry, so the approval card and the Standing-permissions list would render the raw tool identifier through the generic fallback — exactly the #372 bug. Found while grepping to *confirm* the plan's claim rather than to check it. **`workspace_create` has been in that broken state since #551**, so every approval prompt for an agent creating a note has shown a raw identifier; fixed here rather than left as the one unnamed sibling.

3. **Doc drift the plan did not list.** `docs/spec/runtime/manifest.md` asserted "Renaming and deleting stay operator-only" — false as of this branch. Three further spec files enumerate the mutation set. All updated, all still under the 500-line cap.

4. **"Nothing becomes `pub`" holds for production items only.** A child module reaches its ancestors' private items, so `CompanyWorkspace` / `PathIndex` / `echo_path` needed nothing — but the parent's `#[cfg(test)] mod tests` is *not* an ancestor of `lifecycle::test`, so six test helpers took `pub(super)`. Test-only, zero production surface.

Also: **move-to-root is not expressible as a path.** `""` and `"/"` fail `split_logical_path` as a *parse* error, which would have buried the real reason behind a syntax complaint. It is detected before parsing and given the scope answer; the port is never handed `Some(None)`.

## Tests

- [x] `cargo test --locked` — **1979 passed**, 0 failed
- [x] `cargo test --locked --features openhuman,tinycortex --tests` — **2987 passed**, 0 failed
- [x] `frontend`: `npm run typecheck` 0, `npm run build` 0, **`./scripts/ci/assert-design-tokens.sh` 0**

28 new lifecycle tests (99 in `workspace_tools` total). Coverage includes: scope refusals including the **by-id bypass** and a teammate's home; the home-folder refusal; non-empty folder refused **and nothing deleted**; CAS required / stale / string-form; duplicate-target rename refused; move-to-root refused; reparent into and out of the home; binary node delete and rename (pinning the port's post-#553 behaviour); published-note vs permanent-note messages; tenancy mirrors. Grant tests cover read / explicit / `workspace.read` / `workspace.write` / bare `*` / unwired.

## Documentation

Brief write-half rewritten — tidying your own folder is expected upkeep, delete is permanent for direct-created notes, and rename/delete **outside** your folder stays the operator's. The phrasing is pinned by a test, as #570 and #607 both did, because the brief is what actually steers behaviour. Module docs, the `workspace_write` description, and four spec files updated.

## Related

- #551 — agents create in the shared tree; the change that made this asymmetric, and the source of the `workspace_create` label bug fixed here
- #326 — authorship, the accountability mechanism that stands in for a write boundary
- #552 — publish mirroring; why deleting a published note differs from deleting a private one, and why `Agents/<agent>/<task-id>/` makes recursive delete sharp
- #607 — search; stale files now compete for a bounded result set, which is what gave accumulation a cost
- #372 — the raw-identifier-on-the-approval-card bug this re-fixes for `workspace_create`

### DIFF
--- a/docs/spec/company-brain/approvals.md
+++ b/docs/spec/company-brain/approvals.md
@@ -228,7 +228,8 @@ So the two questions are now separate answers from one declaration
   that can execute arbitrary code (`shell`), reach an arbitrary address
   (`http_request`, `curl`, `web_fetch`, `git_operations`), act through a
   third-party server (`mcp_call_tool`), change the company's shared note tree
-  (`workspace_write`, `workspace_create`), or run a saved workflow; refused for
+  (`workspace_write`, `workspace_create`, `workspace_rename`,
+  `workspace_delete`), or run a saved workflow; refused for
   every named
   consequence — Spend, Send, Sign, Publish, Hire, Identity; and refused for any
   tool nobody has declared.
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -211,7 +211,13 @@ which is fixed at agent-build time and never taken from tool arguments. Agents
 reach the same tree through `workspace_list` / `workspace_search` /
 `workspace_read` / `workspace_create` / `workspace_write`, and a created note has its default home
 in the reserved `Agents/<agent-id>/` folder (#551) — a convention the persona
-brief steers toward, not a boundary the routes enforce. Boot scaffolds the
+brief steers toward, not a boundary the routes enforce. `workspace_rename` and
+`workspace_delete` (#671) are the exception: those two *are* bounded to
+`Agents/<agent-id>/`, checked on the resolved node so an `id` argument refuses
+exactly as its path would. Neither restamps authorship; a delete leaves any
+artifact version that pointed at the node with a dangling `workspaceNodeId`,
+which is the same state the `DELETE` route above produces and is read-guarded
+before reuse. Boot scaffolds the
 `Agents/` root empty; an individual `Agents/<agent-id>/` is minted the first
 time that agent writes into it, and the `Desks/` root is minted whole the first
 time a desk produces something (#645) — so a tree read on a fresh company shows
--- a/docs/spec/runtime/manifest.md
+++ b/docs/spec/runtime/manifest.md
@@ -263,20 +263,23 @@ prompt = "Weekly review and operator digest"
     The Usage view surfaces a `Web searches` KPI plus a search status row
     (active / paused at cap 0 / awaiting credential / not granted / not in this
     build).
-  - **`workspace`** (issues #237, #551) grants the company's shared note tree —
+  - **`workspace`** (issues #237, #551, #671) grants the company's shared note tree —
     the `Standards/` / `Playbooks/` / `Product/` documents seeded from
     `companies/<name>/workspace/**`, plus whatever the operator and the agents
     have written since. It is **split**, unlike every other namespace: *reads*
     (`workspace_list`, `workspace_search`, `workspace_read`) follow the
     ordinary rule, so a
     catch-all `*` confers them; *mutations* (`workspace_write`,
-    `workspace_create`) need an **explicit** `workspace` or `workspace.write`
+    `workspace_create`, `workspace_rename`, `workspace_delete`) need an
+    **explicit** `workspace` or `workspace.write`
     entry in `[tools].allow`, because they change a tree every other agent then
     treats as the company's source of truth. `workspace.read` is therefore a
-    genuinely read-only grant. Create and write ride the one flag on purpose:
+    genuinely read-only grant. All four ride the one flag on purpose:
     overwriting an existing standard is strictly more destructive than adding a
-    note beside it, so a grant permitting the first has already permitted the
-    second. `workspace_search` (issue #607) is a read and rides the read side of
+    note beside it, and strictly more destructive than removing or moving
+    something inside the agent's own folder, so a grant permitting the first has
+    already permitted the rest. Issue #671 deliberately added no fifth grant
+    name. `workspace_search` (issue #607) is a read and rides the read side of
     that split — **not** the metered `search` namespace, despite the name.
     `search` is the paid external-credential grant that carries `web_search`;
     reading the company's own notes must not require a billed credential, and
@@ -293,7 +296,20 @@ prompt = "Weekly review and operator digest"
     provisioned at boot. Since issue #552 this call is one of two paths that
     bring it into existence; publishing a deliverable
     (`artifact_mirror::materialize`) is the other, and both go through the same
-    `ensure_agent_folder` seam. Renaming and deleting stay operator-only.
+    `ensure_agent_folder` seam.
+
+    `workspace_rename` and `workspace_delete` (issue #671) are the tidying half.
+    Both act on **one node at a time** and both reach only
+    `Agents/<agent-id>/` — the agent's own folder, never the folder itself,
+    never a teammate's, never shared guidance. `workspace_delete` carries the
+    same required `expected_updated_at` token as `workspace_write` and refuses a
+    folder that still holds anything, so a subtree is removed as N deliberate,
+    individually-parked calls rather than one. `workspace_rename` carries no
+    token, because it destroys nothing: body, id and both authorship stamps
+    survive it. Read the confinement as a division of labour rather than as a
+    security boundary — the same grant already confers *unconfined* overwrite,
+    which reaches further than own-folder lifecycle does. Renaming or deleting
+    anything elsewhere in the tree stays operator-only.
 
     Agent writes are **unconfined**: an agent may create or edit anywhere in its
     company's tree. Confining creation while leaving overwrite free would
--- a/docs/spec/runtime/ports-console.md
+++ b/docs/spec/runtime/ports-console.md
@@ -128,7 +128,8 @@ destroying the diff.
 The Obsidian-style note tree (`src/ports/workspace.rs`), seeded from the
 company's `workspace/**` on first use and thereafter written by both the
 operator (console/REST) and the company's agents (`workspace_create` /
-`workspace_write`).
+`workspace_write`, plus `workspace_rename` / `workspace_delete` within
+`Agents/<agent-id>/` since issue #671).
 
 ```rust
 pub trait WorkspaceStore: Send + Sync {
--- a/frontend/src/lib/language.ts
+++ b/frontend/src/lib/language.ts
@@ -153,6 +153,15 @@ const TOOL_LABELS: Readonly<Record<string, string>> = {
   workspace_write: "Edit a note in its workspace",
   workspace_read: "Read a note in its workspace",
   workspace_list: "List its workspace notes",
+  // Every workspace mutation parks per call, so each of these reaches an
+  // operator on the approval card and again in the Standing permissions list.
+  // `workspace_create` has been missing since issue #551 and was showing as the
+  // generic "Use one of its tools"; the lifecycle pair (#671) would have
+  // arrived with the same gap. "its own folder" is load-bearing on the last
+  // two — the scope is the whole reason an operator can wave them through.
+  workspace_create: "Add a note to its workspace",
+  workspace_rename: "Rename a note in its own folder",
+  workspace_delete: "Delete a note from its own folder",
   memory_store: "Save something to its memory",
   memory_recall: "Look something up in its memory",
   web_fetch: "Fetch a web page",
--- a/src/harness/build.rs
+++ b/src/harness/build.rs
@@ -35,12 +35,14 @@
 //!   the metered `search` one: it reads exactly what `workspace_read` already
 //!   grants, so requiring a billed backend credential for it would price the
 //!   cheap discovery path above the list-then-read crawl it replaces.
-//!   `workspace_write` is added
-//!   only under an **explicit** `workspace` / `workspace.write` grant — a bare
-//!   `*` does not confer it — and is guarded by a required compare-and-swap
-//!   revision token. Unlike the file tools these are scoped by the store, not
-//!   the filesystem: every call resolves through one company-scoped `tree()`
-//!   read, so no host path is ever built from agent input.
+//!   `workspace_create` / `workspace_write` and, since issue #671,
+//!   `workspace_rename` / `workspace_delete` are added only under an
+//!   **explicit** `workspace` / `workspace.write` grant — a bare `*` does not
+//!   confer them. `workspace_write` and `workspace_delete` are each guarded by
+//!   a required compare-and-swap revision token, and the lifecycle pair reaches
+//!   only `Agents/<agent id>/`. Unlike the file tools these are scoped by the
+//!   store, not the filesystem: every call resolves through one company-scoped
+//!   `tree()` read, so no host path is ever built from agent input.
 //! * **Delegation is orchestrator-only.** `query_company` / `spawn_task` /
 //!   `delegate_to_desk` (and the other orchestrator roster/workflow tools) are
 //!   wired only when `is_orchestrator`; a **dispatched** desk/roster agent never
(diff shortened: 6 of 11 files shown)


M3gA-Mind (commented):
Reviewed manually (`gh pr diff` plus `lifecycle.rs`, `workspace_tools.rs`, `consequence.rs`, `policy.rs`, `ports/workspace.rs` and `artifact_mirror.rs` at head `1b3dd2d3`) rather than through our review harness — its `sync` step is openhuman-resident and I didn't want it checking out branches in a clone another agent is using. **Comment only, no approval.**

**Coverage, since 1947 lines is enough to make a shallow pass look thorough.** I read closely: the declaration table changes and the two tests they land in; the whole of `lifecycle.rs` except its test module; `PathIndex` and the containment predicates in `workspace_tools.rs`; the `policy.rs` diff; `language.ts`; the `delete` / `rename_move` port contracts. **I did not read** `lifecycle/test.rs` (910 lines) beyond spot-checking which cases exist, and I did not verify the tests fail with the fix reverted. I skimmed rather than verified `build.rs`'s 106-line brief rewrite and the four spec files. I ran nothing. Treat those as unreviewed.

On CI: `UNSTABLE` is `Console E2E` and `Rust (openhuman, tinycortex)` still pending, nothing failing. Worth noting the CodeRabbit row says **"Review rate limited"** — so there is no automated pass on this PR either, and this comment is the only review it has.

---

This is careful work, and two things in particular are worth naming because they're the parts that usually go wrong.

**The declaration is not optimistic** — which is the specific thing #459 was about, so I checked it against behaviour rather than against the names. `d()` hardcodes `standing: Standing::PerCall`, so `Grantable` requires an explicit struct literal and neither new tool has one; both were added to `arbitrary_code_addresses_and_operator_guidance_are_never_grantable` and to the tools-that-must-still-park list. `Reach::Consequence` is right for both, and the reasoning in the comment — that a narrower scope is not a cheaper reach, because reach is about what the call costs the company — is the correct way round. Nothing here is priced by what the tool is called.

**The containment holds where it's claimed.** I tried to break it and couldn't, because it's built the right way: both scope checks run against the **resolved** node's rendered path from the index, never the caller's argument, so passing an `id` that names a teammate's note refuses identically to passing its path. The rename destination is looked up in `index.by_path` and then checked on `parent.path`, so a `..`-shaped or absolute `new_parent` can't resolve to anything that isn't already a real node in the tree; `is_legal_segment` rejects `.`/`..`/`/`/`\` on `new_name`; the root is expressible and explicitly refused. There's no symlink vector because this is a logical node tree, not a filesystem. And the module doc's refusal to call any of it a security boundary — "coherence, not containment", on the grounds that the same `workspace` grant already confers a `workspace_write` that can empty any note in the tree — is honest in a way that makes the rest easier to trust.

With that said, one finding, and it's in the property the module doc says it protects.

---

### 1. A folder delete can still be recursive, via nodes the index deliberately hides

The module doc states the invariant plainly:

> **Recursion is refused, not implemented.** The port deletes folders recursively, and exposing that would let one agent call remove a subtree with a single approval card naming a single path. Refusing a non-empty folder turns it into N deliberate calls, each announced and each parked.

The emptiness check that enforces it counts rendered paths:

```rust
let prefix = format!("{}/", entry.path);
let held: usize = index
    .by_path
    .iter()
    .filter(|(path, _)| path.starts_with(&prefix))
    .map(|(_, entries)| entries.len())
    .sum();
```

But `PathIndex` deliberately excludes a class of node from `by_path` **and** `by_id`, keeping only a tally:

```rust
None => index.unaddressable += 1,
```

with its own docstring explaining why, and — the part that matters here — where such nodes come from:

> Nodes omitted from the index because they are not addressable by path: a dangling/cyclic ancestor chain, or a name carrying a path separator. […] The `fs` backend rejects such names at creation (`reject_unsafe_name`), **but the sqlite and mongodb backends do not**, so the tool layer stays closed against them regardless of which backend is wired.

So on two of the three backends, a folder inside the agent's own home holding *only* unaddressable children has `held == 0`, passes the "still holds N node(s)" gate, and reaches:

```rust
self.workspace.store.delete(&self.workspace.company, &entry.node.id)
```

whose port contract is `Deletes a node; folders are removed recursively.`

The result is the exact outcome the doc says is refused — one call, one approval card, one path named, an unbounded subtree removed — and it is worse than a plain miscount, because the nodes that go are precisely the ones the tool layer made unreachable so that no tool could touch them. They are never counted, never announced, never named on a card. The `workspace_list` header would have told the operator `N node(s) have no valid path and were omitted entirely`; the delete card says nothing.

I'd treat this as the one blocking item. Two shapes of fix, cheapest first:

- **Refuse a folder delete while `index.unaddressable > 0`.** Conservative and about three lines. It over-refuses (the hidden node may be elsewhere in the tree), but the condition is rare, the message can point at the console, and it cannot be wrong in the dangerous direction.
- **Make the subtree test structural rather than textual.** The emptiness question is "does any node have this one as an ancestor", which is a walk over parent ids and is answerable for unaddressable nodes too — `PathIndex::build` already has `by_id_raw` in hand and currently discards what it knows about them. That's the version that actually closes it.

Worth adding a test either way: a node with a `/` in its name parented to a folder in the agent's home, then a delete of that folder, asserting refusal. That case is representable through the sqlite/mongodb backends by construction, which is what makes it worth pinning.

### 2. Six other per-call tools look like they still hit the generic card label

Not this PR's bug, but this PR is what surfaced the class, and the comment you wrote for it makes the case better than I would:

> `workspace_create` has been missing since issue #551 and was showing as the generic "Use one of its tools"; the lifecycle pair (#671) would have arrived with the same gap.

I cross-checked the whole declaration table against the label maps. Six tools are `Reach::Consequence` — so they park per call and reach an operator on a card — and appear in neither `EFFECT_LABELS` nor `TOOL_LABELS`:

```
curl   git_operations   http_request   mcp_call_tool   publish_artifact   run_workflow
```

`approvalAction` falls through those two maps to `"Use one of its tools"`, so on the face of it an operator approving a `publish_artifact` — externally visible and not reversible by the company alone — is being asked to reason about the same generic string `workspace_create` was showing.

**Caveat I could not resolve from the frontend alone:** `approvalAction` keys on `a.kind`, and some of these may park under an *effect kind* rather than a tool name — `external.publish` is already in `EFFECT_LABELS` and would cover `publish_artifact` if that's the path a publish takes. So this may be six gaps, or fewer. You're better placed to say which of the six actually park under their tool name. If any do, they're the same one-line addition you just made.

### 3. `artifact_mirror` (#683) — flagging for merge order, not verified

I could not test this: #683's delete-then-rename staging is not in this branch, so there is nothing here to run it against. Naming the specific risk so it isn't discovered at merge.

#683's safety argument is that a shape-changing publish stages under a name nothing resolves, so the publish path resolves to exactly one node at every instant. #687 gives agents `rename`/`move` over `Agents/<self>/`. Those two overlap if a publish ever materializes inside the publishing agent's own folder — and if it does, an agent rename landing between #683's stage and its swap is exactly the interleaving that invariant excludes. `workspace_rename` takes no compare-and-swap token (deliberately, and the reasoning given is sound in isolation — a rename destroys nothing), so there's nothing on this side that would notice.

Whichever of the two lands second is the one that should carry a test for the interleaving.

---

Smaller notes:

- `published_record_for_node` does `artifacts.list(company, None)` and scans for a match, so every note delete is a full artifact-table scan. Deletes are per-call parked so the volume is bounded by operator patience — fine as is, just noting it's a scan and not a lookup.
- The delete/rename asymmetry on `expected_updated_at` is argued rather than assumed, and I agree with where it lands.
- `policy.rs`'s changes are entirely inside `mod tests` — no gate logic moved. Given how much traffic that file is seeing right now, that's worth knowing for whoever sequences this against #672.
- All four mutating tools ride the same `can_write` flag. The doc's argument for that (any grant permitting `workspace_write` has already permitted strictly more destruction) holds, and the delete-removes-authorship nuance is acknowledged rather than glossed.

Finding (1) is the one I'd want fixed before merge. (2) is a follow-up unless one of the six turns out to park under its tool name, in which case it's a one-liner here. (3) is a note for whoever merges second.

oxoxDev:
Thanks — finding (1) was real, and it was in the property the module doc claims. Fixed in `61e0b675`.

**Confirmed before touching anything.** `PathIndex::build` drops an unaddressable node from both maps and keeps only a tally; the gate counted `by_path` entries under the folder's prefix. So a folder holding only such children reported `held == 0`, passed the refusal, and reached a port whose `delete` is recursive. Your framing of why it's worse than a miscount is the right one and I've put it in the commit message: the nodes taken are precisely the ones this layer made unreachable so no tool could touch them, so they're never counted, announced, or named on the card.

**Took your option (b), in a cheaper form than the ancestor walk.** `PathIndex` now carries `child_count: HashMap<String, usize>`, built from `parent_id` over every node **before** the addressability filter. Direct children only — a folder with no direct child has no descendants — so it needs no walk and no cycle guard. It's also strictly sharper than the count it replaces: exact per node id, where two folders may share a rendered path but never an id. That incidentally subsumes the shared-path case the old comment was worried about.

**On the test.** I wrote the tool-level test you suggested first and it failed at setup: `FsOps` rejects the name at creation, which is exactly what its docstring says, and `FixedTree` is `tree()`-only and lives in the parent's test module. So the pin is at the index level instead, and it asserts **both** measures so the regression is explicit — `by_path` prefix count is `0` (stated as the precondition: "that is the bug") while `child_count` is `1`. Honest gap: no end-to-end tool test drives the unaddressable case, because no shipped test double can construct it. The addressable path is still covered by the existing non-empty test, which now runs through the new measure.

Gates after the fix: `cargo test --locked` **1979**, gated `--tests` **2988**, both clippy lanes 0, fmt 0. `workspace_tools` is 100 tests (was 99).

---

**(2)** — I tried to settle whether any of the six park under their tool name and could not from the code alone. `ApprovalSummary.kind` is documented as "the parked effect's dotted kind", and `external.publish` appears in `DEFAULT_ALWAYS_APPROVE` as an effect string, which is consistent with your guess that `publish_artifact` parks under the effect rather than the tool. But I couldn't find the park site that populates `kind` for a per-call tool gate, so I'm not going to assert either way or add labels speculatively — a wrong label is worse than a generic one. **Filing it as its own issue** with your list and the caveat intact, rather than guessing here.

**(3)** — noted, and it's the right call to name it now. #683 isn't in this branch either. Whichever of us merges second carries the interleaving test; I'll flag it on whichever PR that turns out to be.

On the smaller notes: agreed on `published_record_for_node` being a scan — bounded by per-call parking, and worth revisiting only if delete ever stops parking. And thank you for stating your coverage boundaries explicitly, including what you didn't read and that you ran nothing. That's more useful than a clean approval.

Issue #668, opened 2026-08-11
publish: a binary artifact's version history records no digest and no content, so superseded versions are unrecoverable and identical


M3gA-Mind:
## What this is

A binary artifact's version history records neither the content nor a digest of it, and
binary re-publishes overwrite the same workspace node. So the older versions of a binary
artifact are both **unrecoverable** and **indistinguishable from each other**.

## The mechanism

`src/harness/publish.rs:559`:

```rust
pub fn artifact_body(&self) -> String {
    match self {
        Self::Text(text) => text.clone(),
        Self::Bytes { bytes, mime } => format!(
            "{mime}, {} bytes — stored as a file in the company workspace. Open it there; \
             this record is the version history, not the content.",
            bytes.len()
        ),
    }
}
```

For text, the version *is* the content, so history works. For bytes, the version is a
sentence containing only the mime type and a byte count — no `sha256`, no pointer to a
retained copy.

Meanwhile the mirror overwrites in place: a re-publish of the same path with a payload of
the same kind takes the `write_payload` branch in
`src/company/artifact_mirror.rs`, replacing the node's bytes.

## The failure path

1. An agent publishes `diagram.png` (120 000 bytes). Version 1 records
   `image/png, 120000 bytes — stored as a file …`.
2. It publishes a revised `diagram.png`, also 120 000 bytes. Version 2 records **the
   identical string**, and the workspace node's bytes are overwritten.
3. The operator wants version 1 back.

Version 1's bytes are gone — overwritten in the workspace, never copied into the artifact
chain. And the two versions cannot even be told apart: same mime, same length, same
sentence. The record asserts "this record is the version history" while carrying nothing
that distinguishes one version from another.

Any two binary versions that coincide in mime and length are literally equal strings. That
is not rare for regenerated assets of a fixed size, and it is guaranteed for a re-publish
of an unchanged file.

## Why it matters

The version history is the operator's only recourse when an agent overwrites a good
deliverable with a worse one, and for binaries it currently provides neither the content
nor a way to identify which version was which. #553's stated aim was that "nothing
published is left pointing at the sandbox" — the pointer is gone, but so is any means of
recovering or even naming a superseded binary version.

At minimum, recording the `sha256` the store already computes (`stamped_binary` sets it)
would make versions distinguishable and let a reader tell whether a re-publish changed
anything. Retaining the bytes is a larger question about where version payloads should
live.

## Verified

Confirmed present on `main` @ `0ea4ed5`.

## Related

[...]

PR #688, opened 2026-08-11
publish: tell the truth about a stored payload, and record its digest

### LINKED ISSUE #663: publish: an over-cap payload is reported to the agent as captured and to the operator as stored, while the workspace refused it
## What this is

When the workspace refuses an over-cap published payload, the agent has **already been
told the file was captured in full** and the artifact version **already says it is stored
in the workspace**. The refusal is only `tracing::error!`d. The result is an artifact
record pointing at a file that does not exist — the exact dangling-record failure #553 set
out to remove.

## The failure path

1. `src/harness/publish.rs:789` — the publish tool returns success to the agent as soon as
   the payload is queued:

   ```rust
   let how = match &payload {
       PublishPayload::Text(_) => "captured in full".to_string(),
       PublishPayload::Bytes { bytes, .. } =>
           format!("captured in full as a {}-byte file", bytes.len()),
   };
   Ok(ToolResult::success(format!(
       "Published `{source}` as \"{title}\" ({kind}) — {how}. {receipt_tail}", …
   )))
   ```

   Nothing has been written to the workspace yet. The message is unconditional.

2. `src/harness/publish.rs:559` — the artifact version body for bytes reads:

   > `{mime}, {n} bytes — stored as a file in the company workspace. Open it there; this
   > record is the version history, not the content.`

   Also unconditional, and written before the store is asked.

3. `src/harness/brain.rs:1620` — the workspace write is attempted. On failure:

   ```rust
   Err(err) => tracing::error!(
       …,
       "[publish] could not put the published file into the company workspace; \
        it is still recorded as an artifact"
   ),
   ```

   The log line states the problem exactly and tells nobody who can act on it.

## Why it matters

The agent believes the publish succeeded, so it will not retry and will report success in
its own turn. The operator opens the artifact, reads "stored as a file in the company
workspace. Open it there", goes there, and finds nothing.

This is the shape #553's own body objects to: *"the most expensive thing the product can
produce, a paid image or video generation, became a dangling digest pointing into a
directory that gets wiped."* The reference record was removed, but the failure mode
returned through the error path rather than the success path — a record that promises a
file the workspace refused to store.

Quota refusal is not an exotic error here. `max_blob_mb` and `tree_quota_gb` exist to be
hit, and an over-cap generated image is precisely the case that trips them.

## Suggested direction

The tool's success message and the artifact body should describe what actually happened,
which means the workspace write has to be attempted before either is composed — or the
artifact record has to be reconciled when the write fails, rather than left asserting
something untrue.

## Verified

Confirmed present on `main` @ `0ea4ed5`.

## Related

- #611 / #553 — the PR and issue that introduced this path.
- The artifact record carries no digest and no content for bytes, so nothing downstream can
  detect or repair the inconsistency either.

M3gA-Mind:
## Summary

Closes #663 and #668. Two defects, one mechanism — which is why they are one PR: #668's fix, done the way its own issue asks ("record the `sha256` the store already computes"), is only reachable once #663's ordering question is answered.

**#663** — a published binary's artifact version asserted *"stored as a file in the company workspace. Open it there"* **unconditionally**, and was composed **before** the store was asked. When the workspace refused the file, the claim stood and the refusal was only `tracing::error!`d. The operator opens the artifact, reads "open it there", goes there, finds nothing — the dangling record #553 set out to remove, returning through the error path.

**#668** — the same body carried no digest, so two versions of one binary that coincided in mime and length were *literally equal strings*. Superseded versions were both unrecoverable and unidentifiable.

The body is now composed **from the outcome**. `PayloadStorage` names which of three states it describes — `Pending`, `Stored { sha256 }`, `Refused` — so all three wordings live in one function and cannot drift into contradicting each other. After the mirror, the drain amends the version: on success with the digest, on failure withdrawing the claim.

### The ordering was the real question, and I did not reverse it

#663 suggests "the workspace write has to be attempted before either is composed". That would invert `artifact_mirror`'s deliberate chain-first ordering, whose own doc says writing the record first is what stops *"a node in the tree with no artifact behind it at all"*. So this takes #663's other branch — **reconcile the record after the write** — which keeps that invariant and is also where the store's digest becomes available. One mechanism, both issues, no ordering regression.

### The digest can only come from the store

`create_binary` now returns the stamped node, exactly as `write_binary` already did — the port was **asymmetric**, and this makes it consistent rather than wider. Hashing the same bytes caller-side would be one algorithm called twice, and two calls can disagree if the bytes differ between them. Returning the node makes provenance structural: a future backend cannot quietly substitute its own digest.

Blast radius was bounded as scoped — 3 backends, 2 thin decorators, 2 test stubs, and **no** unrelated caller broke.

### One thing deliberately not in the record

A store error is **never** written to the artifact body. A version is permanent and a backend error can name host paths; the operator needs to know the file is not there, and the diagnosis belongs in the log. (Directly applying the lesson from #614's review, where a URL's userinfo reached the durable approval journal.)

### A narrowing of #663

#663's first bullet says the tool tells the agent the file was "captured in full" before anything is written. Having read it: the payload genuinely **is** captured in full at that point — into the queue — and the receipt tail says *"It appears on this task's Artifacts tab when the run finishes"*, which is future tense and never claims workspace storage. So the tool's message is defensible and is unchanged. **The false claim lived entirely in the artifact body**, and that is what this fixes. Flagged rather than silently doing half the issue.

## API Or Behavior Changes

- `WorkspaceStore::create_binary` returns `Result<WorkspaceNode>` (was `Result<()>`) — the stamped node, carrying the store's `size` and `sha256`.
- `artifact_mirror::materialize` returns `Mirrored { node_id, sha256 }` (was `Result<String>`).
- New `PublishPayload::artifact_body_for(PayloadStorage)`; `artifact_body()` remains as the `Pending` case.
- New `ArtifactRecord::amend_latest_body`, documented as confined to the drain that created the version — an operator revision is still a new version, so `human_edit_diff` is untouched.
- The post-mirror upsert now runs when the **body** changed as well as when the link did. A *text* re-publish composes an identical body and still stores once, so `an_ordinary_republish_writes_the_artifact_once` holds unchanged; a binary's body gains the digest, so it pays one extra write. That test caught an earlier, broader version of this change — see Tests.

## Tests

Gated lane, plus the all-features check. Exit codes from unpiped redirects to a session-private path.

- [x] `cargo clippy --all-targets -- -D warnings` — CI's form, `--locked --no-deps --features openhuman,tinycortex --all-targets`, `EXIT:0`.
- [x] `cargo build --all-targets` — **`cargo check --locked --all-features --all-targets`, `EXIT:0`.** Run deliberately: this PR changes a port signature all three backends implement, and `mongodb.rs` is compiled by neither the default nor the gated lane.
- [x] `cargo test` — `cargo test --features openhuman,tinycortex`: **2963 passed, 0 failed**, `EXIT:0`.

New tests:

- `a_pending_binary_version_does_not_claim_the_file_is_stored` — the body composed before the store is asked must not assert storage.
- `a_stored_binary_version_records_the_stores_digest` — #668's payoff.
- `two_binary_versions_of_the_same_length_differ_by_their_digest` — #668's defect in one assertion, plus a control showing the no-digest body is the indistinguishable case and says so.
- `a_refused_binary_version_withdraws_the_storage_claim` — and asserts the body does **not** send the operator to look for the file.
- `a_prose_version_is_its_content_whatever_the_store_did` — the control: prose is storage-independent, so a change that made every body outcome-dependent would fail here.
- Conformance (`assert_workspace_store`): `create_binary`'s returned node is asserted with **populated** `sha256`, `size`, `id` and `mime` — never `None`, which would pass against a backend that dropped the field.

**Revert-and-check.** With the old unconditional body restored, **51 tests ran** (47 passed, 4 failed): the four outcome-dependent tests fail, and the prose control plus the pre-existing description test pass. Non-zero count quoted deliberately — a filter that matches nothing reports `0 passed; 0 failed` and exits 0.

**An existing test caught me.** My first version made the post-mirror upsert unconditional, and `an_ordinary_republish_writes_the_artifact_once` failed. It was right: a text re-publish gains nothing from a second write. Narrowed to "only when the body actually changed", which keeps that contract and pays the cost only where the digest makes it necessary.

## Documentation

`create_binary`'s doc explains why it returns the node and why provenance matters more than the value; `PayloadStorage` documents why three states rather than a bool; `amend_latest_body` documents why amending is confined to the creating drain.

### DIFF
--- a/src/company/artifact_mirror.rs
+++ b/src/company/artifact_mirror.rs
@@ -125,7 +125,24 @@ pub enum MirrorPayload<'a> {
     },
 }
 
-/// Put `target`'s body into the shared tree and return the node id holding it.
+/// What a publish left in the tree: the node holding it, and — for bytes — the
+/// digest **the store computed** while writing it (issue #668).
+///
+/// The digest is `None` for prose, whose version body is the content itself, so
+/// there is nothing a hash would add. For bytes it is the only thing that tells
+/// two versions of one deliverable apart, and it comes back from
+/// [`WorkspaceStore::create_binary`] / [`write_binary`](WorkspaceStore::write_binary)
+/// rather than being computed here — see those methods for why the provenance
+/// matters more than the value.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Mirrored {
+    /// The workspace node now holding the deliverable.
+    pub node_id: String,
+    /// The store's `sha256` of the bytes it wrote, when the payload was bytes.
+    pub sha256: Option<String>,
+}
+
+/// Put `target`'s body into the shared tree and return what it left there.
 ///
 /// The layout is `Agents/<agent-id>/<task-id>/<source…>`. The agent's folder is
 /// minted on demand by
@@ -161,7 +178,7 @@ pub async fn materialize(
     workspace: &dyn WorkspaceStore,
     company: &CompanyId,
     target: PublishTarget<'_>,
-) -> Result<String> {
+) -> Result<Mirrored> {
     // The cheap path, and the common one on a re-publish: the node from last
     // time still exists, so revise it in place and keep every reference to it
     // (the console's deep link, an operator's bookmark) working.
@@ -178,8 +195,11 @@ pub async fn materialize(
         && node.kind == NodeKind::File
         && node.is_binary() == matches!(target.payload, MirrorPayload::Bytes { .. })
     {
-        write_payload(workspace, company, existing, target).await?;
-        return Ok(existing.to_string());
+        let sha256 = write_payload(workspace, company, existing, target).await?;
+        return Ok(Mirrored {
+            node_id: existing.to_string(),
+            sha256,
+        });
     }
 
     let segments = split_source(target.source)?;
@@ -231,8 +251,11 @@ pub async fn materialize(
                 workspace.delete(company, &id).await?;
                 return create_payload(workspace, company, Some(parent), filename, target).await;
             }
-            write_payload(workspace, company, &id, target).await?;
-            Ok(id)
+            let sha256 = write_payload(workspace, company, &id, target).await?;
+            Ok(Mirrored {
+                node_id: id,
+                sha256,
+            })
         }
         None => create_payload(workspace, company, Some(parent), filename, target).await,
     }
@@ -244,20 +267,21 @@ async fn write_payload(
     company: &CompanyId,
     node_id: &str,
     target: PublishTarget<'_>,
-) -> Result<()> {
+) -> Result<Option<String>> {
     match target.payload {
         MirrorPayload::Text(body) => {
             workspace
                 .write(company, node_id, body, origin(target.agent_id))
                 .await?;
+            Ok(None)
         }
         MirrorPayload::Bytes { bytes, mime } => {
-            workspace
+            let node = workspace
                 .write_binary(company, node_id, bytes, Some(mime), origin(target.agent_id))
                 .await?;
+            Ok(node.sha256)
         }
     }
-    Ok(())
 }
 
 /// Creates a fresh node of the right kind holding `target`'s payload.
@@ -267,7 +291,7 @@ async fn create_payload(
     parent: Option<String>,
     filename: &str,
     target: PublishTarget<'_>,
-) -> Result<String> {
+) -> Result<Mirrored> {
     let mut node = WorkspaceNode {
         id: crate::ports::generate_id(),
         name: filename.to_string(),
@@ -283,13 +307,20 @@ async fn create_payload(
     match target.payload {
         MirrorPayload::Text(body) => {
             workspace.create(company, &node, Some(body)).await?;
+            Ok(Mirrored {
+                node_id: node.id,
+                sha256: None,
+            })
         }
         MirrorPayload::Bytes { bytes, mime } => {
             node.mime = Some(mime.to_string());
-            workspace.create_binary(company, &node, bytes).await?;
+            let stamped = workspace.create_binary(company, &node, bytes).await?;
+            Ok(Mirrored {
+                node_id: stamped.id,
+                sha256: stamped.sha256,
+            })
         }
     }
-    Ok(node.id)
 }
 
 /// Record an edit to `node_id` on the artifact chain that owns it, when one
@@ -578,7 +609,8 @@ mod test {
 
         let id = materialize(ws, &co, target("launch.md", "# Launch"))
             .await
-            .expect("materialize");
+            .expect("materialize")
+            .node_id;
 
         assert_eq!(
             path_of(ws, &co, &id).await,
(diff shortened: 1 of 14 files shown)


oxoxDev (approved):
Refusing to reverse the ordering is the right call and the reason is the good part: #663 suggests attempting the workspace write before composing either record, which would invert `artifact_mirror`'s deliberate chain-first sequence — whose own doc says writing the record first is what prevents a node in the tree with no artifact behind it. Taking the other branch, reconciling after the write, keeps that invariant *and* is where the store's digest becomes available. Two issues, one mechanism, no ordering regression.

Three things I checked rather than took on trust, all of which hold:

- **All four wordings live in `artifact_body_for`**, so `Pending` / `Stored{Some}` / `Stored{None}` / `Refused` cannot drift into contradicting each other. `Stored { sha256: None }` saying so explicitly, instead of implying the version is identified when it is not, is the detail that makes the three-state split worth having.
- **`Refused` deliberately carries no store error text.** "A version is permanent and a backend error can name host paths" is exactly right, and applying #614's lesson — where a URL's userinfo reached the durable approval journal — rather than re-learning it is the best instinct in this PR.
- **The digest comes from the store**, with `create_binary` returning the stamped node as `write_binary` already did. Making an asymmetric port symmetric is the right framing, and it makes provenance structural: a caller-side hash would be one algorithm run twice on bytes that might differ between the runs.

Running `cargo check --locked --all-features --all-targets` because `mongodb.rs` is compiled by neither the default nor the gated lane is the correct instinct for a port-signature change, and worth keeping as a habit for this crate.

**0 major on this PR. 1 merge-order requirement. Approving.**

### This must land BEFORE #683, and the reconciliation has a silent failure mode

#683 (`workspace: mint a shape-changed deliverable before deleting the old one`, also yours) edits **the same `Some(id)` arm of `materialize`** — it replaces the `delete` + `create_payload` pair with a new `replace_payload`. This PR changes `create_payload` and `materialize` to return `Mirrored { node_id, sha256 }`. They will conflict textually, which is fine and loud.

What is not loud is the reconciliation. `replace_payload` currently returns `Result<String>`, and the smallest edit that makes the merged code compile is:

```rust
Ok(Mirrored { node_id: replace_payload(..).await?, sha256: None })
```

That builds, every test passes, and **every shape-changing binary republish silently records no digest** — the exact case #683 exists to handle, quietly exempted from #668's fix. `replace_payload` needs to return `Mirrored` and thread the digest out of its staged `create_payload`, not discard it.

Landing this one first is the cheaper order — #683 is already awaiting a revision for an unrelated defect, so it is rebasing regardless, and it can pick the digest up in the same pass. I have left the same note there so it is recorded where the work happens.

Worth saying plainly: this is the third pair of your own open PRs to collide this way today (#659/#660, #660/#672, now #683/#688). Each is caught, and each is caught by a reviewer rather than by you — while you document cross-PR conflicts with *other people's* branches meticulously, as the #625 section on #672 shows. A quick `git grep` of your own open branches' file lists before opening the next one would close the gap.

**Before merging, not a finding:** the Rust lanes and CodeRabbit are still running. Approving on the code — don't merge while it is `UNSTABLE`. I reviewed the `artifact_mirror` / `publish` core and the port change; I did not read all fourteen files line by line.

Issue #245, opened 2026-07-31
feat(repo): bind real repositories into agent workspaces with native PR/diff fetch


oxoxDev:
## Summary

Agents carry a full coding toolbelt and have nothing real to point it at. The sandbox is an empty scratch directory that nothing ever populates with a repository, so `code-review` and `bug-triage` can only work on text pasted into chat.

## Problem

The tools are all there: sandboxed file tools (`src/harness/build.rs:474-482`), audited `shell` (`:193-206`), and `apply_patch`/`git_operations`/`csv_export` — where `git_operations` is status/diff/log/commit only, with **no clone, fetch or push** (`src/harness/toolbelt.rs:240-249`).

The sandbox at `<workspace_root>/<company>/<agent>/workspace` (`build.rs:144-149`) is empty and stays empty. The only GitHub reachability is Composio (explicit grant plus per-tenant token, `build.rs:253-283`), which has no clone or diff capability at all.

Steps to reproduce:

1. Paste a PR link and ask the Security Engineer to review it.
2. The agent has no tool that can fetch the diff.
3. You paste the diff by hand, and the review quality is fine — which is the point: the reasoning works, the access does not.

## Solution

**Read tier only in this issue**, behind a new explicit `repo` grant. Write access (opening a PR) is staged into a linked follow-up. That halves the credential blast radius — v1 runs on a fine-grained, single-repo, **read-only** PAT — while still unblocking the skills.

1. **Binding and materialisation, two layers.** An operator binds a repo (URL plus credential) to a company via a new REST ops surface. Materialisation is (a) a persistent **bare mirror cache** *outside* any agent workspace at `<data_dir>/companies/<slug>/repos/<key>.git`, fetched incrementally host-side, restricted to configured branches plus on-demand `refs/pull/N/head`; and (b) on dispatch, a **task-scoped clone** from that cache into `<workspace>/repos/<key>` — same filesystem, so git hardlinks objects and the clone is near-instant with **zero network in the agent path**.
   Rejected: a persistent workspace mount (stale, dirty across tasks, N-agents × repo-size disk, long-lived mutable state inside the sandbox) and clone-from-remote per dispatch (full clone latency plus the credential in the hot path every task).
[...]

PR #689, opened 2026-08-11
feat(repo): bind repositories and mirror them host-side — operator half (#245)

oxoxDev:
## Summary

Agents carry a full coding toolbelt and have nothing real to point it at. The sandbox is an empty scratch directory nothing ever populates, and no tool can fetch a repository — `git_operations` is status/diff/log/commit only, with no clone, fetch or push, and the only GitHub reachability is Composio, which has neither clone nor diff. Paste a PR link and ask for a review and the agent has no way to get the diff; paste the diff by hand and the review is good — the reasoning works, the access does not.

This is **PR-A of two**: the operator half. An operator can bind a repository, the host keeps its own mirror, and **nothing reads it yet**. The `repo` grant and the agent-facing tools are PR-B.

Part of #245.

## API Or Behavior Changes

**The split point is the credential, deliberately.** PR-A ships the credential path *whole* — stored, used host-side, redacted everywhere — with **zero agent surface**. No intermediate state ever ships a half-wired credential. Nothing under `src/harness/**` or `src/policy/**` is touched.

**`RepoManager`** (`src/runtime/repo_manager.rs` + `types.rs` / `git.rs` / `github.rs`): bind, list, revoke, fetch, materialize, pull-request. Bare mirrors live at `companies/<slug>/repos/<key>.git`, fetched incrementally, restricted to the binding's branches plus `refs/pull/N/head` on demand.

**The credential never becomes readable.** Metadata sits in a `repos/bindings` JSON document carrying only a token *fingerprint*; the token itself sits alone under `repos/token/<key>`, so **no list read ever has token bytes in hand**. It reaches git through an askpass helper over a pipe — never argv (`/proc/*/cmdline`), never the environment (`/proc/<pid>/environ` is same-uid readable), never persisted config. The mirror's remote URL is credential-less, and `-c credential.helper=` clears inherited helpers.

**Honest limit, stated because the PR would otherwise imply containment it does not have**: the agent shell and the host-side fetch run as the **same user in the same container**, so this is not a kernel boundary. A prompt-injected agent holding the `shell` grant could still attack `/proc` during a fetch window. The real protections are the fine-grained, single-repo, **read-only** PAT the card insists on, the shell audit log, and an egress policy that is a follow-up rather than a claim.

**Clones use `git clone --shared` (alternates), not hardlinks** — a deliberate deviation from the issue text. Hardlinked inodes would let an agent chmod and rewrite shared mirror objects, silently corrupting every other agent's view. Alternates are read-only into the cache. The consequence is that mirror `gc` must never prune (`gc.pruneExpire=never`), so space is reclaimed by whole-mirror deletion at revoke.

**Per the #459 precedent** — `read_workspace_state` was reclassified because it shells out to git in an agent-writable directory, so repository config decides what runs — every host-side git against an agent-writable or third-party-authored tree passes `-c core.hooksPath=/dev/null -c core.fsmonitor=`.

**Quota refuses rather than evicts**, the second deliberate deviation: an over-cap fetch fails loudly instead of silently discarding a bound mirror.

**Routes**: `POST /repos` (validates https github.com only, **refuses classic `ghp_` tokens**, stores the secret, performs the initial fetch, rolls back on auth failure), `GET /repos` (redacted), `POST /repos/{key}/revoke`. Both scope forms; mutations behind `AdminScopedCompany`.

## A production bug found by accident, and fixed

One test reached `github.com` and **hung for eleven minutes** — `git-remote-https` parked rather than failing. That was not a test bug: **`git` had no deadline at all.** A fetch against a remote that neither accepts nor refuses a credential would hang an HTTP request forever, leaving an orphaned child process **still holding the token**. Now bounded by a 300s deadline plus `kill_on_drop` so the child is reaped. The offending test now collides on a key with no network involved.

## Six places the plan or the issue was wrong

1. **`frontend/src/connections/registry.ts` was the wrong file.** That registry tracks *which OpenCompany host the console is talking to*, not providers. There was no entry to add.
2. **`tempfile` is a dev-dependency only.** The askpass scratch dir is hand-rolled with `create_dir_all` + `0700` — no new shipped dependency.
3. **The "inherited pipe fd" is fd 0, not fd 3.** Handing git an arbitrary descriptor needs `pre_exec` + `dup2` — `unsafe` plus a `libc` dependency — to obtain properties stdin already has. Every command run (`fetch`, `ls-remote`, `clone`, `init`, `config`, `rev-parse`) reads nothing from stdin. Documented in the module header.
4. **`paths::slug()` is private**, so the builder could not reach `DataLayout::company_repos_dir` without duplicating the slug rule. Added `Bundle::repos_dir()` in `src/store/paths.rs` — a file in neither plan.
5. **"Large diffs spill to a file" is not expressible at this tier** — there is no workspace to spill into yet. Truncated at 1 MiB on a UTF-8 boundary with a visible marker; spilling belongs to PR-B.
6. **`storage.md` got no link.** It is 575 lines, already over the 500-line Markdown cap and already split once. Linked from `docs/spec/runtime/README.md` and `docs/spec/README.md` instead.

Plan claims verified **true**: `usage_bytes()` covers the cache for free (now asserted by a test), `SecretStore` has no delete (used the `ops/channels.rs` overwrite convention), and `git_operations` has no clone/fetch/push.

## Tests

- [x] `cargo test --locked` — **2022 passed**, 0 failed
- [x] `cargo test --locked --features openhuman,tinycortex --tests` — **3001 passed**, 0 failed
- [x] `frontend`: `typecheck` 0, `build` 0, `vitest` **418 passed**, **`./scripts/ci/assert-design-tokens.sh` 0**

42 new Rust tests, 7 new frontend. All git tests run against a `file://` fixture remote — no network in CI. The credential test binds with a sentinel token, then walks the mirror and every file written asserting the sentinel appears nowhere, and asserts the same of the real git argv and environment.

**Verified on a live console** (the one gap the implementing agent could not close): the Repositories card renders in Settings → Connections, shows its "Nothing reads these yet" callout, and a classic `ghp_` token entered through the form is refused with the full actionable message surfaced as a toast — and the token field clears itself afterwards. The agent had reported the local console could not be signed into; that turned out not to be the case, so the surface is confirmed rather than assumed.

## Documentation

New `docs/spec/runtime/repos.md`, linked from both spec READMEs. The module header carries the credential-delivery reasoning, the alternates-not-hardlinks argument, and the fd-0 note. The card itself states plainly that nothing reads these bindings yet.

## Related

- #245 — the issue; **PR-B** (the `repo` grant, `repo_checkout` / `repo_pr`, clone lifecycle) follows this
- #459 — `read_workspace_state` reclassified for shelling out to git in an agent-writable directory; the precedent behind `hooksPath=/dev/null`
- #671 — agent workspace lifecycle, developed in parallel; zero file overlap with this PR, but **PR-B will collide** with it in `consequence.rs` and `build.rs`


oxoxDev:
Two CI failures on this branch, both fixed, and both the same underlying assumption — worth naming as a pattern rather than two incidents.

**`903d851b` — the askpass credential test.** The username invocation answers a fixed literal *without* reading stdin, which is the property that test exists to prove, so the helper exits and the test then writes into a closed pipe. `run_git` already tolerates the identical condition deliberately, with a comment saying a git that exited before reading its stdin is expected — production anticipated it and the test did not. `BrokenPipe` is now accepted; every other error still fails, so the tolerance is narrowed rather than widened.

**`98d9db5a` — the deadline test.** It used `Duration::ZERO` and its comment claimed that was "deterministic rather than a race against process spawn". It is not: `tokio::time::timeout` polls the inner future first and consults the deadline only on `Pending`, so a `git version` whose output is already buffered returns `Ok` and the timeout arm never runs. CI returned `GitOutput { ok: true, stdout: "git version 2.54.0" }`. Now on virtual time via `start_paused`, where the clock auto-advances whenever the runtime idles — exactly its state while waiting on a real child's I/O. Verified 10/10, and it now runs in 0.00s instead of racing. `tokio/test-util` is dev-only; under resolver v2 the feature is not unified into normal builds.

**The pattern:** both tests assumed a real subprocess reliably loses a race against a short deadline. Both passed on macOS and failed on Linux CI. Worth watching for in PR-B, which adds more process-driving tests.

---

I also read this adversarially rather than only re-running gates, going after the load-bearing security claims. They hold:

- **The host environment is not inherited.** This was the gap I most expected to find — if the git child inherited the host's env, an agent reading `/proc` during a fetch would get every secret the host holds, not just the repo token, which would make the stated same-uid limitation much worse than described. `cmd.env_clear()` closes it, and the comment names the reasoning: building the environment up rather than filtering it down makes it a property of the code instead of a claim about it.
- **`SpawnPlan` as data** is the right call — "the credential is in neither argv nor env" is inspectable, so the sentinel test asserts it over the real plan rather than over whatever the child happened to print.
- **Hardening is at a single chokepoint.** One `Command::new("git")`, `hardening_flags()` in front of every subcommand, so `core.hooksPath=/dev/null` cannot be forgotten on a future call site — which is the #459 lesson applied structurally rather than by discipline.
- **`truncate_utf8`** walks back off continuation bytes, and its test straddles the limit with a three-byte character asserting no replacement character. That is the #553 byte-slice class, handled.

Gates after both fixes: `cargo test --locked` **2022**, gated `--tests` **3001**, both clippy lanes 0, `--all-features` check 0, fmt 0.

Still worth a second reader: CodeRabbit has been rate-limited across this whole batch and reports `pass` while having read nothing, so nothing here has had an external pass.

src/runtime/repo_manager.rs
```diff
@@ -0,0 +1,882 @@
...
+                "clone",
+                "--quiet",
+                // Alternates, not hardlinks. See this module's header.
+                "--shared",
+                "--branch",
+                &reference,
+                "--single-branch",
+                &mirror_str,
+                &dest_str,
+            ],
```
M3gA-Mind: **[P1] Prevent checkouts from writing back to the host mirror.** `git clone --shared` leaves `origin` pointing at the bare host cache. I reproduced that a commit followed by `git push origin HEAD:refs/heads/main` directly advances this mirror. Because Git resolves the path from `.git/config`, command path guards do not protect it, so an agent could poison future checkouts. Use an isolated object copy or enforce a real read-only filesystem boundary; removing/replacing `origin` alone is insufficient while alternates remain writable.

src/runtime/repo_manager.rs
```diff
@@ -0,0 +1,882 @@
...
+                 repository the account can reach. Create a fine-grained token instead \
+                 (GitHub → Settings → Developer settings → Personal access tokens → \
+                 Fine-grained tokens), scope it to this one repository, and give it \
+                 read-only access to Contents, Metadata and Pull requests."
+                    .to_string(),
+            ));
+        }
+        let branches = normalize_branches(&request.branches)?;
+
+        if self.list().await?.iter().any(|b| b.key == key) {
```
M3gA-Mind, line 308: **[P1] Serialize same-key binds before touching shared state.** The duplicate check runs outside `index_lock`. Two concurrent binds can both pass it, overwrite the same token, and operate on the same mirror. When one later detects the conflict, `bind` calls `roll_back`, blanking the winning request's credential and deleting its mirror while its index entry remains. Hold a per-key lock across validation, token/mirror installation, and index commit, or stage each attempt privately and atomically publish only the winner.

src/runtime/repo_manager/github.rs
```diff
@@ -0,0 +1,224 @@
...
+            return Err(OpenCompanyError::Store(format!(
+                "the GitHub API answered {status}"
+            )));
+        }
+        // Deliberately not `response.text()`: that buffers whatever arrives.
+        let bytes = response
+            .bytes()
+            .await
+            .map_err(|e| OpenCompanyError::Store(format!("reading the GitHub response: {e}")))?;
+        Ok(truncate_utf8(&bytes, MAX_DIFF_BYTES))
```
M3gA-Mind, line 126: **[P2] Stream the response before applying the 1 MiB cap.** `Response::bytes()` collects the complete response before `truncate_utf8` runs, so a very large PR diff still consumes its full size in memory and can exhaust the host. The comment claiming this avoids buffering is therefore incorrect. Read chunks into a bounded buffer, stop after `MAX_DIFF_BYTES + 1`, and then append the truncation marker.

M3gA-Mind (commented):
Four findings from review. The first two affect the integrity/atomicity guarantees of the repository cache; the remaining two cover response bounding and repository identity.

src/runtime/repo_manager/types.rs
```diff
@@ -0,0 +1,402 @@
...
+    pub fn canonical_url(&self) -> String {
+        format!("https://{SUPPORTED_HOST}/{}/{}", self.owner, self.repo)
+    }
+
+    /// The stable cache key (`owner-repo`, lowercased). Used as a single path
+    /// component under the mirror cache and as the id in the REST surface, so
+    /// it is constrained to `[a-z0-9-]` — see [`parse_repo_url`], which is the
+    /// only constructor.
+    pub fn key(&self) -> String {
+        format!("{}-{}", slugify(&self.owner), slugify(&self.repo))
```
M3gA-Mind: **[P2] Make repository keys collision-resistant.** This construction is not injective: valid coordinates such as `a-b/c` and `a/b-c` both become `a-b-c`; punctuation normalization also collapses `foo.bar`, `foo_bar`, and `foo-bar`. Because the key identifies the token, mirror, index entry, and revoke route, distinct repositories cannot coexist. Preserve component boundaries and characters with an encoding, or include a hash of the canonical URL.

Issue #326, opened 2026-08-04
workspace: a note carries no authorship, so agent-written and operator-written files are indistinguishable


oxoxDev:
## Summary

A workspace note records no author or origin. `WorkspaceNode` has no such field anywhere — not on the port, not in either store backend, not in the REST or GraphQL projections, and not in the agent tools. So a file an agent produced and a file the operator typed look exactly alike.

Surfaced while planning #177 (wiring the console Workspace tab to the real backend store). Once the tab reads the shared tree, the operator finally sees agent output — and immediately cannot tell which of it is theirs.

## Why it matters

The workspace is meant to be the single source of truth every agent references. Reviewing what an agent wrote is a core loop: an operator needs to know what to check. Without origin, that review has to happen from memory.

It also blocks the honest version of a question the product already asks elsewhere. `MemoryFact` carries an `origin` field and the console renders it, so the pattern is established in-repo — the workspace simply does not have it.

## What #177 does instead

#177 ships `updatedAt` as an interim signal: a recently-changed file is at least noticeable. That is recency, not authorship — it cannot distinguish an agent write from the operator's own edit thirty seconds ago, and it says nothing about *which* agent.

## Scope

Adding origin touches the `WorkspaceStore` trait, both store backends, the seeder, the REST and GraphQL read projections, the agent write tools, and the console. That breadth is why #177 does not fold it in.

Worth deciding at design time: whether origin is a free-text author id, an enum (`Operator` / `Agent(id)` / `Seed`), and whether it records the *creator*, the *last writer*, or both. A file created by an agent and since edited by the operator is the interesting case, and a single field answers it wrong.

## Related

- #177 — console Workspace tab reads localStorage, not the backend workspace
- #237 — surfaced the company workspace to agents as live read/write tools

oxoxDev:
**Related — `cluster:workspace`**

The Workspace notes surface. All four surfaced from wiring the tab to the real backend store (#177, closed).

- #326 — a note carries no authorship, so agent-written and operator-written files are indistinguishable the moment the operator can finally see agent output.
- #327 — no workspace event on the SSE feed, so an agent-written file appears only after a refresh. #326 and #327 are the two halves #177 deliberately left out; doing them together is cheaper than either alone.
[...]

Issue #71, opened 2026-07-23
Active Runtime Teammates: promote overlay agents to real agents + orchestrator add-agent tool


oxoxDev:
## Problem
Agents added via the Team surface are presentation-only and never become executable runtime teammates, and the orchestrator has no way to add or reach a new agent.

## Evidence
- Overlay agents (`POST .../team` → `record.overlay_agents`) are explicitly "roster-only in v1" (`src/server/ops/team.rs`; `docs/spec/feature-audit/02-active-runtime-teammates.md`). They are merged into the roster at read time but not materialized as harness workers.
- The orchestrator has no `add_agent`/`create_agent` tool (`src/harness/orchestrator.rs`), and `delegate_to_desk` only reaches desk leads that are real manifest agents — an overlay agent is unreachable.
- Observed live: asking the orchestrator to add/use a "researcher" fails; the researcher exists only as an overlay card.

## Proposed fix
- Implement the "Active Runtime Teammates" capability: promote an overlay agent into a real runtime agent (`src/runtime/builder.rs` + `src/harness/`), and add an orchestrator tool to add/assign an agent.

Related: #26

PR #690, opened 2026-08-11
feat(team): give runtime-added teammates a readable roster id (#686)

### LINKED ISSUE #686: team: a runtime-added teammate gets a ULID roster id, so its workspace folder is unreadable
## Problem

A teammate added at runtime gets a ULID as its roster id, so its workspace folder is named after the ULID. Observed live on the `smoke1` staging tenant: of 17 folders under `Agents/`, **8 read like this**:

```
Agents/019fa75dbc9b-000000000005/
Agents/019fad5ada20-000000000003/
Agents/019fb7aaebf5-000000000007/
…
Agents/backend_engineer/
Agents/product_manager/
Agents/qa_engineer/
```

The manifest-defined agents are legible. The runtime-added ones are not. An operator browsing the Workspace tab cannot tell which teammate produced what, and neither can another agent reading the tree.

## Where it comes from

Two paths mint the id, both with `generate_id()`:

- `src/server/ops/team.rs:418` — the console "add member" route
- `src/harness/orchestrator.rs:1641` — the agent-facing `add_agent` tool (#71)

Manifest agents never go through either; their id is the manifest key, which is why they are slugs.

So the split is not cosmetic — it is two different id-minting policies for the same concept, and only one of them produces something a human can read.

## Why it matters now, when it did not before

The roster id used to be an internal key. Three recent changes turned it into **a name people and agents read**:

- **#570** — an agent's folder is `Agents/<roster-id>/`, minted on first use.
- **#552** — published deliverables land at `Agents/<roster-id>/<task-id>/`, so the id is in the path of every artifact.
- **#607** — the tree is searchable, and hit paths are rendered to agents and operators alike.

A ULID in an internal map is fine. A ULID in a folder path that appears in search results is a usability defect.

## The hard part — read before proposing a fix

**The roster id is a durable key, not a label.** It is referenced by:

- workspace folder *names* (`Agents/<id>/`), which are the reverse-lookup key `ensure_agent_folder` uses
- `WorkspaceOrigin::Agent { id }` stamped on every node that agent created or updated (#326)
- budget overrides (`BudgetOverride.agent_id`)
- chat DM routes (`#/chat/dm:<id>`)
- task assignees, and anything else that names a teammate

So **changing an existing agent's id is a migration, not a rename** — every one of those references has to move with it, and the workspace nodes carry the old id in both their path and their authorship stamp. Any proposal that says "just slugify it" without addressing existing records is incomplete.

## Two directions, and they are not exclusive

**A. Derive a slug at creation, for new agents only.** Mirror `company_id_from_name` (`src/runtime/builder.rs`), which already derives `e2e-harness-co` from "E2E Harness Co". Needs a collision rule — two teammates both named "Designer" cannot both be `designer` — and the rule must be deterministic and stated, not incidental. Leaves existing ULID agents as they are.

**B. Stop rendering the id where a human is reading.** The console tree and origin badges could show the teammate's display name (`Dana — Designer`) with the id as a tooltip or secondary. That fixes legibility for *existing* agents too, which A cannot, and it needs no migration. But it does not fix the *path* — `Agents/019fad…/notes.md` is still what search returns and what an agent sees.

My read: **A alone leaves today's tenants broken; B alone leaves the paths ugly.** A is the right fix going forward, B is the right fix for what already exists, and a migration for existing agents is a third, larger question that should be decided explicitly rather than assumed.

## Open questions for the plan

1. **Collision rule.** Deterministic suffix (`designer-2`), or refuse and make the operator pick? Refusing is more honest but worse for the `add_agent` tool, which has no human in the loop.
2. **Does an id need to be stable forever?** If a slug is derived from a name, what happens when the operator renames the teammate? Almost certainly the id must *not* follow the name — but that should be stated, because it means a mis-typed name is permanent.
3. **Is a migration wanted at all?** Existing ULID agents work; they are only ugly. A migration touches authorship stamps, which #326 exists to keep honest — rewriting them is not free.
4. **Reserved names.** A derived slug must not collide with `Agents`, `Desks`, or a manifest agent id.

## Acceptance criteria (direction A, minimum)

- [ ] A teammate added through the console or through `add_agent` gets a readable, slug-shaped id derived from its name.
- [ ] Collisions resolve by a stated deterministic rule, pinned by a test.
- [ ] Reserved and manifest-agent names cannot be taken.
- [ ] Existing ULID-id agents keep working untouched — no silent migration.
- [ ] The workspace folder for a newly added teammate is legible in the tree and in search results.

## Notes and scope

Small-to-medium for direction A. The migration question is deliberately excluded above and should be its own issue if wanted.

Found while manually verifying the workspace on the `smoke1` staging tenant — the first end-to-end pass over this surface against real agent-produced data.

## Related

- #570 — member folders minted on first use; the change that put the id into a path
- #552 — deliverables land under `Agents/<id>/<task-id>/`
- #326 — `WorkspaceOrigin::Agent { id }`; the authorship stamp any migration would have to rewrite
- #607 — search; renders those paths to agents and operators
- #71 — the `add_agent` tool, one of the two minting sites

oxoxDev:
## Summary

A teammate added at runtime got a ULID as its roster id, so its workspace folder was named after the ULID. Observed live on the `smoke1` staging tenant — of 17 folders under `Agents/`, **8 read like this**:

```
Agents/019fa75dbc9b-000000000005/     ← added at runtime
Agents/019fad5ada20-000000000003/
Agents/backend_engineer/              ← from the manifest
Agents/product_manager/
```

Two minting policies for the same concept, and only one produced something a human can read. Runtime-added teammates now get a snake_case slug derived from their display name.

Closes #686.

## The bug this actually prevents

The cosmetics are the symptom. **`build_roster` skips an overlay id that a manifest agent already claims** — so once ids come from names, adding a teammate called "Backend Engineer" to a company with a manifest `backend_engineer` would have **saved a teammate that never materialises**, erroring nowhere. It would sit in the record, absent from the roster, silently doing nothing.

That was unreachable while ids were generated (ULIDs never collide) and became reachable the moment ids came from names. The existing duplicate-*name* guard does not catch it: it compares overlay names only, and the collision here is against a *manifest id*. Suffixing is what makes name-derived ids safe, not what makes them tidy.

## API Or Behavior Changes

**`agent_slug(display_name)`** (`src/ports/ids.rs`) — lowercase, keep ASCII alphanumerics, collapse every other run to a single `_`, trim, cap at 64. Returns the stem `teammate` when the result would not start with a lowercase letter, mirroring `company_id_from_name`'s `"company"` fallback. **Snake_case with underscores, not that function's hyphens**: the roster-id grammar is set by the manifest validator and by every existing agent (`backend_engineer`), and a hyphen dialect would be a third id shape in one system.

**`CompanyRecord::mint_agent_id()`** (`src/ports/types.rs`) — bare slug when free, else the smallest `_2`, `_3`, … Free means not colliding (case-insensitively) with manifest ids, overlay ids, desk ids, or the reserved set. `RESERVED_AGENT_IDS` is built from the real `OPERATOR_CHANNEL` / `AGENTS_ROOT` / `DESKS_ROOT` constants rather than re-typed literals, so a rename upstream cannot silently unreserve a name.

**Both minting sites** — `add_member` (`src/server/ops/team.rs`) and `AddAgentTool` (`src/harness/orchestrator.rs`) — mint through it, inside the shared per-company write lock they already held, so check-then-save is race-safe between the console and the tool.

**`add_agent` now reports the minted id** in its success text. That was the one end-to-end gap this change owed: without it the orchestrator creates a teammate and cannot address it afterwards. The console already had the id in its DTO.

**Existing ULID-id agents are untouched.** No migration, deliberately: it would have to rewrite `WorkspaceOrigin::Agent` authorship stamps — the records #326 exists to keep honest — plus every workspace path. **Zero existing test bodies were modified**, and the ULID-shaped fixtures stayed green untouched; that is the back-compat proof, not an assertion.

## Three places the plan was wrong

1. **A rename route already exists.** The plan stated none did, and asked for a doc comment describing the minted-once policy "for when one arrives". `PATCH …/team/{agent_id}` (`team_agent.rs::edit_agent`) has been writing `agent.name` and leaving the id alone all along. So the policy is **pinned by a test**, not just described — and the practical consequence is that a typo'd name is fixable today, while the folder keeps the original slug.
2. **Reserving desk *display names* would have been superstition.** The plan asked for it. But comparing a minted slug against a raw desk name only bites when the name is already slug-shaped modulo case: a desk called "Content" blocks `content`, while "Content Desk" leaves `content_desk` free, because nothing routes on the slug of a desk name. Not reserved; both directions pinned.
3. **`is_snake_case` was private**, so the grammar sweep needed a visibility change (`pub(crate)`) — the one file outside the planned list. Worth it: the sweep now couples to the real validator instead of a copy of its rules.

Two claims verified by reading rather than asserted by omission: **nothing parses a roster id by shape**, and the frontend uses `agent.id` in exactly three places (equality, avatar-tone hash, rendering) — **no frontend change is owed**.

## Tests

- [x] `cargo test --locked` — **1995 passed**, 0 failed
- [x] `cargo test --locked --features openhuman,tinycortex --tests` — **2978 passed**, 0 failed

20 new tests: slug normalisation including all-punctuation, empty, non-ASCII and digit-leading names → `teammate`; the 64-char cap; a sweep asserting **every** output satisfies the real `is_snake_case` grammar; `"Backend Engineer"` against a manifest `backend_engineer` → `backend_engineer_2` **and present in the built roster** (the silent-drop pin); deterministic `_2`/`_3` ordering; desk-id and reserved-name collisions; the rename route leaving the id alone; and route/tool coverage at both minting sites.

No frontend gates — no frontend file is touched.

## Documentation

`docs/spec/runtime/api.md` describes name-derived ids, the suffix rule, and the minted-once policy.

**Flagged, not fixed**: that file is now **532 lines against the repo's 500-line Markdown cap**. It was **already 506 on `main`**, so this inherits a violation rather than creating one; the addition was tightened from 33 lines to 26 rather than splitting the file, which is separate work.

## Related

- #570 — member folders minted on first use; the change that put the roster id into a path
- #552 — deliverables land at `Agents/<id>/<task-id>/`, so the id is in every artifact's path
- #607 — search renders those paths to agents and operators alike, which is what made this visible
- #326 — `WorkspaceOrigin::Agent { id }`; the authorship stamps a migration would have to rewrite, and the reason there is no migration here
- #71 — the `add_agent` tool, one of the two minting sites

### DIFF
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -223,6 +223,32 @@ is never rewritten. Overlay teammates are addressable: since issue #71 the
 harness builds a real agent for each one, with the company-wide tool grant, no
 cognition tier, and never the orchestrator.
 
+`POST …/team` and the orchestrator's `add_agent` tool both **derive the roster
+id from the display name** (issue #686): "Dana Designer" becomes
+`dana_designer`, in the same snake_case grammar the manifest validator enforces
+on a hand-authored `[[agent]].id`. They used to mint an opaque
+`{millis}-{counter}` id, which #570/#552/#607 render as a workspace folder and
+in search-hit paths — so half a company's tree read as
+`Agents/019fad5ada20-000000000003/` beside `Agents/backend_engineer/`.
+
+- **Collisions suffix, they do not refuse.** A slug already held by a manifest
+  agent, another teammate, a desk id or name, or a reserved word (`operator`,
+  `Agents`, `Desks`) becomes `<slug>_2`, `_3`, … Duplicate display names have
+  always been accepted here, and an unsuffixed collision with a *manifest* id is
+  worse than a refusal: the roster build skips it, so the teammate would persist
+  and never materialise.
+- **Minted once, never re-minted.** `PATCH …/team/{agentId}` renames a teammate
+  and leaves the id alone; a name-keyed id would orphan its workspace folder,
+  budget row, desk memberships and inbox on every correction.
+- **Removal frees the slug**, so re-adding the same name takes the id back and
+  **adopts the old `Agents/<slug>/` folder** — the intended remedy for a typo'd
+  name, and not a way to get a clean slate.
+
+Teammates carrying generated ids are **not migrated**: rewriting them would
+rewrite the `WorkspaceOrigin` stamps issue #326 keeps honest, and every path
+into their folders. They keep working, reachable by display name through
+`crate::runtime::assignee`.
+
 `GET …/team/{agentId}` is the **agent detail** read (issue #264). `GET …/team`
 answers "who is on the roster"; this answers "what is this agent", and before it
 existed neither the console nor any other client could reach an agent's tier,
--- a/src/company/manifest.rs
+++ b/src/company/manifest.rs
@@ -386,7 +386,7 @@ impl CompanyManifest {
 
 /// True when `id` is non-empty, starts with a lowercase letter, and contains
 /// only lowercase letters, digits, and underscores.
-fn is_snake_case(id: &str) -> bool {
+pub(crate) fn is_snake_case(id: &str) -> bool {
     let mut chars = id.chars();
     match chars.next() {
         Some(first) if first.is_ascii_lowercase() => {}
--- a/src/company/mod.rs
+++ b/src/company/mod.rs
@@ -72,6 +72,7 @@ pub use credentials::{Credential, CredentialSource, TinyhumansTokenSource, Token
 /// The roster-id grammar check, shared with the runtime id minter so a slug and
 /// a hand-authored `[[agent]].id` are held to one rule (issue #686). Not `pub`:
 /// outside the crate the validator speaks through `CompanyManifest::validate`.
+#[cfg(test)]
 pub(crate) use manifest::is_snake_case;
 pub use manifest::{LEGACY_MANIFEST_FILE, Located, MANIFEST_FILE, discover};
 pub use skill_file::{SkillDoc, load_dir_skills, parse_skill_md, render_skill_md};
--- a/src/company/mod.rs
+++ b/src/company/mod.rs
@@ -69,6 +69,10 @@ pub mod workspace_seed;
 use std::path::Path;
 
 pub use credentials::{Credential, CredentialSource, TinyhumansTokenSource, TokenTier};
+/// The roster-id grammar check, shared with the runtime id minter so a slug and
+/// a hand-authored `[[agent]].id` are held to one rule (issue #686). Not `pub`:
+/// outside the crate the validator speaks through `CompanyManifest::validate`.
+pub(crate) use manifest::is_snake_case;
 pub use manifest::{LEGACY_MANIFEST_FILE, Located, MANIFEST_FILE, discover};
 pub use skill_file::{SkillDoc, load_dir_skills, parse_skill_md, render_skill_md};
 pub use types::{
--- a/src/harness/mod.rs
+++ b/src/harness/mod.rs
@@ -2423,6 +2423,78 @@ description = "Builds the product."
         );
     }
 
+    /// Issue #686, end to end: the orchestrator adds a teammate whose display
+    /// name slugs onto a **manifest** agent's id, and the teammate still shows
+    /// up in the built roster.
+    ///
+    /// This is the failure the suffix exists to prevent, and it only became
+    /// reachable when ids started coming from names. `add_agent`'s duplicate
+    /// guard compares overlay *names*, so "Engineer" sails past it; an
+    /// unsuffixed `engineer` would then be skipped by
+    /// [`build_roster`](super::build_roster) as already claimed by the manifest
+    /// — saved to the record, never materialised, no error anywhere.
+    #[tokio::test]
+    async fn a_tool_added_teammate_colliding_with_a_manifest_id_still_joins_the_roster() {
+        use openhuman_core::openhuman::tools::Tool;
+
+        use crate::harness::orchestrator::AddAgentTool;
+
+        /// A `CompanyStore` that actually holds the record, unlike
+        /// `RecordingStore` — `add_agent` has to load what it saves.
+        struct SeededStore(StdMutex<CompanyRecord>);
+
+        #[async_trait]
+        impl CompanyStore for SeededStore {
+            async fn load(&self, _id: &CompanyId) -> crate::Result<Option<CompanyRecord>> {
+                Ok(Some(self.0.lock().unwrap().clone()))
+            }
+            async fn save(&self, record: &CompanyRecord) -> crate::Result<()> {
+                *self.0.lock().unwrap() = record.clone();
+                Ok(())
+            }
+            async fn list(&self) -> crate::Result<Vec<CompanySummary>> {
+                Ok(Vec::new())
+            }
+            async fn append_ledger(
+                &self,
+                _id: &CompanyId,
+                _entry: LedgerEntry,
+            ) -> crate::Result<()> {
+                Ok(())
+            }
+        }
+
+        let fx = fixture();
+        let company = CompanyId::new("acme");
+        let store = Arc::new(SeededStore(StdMutex::new(record())));
+        let tool = AddAgentTool::new(company.clone(), store.clone());
+
+        let result = tool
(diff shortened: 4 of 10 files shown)


oxoxDev:
Read this one adversarially before merge rather than only re-running gates — looking for the shape of defect @M3gA-Mind found on #687, where a claimed invariant does not hold.

**The load-bearing claim holds, and is pinned properly.** The silent-drop is the reason this change is more than cosmetic: `build_roster` skips an overlay id a manifest agent already claims, so an unsuffixed collision would persist a teammate that never materialises and errors nowhere. That was unreachable while ids were generated and became reachable the moment they came from names. The test does the right thing — it builds the roster and asserts membership, rather than asserting the suffix and stopping:

```
assert_eq!(ids, vec!["ceo", "engineer", "engineer_2"],
    "a suffixed id materialises; an unsuffixed one would vanish here");
```

Asserting `engineer_2` alone would have passed while the teammate stayed invisible, so this is the version that actually catches the bug.

**One finding, and I am not fixing it — it is cosmetic.** `agent_slug` caps at `AGENT_SLUG_MAX = 64`, then `mint_agent_id` appends `_2`, `_3` … to the already-capped stem, so a suffixed id can exceed the cap by a couple of characters. I checked whether anything downstream enforces a length before deciding: `is_snake_case` has no bound, and I found no length check on roster ids or workspace node names anywhere. So the cap is a soft guard on the derived stem rather than a constraint the system relies on, and exceeding it is harmless. Flagging it because "there is a cap" and "the cap holds" are different statements and only the first is true.

**Also verified rather than assumed:** the reserved set is built from the real `OPERATOR_CHANNEL` / `AGENTS_ROOT` / `DESKS_ROOT` constants instead of re-typed literals, so an upstream rename cannot silently unreserve a name — that is the right way round.

Gates: `cargo test --locked` **1995**, gated `--tests` **2978**, both clippy lanes 0, `--all-features` check 0, fmt 0. All CI lanes green.

Note for whoever reviews: CodeRabbit has been rate-limited across this entire batch and reports `pass` while having read nothing, so this PR has had no external review.

M3gA-Mind:
Post-merge review findings (PR head `0b280301`):

1. **[P1] Duplicate names now silently misroute assignments**

   `POST /team` still permits duplicate display names. Adding “Shane” twice now creates `shane` and `shane_2`; `assignee::resolve("Shane")` performs its case-insensitive ID lookup first, resolves to `shane`, and never reaches the existing `AmbiguousTeammate` safeguard. This violates the explicit no-guessing invariant and can silently send work to the wrong teammate. Reject duplicate console names or distinguish display-name resolution from explicit ID resolution.

   Relevant code: `src/server/ops/team.rs:417-424`; resolution order: `src/runtime/assignee.rs:154-183`.

2. **[P1] Reused slugs inherit the previous agent's state**

   Returning a freed stem makes deletion followed by re-addition reuse an identity, but `DELETE /team/{agent_id}` clears only the agent row and budget. Desk membership/order, inbox metadata and mail, task assignees, usage samples, and workspace authorship remain keyed by the same ID. A new teammate can therefore inherit the former occupant’s desks, inbox, work, and current-day spend; the POST response also hardcodes inbox disabled and spend zero. Preserve tombstones/non-reuse, or explicitly clean or transfer every ID-keyed resource atomically.

   Relevant code: `src/ports/types.rs:2637-2650`; deletion: `src/server/ops/team.rs:506-522`.

3. **[P1] Collision safety lasts only until the roster changes**

   The minting check only covers the current record. On a later rebuild, the builder retains overlay agents while replacing the manifest; if a new manifest agent claims an overlay slug, `build_roster` silently skips the overlay, the Team API exposes duplicate IDs, and manifest-first routes make the overlay uneditable and undeletable. A desk created later can similarly claim the agent ID or name because `create_desk` checks only other desks, and desks win assignment resolution. Enforce the shared namespace during manifest reconciliation and desk creation, or use IDs that cannot be claimed later.

   Relevant code: `src/ports/types.rs:2661-2678`; roster merge: `src/harness/mod.rs:1983-1995`; rebuild persistence: `src/runtime/builder.rs:1382-1390,1892-1903`; desk creation: `src/server/operator.rs:475-489`.

Validation performed: targeted slug/minting/route tests passed (12 total), `git diff --check` was clean, and the current-main merge preview had no textual conflicts.

Non-blocking repository-policy note: `docs/spec/runtime/api.md` is 532 lines versus the repository's 500-line Markdown cap.

Issue #666, opened 2026-08-11
workspace: two same-name uploads in one folder alias a single file on disk, and the blob route serves it with the other node's length and ETag


M3gA-Mind:
## What this is

On the filesystem backend a binary node's on-disk path is derived from the chain of node
**names**, with no uniqueness check among siblings. Two files uploaded under the same name
into the same folder therefore resolve to one file on disk — and the download route serves
that single file under whichever node was asked for, with the *other* node's
`Content-Length` and `ETag`.

## The mechanism

`src/store/fs_ops.rs:1361` — the path is the name chain:

```rust
fn physical_path(&self, company: &CompanyId, index: &HashMap<String, WorkspaceNode>, id: &str)
    -> Result<PathBuf> {
    let mut names = Vec::new();
    let mut cursor = Some(id.to_string());
    while let Some(node_id) = cursor {
        let node = index.get(&node_id)…;
        names.push(node.name.clone());
        cursor = node.parent_id.clone();
    }
    names.reverse();
    let mut path = self.bundle(company).workspace_dir();
    for name in names { path.push(name); }
```

`src/store/fs_ops.rs:1121` — `create_binary` checks the node **id** for collision and the
parent for folder-ness, and never checks whether a sibling already holds this name:

```rust
if index.contains_key(&node.id) { return Err(Conflict…); }   // id, not name
```

There is no sibling-name uniqueness check anywhere in `fs_ops.rs` or
`src/ports/workspace.rs`.

## The failure path

1. Upload `chart.png` (2 MB) to the workspace root. Node A is created; bytes land at
   `…/workspace/chart.png`. A's `size` = 2 MB, A's `sha256` = digest of A's bytes.
2. Upload a different `chart.png` (5 MB) to the same folder — the console offers no
   warning, and the store raises no conflict. Node B is created with a fresh id, and its
   bytes are written to **the same path**, overwriting A's.
3. `GET …/workspace/blob/{A}` streams the file at that path — which is now B's bytes —
   while `src/server/ops/workspace.rs` sets `Content-Length` from `A.size` (2 MB) and
   `ETag` from `A.sha256`.

So the response body is B, the length header claims A's length, and the ETag claims A's
digest. A conditional request against A's ETag will be answered as fresh while serving B's
content. Deleting either node removes the file both point at.

## Why it matters

Duplicate filenames in a folder are an ordinary thing for an operator to do, not an edge
case — and the tree UI presents both nodes as distinct files, so nothing signals that one
[...]

Issue #665, opened 2026-08-11
workspace: a valid-UTF-8 upload bypasses max_blob_mb and tree_quota_gb entirely


M3gA-Mind:
## What this is

The multipart upload route sends any file that is valid UTF-8 down the **text** path,
which is not metered. `max_blob_mb` and `tree_quota_gb` are bypassed entirely, and on
MongoDB the content lands in a `workspace_nodes` document — the 16 MB BSON limit that
GridFS was added in this very PR to escape.

## The failure path

`src/server/ops/workspace.rs:531` (`upload`) classifies on the bytes, then at `:618`:

```rust
match text_body(&mime, &bytes) {
    Some(text) => {
        company.runtime.workspace()
            .create(company.id(), &node, Some(&text))   // <- the unmetered path
            .await?;
```

The `create_binary` arm below it is the one the quota checks sit on. So:

* Upload a 500 MB `.csv`, `.log`, `.json`, `.sql` dump, or any minified asset — all valid
  UTF-8. `text_body` returns `Some`, and the upload is stored with **no** size check
  against `max_blob_mb` and **no** contribution checked against `tree_quota_gb`.
* On the MongoDB backend the text is stored in the node document's `content` field. A
  document over 16 MB is rejected by the driver, so a 20 MB text file fails on a limit
  that #553 introduced GridFS specifically to get out from under — and fails with a
  storage-layer error rather than the operator-facing over-cap message that #647 just
  added for the binary path.

The route's own doc comment explains the text/binary split as a *feature* — notes are
"diffable, backlinkable, searchable and editable" — which is a good reason to keep the
classification. It is not a reason for the classification to also decide whether limits
apply.

## Why it matters

The two knobs exist to bound what one company can put in the tree. As written they bound
only the files that happen not to be valid UTF-8, which is not a property an operator
setting `max_blob_mb` would expect to be buying. A tenant can exceed both without doing
anything unusual — server logs and CSV exports are the ordinary case, not an attack.

`#647` improved the *message* an over-cap upload gets. This is about the uploads that never
reach that check at all.

## Suggested direction

Apply the size and quota checks in `upload` before the text/binary branch, so the limit is
a property of the upload rather than of its encoding.

## Verified

Confirmed present on `main` @ `0ea4ed5`.

## Related

- #611 / #553 — the PR and issue that added this route and the GridFS backend.
- #647 — the over-cap error-message fix, on the metered path only.

M3gA-Mind:
[...]

PR #691, opened 2026-08-11
workspace: never serve a blob as a document a browser will execute (#667)

### LINKED ISSUE #667: workspace: the blob route serves a caller-supplied Content-Type inline with no nosniff — intended?
## A question about intent, not a claim that this is wrong

`GET …/workspace/blob/{node_id}` serves a stored file with a caller-supplied
`Content-Type` and `Content-Disposition: inline`, and I could not find
`X-Content-Type-Options: nosniff` or a Content-Security-Policy anywhere in the repository.
I would like to know whether that combination is intended, or whether a layer I have not
found supplies the missing header.

Filing as a question because the answer may well be "the deployment fronts this with a
proxy that sets it", in which case this should just be closed.

## What the code does

`src/server/ops/workspace.rs:437` (`read_blob`), header set at `:456`:

```rust
let mime = node.mime.unwrap_or_else(|| "application/octet-stream".to_string());
let mut response = Response::builder()
    .status(StatusCode::OK)
    .header(header::CONTENT_TYPE, mime)
    .header(header::CONTENT_DISPOSITION,
            format!("inline; filename=\"{}\"", …));
```

`node.mime` originates from the upload. `src/server/ops/workspace.rs:531` (`upload`) takes
the multipart part's declared `content_type` and passes it to `resolve_mime(&name,
declared.as_deref())`, so the value is influenced by the uploading client rather than
derived solely from the bytes.

The `inline` disposition is deliberate and documented — the console renders images in
place, and that is a reasonable thing to want.

## What I checked before filing

* `grep -rni 'nosniff|content-security-policy|X-Content-Type-Options'` over `src/` and
  `frontend/` — **no matches**.
* No `SetResponseHeaderLayer` or equivalent response-header middleware in `src/server/`.
* No nginx/proxy config in this repository that adds such headers (the only `add_header`
  found is a `Cache-Control` in a vendored `tinyjuice` config, unrelated to this service).

So as far as this repository goes, nothing supplies it. What I cannot check from here is
whether the hosting platform (the `opencompany-manager` control plane, or whatever
terminates TLS in front of a tenant container) adds it.

## The concrete path, stated precisely

The blob route is served from the same origin as the operator console, and the console's
session cookie is scoped to that origin. So:

1. A file is uploaded whose declared type is `text/html` while its bytes are markup — or a
   `.png` whose declared content type is `text/html`.
2. `read_blob` returns it as `Content-Type: text/html` with `Content-Disposition: inline`.
3. A browser navigating to that URL renders it as a document on the console's origin, with
   script executing in that origin's context and the operator's session cookie attached to
   any request it makes.

Whether that is reachable in practice depends on who can upload. The upload route requires
an authenticated principal, so this is not anonymous — it matters where an operator can be
induced to open a link to a file another, less-trusted principal uploaded, or where agent
output reaches the tree.

## The questions

1. Is `Content-Type` from upload metadata the intended behaviour, or should the type be
   derived from a sniff of the stored bytes?
2. Is there a layer above this service that supplies `nosniff` / CSP? If so this can be
   closed and it would be worth a comment in `read_blob` saying where it comes from, since
   the next reader will run the same grep I did.
3. If not — is `nosniff` on this route, or serving unknown types as
   `application/octet-stream`, the right shape? Both preserve the inline image rendering
   the route was written for.

## Verified

Code confirmed as described on `main` @ `0ea4ed5`.

## Related

- #611 / #553 — the PR and issue that added this route.

M3gA-Mind:
## Summary

`read_blob` served `node.mime` verbatim with `Content-Disposition: inline`, and
`node.mime` is the uploader's declared `Content-Type` — or `mime_guess` on a
published deliverable's filename — rather than anything derived from the bytes.
`resolve_mime` rejects only `application/octet-stream`, so any other value a
caller declares is stored and replayed.

The route is same-origin with the operator console. Its session cookie is
`SameSite=Lax` (`src/server/graphql/auth.rs:13`), which **is** sent on a
top-level navigation, and `ScopedCompany` → `CompanyAuth` → `resolve_session`
accepts it. So opening `…/workspace/blob/{id}` in a tab authenticated and
rendered attacker-authored markup as a document on the console's own origin,
with the operator's session attached to anything it then requested.

`serving_for()` now decides the response from a **closed list** instead:

| stored mime | served as | disposition |
| --- | --- | --- |
| raster image (`png`, `jpeg`, `gif`, `webp`, `avif`, `bmp`, `tiff`, `apng`, icon) or `application/pdf` | unchanged | `inline` |
| `image/svg+xml` | unchanged | `attachment` |
| anything else, **including absent** | `application/octet-stream` | `attachment` |

Every response also carries `X-Content-Type-Options: nosniff`, without which the
forced type is a suggestion rather than a decision.

Closes tinyhumansai/opencompany#667.

### Why the read path and not the upload path

Both would close the vector for *future* uploads. Only the read path closes it
for the payloads **already stored** under a mime some caller chose, and those are
the ones that exist right now. The test asserts this directly: the stored mime is
still `text/html` after the fix, and the response is inert anyway.

It also keeps the diff off `upload`/`resolve_mime`, which #665/#666 are actively
changing.

### Why SVG is its own arm rather than folded into either neighbour

An `image/*` prefix rule is the obvious way to write the allow-list and is the
wrong one — `image/svg+xml` is an XML *document* format whose `<script>` runs.
But SVG is genuinely both things at once:

- Inside an `` element it renders in the SVG spec's **secure static mode**:
  no script, no external references. The console previews it there
  (`BinaryNodeView` gates on `node.mime.startsWith("image/")`), and an ``
  will not decode SVG bytes at all without `image/svg+xml` — so the type has to
  survive.
- At the top of a tab the same bytes are a full scripted document on this origin.

Keeping the type and forcing `attachment` serves both. A single allow-list would
have had to sacrifice one of them.

### The PDF call, stated so it is cheap to overrule

I put `application/pdf` on the inline list. The case **against**, which is real:
the console has never previewed a PDF — `BinaryNodeView`'s own comment says it
deliberately shows metadata and a download instead — so excluding PDF would cost
only direct navigation, and would drop pdf.js / plugin surface from a same-origin
path. The case **for**, which I took: `inline` exists so that opening the URL
shows the file, and a browser renders a PDF in a viewer with no reach into the
embedding origin's DOM or cookies. It is a one-line change either way and the
code comment says so.

## API Or Behavior Changes

**The console is unaffected.** It never sees `Content-Disposition`: `fetchBlobUrl`
→ `client.getBlob` → `fetch` → `URL.createObjectURL`, and `fetch` ignores the
header entirely. The Download button is `<a download>` on a `blob:` URL. Image
preview is unchanged because images keep their type; SVG preview is unchanged for
the same reason. No frontend change was needed and none is included.

**What does change** is a direct browser navigation to `…/workspace/blob/{id}`
for a non-image, non-PDF payload: it now downloads under the node's own name
instead of rendering. For the media types this route actually holds — zips,
office documents, video — that is what a browser did anyway. For `text/html`,
`application/xhtml+xml` and SVG it is the fix.

**Out of scope, and named rather than half-done: there is still no
Content-Security-Policy anywhere in `src/` or `frontend/`.** That gap is
repo-wide, not specific to this route, and closing this route does not close it.
I did not build one and this PR should not be read as having addressed it.

Two things I checked and did **not** change:

- `read_blob` is the only non-test caller of `read_bytes` in `src/server/`, so
  this closes the class rather than one door.
- `task_export.rs` is the other HTML-emitting route and is clean: it escapes via
  `escape()` at `:619`, has tests pinning `&lt;script&gt;`, and already serves
  `attachment`.

## Tests

Run on `upstream/main` @ `a90f70f` with submodules populated (verified by
`Cargo.toml` existence, not `submodule status`).

- [x] `cargo build --all-targets` — ran as `cargo check --locked --all-features --all-targets`, which is strictly wider (it is also the lane that catches the MongoDB/SQLite row-construction paths neither default nor gated features compile)
- [x] `cargo test --locked` — **1986 + 10 + 1 passed, 0 failed**
- [x] `cargo test --locked --features openhuman,tinycortex --tests` — **2965 + 10 + 1 + 11 + 2 passed, 0 failed**
- [x] N/A — `--features acp`: this diff touches nothing under `src/server/acp/*`.
- [x] N/A — `--features mongodb` / `sqlite` lanes: the change is one HTTP handler and a pure function; it reads `node.mime` off a value every backend already returns and stores nothing. The `--all-features` check above still compiles both backends.
- [x] N/A — frontend: no `frontend/` file is touched, for the reason given under *Behavior Changes* (the console never sees the header that changed).

### New coverage — 7 tests, and both reverts were checked

Four over HTTP in `write_test.rs`, three unit tests on the classifier in
`workspace.rs`.

**Reaching the vector honestly.** A valid-UTF-8 `text/html` upload is stored as a
prose *note* by `text_body` and never reaches `read_blob` at all, so the HTML
payload carries one trailing `0xff` to take the binary branch. That is not a
contrivance to reach the branch: a browser substitutes U+FFFD for that byte and
runs the script identically. `image/svg+xml` and `application/xhtml+xml` are not
"texty" to `text_body`, so they reach the route with no trick whatsoever.

**Revert-and-check, in two parts so each failure is attributable.** A single
revert would not have distinguished them, and deleting `serving_for` outright
would have failed the unit tests by *compile error*, which proves nothing.

*Part A* — route reverted to `node.mime` + `inline`, classifier left in place:
**0 passed, 4 failed.** Each for its own reason, not a shared one:

```
a_blob_stored_as_html_cannot_be_served_as_a_document
    left: "text/html"   right: "application/octet-stream"
an_svg_keeps_its_type_..._never_rendered_as_a_document
    a top-level navigation must download an SVG: "inline; filename=\"logo.svg\""
an_unrecognised_stored_type_is_served_as_opaque_bytes
    left: "application/xhtml+xml"   right: "application/octet-stream"
an_image_still_renders_inline_with_its_own_type
    no entry found for key "x-content-type-options"
```

*Part B* — classifier gutted to the pre-fix passthrough rule: **0 passed, 3
failed** — `("image/svg+xml", "inline")` vs `attachment`,
`"application/x-invented-2031"` vs `octet-stream`, `"image/png; charset=binary"`
vs `image/png`.

Restored: **7 passed, 0 failed.**

**One test's revert-failure is narrower than its name suggests, so I am saying so
rather than letting the green row imply more.**
`an_image_still_renders_inline_with_its_own_type` fails under revert on the
*missing `nosniff` header* — its `image/png` and `inline` assertions were already
true before this change. It is a regression guard for #611's behaviour plus one
new assertion, and only the new assertion is what its revert-failure proves.

## Documentation

`docs/spec/runtime/ports-console.md` — the classification table, why the fix is
on the read path, why SVG is split out, that the console is unaffected, and an
explicit note that a repo-wide CSP is still absent and this does not close it.
The file is 384 lines, under the 500-line guideline in `AGENTS.md`.

## Related

- #611 / #553 — the PR and issue that added this route and its inline image
  rendering, which this preserves.
- #665 / #666 — the upload path. Untouched here on purpose.

### DIFF
--- a/docs/spec/runtime/ports-console.md
+++ b/docs/spec/runtime/ports-console.md
@@ -229,6 +229,34 @@ including a 17 MiB case that proves the BSON cap is not in play. Over HTTP:
 decode as UTF-8 is stored as a note instead, so an uploaded `.md` keeps its
 editor and backlinks.
 
+**A stored `mime` does not decide how the blob route serves it (#667).** That
+value is the uploader's declared `Content-Type`, or `mime_guess` on a published
+deliverable's filename — a claim, not a property of the bytes — and the route is
+same-origin with the console's `SameSite=Lax` session cookie, which a top-level
+navigation sends. So `read_blob` classifies against a **closed list** instead:
+
+| stored mime | served as | disposition |
+| --- | --- | --- |
+| raster image (`png`, `jpeg`, `gif`, `webp`, `avif`, `bmp`, `tiff`, `apng`, icon) or `application/pdf` | unchanged | `inline` |
+| `image/svg+xml` | unchanged | `attachment` |
+| anything else, including absent | `application/octet-stream` | `attachment` |
+
+Every response also carries `X-Content-Type-Options: nosniff`, without which the
+forced type is a suggestion. SVG is split out rather than folded into either
+neighbour because it is an image *and* a document: an `<img>` will not decode it
+without `image/svg+xml`, and inside an `<img>` the SVG spec's secure static mode
+runs no script — but the same bytes at the top of a tab are a scripted document
+on this origin. Keeping the type and forcing `attachment` serves both.
+
+This lives on the **read** path deliberately. Sanitising on upload would leave
+every payload already stored under a caller-chosen mime exploitable; classifying
+at serve time covers those too. The console is unaffected either way — it fetches
+blobs through `getBlob` and wraps them in an object URL, and `fetch` ignores
+`Content-Disposition` entirely.
+
+A repo-wide Content-Security-Policy is **not** part of this and is still absent;
+its lack is not specific to this route and closing this one does not close that.
+
 **Authorship (#326).** Every `WorkspaceNode` carries `created_by` and
 `updated_by`, both a `WorkspaceOrigin` ∈ `seed | operator | agent{id}`. `write`
 takes the author and stamps `updated_by`; `create` receives a fully-formed node
--- a/src/server/ops/workspace.rs
+++ b/src/server/ops/workspace.rs
@@ -431,9 +431,22 @@ async fn search(
 ///
 /// `ETag` is the payload's sha256 — the digest the store computed from the
 /// bytes it holds, so a conditional request is answered by the thing itself
-/// rather than by a timestamp that a rename would move. `Content-Disposition`
-/// is `inline`: the console renders images in place, and a browser opening the
-/// URL directly should show the file rather than be forced to download it.
+/// rather than by a timestamp that a rename would move.
+///
+/// # A stored `mime` is a caller's claim, so it does not decide the disposition
+///
+/// `node.mime` comes from the upload's declared `Content-Type` (or from
+/// `mime_guess` on a published deliverable's filename), which means it is
+/// influenced by whoever produced the file rather than derived from the bytes.
+/// Serving that value back with `Content-Disposition: inline` handed an
+/// uploader the ability to run script on the console's own origin, with the
+/// operator's `SameSite=Lax` session cookie attached — a top-level navigation
+/// sends it (issue #667).
+///
+/// So the mime no longer selects the disposition; [`serving_for`] does, from a
+/// closed list. Nothing about *storage* changed, which is the point: a payload
+/// already sitting in the tree under a mime some caller chose is neutralised the
+/// next time it is served, rather than only on uploads that happen after this.
 async fn read_blob(
     company: ScopedCompany,
     Path(NodePath { node_id }): Path<NodePath>,
@@ -448,19 +461,22 @@ async fn read_blob(
             "workspace blob {node_id}"
         ))));
     };
-    let mime = node
-        .mime
-        .unwrap_or_else(|| "application/octet-stream".to_string());
+    let serving = serving_for(node.mime.as_deref());
     let mut response = Response::builder()
         .status(StatusCode::OK)
-        .header(header::CONTENT_TYPE, mime)
+        .header(header::CONTENT_TYPE, serving.content_type)
+        // Without this, a browser is free to disregard the type above and sniff
+        // the bytes — which would make forcing `application/octet-stream` a
+        // suggestion rather than a decision.
+        .header(header::X_CONTENT_TYPE_OPTIONS, "nosniff")
         // Quoted per RFC 9110, and escaped defensively: a node name never
         // reaches this header, but a digest that somehow did would break the
         // response rather than the parse.
         .header(
             header::CONTENT_DISPOSITION,
             format!(
-                "inline; filename=\"{}\"",
+                "{}; filename=\"{}\"",
+                serving.disposition,
                 node.name.replace('"', "").replace(['\r', '\n'], "")
             ),
         );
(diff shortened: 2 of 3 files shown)


oxoxDev (approved):
Fixing the **read** path rather than the upload path is the call that makes this worth merging as-is. Both would close it for future uploads; only this one covers the payloads already sitting in stores under a mime some caller chose, and those are the ones that exist. Asserting it directly in the test — the stored mime is still `text/html` afterwards and the response is inert anyway — is the right way to pin that distinction, and keeping the diff off `upload`/`resolve_mime` while #665/#666 are moving there is a bonus rather than the reason.

The `SameSite=Lax` step is what makes the severity concrete rather than theoretical, and it is the step most write-ups skip: `Lax` *is* sent on a top-level navigation, so the route was reached authenticated, on the console's own origin, by opening a URL.

Checked and correct:

- **Closed list, not a blocklist.** The default arm is the safe one, so a media type invented after this was written lands in `application/octet-stream` rather than in the inline bucket.
- **Essence parsing.** Splitting on `;`, trimming and lower-casing means `image/png; charset=binary` matches and `IMAGE/PNG` does too — and every miss falls to the safe arm, so the normalisation can only cost a legitimate inline render, never grant one.
- **Absent mime** → `unwrap_or_default()` → empty string → last arm. Fails closed.
- **`nosniff` on every response**, without which forcing the type is a suggestion.
- **SVG gets its own arm** with the secure-static-mode reasoning, rather than being swept into an `image/*` prefix rule — which is the obvious way to write this and the wrong one.
- **`read_blob` is the only server-side caller of `read_bytes`**, so there is no second route serving stored payloads that this misses.

**0 major. 1 minor, 1 question, 1 nit. Approving.**

### Question — `application/pdf` is inline and is the one arm with no argument

It is the only non-raster entry in `INLINE_RENDERABLE`, and the PR's title is "never serve a blob as a document a browser will execute". PDF is a document format with its own scripting, served inline on the console's origin. I do not think it is a defect — browser PDF viewers are sandboxed in a separate process and cannot reach the embedding origin's cookies or DOM, so it is not the same-origin execution this closes — but you wrote a careful section justifying why SVG is its own arm, and this one got none. Since the table is now the precedent for what may render inline, PDF deserves the same paragraph, even if it only says "sandboxed viewer, not a document context, deliberate".

### Minor — the comment above the `Content-Disposition` escaping contradicts the code

```rust
// Quoted per RFC 9110, and escaped defensively: a node name never
// reaches this header, but a digest that somehow did would break the
// response rather than the parse.
...
node.name.replace('"', "").replace(['\r', '\n'], "")
```

`node.name` reaches the header, two lines below the sentence saying it does not. The escaping is right and load-bearing — node names come from uploads, so this is the untrusted input on this line, and stripping `"` / CR / LF is what keeps a crafted name from ending the quoted string or splitting the header. The risk is that a later reader trusts the comment and removes the `replace` calls as dead defence. Please reword it to say what it does: the name is caller-controlled, so it is escaped.

### Nit — non-ASCII filenames

A UTF-8 node name in a quoted `filename=` is outside what RFC 6266 defines and browsers vary on it. `filename*=UTF-8''<pct-encoded>` alongside the ASCII fallback is the portable form. Cosmetic — it affects the saved name, not the disposition.

**Before merging, not a finding:** the Rust lanes are still running. Approving on the code — don't merge while it is `UNSTABLE`.

PR #693, opened 2026-08-11
fix(desktop): one "This computer" connection, however many launches (#615)

### LINKED ISSUE #615: bug: a stale "This computer" connection accumulates on every desktop launch
## Summary

Every launch of the packaged desktop leaves behind a dead "This computer" connection, and they accumulate without bound.

The embedded host binds an ephemeral port, so its base URL changes on every start:

```
run 1  embedded host listening address=127.0.0.1:65145
run 2  embedded host listening address=127.0.0.1:65275
run 3  embedded host listening address=127.0.0.1:65364
```

`App.tsx:139` adds it as a connection keyed by that URL:

```ts
void embeddedHost().then((host) => {
  if (cancelled || !host) return;
  addConnection({ baseUrl: host.baseUrl, label: "This computer" });
});
```

`addConnection` is idempotent *per base URL*, so a new port is a new connection rather than the same one at a new address. After three launches the persisted registry held:

```json
[{"id":"5pnbp7zfx7w6","baseUrl":"","label":"This host",...},
 {"id":"vad0klxipf59","baseUrl":"http://127.0.0.1:65275","label":"This computer",...},
 {"id":"4g4392soz5vm","baseUrl":"http://127.0.0.1:65364","label":"This computer",...}]
```

Two identically-labelled entries, one live and one pointing at a closed port. Each stale row is probed on every launch, fails, and renders red — so the sidebar fills with indistinguishable "This computer" rows, all but one broken, and the operator has no way to tell which is current.

## Reproduction

1. Launch the packaged app, quit it, launch it again (twice is enough):
   ```bash
   ./target/release/bundle/macos/OpenCompany.app/Contents/MacOS/opencompany-desktop
   ```
2. Note the differing `address=127.0.0.1:<port>` in each run's log.
3. Read the webview's persisted registry and observe one "This computer" entry per launch:
   ```bash
   DB=$(find ~/Library/WebKit/ai.tinyhumans.opencompany -name localstorage.sqlite3 | head -1)
   python3 - "$DB" <<'PY'
   import sqlite3,sys
   c=sqlite3.connect(sys.argv[1]); c.text_factory=bytes
   for k,v in c.execute("select key,value from ItemTable"):
       print(k.decode(), v.decode('utf-16-le')[:600])
   PY
   ```

## Expected behavior

A relaunch should update the existing embedded-host connection rather than mint a new one — one machine, one "This computer" row, however many times the app restarts.

Two candidate fixes:

1. **Key the entry on instance identity, not base URL.** The instance already writes a stable `instance-id` into its data root (visible alongside `.lock` in `~/Library/Application Support/ai.tinyhumans.opencompany/`). If `oc_embedded` returned it and the registry matched on it, a port change would be an address update on a known connection. This is the sounder design — it also covers a remote host that moves address — but it means the registry entry needs an identity field.

2. **Give the embedded host a stable port.** Smaller, but only papers over the general case: any host that changes address still duplicates, and a fixed port can collide with whatever else holds it.

Either way, stale entries already persisted on existing installs need pruning, since the registry is durable.

Related: the same-origin bootstrap connection filed separately — both are reasons the desktop's connection list does not survive contact with a real launch.

CodeGhost21:
## Summary

Every launch of the packaged desktop left behind a dead "This computer" connection, and they accumulated without bound.

The embedded host binds an ephemeral port on purpose — `embedded.rs` explains why a fixed 8080 is wrong — so its base URL differs on every launch. `addConnection` recognises a host *by* its address (`findProfile` keys on `(baseUrl, defaultCompany)`), so every launch read as a first meeting: a new connection id, a new row, and the previous launch's row left behind pointing at a closed port. The registry is durable, so they piled up, all carrying the same label.

The identity needed to tell them apart already existed — `src/app/instance.rs` mints and persists `instance_id` under the data root — it just never crossed the IPC boundary.

## What changed

- **`oc_embedded` reports `instance_id`.** Read off the `AppState` before it moves into `bind`, so the console can be told *who* is listening without waiting to contact it — which matters, because the address it would contact is what changed.
- **`adoptEmbeddedHost` replaces the `addConnection` call in `App.tsx`.** It matches on identity, re-points the remembered connection at the live port, and enforces the invariant that at most one embedded connection exists at a time. Reusing the remembered id is what carries the tour state, last-read channel and mail draft across a relaunch — all keyed by connection id.
- **A different `instance_id` is deliberately not adopted.** A second data root is a different host, and re-pointing the row would merge two hosts' scoped local state — the failure the `(connection, company)` namespace exists to prevent.
- **Rows already on disk are pruned.** They predate the identity, so `embeddedProfiles` also recognises them by the signature the bug left: this client's own label for its host, at a loopback address. Narrow on purpose — a host added by hand is labelled by authority (`127.0.0.1:8080`), never with that string, and a test pins that it survives. One orphan is adopted rather than discarded.

## Behaviour changes

`ConnectionProfile` gains optional `instanceId` and `origin`; both are absent on every profile written before this, so an upgrade reads its existing store unchanged. `Connection` gains `origin`. No API surface moves.

## Verification

```
cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check
cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --no-deps -- -D warnings
cargo test --manifest-path src-tauri/Cargo.toml --locked      # 83 passed
npm run typecheck && npm run typecheck:unit && npm run typecheck:e2e
npm test                                                       # 426 passed
npx playwright test desktop-embedded-identity                  # 2 passed, against a live host
```

The new tests were checked against the pre-fix console rather than assumed to pin something. Driving the built pre-fix bundle in a real browser, the persisted registry grows over three launches:

```
after launch 1: 2 rows    after launch 2: 3 rows    after launch 3: 4 rows
  This host  ""             + This computer :65275    + This computer :65364
  This computer :65145        (distinct id)             (distinct id)
```

With the fix it holds at 2, the connection id is stable, and the address follows `:65145 → :65275 → :65364`. Both E2E tests and five of the seven new unit tests fail against pre-fix code; the other two are margin tests (don't delete the operator's own host, don't duplicate under StrictMode) meant to hold in both directions.

Not covered: a packaged `tauri build` launched twice by hand. Everything below that — real host, real browser, real persistence, real page loads — is.

## One commit is not part of this issue

`chore(desktop): refresh Cargo.lock` is unrelated but the desktop lane cannot run without it. The host crate gained `multer`, `mime_guess`, `tokio-util` and `bytes`; the desktop keeps its own lock (separate workspace, deliberately) and it was never regenerated, so `cargo test --locked` refuses before compiling anything. Reproducible on a pristine `upstream/main` — happy to split it out if you would rather it landed on its own.

Closes #615


oxoxDev (approved):
Deriving the identity from `instance_id` — which `src/app/instance.rs` already mints and persists under the data root, and which simply never crossed the IPC boundary — is the right instinct. A new identity minted for this would have been a second source of truth for "which host is this", and the whole bug is that address was standing in for identity.

Two judgement calls are better than the fix they support:

- **A different `instance_id` is deliberately not adopted.** Re-pointing the row would merge two hosts' scoped local state, which is the exact failure the `(connection, company)` namespace exists to prevent. Recognising that "same machine" and "same host" are different claims is the part that keeps this from becoming a data-mixing bug later.
- **The legacy prune is narrow on purpose.** Recognising pre-identity rows by *this client's own label at a loopback address*, with a test pinning that a hand-added `127.0.0.1:8080` survives, is the correct shape — a broad "looks local, delete it" would have quietly eaten hosts people added themselves.

Checked the invariant rather than taking the description's word: `adoptEmbeddedHost` removes every remembered embedded profile that is not the current one before reseating or adding, so "at most one embedded connection" holds by construction rather than by the caller behaving.

Also worth noting you validated the new tests against the **pre-fix** console rather than assuming they pinned something, and drove a real Playwright run against a live host. That is the step that distinguishes a test that would have caught the bug from one written after the fact to agree with the code.

**0 major on this PR. 1 merge-order requirement. Approving.**

### Conflicts with your own #692 — proven, not predicted

#692 (`fix(desktop): stop opening on a connection to the desktop's own origin`, also yours, also open) shares six files with this one. I merged the two branches locally to check rather than guess, and git reports three content conflicts:

```
CONFLICT (content): frontend/src/App.tsx
CONFLICT (content): frontend/src/connections/registry.ts
CONFLICT (content): frontend/test/unit/connection-registry.test.ts
```

Your `App.tsx` hunk at `@@ -178,15` sits inside #692's `@@ -181,12`, and neither description mentions the other.

The textual half is loud and will be handled. The part worth deciding deliberately:

1. **`adoptEmbeddedHost` still calls `addConnection` on the first-meeting path**, and that call is exactly what #692 guards with `isAddressableBaseUrl`. After the merge the guard has to cover the new path too, or the same-origin bootstrap this PR routes through `adoptEmbeddedHost` walks straight past it.
2. **Both PRs add a pruning pass over the same profile store** — #692 forgets unaddressable rows, this one prunes pre-identity embedded rows. They need to compose rather than each assume it is the only cleanup on that path.

**#692 first is the cheaper order.** It is smaller, already approved, and `adoptEmbeddedHost` replaces the very call site it edits — so this branch absorbs it in the rebase, rather than #692 having to re-derive its guard against a function that did not exist when it was written.

Worth flagging generally: two of your open PRs, same subsystem, same week, and neither names the other. Same note I have left on several of Elvin's today, so it is a team-wide gap rather than a personal one — a `git diff --name-only` across your own open branches before opening the next one catches it in seconds.

**Before merging, not a finding:** four checks still pending. Approving on the code — don't merge while it is `UNSTABLE`. Base is current, which is worth having today.

**Scope:** I read `registry.ts`, the `App.tsx` seam, the description's claims and the cross-PR merge; I did not read all twelve files.

PR #694, opened 2026-08-11
fix(store): spare in-flight uploads from the orphan sweep

### LINKED ISSUE #664: workspace: the boot orphan-blob sweep deletes a concurrent writer's in-flight upload
## What this is

The boot-time orphan-blob sweep scans the **entire** GridFS bucket and deletes any blob
whose node document is not yet present. `create_binary` uploads the blob *before* the node
document, so a concurrent writer's in-flight upload matches that description exactly.

## The two halves

**The sweep** — `src/store/mongodb.rs:2142`, called unconditionally from `MongoStore::new`
(`src/store/mongodb.rs:138`):

```rust
let live: HashSet<(String, String)> = /* every workspace_nodes doc, all companies */;
let bucket = self.blobs();
let mut cursor = bucket.find(doc! {}).await?;        // the whole bucket
while let Some(file) = cursor.try_next().await? {
    let orphan = key.map(|k| !live.contains(&k)).unwrap_or(true);
    if orphan { bucket.delete(file.id).await?; removed += 1; }
}
```

No company filter, no age filter, no upload-in-progress guard.

**The write order** — `src/store/mongodb.rs:2373`:

```rust
self.put_blob(company, &node.id, &node.name, bytes).await?;   // blob first
self.collection("workspace_nodes").insert_one(doc! { … }).await?;   // node second
```

Between those two awaits the blob exists and its node does not.

## The failure path

Any deployment where a second process starts while a first is serving:

1. Process A is mid-`create_binary` for a 40 MB upload. `put_blob` has completed; the
   `workspace_nodes` insert has not.
2. Process B boots — a rolling deploy, a restart, a second replica, or another tenant's
   container on a shared database.
3. B's `sweep_orphan_blobs` reads `workspace_nodes`, does not find A's node, and deletes
   A's blob.
4. A's insert then succeeds. The node document lands.

The company now has a workspace node whose blob 404s forever, and whose recorded `size`
still counts against `tree_quota_gb`. Neither self-heals: the sweep only deletes blobs
without nodes, never nodes without blobs.

The shared-single-DB mode documented in `CLAUDE.md` makes this worse — the sweep's `find`
carries no `company_id` filter, so one tenant's boot sweeps **every** tenant's in-flight
uploads.

## Why it matters

The sweep's own comment calls it "best-effort by design: the cost of skipping it is disk
that is already unreachable". That cost analysis is right for the blobs it was written to
catch and wrong for the ones it also catches: deleting a live upload is not reclaiming
unreachable disk, it is destroying reachable data and leaving a permanently broken node
plus quota that can never be freed.

## Suggested direction

Anything that distinguishes "abandoned" from "in flight" — an age threshold on the blob's
`uploadDate`, a company filter plus a startup barrier, or writing the node document first
and tolerating a node whose blob is still arriving.

## Verified

Confirmed present on `main` @ `0ea4ed5`.

## Related

- #611 / #553 — the PR and issue that introduced the sweep and the GridFS path.

M3gA-Mind:
## Summary

Bound the boot-time GridFS orphan sweep by blob age so a concurrent writer's in-flight upload is not reclaimed between `put_blob` completing and the workspace-node document being inserted.

The sweep now filters server-side on GridFS `uploadDate` and only examines blobs older than one hour. The existing orphan test backdates its staged blobs, while a new regression reconstructs the exposed `create_binary` window and proves the blob remains readable after a concurrent store boot.

Closes #664.

## API Or Behavior Changes

- Orphan GridFS blobs younger than one hour are left untouched because they may still be in flight.
- Genuinely abandoned blobs become eligible at the first boot occurring at least one hour after upload; on a long-lived deployment, that may be weeks later.
- No public API shape changes.

## Tests

Prior verification on this exact branch against a local MongoDB instance:

- [x] `store::mongodb` targeted suite — 26 passed.
- [x] Revert-and-check — removing the age bound made the new regression fail while the pre-existing orphan tests still passed (3 passed, 1 failed).
- [x] `git diff --check` — clean in this continuation.

PR checklist commands below are intentionally delegated to CI by the current manager rule in `WORKFLOW-RULES.md`, which prohibits local Rust compilation/test commands on the shared machine:

- [x] `cargo fmt --all -- --check` — N/A locally; CI verification required.
- [x] `cargo clippy --all-targets -- -D warnings` — N/A locally; CI verification required.
- [x] `cargo build --all-targets` — N/A locally; CI verification required.
- [x] `cargo test` — N/A locally; CI verification required.

## Documentation

No external documentation change is needed. The behavioral rationale and threshold tradeoff are documented next to the sweep and its constant.

### DIFF
--- a/src/store/mongodb.rs
+++ b/src/store/mongodb.rs
@@ -2049,7 +2049,8 @@ const BLOB_BUCKET: &str = "workspace_blobs";
 ///
 /// An hour is far longer than that window and deliberately so. The cost of
 /// being generous is bounded and dull — a genuinely abandoned blob occupies
-/// disk until the next boot an hour later — while the cost of being tight is
+/// disk until the first boot after it becomes old enough, which may be much
+/// later on a long-lived deployment — while the cost of being tight is
 /// destroying a tenant's upload. The asymmetry is the whole argument, so do not
 /// tune this down without one.
 const ORPHAN_BLOB_MIN_AGE_MS: i64 = 60 * 60 * 1000;
--- a/src/store/mongodb.rs
+++ b/src/store/mongodb.rs
@@ -2037,6 +2037,23 @@ impl crate::ports::skills_state::SkillStateStore for MongoStore {
 /// [`MongoStore::blob_filter`].
 const BLOB_BUCKET: &str = "workspace_blobs";
 
+/// How long a blob must have existed before [`MongoStore::sweep_orphan_blobs`]
+/// will treat "no node document" as *abandoned* rather than *in flight*
+/// (issue #664).
+///
+/// The window the sweep must not delete into is between `put_blob` returning
+/// and the `workspace_nodes` insert landing — normally milliseconds, but a
+/// concurrent boot only has to land inside it once to destroy a live upload,
+/// and the damage is permanent: a node whose download 404s and whose `size`
+/// still counts against the quota.
+///
+/// An hour is far longer than that window and deliberately so. The cost of
+/// being generous is bounded and dull — a genuinely abandoned blob occupies
+/// disk until the next boot an hour later — while the cost of being tight is
+/// destroying a tenant's upload. The asymmetry is the whole argument, so do not
+/// tune this down without one.
+const ORPHAN_BLOB_MIN_AGE_MS: i64 = 60 * 60 * 1000;
+
 impl MongoStore {
     /// The workspace payload bucket.
     fn blobs(&self) -> mongodb::gridfs::GridFsBucket {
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
The paragraph that makes this reviewable is **"Not fixed by reordering the writes"** — node-first would replace an invisible orphan with a node whose bytes are absent, which `fs_ops::create_binary` already rejects in as many words. That answers the first question anyone should ask of an age threshold ("why not close the window instead?"), and answers it by showing the window is the *correct* side of a trade already made elsewhere. The ordering is right; the sweep was wrong. Without that paragraph this reads as a heuristic papering over a race — with it, it is the only available fix.

The **shared-single-DB** escalation is the part I would not have expected you to catch and is the strongest reason to take this now: this is the one query in the backend carrying no `company_id`, so in shared mode a boot in one tenant's container reclaims a *different* tenant's in-flight upload. That turns a same-tenant race into a cross-tenant data-destruction path, on the mode whose isolation is application-layer only.

Checked and correct:

- **The filter is server-side**, so a young blob is never read — and your note that this also means it cannot be defeated by a `continue` added later to the loop is the right way to think about where a guard belongs.
- **`uploadDate` is written when the upload completes**, which you say explicitly. That is what makes the bound sound rather than lucky: the window being protected starts *after* the field exists, so the comparison is always against a real timestamp, and a still-streaming upload has no `files` document for `find()` to return at all.
- **An hour against a millisecond window** is generous by four orders of magnitude, and the asymmetry argument for that ("the cost of being generous is bounded and dull, the cost of being tight is destroying a tenant's upload") is stated at the constant where someone would go to tune it down.

### Minor — four checklist boxes are ticked for commands that were not run

```
- [x] `cargo fmt --all -- --check` — N/A locally; CI verification required.
- [x] `cargo clippy --all-targets -- -D warnings` — N/A locally; CI verification required.
- [x] `cargo build --all-targets` — N/A locally; CI verification required.
- [x] `cargo test` — N/A locally; CI verification required.
```

The constraint is legitimate and you were transparent about it in the prose. The rendering is the problem: a tick is what a maintainer skims, and these four say "verified" for work that was explicitly delegated. The repo's own norm is to check only what was verified in this branch or session and to leave the rest unchecked with the reason. Unchecked plus the same sentence carries exactly the information you intended and none of the claim you did not.

Worth keeping the two lines that *are* verification — the targeted `store::mongodb` suite at 26 passed, and the revert-and-check showing the new regression fails without the bound while the pre-existing orphan tests still pass. Those are the ones that matter here and they are real.

### Question — does an abandoned blob ever get collected on a long-lived deployment?

The sweep is boot-time, and the bound now defers a genuine orphan to "a later boot". On a server that runs for weeks between deploys, those two together mean abandoned blobs accumulate for weeks rather than an hour. That is very likely the right trade — disk is cheap and the alternative destroys uploads — but the description frames the cost as "occupies disk until the next boot an hour later", which reads as though a boot is due shortly. Worth stating the real bound (until the next restart, whenever that is) so nobody sizes storage against the optimistic reading.

**Before merging, not a finding:** four checks pending, and this branch is **42 commits behind** `main`. #659 also touches `src/store/mongodb.rs` — a rebase before merge is worth it today in particular, since a green run against a stale base has already been misleading on that PR.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Addressed the retention-window question in `2a00564d` and the PR description: an abandoned blob is collected only on the first boot after it is at least one hour old, which may be much later on a long-lived deployment.

I am keeping the four CI-delegated checklist boxes checked with their `N/A locally; CI verification required` annotations. The current fleet rule in `WORKFLOW-RULES.md` explicitly requires every PR checklist box to be ticked, including non-applicable items, while separately prohibiting local Rust build/test/check commands on this shared machine.

For the stale-base note, I fetched current `upstream/main` (the branch is now 62 behind / 1 ahead) and verified `git merge-tree --write-tree upstream/main HEAD` succeeds without conflicts. The only intervening `src/store/mongodb.rs` commit is the small payload-digest change from #688; the new push is running the full CI gates again.

M3gA-Mind:
@coderabbitai review

Issue #582, opened 2026-08-10
connections: two sources of truth for provider status, so one page says both connected and not connected


graycyrus:
## Symptom

The Connections page contradicts itself about the same provider. Reported in QA: *"GitHub and Gmail show as connected under 'Sign in per provider' (Composio) higher on the page, but still show 'Connect' buttons (as if not connected) under the Developer/Communication sections further down."* And separately: *"There are multiple options to connect the same type of account, duplications are confusing — it shows connected in one place and not connected in one."*

## Where it comes from

There are **two independent sources of truth** rendered on one page, and neither consults the other.

**1 — live Composio state.** `frontend/src/views/connections/ComposioSection.tsx:320` renders a "Sign in per provider" list built from the company's actual Composio connection state. This one is correct.

**2 — a hardcoded catalog.** `frontend/src/lib/connections.ts` carries a static provider list with its own category taxonomy — `"Communication"` (line 6), `"Developer"` (line 8), used at lines 39, 47, 95 and ordered at line 134. It is rendered by category further down the page, with its own Connect buttons, and it has no knowledge of what Composio reports.

So Gmail appears twice: once as connected, once as a Connect button, on the same screen.

## Why it is not just a display bug

The second list's Connect buttons are actionable. An operator who trusts the lower list will start a connect flow for a provider that is already connected — which is at best confusing and at worst produces a second credential for one account, the shape #396 already had to clean up once.

## The change

One list, one source of truth. Either:

- **Fold the catalog into the live state** — keep `lib/connections.ts` as presentation metadata only (display name, icon, category) and drive *every* connected/not-connected decision and every Connect button from the Composio state that `ComposioSection` already holds; or
- **Drop the categorised list** and let the "Sign in per provider" list carry the categories.

The first preserves the category grouping, which is worth keeping as the provider count grows.

Related: **#404** wants a provider to be an openable, manageable object. Whichever detail view that produces should also read from the single source this issue establishes — doing #404 first on top of two lists would bake the split in.

## Acceptance

- A provider appears exactly once on the page.
- Its connected state comes from one place, and matches what an agent can actually reach.
- No Connect button is offered for an already-connected provider.

Issue #404, opened 2026-08-05
connections: open a provider and manage it, the way OpenHuman already does


oxoxDev:
## What we want

OpenHuman already has the connections experience this product needs. Port it.

In OpenHuman, a connection is a thing you can open. Click a provider — Notion, GitHub — and you get its own view: whether it is connected, a connect and a disconnect action, what account it is connected as, what has moved through it, and the surrounding detail that makes the connection feel like a managed object rather than a boolean.

In OpenCompany a provider is a row with a button. There is no detail view at all: no way to see what a connection has been used for, no way to inspect it, and disconnect is not consistently offered.

## Why this is not a straight copy

OpenCompany has **three** connection systems where OpenHuman has one, and they do not agree with each other:

- **Composio** — the path agents actually use. Its OAuth runs on Composio's side; there is no local callback.
- **Remote MCP servers** — a separate per-tenant OAuth with its own console surface.
- **The native OAuth catalog** — currently inert: its credential is written and read by nothing (#396).

So the port has to decide what a "connection" means here before it can render one. A detail view that shows a native-catalog connection as healthy would be actively misleading, because nothing consumes it.

## Scope

- Open a provider from the Connections page and get its own view.
- That view states: connected or not, which account, through which system, and when it was connected.
- Connect and disconnect are both available and both work, with the disconnect saying plainly what it will and will not revoke.
- Usage — what has actually gone through this connection — shown where the data supports it. **Do not invent a number.** If per-connection usage is not recorded today, say so on screen and file what it would take, rather than rendering a plausible-looking zero.
- The view is honest about which of the three systems backs the connection, and does not present an inert one as live.

## Before building

**Read OpenHuman's implementation first** — this issue exists because that experience is already right, and the point is to inherit its decisions rather than re-derive them. The checkout is vendored at `vendor/openhuman`. Take its information architecture, its states, and its wording where they fit; diverge only where OpenCompany's three-system reality forces it, and say where and why.

## Related

Sits on top of #402, which makes the Connections page show real providers and give one answer per provider — that is the list this detail view opens from.
#403 must be respected: connect and disconnect are privileged actions, and a member who cannot change a connection should not be offered a control that will fail.
#396 covers the native catalog being inert; a detail view must not paper over it.
#316 settled one operator per connected account — the detail view is where that becomes visible.

oxoxDev:
**Related — `cluster:connections`**

A chain rather than three independent items — settle them in this order, because each constrains the next.

[...]

PR #696, opened 2026-08-11
feat(connections): make a Composio connection an object with accounts, and let one be revoked (#404)

CodeGhost21:
## Summary

Server groundwork for #404. **This does not close it** — nothing an operator sees has changed yet; the detail view itself is a follow-up. What it removes is the reason that view could not be built.

#404 reads as a console issue, but the blocker was one layer lower: the host destroyed the data a detail view needs before the console ever saw it. The vendored Composio client returns `id`, `status`, `createdAt`, `accountEmail`, `workspace`, `username` per connection (`vendor/openhuman/.../composio/types.rs`). `list_connection_states` folded all of it into `Vec<(String, bool)>`, and `GET …/composio/connections` shipped `{toolkit, connected}`. A connection reached the UI as a boolean, so there was nothing to open — and no amount of frontend work could have changed that.

Second, there was no way to disconnect a Composio connection at all. The router was four routes, none of them a DELETE, while `delete_connection` sat unused in the vendored client. The page's only Disconnect revokes the **native** credential — the one #396 established nothing reads — so the connection agents actually run on could not be revoked, and the one that could be revoked does nothing.

**What this changes**

- `list_connections_detailed` returns one row per connection. `list_connection_states` becomes a projection over it: one network call, one allowlist filter, one scrub, and the per-toolkit OR-fold the tile grid and the `connections_read` probe consume is unchanged in meaning.
- `GET …/composio/connections` gains `accounts[]` per toolkit — **additively**. `toolkit` and `connected` keep their exact previous meaning, so the tile grid and the post-authorize poll are untouched, and the detail view needs no second route and no second round-trip.
- `DELETE …/composio/connections/{connection_id}`, admin-only (#403) on the same grounds as `authorize`: revoking the account the company acts through is a decision made on the company's behalf. It joins the `a_member_cannot_change_what_the_company_reaches_the_world_as` table rather than getting a bespoke test, which is what that table is for.

**Two judgement calls worth reviewing**

*The delete guard is a boundary, not politeness.* The id is checked against this company's own filtered list before dialling. The tenant bearer **can** reach a connection whose toolkit the manifest does not grant, and every read here deliberately hides those; letting an id revoke what no read will show would demote the allowlist to a display filter. Proven end-to-end below.

*The account label is derived server-side* (email → workspace → handle, blank treated as absent), mirroring OpenHuman's `deriveConnectionLabel`, so the precedence is stated and tested once rather than re-derived in the console.

**Not in scope, deliberately**

Per-connection usage. It turns out no server work is needed: `byProvider` on the existing usage route is already keyed by toolkit slug and fed by an `OauthCall` sample per successful `composio_execute`, so the detail view can show honest call counts for Composio. Nothing meters MCP or native, and per #404 those must say so on screen rather than render a plausible zero.

#582 (two sources of truth on one page) is still open and still the sequencing risk it names: whichever list the detail view opens from should be the single source that issue establishes.

## API Or Behavior Changes

- `GET …/composio/connections` — **additive**. New `accounts[]` per toolkit: `{ id, status, connected, createdAt?, account? }`. Existing `toolkit` / `connected` fields unchanged in shape and meaning; no console change is required by this PR.
- `DELETE …/composio/connections/{connection_id}` — **new**. Admin-only. `200 {note}` on success; `404` for an id this company cannot see; `502` when Composio itself fails; `409` on a build without the `composio` feature (never `404`, which would read as "no such connection" — a claim about the company's accounts that such a build cannot make).
- Journals one `ToolAccessChanged { change: "provider_disconnected" }` on success only, attributed to the admin.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` (default build; and `--features composio` on 1.96.1, clean for this crate — the `-D warnings` failures you'll see workspace-wide are pre-existing `large_enum_variant` lints in vendored `tinyagents`)
- [x] `cargo build --bin opencompany --features composio`
- [x] `cargo test --features composio --lib composio` — **102 passed, 1 failed**

The one failure is `server::ops::composio::tests::an_admin_is_unaffected`. It is **pre-existing and environment-dependent**: it dials the real `api.tinyhumans.ai` on the *authorize* route and gets `502` instead of `409`. Reproduced identically on clean `upstream/main` with this branch's changes stashed. This branch does not touch `authorize`.

Six new tests: the account projection (label precedence, allowlist filter, `(toolkit, id)` order), fold-equivalence with the detailed rows, both disconnect guards, the grouping fold, and the route's not-in-build conflict.

**Driven end-to-end against a mock Composio backend** (real host, real magic-link auth, real router), which is what the unit tests could not prove:

- A company with two Gmail accounts renders both, distinguishable by email — the case a boolean structurally could not express.
- Slack reports `connected: false` with `status: "INITIATED"` preserved, so "pending" is distinguishable from "never set up".
- `DELETE conn-gmail-1` → `200`, gone from the next read, `gmail` still `connected: true` via the surviving account, one attributed audit entry.
- A connection outside the allowlist (`notion`, genuinely connected upstream) → `404`, and the mock's request log shows **no DELETE ever left the host**; it survives upstream.

That run also caught a bug the tests missed: both refusals originally returned `502`, because every failure mapped to `OpenCompanyError::TinyHumans`. So a locally-rejected id reported as "the provider is unreachable", and a retry — the obvious response to a bad gateway — would have failed identically forever. Fixed with a typed `DisconnectError` (`NotFound` → 404, `Upstream` → 502); the tests now assert the variant rather than the message, since the variant is what picks the status and the message assertion is exactly what let it through.

## Documentation

Rustdoc on the changed surfaces carries the reasoning: why the fold is lossy and what still reads it, why the delete guard is a boundary rather than politeness, and why the not-in-build answer is `409` rather than `404`. No `docs/` page describes this route's response shape, so none needed updating.

### DIFF
--- a/src/harness/composio.rs
+++ b/src/harness/composio.rs
@@ -263,10 +263,58 @@ fn slug_toolkit(slug: &str) -> String {
     slug.split('_').next().unwrap_or("").to_ascii_lowercase()
 }
 
+/// One connected Composio account, projected for the console (issue #404).
+///
+/// Composio models a connection as an **account**, not as a boolean: a company
+/// can hold two Gmail connections, and telling them apart is the entire point of
+/// a detail view. [`list_connection_states`] deliberately folds this down to one
+/// `(toolkit, connected)` pair per toolkit for the tile grid and the
+/// reconciliation probe, both of which only ever ask "is this provider wired".
+/// Everything that needs to *manage* a connection reads these rows instead.
+///
+/// **Non-secret projection.** Composio returns no token material on this route,
+/// and nothing here is derived from the tenant bearer. The [`id`](Self::id) is a
+/// Composio-side handle, not a credential: it is the argument
+/// [`delete_connection`] takes, and it is useless without the bearer that scopes
+/// it to this company.
+///
+/// Always compiled, so the console DTO that mirrors it (`ops::composio`) can be
+/// defined in a build without the `composio` feature — only the functions that
+/// produce these rows are gated.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ComposioConnectionRow {
+    /// Composio's connection id — what [`delete_connection`] revokes.
+    pub id: String,
+    /// Toolkit slug, normalized (`gmail`, `googlecalendar`).
+    pub toolkit: String,
+    /// Composio's raw status string (`ACTIVE`, `INITIATED`, `EXPIRED`, …).
+    ///
+    /// Forwarded verbatim rather than reduced to [`connected`](Self::connected),
+    /// because "connected: false" reads as *not set up* while an expired
+    /// connection was set up and needs re-authorizing — a different sentence and
+    /// a different button. The console maps the vocabulary; the host does not
+    /// pretend to know every value Composio may add.
+    pub status: String,
+    /// Whether this connection is usable — `ACTIVE` or `CONNECTED`,
+    /// case-insensitively, matching the vendored client's own `is_active`.
+    pub connected: bool,
+    /// When Composio recorded the connection, ISO-8601, when it says.
+    pub created_at: Option<String>,
+    /// The account label this connection acts as, when the provider published
+    /// one: the account email, else a workspace/team name, else a handle.
+    ///
+    /// Derived here rather than in the console so the precedence is stated once
+    /// and tested once — it mirrors OpenHuman's `deriveConnectionLabel`, which
+    /// is the experience this issue ports. `None` is honest: plenty of toolkits
+    /// publish no identity at all, and inventing one from the slug would render
+    /// as a fact the operator cannot check.
+    pub account: Option<String>,
+}
+
 #[cfg(feature = "composio")]
 pub use live::{
-    ComposioMetering, authorize_connect_url, composio_tools, list_catalog_toolkits,
-    list_connection_states,
+    ComposioMetering, authorize_connect_url, composio_tools, delete_connection,
+    list_catalog_toolkits, list_connection_states, list_connections_detailed,
 };
 
 #[cfg(feature = "composio")]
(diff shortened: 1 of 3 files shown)


oxoxDev (approved):
The framing is the right one and I want to note it before the findings: #404 reads as a console issue, and you established that the blocker was a layer lower — `list_connection_states` folded six fields per connection into `Vec<(String, bool)>`, so a connection reached the UI as a boolean and **no amount of frontend work could have built the detail view**. Saying plainly that this does not close the issue, and that what it removes is the reason the view could not exist, is more useful than a partial-scope claim would have been.

**The delete guard is the best thing here, and it is a boundary rather than politeness — I checked it.**

```rust
let known = list_connections_detailed(config).await?;
if !known.iter().any(|row| row.id == connection_id) {
    return Err(DisconnectError::NotFound(...));
}
```

`list_connections_detailed` carries the allowlist filter, so an id whose toolkit the manifest does not grant is refused *before* dialling — and since the tenant bearer genuinely can reach such a connection upstream, without this the allowlist would degrade from an authorization boundary to a display filter. `a refused id must be NotFound, not an upstream failure` pins exactly that distinction, and the blank-id case is refused before any network call at all.

Two smaller calls also right: `409` rather than `404` on a build without the `composio` feature, because `404` would assert something about the company's accounts that such a build cannot know; and folding the admin requirement into the existing `a_member_cannot_change_what_the_company_reaches_the_world_as` table instead of a bespoke test, which is what that table is for.

Being explicit that the one failing test is pre-existing and reproducing it on clean `upstream/main` with the branch stashed is the right way to make that claim — that is the check most people skip.

### Minor — the console's existing Disconnect now sits next to a working one, and still lies

Your own description says it: the page's only Disconnect revokes the **native** credential — the one #396 established nothing reads — "so the connection agents actually run on could not be revoked, and the one that could be revoked does nothing."

That was true before this PR and it is not a regression. What changes is that the working route now exists a few lines away, so the gap is one wiring change rather than a missing capability. Until it is wired, an operator who clicks Disconnect on a Composio provider believes they have revoked a third party's access to their data and has not — which is a worse failure than an absent button, because it is silent and it is about access revocation.

Not asking you to build the detail view here; the scoping argument is sound. Asking that the follow-up be **an issue, linked from this PR**, rather than a paragraph in a merged description. Deferred work that lives only in a merged PR body reliably disappears, and this one has a security-shaped user expectation attached to it. If wiring the existing button to `DELETE …/composio/connections/{id}` is separable from the detail view, it is worth being its own small issue so it can land ahead of the larger one.

### Question — `accounts[]` and the two-sources-of-truth risk

You flag #582 as the sequencing risk and say the detail view should open from whichever list that issue establishes as canonical. Since `GET …/composio/connections` now carries `accounts[]` additively, this PR arguably *creates* the second source it warns about — the tile grid reads the fold, a future detail view reads the rows, and both ship from one payload. That is probably fine precisely because it is one payload and one round-trip. Worth one sentence in the route's docs saying the fold is derived from `accounts[]` and must stay that way, so nobody later populates `connected` from a separate call and reintroduces the divergence #582 is about.

**Before merging, not a finding:** five checks still pending. Base is current, which is worth having today.

Issue #316, opened 2026-08-03
connections: one operator, one connected account — enforce it, and show live connection state


oxoxDev:
## Decision

**One account, one user per company.** The company has a single operator, and **only that operator's own email may be connected**. A connection therefore belongs to the company, the company's single user, and one known address — all the same thing.

That removes the question this issue was originally opened to answer. There is no ownership split to design, no "available to the company's agents" binding, no per-connection attribution, and no ambiguity about which account backs a capability. There is exactly one account it can be.

It matches how the platform already describes itself: the parent backend's instance model calls a hosted OpenCompany instance *"a one-person company"* owned by a team. And it matches what the code already stores — the OAuth callback writes `{token, account}` into the company's `SecretStore` under `oauth/{provider}`.

### 1. Enforce the identity match

Today the connect flow accepts whichever account the operator happens to be signed into at the provider — a personal address, a second Google account, anything. Under this decision that is a misconfiguration, and it should be refused at the callback rather than stored and surfaced later.

The callback already runs `extract_account` on the token response. Compare it to the company's operator address and reject a mismatch with a message naming both, so the operator understands they were signed into the wrong account rather than seeing a generic failure.

This is what makes "one account" real rather than a convention. Without it, the invariant is a comment.

### 2. A connections screen that shows live state

Providers revoke independently of us — consent withdrawn from the account's security page, a refresh token expiring, an admin tightening app access. Our stored token stays exactly where it was; the capability is gone. Today an agent discovers that mid-run and the failure surfaces nowhere an operator looks.

Since there is a single operator and a single connected account per provider, this is a small screen rather than a matrix:

- each provider, its connected address, and its **live** state — connected, needs reconnecting, never connected
- a **reconnect** action that replaces the credential in place, keeping the connection and any agent configuration intact
- a provider rejection recorded against the connection so the state is visible **before** an agent trips over it, not only after

## The one consequence to accept knowingly

With a single operator there is no second member who can re-consent. If that person's account is deprovisioned by their provider admin, nobody inside the company can restore its connections — recovery has to come through the **team** that owns the instance, since instances are team-owned even when the company has one user.

That is a fair trade for the simplicity, but it should be a known trade rather than a surprise during an outage.

## Deferred, deliberately

[...]

PR #699, opened 2026-08-11
fix(connections): one provider grid, one source of truth (#582)

### LINKED ISSUE #582: connections: two sources of truth for provider status, so one page says both connected and not connected

CodeGhost21:
Closes #582.

## The RCA differs from the issue, and it matters for review

Three of the five things #582 describes were **already fixed on main** by #599, #586 and #600. Reviewers should not re-litigate them:

- **Slug aliasing** — fixed by #599. Each tile carries an explicit `toolkit` slug and `connectionStateFor` matches across both spellings, connected-wins, including the `x`→`twitter` alias a normalizer cannot derive.
- **Dead Connect buttons** — fixed by #599. `connectRoute` picks native/composio/managed/unavailable per tile, so a button that could only 400 is no longer rendered.
- **Grid staleness after a sign-in started in the grid** — fixed. `connectComposio` re-reads `GET …/connections` after its poll.

The issue's stated cause — "a hardcoded catalog with no knowledge of what Composio reports" — is therefore not the bug. Since #316 the backend reconciles both namespaces and the grid rendered that live state. Two things actually remained.

### 1. A gate asymmetry between two routes (the root cause)

`composio_view` in `src/server/ops/connections_read.rs` discarded **all** Composio state unless the company explicitly granted the `composio` tool namespace. `GET …/composio/connections` — which the console's other provider list read — never consulted the grant.

**13 of the 21 shipped companies grant no `composio`**, so for most of them "connected here, not connected there" was the steady state, not a race. The console mirrored the split: `composioCanAuthorize` required `reach.granted` while `ComposioSection`'s `showProviders` ignored it entirely.

The grant governs whether **agents receive Composio tools**. It never governed whether a handshake completes — `resolve_tenant` reads the credential and the toolkit allowlist and nothing else — so the sign-in the gate refused to offer worked perfectly well from the surface that ignored it. Reinstating the gate on the *other* surface would have been the wrong repair: it would remove a sign-in operators use today and leave no way to connect an account before granting the namespace. So the grant is now reported (`granted` on `GET …/composio`) and rendered as a caveat beside the connected badge.

### 2. Two lists still rendered

`ComposioSection`'s tile grid and the categorised grid of eleven hardcoded tiles. #600 made this sharper rather than softer: the Composio grid is now a branded tile grid with logos, category chips and search over the backend's ~123-provider catalog, so the two were visually near-identical.

Also fixed: the header badge counted every `/connections` row while the grid drew only the eleven tiles the console had metadata for, so it could read "3 connected" above a grid showing none of them.

## The change

**Which grid survives.** The Composio grid, because it was already the better one: its membership is the backend's live catalog rather than a list compiled into the console, and #600 gave it the names, logos, descriptions and categories that made the hardcoded tiles worth keeping. It absorbs the other grid's job — reconciled status from `GET …/connections`, the native self-hosted route, and Disconnect.

- `frontend/src/lib/provider-grid.ts` (new) merges catalog + native metadata + host status into the rendered rows.
- `frontend/src/views/connections/ProvidersSection.tsx` (new) draws them.
- `frontend/src/lib/connections.ts` stops being a list and becomes native-route metadata: the ids with a `well_known` key, their Composio slugs, brand colours, and the tiles a host with no Composio falls back to.
- `ComposioSection` keeps only the credential layer it documents itself as, and notifies the page on token change so the grid re-probes.
- `reconcile` is split out of `project_connections` as a test seam — the probe is a network call with no injection point, and the removed gate lived on the far side of it, which is how it survived unasserted.

Net **−340 lines**.

## One deliberate behaviour change

**Disconnect is now offered only where `via` includes `native`.** No host route releases a Composio connection (`/composio` exposes status, token, authorize and connections, and nothing else), so the old button called `…/connections/{id}/disconnect`, blanked a secret that was never set, reported success, and left the tile connected on the next refresh. A Composio-only tile now names where the connection lives instead.

## Acceptance

- ✅ A provider appears exactly once on the page.
- ✅ Its connected state comes from one place — `GET …/connections`, which reconciles both namespaces and is what an agent's tool belt is built from.
- ✅ No Connect is offered for an already-connected provider.

## Commands run

```
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings
cargo test                                  # 2038 lib tests, 0 failed
cargo +1.96.1 check --features composio -p opencompany --all-targets
cd frontend && npx tsc -b && npm test        # 434 tests, 0 failed
# real host serving the production bundle, real browser:
PW_BASE_URL=… PW_STORAGE_STATE=… npx playwright test   # 100 passed, 8 skipped, 0 failed
```

Also verified manually against a running host with the console bundle: 13 tiles, zero duplicates, zero connected-tiles-offering-Connect, zero leftover category headings, no console errors.

## What the tests do NOT prove

**The grant-gate divergence itself is not reproduced end-to-end.** It needs a `--features composio` host *and* live Composio credentials for a company that does not grant `composio` — the e2e harness binary is a default-feature build, so `inBuild` is false and the divergence cannot arise there. It is pinned where it is decidable instead:

- `src/server/ops/connections_read.rs` — `composio_state_survives_a_company_that_does_not_grant_composio`
- `frontend/test/unit/provider-grid.test.ts` — the merge, 14 cases

`frontend/test/e2e/connections-one-list.spec.ts` asserts only the structural criteria, and says so in its header.

One existing assertion in `connection-route.test.ts` was **inverted deliberately**: it pinned `composioCanAuthorize(..., granted: false) === false`, which is the exact gate that caused the bug.

## Follow-up

#404 ("a provider as an openable, manageable object") should read from the single source this establishes. Doing it on top of two lists would bake the split in.

### DIFF
--- a/docs/modules/server/README.md
+++ b/docs/modules/server/README.md
@@ -334,6 +334,41 @@ there is **no Slack provider** at all (the backend's only Slack credential is an
 internal alerting bot), so Slack has no hosted route except Composio, which runs
 its own OAuth.
 
+### One connection status, for one console list (issue #582)
+
+`GET …/connections` is the **only** answer to "what is connected". It reconciles
+the native `oauth/{provider}` catalog with a live Composio probe into one row per
+provider, marking which namespaces reported it in `via` — and the console renders
+exactly one provider grid from it (`frontend/src/lib/provider-grid.ts`).
+
+That took removing a gate. `composio_view` used to discard the Composio half
+unless the company explicitly granted the `composio` tool namespace, while
+`GET …/composio/connections` — which the console's *other* provider list read —
+never consulted the grant. The two lists therefore disagreed by construction, not
+by timing: 13 of the 21 shipped companies grant no `composio`, so for most of
+them one screen said both "connected" and "not connected" about the same account,
+and the second list's Connect button was actionable.
+
+The grant governs whether **agents receive Composio tools**. It never governed
+whether a handshake completes — `resolve_tenant` reads the credential and the
+toolkit allowlist and nothing else, so the sign-in the gate hid worked perfectly
+well from the surface that ignored it. It is now reported (`granted` on
+`GET …/composio`) and stated as a caveat next to the connected badge, rather than
+silently deciding what the page may show.
+
+Two consequences worth knowing:
+
+- The probe now runs for any company holding a credential, not only granting
+  ones. It stays bounded by `COMPOSIO_PROBE_TIMEOUT` and still degrades to
+  `unverified` rather than to a confident "not connected".
+- `reconcile` is split from `project_connections` purely as a test seam: the
+  probe is a network call with no injection point, and the removed gate lived on
+  the far side of it, which is how it survived unasserted.
+
+No host route releases a Composio connection — `…/connections/{provider}/disconnect`
+blanks this host's own secret — so the console offers Disconnect only for a row
+whose `via` includes `native`.
+
 ## tiny.place A2A inbound + discovery (`tinyplace` feature)
 
 Behind the `tinyplace` feature the server mounts the agent-to-agent surface
--- a/frontend/ARCHITECTURE.md
+++ b/frontend/ARCHITECTURE.md
@@ -207,14 +207,25 @@ it. Responses mirror the TypeScript models in `src/lib/*` and `src/api/types.ts`
   revenue, byCategory, transactions). **Caveat:** the inference-cost component
   of spend is `0` until openhuman#4940 (as with Usage).
 
-### Connections — `src/views/ConnectionsView.tsx`, `src/lib/connections.ts`
-- OAuth catalog with connect/disconnect and connected-account state.
+### Connections — `src/views/ConnectionsView.tsx`, `src/lib/provider-grid.ts`
+- **One** provider grid with connect/disconnect and connected-account state,
+  plus the credential sections above it (MCP, inference, company key, Composio
+  token, channels).
 - **Source:** ✅ real (feature `oauth`) — `Company.connections` (GraphQL) reads
   manifest intent (`[[connection]]`) + live OAuth status; `POST
   …/connections/{provider}/start` returns the authorize URL,
   `…/disconnect` drops tokens, and `GET /api/v1/oauth/callback` completes the
   flow. Without the `oauth` feature the write routes `404 not_wired` and the
   console shows the read-only catalog.
+- **One list, one answer (issue #582).** The page used to render two provider
+  lists — `ComposioSection`'s grid off `GET …/composio/connections`, and a
+  categorised grid of eleven hardcoded tiles off `GET …/connections` — which
+  applied different rules to the same Composio state and so disagreed on screen.
+  Now: `GET …/connections` is the sole status source, the backend's Composio
+  catalog is the sole list of what can be connected, `src/lib/connections.ts` is
+  native-route metadata rather than a list, and `src/lib/provider-grid.ts` merges
+  the three into the rows `ProvidersSection` renders. `ComposioSection` keeps
+  only the credential layer.
 
 ### Domain & Email (Settings) — `src/components/domain-settings.tsx`, `src/lib/domain.ts`
 - Custom domain with generated DNS records (verification TXT, CNAME, DKIM, SPF)
--- a/frontend/src/lib/composio-catalog.ts
+++ b/frontend/src/lib/composio-catalog.ts
@@ -416,10 +416,10 @@ export function availableCategories(rows: readonly ProviderRow[]): ProviderCateg
  * cleared the chip would be the more common design and the more annoying one:
  * it throws away half of what the operator already told us.
  */
-export function filterByCategory(
-  rows: readonly ProviderRow[],
+export function filterByCategory<T extends ProviderRow>(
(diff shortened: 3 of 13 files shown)


oxoxDev (approved):
Correcting the issue's own RCA before fixing it is the right way round, and the evidence carries it: three of #582's five symptoms were already closed by #599/#586/#600, and the real cause was a gate asymmetry rather than "a hardcoded catalog with no knowledge of what Composio reports". **13 of 21 shipped companies grant no `composio`**, which turns "connected here, not connected there" from a race into the steady state for most of the fleet — that single number justifies the whole change.

The repair direction is the part I most expected to be wrong and is right. Reinstating the gate on the *other* route is the tempting symmetry, and it would have deleted a sign-in operators use today while leaving no way to connect an account before granting the namespace. Removing the gate from `composio_view` and reporting `granted` separately keeps the capability and fixes the contradiction, and the doc comment states the distinction the gate was confusing — the grant answers whether agents receive Composio *tools*, never whether a handshake completes. No new exposure either: these are the company's own connections, already visible from the route that never consulted the grant.

Keeping the Composio grid as the survivor is well argued — its membership is the backend's live catalog rather than a list compiled into the console — and the consolidation preserves what the other grid was for. `buildGridProviders` merges `[...catalog, ...nativeOnly]`, so a host with no Composio still renders the native tiles a self-hoster can genuinely use rather than an empty grid, and the `x`→`twitter` alias handling from #599 survives in the reconciliation rather than being re-derived.

**0 major. 1 finding worth routing, 1 question. Approving.**

### The Composio Disconnect now falls between this PR and #696

```ts
canDisconnect: row.connected && via.includes("native"),
```

For a Composio-only provider that is `false`, so the new grid offers no Disconnect at all. On its own that is an **improvement** — the button it replaces revoked the native credential, which #396 established nothing reads, so it claimed to revoke access and did not. Withholding a lie beats rendering one.

The problem is where it leaves the fleet once both land. **#696 (also yours) ships a working `DELETE …/composio/connections/{connection_id}`**, admin-only, with the allowlist guard. So after both merge, the capability exists on the server and is reachable from nothing: this PR removed the misleading control, that PR added the real route, and neither wires them together. An operator who wants to revoke a Composio account has no way to do it from the console.

That is exactly the seam work disappears into — each PR is individually correct and the gap only exists in their intersection, so neither reviewer of either one would naturally catch it. I raised the same thing on #696 and asked for a linked issue; doing it from this side too, since this is the PR that removes the last button. **One issue, linked from both**, ideally scoped to just wiring Disconnect on a Composio row to the new route rather than waiting on the #404 detail view.

### Question — is the `granted` caveat actionable?

The grant is now surfaced as a caveat beside the connected badge. For the 13 companies with a connection and no `composio` grant, that caveat is the whole explanation of why their agents cannot use an account the page says is connected. Does it say what to do about it — that the namespace is granted in the manifest — or does it only state the fact? A caveat that names a condition without naming its remedy tends to generate the support question it was written to prevent.

**Scope of this review:** I read `connections_read.rs`, the `provider-grid.ts` merge and route/disconnect derivation, and the cross-PR interaction. I did not read all thirteen files — `ProvidersSection.tsx`'s rendering and the E2E specs I took on the strength of the seven green checks.

**Before merging:** one check still pending, and the branch is 27 commits behind `main`. Given how much connections-area work is in flight today — #696 in the same subsystem — a rebase before merge is worth the minute.

PR #712, opened 2026-08-11
workspace: a binary node is not an empty note on GraphQL, and replaced bytes refetch (#669)

### LINKED ISSUE #669: workspace: the GraphQL workspaceFile resolver reports a binary node as an empty note, and the console never refetches replaced bytes
## What this is

`read()`'s "a binary node yields an empty body" contract was handled in the REST handler
and not in the GraphQL resolver, so `Company.workspaceFile` reports a 4 MB PNG as an
ordinary note with no content. The console has a matching hole: the binary viewer refetches
only when the node **id** changes, so a payload replaced in place is never refetched.

Filed together because they are the same omission on both sides of one contract — the
places that consume `read()` for a node that might be binary.

## Backend: the GraphQL resolver has no binary case

The REST route does handle it — `src/server/ops/workspace.rs`, `read_file`:

```rust
if let Some(mime) = &node.mime {
    return Err(ApiError(OpenCompanyError::InvalidRequest(format!(
        "`{}` holds {mime} data, not text; fetch it from \
         `…/workspace/blob/{node_id}` instead", node.name))));
}
```

The GraphQL resolver does not — `src/server/graphql/workspace.rs:154`, `resolve_file`:

```rust
let Some((node, content, backlinks)) =
    file_with_backlinks(runtime.workspace().as_ref(), runtime.id(), id).await? else {
    return Ok(None);
};
Ok(Some(WorkspaceFileGql {
    id: ID(node.id), name: node.name, content, …
}))
```

No mime check, and `WorkspaceFileGql` (`src/server/graphql/workspace.rs:92`) has **no
`mime` and no `size` field** — its seven fields are id, name, content, updatedAt,
createdBy, updatedBy, backlinks. So a GraphQL consumer receives `content: ""` and has no
field available that would let it discover the node is binary. A 4 MB PNG and an empty note
are the same response.

The comment in the REST handler explains exactly why this matters — *"the console would
render a blank editor over a file that is not blank. Naming the route that does serve it
turns a silent wrong answer into a directed one (issue #553)"* — which is the argument for
fixing the GraphQL side too.

## Frontend: the viewer never refetches replaced bytes

`frontend/src/views/WorkspaceView.tsx:1522` — `BinaryNodeView`'s fetch effect:

```tsx
useEffect(() => {
  …
  fetchBlobUrl(client, company, node.id).then(…)
  …
}, [client, company, node.id]);
```

Keyed on `node.id` only. A publish that overwrites a node in place keeps the id and changes
`node.sha256`, `node.size` and `node.updated_at_millis` — none of which are dependencies.
So an operator viewing an image while an agent republishes it keeps seeing the old bytes
until they navigate away and back.

`node.sha256` is the natural dependency: it is already on the node, it is exactly "the
bytes changed", and it is already used as the blob route's `ETag`.

## Why it matters

The GraphQL half means a second read surface silently disagrees with the first about what a
node contains — the same class of divergence `resolve_file` and `resolve_search` go out of
their way to avoid by sharing their scans with the REST routes. The frontend half means the
console can show stale content with no indication it is stale.

## Suggested direction

* Add `mime` and `size` to `WorkspaceFileGql`, and give `resolve_file` the same binary
  answer the REST route gives — or at least a field that lets a caller tell the difference.
* Add `node.sha256` to the `BinaryNodeView` effect dependencies.

## Verified

Both confirmed present on `main` @ `0ea4ed5`.

## Related

- #611 / #553 — the PR and issue that introduced the binary read contract.

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#669.

Two halves of one omission — the places that consume the workspace port's "a
binary node yields an empty body" contract for a node that might be binary.

**Backend.** The REST read route refuses a payload and names the blob route; the
GraphQL `resolve_file` passed the empty body straight through, so a 4 MB PNG and
a genuinely empty note were the same response. `resolve_file` now refuses the
same way. Discovery moves to the tree instead: `FsNodeGql` gains `mime`, `size`
and `sha256`, which the REST `FsNode` has carried since #553 — so a consumer
learns a node is binary *before* asking for text it cannot have. `size` is
`Float`, following the convention `usage` and `Approval.atMillis` set for `u64`
against GraphQL's 32-bit `Int`. `schema.graphql` regenerated.

Adding `mime` to `WorkspaceFileGql` was the alternative and is worse: the two
file DTOs are documented as differing only in timestamp shape, and it would have
left `content: ""` in the response for a client to misread.

**Frontend.** `BinaryNodeView` refetched only when the node id changed, so a
re-publish that overwrites a payload in place — keeping the id, changing the
digest — left the operator viewing and downloading stale bytes. The staleness
rule is now `blobCacheKey()` in `api/workspace.ts`, keyed on `sha256`, and the
effect depends on that. A named function rather than entries in a dependency
array because `vitest.config.ts` confines unit tests to pure functions, and an
array is the kind of thing a later edit shortens without noticing.

## API Or Behavior Changes

- **Breaking for one previously-nonsensical case:** `Company.workspaceFile(id)`
  on a **binary** node now returns a GraphQL error naming the blob route,
  instead of a `WorkspaceFile` with `content: ""`. This matches what the REST
  twin `GET …/workspace/file/{id}` has always done (400). A client that was
  reading `content` for a payload was reading an empty string, never the bytes.
- **Additive:** `FsNode` (GraphQL) gains `mime: String`, `size: Float`,
  `sha256: String`, all null unless the node is binary. `schema.graphql` is
  regenerated accordingly.
- No REST changes. No changes to the workspace port or any store.

## Tests

Commands run locally, all from the repo root:

- `cargo test --features openhuman,tinycortex` — exit 0, **3131 passed, 0 failed**, 3 ignored
- `frontend`: `pnpm typecheck` — exit 0; `pnpm build` — exit 0;
  `pnpm vitest run test/unit/workspace-binary.test.ts` — 12 passed

Each new assertion was proven to fail with its fix reverted, and every
revert-check was re-run **after** the rebase onto current `main`:

| Reverted | Observed |
| --- | --- |
| the `resolve_file` refusal | 1 ran, **1 failed** — `a payload must not resolve as a note: {"name":"hero.png","content":""}` |
| `FsNodeGql`'s `mime`/`size`/`sha256` | 2 ran, **2 failed** — `Unknown field "mime"/"size"/"sha256" on type "FsNode"` |
| `blobCacheKey` → `return node.id` | 12 ran, **1 failed** — `expected 'n-img' not to be 'n-img'` |

Second commit strengthens one of the new tests:
`a_prose_note_projects_no_binary_metadata` was named for a prose note but only
asserted against the `maya` agent **folder**. A folder can never carry a payload,
so its nulls are structural and prove nothing about the rule the console depends
on — `mime`'s presence is what routes a node to the download path, and a note is
a `File` exactly like the binary beside it. It now creates a real text node and
asserts all three stay null for it, plus `kind: "file"` so the case cannot
quietly degrade into a second folder assertion.

Pre-existing and unrelated: a full `pnpm vitest run` has 45 failures across
`tour-resume`, `desktop-bridge` and `connection-registry`, all
`window.localStorage.clear is not a function`. This branch touches none of those
files and no vitest config or setup.

- [x] `cargo build --all-targets` (N/A as spelled — covered by `cargo check --locked --all-features --all-targets`, the stricter lane, since `CompanyRecord` is built from a row in `mongodb.rs`/`sqlite.rs` and is invisible to the default and gated lanes)

## Documentation

Rustdoc on the changed surfaces carries the reasoning: `FsNodeGql`'s three new
fields explain why they are nullable and why `size` is `Float`, and
`resolve_file` documents why a binary node is refused here rather than given a
`mime` field of its own. `blobCacheKey` documents why `sha256` and not `size` or
`updatedAtMillis`. `src/server/graphql/schema.graphql` is regenerated, and being
committed is itself the read-contract documentation for this surface.

No file in `docs/` describes the GraphQL workspace field set, so nothing there
went stale; the REST/GraphQL parity this restores is the behaviour
`docs/spec/runtime/api.md` already describes.

### DIFF
--- a/frontend/src/api/workspace.ts
+++ b/frontend/src/api/workspace.ts
@@ -369,6 +369,33 @@ export async function fetchBlobUrl(
   return URL.createObjectURL(blob);
 }
 
+/**
+ * Identifies the *bytes* a binary node currently holds (issue #669).
+ *
+ * The viewer fetches a payload once and holds an object URL for it, so it needs
+ * to know when that URL has gone stale. The node id alone cannot say: a
+ * re-publish overwrites a payload **in place** and deliberately keeps the id, so
+ * an operator watching an image while an agent regenerated it kept seeing — and
+ * downloading — the old bytes until they navigated away and back.
+ *
+ * `sha256` is the answer because it *is* "these are different bytes", which is
+ * also why the blob route serves it as the `ETag`. The two obvious alternatives
+ * are both wrong in a way that shows up rarely enough to be nasty: `size` misses
+ * a same-length rewrite, and `updatedAtMillis` moves on a rename that changed no
+ * bytes at all, forcing a needless refetch of a payload that may be 60 MB.
+ *
+ * This exists as a named function rather than as entries in a `useEffect`
+ * dependency array because a dependency array is exactly the kind of thing a
+ * later edit shortens without noticing, and because the rule above is worth
+ * asserting in a test.
+ *
+ * A node with no digest falls back to the id, which is the pre-#553 behaviour
+ * and correct: nothing here can detect a change the host did not report.
+ */
+export function blobCacheKey(node: Pick<FsNode, "id" | "sha256">): string {
+  return node.sha256 ? `${node.id}:${node.sha256}` : node.id;
+}
+
 /** A human-readable size, for the metadata a binary node shows instead of a body. */
 export function formatBytes(bytes: number | undefined): string {
   if (bytes === undefined) return "unknown size";
--- a/frontend/src/views/WorkspaceView.tsx
+++ b/frontend/src/views/WorkspaceView.tsx
@@ -24,6 +24,7 @@ import { toast } from "sonner";
 import { ApiError } from "@/api/types";
 import type { OpenCompanyClient } from "@/api/client";
 import {
+  blobCacheKey,
   createNode,
   deleteNode as deleteNodeApi,
   fetchBlobUrl,
@@ -1494,6 +1495,7 @@ function BinaryNodeView({
   const [url, setUrl] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
   const isImage = (node.mime ?? "").startsWith("image/");
+  const blobKey = blobCacheKey(node);
 
   useEffect(() => {
     let revoked = false;
@@ -1519,7 +1521,10 @@ function BinaryNodeView({
       revoked = true;
       if (current) URL.revokeObjectURL(current);
     };
-  }, [client, company, node.id]);
+    // `blobKey`, not `node.id`, is what makes a re-publish visible: it folds in
+    // the digest, so bytes replaced in place re-fetch (issue #669). The rule
+    // lives in `blobCacheKey` and is asserted there.
+  }, [client, company, node.id, blobKey]);
 
   return (
     <div className="mx-auto max-w-3xl px-6 py-6" data-testid="workspace-binary">
--- a/frontend/test/unit/workspace-binary.test.ts
+++ b/frontend/test/unit/workspace-binary.test.ts
@@ -10,7 +10,14 @@
 import { describe, expect, it, vi } from "vitest";
 
 import { OpenCompanyClient } from "@/api/client";
-import { fetchTree, formatBytes, isBinary, uploadFile, type FsNode } from "@/api/workspace";
+import {
+  blobCacheKey,
+  fetchTree,
+  formatBytes,
+  isBinary,
+  uploadFile,
+  type FsNode,
+} from "@/api/workspace";
 
 function client(handler: (req: { method: string; url: string; body?: string }) => unknown) {
   const transport = {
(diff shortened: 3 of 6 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
"A 4 MB PNG and a genuinely empty note were the same response" is the bug in one line, and refusing on `resolve_file` the way the REST route already does is the right repair rather than inventing a third behaviour. Moving discovery to the tree — `FsNodeGql` gaining `mime`, `size` and `sha256`, which REST's `FsNode` has carried since #553 — means a consumer learns a node is binary *before* asking for text it cannot have, which is better than a clearer error at the same point.

Rejecting `mime` on `WorkspaceFileGql` for the stated reason is right and easy to get wrong: the two file DTOs are documented as differing only in timestamp shape, and adding it there would have left `content: ""` in the response for a client to misread — preserving the exact ambiguity being removed.

The frontend half is the one I would have missed. Refetching only on node id change is correct for a *replacement* — after #683, a shape-changing publish promotes a new node and the id moves — but wrong for the same-shape re-publish that goes through `write_binary` and overwrites in place, keeping the id and changing the digest. That is the common case, and the failure is quiet: the operator views and downloads stale bytes with no signal. Keying on `sha256` is the right axis, and putting the rule in a named `blobCacheKey()` rather than a dependency array is well justified — an array is exactly what a later edit shortens without noticing, and this way the rule is assertable.

`size` as `Float` following the `usage` / `Approval.atMillis` convention for `u64` against GraphQL's 32-bit `Int` is consistent with the codebase and fine at any real file size.

**0 major. 1 minor. Approving.**

### Minor — the no-digest fallback can do better with data this PR already adds

```ts
return node.sha256 ? `${node.id}:${node.sha256}` : node.id;
```

The doc is honest — "nothing here can detect a change the host did not report" — and the fallback is not hypothetical: #688 shipped `PayloadStorage::Stored { sha256: None }` precisely because a backend may record no digest, and says so on the artifact body. So for those nodes the stale-bytes bug this PR fixes remains, silently, in the same shape.

`size` is now in the same DTO. `${node.id}:${node.size}` as a second-choice key catches every re-publish that changes the payload's length, which is most of them:

```ts
export function blobCacheKey(node: Pick<FsNode, "id" | "sha256" | "size">): string {
  if (node.sha256) return `${node.id}:${node.sha256}`;
  // No digest reported (a backend that records none — see #688). Length is a
  // weaker witness than a hash and a much better one than the id alone: it
  // catches every re-publish that changes the payload's length.
  if (node.size !== undefined) return `${node.id}:${node.size}`;
  return node.id;
}
```

Still misses a same-length overwrite, so it does not deserve a stronger claim than that — but it is strictly better than the id, costs one line, and the test file you have added is the right place to pin which witness is used when.

**Before merging:** base is current, `MERGEABLE` / `CLEAN`, and only one lane is still running. This is the cleanest-standing PR in your open set right now.

Issue #420, opened 2026-08-05
Remaining unannounced cuts on agent-visible text (operator redirect, memory preamble, publish scan, event tail)


oxoxDev:
The tail of the #410 audit. Four smaller instances of the same class — something that reaches a model is shortened and the result does not say so. Grouped because none of them individually justifies a PR, and they share one fix shape: count what was dropped and say it.

Ranked by how badly the silence misleads.

### 1. `company::steer::cap_redirect` truncates with **no marker at all**

`MAX_REDIRECT_CHARS` is 2000 and the implementation is `instruction.chars().take(MAX_REDIRECT_CHARS).collect()` — not even an ellipsis. The value is an operator's mid-turn redirect, and it becomes the agent's instruction for the rest of the turn (applied on ingest and again in the brain). A cut redirect is byte-indistinguishable from a complete one, to the model *and* to the operator reading the transcript afterwards. This is the most purely silent cut left in `src/`. Low frequency — 2000 characters is a long redirect — but when it bites, the agent acts on a sentence that stops mid-clause and nothing attributes the failure.

### 2. `harness::memory_loop` drops retrieved prior work with no signal

`MAX_SNIPPET_CHARS` (500) clips each hit with a bare `…`, and the `MAX_HISTORY_CHARS` (2000) budget `break`s out of the loop with no marker for the hits it skipped. With `RETRIEVE_TOP_K = 5` at up to ~503 characters each, the budget is exceeded on essentially every warm turn, so the last hit is routinely dropped in silence. Worse: if every hit is over budget the whole preamble is omitted and the turn is byte-identical to a cold start — the model cannot distinguish "no prior work exists" from "prior work existed and was suppressed", which is exactly the distinction the loop exists to communicate.

### 3. `harness::publish` sets a `truncated` flag that nothing reads

[...]

PR #714, opened 2026-08-11
approvals: tell the operator when a workflow run discarded gated calls

### LINKED ISSUE #638: approvals: a workflow node's discarded gated calls are logged, never told to the operator
## Where

`src/workflows/caps/mod.rs` — `park_gated_calls`, the workflow-run counterpart of the chat cycle's `park_approval_requests`.

## What

#561 gave the chat path an operator-visible sentence when a turn gates more calls than `MAX_APPROVAL_REQUESTS_PER_TURN` allows: `drain` now returns `DrainedRequests { requests, discarded, cap }` and the cycle pushes `overflow_notice()` as its own operator bubble.

The workflow path bounds itself with the same cap and discards the same way, but the only trace is a `tracing::warn!`. An operator watching a run get gated sees the first 8 cards on the Approvals page and nothing at all about the rest — and a log line is not something an operator reads. This is the same defect #561 fixed, in the other caller.

It is genuinely less bad than the chat path was before #561: the count *is* logged, so the information exists somewhere. It is still not the operator learning anything.

## Why it is separate from #561

The workflow path has no channel to push a bubble onto — a run has no conversation, no board card, and its approvals are Approvals-page-only (#333/#379). So the fix is not "call `overflow_notice()` here"; it needs a surface decision first: a run-scoped notice on the run's own view, an extra Approvals-page entry, or something on the run report. That design question is why #561 scoped it out rather than half-answering it.

## Suggested shape

- Reuse `DrainedRequests` (or the same `overflow_notice()` wording) so the two paths cannot word this differently.
- Decide the surface: most likely the run's outcome/report, which the operator already opens.
- Test: a node whose turn gates more than the cap must produce something an operator-facing surface renders — not only a log assertion.

## Also worth deciding here

Whether `MAX_APPROVAL_REQUESTS_PER_TURN = 8` is the right number at all. #561 deliberately did not touch it; the value is a separate question from making the discard visible, and this is a reasonable place for it to live.

## Related

Follow-up from review of #625 (`approvals: tell the operator when a turn's approval requests were dropped`).

M3gA-Mind:
## Summary

A workflow node whose turn gates more tool calls than `MAX_APPROVAL_REQUESTS_PER_TURN`
allows discards the excess. Until now the only trace was a `tracing::warn!` — the
operator was never told. The chat path already says it out loud (#561,
`DrainedRequests::overflow_notice`); a run has no conversation to speak on, so this adds
the surface a run can speak on.

Closes tinyhumansai/opencompany#638

**How it works.** `RunNotices` is a cheap-to-clone shared handle owned by the run. An
agent node pushes onto it sideways rather than returning a notice, because
`AgentRunner::run_agent` hands the engine a `Value` that becomes the node's *output* — a
system notice riding that channel would land in a downstream `=item` binding and in the
run's persisted output snapshot. The run drains it onto `WorkflowRun::notices`, which is
journaled onto `WorkflowRunFinished` and rendered by `RunHistoryPanel` as an amber line
beside the outcome.

Three deliberate choices, each pinned by a test:

- **Not `error`.** A run that overflowed the cap did not fail — its nodes ran and its
  output is valid. Marking it failed would inflate the failure count and hide real
  failures among them. Same reasoning as `cancelled` being a flag.
- **Rendered alongside the outcome, not as one more branch of it.** A run can succeed, be
  stopped, *or* fail and still have discarded gated calls. Folding it into the outcome
  chain would make it invisible for every outcome but the one branch it sat in.
- **An ordering fix rides along.** The overflow branch now runs *before* the
  `parking`-is-`None` guard. That guard `return`s, so on a runtime with no approvals gate
  the discard was not reaching even the log. That is the worst case, not a corner: the
  survivors could not be parked either, so the notice is the only thing the operator can
  be told.

### Also in here: number agreement in the overflow notice

`overflow_notice` branched its *nouns* on the count but not its verbs or pronouns, so a
single discard read "1 further gated tool call **were** not raised … **They were** not
run and **they are** not on the Approvals page". This was raised on #672, which merged
with the bug still in it, so it is live on `main` today; `overflow_notice` is this PR's
surface, so it is fixed here.

Fixing only the first verb would produce "They **was** not run", so the pronouns move
with it. n=1 now reads: "…1 further gated tool call **was** not raised … **It was** not
run and **it is** not on the Approvals page." Both directions are pinned — one test that
fails if it stays plural, one that fails if the fix over-corrects the plural case.

## API Or Behavior Changes

- **`WorkflowRun` gains `notices: Vec<String>`** and `CompanyEvent::WorkflowRunFinished`
  gains the matching field. Both `#[serde(default)]`, so payloads written before the
  field existed still load — as empty, which is the overwhelmingly common case.
- **`WorkflowRunOutcome.notices?: string[]`** on the console API. Absent means "nothing
  to tell you", never "unknown". It is a warning, never an error.
- **`build_capabilities` signature**: its five per-run arguments are now bundled into a
  new `pub struct RunContext<'a>`, taking it from 8 parameters to 4. Issue #638 added the
  fifth and tipped it over clippy's arity limit; the five genuinely travel together, so a
  struct is the honest shape rather than a way of quieting the lint. Three call sites
  updated (1 production, 2 tests).
- The operator-facing overflow sentence now agrees in number for a single discard. No
  change for the plural case, which was already correct.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — run as
      `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`
      (`--no-deps` keeps pre-existing `vendor/tinyagents` lints, which CI never sees, from
      masking the result). Exit 0, zero errors.
- [x] `cargo build --all-targets` — exit 0.
- [x] `cargo test` — run as `cargo test --features openhuman,tinycortex`. Exit 0:
      **3088 passed, 0 failed, 3 ignored** in the lib target, all other targets green.

Also run: `cargo check --locked --all-features --all-targets` (exit 0). `--all-features`
matters here because `mongodb.rs` / `sqlite.rs` build `CompanyRecord` from a row rather
than a struct literal, which the default and gated lanes cannot see.

**Test count is accounted for exactly.** Lib tests at the base commit: 3085. At this
branch: 3091. A name-diff of the two `--list` outputs (not just a net count, which could
hide a rename) shows **6 added, 0 removed** — the 5 this feature adds plus 1 for the
grammar fix.

**Every new test was verified to fail with its mechanism reverted** (7 mutations; each
run exited non-zero with the rest of the set still passing, so none was a silent no-op):

| Mutation | Test that failed |
| --- | --- |
| Undo the number-agreement fix | `the_overflow_notice_is_singular_for_a_single_dropped_request` |
| Over-singularise (always "was") | `the_overflow_notice_stays_plural_for_several_dropped_requests` |
| Suppress `notices.push` | `an_overflowing_node_leaves_the_operator_a_notice` **and** `the_notice_survives_a_runtime_with_no_approvals_gate` |
| Move the notice back below the parking guard | `the_notice_survives_a_runtime_with_no_approvals_gate` (alone — the ordering fix is pinned on its own, not riding on the push) |
| Fire the notice unconditionally | `a_node_within_the_cap_raises_no_notice` |
| Ok arm drops `notices` | `a_runs_notices_reach_the_journaled_outcome` |
| Err arm carries `notices` | `an_ordinary_run_and_a_failed_run_carry_no_notices` |

### Console lane

Run in full locally, all green: `npm run typecheck`, `typecheck:e2e`, `typecheck:unit`,
`npm test` (**426 passed, 38 files**), `npm run build`, and
`./scripts/ci/assert-design-tokens.sh`.

The design-tokens gate caught a real violation in the first push of this branch, now
fixed. The notice line was written as `text-[11px] text-amber-600 dark:text-amber-500` —
an arbitrary font size *and* a raw palette colour, two of the three things that script
forbids. The gate postdates the code: it arrived on `main` while this work was in
progress, so the rebase is what brought the two together. It now reads
`text-2xs text-[var(--status-blocked-text)]` — the same named 11px rung the sibling lines
in this component already use, and the console's "needs your attention, nothing is
broken" state token, which themes for both schemes on its own rather than restating
itself in a `dark:` variant.

Note for anyone reproducing locally: the unit suite must run on **Node 22**, as CI does.
On Node 23+ it fails 45 tests across 3 files with `window.localStorage.clear is not a
function` — newer Node ships a native `globalThis.localStorage` global that collides with
jsdom's. That is a local-environment artifact only; the suite is green on `main` and
green here.

## Documentation

The new field and its constraints are documented where they are declared, which is where
a reader meets them: `WorkflowRun::notices` in `src/ports/workflow_runner.rs` carries the
reasoning for not being `error` and not being a `DeliveryReport` row, and
`WorkflowRunOutcome.notices` in `frontend/src/api/workflows.ts` carries the
render-as-warning-never-error contract. `overflow_notice` gains an `# Agreement` section
explaining why every countable word in the sentence branches on the count.

No `docs/` change: this adds a field to an existing journaled outcome and a line to an
existing panel rather than changing architecture or expected usage. No new UI string
literal is introduced — the panel renders backend-supplied text — so there is no `useT()`
key and no locale files to update.

### DIFF
--- a/frontend/src/api/workflows.ts
+++ b/frontend/src/api/workflows.ts
@@ -341,6 +341,19 @@ export interface WorkflowRunOutcome {
    * and absent must read as "not cancelled".
    */
   cancelled?: boolean;
+  /**
+   * System notices raised about this run (issue #638) — today, that a node
+   * gated more tool calls than the per-batch cap allows and the excess was
+   * discarded.
+   *
+   * NOT a failure. A run that overflowed the cap still succeeded: its nodes
+   * ran and its output is valid. Render it as a warning beside the outcome,
+   * never as an error, or a run that worked reads as one that broke.
+   *
+   * Omitted on the wire for the overwhelming majority of runs, which raise
+   * nothing — so absent must read as "nothing to tell you", not "unknown".
+   */
+  notices?: string[];
 }
 
 export function listWorkflows(
--- a/frontend/src/hooks/use-events.ts
+++ b/frontend/src/hooks/use-events.ts
@@ -144,6 +144,12 @@ export type CompanyStreamEvent =
        * just pressed Cancel that the run finished fine.
        */
       cancelled?: boolean;
+      /**
+       * System notices raised about the run (issue #638), e.g. that gated tool
+       * calls past the per-batch cap were discarded. Absent on the vast
+       * majority of runs. Not a failure — see `WorkflowRunOutcome.notices`.
+       */
+      notices?: string[];
     }
   // Issue #371/#382: the live per-node progress trail. A run announces itself,
   // then brackets each non-trigger node with a *started* frame as it begins and
--- a/frontend/src/views/workflows/RunHistoryPanel.tsx
+++ b/frontend/src/views/workflows/RunHistoryPanel.tsx
@@ -308,11 +308,18 @@ function RunHistoryRow({
 
           Deliberately not a destructive Alert: nothing failed. It is the same
           tone as the cancelled line — something happened that you need to know,
-          not something that went wrong. */}
+          not something that went wrong.
+
+          Coloured with `--status-blocked-text` rather than a palette amber:
+          that token is the console's "needs your attention, nothing is broken"
+          state, it is the one a gated call already reads as elsewhere, and it
+          themes for both schemes on its own — which the `dark:` pair it
+          replaced had to restate by hand. `text-2xs` is the same 11px rung the
+          sibling lines above use, by name. */}
       {(run.notices ?? []).map((notice, i) => (
         <p
           key={i}
-          className="text-[11px] text-amber-600 dark:text-amber-500"
+          className="text-2xs text-[var(--status-blocked-text)]"
           data-testid="workflow-run-notice"
         >
           {notice}
--- a/frontend/src/views/workflows/RunHistoryPanel.tsx
+++ b/frontend/src/views/workflows/RunHistoryPanel.tsx
@@ -300,6 +300,24 @@ function RunHistoryRow({
           Finished — this run routed no reports.
         </p>
       )}
+      {/* Issue #638. Rendered ALONGSIDE the outcome above rather than as one
+          more branch of it, because a notice is not a terminal state — a run
+          can succeed, be stopped, or fail and still have discarded gated calls
+          the operator needs to know about. Folding it into the chain would have
+          made it invisible for every outcome except the one branch it sat in.
+
+          Deliberately not a destructive Alert: nothing failed. It is the same
+          tone as the cancelled line — something happened that you need to know,
+          not something that went wrong. */}
+      {(run.notices ?? []).map((notice, i) => (
+        <p
+          key={i}
+          className="text-[11px] text-amber-600 dark:text-amber-500"
+          data-testid="workflow-run-notice"
+        >
+          {notice}
+        </p>
+      ))}
     </div>
   );
 }
--- a/src/brain/medulla/effects.rs
+++ b/src/brain/medulla/effects.rs
@@ -629,6 +629,7 @@ mod test {
             pending_approvals: Vec::new(),
             error: None,
             cancelled: false,
+            notices: Vec::new(),
         };
 
         let wired = wire_event(7, &event);
@@ -667,6 +668,7 @@ mod test {
             pending_approvals: Vec::new(),
             error: None,
             cancelled: true,
+            notices: Vec::new(),
         };
 
         let wired = wire_event(7, &event);
--- a/src/harness/orchestrator.rs
+++ b/src/harness/orchestrator.rs
@@ -3214,6 +3214,7 @@ mod tests {
             pending_approvals: Vec::new(),
             error: None,
             cancelled: false,
+            notices: Vec::new(),
         });
 
         assert!(!summary.contains(RECIPIENT), "{summary}");
@@ -3240,6 +3241,7 @@ mod tests {
             pending_approvals: Vec::new(),
             error: None,
             cancelled: true,
+            notices: Vec::new(),
         });
 
         assert!(summary.contains("stopped"), "{summary}");
@@ -4672,6 +4674,7 @@ name = "Morning"
                 deliveries: Vec::new(),
                 cancelled: false,
                 nodes: Vec::new(),
+                notices: Vec::new(),
             })
         }
     }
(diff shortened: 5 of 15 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
The reason the notice goes sideways rather than out through the return value is the part worth having in the doc comment, and it is there: `AgentRunner::run_agent` hands the engine a `Value` that becomes the node's *output*, so a system notice returned normally would ride into a downstream `=item` binding and into the run's persisted output snapshot. That is a subtle, expensive bug avoided by construction rather than by remembering, and "a system notice is emphatically not node output" is the right way to say it.

`RunNotices` is the right shape for it — `Arc<Mutex<Vec<String>>>`, `push` appends, `take` drains with `mem::take`. A `std::sync::Mutex` is correct here for the same reason `workspace_failures` uses one: both critical sections are a single `Vec` operation with no `await` inside, so an async lock would buy nothing and cost a scheduling point on the run path.

The three choices are each argued from the consequence rather than from taste, and I agree with all three:

- **Not `error`.** A run that overflowed the cap did not fail — inflating the failure count would hide real failures among them. Same reasoning as `cancelled` being a flag rather than an outcome.
- **Beside the outcome, not a branch of it.** A run can succeed, be stopped, *or* fail and still have discarded gated calls; folding it into the outcome chain makes it invisible for every branch it does not sit in.
- **The ordering fix is the real bug in here.** The overflow branch running *after* a `parking`-is-`None` guard that `return`s meant the notice was lost outright on a runtime with no approvals gate — the operator silently uninformed in exactly the configuration where nothing else would tell them either. Finding that while building the surface is the kind of thing that justifies writing the surface.

Checked and clean: the branch merges against current `main` with **no conflicts**, despite being 32 commits behind — and #672, which this builds on, landed inside that range, so the `DrainedRequests` shape it depends on is the merged one.

### Question — `notices` is unbounded, on a path whose subject is a bound

Every overflowing agent node appends a string, the list is journaled onto `WorkflowRunFinished`, and nothing caps it. It is bounded in practice by node count, so this is not a leak — but the event is persisted, and this repo has already had to add a size guard to a journaled payload for the same reason (#420's insight-doc guard), while the notice itself exists because #561 decided an unbounded set of approval requests needed a cap.

A run with many agent nodes each overflowing writes a durable event that grows with the workflow's size. Two cheap answers, either fine: cap the list and let the last entry say "and N more nodes discarded calls", or state at `RunNotices` that node count is the bound and it is small by construction. What I would avoid is leaving the one unbounded collection in a change whose whole subject is what happens when a bound is exceeded — that is the shape a later reader will find surprising.

**Before merging:** the branch is `CLEAN` and merges without conflict, but it is 32 commits behind. Given that today has twice produced a green run against a base that had moved underneath it, a rebase before merge is worth the minute even with the conflict check clean.

PR #715, opened 2026-08-11
approvals: make always_approve gate what it claims, and default it to empty

### LINKED ISSUE #684: approvals: the default always_approve list matches tool names against dotted effect kinds, so it gates nothing
## The default `always_approve` list matches nothing, so it gates nothing

`always_approve` is the operator's override — the one list that, per `policy.rs:123`, *"wins over everything else, including Full autonomy"*. It is the mechanism an operator is told to reach for when they want a specific effect to always stop for a human.

**The shipped default is inert.**

The list is matched against the **tool name**. The default (`src/company/types.rs:52`) is:

```rust
pub const DEFAULT_ALWAYS_APPROVE: &[&str] = &["payment.send", "filing.submit", "external.publish"];
```

Those are **dotted effect kinds, not tool names**. None of them matches any tool in the declaration table, and `external.publish` appears nowhere else in the codebase at all.

## Why this matters more than a dead default

Every company shipping the default believes **payments and publishing are gated. They are not.** The list is the thing an operator is pointed at when they ask "how do I make sure it never sends money without asking" — and the answer currently returns a list that cannot match.

It also undercuts the reasoning in **#658**, where the argument that `full` may publish unattended rests on "the operator can always write `always_approve`". That is only true if `always_approve` works.

## What the fix has to settle

1. **Which namespace is authoritative** — tool names, or effect kinds? The matcher and the default disagree, and one of them is wrong. `EffectGroup` already exists (`Send`, `Publish`, `Spend`, `Other`) and the parked `Effect` carries a dotted `kind`, so an effect-kind match is expressible; a tool-name match is what the code does today.
2. **Whether the default should be non-empty at all.** A default that silently does nothing is worse than an empty one, because it reads as protection. If the intended entries cannot be expressed today, ship `[]` and say so.
3. **A test that fails if the default stops matching.** This is the shape that let it ship: nothing asserts the default list resolves to anything. Whatever the fix, pin each default entry to the thing it is supposed to gate — a list of strings nobody validates will drift again.

## How it was found

Reviewing #658 (should `full` publish without asking). The answer there is "yes, because the operator has an override" — which prompted checking whether the override actually fires. It does not, for the shipped default.

Related: #658, #338, #558, #560.

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#684.

`always_approve` is the operator's override — the list that wins over every
tier, `full` included. The shipped default was
`["payment.send", "filing.submit", "external.publish"]`, and it did not do what
it read as doing.

**The issue's premise is right about the harm and wrong about the mechanism, in
a way that changes the fix.** There are **two** consumers of `always_approve`,
not one, and they were matching different things under different rules:

| path | where | matched against | rule |
|---|---|---|---|
| native effect (hosted / sidecar brain emits an effect frame) | `ManifestApprovalGate::evaluate` | `Effect.kind` | exact only |
| harness tool call (an openhuman agent calls a tool) | `ApprovalPolicy::always_requires_approval` | tool name | exact **or leading dotted segment** |

So the default was **not** globally inert — `gate.rs`'s own test
`always_approve_overrides_full` proves `payment.send` parks a `payment.send`
effect today. What is true is narrower and still bad: it was inert on the
**harness** path, which is the path every company using the openhuman toolbelt
actually runs, and the path #658's ruling was about (`always_approve =
["publish_artifact"]` — a tool name).

Two further defects the issue did not name, both found on the way in:

- **The matchers disagreed.** `always_approve = ["payment"]` parked
  `payment.send` on the harness path and silently did nothing on the native one.
  One operator list, two answers, depending on which brain was running.
- **The spec contradicted itself.** `docs/spec/runtime/manifest.md:198` said
  "`always_approve` lists effect kinds"; line 257 told the operator to write
  `always_approve = ["web_search"]` — a tool name.

## 1. Which namespace is authoritative

**Neither, because there was never a conflict to resolve — there is one
namespace and it was being read twice.**

`ApprovalPolicy::effect_for` projects a flagged tool call onto an `Effect` by
making **the tool name the effect kind, verbatim**. A tool name already *is* an
effect kind: the single-segment, undotted case of one. `publish_artifact` and
`payment.send` are the same syntax at different segment counts.

So the fix is not to pick a winner and delete the loser. Both call sites now
read one matcher — `policy::always_approve::matches` — with one rule: exact, or
a **leading dotted segment** (`payment` gates `payment.send`, never
`payroll.export`; the segment boundary is what stops a bare `starts_with` from
gating a capability the operator did not name). Both syntaxes keep working
because they were always one syntax. No half-supported second syntax is left
behind.

The rejected alternative was an explicitly tagged syntax (`kind:payment.send` vs
a bare tool name). It validates better, but it turns every operator list of
dotted kinds into a load-time validation error, which bricks a running company
on upgrade. Not worth it for a distinction the type system already collapses.

## 2. Should the default be non-empty

**No. It ships `[]`.**

- **Two of the three named capabilities this product does not have.** The
  declaration table has no payment tool and no `Sign`-group tool at all, and
  outside test code nothing emits either kind. A default cannot gate a
  capability that does not exist.
- **The third must not be defaulted.** The real name behind `external.publish`
  is `publish_artifact`, and **#658 ruled that `full` publishes unattended** —
  an operator who wants otherwise writes `always_approve =
  ["publish_artifact"]`. Defaulting it would overturn that ruling silently, in
  a PR about a different bug.
- **It costs no protection that was ever real.** The default mode is
  `supervised`, and `evaluate_supervised` already parks every `Spend`, `Sign`
  and `Publish` effect on its own. The three entries added nothing to the
  default configuration.

## ⚠️ Behaviour change on next load — and it runs the *opposite* way to the obvious one

The expected risk here was widening: make an inert default match, and companies
start parking calls they never parked. **That is not what happens**, because the
default was only inert on one of the two paths. Shipping `[]` is a
**narrowing**, and it is the thing to review hardest:

- `readonly` / `supervised` / `auto` — **no change whatsoever.** The supervised
  taxonomy parks `Spend`/`Sign`/`Publish` on its own, and `auto` falls to
  `gate.rs`'s `_ => RequireApproval` arm.
- **`full` + a hosted or sidecar brain — a real loss.** A company whose brain
  emits an effect kind spelled *exactly* `payment.send`, `filing.submit` or
  `external.publish` will **stop parking it** on next load.

Argued rather than waved through: the protection removed was arbitrary even
where it fired — three exact spellings out of a namespace the Medulla wire spec
types as a free `string(1..64)`, so `payment.transfer` or `pay.send` were never
gated — and `full` means "unattended except what the operator listed", which an
operator who never wrote a list never did. Nothing in this repo emits those
kinds. **The honest limit: I cannot see what Medulla emits in production, so the
loss is speculative rather than zero.**

If a maintainer weighs that unknown differently, the alternative is
`["payment", "filing", "publish"]`, which with the new leading-segment matcher
gates strictly *more* natively than the old default (`payment.*`, not just
`payment.send`) and narrows nothing. Its cost is that it stays inert on the
harness path — reproducing the exact confusion #684 is about. I did not take it
for that reason, but it is a one-line change.

## 3. Why operator-authored entries are not registry-validated

The first implementation attempted to fail company loading when an
`always_approve` entry named neither a declared tool nor a
consequence-classified effect. Inspection after rebasing found that check was
unsound, so it is deliberately **not** part of this PR.

- **Tool names are closed.** `declared_tools()` enumerates every tool the crate
  can wire.
- **Effect kinds are open by specification.**
  `docs/spec/integrations/medulla.md` types the wire field as
  `"string(1..64)"`, and `effect_from_frame` copies `frame.kind` verbatim. A
  hosted brain can emit a legitimate custom kind this repository has never
  seen.

Most importantly, the native gate checks `always_approve` **before**
`EffectGroup`. An otherwise-unclassified custom kind can therefore be exactly
the one an operator meant to stop. Using `effect_group_for` as a registry would
turn that working fence into a load error; it also would not reliably catch
typos because the classifier uses broad substrings such as `pay` and `sign`.

> The breadth of those substrings is a live defect in its own right —
> `effect_group_for` (`src/brain/medulla/effects.rs:459`) classifies
> `design.review` as `Sign` and `paymnt.send` as `Spend`. It is **adjacent to
> but not part of** this PR, which only stops depending on the classifier for
> validation; it is being filed separately. Nothing here changes
> `effect_group_for`.

The shipped default is different from open operator input: this repository owns
it and can require executable proof. The drift test therefore requires every
future default entry to add an explicit intended target and prove that target is
a declared tool the shared matcher gates. The list is empty today, and the test
pins that as a deliberate decision.

## Tests that asserted the default's *contents* rather than its *effect*

These passed before this PR and would have passed after a fix that changed
nothing. This is the shape that let the defect ship, so all three are converted:

- `src/company/manifest.rs` `defaults_are_prosumer_safe` — asserted the
  three-string vector verbatim. Now asserts emptiness, with the comment naming
  `supervised` as the thing that actually makes the defaults safe.
- `src/company/manifest.rs` template test — asserted only
  `!always_approve.is_empty()`. A list's length says nothing about whether it
  fires; it now proves the template fence matches `publish_artifact` on the
  harness path the template uses.
- `src/policy/gate.rs` test helper — seeded its fence from
  `DEFAULT_ALWAYS_APPROVE`, so ~27 tests depended on a constant they do not own.
  Replaced with an explicit `FENCE` fixture. A test that borrows a shipped
  default tests the default, not the mechanism.

## Also fixed: the shipped template carried the same false fence explicitly

`companies/signals_opportunity_studio/company.toml:73` did not inherit the
default — it wrote all three entries out. Fixing only the constant would have
left the flagship template advertising a fence that gates nothing. It now reads
`always_approve = ["publish_artifact"]`: a real name, behaviour-neutral today
(`supervised` already parks every publish), and a shipped example of the correct
namespace.

## API Or Behavior Changes

- **`DEFAULT_ALWAYS_APPROVE` is now `&[]`** (was three dotted kinds). See the
  behaviour-change section above — it is a narrowing under `full` + a hosted
  brain, and a no-op under every other mode.
- **Operator-authored effect kinds remain open-ended.** There is no new
  registry-based manifest rejection; a hosted brain may emit a valid kind this
  repository does not know in advance.
- **The native-effect gate now honours leading-segment entries.** `["payment"]`
  parks `payment.send` natively, where before it parked nothing. This widens
  that arm, in the fail-safe direction, and makes it agree with the harness.
- **Matching is ASCII-case-insensitive and trims surrounding space** on both
  paths. Tool names reach the harness from openhuman's registry and
  `consequence_of` already lowercases defensively before consulting the
  declaration table, so a case difference would classify correctly and then slip
  the override — the same silent-miss class as this issue.
- New public module `crate::policy::always_approve` with the shared `matches`
  rule.

## Tests

All four lanes were run locally on the final head, after the rebase onto current
`main`:

- [x] `cargo fmt --all -- --check` — passed.
- [x] `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` — passed, zero lints. (`--no-deps` keeps pre-existing `vendor/tinyagents` lints, which CI does not see, out of the run.)
- [x] `cargo build --all-targets --features openhuman,tinycortex` — passed. Also `cargo check --locked --all-features --all-targets`, which is the lane that sees the `store::mongodb` / `store::sqlite` row-built `CompanyRecord` paths.
- [x] `cargo test --features openhuman,tinycortex` — **3089 passed, 0 failed, 3 ignored** (lib), plus 10 / 1 / 11 / 2 / 0 across the five other targets. No failures.

### Every added test was checked by reverting its fix

A test that cannot fail proves nothing, so each mechanism was reverted
individually and the suite re-run. Counts are quoted so a run that silently
selected nothing is distinguishable from a pass:

| reverted mechanism | result | test that caught it |
|---|---|---|
| segment boundary → bare `starts_with` | 16 ran, 1 failed | `a_prefix_that_is_not_a_whole_segment_does_not_gate` |
| matcher → exact-only | 16 ran, **5** failed | `a_leading_segment_gates_everything_under_it`, `both_approval_paths_agree_…`, +3 harness tests |
| case-insensitivity and trim dropped | 16 ran, 1 failed | `case_and_surrounding_space_do_not_defeat_the_override` |
| `gates` empty-entry guard dropped | 1 ran, 1 failed | `an_empty_entry_gates_nothing` |
| default restored to the three kinds | 16 ran, 1 failed | `every_default_entry_names_a_declared_target_and_the_default_is_empty` |
| default restored to the three kinds | 1 ran, 1 failed | `defaults_are_prosumer_safe` |
| native gate → exact-only compare | 16 ran, 1 failed | `both_approval_paths_agree_on_the_same_always_approve_list` |
| template's false fence restored | 1 ran, 1 failed | `signals_opportunity_studio_template_passes_lint` |

#### One of these tests was vacuous as first written, and this PR says so

`an_empty_entry_gates_nothing` **failed its revert-check**: it went on passing
with `gates`'s `entry.is_empty()` early return deleted — 16 ran, 0 failed. It
was not testing the guard at all.

The reason is the failure mode worth naming, because it is not obvious from
reading the test. Its assertions ran an empty entry against
`publish_artifact`, and for that target an empty entry already falls out of
both arms of `gates` unaided: `"".eq_ignore_ascii_case("publish_artifact")` is
false, and the segment arm reads `target.as_bytes()[0]`, which is `p`, not `.`.
The assertion was routed *through* the matcher and satisfied by the matcher's
own behaviour, so it pinned the stated contract while never reaching the
mechanism named in its own doc comment.

The empty **target** is the only input the guard actually decides — it is the
one case where the exact arm would otherwise return true, since `"" == ""`.
That case is now asserted, and the test fails when the guard is removed: **1
ran, 1 failed**. The guard is load bearing, and a whitespace-only line in an
operator's list cannot gate a blank effect kind.

Recorded rather than quietly repaired, because the test as first written is the
same shape as the defect this PR fixes: something that reads as protection,
passes, and checks nothing. A green assertion is not evidence until it has been
shown it can fail.

## Documentation

- `docs/spec/company-brain/approvals.md` — new section on what an entry names,
  why there is one namespace, why open effect kinds cannot be registry-validated,
  and why the default is empty.
- `docs/spec/runtime/manifest.md` — resolved the tool-name-vs-effect-kind
  contradiction between the `[policy]` prose and the `web_search` example, and
  updated the sample block.
- `src/policy/always_approve.rs` module docs carry the argument in full.

### DIFF
--- a/companies/signals_opportunity_studio/company.toml
+++ b/companies/signals_opportunity_studio/company.toml
@@ -70,7 +70,15 @@ enabled = true
 
 [policy]
 mode = "supervised"
-always_approve = ["payment.send", "filing.submit", "external.publish"]
+# `always_approve` names the thing to gate the way the runtime names it: a tool
+# name, or a dotted effect kind a brain emits. This template used to list
+# `payment.send` / `filing.submit` / `external.publish`, none of which is a tool
+# this company can run — so the fence it advertised gated nothing on the harness
+# path (issue #684). `publish_artifact` is a real one, and it is the entry worth
+# demonstrating: under `supervised` the gate already parks every publish, so
+# this changes nothing today and keeps holding if the studio is ever moved to
+# `auto` or `full`.
+always_approve = ["publish_artifact"]
 auto_approve_under_usd = 1.0
 
 [place]
--- a/docs/spec/company-brain/approvals.md
+++ b/docs/spec/company-brain/approvals.md
@@ -347,24 +347,25 @@ a leading dotted segment (`payment` gates `payment.send`, never
 `payroll.export`). They previously held two matchers with two rules, so one
 operator list meant different things depending on which brain was running.
 
-An entry must **resolve**: name a declared tool, or carry a consequence word so
-that a brain emitting it produces a gateable group. One that does neither stops
-the company loading, because a fence that can never fire is worse than no fence
-— it reads as protection. That check is exact for tool names (a closed set) and
-heuristic for effect kinds (open by specification: the Medulla wire types
-`kind` as a free string, so the brain names its own effects and there is no set
-to check against).
+Operator-authored entries are intentionally not restricted to the declared tool
+table. Native effect kinds are open by specification: the Medulla wire carries
+`kind` as a free string, and a hosted brain may emit one this repository has
+never seen. The gate checks `always_approve` before its `EffectGroup` fallback,
+so even an otherwise-unclassified custom kind can be exactly the one the
+operator meant to stop. Treating the classifier as a registry would turn a
+working fence into a load error.
 
-That limit is why **the default is empty**. A default the runtime cannot prove
-will ever fire must not ship as though it were protection: the shipped default
-used to be `payment.send` / `filing.submit` / `external.publish`, none of which
-named a tool, so every company running it believed payments and publishing were
-gated on the harness path and none were. Two of the three name capabilities the
-product does not have; the real name behind the third is `publish_artifact`,
-which must not be defaulted because `full` publishing unattended is the ruling
-on issue #658. Under the default `supervised` mode the checkpoint taxonomy
-parks every `Spend` / `Sign` / `Publish` effect anyway, so the empty default
-costs no protection that was ever real.
+The **default is empty** because shipped defaults can and should meet a stricter
+standard than open operator input. The old default was `payment.send` /
+`filing.submit` / `external.publish`; none named a tool, so every company using
+the harness path believed payments and publishing were gated and none were. Two
+of the three name capabilities the product does not have; the real name behind
+the third is `publish_artifact`, which must not be defaulted because `full`
+publishing unattended is the ruling on issue #658. Under the default
+`supervised` mode the checkpoint taxonomy parks every `Spend` / `Sign` /
+`Publish` effect anyway, so the empty default costs no protection that was ever
+real. A drift test requires every future default entry to name its intended,
+declared tool target explicitly.
 
 ### Precedence at the tool gate
 
--- a/docs/spec/company-brain/approvals.md
+++ b/docs/spec/company-brain/approvals.md
@@ -335,6 +335,37 @@ with actor and timestamps, but each admitted call writes no journal record.
 Defensible because a standing grant only ever admits tools declared grantable
 and never a priced call; a per-use record is additive later.
 
+### What an `always_approve` entry names (issue #684)
+
+One namespace, not two. An entry is an **effect kind**, and a **tool name is an
+effect kind** — the harness projects a flagged tool call onto an `Effect` by
+making the tool name the kind verbatim, so `publish_artifact` is the
+single-segment case of the same thing `payment.send` is the two-segment case of.
+Both approval paths — the native-effect gate and the harness tool policy — read
+one matcher, `policy::always_approve::matches`, which matches the exact entry or
+a leading dotted segment (`payment` gates `payment.send`, never
+`payroll.export`). They previously held two matchers with two rules, so one
+operator list meant different things depending on which brain was running.
+
+An entry must **resolve**: name a declared tool, or carry a consequence word so
+that a brain emitting it produces a gateable group. One that does neither stops
+the company loading, because a fence that can never fire is worse than no fence
+— it reads as protection. That check is exact for tool names (a closed set) and
+heuristic for effect kinds (open by specification: the Medulla wire types
+`kind` as a free string, so the brain names its own effects and there is no set
+to check against).
+
+That limit is why **the default is empty**. A default the runtime cannot prove
+will ever fire must not ship as though it were protection: the shipped default
+used to be `payment.send` / `filing.submit` / `external.publish`, none of which
+named a tool, so every company running it believed payments and publishing were
+gated on the harness path and none were. Two of the three name capabilities the
+product does not have; the real name behind the third is `publish_artifact`,
+which must not be defaulted because `full` publishing unattended is the ruling
+on issue #658. Under the default `supervised` mode the checkpoint taxonomy
+parks every `Spend` / `Sign` / `Publish` effect anyway, so the empty default
+costs no protection that was ever real.
+
 ### Precedence at the tool gate
 
 A tool call is decided in this order:
@@ -449,8 +480,10 @@ Prosumers adjust the fence in plain language, which compiles to policy:
 - "Auto-approve spending under $5" → `auto_approve_under_usd = 5.0`
 - "Never contact my customers directly" → `never_do` → `Deny` on
   `dm.external` matching the customer list
-- "You can post to the blog without asking" → remove `external.publish`
-  from `always_approve` for that channel
+- "You can post to the blog without asking" → remove `publish_artifact` from
+  `always_approve` for that channel. Nothing to remove unless the operator put
+  it there: `always_approve` defaults to empty, and under `supervised` it is
+  the checkpoint taxonomy — not the list — that parks a publish
 
 Standing-rule changes are themselves Charter edits with provenance and audit
 ([charter.md](charter.md)); loosening a rule takes effect for *future*
--- a/docs/spec/runtime/manifest.md
+++ b/docs/spec/runtime/manifest.md
@@ -71,9 +71,8 @@ search_daily_calls = 200           # per-company daily web_search cap (0 = pause
 
 [policy]                           # see company-brain/approvals.md
 mode = "supervised"                # readonly | supervised (default) | auto | full
-always_approve = ["publish_artifact"]   # default []; names a tool or an effect
-                                   # kind — see approvals.md. An entry that
-                                   # matches neither is a validation error
+always_approve = ["publish_artifact"]   # default []; names a tool or an open
+                                   # effect kind — see approvals.md
 auto_approve_under_usd = 1.0
 
 [place]                            # see company-as-agent/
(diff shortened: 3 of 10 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
The analysis is the strongest part and it corrects the issue rather than implementing it. #684 says the default is inert; you established it is inert on **one of two** paths — `gate.rs`'s own `always_approve_overrides_full` proves `payment.send` parks a native effect today — and that the two consumers were matching different things under different rules. That reframing is what turns "delete a dead default" into "one namespace read twice", and it is why the fix is a shared matcher rather than a winner.

"A tool name already *is* an effect kind: the single-segment, undotted case of one" is the sentence the whole PR rests on, and it is right. `effect_for` makes the tool name the kind verbatim, so `publish_artifact` and `payment.send` are one syntax at different segment counts. Unifying on exact-or-leading-dotted-segment keeps both working with no half-supported second syntax, and the segment boundary is what stops `payment` from gating `payroll.export`. Rejecting the tagged `kind:` syntax because it would turn every existing dotted list into a load-time error and brick a running company on upgrade is the right trade for a distinction the type system already collapses.

Declining to registry-validate operator entries, and saying plainly that the first attempt was unsound because effect kinds are open by specification (`string(1..64)`, `effect_from_frame` copies verbatim) while tool names are closed, is better than shipping a validator that rejects a hosted brain's legitimate custom kind.

**0 major. Approving — with data on the one question you left open.**

### The `full` + native narrowing: bounded, and smaller than you could see

You flagged this as the thing to review hardest and stated the honest limit — "I cannot see what Medulla emits in production, so the loss is speculative rather than zero". I checked the half that is visible from here, and it is narrower than that:

- **21 company manifests ship in this repo. Exactly one sets `[policy] mode = "full"`, and it is `e2e_harness`** — a test fixture, not a tenant.
- `signals_opportunity_studio`, the only company declaring `always_approve` and the one this PR edits, is `supervised`, where you have shown there is no change at all.

So across every manifest in the tree, the narrowing has **zero exposure**: the tiers that shipped are the ones you proved unchanged. That does not close your question — a provisioned hosted tenant's manifest never lands in this repo, and those are exactly the companies running a Medulla brain — but it reduces the unknown from "any `full` company" to "a provisioned tenant that both runs `full` and emits one of three exact spellings".

Given that, I think `[]` is the right call and the alternative is not worth its cost: `["payment", "filing", "publish"]` would preserve a native gate with no demonstrated user while reintroducing the harness-path lie that #684 exists to remove — and on the new matcher `publish` would not gate `publish_artifact` anyway, since that name has no segment boundary. The confusion would survive in the exact place the issue complains about it.

**One ask before merge, and it is a maintainer's call rather than mine:** someone with production visibility should confirm no provisioned tenant runs `full` with a hosted or sidecar brain emitting `payment.send` / `filing.submit` / `external.publish`. You have already scoped the fallback to one line, which is the right way to leave a decision you cannot make yourself — please get that confirmation on the PR rather than merging on the absence of an objection, since the failure mode is silent and lands on the tier with no other brake.

### Also worth having

Fixing the spec's self-contradiction — `manifest.md:198` saying "effect kinds" while `:257` tells the operator to write `always_approve = ["web_search"]` — matters more than it looks. That contradiction is how an operator ends up with a list that does nothing, and it is the documentation half of the same bug.

**Before merging:** five checks pending, and the branch is **32 commits behind** `main`. #659 merged in that range and adds `PolicyOverride::always_approve`, which is a *second* writer of this same list through the console — worth confirming the unified matcher reads console-set entries by the same rule, since that path did not exist when this branch was cut.

Issue #610, opened 2026-08-11
approvals: #457's per-toolkit grant scoping has no live caller in any tier after #559


M3gA-Mind:
## What this is

#457 shipped per-toolkit scoping for standing grants: an operator who approves "read from
GitHub" gets a grant scoped to the `github` toolkit, so a *second* GitHub read runs while
a Gmail read still has to ask. It is closed and it works.

**After #559 lands, nothing routes a Composio read to that check in any tier.** The
machinery is still correct; it has become unobservable.

This is not a defect introduced by #559's diff — it is a consequence of where the tier
answer sits relative to the grant checks, which #559 makes visible for the first time.
Filing it separately because a closed, shipped feature quietly losing its only caller is
not a paragraph in a PR body.

## The mechanism

`ApprovalPolicy::check()` (`src/harness/policy.rs`) evaluates in this order:

1. **`readonly` brake** — `src/harness/policy.rs:773`. Denies every external effect,
   **above both grant checks**. This ordering is deliberate and documented there: it is
   the #243 emergency-brake rule, "consent does not survive the brake."
2. single-use grant (#243)
3. standing grant (#374 / #457) — where `standing_grant_allows` consults the toolkit scope
4. …
5. the tier dispatch — `parks_under_supervision()` / `denied_under_readonly()`

Walk a Composio **read** through each tier after #559:

| tier | what happens | is the scope consulted? |
|---|---|---|
| `full` | allowed at the tier dispatch | no |
| `supervised` | #559 moves the read to `Reach::ExternalRead`, which does not park — allowed at the tier dispatch | no |
| `readonly` | denied at **step 1**, above the grant checks | no |

So the answer is *no* in all three.

## This makes #559's stated premise false

#559's body says:

> Keep `Standing::Grantable` either way — it stops mattering for the supervised path but
> still governs `readonly` and any future tier.

**It does not govern `readonly`.** The brake at step 1 denies before any grant is
consulted, so a scoped standing grant cannot admit a Composio read on a `readonly` desk —
with or without #559. The "any future tier" half of the sentence still stands, and is
now the *only* thing keeping `Standing::Grantable` meaningful for reads.

Worth being precise about what is and is not affected: `standing_scope_of`,
`StandingGrant::admits_scope` and `match_standing` are all still correct, still tested,
and still reachable for any *other* grantable tool. What is gone is the routing for
`composio_execute` — which, per `standing_scope_of`'s own doc comment, is
[...]

Issue #264, opened 2026-08-02
feat: an agent has no detail view — cannot be opened, read, or edited


oxoxDev:
## Problem

An agent on the Team tab is a dead end. The card renders a name and a role, and the only action behind its menu is a destructive Remove. There is no way to open an agent, read what it is, or change it.

Concretely, none of this is reachable from the console once an agent exists:

- the description and instructions it was defined with
- which tools it is allowed to use
- which desks it belongs to
- its tier (orchestrator vs worker)
- any edit at all — the "Define an agent" dialog is create-only

So the roster is write-once per member: define it, then live with it or delete it and start over. That mirrors the same shape as #259 for workflows, and it bites harder here because an agent's instructions are the thing most likely to need iteration.

It also blocks verification. Checking what tools a company actually grants its agents — the behavior #230 changed — has no console surface and no read endpoint either: `GET …/team` returns `tools: null` for every member, and there is no `…/tools` route.

## Proposal

Make an agent openable, and give it a detail view that shows what the agent is:

- identity (id, name, role, tier)
- description / instructions, as authored
- effective tool grants, resolved — not just the raw manifest list, but what the agent actually ends up with after the company's `[tools].allow` is applied
- desk membership
- edit for the fields that are safe to edit

The user's framing: there should be a way to reach "the `AGENT.md` or similar file for that agent". The manifest already carries this content — the console just never surfaces it after creation.

Suggested smallest useful slice, if this needs splitting: a read-only detail panel first (identity, description, resolved tools, desks), with editing as a follow-up. The read half alone closes the verification gap and makes the roster inspectable.

## Alternatives

- Expose the data on `GET …/team` and leave the UI alone. Cheaper, unblocks verification, does nothing for the operator staring at an unopenable card.
- Reuse the create dialog in an edit mode, as #259 plans to do for workflows. Consistent, and worth considering together — but the detail *view* is the part that is missing entirely, and a form is not a good reader.

---

## Testing — required, and this one is bound to the UI

**This issue has a user-facing surface, so it is verified on that surface.** A green unit or component test is not evidence that this is fixed, and does not close it.

Before this is done:

- Run the console against a live host and walk the exact path described above, by hand, in a browser.
- Exercise the **failure** this issue describes, not only the happy path — the fix is that the described behaviour is gone, so the old behaviour is what you have to try to reproduce.
[...]

PR #717, opened 2026-08-11
fix(policy): let an overlay teammate be scoped, and bound what add_agent may mint

### LINKED ISSUE #619: approvals: an overlay teammate cannot be scoped, so it silently holds the company's widest grant
## Problem

`OverlayAgent` (`src/ports/types.rs:2104`) carries only `id`, `name`, `role`, `description`. There is no tools field, so an operator-added teammate cannot be scoped at all.

Both the harness and the console resolve that absence to the same thing — the widest grant the company has:

```rust
// src/harness/mod.rs:2014
fn overlay_agent_to_manifest(overlay: &OverlayAgent) -> ManifestAgent {
    ManifestAgent { id, role, description, tier: None, tools: Vec::new(), budget_usd_daily: None }
}
```

`src/server/ops/team_agent.rs:346` mirrors it deliberately — "no per-agent tool line: it holds the company's standard grant". Empty `tools` then falls through `agent_effective_grants` to the full company `allow`. That inheritance rule is #264 working as designed and is not what this issue disputes.

The gap is narrower: a manifest teammate can be scoped with a `tools` line in `company.toml`; an overlay teammate has nowhere to write one. Every overlay teammate holds everything the company holds, permanently.

## Why it matters now rather than later

`add_agent` is declared `Reach::Nothing` (`src/policy/consequence.rs:169`) and sits in `INTRINSIC_TOOLS` (`src/harness/steps.rs:491`), so it is always present and never asks. A model can therefore mint a teammate that holds the company's full grant with no approval anywhere in the path.

That is tolerable only while every write still parks per call. It stops being tolerable the moment #560 lands a broader policy tier or #563 lands durable allow rules: an overlay teammate would silently inherit the widest standing grant in the company, and there would be no way to express "this one is read-only". The scoping hole should close before, or with, whichever of those lands first.

## Scope

1. Add an optional tools field to `OverlayAgent`, serde-additive (skip when empty) so existing company records serialise byte-identically.
2. Thread it through `overlay_agent_to_manifest` and the `team_agent.rs` projection **in lockstep** — if they drift, the console shows a grant the harness does not honour, which is the exact failure #264 exists to prevent.
3. Decide what `add_agent` may set. Recommendation: a model-minted teammate defaults to no wider than the minting agent's own effective grant, rather than to the company's.
4. Console: the teammate detail view should state where a teammate's grant came from — an explicit line, or inherited company allow — so an operator can tell a scoped teammate from an unscoped one by looking.

## Out of scope

The empty-means-inherit rule itself (#264). That is correct and stays. This issue is only about being able to write a non-empty value for an overlay teammate.

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#619.

`OverlayAgent` carried only `id`, `name`, `role`, `description`. With no tools field, an operator- or orchestrator-added teammate could not be scoped at all: empty `tools` falls through `agent_effective_grants` to the full company `[tools].allow`, so **every overlay teammate held everything the company held, permanently**, with no way to express "this one is read-only".

This adds the field and threads it through every reader, bounds what `add_agent` may mint, and makes the mint observable.

The empty-means-inherit rule itself (#264) is **not** touched. It is correct and stays. This is only about being able to write a *non-empty* value.

### The field

`OverlayAgent.tools: Vec<String>`, serde-additive — `#[serde(default, skip_serializing_if = "Vec::is_empty")]`. A record written before this field existed still parses, and an unscoped teammate still serialises **without** the key, so the first save after upgrade does not rewrite every overlay row in the store for no behaviour change. Both halves are pinned by `an_overlay_teammates_tool_line_is_serde_additive`.

### Three readers, threaded in lockstep

The issue calls for the harness and the console projections to move together — if they drift, the console shows a grant the harness does not honour, which is the exact failure #264 exists to prevent. There are **three** readers, not two:

| reader | what it decides |
|---|---|
| `harness::overlay_agent_to_manifest` | what `build_agent` actually wires onto the agent |
| `server::ops::team_agent::requested_grants` | the console's agent-detail read |
| `server::ops::mcp::roster_grants` | **every MCP server row's `reachableBy`** |

The third is the one easiest to miss on review. It computes who can reach each MCP server, and it hard-coded the overlay half to `&[]`. Left alone, a teammate scoped away from `mcp:*` would still be listed as reaching every enabled server — the console asserting a connection that is not there, which is the same inversion the issue is about wearing a different hat.

### The fingerprint axis

**`overlay_fingerprint` is the fingerprint covering this field**, and it did not cover it until this PR. It hashes `overlay_agents` only — which is the *correct* axis here precisely because `tools` lives on `OverlayAgent`, so no new axis was needed; `agent.tools.hash(&mut hasher)` joins the existing id/name/role/description hashes.

Without that line, narrowing a teammate from the console is a **silent no-op until the process restarts**: `HarnessPool::ensure`'s staleness fast-path would not notice, and the roster would not rebuild. This repo has shipped exactly that bug twice already (`policy_fingerprint` in #562, and this area). `narrowing_a_teammates_scope_changes_the_overlay_fingerprint` fails when the hash line is removed.

No other overlay field is written by this change, so no second axis is implicated. `fresh_company`'s pre-rebuild fold-in is untouched — it clones `overlay_agents` wholesale, so the new field rides along unchanged.

### `add_agent`: bounded, and no longer silent

`add_agent` is `Reach::Nothing` and sits in `INTRINSIC_TOOLS`, so it is always present and never asks. A model could mint a teammate holding the company's full grant with no approval anywhere in the path. The rule now: **a minted teammate is never wider than the agent that minted it.**

- **No `tools` argument** → the teammate copies the minter's own `tools` **line**, verbatim. Copying the line rather than the resolved grant is deliberate: an unscoped minter mints an unscoped teammate that keeps tracking `[tools].allow`, instead of freezing today's allow-list into the record as an explicit scope a later company-wide narrowing would not reach.
- **With a `tools` argument** → narrowed against the minter's own *effective* grant, so the tool cannot hand out what its caller does not hold.
- **A request that narrows to empty is a tool error, not a stored empty list.** This is the sharp edge and it is worth stating plainly, because it is the kind of thing a later refactor undoes without noticing: an empty `tools` list means *inherit the company's standard grant*. Storing the empty result of a narrowing would turn the single most deliberate narrowing an agent can ask for — "scope this teammate to something I don't have" — into the **widest grant in the company**.

**Observability is required and included.** Every mint logs the minter, the teammate, and the resolved scope at `info`, naming inheritance explicitly when the grant is inherited. A narrowing that happens silently is the defect being fixed, repeated one layer down.

### Console

`PATCH …/team/{agentId}` accepts `tools` for an overlay teammate. An omitted key leaves the scope alone; `[]` is the deliberate way back to the company's standard grant (so `null` never has to mean a third thing); a blank glob inside the list is a `400`, because `""` matches nothing an operator meant and would read as a scope that grants nothing while looking like a scope that was set. Globs are stored verbatim exactly like a manifest line — the `[tools].allow` ceiling is applied at read time, so a glob the company does not cover surfaces as asked-for-but-not-granted rather than vanishing on save.

The console can only ever *narrow* a teammate within a grant the company already made; this route does not touch `[tools].allow`.

**No new UI strings, so no i18n changes.** The console already distinguishes inherited-from-explicit through `summarizeGrants().standardGrant`, and that rendering now tells the truth for overlay teammates because `requested_grants` reads their line. This is the issue's item 4, satisfied by existing rendering. Zero frontend files changed.

## API Or Behavior Changes

- **`OverlayAgent` gains `tools`** (optional, additive). Existing records round-trip byte-identically; an unscoped teammate behaves exactly as before.
- **`PATCH {scope}/team/{agentId}` accepts `tools`** for an overlay teammate, and `editable` now lists `"tools"`. A **manifest** teammate's `tools` is still a `409` — its line lives in the version-controlled `company.toml`. **The scope of this change is the overlay half only.**
- **`add_agent` gains an optional `tools` parameter** and now bounds every mint by its minter (see above). Its success text names the resulting scope.
- **MCP `reachableBy`** now reflects a scoped overlay teammate's real grant instead of always reporting the company's.
- `POST {scope}/team` is unchanged — a console-created teammate is still unscoped on creation, and scoping is a `PATCH` away.

## Tests

Twelve tests added or changed; the full suite is green.

```
cargo fmt --all -- --check                                                      → 0
cargo check  --locked --all-features --all-targets                              → 0
cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings  → 0
cargo build  --all-targets                                                      → 0
cargo test   --features openhuman,tinycortex   → 0 — 3140 + 10 + 1 + 11 + 2 = 3164 passed, 0 failed, 3 ignored
```

`--all-features` is not redundant here: `mongodb.rs` / `sqlite.rs` build `CompanyRecord` from a row rather than a struct literal, so a field addition is invisible to the default and gated lanes and caught only by that one.

**Every new test was revert-checked** — its own fix removed, the named test re-run, and confirmed to fail. Counts quoted so a run that silently executed nothing cannot pass for a pass:

| revert | test | result |
|---|---|---|
| drop `skip_serializing_if` | `an_overlay_teammates_tool_line_is_serde_additive` | FAILED — 0 passed; 1 failed |
| `overlay_agent_to_manifest` → `Vec::new()` | `an_overlay_teammates_tool_line_reaches_the_manifest_shape` | FAILED — 0 passed; 1 failed |
| drop `tools` from `overlay_fingerprint` | `narrowing_a_teammates_scope_changes_the_overlay_fingerprint` | FAILED — 0 passed; 1 failed |
| drop the overlay lookup in `requested_grants` | `an_overlay_teammate_can_be_scoped_and_the_scope_is_enforced` | FAILED — 0 passed; 1 failed |
| `trimmed_globs` never refuses | `a_blank_tool_glob_is_refused` | FAILED — 0 passed; 1 failed |
| `OVERLAY_EDITABLE` back to three fields | `the_host_states_which_fields_are_editable` | FAILED — 0 passed; 1 failed |
| `mcp::roster_grants` → `&[]` | `a_scoped_overlay_teammate_does_not_read_back_as_reaching_everything` | FAILED — 0 passed; 1 failed |
| `add_agent` stores `Vec::new()` | `a_minted_teammate_copies_the_minters_scope`, `an_explicit_scope_is_narrowed_to_what_the_minter_holds` | both FAILED — 0 passed; **2 failed** |
| drop the empty-narrowing refusal | `a_scope_entirely_outside_the_minters_grant_is_refused` | FAILED — 0 passed; 1 failed |

The working tree was restored and re-verified after the reverts, and the five lanes above were run **after** restoration.

Three further tests are **regression guards on unchanged behaviour** and pass before and after by design — they are not claimed as revert-proven: `an_unscoped_minter_mints_an_unscoped_teammate`, `an_unscoped_overlay_teammate_still_inherits_the_company_grant` (both pin #264's empty-means-inherit), and `a_manifest_teammates_tools_are_still_not_editable_here` (pins the `409`).

## Documentation

`docs/spec/runtime/api.md` updated. It previously stated outright that **"`tier` and `tools` are read-only for both kinds"**, which this PR makes false for `tools` on the overlay half — so the correction is required, not cosmetic. The revised text is explicit that **only the overlay half changed**: a manifest teammate's tool line still lives in git and is still a `409`. The same claim in the `server/ops/team_agent.rs` module docs is corrected identically.

Also documented: the overlay `tools` field and its empty-means-inherit rule, the `PATCH` contract for it, and the `add_agent` minting rule including why a narrowing-to-empty is refused rather than stored.

`docs/spec/runtime/api.md` was already 587 lines — over the repo's 500-line guideline — before this change, and this adds ~32. Splitting it is out of scope here; flagging rather than growing it silently.

## Not in this PR

- **#610** (retaining #457's per-toolkit grant scoping, pinned by a direct unit test on `standing_scope_of` / `admits_scope`) is adjacent and assigned elsewhere. **Neither function is touched here** — #619 did not require them.
- The empty-means-inherit rule (#264) is unchanged, deliberately.

### DIFF
--- a/docs/spec/runtime/api.md
+++ b/docs/spec/runtime/api.md
@@ -229,8 +229,16 @@ exactly one root and no member folders.
 Team writes are an **operator overlay** persisted through the store, merged
 into the manifest roster at read time — the version-controlled `company.toml`
 is never rewritten. Overlay teammates are addressable: since issue #71 the
-harness builds a real agent for each one, with the company-wide tool grant, no
-cognition tier, and never the orchestrator.
+harness builds a real agent for each one, with no cognition tier and never the
+orchestrator.
+
+Since issue #619 an overlay teammate carries its own `tools` list — the
+overlay's answer to a manifest `[[agent]].tools` line, and the thing that was
+missing when the only readable answer for every teammate was "everything the
+company allows". It obeys the same rule a manifest line does: **empty means the
+company's standard grant** (#264), so a teammate written before the field
+existed reads back exactly as it did. What changed is that a non-empty value is
+now writable at all.
 
 `POST …/team` and the orchestrator's `add_agent` tool both **derive the roster
 id from the display name** (issue #686): "Dana Designer" becomes
@@ -270,7 +278,7 @@ lists, because only the third is the answer:
 
 | field | meaning |
 |---|---|
-| `requested` | the agent's own `[[agent]].tools` globs. **Empty means the company's standard grant**, not "no tools" |
+| `requested` | the agent's own globs — a manifest `[[agent]].tools` line, or an overlay teammate's `tools` list (#619). **Empty means the company's standard grant**, not "no tools" |
 | `companyAllow` | the `[tools].allow` ceiling the request is intersected with |
 | `effective` | what the agent actually holds |
 
@@ -280,18 +288,42 @@ when it builds the agent, so the readout cannot drift from what is enforced.
 "orchestrator"` agent, else the first declared) rather than read off `tier`, so
 a company that tags nobody still names its orchestrator.
 
-`PATCH …/team/{agentId}` edits an **overlay** teammate's `name`, `role` and
-`description`. It is a patch: an omitted key is left alone, and `"description":
-null` clears it — the two must stay apart or every partial save would erase an
-agent's instructions. A blank `name`/`role` is `400`, an unknown teammate `404`,
+`PATCH …/team/{agentId}` edits an **overlay** teammate's `name`, `role`,
+`description` and `tools`. It is a patch: an omitted key is left alone, and
+`"description": null` clears it — the two must stay apart or every partial save
+would erase an agent's instructions. `tools` needs no such distinction: an
+omitted key leaves the scope alone and `"tools": []` is the deliberate way back
+to the company's standard grant, so `null` never has to mean a third thing. A
+blank glob inside the list is a `400` — `""` matches nothing an operator meant,
+and storing it would read as a scope that grants nothing while looking like a
+scope that was set. Globs are stored verbatim, exactly like a manifest line: the
+`[tools].allow` ceiling is applied at read time, so a glob the company does not
+cover surfaces as asked-for-but-not-granted rather than vanishing on save.
+
+A blank `name`/`role` is `400`, an unknown teammate `404`,
 and a **manifest** teammate is `409`: its fields live in the version-controlled
 `company.toml`, and the console does not rewrite the blueprint. The one thing
 that *is* changeable on such a teammate is its daily budget, and that works
 because #343 modelled it as an override rather than as a rewrite. Every detail
 response carries an `editable` list naming the fields this route will accept, so
 a client renders read-only from the host's answer instead of re-deriving the
-rule. `tier` and `tools` are read-only for both kinds: there is no override
-layer for either, and adding one is a policy decision rather than a form field.
+rule. `tier` stays read-only for both kinds: there is no override layer for it,
+and adding one is a policy decision rather than a form field. `tools` was
+read-only for the same reason until #619 made that policy decision for the
+overlay half — a manifest teammate's `tools` line still lives in the blueprint
+and is still `409`.
+
+The orchestrator's `add_agent` tool writes the same field, under one added rule:
+**a minted teammate is never wider than the agent that minted it** (#619).
+Omitting `tools` copies the minter's own line — so an unscoped minter still
+mints an unscoped teammate, which keeps tracking `[tools].allow` instead of
+freezing a copy of it. Passing `tools` narrows the request against what the
+minter actually holds, and a request that narrows to nothing is a clean tool
+error rather than a stored empty list, because an empty list means *inherit
+everything* — silently storing one would turn a deliberate narrowing into the
+widest grant in the company. Every mint is logged with the minter, the teammate
+and the resolved scope: `add_agent` is `Reach::Nothing` and never asks, so the
+log is the only place the grant is observable at all.
 
 The two **budget** routes (issue #343) are how a teammate's `budget_usd_daily`
 becomes changeable without a redeploy. Both are **admin-only** — a member gets
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -4597,6 +4597,7 @@ members = ["engineer"]
             name: "Nova".into(),
             role: "Growth".into(),
             description: None,
+            tools: Vec::new(),
         });
         tasks
             .upsert(&CompanyId::new("acme"), &card("t1", "nova"))
@@ -4925,6 +4926,7 @@ members = ["eng1", "eng2"]
                 name: "Cto".to_string(),
                 role: "CTO".to_string(),
                 description: None,
+                tools: Vec::new(),
             }],
             overlay_desk_members: vec![crate::ports::types::OverlayDeskMember {
                 desk_id: "eng".to_string(),
--- a/src/harness/build.rs
+++ b/src/harness/build.rs
@@ -610,6 +610,13 @@ pub fn build_agent(
             // the `read_run_output` companion reads back, so a clipped preview
             // is reachable within the turn. Orchestrator-only, like the tools.
             deps.run_outputs.clone(),
+            // Issue #619: who is minting, and how wide they are. `add_agent`
+            // bounds the teammate it mints by this agent's own scope, and names
+            // this agent in the mint log — a teammate can no longer be minted
+            // wider than its minter, or minted invisibly.
+            manifest_agent.id.clone(),
+            manifest_agent.tools.clone(),
+            grants.to_vec(),
         ));
     }
 
--- a/src/harness/mod.rs
+++ b/src/harness/mod.rs
@@ -1867,6 +1867,12 @@ fn overlay_fingerprint(agents: &[OverlayAgent]) -> u64 {
         agent.name.hash(&mut hasher);
         agent.role.hash(&mut hasher);
         agent.description.hash(&mut hasher);
+        // Issue #619: the teammate's own tool line is roster-shaping, not
+        // display data — it decides what `build_agent` wires onto the agent. A
+        // scope narrowed through the console would otherwise not take effect
+        // until the next restart, which is the silent-no-op trap this repo has
+        // already hit twice (`policy_fingerprint` in #562, and this axis).
+        agent.tools.hash(&mut hasher);
     }
     hasher.finish()
 }
@@ -2129,10 +2135,11 @@ pub(crate) fn build_roster(
 }
 
 /// Converts an operator-added [`OverlayAgent`] into the manifest agent shape
-/// [`build::build_agent`] consumes: an empty `tools` list (so
-/// [`agent_effective_grants`] falls back to the full company `[tools].allow`
-/// — the "standard tool grant"), no cognition tier (→ the default `chat-v1`
-/// model), and no manifest budget cap — an overlay teammate has no manifest row
+/// [`build::build_agent`] consumes: the teammate's own `tools` line verbatim
+/// (issue #619 — empty, as it is for every teammate minted before that field
+/// existed, so [`agent_effective_grants`] falls back to the full company
+/// `[tools].allow`, the "standard tool grant"), no cognition tier (→ the default
+/// `chat-v1` model), and no manifest budget cap — an overlay teammate has no manifest row
 /// at all, so its cap (if any) comes from the record's budget overrides via
 /// [`CompanyRecord::effective_budget`], resolved by the caller. The overlay's
 /// `name` is a display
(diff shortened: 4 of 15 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
The sharp edge is identified correctly and closed correctly, and it is the reason this PR needed writing carefully rather than quickly:

> an empty list means "inherit everything", so silently storing one would turn a deliberate narrowing into the widest possible grant

I checked it in code rather than in its doc:

```rust
let narrowed = agent_effective_grants(&self.minter_grants, &globs);
if narrowed.is_empty() {
    return Ok(ToolResult::error(...));
}
```

A narrowing that survives empty is a clean, actionable error naming what the minter holds — not a stored `[]`. That inversion would have turned the fix for #619 into a wider version of #619, and it is exactly the kind of thing a later refactor "simplifies" away, so stating it at the type docs *and* pinning it is the right amount of noise.

Carrying **both** `minter_tools` (the verbatim line) and `minter_grants` (the effective ceiling) is what makes the two cases correct for different reasons, and the argument for copying the *line* on the no-argument path is subtle and right: an unscoped minter mints an unscoped teammate that keeps tracking `[tools].allow`, rather than freezing today's allow-list into a record that a later company-wide narrowing would never reach.

Two more things verified:

- **The third reader.** `mcp::roster_grants` hard-coding the overlay half to `&[]` is the one I would have missed, and you are right about the shape of the failure: a teammate scoped away from `mcp:*` still listed as reaching every enabled server is the console asserting a connection that is not there — the same inversion in different clothes.
- **The fingerprint line.** `agent.tools.hash(..)` joining `overlay_fingerprint` is what stops a console narrowing from being a silent no-op until restart. This repo has now shipped that exact class twice; I verified the same axis on #659 earlier today, so the instinct to check it here is well-founded.

Also worth noting the PR already absorbed the concurrency finding (per-company mutex so `add_agent` and `POST …/team` cannot clobber each other's `overlay_agents` in a load→push→save race) and the duplicate-name finding.

**0 major. 1 question I could not answer from outside, and it is the one to answer before merging.**

### Question — `OVERLAY_EDITABLE` just gained a security field alongside three cosmetic ones

```rust
const OVERLAY_EDITABLE: [&str; 4] = ["name", "role", "description", "tools"];
```

`name`, `role` and `description` are presentation. `tools` is a privilege boundary. Whatever authorization currently guards that route now guards a materially different class of decision, and I could not establish from the diff which extractor it sits behind — I looked for `require_admin` / `AdminScopedCompany` in `team_agent.rs` and found neither, though the handler may be gated elsewhere or by an extractor I did not match.

Please confirm explicitly on the PR. If it is admin-gated, this is a non-issue and worth one line in the module docs saying the gate is now load-bearing for a reason it was not before. If it is member-reachable, a member can restrict another teammate's scope — bounded, since you can only ever narrow within the company's existing grant and this route never touches `[tools].allow`, so it is capability-denial rather than escalation. Still not a decision a member should make, and it would sit oddly beside #685, which just established that a member may not read an approval's payload.

I am not blocking on it because I have no evidence of a defect — only an unverified boundary on the one field in that list where the boundary matters.

**Before merging:** four checks still pending. Base is current, which is worth having today.

**Scope:** I read the `add_agent` bounding path, the three readers, the fingerprint line and the route's editable set. I did not read all fifteen files.

senamakel:
Confirmed explicitly, as you asked.

**The `OVERLAY_EDITABLE` / `tools` edit on `PATCH {scope}/team/{agent_id}` is member-reachable, not admin-gated.** `edit_agent` carries no `require_admin`; it is deliberately "open to any signed-in member, matching `POST …/team`" — that is stated in the module doc in `src/server/ops/team_agent.rs`, which also carries the privilege analysis for `tools` (it `only ever narrow[s] a teammate within a grant the company already made`).

I want to be honest that your second scenario is the one we're in:

- **Bounded**: the field can only narrow a teammate within the grant the company already holds, and this route never touches `[tools].allow`, so it is capability-denial rather than escalation. The empty-narrowing sharp edge (store-empty would read back as "inherit the whole company grant") is guarded the same way here as on the `add_agent` path.
- **Consistent with the server's existing foothold**: `POST …/team` was never admin-only, so the PATCH matching it is by design. And now that #661 is in `main`, that same member-reachable `POST` can already carry a `tools` scope at creation — so member-reachability of a *narrowing* scope is the established server posture, not something this PR introduces.

I agree the `#685` comparison is fair and it reads a little oddly. If the maintainers want the narrowing action gated to admins, that is a small, clean follow-up (hoist an admin check into `edit_agent` the way the budget route does), and I'm happy to make it. I did not want to silently re-architect the authorization boundary of an already-approved PR in this merge pass, so the decision stays with maintainers rather than being smuggled in.

M3gA-Mind:
Superseded by #745. Closing rather than rebasing, and **retaining this branch**.

## Why not a rebase

#713 (#661) landed the foundation this PR was building while it sat approved: `OverlayAgent.tools`, the `overlay_fingerprint` hash **and its test**, the harness carry through `overlay_agent_to_manifest`, and the team-detail read. That is roughly 60% of this PR.

Rebasing would mean resolving conflicts across fifteen files by taking `main`'s side on most of them — hand-deleting my own work to arrive at what is already merged. The remainder is a different, smaller change and deserves its own diff.

## What was dropped as redundant

All of it now on `main` via #713:

- `OverlayAgent.tools` and its serde-additive shape
- `overlay_fingerprint` hashing the field, plus the freshness test
- `overlay_agent_to_manifest` carrying it
- `team_agent::requested_grants` reading it
- scope-at-creation on `POST …/team`

Credit where it is due: #661 got the fingerprint axis right, which is the trap this repo has hit twice (`policy_fingerprint` in #562, and this axis).

## What carried over into #745

The parts #661 did not cover:

1. **The third reader** — `roster_grants` in `src/server/ops/mcp.rs` still passed an empty grant, so a scoped teammate reported as reaching **every** enabled MCP server. Filed as #740; a live regression on `main`, not a hypothetical.
2. **`add_agent` bounded by the *minter*, not the company.** #661 clamps an explicit `tools` argument to the company grant — its own comment says "can only narrow below the company grant" — so omitting `tools` still yields the company's whole grant. That is #619's original defect, and bounding by the minter's effective grant is what closes it.
3. **The observability** that was the condition of sanctioning that narrowing at all — minter, teammate, resolved scope. `add_agent` is `Reach::Nothing` and never asks, so the log is the only place the decision is visible.
4. **`PATCH …/team/{id}` accepting `tools`**, so a teammate can be narrowed *after* creation rather than only at it — otherwise correcting a grant means deleting and recreating the teammate, orphaning its workspace folder, budget row, desk memberships and inbox.

No review comments on this PR are left unaddressed — it was approved and green; it simply stopped being the right diff.

Issue #563, opened 2026-08-10
approvals: no durable allow rules, so consent given once must be given again


oxoxDev:
## Symptom

An operator who approves the same shape of call repeatedly has no way to say "and from now on". The two mechanisms that exist both fall short:

- **A standing grant** is offered on the approval card, is scoped to one teammate and one tool until a deadline, and expires. It also only applies to tools declared `Standing::Grantable` — which excludes `shell`, `http_request`, `workspace_write` and `git_operations` by deliberate design, and (today) includes Composio reads only because they were classified as a consequence in the first place.
- **`always_approve`** is the inverse — a list of things to *always* ask about — and it is manifest-only.

There is no persisted "these are fine, stop asking" rule set. Claude Code's allowlist is exactly that: a durable pattern list an operator grows over time, where approving once can be promoted into never being asked again.

## The change

A durable per-company allow rule set, consulted above the tier, that an operator grows from the approval card itself.

Design questions worth settling before code:

- **What is a rule keyed on?** Tool name alone is too coarse for `composio_execute`, where one name carries every action across every connected toolkit. `src/policy/consequence.rs` already has the notion of confining a grant to a slice of a tool for exactly this reason — that concept is the right seam.
- **Does a rule expire?** The standing grant's deadline exists so a capability does not sit open forever. A durable rule trades that safety for the convenience the request is asking for. A reasonable middle: durable for reads and sandbox writes, deadline-bound for anything outward.
- **Can a rule be revoked, and can the operator see what is currently allowed?** A permission set nobody can enumerate is one nobody can audit. Whatever the storage, there should be a screen that lists the live rules and lets one be removed.
- **What may never become a rule?** Some things should stay per-call however often they are approved — moving money, filing, changing identity. `always_approve` is the existing expression of that idea and the two should not be able to contradict each other.

## Acceptance

- Approving a card offers "don't ask again for this", and taking it means the next matching call runs unattended.
- The rule survives a restart.
- The live rules are listable and individually removable.
- A rule cannot be created for anything in `always_approve`.

M3gA-Mind:
**Scoping pass — verified against `upstream/main` @ `843d5c2`.**

Before this is scoped as new construction: **more of it exists than the body suggests.** In `src/runtime/grants.rs`:

[...]

PR #724, opened 2026-08-11
approvals: pin #457's per-toolkit grant scoping directly, and record why it has no live caller (#610)

### LINKED ISSUE #610: approvals: #457's per-toolkit grant scoping has no live caller in any tier after #559

M3gA-Mind:
## Summary

**No behaviour change.** #457's per-toolkit grant scoping is correct and shipped, and
after #559 no live tier routes a Composio call to it. Working, reviewed security
machinery with no caller and no test is what a future reader deletes as dead code. This
retains it deliberately and says so where that reader will look.

### The test: the two halves, pinned against each other

`standing_scope_of` (`src/policy/consequence.rs`) mints a scope and
`StandingGrant::admits_scope` (`src/runtime/grants.rs`) spends it. Both are called
**directly** — not through a tier, which would pass vacuously since no tier reaches them.

Each half already has direct coverage, so this does not duplicate it. What was unpinned is
the **seam**: `a_composio_call_is_scoped_to_its_toolkit` spells the toolkit as its own
`"github"` literal, and `grants.rs`'s scope tests spell it as theirs. Two literals in two
files are not an agreement — change what `standing_scope_of` returns and both suites can
be brought to green independently while the pairing they describe is broken, and there is
no live caller left to notice.

So `the_minted_scope_is_the_scope_a_grant_admits` writes no scope string down. Every scope
comes out of `standing_scope_of` and goes straight into a `StandingGrant` or into
`admits_scope`, which makes it a test of whether the two still agree rather than of what
either one says. It covers the four cases that matter: a second action from the same
provider is admitted (toolkit grain, not slug grain), another provider's action is
refused, an action the catalogue cannot place resolves to `None` and is refused rather
than guessed at, and a pre-#457 unscoped grant still admits whatever its `(agent, tool)`
pair did.

**What this test deliberately does not catch, and what does.** Because both sides derive
from `standing_scope_of`, a change that moves both together — returning `GITHUB` instead
of `github` — leaves it green. That is inherent to testing an agreement. The hazard such a
rename actually poses is to *stored* grants, where a journal line written under the old
spelling would silently stop matching, and that is caught by the existing
`a_composio_call_is_scoped_to_its_toolkit`, which asserts the literal `Some("github")`.
The two tests are complementary and neither alone is sufficient; this is recorded here
because a reader who assumes the new test subsumes the old one could delete the wrong one.

### The comment: why it is uncalled, at the definition

`standing_scope_of` now carries a "no live caller, and retained on purpose" section
stating the justification directly rather than by reference to a ticket a reader would
have to go and interpret. It names the three tiers and why each misses the scope
(`supervised` allows a catalogue read at the tier since #559; `readonly` denies at the
#243 brake, which sits *above* the grant checks; `full` allows everything), what would be
re-derived if it were deleted, and why #563 closed — its card-promotion shape cannot work
under `auto`, because `parks_under_auto()` is `reach.parks_under_supervision() &&
!standing.is_grantable()` while the standing-grant control is offered only for `Grantable`
tools: exact complements on the parked set, a partition rather than a gap. A console- or
manifest-managed allowlist consulted **before** the tier check has no such problem and
would use this.

`admits_scope` gets a short pointer to that one explanation rather than a second copy.

Closes tinyhumansai/opencompany#610

## API Or Behavior Changes

None. Two doc comments and one test; no runtime code path is touched. `cargo build` output
differs only by the added test.

## Tests

- `cargo test --features openhuman,tinycortex` — **3129 + 10 + 1 + 11 + 2 = 3153 passed, 0 failed**
- `cargo test` (default features) — **2083 passed, 0 failed**

Revert-and-check on `the_minted_scope_is_the_scope_a_grant_admits`, each run selecting
exactly one test:

| Reverted | Test outcome |
|---|---|
| `admits_scope` always returns `true` | 1 selected, **1 FAILED** ("'read from GitHub' is not consent to read the company's mail") |
| a scoped grant admits `None` | 1 selected, **1 FAILED** ("unknown is a send here too") |
| `standing_scope_of` scopes by action slug, not toolkit | 1 selected, **1 FAILED** ("the operator consented to a provider, not to one action slug") |
| `standing_scope_of` uppercases the toolkit | 1 selected, **PASSED — not caught** |

The last row is reported rather than omitted: it is the limitation described above, and it
is covered by `a_composio_call_is_scoped_to_its_toolkit`'s literal assertion instead.

The test is gated `#[cfg(feature = "openhuman")]`, matching the neighbouring scope tests —
without the catalogue every toolkit resolves to `None`, and that build cannot mint a
Composio standing grant at all. It is run by the `Rust (openhuman, tinycortex)` lane.

- [x] `cargo clippy --all-targets -- -D warnings` — run as `--locked --no-deps --features openhuman,tinycortex --all-targets`; `--no-deps` keeps pre-existing `vendor/tinyagents` lints out, which CI does not see either
- [x] `cargo test` — plus the gated `--features openhuman,tinycortex` lane, which is the one that runs the new test
- [x] N/A — no i18n keys: this change adds no user-facing UI text, only Rust doc comments and a unit test

## Documentation

The documentation *is* the change: the retention rationale lives at `standing_scope_of` in
`src/policy/consequence.rs`, with a pointer from `StandingGrant::admits_scope` in
`src/runtime/grants.rs`. No `docs/` page describes this internal seam, and adding one would
put the explanation somewhere the reader deciding whether to delete the function would not
be looking.

### DIFF
--- a/src/policy/consequence.rs
+++ b/src/policy/consequence.rs
@@ -730,6 +730,35 @@ pub(crate) fn composio_action_slug(args: &serde_json::Value) -> Result<&str, Act
 /// the first place — `without_the_catalogue_every_composio_action_is_a_send`
 /// pins that — so there is no scoped grant there for `None` to widen.
 ///
+/// # No live caller, and retained on purpose (issue #610)
+///
+/// **Nothing in any current tier routes a Composio call to this.** Since #559 a
+/// catalogue read no longer parks under `supervised`, so the tier allows it well
+/// above the grant checks; under `readonly` the #243 emergency brake denies
+/// every external effect *above* them; under `full` everything is allowed. The
+/// scope is minted and stored, and no tier spends it.
+///
+/// That is dormancy, not death, and the distinction is written here because a
+/// mechanism with no caller and no note is indistinguishable from dead code —
+/// and this one is a reviewed security boundary. It is what any future durable
+/// allowlist would consult, and rebuilding it later would mean re-deriving the
+/// same decision: that consent attaches to a *provider*, that a slug the
+/// catalogue cannot place resolves to `None`, and that a scoped grant refuses
+/// `None` rather than guessing permissively. #563 proposed one such feature and
+/// was closed because its **card-promotion** shape cannot work under `auto`: a
+/// tool parks under `auto` exactly when it is not [`Standing::Grantable`], and
+/// the standing-grant control is offered only for `Grantable` tools, so no card
+/// under `auto` can ever offer "don't ask again" — exact complements on the
+/// parked set, a partition rather than a gap. A console- or manifest-managed
+/// allowlist consulted **before** the tier check has no such problem, and would
+/// use this.
+///
+/// Deleting it is therefore a decision to re-derive it later, and should be made
+/// as one. `the_minted_scope_is_the_scope_a_grant_admits` pins this function
+/// against [`StandingGrant::admits_scope`] directly, so the pairing keeps its
+/// coverage while no caller connects them.
+///
+/// [`Standing::Grantable`]: crate::policy::consequence::Standing::Grantable
 /// [`StandingGrant::admits_scope`]: crate::runtime::grants::StandingGrant::admits_scope
 pub fn standing_scope_of(tool: &str, args: &serde_json::Value) -> Option<String> {
     if !tool.eq_ignore_ascii_case(COMPOSIO_EXECUTE) {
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
The test is designed around the right observation, and it is one most coupling tests get wrong:

> Two literals in two files are not an agreement — change what `standing_scope_of` returns and both suites can be brought to green independently while the pairing they describe is broken.

Writing no scope string down, and passing every scope straight from `standing_scope_of` into `StandingGrant` / `admits_scope`, makes this a test of whether the two halves still agree rather than of what either one says. That is the only shape that survives a refactor of the scope format, which is exactly the change that would otherwise break the pairing silently. Calling both halves directly rather than through a tier is right for the same reason — a tier route would pass vacuously, since none reaches them.

The four cases are the ones that matter, and the third is the one I would have wanted most: an action the catalogue cannot place resolving to `None` and being **refused rather than guessed at**. That is the fail-closed property, and it is the one a well-meaning future edit ("fall back to the provider prefix") would quietly invert.

Retaining reviewed security machinery rather than letting the next reader delete it as dead code is a defensible call, and doing it with a test plus a note beats a comment alone.

### Question — record *why* there is no live caller, not just *that* there isn't

The PR says "after #559 no live tier routes a Composio call to it" and stops there. That is the fact, but the fact a future reader needs is the **condition**: what changed such that having no caller is correct, and what would have to become true again for one to exist.

It matters because the two readings look identical from here. Either #559 legitimately removed the need — a Composio call no longer reaches the mint path, so no scope is owed — or the scoping was orphaned as a side effect and standing grants for Composio are now coarser than #457 intended. The first is housekeeping; the second is a live weakening that this PR would be documenting rather than reporting. I could not settle it from the diff, and `consequence.rs`'s own note that a Composio read is `Grantable` while carrying `Reach::Consequence` reads, at a glance, like the mint path is still theoretically reachable.

One or two sentences at the retention note — naming the condition #559 changed, and what would restore a caller — turns this from "kept because it looked important" into something the next reader can act on. If the answer is that the mint path genuinely cannot be reached today, saying so also makes the test's own value clearer: it is guarding a seam for the moment it becomes live again, which is a better argument than the one currently written down.

**Before merging:** base is current and the diff is purely additive. Just let the lanes finish.

Issue #558, opened 2026-08-10
EPIC: approvals interrupt for things that carry no consequence, and every interruption dead-ends the turn


oxoxDev:
## The problem

Ask an agent to generate anything and the operator gets a wall of approval cards — on the task card and in the Approvals tab — for calls that carry no consequence worth interrupting a person over. Reading a mailbox asks. Writing a scratch file asks. Listing a repository's pull requests asks.

That is not a rendering bug. Four separate design decisions compose into it, and each one is defensible alone:

1. **The default tier parks every write and every external read.** `default_policy_mode()` is `supervised` (`src/company/types.rs`), and no shipped preset except `signals_opportunity_studio` sets `[policy]` at all — so every company created from a preset, including the marketing one, runs supervised.
2. **Under supervised, anything classified `Reach::Consequence` parks.** That includes `file_write`, `edit`, `apply_patch`, `memory_store`, `workspace_write`, `run_workflow`, `mcp_call_tool` — and every Composio **read**.
3. **A parked call is refused, not suspended.** openhuman resolves `RequireApproval` inline and fail-closed: the tool call is blocked, the model is handed a refusal, and the turn continues without the result. Approving afterwards mints a single-use grant and the agent has to be **re-dispatched to re-issue the same call**. So each approval costs a round trip and the work visibly dead-ends in between.
4. **There is no tier between the two extremes.** `supervised` parks every consequence; `full` parks nothing but the `always_approve` list. An operator who wants "don't ask me to read things, do ask me before you send money" has no setting to express that — and cannot change the mode from the console at all, because `[policy].mode` is manifest-only.

## What we want instead

The model Claude Code uses, which the team already lives in daily:

- **Reads never prompt.** Not local reads, not reads through a connected third-party account.
- **Writes inside the agent's own sandbox never prompt.** Scratch files, patches, the company's own memory.
- **Only consequence that leaves the building prompts** — pushing to a remote, sending mail or a message, publishing, spending money, moving cash, filing something, changing an identity.
[...]

PR #725, opened 2026-08-11
brain: classify an effect kind by segment, not by bare substring (#704)

### LINKED ISSUE #704: approvals: effect_group_for matches bare substrings, so design.review is classified as a signature and always parks
`effect_group_for` classifies a brain-emitted effect kind by bare substring, and two of the substrings are far too short to be safe. Live on `main` @ `9af6d672`.

## The code

`src/brain/medulla/effects.rs`, `effect_group_for`:

```rust
} else if k.contains("payment") || k.contains("spend") || k.contains("x402") || k.contains("pay") {
    EffectGroup::Spend
} else if k.contains("sign") || k.contains("filing") || k.contains("contract") {
    EffectGroup::Sign
}
```

`contains("sign")` and `contains("pay")` match *inside words*, not at a segment boundary.

## What that misclassifies

| effect kind | classified as | why | consequence |
|---|---|---|---|
| `design.review` | `EffectGroup::Sign` | de-**sign** | always parks under `supervised` |
| `assignment.create` | `EffectGroup::Sign` | as-**sign**-ment | always parks under `supervised` |
| `paymnt.send` | `EffectGroup::Spend` | **pay**mnt | treated as money movement |

The `Sign` group is the expensive direction: an effect wrongly in it parks for human approval on every single call, so a company doing design review under `supervised` interrupts an operator for something that carries no consequence at all. That is the exact failure EPIC #558 exists to remove.

`Spend` is the dangerous direction: the gate reads `amount_usd` for a group it should never have been in.

## Why the shortening is not defensible

The longer alternates in the same arms (`payment`, `filing`, `contract`, `x402`) are specific enough to be safe. `pay` and `sign` are the two that carry no word boundary, and they are also the two that are redundant — `payment` already covers the intended case.

## Suggested fix

Match on dotted **segments** rather than raw substrings: split `kind` on `.` and `_` and compare whole tokens. That keeps every intended match (`payment.send`, `document.sign`, `contract.execute`) and drops every one in the table above.

Whatever the fix, it wants a test that walks a list of kinds that must NOT be `Sign`/`Spend` — the defect is a false positive, so a test that only asserts the true positives still pass will not catch a regression.

## Provenance

Found while working #684 (`always_approve` name matching). It is a **different** mechanism in a different file — #684 is about the operator's allow-list, this is about the brain's own classifier — so it is filed separately rather than folded in.

M3gA-Mind:
## Summary

Closes #704.

`effect_group_for` matched its needles with `contains`, which fires *inside* a
word. Three kinds were misclassified, and a fourth turned up while writing the
test table:

| effect kind | was | why | now |
| --- | --- | --- | --- |
| `design.review` | `Sign` | de-**sign** | `Other` |
| `assignment.create` | `Sign` | as-**sign**-ment | `Other` |
| `hire.contractor` | `Sign` | **contract**-or, and `Sign` is tested before `Hire` | `Hire` |
| `paymnt.send` | `Spend` | **pay**-mnt | `Other` |

Both directions are worth naming. A wrong `Sign` parks a routine effect for a
human on **every** call under `supervised` — the standing interruption EPIC #558
exists to remove. A wrong `Spend` hands the effect to the money gate, which then
reads an `amount_usd` that a non-payment effect never carried.

Kinds are now split into `.`/`_` segments, and each needle is matched as a
**contiguous run** of segments.

**Why both delimiters.** `kind` is free `string(1..64)` on the Medulla wire —
`effect_from_frame` copies it verbatim out of the `orch:effect:<kind>` event
name — and live traffic uses each: `payment.received` and `x402.spend` are
dotted, while **`send_dm` is underscore-joined and carries no dot at all**
(exercised in `brain/hosted/test.rs`, `medulla/mock/test.rs`, `medulla/wire/test.rs`).
Splitting on `.` alone would leave `contract_sign.execute`-shaped kinds
unreachable; splitting on `_` is therefore required.

**Why a run, not a per-segment match.** Because `send_dm` is itself two
segments, matching segments individually forces a lone `send` needle into the
first arm — and since `Send` is tested before `Spend`, that would route
`payment.send` into the messaging group and past the `amount_usd` gate. The run
keeps `send_dm` matching its own kind while leaving `payment.send` in `Spend`.
This is asserted, not described (`a_payment_that_sends_is_spend_rather_than_send`).

**One deliberate divergence from the issue's suggestion.** The needle set is
unchanged, including `pay`, which the issue proposed dropping as redundant. That
reasoning was a property of substring matching: under segment matching `pay`
cannot fire inside `paymnt`, so the false positive that motivated removing it is
gone either way, and dropping it would only lose a literal `pay` segment. Losing
a *true* `Spend` is the one error here with real consequence — an ungated
payment rather than an annoyed operator — so it stays. Happy to remove it if you
would rather have the shorter list.

## API Or Behavior Changes

No API change. Classification behaviour changes exactly where it was wrong: the
four rows above, plus any kind that spelled a needle inside a longer word. Every
intended match is preserved and pinned by a table test. `EffectGroup`, the wire
format, and the gate are untouched.

One consequence worth stating plainly: a kind whose needle only ever appeared as
an infix now falls to a *different* group, not merely to `Other` —
`redesign.publish` was captured by `Sign` before reaching the `Publish` arm and
is now `Publish`. That is the fix working, and it is asserted directly.

## Tests

Commands run locally from the repo root:

- `cargo test --features openhuman,tinycortex` — exit 0, **3132 passed, 0 failed**, 3 ignored

Four tests added. The defect is a **false positive**, so the intended-match table
cannot guard it — a bare `contains` satisfies every row of that table. The
must-**not**-match table is the real guard.

Revert-and-check, with the module's counts (8 tests: 4 pre-existing, 4 new):

| Reverted to | Observed |
| --- | --- |
| the original bare-`contains` matcher | 8 ran, **3 failed** — the must-not table (`design.review`), the arm-stealing test (`redesign.publish`), and the intended-match table (on `hire.contractor`) |
| the naive `_`-split (`send_dm` run replaced by a lone `send`) | 8 ran, **2 failed** — including `a_payment_that_sends_is_spend_rather_than_send` |

The second revert is run so the `payment.send` assertion is demonstrably load-
bearing rather than decorative: it is the only thing standing between this fix
and the obvious-but-wrong simplification.

- [x] `cargo build --all-targets` (N/A as spelled — covered by `cargo check --locked --all-features --all-targets`, the stricter lane, which builds every target under every feature)

## Documentation

Rustdoc on `effect_group_for` carries the reasoning: what bare-substring matching
did, why both delimiters are live, and why the needle set — `pay` included —
stayed as it is. The two new helpers (`segments`, `has_segment_run`) document why
a contiguous run is required rather than a per-segment test.

No file in `docs/` pins the needle list or the classification rule, so nothing
there went stale; the supervised-policy groups themselves are unchanged.

### DIFF
--- a/src/brain/medulla/effects.rs
+++ b/src/brain/medulla/effects.rs
@@ -474,25 +474,69 @@ pub(crate) fn effect_from_frame(frame: &EffectFrame) -> Effect {
     }
 }
 
+/// Splits an effect kind into its lowercase segments.
+///
+/// Both delimiters are live, which is why neither can be dropped (issue #704).
+/// `kind` is free `string(1..64)` on the wire — [`effect_from_frame`] copies it
+/// verbatim out of the `orch:effect:<kind>` event name — and real traffic uses
+/// each: `payment.received` and `x402.spend` are dotted, while `send_dm` is
+/// underscore-joined and carries no dot at all.
+fn segments(kind: &str) -> Vec<&str> {
+    kind.split(['.', '_']).filter(|s| !s.is_empty()).collect()
+}
+
+/// Whether `needle`'s own segments appear as a **contiguous run** in `segments`.
+///
+/// A run rather than a single segment because one needle — `send_dm` — is itself
+/// two segments, and it must keep matching the kind of the same name. Matching it
+/// as a run also keeps it from matching `payment.send`, which would be the worst
+/// possible reclassification here: `Send` is tested before `Spend`, so a bare
+/// `send` needle would route money movement into the messaging group and past the
+/// gate that reads `amount_usd`.
+fn has_segment_run(segments: &[&str], needle: &str) -> bool {
+    let want: Vec<&str> = needle.split(['.', '_']).filter(|s| !s.is_empty()).collect();
+    if want.is_empty() || want.len() > segments.len() {
+        return false;
+    }
+    segments.windows(want.len()).any(|window| window == want)
+}
+
 /// Maps a dotted effect kind to its supervised-policy [`EffectGroup`].
+///
+/// # Segments, not bare substrings (issue #704)
+///
+/// Each needle below is matched against whole `.`/`_` segments. It used to be
+/// matched with `contains`, which fires *inside* a word: `de-sign.review` and
+/// `as-sign-ment.create` both classified as [`Sign`](EffectGroup::Sign), and a
+/// misspelled `pay-mnt.send` classified as [`Spend`](EffectGroup::Spend).
+///
+/// Both directions of that were wrong in a way worth naming. A wrong `Sign`
+/// parks a routine effect for a human on *every* call under `supervised` — the
+/// standing interruption EPIC #558 exists to remove — while a wrong `Spend`
+/// hands an effect to the money gate, which then reads an `amount_usd` that a
+/// non-payment effect never carried.
+///
+/// The needle set is deliberately unchanged, including `pay`. Under segment
+/// matching `pay` can no longer fire inside `paymnt`, so the false positive that
+/// made it look redundant is gone; dropping it would only lose a literal `pay`
+/// segment, and losing a *true* `Spend` is the one error here with real
+/// consequence — an ungated payment rather than an annoyed operator.
 pub(crate) fn effect_group_for(kind: &str) -> EffectGroup {
-    let k = kind.to_ascii_lowercase();
-    if k.contains("send_dm") || k.contains("message") || k.contains("email") || k.contains("reply")
-    {
+    let lowered = kind.to_ascii_lowercase();
+    let segments = segments(&lowered);
+    let has = |needle: &str| has_segment_run(&segments, needle);
+
+    if has("send_dm") || has("message") || has("email") || has("reply") {
         EffectGroup::Send
-    } else if k.contains("payment")
-        || k.contains("spend")
-        || k.contains("x402")
-        || k.contains("pay")
-    {
+    } else if has("payment") || has("spend") || has("x402") || has("pay") {
         EffectGroup::Spend
-    } else if k.contains("sign") || k.contains("filing") || k.contains("contract") {
+    } else if has("sign") || has("filing") || has("contract") {
         EffectGroup::Sign
-    } else if k.contains("publish") {
+    } else if has("publish") {
         EffectGroup::Publish
-    } else if k.contains("hire") || k.contains("engage") {
+    } else if has("hire") || has("engage") {
         EffectGroup::Hire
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
`de-**sign**.review` and `as-**sign**-ment.create` are the kind of bug that survives for years because every individual needle looks obviously right, and the table naming each false match by the substring that caused it is the clearest possible way to present it. Matching needles as contiguous **segment** runs is the correct fix, and catching `hire.contractor` — misclassified `Sign` via *contract-or*, and only because `Sign` is tested before `Hire` — while writing the test table is exactly what a test table is for.

Naming both directions of harm is right too, and the `Sign` direction is the one with a live cost: a wrong `Sign` parks a routine effect for a human on **every** call under `supervised`, which is the standing interruption #558 exists to remove.

The delimiter argument is well evidenced rather than assumed — `send_dm` is underscore-joined and carries no dot at all, exercised in three separate test modules, so splitting on `.` alone would leave `contract_sign.execute`-shaped kinds unreachable. Checking live traffic for both shapes before choosing is the step that makes this safe to generalise.

### Question — the fourth row is a loosening, not a fix, and it is the one where direction matters

Three rows in your table remove a false *positive*: `design.review` and `assignment.create` are genuinely not signings, and `hire.contractor` moves to the class it belonged in. Those are strictly better.

`paymnt.send` is different. It is a **misspelled payment**, and today `contains("pay")` accidentally classifies it `Spend` — which routes it to the money gate. After this change it is `Other`, so it does not. The old behaviour was accidentally correct; the new one is precisely wrong on the one class of effect where being wrong costs money.

That matters because `kind` is free `string(1..64)` on the Medulla wire, copied verbatim out of the event name — a field typed to accept typos, from a brain this repository does not control. Substring matching failed *closed* on malformed money kinds; segment matching fails *open*. For `Sign` that trade is clearly right, since over-parking is the harm you are removing. For `Spend` it runs the other way.

Not asking you to keep `contains` — it is wrong for the reasons you give. Worth deciding deliberately, though, and saying which you chose:

- accept it, on the grounds that an unrecognised kind is not a payment and the money gate reading an absent `amount_usd` is its own defect (your argument, and it may well be the right one); or
- keep a cautious arm for the money direction only, so a kind that *segments* to something unrecognised but resembles money still reaches the gate.

Either way the row deserves a different column heading from the other three — it is the one place this PR removes a gate rather than a false alarm, and a reader skimming the table will not notice that on their own.

**Before merging:** base is current, one file, and the lanes are still running. Nothing else outstanding from me.

PR #727, opened 2026-08-11
feat(repo): hand an agent a confined checkout and the two tools (#245)

oxoxDev:
## Summary

The agent tier for #245: an agent with the `repo` grant can check out a bound repository and read a pull request, over a checkout that has no route back to the host's mirror.

**Stacked on #689** (`feat/245-repo-binding`, the operator tier). Until that merges, the diff here includes its commits; rebasing after it lands drops them. Review the eight commits from `9dc20bc4` onward.

Part of #245.

## Problem

Agents carry a full coding toolbelt over a sandbox nothing populates. PR-A shipped bind/mirror/credential/revoke deliberately **without** a checkout layer, because its original `materialize` used `git clone --shared` and @M3gA-Mind reproduced the consequence: a commit followed by `git push origin HEAD:refs/heads/main` advances the host mirror directly. Git resolves that path out of the checkout's own `.git/config` and `.git/objects/info/alternates`, so no command guard protects it, and removing `origin` alone does not close it while the alternates remain.

## Solution

**Clone over `file://` transport from the local mirror, then sever.** The scheme is load-bearing: it forces the fetch/pack path and produces a real object copy. A bare local-path clone hardlinks objects — shared inodes an agent can rewrite — which is the poison case, so a test asserts `st_nlink == 1` on the checkout's object files. After the clone, `origin` is removed and `FETCH_HEAD`/`ORIG_HEAD` deleted (both record the mirror URL), and an alternates file is a hard error.

The attack test performs the reported attack rather than approximating it: materialize, commit a poison file, push to `origin` (must fail — no such remote), push to the explicit `file://` mirror URL (must be refused by the mirror's `pre-receive` hook, and the refusal must come from *that* hook), then assert the mirror's refs and object list are **byte-identical** across both attempts. Isolation is also proved by destruction — delete the mirror entirely and `git fsck --strict` and `git log` still succeed, which a `--shared` clone cannot do.

### The hook was a no-op, and the test caught it

The `pre-receive` hook did nothing on the first machine it ran on: the push succeeded and advanced the mirror. Cause — a `core.hooksPath` in the **host's global git config** silently disables every hook in every repository on that machine. `init_mirror` now also pins `core.hooksPath` to the mirror's own hooks directory in the mirror's repository config, absolute, since git resolves a relative one against the process cwd. That is a real hardening, not a test fixture.

**Stated honestly:** the hook and the mirror are same-uid writable. An agent that escapes shell confinement can tamper with them. This raises the bar from one sanctioned command to deliberate multi-step tampering outside the workspace — it is not a kernel boundary. The primary defence is that the sanctioned-command attack has no address left to aim at. Per-agent uid / read-only bind mount / egress policy remains the platform follow-up `repos.md` already names.

## Submission Checklist

- [x] `cargo clippy --locked --no-deps --all-targets -- -D warnings`, and under `--features github` and `--features openhuman,tinycortex`
- [x] `cargo check --locked --all-targets` and `--all-features --all-targets`
- [x] `cargo test --locked` — 2119 passed, 0 failed, 1 ignored
- [x] `cargo test --locked --features openhuman,tinycortex --tests` — lib 3200 passed / 0 failed / 3 ignored, plus 10 / 1 / 11 / 2 on the integration targets
- [x] `npx vitest run` — 457 passed across 41 files
- [x] `npx tsc --noEmit`, `npm run build`, `scripts/ci/assert-design-tokens.sh`
- [ ] N/A: no manual run against a live forge — the tier is exercised against real `git` over `file://` fixtures, and `repo_pr` against a mock host

19 new Rust tests plus frontend unit coverage. Counts are from unfiltered runs; the repo tests sit behind `openhuman`, where a filter that selects nothing would exit 0 and prove nothing.

## Impact

Fail-closed and opt-in. `repo_checkout`/`repo_pr` are wired only when the agent holds an explicit `repo` grant **and** the manager is present **and** at least one repository is bound; granted-but-unbound warns and wires nothing. Bare `*` does not confer `repo`, matching the search/media/composio precedent. Both tools are `Reach::Consequence`/`PerCall` — they park on every call — and are denied under readonly, parked under supervised and auto, allowed under full. Not feature-gated: with no forge client, `repo_pr` degrades through the existing honest `Unimplemented` answer.

Checkouts are deleted by an RAII janitor when the turn ends — on success, error, redirect exhaustion, and panic-unwind alike — with a boot sweep for crash orphans. `repos/` is excluded from the publish scan: cloned source and PR spills are third-party content, never a deliverable.

Three things the plan got wrong, corrected here:

- **The quota pre-check is reachable less often than planned.** The mirror refresh runs first (necessarily, for correctness), and `fetch`'s own post-fetch cap fires first when the cache alone is already over. The checkout refusal only fires when the cache is under cap and cache + workspace + 2×size is not. The test targets the reachable window and the behaviour is documented at the refusal.
- **The `repo_pr` spill is bounded by the operator tier, not this one** — `pull_request` truncates at 1 MiB before the harness sees it, so "spill the full diff" can only mean "spill what the host obtained". Over 32 KiB goes to a file the reply names; the reply separately says when the host's own cut is also present. The value is that a file survives the tool-result context cut, which an in-band megabyte would not.
- **`unlimited` had to gain the namespace** — widening `GATEABLE_NAMESPACES` to 8 broke two plan tests, because the built-in `unlimited` plan enumerates namespaces.

## Related

- #689 — the operator tier this stacks on
- #245 — the epic
- `docs/spec/runtime/repos.md` — "not in this tier" replaced with what shipped


oxoxDev:
**Union-build finding — a compile break neither PR's CI can see.**

I merged all 14 currently-open PRs into a throwaway branch off `upstream/main` and ran the full gates. Individually every PR is green and `MERGEABLE`; together, this one does not build:

```
error[E0063]: missing field `delegates_to` in initializer of `company::types::Agent`
  --> src/harness/build.rs:1594:30
```

**#722** (`feat/176-recursive-desk-delegation`) adds a required `delegates_to` field to `Agent`. **#727** (`feat/245-agent-tier`) adds a new `ManifestAgent` literal in a test that compiles **only** under `openhuman,mcp,telegram`.

Three layers of blindness, which is why this is worth a comment rather than a quiet fix:

1. #722's CI never sees #727's literal.
2. #727's CI never sees #722's field.
3. **No CI lane builds `openhuman,mcp,telegram`.** So even once both are merged, CI stays green while `main` will not compile under that combo — it surfaces only when someone next runs the gated suite locally, and they will reasonably blame their own work rather than two PRs that landed days earlier.

**The fix is one line**, and it belongs to **whichever of the two merges second** — the first cannot carry it, because on its own branch either the field or the literal does not exist:

```rust
// src/harness/build.rs, in the `ManifestAgent { … }` literal
tools: Vec::new(),
delegates_to: Vec::new(),   // <- add this
budget_usd_daily: None,
```

Verified: with that single line, the full 14-PR union is green — `cargo test` **2177 passed**, `cargo test --features openhuman,mcp,telegram` **3283 passed**, and the console suite **513 passed** (typecheck ×3, build, design tokens).

**Two other cross-PR conflicts** turned up in the same run, neither visible on any PR page. Recording them here so whoever merges does not have to rediscover the resolutions:

- `src/harness/mod.rs` — **#722 vs #713**, the *same struct literal*. #722 adds `delegates_to: Vec::new()`; #713 changes the adjacent `tools: Vec::new()` to `overlay.tools.clone()` (issue #661/L5's per-teammate grant). Taking "ours" **compiles cleanly and silently reverts the grant narrowing** — no error, no failing test. Keep **both** sides.
- `Cargo.toml` — **#730 vs #689/#727**. Both append to `[dev-dependencies]`, so "keep both" looks safe, but #730's block *ends* with a `[[bench]]` table; appending #689's `tokio` dev-dependency after it files `tokio` **inside `[[bench]]`**. Cargo only warns (`unused manifest key: bench.0.tokio`) and the failure lands elsewhere as `no method named 'start_paused' found for struct 'tokio::runtime::Builder'` in `src/runtime/repo_manager/git.rs`. Put the dev-dependency **above** the `[[bench]]` table.

Also confirmed clean: `src/company/mod.rs` (#722 vs #727) is a plain union of two `pub use` name lists, and `docs/spec/runtime/api.md` (#719 vs #702) is the already-known docs conflict.

PR #729, opened 2026-08-11
fix(tasks): restrict an irreversible effect's amount to admins

### LINKED ISSUE #705: approvals: the task-detail read shows an irreversible effect's USD amount to any Member, bypassing #618's redaction
#685 redacted the approval payload and its amount for non-admins. The **task-detail read** exposes the same class of number through a different DTO, and it was not covered. Live on `main` @ `9af6d672`.

## The exposure

`src/server/ops/tasks.rs:766`:

```rust
pub(crate) struct IrreversibleEffect {
    kind: String,
    at_millis: u64,
    /// The USD amount involved, if any — what makes "sent a payment" into
    /// "sent a payment of $2,400".
    #[serde(skip_serializing_if = "Option::is_none")]
    amount_usd: Option<f64>,
}
```

It is assembled unconditionally into `TaskDetail` (`tasks.rs:1290`) and the handler takes a bare `ScopedCompany` with **no role check**:

```rust
async fn task_detail(
    company: ScopedCompany,
    Path(TaskPath { task_id }): Path<TaskPath>,
    Query(query): Query<TaskDetailQuery>,
) -> Result<Json<TaskDetail>, ApiError>
```

So any Member of the company scope can `GET` a card and read the dollar amount of every irreversible effect on it.

## Why this is the same defect as #618/#685, not a new policy question

`src/server/approval_visibility.rs` already states the rule and already has the machinery — `for_principal` gives an `Admin` the amount and a `Member` the redacted form. The task-detail path simply does not call it. The struct's own doc comment shows the intent was scrub discipline (*"Deliberately no payload... a recipient or a message body cannot reach this response even by accident"*) — the payload was scrubbed and the amount was not.

## Second path to the same value

`assemble_detail` is deliberately shared with the export document (`src/server/ops/task_export.rs:58`), and that sharing is documented as *"the export's redaction guarantee"*. That is a good property here: fixing it in `assemble_detail` fixes the export too. But it also means the amount is currently reachable in an exported HTML document as well as in the JSON.

## Suggested fix

Route `IrreversibleEffect`'s amount through the existing `approval_visibility` principal check rather than adding a second redactor. Pin with a test that asserts a `UserRole::Member` request omits `amountUsd` while an `Admin` request retains it — asserted on the serialized JSON, not on the struct, since `skip_serializing_if` means the wire shape is the thing that matters.

## Provenance

Noticed while working #618. It is a different DTO in a different module, so it was not in that PR's scope.

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#705.

`GET …/tasks/{task_id}` projects every irreversible effect a card executed, including `amountUsd`. The handler takes a bare `ScopedCompany` — membership, not authority — so **any Member of the company could read the dollar value of every irreversible effect on any card**.

### This is #618/#685's defect in a different DTO, not a new policy question

The rule already exists and is already written down. `src/server/approval_visibility.rs` states it, and `for_principal` already implements it: an Admin reads an effect's money and arguments, a Member gets the row without them. Issue #618 applied that to `ApprovalSummary` — the effect nobody has signed off **yet**. The identical amount on the effect **once it has run** is projected by `IrreversibleEffect` on a different route, and that route simply never called the machinery.

So there is nothing to decide here about whether Members should see amounts — that was settled by #618, and this only extends the existing answer to the DTO it missed. Concretely, the fix **calls the same predicate** rather than adding a second redactor: two role checks that can disagree about what an operator may see is a worse outcome than the leak either was written to close.

The struct's own doc comment already showed the intent was scrub discipline — *"Deliberately no payload… a recipient or a message body cannot reach this response even by accident"*. The payload was scrubbed; the money was not.

### How it is wired

| piece | what it does |
|---|---|
| `ScopedCompany::may_read_contents` | the **decision**, resolved at the edge by `may_read_approval_contents` |
| `approval_visibility::effects_for_principal` | applies it to the effect list, beside `for_principal` |
| `assemble_detail_with_cursor` | the single call site both readers pass through |

**The decision travels, not the role.** `ScopedCompany` deliberately drops the role at the edge — *"Keep the person, drop the credential… never the email or the role."* Rather than working around that rule, the extractor now resolves the question while the principal still exists and carries the **answer** forward. Nothing downstream can re-derive it differently, because nothing downstream holds the input.

**`effects_for_principal` lives in `approval_visibility.rs`, beside `for_principal`** — not in the tasks module — because the thing that must not fork is the *rule*. It takes the whole list for the same reason `for_principal` does: so a caller cannot redact some entries and forget others.

**Hidden is not absent**, exactly as #618 reasoned. A withheld amount is omitted **and** `amountHidden: true` is set — but only where an amount actually existed. Without the flag a withheld payment and a free tool call are the same bytes on the wire, and a reader cannot tell "this cost nothing" from "you may not see what this cost". Skipped when `false`, so an admin's response — and every response produced before this field existed — serializes byte-identically.

### The export document is covered, and that is why one fix is enough

`assemble_detail` is deliberately shared between the JSON route and the export document (`task_export.rs:58`), and #352 documents that sharing as **"the export's redaction guarantee"**. The redaction is applied **inside** `assemble_detail`, so the exported HTML is covered by construction rather than by a second caller remembering. A reviewer checking only the API route would not see that this single change covers both.

**One accuracy note, having read the template:** the export does **not** leak the amount today — `task_export.rs` never renders `irreversible_effects` at all (the strings `effect`, `irreversible` and `retry` do not appear in that file; it renders task, durations, timeline and lineage). So the live leak was the JSON read only. The fix is still placed in `assemble_detail` for exactly the reason above: it is the shared seam, and a future template that renders effects inherits the guarantee instead of reopening the hole. The tests assert both halves.

## API Or Behavior Changes

- **`GET …/tasks/{task_id}` now redacts `irreversibleEffects[].amountUsd` for a non-admin.** The rows themselves survive — a Member must still see what a retry would re-do; that is #468's stalled-work signal and #351's retry warning, and neither is weakened.
- **New optional field `irreversibleEffects[].amountHidden`** (`true` only when an amount was withheld, skipped otherwise). Additive: an admin's response is byte-identical to before, and a client that reads no such field is unaffected.
- The **platform bearer** is fail-closed, inherited from `may_read_approval_contents` — #618's deliberate choice, not a new one.
- **No console change and no i18n.** #618's own `contentsHidden` is likewise not consumed by the frontend today; this matches that precedent rather than inventing operator wording and 14 locale entries the issue does not ask for. Zero frontend files changed.

## Tests

Two tests added, both driven with **two accounts** — the harness signs every request in as an admin, so a redaction verified only as an admin passes identically against no redaction at all.

**Both assert on the serialized JSON, never on the struct.** `amount_usd` carries `skip_serializing_if`, so the wire shape is the only thing that settles whether the field shipped; a struct-level assertion can pass while the bytes still carry the amount. The HTTP test reads the response body; the shared-seam test calls `serde_json::to_value` on the projection before asserting.

- `a_member_does_not_see_an_irreversible_effects_amount` — drives the real route as Admin and as Member. Admin: `amountUsd == 2400.0`, no `amountHidden`. Member: `amountUsd` **absent from the wire**, `amountHidden == true`, and `kind`/`atMillis` still present. Includes a **control effect that never carried money**, asserting it is *not* flagged as hidden — otherwise "nothing to show" and "not shown to you" collapse.
- `the_export_document_is_built_from_the_redacted_detail` — drives `assemble_detail` (what `export_task` calls) directly as a member **and** as an admin, asserting on the serialized effects both ways, plus a secondary check that the rendered HTML carries no amount.

```
cargo build  --all-targets                                                                     → 0
cargo test   --features openhuman,tinycortex   → 0 — 3130 + 10 + 1 + 11 + 2 = 3154 passed, 0 failed, 3 ignored
```

**Revert-checked**, counts quoted so a run that executed nothing cannot pass for a pass:

| revert | result |
|---|---|
| `assemble_detail` stops calling `effects_for_principal` | **both** FAILED — 0 passed; 2 failed |
| `effects_for_principal` blanks nothing | **both** FAILED — 0 passed; 2 failed |
| `ScopedCompany` hardcodes `may_read_contents = true` | **1 of 2** FAILED — 1 passed; 1 failed |
| `amount_hidden` set `false` instead of `true` | **both** FAILED — 0 passed; 2 failed |

The 1-of-2 is **correct rather than a gap**, and is flagged rather than left to read as a clean sweep: the export test constructs `ScopedCompany` directly in order to drive the shared projection, so by design it does not exercise the extractor. The HTTP test, which does, failed.

**One test was rewritten because it was vacuous.** The first version of the export test asserted only that the exported HTML lacks `2400` — which passes with or without the fix, since the template renders no effects. That is coverage that cannot fail. It was replaced with the member-and-admin assertion on the shared projection described above, which fails on three of the four reverts; the HTML check survives only as an explicitly-labelled secondary guard.

## Documentation

`docs/spec/runtime/events.md` — the executed-effect section stated that the task-detail read *"scrubs by construction"*. True of the payload, and it read as an all-clear that covered the amount too. The amount is retained **deliberately** (the record exists to say what a retry would repeat, and "sent a payment" is a materially weaker warning than "sent a payment of $2,400"), so it is restricted at read time rather than dropped at write time.

The revised section states: the restriction and that it reuses #618's predicate; that the decision travels rather than the role; that hidden is not absent and why the flag is set only where something was withheld; and that the redaction sits in `assemble_detail` so the export inherits it. File is 391 lines, within the 500-line guideline.

### DIFF
--- a/docs/spec/runtime/events.md
+++ b/docs/spec/runtime/events.md
@@ -305,6 +305,32 @@ There is **no payload**. The record is read back onto an operator's screen
 through `GET …/tasks/{task_id}`, which scrubs by construction, so recipients and
 message bodies are never retained in the first place.
 
+**The amount is admin-only** (issue #705). Unlike the payload, the amount *is*
+retained — the whole point of the record is to say what a retry would repeat,
+and "sent a payment" is a materially weaker warning than "sent a payment of
+$2,400". So it is restricted at read time rather than dropped at write time,
+through the *same* predicate that restricts an approval's amount for issue #618
+(`approval_visibility::may_read_approval_contents`). A Member gets the row —
+what a retry would re-do stays visible to whoever is doing the work — without
+the number.
+
+Two consequences worth stating, because both were a defect before:
+
+* **The decision travels, not the role.** `ScopedCompany` deliberately drops the
+  role at the edge, so it carries what the predicate *decided*. Nothing
+  downstream can answer the question differently, because nothing downstream
+  holds the input.
+* **Hidden is not absent.** `amountUsd` is omitted from a Member's response and
+  `amountHidden: true` is set in its place — but only where an amount was
+  actually withheld. Without that flag a withheld payment and a free tool call
+  are the same bytes, and a reader cannot tell "this cost nothing" from "you may
+  not see what this cost". The flag is skipped when false, so an admin's
+  response is byte-identical to before.
+
+The redaction is applied inside `assemble_detail`, which the JSON route and the
+export document (issue #352) both call, so the guarantee covers the exported
+record too rather than depending on two callers remembering.
+
 The task attribution comes from the cycle that ran the effect. Under
 `supervised` an irreversible effect never executes in the cycle that emitted it
 — it parks, and the operator's approval opens a fresh cycle carrying only
--- a/src/server/approval_visibility.rs
+++ b/src/server/approval_visibility.rs
@@ -84,6 +84,48 @@ pub(crate) fn hide_contents(approval: &mut ApprovalSummary) {
     approval.contents_hidden = true;
 }
 
+/// The same rule applied to the **executed** effects on a task's detail read
+/// (issue #705).
+///
+/// #618 restricted the money on an approval — the effect a person has *not yet*
+/// signed off. The identical amount on the effect once it has run was projected
+/// by a different DTO on a different route, and that route was never covered:
+/// any Member could `GET` a card and read the dollar value of every irreversible
+/// effect on it.
+///
+/// It lives here, beside [`for_principal`], rather than in the task module,
+/// because the thing that must not fork is the *rule*. A second redactor in the
+/// tasks route could drift from this one, and two role checks disagreeing about
+/// what an operator may see is worse than the leak either was written to close.
+///
+/// Takes `may_read_contents` rather than the principal because
+/// [`ScopedCompany`](crate::server::ops::ScopedCompany) deliberately drops the
+/// role at the edge, carrying this decision forward instead — see the field's
+/// own note. The decision is still made by
+/// [`may_read_approval_contents`] and nowhere else.
+///
+/// Takes the whole list, like [`for_principal`], so a caller cannot apply it to
+/// some entries and forget others.
+pub(crate) fn effects_for_principal(
+    may_read_contents: bool,
+    mut effects: Vec<crate::server::ops::tasks::IrreversibleEffect>,
+) -> Vec<crate::server::ops::tasks::IrreversibleEffect> {
+    if may_read_contents {
+        return effects;
+    }
+    for effect in &mut effects {
+        // `amount_hidden` only where something was actually withheld: an effect
+        // that never carried money must not report itself as redacted, or
+        // "nothing to show" and "not shown to you" stop being distinguishable
+        // in the direction that matters.
+        if effect.amount_usd.is_some() {
+            effect.amount_usd = None;
+            effect.amount_hidden = true;
+        }
+    }
+    effects
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/src/server/ops/scope.rs
+++ b/src/server/ops/scope.rs
@@ -71,6 +71,21 @@ pub(crate) struct ScopedCompany {
     /// distinction [`CompanyEvent::OperatorMessage`](crate::ports::types::CompanyEvent::OperatorMessage)'s
     /// `by` already draws.
     pub(crate) actor: Option<Actor>,
+    /// Whether this principal may read the *money and arguments* an effect
+    /// carries — issue #618's rule, answered once here for routes that project
+    /// such a field (issue #705).
+    ///
+    /// **The decision, not the credential.** The comment below on `actor` is
+    /// the rule for this struct: the role does not travel. So this carries what
+    /// [`may_read_approval_contents`](crate::server::approval_visibility::may_read_approval_contents)
+    /// *decided*, resolved at the edge where the principal still exists, rather
+    /// than the role that decided it. Nothing downstream can re-derive the
+    /// answer differently, because nothing downstream has the input.
+    ///
+    /// Deliberately the same predicate the Approvals read uses rather than a
+    /// second one. Two role checks that can disagree about what an operator may
+    /// see is a worse outcome than the leak either was written to close.
+    pub(crate) may_read_contents: bool,
 }
 
 impl ScopedCompany {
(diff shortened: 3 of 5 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Framing this as #618/#685's rule reaching a DTO it missed — rather than as a fresh policy question — is right, and it is what makes the fix small. Calling the same predicate instead of adding a second redactor is the important half: two role checks that can disagree about what an operator may see is a worse outcome than the leak either was written to close.

**The decision travels, not the role.** `ScopedCompany` deliberately drops the role at the edge, and rather than working around that rule you resolved the question while the principal still exists and carried the answer forward. Nothing downstream can re-derive it differently because nothing downstream holds the input — that is a stronger property than "every caller remembers to check", and it is the reason this will not rot.

Putting `effects_for_principal` in `approval_visibility.rs` beside `for_principal`, and having it take the whole list so a caller cannot redact some entries and forget others, follows the same instinct. So does keeping `amountHidden` set only where an amount existed: without it a withheld payment and a free tool call are identical bytes, which is the ambiguity #618 already reasoned through.

I checked the completeness rather than taking it on trust, since that is where a redaction change fails:

- `IrreversibleEffect` is constructed in exactly one place — `tasks.rs:1321`, inside `assemble_detail_with_cursor` — and `effects_for_principal` is applied six lines above it. Both readers (`:1164`, `:1181`) pass through that function, so there is no second path to the DTO.
- `runtime::journal::ExecutedEffect` reaches the server layer through that `From` impl and nowhere else, so the underlying record is not exposed by some other route carrying its amount.

The struct's own doc already showed the intent — *"Deliberately no payload… a recipient or a message body cannot reach this response even by accident"* — and you are right that the payload was scrubbed while the money was not. Quoting it is the most convincing part of the case.

### Question — the adjacent surfaces, since that is where this rule goes next

`amount_usd` also appears in `ops/finances.rs`, `graphql/finances.rs`, `graphql/company.rs` and `a2a.rs`. I could not establish from this diff which of those are member-reachable, and it matters for the same reason this PR matters: if a Member can read per-transaction rows from the finances surface, the amount withheld on a task card is available one screen over, and the policy reads as inconsistent rather than enforced.

Aggregate spend is a different claim from "this specific effect cost $2,400", so this is not necessarily a hole — but it is the next place the same leak would live, and you now hold the machinery to answer it uniformly. Worth either confirming those routes are admin-gated, or opening the follow-up while the reasoning is fresh, rather than waiting for a third DTO to be found the same way this one was.

Related: `a2a.rs` is worth a look on its own terms, since an agent-to-agent surface is not a person at all and `may_read_approval_contents` is framed around member-versus-admin.

**Before merging:** base is current and the branch is `CLEAN`. Nothing outstanding from me.

PR #732, opened 2026-08-12
approvals: give a newly provisioned company the `auto` tier (#605)

### LINKED ISSUE #605: approvals: decide whether `auto` becomes the default policy mode
## The decision

`default_policy_mode()` (`src/company/types.rs`) returns `"supervised"`. #560 added `auto` and deliberately did **not** flip this, because the change is one line and the consequence is not. This issue is that decision, on its own, where it can be argued and reverted alone.

```rust
fn default_policy_mode() -> String {
    "supervised".to_string()
}
```

It feeds `#[serde(default = "default_policy_mode")]` on `Policy.mode` and `impl Default for Policy`. So it governs **every company with no `[policy]` block, and every company whose `[policy]` block omits `mode`** — not just newly created ones.

## What changes, concretely

`auto` is strictly more permissive than `supervised`. On next load, for every affected company, these stop parking and start running unattended:

- the agent's own sandbox writes — `file_write`, `edit`, `apply_patch`, `csv_export`, `memory_store`
- outward reads — Composio reads, `mcp_list_tools`, catalogue listings

These continue to park, unchanged: `shell`, `http_request`, `curl`, `web_fetch`, `git_operations`, `workspace_write`, `workspace_create`, `publish_artifact`, `media_generate_*`, `mcp_call_tool`, `mcp_registry_tool_call`, `run_workflow`, `composio_authorize`, Composio writes, anything undeclared, and everything in `always_approve`.

Also unchanged: `readonly` and `full` companies, and any company that names `mode` explicitly.

## Why `auto` is the defensible default

The argument in #560 is that `supervised` is not a *cautious* default, it is an *unusable* one — it parks the agent's own scratch files, so an operator running the shipped default drowns in cards that carry no consequence and learns to approve without reading. A default nobody can live with degrades the gate it was meant to provide: the observed escape hatch is switching the company to `full`, which throws away the boundary that actually mattered.

`auto` is the first default that is both survivable and meaningful — every card it raises is a card about something leaving the company or spending money.

## The argument against, which deserves a real answer

This is a **silent** widening. A company that has been running on the implicit default has an operator who never chose `supervised` and will not be told it changed. They may be relying on seeing a card before the agent writes to its workspace — and unlike a config migration, there is nothing to review.

The counter-argument is that the widened set is exactly the set `Standing::Grantable` already marks as consequence-free *inside* the company, and that no outward or billed effect changes. But "we reclassified it as harmless" is precisely the sentence a security review should be suspicious of when it arrives with no notification.

## Options

1. **Flip it.** One line. Every affected company changes on next load.
2. **Flip it for new companies only** — have provisioning write `mode = "auto"` explicitly into the manifest it generates, and leave `default_policy_mode()` at `"supervised"`. New companies get the good default; existing ones are untouched; the difference is visible in the manifest rather than implied by a serde default. **Recommended.**
3. **Flip it, with a one-time boot notice** naming the change per affected company.
4. **Leave it**, and let #562's console control be how operators reach `auto`.

Option 2 is the one that gets the benefit without the silent part, and it also makes the tier *legible* — an operator reading their own `company.toml` sees which tier they are on instead of inferring it from an absent block.

## Prerequisites

- #560 (the `auto` tier itself) — merged or in review.
- Worth sequencing after #562, so an operator who dislikes the new default has a way to change it that is not hand-editing TOML.

## Acceptance

- A deliberate, recorded decision among the options above — not a drive-by edit.
- If the default changes for existing companies, the migration story is written down in `docs/spec/runtime/manifest.md` and `docs/spec/company-brain/approvals.md`.
- Whatever is chosen, the shipped presets under `companies/` declare `mode` **explicitly** rather than relying on the default, so the tier a shipped company runs on is readable from its manifest.

Related: #560, #558, #562.

M3gA-Mind:
## Summary

Issue #605 asked whether `auto` becomes the default policy mode. The answer here is **yes for a new company, never retroactively for an existing one** — which is the issue's own recommended option 2.

`auto` is delivered by **writing the tier into a new company's manifest at provisioning**. `default_policy_mode()` stays `"supervised"`, deliberately.

Two knobs, and the distinction is the whole decision:

- **`default_policy_mode()`** — what a manifest with no stated `mode` *parses* to. Stays **`supervised`**.
- **`PROVISIONED_POLICY_MODE`** — what provisioning *writes into* a new company's manifest when its author stated no tier. **`auto`**.

## Why `default_policy_mode()` is not flipped

A reader who knows this issue's title will expect the one-line flip, so the argument is written down rather than left as an omission — in the `Policy::mode` doc comment, in `approvals.md`, and below.

That value answers for **every manifest ever parsed**, including the re-parse of a company that has been running for months. Moving it re-tiers existing companies rather than creating them differently. Two things go wrong:

1. **The obvious one.** Every company with no `[policy]` block silently widens on its next load. Nobody edited anything and nothing is written down.

2. **The one that is not obvious, and is worse.** The persisted record stores the *defaulted* value, not what the file said. So a silent `company.toml` that parsed to `supervised` last boot parses to `auto` this boot — and `carry_policy_override` (`runtime/builder.rs:232`) is literally `(previous_seed == next_seed).then(...)` over the whole `[policy]` block, with `Policy` deriving `PartialEq`. It reads that as version control having spoken, and **discards the operator's console `[policy]` override — including one that had *tightened* the tier to `readonly`.**

A migration that rewrites *stored* manifests cannot prevent (2): the right-hand side of that comparison is a fresh parse of the operator's own file, which must not be rewritten.

Writing the tier at provisioning avoids both, and makes the tier legible: a platform-provisioned tenant has no `company.toml` anywhere, so the stored manifest is the only place its tier is written at all.

**This was never a one-line change.** Revert 5 below shows a pre-existing test, `company::manifest::tests::defaults_are_prosumer_safe` (`manifest.rs:454`), already pins the parse default at `supervised`.

## API Or Behavior Changes

**New:** `crate::company::PROVISIONED_POLICY_MODE`.

**Behaviour:** a company provisioned through `POST /api/v1/companies` whose manifest states no `[policy].mode` is now recorded on `auto` instead of `supervised`. One whose manifest **does** state a tier is untouched, whichever tier it is.

**Unchanged:** every existing company, on every backend. Nothing re-tiers anything already stored, and `default_policy_mode()` is the same value it was.

### Creation paths — exactly one needed the change

1. `serve --company <dir>` (`bin/opencompany.rs:273`) — reads a `companies/*/company.toml`; every shipped preset now declares `mode`.
2. The desktop app (`desktop.rs:138`) — `include_str!` of the same presets.
3. **`POST /api/v1/companies`** (`server/provision.rs`) — the only creation path with no template behind it. This is where the write happens.

There is no `createCompany` GraphQL mutation (every `create_company_*` symbol is a *workflow* creator). `runtime/rebuild.rs` rebuilds from the persisted record, so a provisioned tier survives restarts.

### Why the check reads the raw TOML

`declares_policy_mode` parses the submitted **text**, not the resulting `CompanyManifest`. `Policy::mode` is a plain `String` behind a serde default, so a parsed manifest *always* carries a mode and can never be asked what its author actually wrote. Only the text distinguishes "the author chose `supervised`" from "the author said nothing".

## Tests

- `a_provisioned_company_with_no_stated_tier_is_recorded_on_auto` — the new-company half. It also asserts `PROVISIONED_POLICY_MODE != Policy::default().mode`, so it cannot pass vacuously if the two ever coincide.
- `a_provisioned_company_keeps_whatever_tier_it_states` — **preserve, never widen**. Walks `POLICY_MODES` rather than spot-checking, so a fifth tier cannot escape the guarantee without failing a test. `supervised` is the sharp case and the reason it is a walk: it is the value the parse default *also* produces, so a broken declared-mode check is invisible on every other tier and caught only there.

Both read the **stored record**, not the HTTP response — the record is the manifest a rebuild re-reads, and the only place a provisioned tenant's tier is written down.

### Verified locally

- [x] `cargo clippy --all-targets -- -D warnings` — ran as `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`; exit 0, no lints.
- [x] `cargo build --all-targets` — N/A as written; ran the stronger `cargo check --locked --all-features --all-targets`, exit 0, because `mongodb.rs` / `sqlite.rs` build `CompanyRecord` from a row and neither the default nor the gated lane sees it.
- [x] `cargo test` — ran as `cargo test --features openhuman,tinycortex`; exit 0. **3130 passed, 0 failed, 3 ignored** in the lib target, plus 5 integration targets green.

### Revert-and-check

Each row applies one revert, proves the file actually changed (a pattern matching nothing is a silent no-op that would score green), runs the full gated lib suite, and restores.

| # | Revert | Tests failed |
|---|---|---|
| 1 | remove the provisioning write | 1 — `..._recorded_on_auto` |
| 2 | never detect a declared tier | 1 — `..._keeps_whatever_tier_it_states` |
| 3 | always claim a tier was declared | 1 — `..._recorded_on_auto` |
| 4 | set `PROVISIONED_POLICY_MODE` to the parse default | 1 — `..._recorded_on_auto` |
| 5 | flip the parse default to `auto` | **2** — plus `defaults_are_prosumer_safe` |

Row 4 is the vacuity guard doing its job. Row 5 is the change this PR declined to make, and it fails a test that was already in the tree.

## Documentation

- `docs/spec/company-brain/approvals.md` — a new **"Which tier a new company gets"** section: the two knobs, why the parse default does not move, and the `carry_policy_override` mechanism spelled out.
- `docs/spec/runtime/manifest.md` — the `[policy]` entry now distinguishes the parse default from what a new company gets, and links to the section above.
- Both remain within the repo's 500-line cap (`approvals.md` is at exactly 500).

## Notes for review

The `carry_policy_override` behaviour described above is **latent, not currently firing**: with the parse default unchanged, a silent manifest parses to `supervised` on both sides of the comparison and the override is correctly preserved. It is a tripwire armed by any future change to a `[policy]` serde default — `default_policy_mode` or `default_always_approve`. Recorded as a comment on #605 rather than silently fixed here, since it is out of this PR's scope.

Closes #605

## Summary by CodeRabbit

* **New Features**
  * New companies now default to automatic approval mode when no policy tier is specified.
  * Explicitly configured approval tiers remain unchanged.
  * Automatic policies allow unattended sandbox writes and outward reads while retaining approval requirements for external actions and spending.

* **Documentation**
  * Updated policy and manifest documentation to clarify provisioning defaults and parsing behavior.

* **Tests**
  * Added coverage verifying default policy assignment and preservation of explicitly selected tiers.

### DIFF
--- a/companies/agentic_accounting_firm/company.toml
+++ b/companies/agentic_accounting_firm/company.toml
@@ -36,3 +36,13 @@ description = "Assemble documentation for audits."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["month_end_close"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
--- a/companies/agentic_consultation_firm/company.toml
+++ b/companies/agentic_consultation_firm/company.toml
@@ -63,3 +63,13 @@ description = "Turn strategy into an execution roadmap."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["engagement_pipeline"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
--- a/companies/agentic_customer_support/company.toml
+++ b/companies/agentic_customer_support/company.toml
@@ -36,3 +36,13 @@ description = "Process refunds within policy."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["ticket_pipeline"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
--- a/companies/agentic_design_studio/company.toml
+++ b/companies/agentic_design_studio/company.toml
@@ -53,3 +53,13 @@ description = "User testing and design validation."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["design_pipeline"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
--- a/companies/agentic_enterprise_sales/company.toml
+++ b/companies/agentic_enterprise_sales/company.toml
@@ -41,3 +41,13 @@ description = "Nurture and follow up on pipeline."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["sales_pipeline"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
--- a/companies/agentic_game_business/company.toml
+++ b/companies/agentic_game_business/company.toml
@@ -46,3 +46,13 @@ description = "Resolve player issues and refunds."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["liveops_pipeline"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
--- a/companies/agentic_game_studio/company.toml
+++ b/companies/agentic_game_studio/company.toml
@@ -46,3 +46,13 @@ description = "Market and promote the launch."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["game_build_pipeline"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
--- a/companies/agentic_influencer_business/company.toml
+++ b/companies/agentic_influencer_business/company.toml
@@ -51,3 +51,13 @@ description = "Source and negotiate sponsorships."
 # directory (one file per id), not inline here.
 [workflows]
 enabled = ["content_pipeline"]
+
+# Approval tier for this company (issue #605). Declared explicitly rather than
+# left to the serde default, so the tier a shipped company runs on is readable
+# from its own manifest instead of inferred from an absent block.
+#
+# `auto` runs the agent's own sandbox writes and outward reads unattended and
+# still parks everything that leaves the company or spends money. `always_approve`
+# is omitted, so it keeps its default.
+[policy]
+mode = "auto"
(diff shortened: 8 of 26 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Splitting the question into two knobs — what a silent manifest *parses* to, versus what provisioning *writes* — is the whole review, and the second hazard you found is the one that makes it more than a style preference:

> `carry_policy_override` is literally `(previous_seed == next_seed).then(...)` over the whole `[policy]` block … It reads that as version control having spoken, and **discards the operator's console `[policy]` override — including one that had *tightened* the tier to `readonly`.**

A default flip silently discarding an operator's deliberate *tightening* is a security regression that no amount of testing the new default would have surfaced, because the bug is in the comparison rather than in the value. And you are right that a migration over stored manifests cannot fix it — the right-hand side is a fresh parse of the operator's own file, which must not be rewritten. That argument is worth keeping in the doc comment exactly where you put it.

I checked the two things that would make this dangerous, and both are right:

- **The write is guarded.** `if !declares_policy_mode(&manifest_toml)`, so a provisioned company that states `readonly` keeps `readonly`. A bare assignment here would have overwritten a deliberately tightened tier at creation.
- **The guard reads the raw TOML, structurally.** `toml::from_str::<toml::Value>` then `policy.mode` — not the parsed struct, which would already carry the serde default and make "stated" indistinguishable from "defaulted", and not a substring scan, which would false-positive on the `mode` key that appears under other sections in shipped manifests. Given you have just spent #725 on exactly that class of mistake, it is good to see the careful version here.

Making every shipped preset declare `mode` explicitly, rather than leaning on a default, is the right accompanying move — it means the presets say what they are instead of inheriting a value that this PR is in the business of changing.

`auto` is also the right tier to land on: sandbox writes and outward reads run unattended while anything that leaves the company or spends on submit still parks, so a new tenant gets a usable company without the thing an operator would most regret being unattended.

### Question — is a provisioned tenant *told* what tier it got?

You make the case that writing the tier explicitly is what makes it legible: a provisioned tenant has no `company.toml`, so the stored manifest is the only place its tier is written at all, and leaving it implicit would make it unreadable rather than merely unstated. Agreed — but legible is not the same as announced.

The operator of a newly provisioned company is the one person who did not choose this tier and has no file to read it from. #659 shipped the console surface, so they *can* find it. The question is whether anything puts it in front of them — the provisioning response, or first-run console copy — or whether a hosted tenant simply runs at `auto` until someone thinks to look.

Not a blocker, and possibly already covered by the console work. But a default that loosens what runs unattended is the one worth stating out loud rather than merely recording, and you now have the surface to state it on.

**Before merging:** five checks still pending and the branch is 11 commits behind. Nothing in that range looks related, but today has twice produced a green run against a base that moved underneath it.

Issue #700, opened 2026-08-11
workspace: pre-#570 companies carry an empty folder per agent, with no way to clean them up


oxoxDev:
## Problem

Every company provisioned before #570 permanently carries one empty folder per agent under `Agents/`, and nothing will ever clean them up.

Observed live on the `smoke1` staging tenant: **17 folders under `Agents/`, all empty**, created in a single burst on 2026-08-10 — the root at `18:01:51.933` and the children at `18:01:51.967`, `18:01:52.001`, `18:01:52.036`, roughly 34 ms apart. That timing is the signature of eager provisioning at boot; folders minted by producers appear at scattered times as work actually happens.

The tenant's seeded content (`Playbooks/`, `README.md`, `Product/`, `Standards/`) dates from 2026-07-20 and is stamped `operator`. The agent folders are stamped `agent` and postdate it by three weeks — a #551-era image provisioning `Agents/<id>/` for every roster member at boot, whether or not that member ever produced anything.

**This is not a bug in current code.** #570 stopped creating them and deliberately chose no backfill and no migration, so existing nodes are left alone. That decision was right at the time. This issue is about what it looks like a month later, on real tenants, now that a second change has raised the cost.

## Why it matters more than it did

Three things have landed since:

- **#570** made a folder mean *"this member produced something"*. On a pre-#570 tenant it means nothing, so the two populations are indistinguishable to a human reading the tree.
- **#552** publishes deliverables into `Agents/<agent>/<task-id>/`, so an empty folder now reads as "this agent has produced nothing" — which is a claim, and on these tenants it is unfounded.
- **#607** made the tree searchable with a bounded result set (`MAX_SEARCH_RESULTS = 50`). Empty folders are name-matchable, so they compete for slots against files that actually contain something.

Eight of the seventeen on `smoke1` are also ULID-named (see #686), so the tab opens on rows like `Agents/019fad5ada20-000000000003/` that are both empty and unreadable.

## Proposal

A one-time, opt-in sweep that removes **provably empty** agent folders — and nothing else.

The safety condition has to be exact, because the failure mode is deleting someone's work:

- the node is a direct child of the `Agents/` root, **and**
[...]

PR #742, opened 2026-08-12
test(workspace): unbreak main — seed the sweep's ambiguous root instead of creating it (#711 × #719)

M3gA-Mind:
## Summary

**`main` is red on the default `Rust` lane. This is one of the two reasons.**

```
company::workspace_sweep::tests::an_ambiguous_root_is_refused_and_removes_nothing
called `Result::unwrap()` on an `Err` value:
    Conflict("workspace already contains `Agents` in that folder")
```

A latent conflict between two already-merged PRs, each green against a `main` that lacked
the other:

- **#711 (#666)** made the filesystem store refuse a write whose resolved path another node
  already holds — two siblings under one name are one path on that backend.
- **#719 (#700)** added `src/company/workspace_sweep.rs`, whose ambiguity test builds its
  fixture by calling that same store's `create` twice with `Agents`.

So the fixture now asks the store for exactly the state the store exists to refuse, and
fails in setup — before reaching the sweep at all.

**The sweep's behaviour is not in question and is not changed here.** An ambiguous root is
still refused, still names `Agents`, still removes nothing. What changes is how the test
*obtains* an ambiguous tree: it writes the workspace index directly instead of requesting an
illegal create.

That is also the more honest fixture. The sweep's job is to fail closed on an ambiguous tree
it **finds** — one written before #666's refusal existed, or one an id-keyed backend can
still represent legally — not on one it was able to create a moment earlier. #711 made the
same change to the two `workspace_scaffold` fixtures for the same reason; this uses the same
helper shape, so the two modules seed the case identically.

Verified on pristine `upstream/main` @ `a9c75878` (none of this branch's commits) before
writing anything, so the failure is attributed rather than assumed.

## API Or Behavior Changes

None. One test fixture. No production code is touched — the diff is confined to
`#[cfg(test)]`.

## Tests

- `cargo test --locked` — **2202 + 10 + 1 passed, 0 failed** (this is the lane `main` is red
  on; it is green with this change)

**Revert-and-check**, because a reseeded fixture is exactly the kind of change that can
quietly stop testing anything. With the sweep's ambiguity refusal replaced by "return no
candidates" (and the refusal's absence confirmed by `grep` before running):

| reverted | test | result |
|---|---|---|
| `Found::Collision(_) => return Ok(Vec::new())` | `an_ambiguous_root_is_refused_and_removes_nothing` | 1 selected, **1 FAILED** |

The test still discriminates: it fails when the sweep stops refusing, which is what it is for.

### The other reason `main` is red

`harness::policy::tests::both_approval_paths_agree_on_the_same_always_approve_list` also
fails on pristine `main` — a separate latent conflict between #660 (#338) and #715 (#684),
being fixed in its own PR. It is untouched here and is not weakened or skipped. `main` needs
both fixes.

- [x] `cargo clippy --all-targets -- -D warnings` — run as `--locked --no-deps --features openhuman,tinycortex --all-targets`; `--no-deps` keeps pre-existing `vendor/tinyagents` lints out, which CI does not see either
- [x] `cargo build --all-targets` — covered by `check --locked --all-features --all-targets` plus the full `cargo test --locked` build above
- [x] `cargo test`
- [x] N/A — no i18n keys: the diff is one Rust test fixture, with no user-facing text

## Documentation

None needed: no behaviour, API or architecture changes. The reason the fixture is seeded
rather than created is recorded on the helper itself, where the next person tempted to
"simplify" it back into a `create` loop will read it.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for ambiguous workspace states involving duplicate roots.
  * Confirmed ambiguous workspaces are refused without deleting existing data.

### DIFF
--- a/src/company/workspace_sweep.rs
+++ b/src/company/workspace_sweep.rs
@@ -254,6 +254,37 @@ mod tests {
         (dir, ops)
     }
 
+    /// Seeds root folders that share a name by writing the workspace index
+    /// directly.
+    ///
+    /// The filesystem store refuses to *create* two siblings under one name,
+    /// because on that backend they resolve to one path (issue #666). The tree
+    /// below is the one this test needs to find rather than to build: an index
+    /// written before that refusal existed, or one an id-keyed backend can
+    /// still represent legally. Going through `create` would only re-assert the
+    /// store's refusal and never reach the sweep.
+    async fn seed_duplicate_roots(
+        dir: &std::path::Path,
+        company: &CompanyId,
+        name: &str,
+        ids: &[&str],
+    ) {
+        let index: std::collections::HashMap<String, WorkspaceNode> = ids
+            .iter()
+            .map(|id| ((*id).to_string(), folder(id, name, None)))
+            .collect();
+        let bundle = crate::store::Bundle::new(dir.to_path_buf(), company);
+        tokio::fs::create_dir_all(bundle.workspace_dir())
+            .await
+            .expect("workspace dir");
+        tokio::fs::write(
+            bundle.workspace_index_json(),
+            serde_json::to_vec(&index).expect("index json"),
+        )
+        .await
+        .expect("seed index");
+    }
+
     /// The sorted names in the tree, so an assertion says what survived rather
     /// than counting.
     async fn names(ws: &Arc<dyn WorkspaceStore>, company: &CompanyId) -> Vec<String> {
@@ -542,13 +573,9 @@ mod tests {
     /// one and deleting beneath it.
     #[tokio::test]
     async fn an_ambiguous_root_is_refused_and_removes_nothing() {
-        let (_dir, ws) = store().await;
+        let (dir, ws) = store().await;
         let company = CompanyId::new("odd");
-        for id in ["dup-a", "dup-b"] {
-            ws.create(&company, &folder(id, AGENTS_ROOT, None), None)
-                .await
-                .unwrap();
-        }
+        seed_duplicate_roots(dir.path(), &company, AGENTS_ROOT, &["dup-a", "dup-b"]).await;
         ws.create(&company, &folder("stray", "ceo", Some("dup-a")), None)
             .await
             .unwrap();


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Right fix, right reason, and it unblocks `main`. Approving.

The fixture was asking the store for exactly the state the store exists to refuse, so it failed in setup before reaching the sweep at all. Seeding the index directly is not a workaround — it is the more honest fixture, and your sentence for it is the one that matters:

> The sweep's job is to fail closed on an ambiguous tree it **finds** … not on one it was able to create a moment earlier.

A test that builds its subject through the API that now forbids it is testing the guard twice and the sweep never. Matching the helper shape #711 already used for the two `workspace_scaffold` fixtures means the two modules seed the case identically, which is worth more than either fixture being individually clever.

Verifying on a pristine `upstream/main @ a9c75878` before writing anything — so the failure is attributed rather than assumed — is the step that makes "latent conflict between two merged PRs" a claim rather than a guess. Confining the diff to `#[cfg(test)]` keeps the blast radius where it belongs.

**0 major. Approving.**

### Worth strengthening one line of your reasoning

You justify the fixture as representing "an index written before that refusal existed, **or one an id-keyed backend can still represent legally**". That second clause is not hypothetical — it is the current state of the codebase, and it makes this sweep load-bearing rather than legacy-only.

`reject_sibling_name_collision` lives in `fs_ops.rs`, and #711 did not touch `src/store/conformance.rs`. SQLite and MongoDB key nodes by id, so two same-named siblings are two rows on those backends and nothing refuses them. **Hosted tenants run MongoDB.** So an ambiguous tree is not merely a pre-#666 artifact an old filesystem deployment might carry — it is a state the deployed backend can reach today, which is exactly why the sweep needs to fail closed on finding one.

I raised the same gap on #711 as a minor: the port contract now differs by backend and `conformance.rs` does not pin which answer is correct. This PR is the second place that divergence has surfaced in a day. Worth resolving it there — either the refusal is contract and belongs in conformance with implementations in all three backends, or it is an fs implementation detail and the port doc should say so — because right now two modules are reasoning about it from comments rather than from a pinned contract.

**Before merging:** base is current and this is one test fixture. `main` is red on two independent counts — this and #741 — so both want landing rather than sequencing.

M3gA-Mind:
## Closing in favour of #741, which carries this fix.

Not withdrawn — **folded in**. The sweep fixture change now lives in #741 at
`src/company/workspace_sweep.rs`.

### Why one PR rather than two

`main` was red for **two independent reasons**, and each PR fixed one and lacked the other:

| lane | failing test | fixed by |
|---|---|---|
| `Rust` (default) | `an_ambiguous_root_is_refused_and_removes_nothing` | this PR |
| `Rust (openhuman, tinycortex)` | `both_approval_paths_agree_on_the_same_always_approve_list` | #741 |

So neither could ever go green on its own: whichever branch you stood on, the other break was
still there, and every other open PR was queued behind both. A single PR carrying both fixes
is the only shape that can be **demonstrated** green rather than argued green — which #741
now is, on all three Rust lanes with zero failures.

That is the whole reason this one closes. The change itself was never in question.

### What this PR established

That `main` was broken a **second** time, by a collision nobody had looked for: my own merged
#711 (#666 — the filesystem store refusing a write whose resolved path another node holds)
against #719 (#700 — the empty-agent-folder sweep), whose ambiguity test built its fixture by
calling `create` twice with `Agents`. Each was green against a `main` that lacked the other;
they merged **four minutes apart** (#719 at 06:21:22Z, #711 at 06:25:39Z), which is how both
passed review and still broke on landing.

The prevailing account at the time was that the red was the #660 × #684 policy conflict and
that everything else merely inherited it. That account was complete for the gated lane and
wrong for the default one. The failure was verified on pristine `upstream/main` @ `a9c75878`,
with none of this branch's commits, before any fix was written.

One detail worth keeping, because it caused a reporting disagreement: this test is **not
feature-gated** — `pub mod workspace_sweep;` and `pub mod fs_ops;` are both unconditional, and
the test module is a bare `#[cfg(test)]`. It therefore fails identically under every feature
configuration, and its result can only vary by **base commit**, never by flags. Any tree based
before 06:25:39Z passes it cleanly, which is the likely source of a "gated lane is clean"
reading.

### The fix, for anyone reviewing it in #741

The sweep's behaviour is unchanged: an ambiguous root is still refused, still names `Agents`,
still removes nothing. Only the fixture changed — it **seeds** the ambiguous index directly
instead of asking the store to create a state the store exists to refuse. That is also the
truer case, since the sweep's job is to fail closed on an ambiguous tree it *finds* (an index
written before #666, or an id-keyed backend), not one it could have created a moment earlier.
#711 made the same change to the two `workspace_scaffold` fixtures, so both modules now seed
the case identically.

Revert-checked so the reseeded fixture is not hollow: with the sweep's ambiguity refusal
replaced by "return no candidates", the test fails — 1 selected, 1 failed.

Branch retained.

PR #747, opened 2026-08-12
fix(desktop): require https for a credentialed host, permit anonymous http

### LINKED ISSUE #731: desktop: a remote connection over plain http sends its credential in cleartext
## Where

- `src-tauri/src/proxy/mod.rs` — `ProxyRegistry::upsert` (registration), `request` / `subscribe` (credential attachment, `src-tauri/src/proxy/mod.rs:374`).
- `frontend/src/api/transport/index.ts` — `isAddressableBaseUrl`, the console-side twin of the same rule.

## What

Registration accepts any `http:` or `https:` url with a host:

```rust
let addressable = reqwest::Url::parse(&connection.base_url)
    .is_ok_and(|url| matches!(url.scheme(), "http" | "https") && url.has_host());
```

and every later request attaches the connection's stored credential:

```rust
Credential::Device(value) => builder.header("x-opencompany-session", value),
```

So a desktop pointed at `http://192.168.1.20:8080` sends its session token over the wire in cleartext, to anyone on that network path. `isAddressableBaseUrl` admits the same pair on the console side.

Nothing here is wrong for **loopback** — `http://127.0.0.1:<port>` is how the embedded host is reached, and it never leaves the machine. The gap is remote hosts, where `http:` and `https:` are currently treated as equally acceptable.

## Proposed rule

Require HTTPS for a non-loopback host; permit `http:` only for loopback. Applied in both places, since they are two copies of one decision — a console-side check alone is bypassed by anything that reaches the proxy directly, and a proxy-side check alone lets the console offer a connection it cannot use.

## The part that needs a decision, not just a patch

**This refuses setups that work today.** A LAN host, a home-lab box, a staging deployment reachable over plain HTTP, and a colleague's machine on the office network all stop being addable. Local development is unaffected (loopback stays permitted), and so is anything already on HTTPS.

Options, roughly in order of how much they cost the people affected:

1. **Refuse outright.** Simplest, safest, and the one that generates support questions from anyone running an internal host without a certificate.
2. **Refuse only when a credential would be attached.** An anonymous connection over HTTP is no worse than reading a public page; the exposure is entirely in the credential. Narrower, and it keeps read-only LAN use working.
3. **Warn and require explicit confirmation**, recording the choice on the profile. Keeps every current setup working, at the cost of a dialog nobody reads.

I would take (2) as the default, but that is a product call rather than a security one, and it belongs to whoever owns the desktop's connection UX.

## Acceptance

- A remote `http:` base url is refused (or gated, per the option chosen) at **both** `ProxyRegistry::upsert` and `isAddressableBaseUrl`.
- Loopback `http:` still registers, and the embedded host still works — `desktop-connections.spec.ts` and `desktop-embedded-identity.spec.ts` stay green.
- E2E coverage for the rejection path, alongside the existing loopback coverage.
- The refusal names the reason, so an operator gets "this host is not encrypted" rather than the existing "could not be reached" — the two failures look identical otherwise, which is the confusion #613 was about.

## Related

Raised by CodeRabbit in review of #692, which touched `upsert`'s validation to refuse a base url naming no host. Split out rather than folded in: #692 is a desktop-bootstrap fix that is green and approved, and this is transport policy that changes which existing setups keep working — it deserves its own review rather than riding along.

CodeGhost21:
Closes #731.

## The rule

**A credential travels over HTTPS, or to a host on this machine, and nowhere else.** Gated on the credential rather than on the scheme — option (2) from the issue — so an anonymous connection to an unencrypted LAN, home-lab or staging host still registers and still reads. Nothing is exposed by reading one that a passer-by could not have asked the host for themselves; the exposure is entirely in the credential.

Loopback is exempt because `http://127.0.0.1:<port>` is how the embedded host is reached, on an ephemeral port that can never carry a certificate. `localhost` and its subdomains come with it, per RFC 6761. The private ranges are deliberately **not** exempt — an office LAN is precisely where someone else is on the path.

## What the issue did not account for

Three things turned up while tracing it, and all three are in the diff:

1. **The pairing claim bypasses the proxy entirely, and it is the worse leak.** `oc_pair_device` builds its own client and never touches the registry, so `upsert`'s refusal would not have covered the one exchange in which a session token is *created* rather than replayed — the code goes out in the request, the token comes back in the response body. Its client also used `reqwest`'s default redirect policy: a 307 from an HTTPS base to an HTTP one re-sends the claim, body and all, over the wire the check just refused. Now `Policy::none()`, as `ProxyRegistry`'s client already was.

2. **`isAddressableBaseUrl` is not on the add-host path.** It has two callers — the bootstrap add and restore-time pruning — and the "Add a host" dialog uses neither. It is also the wrong lever here: it must not narrow, or the web build's same-origin discovery breaks. Left unchanged; the console-side twin is the new credential-aware `mayCarryACredential`, kept in the same file beside it.

3. **"The refusal names the reason" needed more than a check.** The core's refusal is discarded three times on the way up — `ProxyErrorWire` drops the transport's words by design, `registerConnection` catches the rejection into `console.error`, and `client.ts` flattens every transport failure into `cannot reach the company host at …`. So `probe` answers the same question console-side and marks the row `down` with the reason before contacting anything.

## Behaviour changes

| Setup | Before | After |
|---|---|---|
| Paired device on `http://<lan>` | session in cleartext on every request | refused, row says why |
| Anonymous `http://<lan>` | works | works, unchanged |
| Pairing against `http://<lan>` | code out and token back in cleartext | refused before a socket opens |
| Loopback `http://127.0.0.1:<port>` | works | works, unchanged |
| Anything on HTTPS | works | works, unchanged |

An existing profile paired over plain HTTP stops connecting and says why, rather than being silently dropped.

## Verification

Beyond the test suites, I ran the leak on a real socket — a capture server bound to this machine's actual LAN address, driven by the same test against pre-fix and post-fix source.

Pre-fix, the token came off the wire verbatim:

```
[wire] REGISTERED a credentialed connection to http://192.168.29.56:55404
GET /api/v1/companies HTTP/1.1
x-opencompany-session: acme.SUPER-SECRET-SESSION-TOKEN
[wire] VERDICT: connections=true leaked_secret=true
```

Post-fix, nothing connected at all:

```
[wire] refused at registration: this host is not encrypted, so a credential cannot be sent to it
[wire] VERDICT: connections=false leaked_secret=false
```

Against a live `opencompany serve`, `claim()` reached the host over loopback and was refused at the same host's LAN address. The real claim endpoint answers directly with no redirect, so `Policy::none()` costs nothing.

Commands run locally:

- `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` in `src-tauri/` — 71 lib + 12 acp + 7 proxy_parity, all pass
- `npm test` — 469 pass across 41 files
- `npm run typecheck`, `typecheck:unit`, `typecheck:e2e` — clean
- `npm run e2e` — 108 pass, 8 skipped (live-brain gated), 0 fail

`desktop-connections.spec.ts` and `desktop-embedded-identity.spec.ts` stay green, with two tests added: the rejection path, and the anonymous-HTTP path this option keeps working. I checked the first has teeth by neutering the `probe` guard and watching it fail.

## Unrelated bug found while testing

Not fixed here, and worth its own issue: **desktop device pairing has never worked against a real host.** `oc_pair_device` posts to `{base}/api/v1/devices/claim`, which has no route — the real one is company-scoped (`src/server/users/devices.rs:97` via `public_scoped`).

```
POST /api/v1/devices/claim         -> 404 Not Found
POST /api/v1/company/devices/claim -> 400 {"error":"invalid or expired pairing code"}
```

Present on `main` and independent of this change, which touches the guard and the client builder, not the path. Left out because the desktop has no company id at pairing time, so choosing between the single-company alias and the scoped form is a real decision rather than a typo fix.

## Summary by CodeRabbit

* **Security Enhancements**
  * Credentialed connections now require HTTPS or a loopback address.
  * Remote HTTP connections carrying credentials are blocked before network requests.
  * Redirects are disabled during device pairing to prevent unsafe credential transport.
  * Anonymous HTTP connections remain available where permitted.

* **Bug Fixes**
  * Improved validation for localhost, IPv4, and IPv6 loopback addresses.
  * Added clear errors when a connection or pairing attempt uses an unsafe transport.

### DIFF
--- a/docs/spec/runtime/desktop.md
+++ b/docs/spec/runtime/desktop.md
@@ -63,6 +63,10 @@ second is what someone types into "Add a host". Parsing is not enough either —
 `URL` accepts `tauri://localhost` and `file:///x`, and neither is a company
 host — so the check is `http:` or `https:`.
 
+Whether that host may then be *trusted with a secret* is a separate question,
+answered separately — see "Where a credential may travel" below. Collapsing the
+two would either forbid anonymous HTTP or permit credentialed HTTP.
+
 Two consequences follow, and both are load-bearing
 ([#613](https://github.com/tinyhumansai/opencompany/issues/613)):
 
@@ -121,6 +125,48 @@ is attached. `RequestBuilder::header` appends and axum reads the *first* value,
 so a header from the webview would otherwise be the one the host honoured.
 Keeping the token out of the webview is worth little if the webview still
 decides what a request authenticates as.
+
+### Where a credential may travel
+
+Addressability is not the only question a base url has to answer. A host can be
+perfectly reachable over a wire anyone on the path can read, and the desktop's
+credential is a device session — a person's standing authority on a company,
+attached to every request and to the whole life of the event stream, and
+replayable by whoever copies it down
+([#731](https://github.com/tinyhumansai/opencompany/issues/731)).
+
+So a second rule sits beside the first: **a credential travels over HTTPS, or to
+a host on this machine, and nowhere else.** `may_carry_a_credential` in
+`src-tauri/src/proxy/mod.rs` is the one that enforces it, with
+`mayCarryACredential` in `frontend/src/api/transport/index.ts` as the console's
+copy — the same arrangement as `isAddressableBaseUrl`, and for the same reason:
+a check in the console alone is bypassed by anything reaching the proxy
+directly, and a check in Rust alone arrives as an opaque IPC rejection that
+`client.ts` flattens into "cannot reach the company host".
+
+Loopback is exempt because `http://127.0.0.1:<port>` is how the embedded host is
+reached, on a port that changes every launch and so can never carry a
+certificate; `localhost` and its subdomains come with it, per RFC 6761. The
+private ranges are deliberately **not** exempt — an office LAN is precisely
+where someone else is on the path.
+
+The rule turns on the credential rather than on the scheme, which is what keeps
+a home-lab or staging host without a certificate usable: an anonymous connection
+to one still registers and still reads, because nothing is exposed that a
+passer-by could not have asked the host for themselves. Three surfaces apply it:
+
+- `ProxyRegistry::upsert` refuses to register a credentialed connection to such
+  a host, by name — `this host is not encrypted`, not `not an absolute host url`,
+  because an operator told the second goes to debug a network that is working.
+- `claim` in `commands.rs` refuses the pairing exchange before opening a socket.
+  This is the one place a session token is *created* rather than replayed — the
+  code goes out in the request and the token comes back in the response — and it
+  never touches the registry, so `upsert`'s refusal does not cover it. Its client
+  also refuses redirects: a 307 from an HTTPS base to an HTTP one would re-send
+  the claim, body and all, over the wire the check just refused.
+- `probe` in `registry.ts` marks such a connection `down` with the reason,
+  before contacting it, so the row says what is wrong instead of blaming the
+  network.
 
 The webview also runs under a CSP (`src-tauri/tauri.conf.json`) whose
 `connect-src` allows the IPC origin only. All host traffic goes through Rust and
@@ -193,7 +239,9 @@ a paired session by connection id. Pairing runs entirely in Rust —
 answers with the company, device id and expiry — so the token exists for one
 HTTP response that the webview is not on the path of. That is the difference
 between a design where the webview *should not* hold the credential and one
-where it *cannot*.
+where it *cannot*. The claim itself is a plain `claim()` rather than command
+logic, so the rules on it can be tested without starting a GUI — see "Where a
+credential may travel" for the one it enforces.
 
 The console's `Credential { kind: "device", ref }` is therefore a record that
 this machine is paired, not something the core is told. `ref` is the host's
--- a/frontend/src/api/transport/index.ts
+++ b/frontend/src/api/transport/index.ts
@@ -71,6 +71,71 @@ export function isAddressableBaseUrl(baseUrl: string): boolean {
 }
 
 /**
+ * Whether a secret may be sent to `baseUrl`.
+ *
+ * **HTTPS, or a host that is this machine.** A device session is a person's
+ * standing authority on a company, and plain HTTP puts it in front of every
+ * device on the path — a LAN, a café, an office switch. Reading an unencrypted
+ * host anonymously is not this; the exposure is entirely in the credential, so
+ * that is what this gates and nothing else (issue #731).
+ *
+ * Loopback is exempt because `http://127.0.0.1:<port>` is how the embedded host
+ * is reached and a certificate for an ephemeral port cannot exist. `localhost`
+ * comes with it — RFC 6761 reserves the name, and it is what a developer types.
+ *
+ * ## This is a second copy of a rule that lives in Rust
+ *
+ * Deliberately, and the duplication is the point rather than a smell:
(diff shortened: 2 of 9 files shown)


oxoxDev (approved):
Gating on the **credential** rather than the scheme is the right rule, and the sentence that earns it is "nothing is exposed by reading one that a passer-by could not have asked the host for themselves; the exposure is entirely in the credential". That keeps anonymous LAN and home-lab hosts working while closing the thing that actually leaks — option (2) was the correct read of the issue.

**Refusing to exempt the private ranges is the call most implementations get wrong**, and your reason is exactly right: `10.0.0.0/8` and `192.168.0.0/16` are the networks this exists to protect against, so treating them as trusted would invert the rule while looking like a refinement of it.

The loopback check is the part I went looking for, and it holds up against the ways this normally breaks:

```rust
if let Ok(address) = bare.parse::<std::net::IpAddr>() {
    return address.to_canonical().is_loopback();
}
bare == "localhost" || bare.ends_with(".localhost")
```

Real URL parsing rather than string matching, fail-closed on a parse error, brackets stripped before the IPv6 literal is parsed, and `to_canonical()` so `::ffff:127.0.0.1` is recognised as the loopback it routes to. `127.0.0.2` and `::1` come along correctly for free, which a `== "127.0.0.1"` check would have missed. Names limited to what RFC 6761 actually guarantees rather than to what usually works.

**The three things you found by tracing are each worth more than the check itself**, and the first is the most important: `oc_pair_device` bypasses the registry entirely, so `upsert`'s refusal would never have covered the one exchange where a token is *created* rather than replayed. That its client also carried `reqwest`'s default redirect policy — a 307 from an HTTPS base to an HTTP one re-sending the claim over the wire the check just refused — is the exact shape I raised on Elvin's #739 about standing grants and redirects, found here independently and fixed with `Policy::none()`.

You are also right to push back on my `isAddressableBaseUrl` suggestion from #692. I proposed tightening it; it has two callers, neither is the "Add a host" dialog, so it would not have covered this path — and narrowing it risks the web build's same-origin discovery. A separate credential-aware predicate beside it is the better shape.

**0 major. 1 question, 1 nit. Approving.**

### Question — the console twin is a second copy of the rule you just deduplicated

`may_carry_a_credential` is `pub` precisely so the pairing claim shares it: *"Two copies of one rule is how the console and the core came to disagree about addressability; one function is the fix."* Agreed — and `mayCarryACredential` in TypeScript is, unavoidably, that second copy.

The architecture is right: the core **enforces**, the console **explains** (so `probe` can mark the row `down` with a reason the three layers above would otherwise discard). So a drift between them yields a confusing UI rather than a leak, which is the correct direction. What I would like is for that to be written down at the TypeScript twin — that it is advisory, that the core is authoritative, and that a future change must not "optimise" by trusting it and skipping the core's refusal. Right now the two functions look like peers, and the one place this repo has been bitten repeatedly today is a rule that existed in two places with no test coupling them.

A shared fixture list of URLs asserted identically on both sides would close it cheaply, if that is easy to reach from the console tests.

### Nit — `http://localhost./`

The trailing-dot FQDN form matches neither `== "localhost"` nor `ends_with(".localhost")`, so a credentialed connection to it is refused. That is the safe direction and hardly anyone types it — noting it only so the behaviour is deliberate rather than discovered.

**Before merging:** the branch is **107 commits behind** `main` and GitHub still reports `mergeable: UNKNOWN` — by far the largest drift in either contributor's open set. `main` has been red twice today from semantic collisions between branches that were each green against their own base, so this one wants a rebase and a full green run before it lands, not after.

PR #760, opened 2026-08-12
docs(spec): land #702 review-doc fixes that missed the merge

senamakel:
PR #702 split `storage.md` and `api.md` ([`dc4ae0b`](https://github.com/tinyhumansai/opencompany/commit/dc4ae0b)) but merged with its old head, before the documentation corrections tinysweeper and coderabbit flagged were pushed. This follow-up lands those fixes in `main`, verified against the code:

- **workspace-layout.md** — tinysweeper finding: the merged prose said the journal failed at `{workspace}/tinyagents_store/`, which contradicted the concrete path right below it. Now states that the vendored resolver nests a `workspace/` level under `OPENHUMAN_WORKSPACE`, so `<root>/workspace/tinyagents_store/` = `<data-dir>/openhuman/workspace/tinyagents_store/`. Matches `WORKSPACE_SUBDIR`/test in `src/app/journal.rs`.
- **api-write-plane.md** — tinysweeper finding: the intro named "two deliberate read exceptions" while the same file's route table lists far more. Now enumerates the full set (two inbox `GET`s, three workspace `GET`s incl. `search`, task export, skill-registry browse, agent detail, policy read, credential status). One prose spot and the line counts updated accordingly.
- **memory-engine.md** — coderabbit: `<workspace-name>` is now derived from `EngineCortex::workspace_name` (sanitized id + stable hash, matching `src/store/tinycortex_engine.rs`), the no-network-call claim is qualified to "the local workspace persistence layer", and the embedding dimension is stated (1024 by default).
- **api-graphql.md** — coderabbit: explicit link to the REST write plane, and the read-exception set is corrected to reference api-write-plane.md; GraphQL mutations/subscriptions are stated out of scope with streaming wired over REST.

Docs-only change; each file stays under the 500-line cap. Commands run: `git diff upstream/main...HEAD --stat` (4 files, +64/−35).

### DIFF
--- a/docs/spec/runtime/api-graphql.md
+++ b/docs/spec/runtime/api-graphql.md
@@ -7,12 +7,15 @@ Split out of [`api.md`](api.md), which was over the repository's 500-line ceilin
 ## Read plane — GraphQL (`/graphql`)
 
 Every console **read** is served by a single async-graphql query surface at
-`POST /graphql` (with a `GET /graphql` GraphiQL explorer in development) — the
-sole exceptions being the two inbox `GET`s and the two workspace `GET`s above,
-which exist because the console ships no GraphQL client and those two views need
-a reachable read (issues #173 and #177 respectively). The schema is
-query-only — REST otherwise owns writes — and is **built once at startup** and
-stored on `AppState`; each request injects its resolved `GqlAuth` principal.
+`POST /graphql` (with a `GET /graphql` GraphiQL explorer in development). The
+REST **read exceptions** are the console reads that ship over REST instead —
+the two inbox `GET`s and the three workspace `GET`s, the task export, the
+skill-registry browse, the agent detail `GET`, the policy read, and the
+credential status `GET` — each because the console ships no GraphQL client and
+that view needs a reachable read (issues #173, #177, #607, #352, #264, #562;
+see the [write plane](api-write-plane.md)). The schema is query-only — REST
+otherwise owns writes — and is **built once at startup** and stored on
+`AppState`; each request injects its resolved `GqlAuth` principal.
 
 The schema is rooted at a **`Company` aggregation object** so a view fetches
 everything it needs in one round trip; the only top-level queries are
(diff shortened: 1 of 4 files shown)


oxoxDev (approved):
Verified each correction against the code rather than against the description, since a docs PR is only worth what its claims are:

- **`workspace-layout.md`** — `journal.rs:61` has `WORKSPACE_SUBDIR = "workspace"`, `:64` has `STORE_SUBDIR = "tinyagents_store"`, and `:120` is `self.root.join(WORKSPACE_SUBDIR).join(STORE_SUBDIR)`. So `<root>/workspace/tinyagents_store/` is right and the merged prose was genuinely wrong — it dropped the nested level, which is exactly why it contradicted the concrete path printed below it.
- **`memory-engine.md`** — `EngineCortex::workspace_name` exists at `tinycortex_engine.rs:363`, maps anything outside `[A-Za-z0-9-_]` to `_`, and is pinned by `workspace_name_is_injective_and_stable`. "Sanitized id + stable hash" describes it accurately.
- **The 500-line cap** holds with room: 141 / 312 / 151 / 302.

Qualifying the no-network claim to "the local workspace persistence layer" is the right kind of correction — the original was the sort of absolute that is true of the thing being described and false of the sentence as written.

**0 major. Approving.** Two observations, neither blocking.

### The interesting failure here is not in the docs

> #702 … merged with its old head, before the documentation corrections tinysweeper and coderabbit flagged were pushed.

That is a merge landing on a state that had already moved, and it is the third distinct instance of that shape in this repo today — `main` went red twice from branches that were each green against a base lacking the other, and now a merge that silently dropped pushed review fixes. The first two announced themselves with a red lane. This one did not: the fixes simply were not there, and it took someone noticing to recover them.

Worth understanding how it happened, because the recovery is manual and the detection is luck. If it was a merge clicked from a stale page, that is one class of fix; if it went through an API call carrying an explicit SHA, that is another. Silently losing bot findings is worse than ignoring them, since nobody is left holding the knowledge that they existed.

### The read-exception list will drift again

`api-write-plane.md`'s intro said "two deliberate read exceptions" while the route table in the same file listed many more, and the corrected enumeration now includes `policy read` and `credential status` — which arrived with #659 and #637, both merged today. So it is accurate as of this afternoon and structurally guaranteed to go stale the next time a read lands on that plane.

This repo just solved the identical problem one layer over: #728 generates `gated-tools.generated.ts` from the typed table with a non-ignored test that fails the Rust lane when the checked-in copy drifts, precisely because a hand-maintained list of what-the-code-does cannot be trusted to report its own omissions. A prose count of read exceptions beside a table of routes is the same shape. Not this PR's job — but worth an issue, because the fix that just landed next door is a template rather than an idea.

Issue #555, opened 2026-08-10
tests: the MongoDB store's suite never runs in CI, so the hosted backend is unpinned


oxoxDev:
## Summary

The MongoDB store's test suite never runs in CI. Every test in `src/store/mongodb.rs`'s test module goes through a `store()` helper that reads `OPENCOMPANY_TEST_MONGODB_URI` and returns `None` when it is unset (`src/store/mongodb.rs:1940-1942`), printing `skipping: OPENCOMPANY_TEST_MONGODB_URI is not set`. `.github/workflows/ci.yml` sets no such variable, runs no `services:` container, and contains **no MongoDB reference at all**.

Twenty tests are gated this way — including the shared conformance suite (`src/store/conformance.rs`), the one thing that pins all three backends to answer identically.

MongoDB is the backend hosted tenants run (`StorageKind::Mongodb`, `src/store/select.rs`). So the storage layer under the deployed product is the only one with no CI coverage, and a green pipeline says nothing about it.

## Problem

A skipped test is indistinguishable from a passing one in the CI summary. The suite exists, looks maintained, and asserts nothing on every pull request.

The failure mode this permits is specific: `WorkspaceStore`, `ArtifactStore`, `TaskStore` and the rest are implemented three times over, and `conformance.rs` is what stops the implementations from drifting. With the MongoDB third of that suite dark, a change that satisfies the fs and sqlite backends can ship having broken the hosted one, and the first signal is a tenant.

The precedent for fixing it is already in the file. CI carries a dedicated sqlite lane (`ci.yml:170`, `cargo test --locked --features sqlite --lib store::sqlite`) whose own comment explains it exists because "`cargo test` above omits `--features sqlite`", so those tests had never executed. That reasoning applies unchanged to `--features mongodb`, which additionally needs a server.

Steps to reproduce:

1. Open any pull request touching `src/store/mongodb.rs`.
2. Watch CI go green.
3. `rg -n "mongo" .github/workflows/ci.yml` → no matches.

## Solution

1. **A MongoDB job in `ci.yml`**, modelled on the existing sqlite lane, with a `mongo` service container and `OPENCOMPANY_TEST_MONGODB_URI` pointed at it:

   ```yaml
   services:
     mongodb:
       image: mongo:7
       ports: ["27017:27017"]
       options: >-
         --health-cmd "mongosh --quiet --eval 'db.runCommand({ping:1})'"
         --health-interval 10s --health-timeout 5s --health-retries 10
   ```

[...]

Issue #499, opened 2026-08-06
vendor: the openhuman pin is 102 commits behind, and tinyflows only moves with it


M3gA-Mind:
## Summary

`vendor/openhuman` is pinned at `9a484aa50`, which is **102 commits behind** `tinyhumansai/openhuman` main (`e29bfc66f`). Eight of those commits are the kernelization — the reorg that collapsed `src/openhuman/` from 124 flat domain directories into 31 gate-aligned families.

Filing this because the gap is currently invisible: it is a submodule pointer, so nothing in CI reports it and nothing fails until something needs a symbol the pinned copy does not have. #382 is the first issue to hit it, and it will not be the last.

## Why it matters

The vendored copy is not just openhuman — it is how opencompany reaches **tinyflows**:

```
[patch.crates-io] tinyflows = { path = "vendor/openhuman/vendor/tinyflows" }
```

That path resolves *inside* openhuman, so opencompany cannot move tinyflows independently. Moving tinyflows means moving the openhuman pin, which moves the whole vendored runtime at once.

Concretely, what the current pin costs:

| | tinyflows commit | `on_step_start` declared | fired by engine |
|---|---|---|---|
| what opencompany builds today | `fb24363` (via openhuman `9a484aa`) | no | no |
| openhuman main (`e29bfc66`) | `80546469` | yes | yes |

`on_step_start` landed upstream in tinyflows at `fa7a31b3` (2026-07-22) — the commit immediately after our pin — and fires once per non-trigger node, before the first execution attempt, with the node id. #382 needs exactly that and cannot have it.

## Why this is not a routine bump

Eight kernelization commits sit in the span. The vendored tree's directory layout changes wholesale, so this is a migration rather than a pointer move, and the `Rust (openhuman, tinycortex)` lane is what has to prove it. That is why it belongs in its own PR with its own review, rather than riding in on a UI issue.

A narrower alternative was considered and rejected: landing an openhuman commit that moves only its `vendor/tinyflows` pin, then bumping opencompany to that. It would need branching off a 102-commit-old base, so it is unlikely to be clean.

## What would close this

- `vendor/openhuman` moved to a current openhuman commit, with the `Rust (openhuman, tinycortex)` lane green.
- A note in the PR on what moved and what broke, so the next person can see the shape of the migration.
- Ideally, something that makes the drift visible — the gap grew to 102 commits with nothing reporting it.

## Blocks

[...]

PR #763, opened 2026-08-12
ci: extract the vendored submodule init to one script, and un-break release.yml (#592)

### LINKED ISSUE #592: ci: the vendored-submodule init is copied into three jobs, so a missed copy breaks exactly one lane
## Summary

`.github/workflows/ci.yml` carries the same vendored-submodule init block three times — `rust` (line 47), `Rust (mongodb)` (line 276) and `rust-gated` (line 388). Each copy runs the same two `git -C vendor/openhuman submodule update --init --depth 1 …` commands.

Surfaced by CodeRabbit on #589, which added the third copy. It flagged it as a nitpick and said to treat it as out of scope for a PR focused on the MongoDB lane — correct on both counts, so it is filed here rather than folded in.

## Problem

A submodule added under `vendor/openhuman` must be added in three places, and **a missed copy fails in exactly one lane**. That is the same defect shape #555 was filed for: a lane that quietly is not doing what the other lanes do, discovered late and by accident.

It is worse than ordinary duplication because the copies are not adjacent — they sit ~340 lines apart in a file long enough that a reader edits one and never learns the others exist. The failure is also asymmetric: a lane missing a submodule fails loudly if it needs it to compile, and passes silently if the missing piece only matters to a test that then does not run.

## Solution

Extract the two commands into an executable script under `scripts/ci/` and call it from all three jobs. The repo already uses this pattern — `scripts/ci/assert-integration-targets-run.sh` is the precedent, so this adds no new convention.

Keep the command order and the `--depth 1` flags exactly as they are; this is a de-duplication, not a re-tuning of what gets fetched.

## Acceptance criteria

- [ ] One script under `scripts/ci/`, invoked by `rust`, `Rust (mongodb)` and `rust-gated`; no `git submodule update` remains inline in `ci.yml`.
- [ ] The script is committed executable (`chmod +x`, mode `100755`) — a non-executable script fails every lane at once.
- [ ] CI is green across all three jobs, proving each still gets the submodules it needs.
- [ ] Adding a submodule now requires editing one file, and that is stated in a comment at the call site so the next person does not re-inline it.

## Notes and scope

Purely mechanical, and the failure mode is loud. Worth doing in one pass across all three jobs — extracting it for two of the three would leave precisely the inconsistency the change exists to remove.

Worth checking while in there whether the other workflows (`deploy-staging.yml`, `release.yml`) carry a fourth and fifth copy.

Scope: **small**.

## Related

- #589 — added the third copy and surfaced this
- #555 — the MongoDB lane; same defect family (a lane silently not doing what its siblings do)

oxoxDev:
## Summary

Extracts the vendored-openhuman submodule init into `scripts/ci/init-vendored-submodules.sh` and calls it from every lane, following the `scripts/ci/assert-integration-targets-run.sh` precedent. A grep in the `Actions pinned` meta job keeps it that way.

**Read this part first — the de-duplication is the smaller half of the change.**

The issue was filed against three copies. There are **five**, and the drift it warned about had already happened: **`release.yml` is latently broken on `main` today.** Its copy initializes five submodules, omitting `vendor/tinyhumans-sdk`, and skips the nested `tinycortex` init entirely. `tinyhumans-sdk` became an *unconditional* path dependency of the vendored openhuman crate in #499 (`vendor/openhuman/Cargo.toml:89`), and Cargo reads every path-dependency manifest during resolution — so that job's `cargo build --locked --all-targets` cannot resolve the workspace at all.

Reproduced locally by running `release.yml`'s exact command list against a checkout mirroring its CI state:

```
error: failed to load manifest for workspace member `.`
Caused by: failed to load manifest for dependency `openhuman`
Caused by: failed to load manifest for dependency `tinyhumans-sdk`
Caused by: failed to read vendor/openhuman/vendor/tinyhumans-sdk/Cargo.toml
```

It has never gone red only because the workflow is `workflow_dispatch`-only and has not run in retained history. **The next release dispatch would have failed.** Its step comment still cited the `[patch]` entries as the rationale — the signature of a missed sweep rather than a deliberate divergence.

This is exactly the defect shape #592 and #555 were filed for: a lane quietly not doing what its siblings do, found late and by accident.

### What changed

| Lane | Before | After |
|---|---|---|
| `ci.yml` — `rust`, `Rust (mongodb)`, `rust-gated`, `Desktop` | 4 character-identical inline copies (md5-verified), ~340 lines apart | call the script |
| `release.yml` | drifted copy: 5 submodules, no `tinyhumans-sdk`, no nested init | call the script — **behaviour change**, now fetches the same set as every other lane |
| `deploy-staging.yml` | root-level `--recursive` init over a `submodules: false` checkout, also taking `vendor/tinyagents` | **untouched** — different shape; converting it would change what the Docker build context fetches |

The script's command text is verbatim from the `ci.yml` block: same order, same paths, same `--depth 1`. The rationale scattered across the five step comments is consolidated into its header rather than deleted; lane-specific comments stay at their call sites.

### The guard, and the commit order that proves it

`Actions pinned` gains a step making the invariant enforceable, on that job's own stated argument — a comment is a guard only for people who read it, and its header already records a comment failing to stop the `Desktop` job landing unpinned.

1. Rejects the inline-copy signature `git -C vendor/<path> submodule update` in any workflow.
2. Anti-vacuity per #555: the script must be called by at least one workflow, so a rename or mass delete cannot leave check 1 green while guarding nothing.

**The commits are ordered so the guard is seen to fail on a real runner**, rather than asserted to work:

- `208fd2a6` — script only.
- `ddd29849` — guard only, landed **while all five inline copies still existed**. [**Red run — both checks bite**](https://github.com/oxoxDev/opencompany/actions/runs/31587863922/job/94085773260): annotations `Inlined vendored submodule init` (listing all 9 offending lines) and `Submodule init script unreferenced`.
- `aaecb421` — the conversion. Green.

Please keep both runs in the history when reviewing; the red one is the evidence.

Two precision notes on the pattern, both found by testing it rather than reasoning about it:

- It is deliberately narrower than `git -C vendor/`. That broader form also matches the `Report vendored openhuman drift` step's `rev-parse`/`fetch`/`merge-base`/`rev-list` (`ci.yml:169-172`), which belongs inline — the guard would have been permanently red and unfixable without deleting a good step.
- The anti-vacuity check is anchored to the whole `run:` line so it counts **call sites, not mentions**. Unanchored, it matched the prose inside the guard step itself and could never fail — the precise vacuity it exists to catch. The first draft of this PR had both bugs; the local dry-run caught them.

## API Or Behavior Changes

- **`release.yml` now fetches `vendor/tinyhumans-sdk` and the nested `tinycortex` → `tinyagents` submodule**, which it did not before. This un-breaks the release build. The nested init is harmless for a default build — seconds of shallow fetch.
- CI lane behaviour is otherwise unchanged: the four `ci.yml` copies were character-identical to the script's contents.
- The script deliberately does **not** initialize top-level submodules, so no caller's checkout strategy changes.
- New failure mode, on purpose: a workflow that inlines the init now fails `Actions pinned`.

## Tests

The script cannot silently no-op and stay green: without the fetch, Cargo fails manifest resolution before compiling anything. So green `rust` / `Rust (mongodb)` / `rust-gated` / `Desktop` on this head *is* proof it ran and fetched the right set. `rust-gated` additionally compiles all six crates plus nested `tinyagents` under `--features openhuman`.

Locally, in a scratch clone whose state mirrors CI's (`actions/checkout` with `submodules: true`, i.e. top level only, all nested submodules uninitialized):

- `sh -n` and `shellcheck` — clean.
- Ran it from a **subdirectory** (`src/server`) — resolves the repo root correctly, exit 0.
- `git -C vendor/openhuman submodule status` afterwards — all six at pinned commits, **SHA-for-SHA identical** to a known-good checkout. `tauri-cef` correctly left uninitialized.
- `cargo metadata --locked` — OK. A correct SHA in `submodule status` does not prove the objects landed; this does.
- Ran a **second time** — exit 0, idempotent.
- Precondition path: with `vendor/openhuman` unchecked-out, fails with the actual mistake named. Worth noting this is not hypothetical — during testing, `git -C vendor/openhuman submodule status` against an empty directory silently reported the *parent* repo's submodules, which is the confusing failure the check replaces.
- Guard dry-run: red against the unconverted tree (9 matches, both checks), green against the converted one (5 call sites).
- `actionlint` — clean on both workflow files.

`release.yml` cannot be exercised on this PR — it is dispatch-only and it publishes — so a YAML typo there is the residual risk. `actionlint` is the mitigation.

Not run: `cargo fmt` / `clippy` / `build` / `test` locally — this PR changes no Rust source. CI runs them.

- [ ] `cargo fmt --all -- --check` — N/A: no Rust source changed; CI runs it.
- [ ] `cargo clippy --all-targets -- -D warnings` — N/A: no Rust source changed; CI runs it.
- [ ] `cargo build --all-targets` — N/A: no Rust source changed; CI's four lanes are the real assertion here.
- [ ] `cargo test` — N/A: no Rust source changed; CI runs it.

## Documentation

No docs change. The reasoning lives where the next person will hit it: the script header states plainly that adding a submodule under `vendor/openhuman` means editing that file and nothing else, and each call site points at it.

### Deliberately not done

- An objects-landed post-check inside the script (`submodule foreach 'test -f Cargo.toml'`) — one nested submodule is a JS repo with no `Cargo.toml`, and Cargo already fails loudly downstream on every converted lane.
- Converting `deploy-staging.yml`. Worth a follow-up look, but it is a genuinely different init and folding it in here would change the image build.

Closes #592

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,27 +165,18 @@ jobs:
           persist-credentials: false
           submodules: true
 
-      # `openhuman_core` is an optional path dependency, so Cargo must resolve
-      # its vendored-crate manifests even for the default (offline) build where
-      # the `openhuman` feature is off. `submodules: true` is not recursive, so
-      # init the nested crates its manifest names.
+      # `submodules: true` is not recursive, so the crates nested under
+      # `vendor/openhuman` still have to be fetched. Required even for this
+      # default (offline) build, where the `openhuman` feature is off:
+      # `openhuman_core` is an optional path dependency, but Cargo reads every
+      # path dependency's manifest during resolution regardless of features.
       #
-      # `tinyhumans-sdk` is NOT one of the `[patch]` targets — openhuman consumes
-      # it as a plain unconditional path dependency because it is unpublished
-      # (issue #499). It still has to be on disk: Cargo reads every path
-      # dependency's manifest during resolution, so a missing checkout fails the
-      # DEFAULT build too, not just `--features openhuman`.
+      # The list of crates lives in the script and only there — add a submodule
+      # under `vendor/openhuman` there, not here (issue #592). The script also
+      # carries the rest of the reasoning, including why `tinyhumans-sdk` is in
+      # the list despite having no `[patch]` entry (issue #499).
       - name: Init vendored openhuman crate submodules
-        run: |
-          git -C vendor/openhuman submodule update --init --depth 1 \
-            vendor/tinyagents vendor/tinybus vendor/tinyflows vendor/tinycortex \
-            vendor/tinydocs vendor/tinymemory vendor/tinywallet \
-            vendor/tinyjuice vendor/tinychannels vendor/tinyplace vendor/tinyhumans-sdk
-          # tinycortex declares its own `tinyagents` path dependency, so its
-          # manifest must resolve too. Harmless for the default build; required
-          # for any `--features openhuman` build.
-          git -C vendor/openhuman/vendor/tinycortex submodule update --init --depth 1 \
-            vendor/tinyagents
+        run: scripts/ci/init-vendored-submodules.sh
 
       # Issue #499: say out loud how far the vendored pin has drifted.
       #
@@ -406,14 +397,10 @@ jobs:
       # its manifest during resolution even where the feature selection excludes
       # `openhuman`. A missing checkout fails THIS build too, not just a gated
       # one. `submodules: true` is not recursive, hence the explicit init.
+      #
+      # The list lives in the script and only there (issue #592).
       - name: Init vendored openhuman crate submodules
-        run: |
-          git -C vendor/openhuman submodule update --init --depth 1 \
-            vendor/tinyagents vendor/tinybus vendor/tinyflows vendor/tinycortex \
-            vendor/tinydocs vendor/tinymemory vendor/tinywallet \
-            vendor/tinyjuice vendor/tinychannels vendor/tinyplace vendor/tinyhumans-sdk
-          git -C vendor/openhuman/vendor/tinycortex submodule update --init --depth 1 \
-            vendor/tinyagents
+        run: scripts/ci/init-vendored-submodules.sh
 
       - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
         with:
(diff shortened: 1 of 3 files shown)


oxoxDev:
Rebased onto `ea40b004` after the conflict in `ci.yml`.

Worth flagging what the conflict actually was, because a blind `--ours` would have shipped a regression: **`main`'s submodule list grew from 6 entries to 11 while this PR was open** — `tinyagents`, `tinybus`, `tinydocs`, `tinymemory` and `tinywallet` were added to all four inline copies. Taking this branch's side verbatim would have silently dropped those five from every lane.

So the resolution updates the script to the current 11-entry list first, then converts the call sites. The script is now a superset-of-nothing: it names exactly what `main` names today, plus the nested `tinycortex` → `tinyagents` init.

That is also the issue restating itself while the fix was in review — five copies drifting together this time, which is the benign direction, but only because someone remembered all four. The sixth copy in `release.yml` was not updated in that same change, which is the whole point of #592.

Verified after the rebase: no inline `git -C vendor/… submodule update` remains in any workflow, all five call sites reference the script, `sh -n` clean, and `actionlint` clean on both workflow files. Also corrected a now-stale comment in the gated lane that said "all six crates".

Issue #392, opened 2026-08-05
runtime: decide whether a journal append should survive losing the host, not just the process


oxoxDev:
## The question

The runtime journal's append now completes its write syscall before returning, so a **process** crash cannot lose a record that was reported durable (#386). It does not call `sync_data` or `sync_all`, so the bytes sit in the kernel page cache rather than on stable storage: a **host** crash or power loss between the append and the kernel's flush can still lose a record the caller was told was safe.

That gap matters specifically for the at-most-once contract. The journal writes an `EffectExecuted` key *before* the side effect runs, precisely so a restart cannot run it twice. If the machine dies after the effect happened but before the record reached disk, the restart replays without that key and the effect can fire a second time — a duplicate payment, a duplicate external message.

## Why it is not simply "add fsync"

An `fsync` per append is a real cost on a path that runs on every parked approval, every resolution, every effect and every run outcome. On a hosted tenant that is a lot of forced flushes, and the durability it buys is against a failure mode — losing the machine — that the rest of the system does not currently defend against either. The boot reaper and the interrupted-run sweep are both built around a process dying, not a host disappearing.

So this is a deliberate decision to make, not an oversight to patch:

- Is losing the machine in the write window a risk this product accepts?
- If not, is the answer `sync_data` on every append, or only on the records where duplication is expensive (`EffectExecuted` and friends) rather than all of them?
- Does the hosting platform already give us something here — a durable volume, replication — that changes the calculation?

## What would settle it

- Measure the cost: append throughput with and without `sync_data`, at a realistic record size, on the storage tenants actually run on.
- Enumerate which record kinds carry duplication risk if lost. If it is a small set, a per-kind policy is cheaper than a blanket flush and easier to justify.
- Decide and write it down. The append path currently documents the guarantee as process-crash durability; whichever way this goes, that doc should end up stating a decision rather than a limitation.

## Related

Follows #386, which fixed the larger problem: appends were returning `Ok(())` with the bytes still queued, so even a process crash could lose them, and an `ENOSPC` reached nobody. This issue is the remaining, much narrower window.
Raised in review of PR #389.

oxoxDev:
**Related — `cluster:journal`**

The company event journal: what it guarantees, what it keeps, and what it can be made to forget.

[...]

PR #766, opened 2026-08-12
feat(journal): back the runtime journal with a storage port (#726)

### LINKED ISSUE #726: runtime: the journal bypasses the storage ports, so a hosted mongodb tenant's at-most-once set lives on ephemeral /data
## The gap

`RuntimeJournal` is constructed **unconditionally on the filesystem path** — `src/runtime/builder.rs:1123-1125` calls `Bundle::journal_jsonl()` regardless of the configured storage backend. It sits outside the fourteen storage ports that every other piece of persistence goes through.

For a tenant running `OPENCOMPANY_STORAGE=mongodb`, that is a problem the fs-durability discussion in #392 does not reach. `docs/spec/runtime/storage.md` documents `/data` for these tenants as **ephemeral scratch** — which is exactly why the tinycortex boot refusal exists. So `journal.jsonl` does not survive a **container replacement**, never mind a host crash.

## Why this matters more than the fsync question

The journal is where the at-most-once contract lives. It writes an `EffectExecuted` key before a side effect runs precisely so a restart cannot run it twice. On a hosted mongodb tenant that key is on scratch storage: replace the container — a deploy, a reschedule, a node drain, an OOM kill — and the whole set is gone. Every previously-executed effect is eligible to fire again.

#392 addresses a window of **seconds** (kernel page cache between append and writeback) on the fs-on-durable-volume shape. This is a window of **the container's lifetime**, on a shape that is a routine operational event rather than a hardware failure.

The two are not alternatives. #392's per-kind durability declaration travels with the record kind and survives any backing-store move, so it is not wasted work — but no amount of `sync_data` helps a volume that is discarded on purpose.

## Shapes, for clarity

| Deployment | `/data` | Journal survives host crash? | Survives container replacement? |
|---|---|---|---|
| Reference ECS (`deploy/aws-ecs-task-definition.json`) | EFS mount | Only with the #392 fix (NFS client page cache) | Yes |
| Hosted multi-tenant, `OPENCOMPANY_STORAGE=mongodb` | ephemeral scratch | No | **No** |
| Local dev | local disk | Only with #392 | n/a |

## What a fix probably looks like

Put the journal behind a port like every other persistence surface, so a mongodb tenant's journal lands in its tenant database alongside the rest of its durable state. That is a larger change than #392 and deliberately not folded into it — it touches the port surface, the builder, and the replay path.

Worth settling first, because it changes the shape:

- Does the journal become a fourteenth-plus port with per-backend implementations, or does the fs journal stay and the *record set that carries at-most-once* move to the backend?
- Replay currently reads a JSONL file start-to-finish. What does replay mean against a document store — a query, or a full scan?
- Does an append still need to be synchronous with respect to the side effect, and what is the mongodb equivalent of the `sync_data` decision #392 makes?

## Related

- #392 — per-kind host-crash durability for the fs journal. Surfaced this gap; does not fix it.
- `docs/spec/runtime/storage.md` — the ephemeral-`/data` documentation and the tinycortex boot refusal that exists for the same reason.

oxoxDev:
## Summary

`RuntimeJournal` was constructed unconditionally on the filesystem, outside the port surface `StorageHandles` swaps for every other durable store. The journal is where the **at-most-once** contract lives — the `EffectExecuted` key is committed *before* a side effect runs. On a hosted mongodb tenant `/data` is documented ephemeral scratch, so container replacement (deploy, reschedule, node drain, OOM kill) discarded the whole set: every previously executed effect became eligible to fire again, and every parked approval, single-use grant and standing grant silently vanished.

This adds a `JournalStore` port with fs / sqlite / mongodb backings, a receipt-gated one-time import of an existing `journal.jsonl`, and conformance coverage across all three.

The port is deliberately two byte-level operations (append one opaque line, read every line in append order) rather than a mirror of `RuntimeJournal`'s ~25 methods. Everything semantic — the record enum, the corrupt-line skip, the merged-line recovery, the in-memory replay — stays in `RuntimeJournal` and is backend-agnostic by construction, so a new record variant needs no backend change. It does not ride `EventLog`: `CompanyEvent` is a closed enum, and the event log is *pruned* under a retention policy — rotating away an `EffectExecuted` key would silently un-commit it.

### `Durability` had to become part of the port — the most important thing to review

#730 (#392, journal host durability) merged mid-implementation and rewrote both files this PR rewires. It had made `RuntimeJournal::append` branch on a per-record `Durability` **twice** — `create_dir_all_durable` vs `create_dir_all`, then `append_line_durable` vs `append_line`. A port with a plain `(id, line)` signature would have flattened that to `Process` and **silently reverted #392**.

So `Durability` now lives in `src/ports/journal.rs` and `append_journal` carries it; the per-record *decision* stayed on `JournalRecord::durability`. Both halves of the fs branch — including `create_dir_all_durable` — moved into `FsJournalStore::append_journal` intact, and #730's two tests still fail if it is flattened.

**Reviewers: `FsJournalStore::append_journal` in `src/store/fs.rs` is the place to look hardest.**

### Two durability decisions the plan did not anticipate

Taken because the alternative was a silent weakening:

- **sqlite** honours `Host` by running the insert under `PRAGMA synchronous=FULL`, and `Process` under the standing `NORMAL` (sound only under WAL). The import transaction is `Host`-durable for the same reason `EffectExecuted` is.
- **mongodb** applies `j:true` **only to `Host` records**, not to every journal write. Blanket `j:true` would tax `CycleStarted`/`CycleFinished` — a pair per cycle — with a disk flush each, which is exactly the trade #392 declined.

### The import gate closes even when there is no `journal.jsonl`

An import of zero lines still writes the receipt. Without that, a file appearing *later* — a rollback, a stray copy into the data dir — would wipe and replace a journal the backend had since accumulated, because `complete_import` clears before it copies. The source file is left in place: the receipt makes a second import impossible, and a rollback to an older binary still finds history it can read, where renaming the file would hand it an *empty* at-most-once set.

## API Or Behavior Changes

- **New port**: `crate::ports::journal::{JournalStore, Durability}` — `append_journal`, `read_journal`, `journal_imported`, `complete_import`.
- **Breaking for in-flight PRs**: `StorageHandles` gains a **non-optional** `journal: Arc<dyn JournalStore>` field, so any open PR constructing it will fail to compile until it adds one. Deliberate — an `Option` would invite the silent fs fallback this issue exists to remove.
- **Behavior**: with `OPENCOMPANY_STORAGE=sqlite|mongodb` the journal now lives in the backend rather than in the company bundle. First boot performs a one-time verbatim import of any existing `journal.jsonl`, then records a receipt. The fs backend is unchanged — its store *is* the file at today's path, so it migrates nothing and answers `journal_imported` `true` unconditionally.
- **Errors stay fail-closed**: an `Err` from `append_journal` reaches the caller *before* the side effect, so the effect does not run. The residual ambiguity (a timeout on a write the server did commit) leaves a committed key with no effect — the at-most-once contract's documented safe direction.

### One honest limitation

An invalid-UTF-8 line cannot round-trip byte-for-byte through a sqlite `TEXT` column or a BSON string. Such a line migrates as a corrupt line that still fails to parse and is still reported — the same outcome as today — but "verbatim" is approximate in that one case.

### A residual coverage gap, named rather than left implicit

The durability *level implementations* are pinned on all three backends (`the_host_durable_write_runs_under_full_sync_and_restores_normal`, `only_the_host_durable_journal_write_asks_for_j_true`, and #730's fs `append_probe` counts). But the per-call **branch selection** inside sqlite's and mongodb's `append_journal` is not directly observable — there is no database equivalent of the fs `append_probe`. A mutation swapping those two arms would go undetected.

## Tests

Run locally on this branch, rebased on `main` @ `cae21623`:

- [x] `cargo build --all-targets` (covered by the `--all-targets` clippy lanes below)
- [x] `cargo test`

| Lane | Result |
|---|---|
| `cargo clippy --no-deps --features openhuman,mcp,telegram --all-targets -- -D warnings` | clean |
| `cargo test --locked` | **2246 passed**, 0 failed, 1 ignored (lib) |
| `cargo test --locked --features sqlite --lib store::sqlite` | **34 passed**, 0 failed, 2247 filtered out |
| `cargo test --locked --features openhuman,mcp,telegram` | **3372 passed**, 0 failed, 3 ignored (lib) |
| `cargo test --locked --features mongodb --lib store::mongodb` | **31 passed**, 0 failed, 2248 filtered out |

The mongodb lane ran against a local `mongo:7` with `OPENCOMPANY_TEST_MONGODB_URI=mongodb://localhost:27017` and `OPENCOMPANY_TEST_MONGODB_REQUIRED=1`. `REQUIRED=1` matters: without it a missing server skips silently and still exits 0. `conformance_journal_import` is in the 31 that ran, so the server-backed path is genuinely covered rather than vacuously green.

### Teeth

Each was verified by reverting the production change, observing RED, and restoring.

The headline one: with the builder ignoring the backend sink, `a_backend_journal_survives_the_loss_of_the_whole_data_directory` fails with *"the committed key must survive the container: without it the effect is eligible to fire a second time"*. It boots, executes an effect and parks an approval, deletes the entire data directory (asserting the file is really gone first), then boots again on the same sink.

Also covered: `a_filesystem_journal_is_imported_once_and_the_receipt_blocks_the_rest` (a second import would clear a key committed after the migration and re-arm an effect), and `an_import_interrupted_before_its_receipt_is_re_run_whole` (a crash between the copy and the receipt re-runs the whole wipe-and-copy rather than replaying a truncated prefix).

One test's teeth are structural rather than demonstrated: `a_journal_over_a_non_filesystem_store_replays_identically` cannot be written against pre-change code at all, since the seam it uses does not exist there. Its non-vacuity comes from a paired negative assertion in the same test.

`assert_journal_store` holds fs, sqlite and mongodb to identical read-back semantics — including that the fs backend's trailing-newline artefact is dropped so its read agrees element-for-element with a database backend's.

### Manual verification

Real host on `OPENCOMPANY_STORAGE=mongodb` against `mongo:7`, console built and served. Boot logged `sink="backend"` with no `journal.jsonl` anywhere under the data dir. A chat turn parked an approval; `kill -9` plus `rm -rf` of the entire data dir (simulating container replacement); re-serve.

In the browser: the Approvals badge showed 1 and the page named the approval parked before the wipe, under its original id and timestamp. Approving it from the console produced `ApprovalResolved` then `EffectExecuted`. A second wipe-and-restart confirmed the resolution survived and the effect had fired **exactly once**.

## Documentation

- `docs/spec/runtime/journal.md` — new: the port, the two durability levels and what each backend does for them, and the migration's clear-copy-receipt ordering with the safety argument.
- `docs/spec/runtime/ports.md` — `JournalStore` added to the port surface.
- `docs/spec/runtime/storage.md` — the journal is no longer an exception to backend selection.
- `docs/spec/runtime/data-root.md` — `journal.jsonl` is the fs backend's file, not an unconditional one.
- `docs/spec/runtime/lifecycle.md` — boot-time import step.
- `docs/spec/runtime/README.md` — links the new page.

Rustdoc on `src/ports/journal.rs` carries the per-backend durability table and the "why not `EventLog`" reasoning.

Closes #726

### DIFF
--- a/docs/spec/runtime/README.md
+++ b/docs/spec/runtime/README.md
@@ -20,6 +20,10 @@ Supporting docs:
   - [ports-console.md](ports-console.md) — the WS3 console-surface stores
   - [ports-runs.md](ports-runs.md) — `RunStore`: one attempt at a task, its
     trace, and who writes it
+  - [journal.md](journal.md) — `JournalStore`: the runtime journal's durable
+    sink (at-most-once effect keys, parked approvals, grants, cycle brackets),
+    the per-backend shapes, and the one-time receipt-gated import off the old
+    `journal.jsonl` (issue #726)
 - [storage.md](storage.md) — how a backend is chosen at boot, the shipped
   backends, and the conformance suite that pins them to identical answers
   - [workspace-layout.md](workspace-layout.md) — the on-disk layout inside the
--- a/docs/spec/runtime/data-root.md
+++ b/docs/spec/runtime/data-root.md
@@ -47,6 +47,15 @@ on `<root>/.lock` (`flock`/`LockFileEx` via `fs2`) and hold it for the life of
 the process. A second instance is refused immediately with a message naming the
 directory and `OPENCOMPANY_DATA_DIR`.
 
+Since issue #726 the journal's *bytes* may live in the storage backend rather
+than under this root (see [journal.md](journal.md)), and that changes nothing
+here: single-writer-per-company is still the contract. Two live hosts on one
+tenant database no longer corrupt each other's records — sequences are allocated
+server-side, so appends interleave without collision — but each keeps its own
+in-memory replay of the executed-key set, so neither sees the other's commits
+until it reloads. That gap is pre-existing and unchanged; the root lock above is
+still what a single-node deployment relies on.
+
 An OS advisory lock rather than a pid file: the lock belongs to the open file
 description, so the kernel drops it when the process exits for any reason —
 clean exit, panic, `SIGKILL`, power loss. There is no stale state to detect and
--- a/docs/spec/runtime/journal.md
+++ b/docs/spec/runtime/journal.md
@@ -1,0 +1,167 @@
+# The runtime journal and its storage port
+
+The runtime journal (`src/runtime/journal.rs`) holds four things no other store
+holds:
+
+- the **at-most-once effect set** — the idempotency key of every side effect that
+  was committed to run;
+- the **parked-approval queue**, with each approval's original `ApprovalId`, the
+  effect it parked, its board task and its conversation;
+- **single-use and standing grants**, minted when an operator approves a tool
+  call;
+- **cycle brackets**, which is how a cycle interrupted by a host restart is
+  recognised and settled at the next boot.
+
+It is deliberately not the [`EventLog`](events.md). `CompanyEvent` is a closed,
+binding enum with no marker variants, and the event log is *pruned* under a
+retention policy — rotating away an `EffectExecuted` key would silently
+un-commit it and let an at-most-once effect fire a second time.
+
+## Why it is a port (issue #726)
+
+The journal used to be constructed unconditionally on the filesystem, at
+`<home>/companies/<slug>/journal.jsonl`, outside the port surface
+`StorageHandles` swaps for every other durable store.
+
+On a hosted tenant with `OPENCOMPANY_STORAGE=mongodb` the container's `/data` is
+documented **ephemeral scratch** — the same fact the TinyCortex boot refusal
+exists for (see
+[memory-engine.md](memory-engine.md#durability-contract--the-data-is-scratch-caveat)).
+Container replacement — a deploy, a reschedule, a node drain, an OOM kill —
+discarded the file. Every previously executed effect became eligible to fire
+again, and every parked approval, grant and standing grant silently vanished.
+
+That is a data-integrity defect, not log noise, so the fix is the port
+(`src/ports/journal.rs`) rather than a warning.
+
+## `JournalStore` — two byte-level operations
+
+```rust
+async fn append_journal(&self, id: &CompanyId, line: &str) -> Result<()>;
+async fn read_journal(&self, id: &CompanyId) -> Result<Vec<String>>;
+async fn journal_imported(&self, id: &CompanyId) -> Result<bool>;
+async fn complete_import(&self, id: &CompanyId, lines: Vec<String>) -> Result<()>;
+```
+
+Not a mirror of `RuntimeJournal`'s ~25 methods. The journal's whole persistence
+contract is "append one opaque line, read every line back in order". Everything
+semantic — the record enum with its `#[serde(default)]` archaeology, the
+corrupt-line skip, the merged-line recovery, the replayed in-memory state — stays
+in `RuntimeJournal` and is backend-agnostic by construction. A backend stores
+strings and never learns what a `JournalRecord` is, so a new record variant needs
+no backend change.
+
+Two obligations a backend must not weaken:
+
+- **Durability before return.** The at-most-once guarantee is that a key is
+  durable *before* the side effect runs. An `Err` therefore reaches the caller
+  before the effect, which fails closed. The residual case — a timeout on a write
+  the server did commit — leaves a committed key with no effect, the contract's
+  documented safe direction.
+- **Bytes and order.** A rewritten, trimmed or re-encoded line no longer parses,
+  and a skipped `EffectExecuted` un-commits its key. A park read back *after* the
+  resolution that drained it resurrects a resolved approval.
+
+`src/store/conformance.rs` holds every backend to both:
+`assert_journal_store` (all three backends) and `assert_journal_import`
+(sqlite and mongodb — the fs backend has nothing to import). Both levels of
+`Durability` are exercised there, because a backend that ignored the parameter
+would otherwise pass every assertion.
+
+## Durability travels through the port (issue #392)
+
+Durability is chosen **per record kind**, not once for the journal:
+`EffectExecuted`, `GrantConsumed` and `StandingGrantRevoked` must survive losing
+the machine, because losing them makes the runtime repeat an external action; the
+other ten need only survive the process, because losing them makes it *re-ask*.
+The per-record decision and its reasoning live in
+[lifecycle.md](lifecycle.md#which-crash-a-journal-record-survives-issue-392).
+
+`append_journal` therefore carries a `Durability`, and a backend MUST NOT flatten
+the two levels into one. Flattening upward taxes the journal's highest-volume
+records with a flush they do not need; flattening downward silently drops the
+guarantee that keeps an already-fired effect from firing again after a power
+loss, which is the one thing the split was bought for.
+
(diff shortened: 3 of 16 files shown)


M3gA-Mind:
Reviewed for durability — whether moving the journal behind a port can lose a write that the filesystem path did not. **Not an approval.**

## What holds up

**The `Durability` split survives the move intact, and that is the thing that mattered.** The three `Host` kinds (`EffectExecuted`, `GrantConsumed`, `StandingGrantRevoked`) are exactly the records whose loss would *repeat an external action*; the other ten make the runtime re-ask and never re-fire. Keeping that as an explicit parameter on the port — rather than letting each backend pick — is what stops the refactor from quietly downgrading the at-most-once contract.

**All three backends actually honour it**, which is the part I most expected to find broken:

| | Process | Host |
|---|---|---|
| fs | `O_APPEND` `write_all` | + `sync_data` + parent chain flushed |
| sqlite | `synchronous=NORMAL` | `synchronous=FULL` |
| mongodb | acknowledged insert | `WriteConcern` with `journal(true)` |

The mongodb one is the new risk surface and it is right: `j:true` means the server has committed to its own on-disk journal before the insert returns. Scoping that write concern to the journal collections rather than database-wide is the correct granularity — a database-wide `j:true` would tax every unrelated write.

The fs parent-directory flush is the detail most implementations miss: a record `fsync`ed under a directory entry that was never itself written down is lost with it. Good that #392's reasoning came across with the code.

**The conformance suite alternates both levels** and you say plainly that it *cannot* catch a backend ignoring the parameter, then cover that with per-backend tests. That is the honest structure — a conformance suite that implied it checked durability would be worse than one that says it does not.

`$max` rather than `$set` on the counter after import, so a counter somehow ahead is never wound back, is a nice touch.

## One narrow window, and I want to know if it is deliberate

`complete_import` is `delete_many` → `insert_many` → counter → receipt, with **no transaction**, and the receipt is written last. The receipt-last ordering is right — it is what makes an interrupted import re-run the whole copy rather than leave a partial one.

The window I can construct: crash **after** `insert_many` but **before** the receipt lands. The backend now holds a correct journal with no receipt, so the next boot re-imports — and `complete_import` clears before it copies. That is safe while the filesystem journal is still the source of truth, which it normally is.

It stops being safe if the fs journal is gone by that next boot, because the caller would then hand `complete_import` an **empty** `lines` and `delete_many` would erase a good journal. Your zero-line gate is deliberately "import even when there is no file", so an empty import is a real path, not a hypothetical.

I could not make that reachable in a way I would call a defect: a host with a journal worth importing normally has persistent `/data`, and a tenant whose `/data` is ephemeral scratch has an empty fs journal on every boot anyway, so its import copies nothing and the receipt lands immediately. The combination needs `/data` persistent enough to hold a journal and ephemeral enough to lose it inside that one window.

So this is a question rather than a finding: **is that window accepted deliberately, and is it worth a note next to `complete_import`?** If mongodb here is always a replica set, wrapping the clear-and-copy in a transaction would close it outright; if the window is accepted, saying so where the clear happens would stop the next reader re-deriving this exact analysis.

## Not covered

I reviewed the port contract, the three backends' durability handling, and the import/receipt path. I did **not** review the 201/144 rewrite of `runtime/journal.rs` line by line, the 167-line journal doc, or the conformance suite's non-durability assertions, and I did not build or run anything.

PR #771, opened 2026-08-12
fix(delegation): scope the delegation queue per claimant (#661)

oxoxDev:
## Summary

`DelegationQueue` held a **global** claim and `claim_as()` cleared on entry, so any claimant taking the queue destroyed every other claimant's staged delegations, refusals and scope chain. That is safe only for as long as every claimant serialises under the chat cycle lock.

**Workflow runs are `tokio::spawn`ed and are not under that lock** — several overlap, bounded only by the #401 in-flight cap. This is byte-for-byte the race `ApprovalRequestQueue` had before #439, one queue over, so it is fixed the same way rather than a new way: entries bucket by a `DelegationScope` key, a claimant takes a claim, and calls file into the surrounding claim's bucket.

**One half of this is already live, not latent.** `DelegateToDeskTool` calls `push_refusal` on the ungrounded path *before* it consults the claim (`orchestrator.rs`, `execute`), so an invented desk named by a workflow agent node already reaches the shared vector today — with no drain wired and nothing else changed. A concurrent chat turn's `drain_refusals` then takes it, records it on that turn's card, and clears it: a hand-off attempt nobody on that turn made, attributed to that turn, and destroyed for the run that made it. A workflow node's turn carries the whole toolbelt, so an orchestrator-tier `agent_ref` genuinely can reach `delegate_to_desk` (documented at `workflows/caps/mod.rs`).

The other half is latent and would trip the moment anything on the workflow path claims: the entry `clear()` would wipe a concurrent chat turn's staged card after that turn had already told the operator it was opened.

`Drop` now discards only its own bucket, so a cancelled run's staged writes die with the run while a concurrent chat turn's survive — matching the stance `ApprovalClaim` takes.

Also adds `DrainClaim::Board`, a claim variant permitting `spawn_task` and `assign_task` and refusing `review_task` and `delegate_to_desk` in the model's own turn. Nothing claims it yet; the drain is a follow-up. It is here because it is the piece that makes the scoping testable end to end, and because its refusals encode a decision worth reviewing on its own (below).

Context, not the headline: this surfaced while designing M5 of #661 (workflows creating task cards). The concurrency fix is correct regardless of how that lands.

### The chat path has zero behaviour change

`DelegationScope::Unscoped` is the default, and **every existing chat and task call site compiles unchanged as one** — no call site in this diff. `run_cycle`, the operator turn, its relay and a dispatched card all serialise under the cycle lock, so one bucket is all they ever needed. An unclaimed push landing in `Unscoped` is deliberately not an error: a claimant added later that forgets to name a scope degrades to today's behaviour rather than to a silently dropped delegation, the same reasoning #439 applied.

**The existing orchestrator test battery passes completely unmodified** — 102 pre-existing orchestrator tests, zero edited (the test commit is +323 insertions, 0 deletions).

### Two scopes with confusingly similar names, kept distinct

#176 put a scope *chain* on this queue with an RAII guard where depth **is** `chain.len()`. That is a different thing from these buckets: the key is *which claimant*, the value is *how deep that claimant is nested*. The field is renamed `scope` → `chains` for that reason.

**Depth accounting is unchanged.** Depth is still exactly `chain.len()`, still has no counter beside it, and is still read and written only within one claimant's own chain — so a concurrent run can neither deepen nor shallow another's. `ScopeGuard` now records which scope it pushed onto and pops from that one rather than whatever is ambient at drop time, so a pop cannot take a live level off another claimant's chain. Pinned by a test that proves the bound still fires at the same depths.

## API Or Behavior Changes

Public, additive:

- `DelegationScope { Unscoped, Run(String) }`, and a `tokio::task_local` current scope.
- `DelegationQueue::claim_board(run_id)` and `DelegationClaim::scoped(fut)` / `::scope()`.
- `DrainClaim::Board`; `NoDrainReason::{WorkflowLifecycle, WorkflowHandOff}`; `Delegation::writes_board_only()`.

No existing signature changed. `push`, `drain`, `clear`, `queued`, `push_refusal`, `refusals_queued`, `drain_refusals`, `drain_refusals_after`, `claim_state`, `scope_depth`, `scope_chain` and `enter_scope` all keep their signatures; only *which* entries they see changes, and for every existing caller that is the same single bucket as before.

**Why `Board` refuses what it refuses** — the part worth arguing with:

- `review_task` — `in_review → done` is the operator accept lane, and column movement is operator-only (`todo → planning` is only ever an operator drag, `planning → in_progress` is the dispatch gate). The no-column-moves rule is **load-bearing for loop safety, not taste**: run → card → dispatch → run cycles are bounded precisely because every dispatch requires an operator act. Relaxing it takes the bound with it.
- `delegate_to_desk` — its only value is a synchronous reply, and a run has no conversation for one to land in.

Two `NoDrainReason`s rather than one, with their own sentences and their own log labels, because the causes are unrelated: a model told the wrong one is told something false about what it may do next. Each sentence names what the run **can** do instead, since that is what a model reads and reacts to.

## Tests

Seven new tests, each with teeth proved by reverting the production change it guards and observing red, then restoring:

| Guard reverted | Result |
|---|---|
| `clear_scope` → pre-change global clear | 3 red, incl. the headline: chat's staged card `left: 0, right: 1` |
| `push_refusal` → global bucket | red: chat turn sees a refusal it never made (`left: 1, right: 0`) |
| chain → single global chain | red: run opens at depth `left: 1, right: 0` |
| `Board` permit gate disabled | red: `review_task` `left: Queued, right: NoDrain(WorkflowLifecycle)` |
| hand-off wording → copy of `Unwired` | red on the borrowed-cause assertion |

Covered: two `Run` scopes and `Unscoped` interleaved with no cross-drain or cross-clear; `Drop` discarding only its own bucket and un-claiming only its own scope; scope-chain bucketing with depth accounting proven unchanged; the `Board` permit matrix with both permitted kinds accepted and both refusal wordings asserted — including that neither borrows the other's cause, and that all five `NoDrainReason` log labels stay distinct.

Commands run locally, on this branch, rebased onto `main` @ `f1a7a278`:

- [x] `cargo fmt --all -- --check` — pass
- [x] `cargo clippy --all-targets -- -D warnings` — exit 0, 0 errors
- [x] `cargo clippy --no-deps --features openhuman,mcp,telegram --all-targets -- -D warnings` — exit 0, 0 errors
- [x] `cargo build --all-targets` — exit 0
- [x] `cargo test --locked` — 2251 passed, 0 failed
- [x] `cargo test --features openhuman,mcp,telegram` — **3397 passed, 0 failed** (3373 + 10 + 1 + 11 + 2)

The gated lane is the one that matters here: essentially everything touched is gated-only, and the default lane reports 0 for several integration targets — a filter selecting nothing still exits 0.

## Documentation

No doc file changes. The reasoning lives in rustdoc on the types it constrains, which is where this module keeps it: why the handle must stay one per company, why `Unscoped` is not an error, why depth accounting is unaffected by bucketing, and why each `Board` refusal says what it says.

## Related

Part of #661

Mirrors #439 (`ApprovalRequestQueue` scoping) deliberately — two proven-identical solutions rather than one clever one. Follows the #267 `Answering` precedent, inverted. Preserves the #419, #453, #272 and #176 invariants; the #176 scope chain is bucketed with its depth semantics intact.

### DIFF
--- a/src/harness/orchestrator.rs
+++ b/src/harness/orchestrator.rs
@@ -52,7 +52,7 @@
 //!
 //! Compiled only under `feature = "openhuman"` (the whole `harness` module is).
 
-use std::collections::VecDeque;
+use std::collections::{BTreeMap, VecDeque};
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex, OnceLock, Weak};
 
@@ -329,6 +329,66 @@ impl Delegation {
     }
 }
 
+/// Which claimant a queued delegation belongs to (issue #661).
+///
+/// The queue handle is one per company and cannot be otherwise — see
+/// [`DelegationQueue`] — so the separation between concurrent claimants lives
+/// here, in the key, rather than in separate queues. Exactly the shape
+/// [`ApprovalScope`](crate::harness::policy::ApprovalScope) took for the same
+/// race one queue over (issue #439), and deliberately so: two identical
+/// solutions to one problem are worth more than two clever ones.
+///
+/// # Not to be confused with the scope *chain*
+///
+/// [`DelegationQueue::scope_chain`] and [`ScopeGuard`] (issue #176) also say
+/// "scope", and mean something else entirely: the stack of desk ids a hand-off
+/// is currently nested through, whose length **is** the delegation depth. That
+/// chain is per-claimant state like everything else here — each scope gets its
+/// own — but a `DelegationScope` is *which claimant*, never *how deep*.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum DelegationScope {
+    /// A delegation staged outside any run-scoped claim.
+    ///
+    /// The default, and every chat and task path lands here: `run_cycle`, the
+    /// operator-message turn and its relay, and a dispatched card all serialise
+    /// under the cycle lock, so one bucket is all they have ever needed and
+    /// their behaviour is unchanged by this scoping.
+    ///
+    /// Deliberately **not** an error, for the same reason
+    /// [`ApprovalScope::Unscoped`](crate::harness::policy::ApprovalScope)
+    /// is not: a claimant added later that forgets to name a scope degrades to
+    /// today's behaviour rather than to a silently dropped delegation.
+    #[default]
+    Unscoped,
+    /// One workflow run, keyed by its run id.
+    ///
+    /// Workflow runs are `tokio::spawn`ed and are **not** under the cycle lock —
+    /// several genuinely overlap, bounded only by the #401 in-flight cap. That
+    /// is the concurrency this scoping exists for.
+    Run(String),
+}
+
+tokio::task_local! {
+    /// The scope this task's delegation calls file into, installed for the
+    /// duration of a claim by [`DelegationClaim::scoped`] (issue #661).
+    ///
+    /// # Why ambient, and why that is sound here
+    ///
+    /// The delegation tools are constructed with a queue handle and nothing
+    /// else, and they are wired **statically**: belts are cached per roster by
+    /// [`HarnessPool::ensure`](crate::harness::HarnessPool::ensure) and rebuilt
+    /// rarely, so a tool cannot be handed a run id at call time — the same
+    /// constraint that put the #176 depth chain on this queue rather than in a
+    /// parameter.
+    ///
+    /// A task-local is sound on this path because a turn does not leave its
+    /// task, and this is not a new dependency: the queue's neighbours
+    /// ([`ApprovalRequestQueue`](crate::harness::policy::ApprovalRequestQueue))
+    /// and `with_stop_hooks` are already task-local scopes over these same
+    /// turns.
+    static CURRENT_SCOPE: DelegationScope;
+}
+
 /// A shared, in-memory queue the delegation tools push onto and the harness
 /// brain drains. Cheap to [`Clone`] (a shared handle); the same underlying
 /// queue is seen by the tools captured into the orchestrator agent and by the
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
Reviewed for concurrency — whether two claimants can still reach each other's state, and whether a scope can leak. **Not an approval.**

## What holds up

**The isolation is by key, not by lock, and that is the right shape here.** Each of the four maps is keyed by `DelegationScope`, and a workflow run's scope is its own run id, so two concurrent runs cannot collide by construction rather than by careful locking. The commit message's framing is accurate: the old global claim meant *any* claimant's claim destroyed every other's staged delegations and reset a running chain's depth — safe only while everything serialised under the cycle lock, which spawned workflow runs demonstrably do not.

**Lock discipline is enforced by the type system.** These are `std::sync::Mutex`, whose guards are not `Send`, so "held across an `await`" is a compile error rather than a review item. Worth stating because it is the failure mode I went looking for and it cannot occur here.

**No scope leak.** `release` removes from all four maps — `committed.remove`, then `clear_scope` (inner + refused), then `reset_chain` (chains) — so a host that runs many workflows does not accumulate a `Run(id)` entry per run. I checked this specifically because three of the four removals are in helpers and it would have been easy for one to drift.

**`Drop` carrying its own scope is the subtle part and it is correct.** A claim is very often dropped outside its own `scoped` future, so it cannot read the ambient scope — the doc comment says exactly that, and `dropping_a_claim_discards_only_its_own_bucket` pins it. Discarding a cancelled or panicking run's staged writes is also the right semantics, and matching `ApprovalClaim`'s existing stance means there is one rule here, not two.

## One forward-looking concern

`DelegationScope::Unscoped` is `#[default]`, justified as degrading "to today's behaviour rather than to a silently dropped delegation". I agree with the direction — dropping delegations silently would be worse.

The residual is that *today's behaviour is only safe because the Unscoped claimants serialise under the cycle lock*. That is stated in the variant's own doc, but it is a property of the callers, not of the type. A future claimant that is **not** under the cycle lock — another `tokio::spawn`ed path, which is precisely the shape that motivated this PR — and which forgets to name a scope will silently land in the shared `Unscoped` bucket and reintroduce the exact cross-claimant destruction this PR removes. It will do so with no error, no panic, and passing tests, because `Unscoped` is a legitimate value.

Not asking for a redesign; the default is the right trade. But something that makes the invariant enforceable rather than remembered would be worth it — even a sentence on the `Unscoped` variant along the lines of *"only sound for claimants serialised under the cycle lock; a concurrent claimant must name a `Run` scope"* gives the next person the rule at the point they would break it. A `debug_assert` on the concurrent path would be stronger if there is a natural place for one.

This is the same class of thing as the `CallPath::AuthoredWorkflowNode` comment I raised on #776: an invariant that is true today because of who the callers happen to be, recorded where the next caller will read it.

## Not covered

I reviewed the queue's scoping, lock discipline, and claim lifecycle. I did **not** review the drain/execute paths in `orchestrator.rs` outside the queue itself, nor the test additions in detail, and I did not build or run anything.

PR #772, opened 2026-08-12
fix(desktop): package the app in CI, and fix the two bugs that shipped behind a green lane (#616)

### LINKED ISSUE #616: bug: the Desktop CI lane never packages or launches the app, so two blocking bugs shipped green
## Summary

The `Desktop` CI lane proves the desktop crate *compiles*. Nothing proves it *packages* or *runs*. Two bugs shipped to `main` behind a fully green lane, and neither was reachable by any test the lane runs.

The lane (`.github/workflows/ci.yml`) does:

```yaml
- run: npm ci                 # working-directory: frontend
- run: npm run build          # working-directory: frontend
- name: Format   → cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check
- name: Clippy   → cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --no-deps -- -D warnings
- name: Test     → cargo test --manifest-path src-tauri/Cargo.toml --locked
```

It never invokes `cargo tauri build` or `cargo tauri dev`. What that let through:

**1. The app could not be packaged at all.** `tauri.conf.json` carried `"beforeBuildCommand": "npm --prefix ../frontend run build"`. Tauri runs that hook from the *project root*, not from `src-tauri/`, so `../frontend` escaped the repository entirely. Both `cargo tauri build` and `cargo tauri dev` failed on their first step for every developer. CI never executed the hook, because it builds the frontend itself with `working-directory: frontend` and then only calls `cargo` directly.

**2. The desktop bridge never resolved.** Tauri v2 namespaces `invoke` and `Channel` under `window.__TAURI__.core`; the bare `__TAURI__.invoke` is the **v1** shape and is `undefined`. Three readers (`transport/index.ts`, `transport/proxy.ts`, `transport/desktop.ts`) read one level too shallow, so `bridge()` returned `null`, every command silently no-opped, no connection was registered, and the console fell back to a browser error about not reaching a host.

The unit tests could not have caught (2), because they *asserted the wrong shape*: the mocks installed `window.__TAURI__ = { invoke: vi.fn() }`, a shape Tauri v2 never produces. One test's comment even reasoned explicitly that `__TAURI__` "is the one `ProxyTransport.bridge()` calls". 82 desktop tests and 276 console tests passed against a fixture that did not correspond to the runtime.

This is the shape of #475 — a suite that reads as coverage but is selected by nothing that would fail — moved up a level: here it is the *packaging entry point* rather than a test target that no lane exercises.

## Reproduction

On `main` before the accompanying fixes:

```bash
cd src-tauri && cargo tauri build --bundles app
# Error beforeBuildCommand `npm --prefix ../frontend run build` failed with exit code 254
# npm error path /…/tinyhumans.ai/frontend/package.json   ← one level above the repo
```

Then, with only that fixed, the packaged app launches, the embedded host serves, and the console still shows "Can't connect … at this origin" because the bridge is `null`.

Meanwhile `cargo test --manifest-path src-tauri/Cargo.toml --locked` is green throughout, as is the whole `Desktop` lane.

## Expected behavior

The lane should fail when the app cannot be packaged, and ideally when it cannot boot.

Suggested, in increasing order of cost:

1. **Package in CI.** Add `cargo tauri build --bundles app` (or `--no-bundle` for speed) to the `Desktop` lane. This alone would have caught (1) at the commit that introduced it. The lane already installs webkit/GTK, so the marginal cost is the release compile.
2. **Assert the bridge shape against the real API.** A test that reads `invoke` off `@tauri-apps/api` the way the runtime injects it — rather than off a hand-written mock — would have caught (2). At minimum, the mocks must mirror the v2 namespace, and a test should assert the **v1 shape is rejected** so a regression fails rather than passes. (Added in the accompanying PR.)
3. **Smoke-launch the packaged binary headlessly** (`xvfb-run` on the Linux runner), assert the process stays up and the embedded host answers `/healthz`. This is what would have caught the class of failure that only appears when the webview actually loads.

Without at least (1), the lane's green is a statement about compilation only, and the desktop's first real launch stays the integration test.

CodeGhost21:
Closes #616.

The `Desktop` lane proved the crate compiles and nothing else. Two bugs shipped behind a fully green lane, and neither was reachable by any test it runs. This packages the app in CI and fixes both.

> **Rebased onto `main` after #779.** An earlier revision of this PR also deleted `frontend/src-tauri/`, on the grounds that it was an unbuilt workspace member. `85ed495d` has since made it an independent workspace with its own lockfile and added three CI lanes for it, which falsifies that reasoning — so that commit is gone and this PR touches neither the crate, the root manifest, nor the lockfiles. See the note at the end for the one observation that survives.

## What the RCA changed

The issue's two bugs are real. Its prescription for the first is not, and that only shows up when you probe the hook rather than read it.

**Tauri derives `beforeBuildCommand`'s working directory from where the CLI was invoked.** So each candidate value passes in exactly the directory that hides it:

| hook value | from repo root | from `src-tauri/` |
| --- | --- | --- |
| `../frontend` — what shipped | passes | **fails** |
| `frontend` — the obvious fix | **fails** | passes |
| `$(git rev-parse --show-toplevel)/frontend` | passes | passes |

The issue says adding `cargo tauri build` to the lane "alone would have caught (1)". It would not: every other step in that lane runs from the repository root, where the shipped value passes. So the `Package` step runs from **both** directories — the only two that reach this app, since the CLI finds a project by searching *subfolders* of the working directory rather than ancestors.

`frontendDist` is resolved relative to `src-tauri/` and is unchanged. That asymmetry between two keys in one file is the whole trap.

**The bridge bug is confirmed.** `tauri-2.11.5/scripts/bundle.global.js` ends `window.__TAURI__ = __TAURI_IIFE__`, whose top-level keys are `app, core, dpi, event, image, menu, mocks, path, tray, webview, webviewWindow, window` — `invoke` and `Channel` exist only under `core`. One correction to the report: it names three bad readers, but `transport/index.ts` is a presence check and was correct. Two readers, plus a comment that taught the wrong shape.

## Verification

Every claim below was checked by breaking the fix and watching the check fail, not by observing green.

**The bridge fix is now selected by tests.** Reverting the reader to the v1 path turns **11 unit tests** and **5 of 6 desktop e2e specs** red. Before this PR it turned *none* — 82 desktop tests passed against a fixture the runtime never produces, because every mock hand-wrote `{ invoke, Channel }` at the top level. The mocks now install the v2 shape, and a new suite reads the shape off `@tauri-apps/api` itself and asserts the v1 form is **refused**, so the fixture is tied to something other than itself.

Two fixtures added on `main` while this branch was open (`transport-seam.test.ts`, `connection-registry.test.ts`) were still v1-shaped. They pass either way — both only exercise the presence check — but they are normalised here so the tree stops teaching the shape.

**The packaging gate is selected by both regressions.** Each is caught by exactly one of the two `Package` steps; the committed hook passes both. A single-directory step would have gated half the class, and which half would depend on the directory it picked.

**The app really launches.** Built `OpenCompany.app` and ran it, then reran the identical launch with the bug reintroduced as a control:

| | fixed | v1-bridge build |
| --- | --- | --- |
| bundles | yes | yes |
| embedded host listening | yes | yes |
| console→host sockets | **4 ESTABLISHED** | **0 — LISTEN only** |

Those sockets are the core's own connections to its embedded host; they exist only if the console resolved the bridge and drove `oc_request`. Backend confirmed behind it: `/healthz` → 200, `/spec` → the module list.

That table also settles suggestion (3) in the issue: **the embedded host starts and answers `/healthz` even in the broken build**, so a headless `/healthz` smoke test would have passed while the app was unusable. It is proposed there as the strongest option; it is the one that would not have worked.

## Commands run locally, on the merged tree

- `npm test` — 546 passed
- `npm run typecheck:unit`, `npm run typecheck:e2e`
- `npm run e2e` — 115 passed, 10 skipped, 0 failed (real Rust host; run pre-merge)
- `cargo fmt` / `clippy` / `test` on `src-tauri`, and `cargo fmt` on `frontend/src-tauri`
- `tauri build --debug --no-bundle` from the repo root **and** from `src-tauri/`

## Two things a reviewer should weigh

**The hook assumes a POSIX shell and a git checkout.** A Windows `cmd` host would not expand `$(…)`. The `Desktop` lane is ubuntu-only today and `docs/spec/runtime/desktop.md` records this as the line to revisit when a Windows runner appears. The alternative — dropping the hook and requiring `frontend/dist` to be built first — is cwd-independent for free but removes `tauri dev` as a one-command workflow.

**`src-tauri/Cargo.lock` is deliberately untouched.** It fails `--locked` locally (wanting `"url"` added to the `opencompany` entry), but that reproduces on a pristine `main` checkout, so it is not this branch. `rust-toolchain.toml` pins floating `stable`, so a runner's cargo and a local cargo disagree about it. Note that `tauri build` is not `--locked`, so a local packaging run dirties that file.

**Not done, deliberately:** the headless smoke-launch from the issue's suggestion (3), for the reason the control table shows — it would not have caught either bug.

## One observation, not a change

`frontend/src-tauri` and `src-tauri` are both Tauri apps sharing a `productName` of `OpenCompany`, and which one a `tauri` invocation picks up depends on the working directory — from `frontend/` you get the wrapper, from the root or `src-tauri/` you get the desktop client. `npm run tauri:build` therefore builds the wrapper. That is left exactly as `main` has it here.

Worth noting only because the wrapper has the same class of bug this PR fixes: its `beforeBuildCommand` is `pnpm build` with `frontendDist: "../dist"`, and no lane packages it either — the new lanes on `main` run `fmt`/`clippy`/`test`, which is precisely the trio #616 is about not being enough. Happy to open a separate issue if that is wanted.


oxoxDev (approved):
"The `Desktop` lane proved the crate compiles and nothing else" is the sentence this whole repo has needed today, and the RCA behind it is the best part of the PR.

**Probing the hook rather than reading it changed the answer.** The issue's prescription — that adding `cargo tauri build` "alone would have caught (1)" — is wrong for a reason only a table makes visible: Tauri derives `beforeBuildCommand`'s working directory from where the CLI was invoked, so `../frontend` passes from the repo root and fails from `src-tauri/`, and the obvious fix inverts it. Every other step in that lane runs from the root, where the shipped value passes. Running `Package` from **both** directories — the only two the CLI accepts, since it finds a project by searching subfolders rather than ancestors — is the only version of this gate that would have caught the bug it exists for.

That `frontendDist` resolves relative to `src-tauri/` while `beforeBuildCommand` resolves relative to the invocation directory, in the same file, is a genuinely nasty asymmetry and worth the paragraph you gave it.

**Correcting the report rather than implementing it** is right again: the bridge issue named three bad readers, and `transport/index.ts` was a presence check that was already correct. Two readers and a comment teaching the wrong shape is a smaller, truer claim.

**The third desktop app is the finding I would not have gone looking for.** `frontend/src-tauri/` still a tracked root workspace member, still what `npm run tauri:build` packaged because the CLI searches subfolders and that script ran from `frontend/`, and its membership dragging 148 packages — webkit2gtk and GTK included — into the root `Cargo.lock`, which is precisely the tree `desktop.md` documents as avoided. A third instance of #475's shape, and the `-2136` lockfile lines are the receipt.

I checked the two things that would undercut it, and both hold: the `Package` steps carry no `continue-on-error`, so a hook that resolves wrong fails the lane rather than warning; and `frontend/package.json`'s `tauri:build` is now `cd .. && tauri build`, so the script that used to package the stale app now packages the real one. `Cargo.toml` carries the reason for the removed member at the site, which is where the next person will look.

**0 major. 1 question. Approving on the code.**

### Question — `--no-bundle` leaves out the failure class this repo actually ships

Both steps run `tauri build --debug --no-bundle`. That exercises `beforeBuildCommand`, `frontendDist` and the crate build — the two bugs at hand — and it is the right trade for lane time. It does not exercise bundling: icon and installer configuration, and above all **signing**.

That matters here specifically. Signing failures are a live class in this repo — an ad-hoc-signing failure has previously presented as a `-67030` and been misread as a secrets problem — and `--no-bundle` is exactly the flag that skips it. So after this lands, `Desktop` proves the app *builds and its hooks resolve from both directories*, which is a large improvement over "the crate compiles", and still not "the app packages".

Worth stating that boundary in the workflow comment, since the PR's own framing is about a lane proving less than its name implies. If a bundling step is out of scope on runtime cost, saying so at the step is what stops the next person reading `Package` as covering more than it does — the same trap, one layer out.

**Before merging:** the branch is `CONFLICTING` / `DIRTY` and **57 commits behind**. Root `src-tauri/` has taken several changes since you branched — your own #692, #693 and #747 among them — so the rebase is not mechanical and the conflict wants resolving with those in mind rather than by taking either side wholesale.

PR #773, opened 2026-08-12
fix(desktop): give a fresh install a company it can sign into (#632)

### LINKED ISSUE #632: bug: the desktop cannot be signed into — its embedded host has no company and none can be created
## Summary

The packaged desktop cannot be signed into, because its embedded host has no company and the desktop offers no way to create one. Sign-in fails not because authentication is broken but because there is nothing to authenticate against.

The three facts that close the loop:

1. **Sign-in is per-company.** The routes are `/api/v1/companies/{company}/auth/login` and `/api/v1/companies/{company}/auth/verify`. A session is scoped to a company; there is no host-level sign-in.

2. **The embedded host boots with no companies.** `src-tauri/src/embedded.rs:74` starts it from defaults:

   ```rust
   let config = AppConfig {
       bind: "127.0.0.1:0".to_string(),
       ..AppConfig::default()
   };
   let state = AppState::new(config).with_home(instance.home().to_path_buf());
   ```

   No company is loaded and no admin email is set. The data root confirms it — after several launches it holds only `.lock`, `instance-id`, and `openhuman/`.

3. **Creating a company requires the `platform` scope**, which a desktop user has no way to hold. `src/server/provision.rs` gates `POST /api/v1/companies` on it — that endpoint exists for the hosting control plane, not for a person on their own laptop.

So the console does the right thing at every step and still dead-ends: `GET /api/v1/companies` returns 401, `ConnectionConsole.tsx` maps 401 to the login view, and the login view targets a company that does not exist and cannot be created.

## Reproduction

Against the embedded host of a launched packaged app (its address is in the log line `embedded host listening address=127.0.0.1:<port>`):

```bash
P=127.0.0.1:<port>

curl -s -o /dev/null -w '%{http_code}\n' http://$P/healthz            # 200 — host is healthy
curl -s -o /dev/null -w '%{http_code}\n' http://$P/api/v1/companies   # 401 — needs a session

# A session needs a company. Try to create one, with a valid manifest:
curl -s -w '\nHTTP=%{http_code}\n' -X POST http://$P/api/v1/companies \
  -H 'content-type: application/toml' \
  --data-binary 'name = "Acme"
[users]
admins = ["someone@example.com"]
'
# {"error":"unauthorized","code":"unauthorized"}
# HTTP=401
```

There is no path from a fresh install to a usable console.

## Expected behavior

A person who installs the desktop and opens it should end up somewhere they can work, without a terminal and without a platform credential.

The shape is a product decision, so listing options rather than asserting one:

- **Provision a local company on first run.** The desktop *is* the platform for its own machine, so it is the natural holder of platform scope against its own embedded host — it could create a default company on first launch and sign the local operator in. This keeps the deployed authorization model intact while giving the desktop the one thing it is missing.
- **Ship a default single-company manifest** and have `embedded::start` load it, the way `opencompany serve --company companies/<name>` does. Simpler, but bakes in a company identity the operator may not want.
- **Add a first-run screen** that collects a company name and admin email, then does whichever of the above.

Whatever the choice, `embedded::start` needs to pass more than `AppConfig::default()`, and the console needs a state for "this host has no companies yet" that offers to create one rather than showing a login form for a company that does not exist.

## Related

- #613 — the desktop opens on a same-origin connection that can never work. Fixing that reveals *this*: with the correct connection selected, the login dead-end is what remains.
- #616 — the Desktop CI lane never launches the app, which is why none of this surfaced before.

CodeGhost21:
Closes #632.

## The dead end

A packaged desktop booted an empty registry, and an empty registry cannot be signed into. Sign-in is per-company (`/api/v1/companies/{id}/auth/…`, or the sole-company alias), so a fresh install rendered a login form addressing a company that did not exist — and neither way a company normally reaches the registry is open to it:

- nobody types `serve --company <dir>` at a double-clicked application;
- `POST /api/v1/companies` requires the `platform` scope, which `PlatformScope` grants only against a configured `platform_auth`. A prosumer host has no machine credential to hand out, deliberately (`src/server/platform_auth.rs`), so every provisioning attempt is a 401 by design rather than by accident.

## Root cause

The bootstrap existed and the desktop rewrite orphaned it. `src/desktop.rs::start_local` still loads a bundled preset and adds `DESKTOP_OPERATOR_EMAIL` as a standing admin, and `tests/desktop_e2e.rs` still proves that path signs in end to end. Its only caller is `frontend/src-tauri/` — the *old* shell (`ai.opencompany.desktop`). The packaged app is the root `src-tauri/` (`ai.tinyhumans.opencompany`, the only one the `Desktop` CI job builds), which carried over the host, proxy, keychain and ACP but not the company bootstrap. #616 is why nobody saw it.

## The fix

**`bootstrap_companies` (`src/desktop.rs`)** — adopts every bundle the data root already holds, then seeds the default preset only when it holds none:

- adoption before seeding, because seeding unconditionally hands the operator a second starter company per launch, and skipping adoption makes the one they have unreachable;
- `archived` companies are skipped — archiving removes a company from the registry on purpose, and re-registering it next launch would undo that quietly;
- the bundle is the only authority (a desktop company has no source directory), and `RuntimeBuilder::build` carries the persisted record's console-created desks, agents and workflows forward.

**`embedded::start`** calls it before binding and sets `admin_email` to the local operator — the same seam the hosted control plane fills with `OPENCOMPANY_ADMIN_EMAIL`. Without an eligible address, `eligibility` admits nobody and the seeded company would still be unenterable.

**`oc_embedded` + the console** carry `operatorEmail`, offered on the embedded connection only. Adopted through an effect rather than initial state: a relaunch restores the connection from its profile before the bridge answers, so on every launch but the first the form mounts first. Only ever into an empty field, so it never overwrites what someone typed and never undoes a deliberate clear.

Nothing else changed: the host binds loopback, so `auth/request` already returns the code in its own response and the console already redeems it.

## Testing

- `cargo test` (2215 pass), `cargo test --manifest-path src-tauri/Cargo.toml --locked` (87 pass), `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` on both crates.
- `cargo test --manifest-path src-tauri/Cargo.toml` adds `a_fresh_install_can_be_signed_into`, which drives a real embedded host over HTTP: fresh root → 401 on `/api/v1/companies` → `auth/request` → `dev_code` → `auth/verify` → 200 → the refused call answers. Plus `a_relaunch_reuses_the_company_the_root_already_holds`.
- Console: 3 typecheck configs, 523 unit tests, and a new Playwright spec (`desktop-first-run-signin.spec.ts`) covering the late-arriving suggestion. Confirmed it is a real regression test — with the prefill disabled it fails `Expected "operator@opencompany.local", Received ""`.
- **Manually, against the built `opencompany-desktop` binary on a fresh data root:** it logs `seeded the first-run company`, the root gains `companies/agentic-marketing-agency`, the issue's own curl repro now completes (401 → `dev_code` → 200 → `/api/v1/companies` lists the company), and the real console bundle in a real browser against that host signs in and lands on the org graph. A relaunch on the same root re-adopts (`companies=1`, no second bundle) and signs in again.

## Notes

- Not covered: the packaged `.app` was not launched and clicked (that is #616's gap). The console half was driven in a real browser against a real embedded host with only `window.__TAURI__` shimmed, since that object exists nowhere else.
- `src-tauri/Cargo.lock` is deliberately left at `main`. `--locked` fails locally, wanting `url` under the `opencompany` entry, while the `Desktop` lane passes — a floating `stable` toolchain disagreeing with the runner's cargo, not a real drift. The desktop suite was run without `--locked` here.
- Out of scope, worth their own issues: the embedded host still starts no schedulers or mailbox pollers for its companies, and `frontend/src-tauri/` is now fully orphaned.

## Summary by CodeRabbit

- **New Features**
  - Desktop startup now automatically prepares the default company on first launch.
  - Existing non-archived companies are reused across relaunches.
  - Local sign-in forms can suggest the desktop operator’s email address, which remains editable.
  - Sign-in now provides a local login code without sending email.
  - Company listings are available after signing in.

- **Documentation**
  - Added documentation describing first-run desktop setup and embedded-host sign-in behavior.

- **Tests**
  - Added coverage for first-run setup, relaunch persistence, sign-in suggestions, and archived-company handling.

### DIFF
--- a/docs/spec/runtime/desktop.md
+++ b/docs/spec/runtime/desktop.md
@@ -193,6 +193,48 @@ valid evidence about embedded mode too.
 data root. The console renders that as a row; the desktop still holds remote
 hosts, which is the point of holding several.
 
+### First run
+
+`embedded::start` calls `opencompany::desktop::bootstrap_companies` before it
+binds, because a host with an empty registry cannot be signed into
+([issue #632](https://github.com/tinyhumansai/opencompany/issues/632)). Sign-in
+is per-company — `/api/v1/companies/{id}/auth/…`, or the sole-company alias —
+so an empty registry leaves the console rendering a login form for a company
+that does not exist.
+
+The two ways a company normally reaches the registry are both closed to a
+packaged application. Nobody types `serve --company <dir>` at a double-clicked
+app, and `POST /api/v1/companies` demands the `platform` scope, which
+`PlatformScope` grants only against a configured `platform_auth` — a prosumer
+host has no machine credential to hand out, deliberately. So the desktop
+bootstraps its own:
+
+1. **Adopt** every company bundle the data root already holds, skipping
+   `archived` ones (archiving removes a company from the registry on purpose,
+   and re-registering it at the next launch would undo that quietly). The
+   bundle is the only authority — a desktop company has no source directory to
+   re-read — and `RuntimeBuilder::build` carries the persisted record's console
+   -created desks, agents and workflows forward.
+2. **Seed** the `DEFAULT_PRESET_ID` preset when there were none, stamping the
+   preset slug as the record's template provenance. Fallback rather than
+   unconditional: seeding on every launch would hand the operator a second
+   starter company per run.
+
+`AppConfig.admin_email` is set to `DESKTOP_OPERATOR_EMAIL` — the same seam the
+hosted control plane fills with `OPENCOMPANY_ADMIN_EMAIL`, and the reason a
+person is eligible to sign in at all (`eligibility` in `src/server/users/`
+admits an existing user, a bootstrap admin, or an invite, and a fresh install
+has none of the three). The seeded manifest names the same address in its own
+`[users].admins`, so the company is self-describing if it is ever served
+elsewhere.
+
+Nothing is mailed: the host binds loopback, so `is_local_only` holds and
+`auth/request` returns the login code in its own response (`dev_code`), which
+the console redeems in place. `oc_embedded` carries `operatorEmail` so the
+sign-in form can offer the address — a person cannot guess it, and every other
+address gets the same silent `202`. It is a suggestion, not a lock: the field
+stays editable, which is what an operator who invites someone else needs.
+
 ### One row, however many launches
 
 The ephemeral port is not free: it means the embedded host's *address* is
--- a/docs/spec/runtime/desktop.md
+++ b/docs/spec/runtime/desktop.md
@@ -213,8 +213,8 @@ bootstraps its own:
    `archived` ones (archiving removes a company from the registry on purpose,
    and re-registering it at the next launch would undo that quietly). The
    bundle is the only authority — a desktop company has no source directory to
-   re-read — and `RuntimeBuilder::build` carries the persisted record's console
-   -created desks, agents and workflows forward.
+   re-read — and `RuntimeBuilder::build` carries the persisted record's
+   console-created desks, agents and workflows forward.
 2. **Seed** the `DEFAULT_PRESET_ID` preset when there were none, stamping the
    preset slug as the record's template provenance. Fallback rather than
    unconditional: seeding on every launch would hand the operator a second
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -221,6 +221,7 @@ function Console() {
         id: host
           ? adoptEmbeddedHost({ baseUrl: host.baseUrl, instanceId: host.instanceId })
           : null,
+        operatorEmail: host?.operatorEmail,
       });
     });
     return () => {
@@ -415,6 +416,12 @@ function Console() {
             defaultCompany={active.defaultCompany}
             notice={active.id === bootstrapId ? auth.notice : undefined}
             forceLogin={active.id === bootstrapId && auth.failed === true}
+            // Only ever the embedded host's own operator. Offering it on a
+            // remote connection would put a local address in front of someone
+            // signing in to a server that has never heard of it.
+            suggestedEmail={
+              active.id === embedded.id ? embedded.operatorEmail : undefined
+            }
           />
         ) : (
           <NoConnection starting={!embedded.resolved} />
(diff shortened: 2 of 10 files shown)


oxoxDev (approved):
The dead end is stated exactly right: sign-in is per-company, an empty registry has no company to address, and **neither way a company normally reaches the registry is open to a double-clicked application** — nobody types `serve --company`, and `POST /api/v1/companies` is a 401 by design because a prosumer host has no machine credential to hand out. Naming that second one as deliberate rather than broken is what makes this a missing-bootstrap problem rather than an auth bug.

The ordering inside `bootstrap_companies` is the part worth having got right, and each rule has a consequence behind it:

- **adoption before seeding**, because seeding unconditionally hands the operator a second starter company per launch, and skipping adoption makes the one they have unreachable;
- **`archived` skipped**, because archiving removes a company from the registry on purpose and re-registering it next launch would quietly undo an operator's decision;
- **the bundle as sole authority**, since a desktop company has no source directory.

Setting `admin_email` to the local operator — the same seam the hosted control plane fills with `OPENCOMPANY_ADMIN_EMAIL` — is the right reuse. Without an eligible address `eligibility` admits nobody and the seeded company is unenterable, so the two halves genuinely have to land together.

The console detail is careful in a way that is easy to skip: adopting `operatorEmail` through an effect rather than initial state, because a relaunch restores the connection from its profile before the bridge answers, and only ever into an **empty** field so it never overwrites what someone typed nor undoes a deliberate clear.

**0 major. 1 sequencing item. Approving on the code.**

### This and #772 finish each other's job, and the second one to land leaves a dead path

They share three files and merge with **0 conflicts**, so nothing will complain. The interaction is semantic:

- Your RCA here is that the desktop rewrite orphaned the bootstrap, and `start_local`'s only caller is the old shell, `frontend/src-tauri/`.
- **#772 deletes `frontend/src-tauri/` outright.**
- This branch still carries `frontend/src-tauri/src/main.rs:23` calling `start_local`, because it was cut before that deletion.

After both land, `start_local` has no production caller at all — the new path goes through `embedded::start` → `bootstrap_companies` — and its remaining callers are `src/desktop.rs:324` and `tests/desktop_e2e.rs`. Which means the e2e you cite as proof "that path signs in end to end" would be proving an entry point the product no longer has. That is the same shape #772 is written to remove: a green test standing in for a thing nobody runs.

Two clean outcomes, and the choice is yours:

- **Retire `start_local` with the shell** and repoint `desktop_e2e.rs` at `embedded::start` + `bootstrap_companies`, so the end-to-end sign-in proof covers the flow that actually ships;
- **Keep it deliberately** as the library-level entry point, with a note saying so — the #724 pattern — and be explicit that the e2e covers the library path rather than the packaged app.

What I would avoid is landing both and discovering later that the one test proving a fresh install can sign in has quietly stopped describing the shipped one.

**Before merging:** `CONFLICTING` / `DIRTY` and **57 commits behind**, same as #772. Root `src-tauri/` has moved under both of these — your own #692, #693 and #747 among the reasons — so resolve with those in view rather than taking a side wholesale, and land #772 and this one in a deliberate order rather than whichever goes green first.

Issue #752, opened 2026-08-12
security: agent shell confinement is policy-level, not a boundary — and the write tier widens what that costs


oxoxDev:
## Summary

`shell` confinement for agent tools is path- and policy-level, not a kernel boundary. #245 said so plainly and shipped anyway, because a read-only single-repo PAT made the trade acceptable. The write tier changes that trade, and nothing tracks the isolation work.

## Problem

#245 completed a triangle it named in its own Notes section: arbitrary code execution via native-runtime `shell`, a workspace full of third-party-authored content, and a live GitHub credential with egress. Every design choice in that issue exists to keep those three from composing — the credential never touches workspace, argv, env or persisted git config; the workspace clone's `origin` is a local path; the mirror refuses every push it could receive.

Those are good defences and they hold against the paths they cover. What they do not do is establish a boundary. From #245:

> `workspace_only` confinement for file tools is robust; shell confinement is path- and policy-level, **not a kernel boundary**. […] Honest verdict: adequate against accident and acceptable for a read-only v1; **not a hard boundary against a determined prompt-injected agent**. Container-level egress policy and per-agent user namespaces are the real fix and should be filed as a follow-up, not claimed here.

That follow-up was never filed. This is it.

The reason it matters now rather than at leisure: the write tier (#247, #734, #735, #736, #738) moves the credential from "read one repository" to "push to it". The blast radius of a prompt-injected agent goes from *reads code it was going to be shown anyway* to *writes code a human may merge*. #735's approval gate is a real mitigation and I am not arguing against it — but an approval gate is a policy control living in the same process as the agent that would need to be contained.

Note also that the harness process holding the secret runs in the same container as `shell`, and the fs storage backend keeps secret bytes on `/data`. Hosted tenants on mongodb storage avoid the second half; a local or fs-backed deployment does not.

## What would actually help

Roughly in order of value per unit of work — this is a starting position, not a design:

1. **Egress policy at the container**, so an exfil attempt fails at the network rather than at a tool-name check. Today `web_allowed_domains` is a policy list an agent's own tools consult; it is not a boundary either.
2. **Per-agent user namespace / uid separation**, so one agent's workspace is not readable by another's shell, and neither can read the harness process's memory or the secrets path.
3. **An audit trail that survives the agent** — shell calls are already audited; the question is whether the audit sink is reachable from inside the blast radius.

## Scope and honesty

This is not a small piece of work and it should not be smuggled into a feature PR. Filing it so that:

[...]

Issue #707, opened 2026-08-11
runtime: five operator-writable overlay fields have no fingerprint, so a console change may not reach a resident company


M3gA-Mind:
**Status: the mechanism is confirmed by reading the code; the runtime consequence is NOT yet reproduced.** This is filed as an investigation with the evidence and the exact experiment, deliberately not as a defect report — see "What is not established" below.

## Confirmed

`HarnessPool::ensure`'s staleness fast-path (`src/harness/mod.rs:990-1007`) returns early unless one of **seven** fingerprints changed: mcp, overlay, capability, composio, skill, budget, and policy (the last added in #659).

`CompanyRecord` carries **eight** operator-writable overlay fields:

| field | covered by a fingerprint? |
|---|---|
| `overlay_agents` | yes — `overlay_fingerprint` |
| `overlay_budgets` | yes — `budget_fingerprint` (#343) |
| `overlay_policy` | yes — added in #659 |
| `overlay_desk_members` | **no** |
| `overlay_desk_order` | **no** |
| `overlay_desks` | **no** |
| `overlay_workflows` | **no** |
| `disabled_workflows` | **no** |

`overlay_fingerprint`, despite the general name, hashes `overlay_agents` **only** — the signature settles it:

```rust
fn overlay_fingerprint(agents: &[OverlayAgent]) -> u64   // src/harness/mod.rs:1859
```

The desk overlays are consumed at **build time**: `src/runtime/builder.rs:1450` reads `overlay_desk_order` off the existing record and seeds the brain with it, with the comment at `builder.rs:1838-1843` stating why —

> *"Seed the brain from the persisted overlays so desk routing (`desk_lead` → `effective_desk_members` → `overlay_desk_order`) reflects the operator's current hierarchy, not the blueprint default."*

The pre-rebuild fold-in sets `overlay_agents`, `overlay_budgets` and now `overlay_policy` — **but not the desk overlays** — so even a rebuild triggered by some *other* axis takes desk state from a possibly-stale snapshot.

## What is not established

Whether a resident company actually routes on stale desk order in practice. Build-time consumption plus an uncovered fingerprint is the right *shape* for the bug, but the consequence has not been observed. Filing it as a confirmed defect would send someone building against something that may not be there — which is what happened on #681.

## The experiment that settles it

1. Add a desk-order overlay to a **resident** company through the console route.
2. Dispatch a turn **without** touching agents, budgets, policy, mcp, or skills — so no covered axis moves.
[...]

PR #774, opened 2026-08-12
security: agent-isolation threat model, and fail closed on the fs secret backend (#752 C0+C3)

oxoxDev:
## Summary

The in-repo half of #752: the threat model **C0**, and the fail-closed check
**C3**. `Part of #752` — the issue stays open. The load-bearing children (C1
default-deny egress, C2 pod `securityContext`, C5 short-lived installation
tokens) are in `opencompany-microservice` and cannot be closed from here.

**C0 — `docs/spec/security/agent-isolation.md`.** The threat model #245
deferred when the repo read tier shipped ("should be filed as a follow-up, not
claimed here") and nobody filed. It states what is enforced today and where,
what is not, and — the section that is the point — the attacker capability that
survives **every** control C1–C5, written as if they have all landed. That
section exists so nobody reads a closed child issue as "the shell is contained
now".

Three findings behind it, each re-verified in this branch:

1. **OpenCompany never sets a sandbox mode.** `ShellTool` routes to a sandbox
   backend only when the mode is `Sandboxed`
   (`vendor/openhuman/.../tools/impl/system/shell.rs`); `SandboxMode::default()`
   is `None` and grepping `src/` for `SandboxMode` / `jail` / `landlock` returns
   doc comments only. Every shell command is a direct child of the server
   process. `cwd_jail` is neither wired nor compiled — `sandbox-landlock` is
   absent from the tenant feature set in `deploy-staging.yml`.
2. **Landlock would not close the exfiltration channel.** From
   `vendor/openhuman/src/openhuman/sandbox/cwd_jail/README.md`, verbatim:
   *"Landlock does not gate network at all."*
3. **Secrets are plaintext on `/data`.** `FsSecretStore` says so in its own doc
   comment; `sqlite` puts the same bytes on the same disk. Readable by the uid
   the shell runs as.

A fourth turned up while writing it and is in the doc: **the shell audit log is
`<workspace>/audit.log`** — the agent's own working directory. Withholding
shell when the logger cannot initialise makes the log's *existence*
fail-closed; nothing makes its *integrity* so. An agent holding `shell` can
truncate the record of what it did.

**C3 — a repository credential requires `OPENCOMPANY_STORAGE=mongodb`.**

### Where the refusal went, and why

Three places, because each catches a case the others structurally cannot:

| Where | Catches | Why not the others |
| --- | --- | --- |
| `RepoManager::bind_validated` → `409 Conflict` | A new credential | The last moment the operator is still holding the token and can act — the same argument the classic-PAT refusal already makes one line below. In `bind_validated`, not `bind`, because that is the one funnel every bind path reaches including the one tests drive; a gate in the public wrapper is a gate the most realistic exercise walks past. |
| `RuntimeBuilder::build` → the company does not come up | A tenant that bound a credential **before** this gate existed | Bind-time cannot see it — that token is already on `/data` and its agents would keep checking out under it forever. Read over the **effective roster**, not `[tools].allow` alone: an agent naming `tools = ["repo"]` under a company `allow = ["*"]` holds an explicit grant the company line does not show, because the wildcard deliberately does not confer the namespace. |
| `build_agent` → repo tools withheld (gate 4 of 4) | A teammate added through the console on a **live** runtime | `HarnessPool::ensure` rebuilds that agent's belt each turn without going back through `build`, so the boot check never sees it. |

Bind-time alone leaves an already-bound tenant running. Boot alone is a worse
operator experience and misses the console path. The condition is asked once, in
`StorageKind::secrets_are_plaintext_on_disk`, and the message comes from one
function so an operator who meets it in a console toast and again in a boot log
does not have to work out whether they are two problems.

Deliberately **not** proposed: a shell command allowlist or regex blocklist.
That is this codebase's documented rejected anti-pattern and would be the wrong
centrepiece. Nothing here sandboxes or contains the shell, and the doc says so
in as many words.

## API Or Behavior Changes

**Breaking, and intended.** A deployment on `OPENCOMPANY_STORAGE=fs` (the
default) or `sqlite` that has bound a repository, or whose roster grants `repo`,
**will stop working**:

- `POST …/repos` answers `409 Conflict` instead of binding.
- A company whose effective roster grants `repo` **fails to start**, with
  `OpenCompanyError::Config`.
- On a live runtime, an agent granted `repo` gets no `repo_checkout` /
  `repo_pr`, with an `error`-visible warning.

Migration, named in the refusal itself: set `OPENCOMPANY_STORAGE=mongodb` with
`OPENCOMPANY_MONGODB_URI`, **or** drop the `repo` grant from the company's
`[tools]` allow list and from every agent naming it.

This is failing closed rather than warning on purpose. The deployments it
breaks are exactly the ones carrying a plaintext repository token on a disk
their own agents can read; a warning would leave that in place and call it
handled. Operators hitting it should also **revoke the bound credential at the
forge** — it has been readable for as long as it has been installed.

Hosted tenants are unaffected: the manager injects
`OPENCOMPANY_STORAGE=mongodb`.

New public API: `StorageKind::secrets_are_plaintext_on_disk()`,
`store::plaintext_secret_refusal()`, `RuntimeBuilder::with_storage_kind()`,
`RepoManager::with_storage_kind()` / `::secrets_are_plaintext_on_disk()`,
`AppState::storage_kind()`. `RuntimeBuilder` and `RepoManager` default to `Fs`
— the refusing side — so a caller that forgets to thread the backend through is
treated as unsafe rather than as safe.

## Tests

Modelled on #245's own `a_checkout_cannot_reach_the_mirror_it_came_from`: an
attack that must fail, not a unit test of the guard.

- `a_credential_cannot_be_installed_where_the_agent_shell_could_read_it` —
  installs a credential on a real `FsSecretStore`, then walks the company's
  whole on-disk footprint the way a shell with `cat`/`grep` would and finds
  nothing. Same call, same fixture remote as the accepted case; the only
  difference is the backend.
- `the_same_bind_is_accepted_where_secrets_leave_the_container` — the other
  half, without which the first only proves the feature is off. `sqlite` is
  refused alongside `fs` in the same test.
- `a_repo_granted_company_does_not_boot_on_a_plaintext_secret_backend` and
  `an_agent_that_names_repo_under_a_wildcard_company_is_caught_too`, against
  `the_same_company_boots_where_secrets_leave_the_container`.
- `binding_is_refused_on_a_host_that_keeps_secrets_on_its_own_disk` — the route,
  asserting `409`, the message, and that nothing was stored.
- Gate 4 folded into
  `repo_tools_are_wired_only_by_explicit_grant_a_manager_and_a_binding`.
- `only_mongodb_keeps_secrets_off_the_local_disk` and
  `the_refusal_names_the_condition_and_both_remedies`.

No test asserts on a real credential; every fixture token is an obvious dummy,
and two tests assert the refusal does **not** echo it.

**Teeth proof.** Each refusal was reverted in turn and every test observed
going RED, then restored. Verbatim:

```
---- runtime::repo_manager::test::a_credential_cannot_be_installed_where_the_agent_shell_could_read_it stdout ----
called `Result::unwrap_err()` on an `Ok` value: RepoBinding { key: "acme-widgets-…", … }

---- runtime::builder::test::a_repo_granted_company_does_not_boot_on_a_plaintext_secret_backend stdout ----
an fs host must refuse to bring up a repo-granted company: CompanyRuntime { id: CompanyId("acme"), … }

---- runtime::builder::test::an_agent_that_names_repo_under_a_wildcard_company_is_caught_too stdout ----
an agent-level `repo` grant must be caught at boot: CompanyRuntime { id: CompanyId("acme"), … }

---- harness::build::tests::repo_tools_are_wired_only_by_explicit_grant_a_manager_and_a_binding stdout ----
a repo grant on fs must wire nothing: [… "repo_checkout", "repo_pr"]
```

The accepted-case tests stayed green throughout, which is what makes the red
ones mean something.

Gate 4 also narrowed `every_registered_tool_is_declared`, whose widest-belt
fixture built its `RepoManager` with the default backend — and that test caught
it by name ("the belt builder stopped wiring `repo_checkout`, so this check has
narrowed without anyone deciding to narrow it"). The fixture now declares a
mongodb backend for the same reason it already declares a binding. Worth
recording: the coverage guard did exactly the job it was written for.

Commands run locally, with counts (a filter selecting nothing exits 0, so the
counts are the receipt):

- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo clippy --no-deps --features openhuman,mcp,telegram --all-targets -- -D warnings` — clean
- [x] `cargo build --all-targets` — via `cargo check --all-targets` and the test builds below
- [x] `cargo test --locked` — **2259 passed, 0 failed, 2 ignored** (2248 lib + 10 bin + 1 `desktop_e2e`)
- [x] `cargo test --features openhuman,mcp,telegram` — **3398 passed, 0 failed, 4 ignored** (3374 lib + 10 bin + 1 `desktop_e2e` + 11 `one_card_per_message` + 2 `product_identity`) (no CI lane builds this combo)

The MongoDB lane is not exercised: it skips silently without
`OPENCOMPANY_TEST_MONGODB_URI` while still exiting 0. Every "accepted on
mongodb" assertion here is against the **selection**, not a live MongoDB — which
is the right level, since the gate is on `OPENCOMPANY_STORAGE` and nothing else.

## Documentation

- `docs/spec/security/agent-isolation.md` — new (243 lines, under the 500-line
  cap). Linked from `docs/spec/README.md` (index + a Security reading path) and
  `docs/spec/runtime/README.md`.
- `docs/spec/runtime/storage.md` — a "Secrets at rest, and what that costs"
  section: the three enforcement points, the breaking change, and the migration.
- `docs/spec/runtime/repos.md` — the requirement as an intake rule, and its
  "The honest limit" section now points at the threat model it was owed.

`docs/spec/security/` has no `README.md`. The repo convention asks for one when
a topic *outgrows a file and splits into a directory*; this is one file. Happy
to add one if maintainers would rather every `docs/spec/` subdirectory have an
entrypoint.

Part of #752

### DIFF
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -54,6 +54,8 @@ L0  Substrate       api.tinyhumans.ai, openhuman-core, tiny.place, filesystem
 - **Runtime engineering**: [runtime/](runtime/README.md) →
   [company-brain/](company-brain/README.md) →
   [integrations/](integrations/README.md)
+- **Security**: [security/agent-isolation.md](security/agent-isolation.md) —
+  read this before assuming any agent capability is contained
 - **Where this is going**: [roadmap.md](roadmap.md) →
   [feature audit](feature-audit/README.md) →
   [vision/](vision/README.md)
@@ -94,6 +96,7 @@ L0  Substrate       api.tinyhumans.ai, openhuman-core, tiny.place, filesystem
 | [runtime/repos.md](runtime/repos.md) | Bound repositories: mirror cache, credential handling, quota |
 | [runtime/data-root.md](runtime/data-root.md) | Data-root resolution, the single-writer lock, instance identity |
 | [runtime/desktop.md](runtime/desktop.md) | The desktop client: connections, transport seam, embedded host |
+| [security/agent-isolation.md](security/agent-isolation.md) | What confines an agent and what does not — enforced controls, the gaps, and the capability that survives every planned control |
 | [company-as-agent/README.md](company-as-agent/README.md) | Companies as economy citizens |
 | [company-as-agent/identity.md](company-as-agent/identity.md) | Wallet, handle, Agent Card |
 | [company-as-agent/commerce.md](company-as-agent/commerce.md) | Selling, hiring, delegated signers, ledger |
--- a/docs/spec/runtime/README.md
+++ b/docs/spec/runtime/README.md
@@ -68,6 +68,10 @@ Supporting docs:
   credential reaches git without entering argv, the environment or any file, the
   alternates-not-hardlinks and refuse-not-evict departures, and the honest limit
   of same-user confinement
+- [../security/agent-isolation.md](../security/agent-isolation.md) — the threat
+  model behind that limit: what confines an agent today, what does not, and what
+  a prompt-injected agent with `shell` can still do after every planned control
+  lands
 - [users.md](users.md) — human collaborators: magic-link/password sign-in,
   sessions, invites, and chat attribution
 
--- a/docs/spec/runtime/repos.md
+++ b/docs/spec/runtime/repos.md
@@ -147,6 +147,11 @@ different ones, and they are named rather than implied:
 - container-level egress policy and per-agent user namespaces, which are the
   real fix and are a follow-up, not a claim made here.
 
+That follow-up is issue #752, and the threat model it owed is now written:
+[../security/agent-isolation.md](../security/agent-isolation.md). Read it before
+concluding that any of the above adds up to confinement. Its residual section
+is the part that matters — it holds after every control #752 tracks has landed.
+
 ### Repository configuration decides what runs
 
 Issue #459 reclassified `read_workspace_state` for exactly this reason: it
@@ -334,6 +339,16 @@ of knowing what the company is, and the body carries no credential material.
 
 ### Intake rules
 
+- **The host must keep secrets off its own disk** (issue #752). On
+  `OPENCOMPANY_STORAGE=fs` or `sqlite` the credential would land as plaintext
+  under the company's data directory, readable by the uid the agent shell runs
+  as, so the bind is refused with `409 Conflict` before anything is parsed or
+  written. The same condition refuses **boot** for a company whose roster grants
+  `repo`, and withholds the repo tools at agent build. Remedies:
+  `OPENCOMPANY_STORAGE=mongodb` with a Mongo URI, or drop the `repo` grant.
+  This is a breaking change for `fs` deployments that already bound one — see
+  [storage.md](storage.md) and, for why a warning would not have been honest,
+  [../security/agent-isolation.md](../security/agent-isolation.md).
 - **https://github.com/&lt;owner&gt;/&lt;repo&gt; only.** Another forge, `http`,
   `ssh`, a userinfo section, a port, extra path segments, a query string, a `..`
   — each is refused rather than normalized. A URL is the thing a credential is
--- a/docs/spec/runtime/storage.md
+++ b/docs/spec/runtime/storage.md
@@ -39,6 +39,41 @@ MongoDB settings:
 - `OPENCOMPANY_TENANT_ID` — tenant identity for **shared-single-DB** mode
   (default unset). See [Shared single database](#shared-single-database-mode).
 
+## Secrets at rest, and what that costs (issue #752)
+
+`fs` writes one **plaintext** file per secret under
+`<data-dir>/companies/<slug>/secrets/`; `sqlite` puts the same bytes in a
+database file on the same disk. Both are readable by the uid the server runs
+as — which is the uid an agent's `shell` tool runs as, in the same container.
+There is no boundary in between; see
+[../security/agent-isolation.md](../security/agent-isolation.md). `mongodb` is
+the only backend that keeps secrets out of the container, in the tenant
+database.
+
+**Repository credentials therefore require `OPENCOMPANY_STORAGE=mongodb`.**
+This is enforced, not advised, in three places:
+
+| Where | What happens |
+| --- | --- |
+| `POST …/repos` (bind) | `409 Conflict` with the refusal message; nothing is stored |
+| Company boot / rebuild | The company does not come up — `OpenCompanyError::Config` — when its **effective roster** grants `repo`, including an agent naming `repo` under a wildcard company allow-list |
+| Agent build | Repo tools are withheld (fail-closed, with a warning), which covers a teammate added through the console on a live runtime |
+
+The refusal names both remedies: set `OPENCOMPANY_STORAGE=mongodb` with
+`OPENCOMPANY_MONGODB_URI`, or drop the `repo` grant.
+
+**This is a breaking change for `fs` and `sqlite` deployments that have already
+bound a repository.** That is deliberate. Those are precisely the deployments
+carrying a plaintext repository token on a disk their agents can read, so a
+warning would leave the exposure in place and call it handled. Migration is one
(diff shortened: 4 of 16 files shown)


M3gA-Mind:
Reviewed the security half closely — the fail-closed decision, where the gates sit, and whether any path reaches a credential around them. **Not an approval**, just findings.

## What holds up

**The default is the refusing side, and that is load-bearing.** `RepoManager::new` seeds `storage_kind: StorageKind::default()`, `StorageKind`'s `#[default]` is `Fs`, and `Fs.secrets_are_plaintext_on_disk()` is `true` — so a manager assembled without being told refuses rather than permits. `only_mongodb_keeps_secrets_off_the_local_disk` pins exactly that with `StorageKind::default()`, which is the assertion that keeps it true when a fourth backend is added. That is the right shape for a control like this.

**Gate placement in `bind_validated` rather than `bind` is correct**, and the doc comment gives the real reason: it is the one funnel every bind path reaches, including the tests that drive the credentialed path directly. A gate in the public wrapper is a gate the most realistic exercise walks past — that is how a control ends up proven only against itself.

**Gate 4 in `build_agent` is the non-obvious one and it is justified.** A teammate added through the console lands in `overlay_agents` on a *live* runtime, and `HarnessPool::ensure` rebuilds that belt without going back through `RuntimeBuilder::build`, so the boot check never sees it. I verified that seam separately today on #707 — a build-time check genuinely does not run again for a console-added teammate. Without gate 4 this would be reachable.

I checked the grant matching against the implementation rather than the description: `grants_repo_explicit` is `grant == "repo" || grant.starts_with("repo.")`, so `repo.read` also trips the gate. For a *refusal* that is the conservative direction, and the threat-model table's claim that `*` does not confer `repo` matches the code.

Every production construction site passes the real kind (`bin/opencompany.rs:290`, `:1149`, `server/provision.rs:258` → `RuntimeBuilder`, then `builder.rs:1637` → `RepoManager`). The remaining `RepoManager::new` call sites are all under `#[cfg(test)]`.

## One finding: the breaking-change paragraph understates who is affected

`docs/spec/runtime/storage.md`:

> **This is a breaking change for `fs` and `sqlite` deployments that have already bound a repository.**

The boot refusal keys on **grants, not bindings**. `builder.rs` computes `roster_holds_repo` from the company allow-list and each manifest/overlay agent's effective grants, and returns `Config` on that alone — no binding is consulted anywhere in that branch. Your own test is the proof: `a_repo_granted_company_does_not_boot_on_a_plaintext_secret_backend` asserts `!grants_repo_explicit(&manifest.tools.allow)` and still fails the boot on the *agent's* grant, with no repository bound at all.

Concrete failure: a self-hoster on the default `fs` backend with `allow = ["repo"]` (or one agent naming `tools = ["repo"]`) and **zero bindings** reads that sentence, concludes they are unaffected because they never bound anything, upgrades, and the company does not come up. The table three lines above states it correctly — "when its **effective roster** grants `repo`" — so the prose and the table currently disagree, and the prose is the part an operator skims for "does this hit me".

Suggest the paragraph match the table: breaking for `fs`/`sqlite` deployments **whose effective roster grants `repo`**, with the already-bound case called out as the subset that must also revoke at the forge. The remedy sentence is fine as-is.

That is a documentation accuracy point, not a code defect — I think refusing on the grant rather than the binding is the right call, since a granted-but-unbound company would only hit the `409` later anyway, and refusing at boot is the more honest moment.

## Not covered

I did not exercise this at runtime (no build), and I did not review the 252-line threat model line by line for claims outside the storage/grant surface — I checked the claims it makes about `grants_repo_explicit`, the fs/sqlite/mongodb split, and the three gate locations, all of which match the code.

Issue #681, opened 2026-08-11
workflows: an undeclared slug parks for approval instead of routing as an error (#338 fail-closed vs the runner's error path)


M3gA-Mind:
## The defect

#338's judgement gate (PR #660) **fails closed on an undeclared tool**: a slug nobody has declared stops for a human rather than passing. That is the right instinct for an agent turn — nobody can say what the tool does, so a person should look.

It is **wrong for a workflow's error path**, where an unknown slug is not a mystery to be adjudicated but a failure to be routed.

## What breaks

```
workflows::runner::tests::t2_unknown_slug_retries_then_continues_with_error_item
workflows::runner::tests::t3_on_error_route_sends_failure_down_the_error_edge
```

Both use the slug **`bogus_tool`**, deliberately: it cannot execute, which is the whole point. They exercise **error routing** — retry-then-continue-with-an-error-item, and the on-error edge — and have nothing to do with approvals.

With the gate in place, `bogus_tool` is undeclared, so it parks for a human *before* the runner can produce the error the test is about. "Unknown tool → error item" becomes "unknown tool → parks for a human".

## Why this is its own issue

It is **independent of #674**. That one is a contradiction between #338's unbounded-reach rule and #614's position on `tool_call` nodes. This is a different rule (`StopReason::Undeclared`) failing for a different reason (an unknown slug in a workflow is a routing case, not a consent case), and it would still be wrong whichever way #674 is decided.

## What a fix might look like

- **Resolve the slug first, park second.** If the runner already knows a slug does not resolve, that is a failure and should route as one — the approval gate never needs an opinion about a tool that cannot run. This looks closest to correct: it fixes the ordering rather than weakening the rule.
- **Scope fail-closed to the harness path**, leaving workflow nodes to their own resolution errors. Needs an argument for why an unknown slug is safe in one caller and not the other.
- **Keep the park.** Defensible if an unknown slug in a workflow is genuinely a thing an operator should see — but it changes what a failing workflow does, and the tests above say the current answer is "route the error".

## Related

- #338 / #660 — the gate that introduces the fail-closed rule
- #674 — the other, separate collision between #660 and #614/#627
- #558 — the approvals epic

M3gA-Mind:
## Not actionable yet — this describes a consequence of #660 landing, not a live defect. Parking.

A worker claimed this, applied the re-verify-before-building rule, and found it does not reproduce. Verified independently against `upstream/main` @ `a90f70f0`:

[...]

Issue #753, opened 2026-08-12
workflows: a create-time co-pilot — describe a workflow in the New-workflow dialog and get a grounded draft graph to review


oxoxDev:
## Summary

The console has two workflow-authoring aids today, with a gap between them. The **per-workflow copilot** (#574 / #416) refines an *already-saved* graph over chat on the `workflow-copilot:<id>` thread, emitting a `workflow-proposal` op list applied with a versioned `PUT`. The **task→workflow builder** (#580) turns a *task card + plan* into a whole graph server-side. Neither helps the operator who opens **New workflow** with only an intent in their head: they face a blank hand-form and must wire every node, edge, and config field by hand — the exact surface #661 showed is easy to get silently wrong (field-less conditions, unwired `tool_call`s, dead branches, all of which save clean and fail at runtime).

This asks for a **create-time co-pilot**. In the New-workflow dialog the operator types a plain-language description ("every weekday morning, triage new GitHub issues and post a summary to the eng channel"); the copilot proposes a complete draft graph grounded in the company roster, granted tools, and existing workflow names; the operator reviews it, it loads into the dialog's draft form, and they save through the normal create path. The host stays the authority — no new persist path, no new validation surface.

## Motivation

- Blank-form authoring is the highest-friction, highest-error entry point. Every silent-broken-save class in #661 starts with a human (or agent) hand-writing config nothing enforces.
- The hard pieces already exist: an LLM that emits a whole, validated graph (`WorkflowBuilder`, `src/harness/workflow_build.rs`), a *pure* client-side apply into an empty graph (`applyProposal(EMPTY_GRAPH, …)`, `frontend/src/lib/task-workflow-proposal.ts`), and the review UI. What is missing is a **request-shaped describe→graph entrypoint** and a create-dialog surface to drive it.

## Proposed approach

- **Backend** — a new small request endpoint (not the chat thread): `POST …/workflows/draft-from-description` (name TBD). Takes `{ description }`, assembles the same `Evidence` the builder already gathers (roster for `agent` nodes, granted tools for `tool_call`, existing workflow names for uniqueness), runs the builder's LLM + full-graph prompt, validates with `courtesy_validate_draft` (validate **without** persisting), and returns a `WorkflowGraphSpec` + a one-line summary — or `{ automatable: false, reason }` when the intent can't be a graph. Reuses `WorkflowBuilder` and `courtesy_validate_draft`; does not touch the create/persist core.
[...]

Issue #708, opened 2026-08-11
workflows: delete+recreate inherits the old schedule's fire ledger


oxoxDev:
## Problem

A saved workflow's cron trigger is keyed in the durable `ScheduleFireStore` on the workflow id alone — `workflow_schedule_id(workflow_id)` returns `"workflow-<workflow_id>"` (`src/runtime/workflow_scheduler.rs`). That id is intentionally restart-stable and index-independent.

The consequence: deleting a workflow and recreating another with the **same id** makes the new workflow inherit the deleted one's entire fire ledger (its `claim_fire` / `latest_fire` rows). `delete_workflow` → `delete_company_workflow` (`src/server/ops/workflows.rs:784`) tears down the graph body, revisions, and events, but never purges the schedule's claim rows.

Two bounded but real effects, each capped by `CATCHUP_WINDOW_MINUTES`:

- **One suppressed minute** — if the recreated workflow's schedule matches a minute the old one already claimed, the durable `claim_fire` returns `false` and that occurrence is silently skipped.
- **One inherited make-up** — the first-sight restart catch-up reads the stale anchor and may fire (or decline to fire) a make-up based on the deleted workflow's history rather than the new one's.

## Why the key was NOT re-keyed (deferred from #661 F5)

#661 F5 considered re-keying the schedule identity to make delete+recreate a clean slate and **rejected** it, for three reasons:

1. **Per-deploy catch-up loss** — re-keying orphans every deployed tenant's existing anchors, so the first boot after the change treats every armed schedule as a fresh install and forfeits its legitimate restart catch-up.
2. **Rolling-deploy double-fire (#241 regression)** — during a rolling deploy, mixed-version replicas would key the same workflow's minute under two different ids and both could win a claim → the exact cross-replica double-fire #241's durable claim exists to prevent.
3. The id is deliberately the natural `(company, workflow)` identity so it survives a restart without depending on a positional index.

## The real fix

Purge the schedule's fire-ledger rows at delete time instead of changing the key. Likely shape:

- Add a `ScheduleFireStore` method (e.g. `delete_fires(company, schedule_id)`) implemented across the fs / sqlite / mongodb backends.
- Call it from `delete_company_workflow` (`src/server/ops/workflows.rs`) with `workflow_schedule_id(wid)` so a deleted workflow leaves no anchor behind and a recreate is a genuine fresh install.

[...]

PR #776, opened 2026-08-12
feat(harness): read/update/delete workflow tools for agents (#661 M7)

oxoxDev:
## Summary

Closes **M7** of #661 — *no edit/delete-workflow tool for agents*. An agent could
author a workflow (`create_workflow`, #112) and run one (`run_workflow`, #67),
and nothing else: a graph it got wrong could only be replaced by a second graph
beside it, under a different id and a different name, forever. The company layer
has had the writes since #259/#274; only the tool surface was missing.

**Three tools, not the two the audit predicted.** `update` is a whole-graph
replacement and `query_company` lists ids and names only, so shipping the two
writes alone would make every edit a blind rewrite of a graph the agent has
never seen — a fresh instance of exactly the class this issue exists to close.
So:

| Tool | What it does | Reach |
|---|---|---|
| `read_workflow` | the graph in **exactly the JSON `update_workflow` accepts**, plus the `version` token, `editable`, `enabled`, and any operator-only fields | `Nothing` |
| `update_workflow` | full-graph replacement, **`expected_version` required** | `Nothing` |
| `delete_workflow` | remove the workflow and its revision history | `Consequence` |

The required token is what turns "read before you write" from advice into a
mechanical property: the read is the only place to get one, so an agent that
never read the graph has nothing to send, and a console edit landing in between
is the company layer's own 409 rather than a silent clobber.

## API Or Behavior Changes

**New agent tools** (orchestrator-only, registered straight after
`create_workflow` whose lifecycle they are). No REST, GraphQL or console API
changes.

**New refusals on the agent surface only.** An agent-authored graph is
manual-run only — `schedule`, `on_error`, `retry` and `requires_approval` are
unattended-run policy the authoring schema deliberately omits. Create can honour
that by never emitting them; a full-replacement *edit* cannot, because replaying
a graph through the narrow schema drops whatever the stored body carried. So
both writes refuse a scheduled target, and `update` also refuses a graph
carrying per-node run policy, naming the node and the field. Verified: with the
guard removed, the same update really does turn `requires_approval: Some(true)`
into `None` — that is removing an operator's approval gate, not forgetting a
field.

Two consequences of the schedule half, both deliberate:

* it **closes the agent-driven half of #708** outright (an agent cannot delete a
  scheduled workflow, and cannot recreate one *with* a schedule because the
  schema has no field for it), so no agent path reaches
  delete-then-recreate-inherits-stale-ledger. **#708 is not fixed here** — the
  console path is where its real fix belongs;
* the guard lives in the tools, not the company layer, so **console behaviour is
  untouched** and it is one line to lift per field if agent self-scheduling ever
  becomes a yes.

**Gating: delete gets one, create and update do not.**

`delete_workflow` is `Reach::Consequence` + `Standing::PerCall` — `workspace_delete`'s
(#671) argument, with a strictly worse object: `delete_company_workflow` cascades
the #274 revision history away in the same call, so unlike an update there is
nothing left to restore from. Per-call parking makes each removal its own card
naming its target.

`update_workflow` takes `create_workflow`'s classification, re-derived rather
than copied. Four properties separate it from `workspace_write`: every
content-changing update is **undoable by construction** (#274 snapshots the prior
body inside the write lock); after the scheduled-target refusal every editable
workflow is manual-run only, so a bad edit's consequence materialises solely
through `run_workflow`, which already parks; the required token means it cannot
clobber unseen state; and validation is identical to the console's. Parking every
fix-up save of the agent's own just-created draft would recreate the #558/#561
no-consequence-interrupt pattern against this feature's headline flow.

**Honest residual:** `Reach::Nothing` means a `readonly` desk can edit workflows.
But `create_workflow` is *already* `Nothing`, so such a desk can already create
them — this is a pre-existing classification decision, not a new hole. If it is
wrong, create and update are reclassified **together**; this PR does not smuggle
a unilateral change.

**Seeds are protected by inheritance, not by a new check.** `locate_editable_overlay`
is the single shared resolver behind both writes (seed-backed → 409,
bodiless-enabled → 409, unknown → 404, with the `is_safe_workflow_id` traversal
guard in front), so an agent can never touch a shipped seed regardless of what a
tool does. The Conflict message is passed through verbatim rather than forked —
`seed_file_exists`'s own doc warns that duplicated probes are how the three
surfaces drift. Its "from the console" wording reads slightly oddly in an agent's
mouth; that is wording debt on the shared string, noted rather than forked.

**Why there is no console change.** The tools write through the same company
functions the REST routes call and journal the same `WorkflowUpdated` /
`WorkflowDeleted` events, so the workflow list, the editor and the events feed
pick an agent's change up with nothing new wired; and the editor's own
`expected_version` 409 already protects an operator mid-edit from any concurrent
writer. The one frontend edit here is unrelated to the editor: a
`Reach::Consequence` tool with no entry in `language.ts` renders as "Use one of
its tools" on the approval card, so `delete_workflow` gets words there.

## Tests

Thirteen new tests in `src/harness/workflow_admin/tests.rs`, over the tool
surface only — the company layer is already covered in `workflow_create.rs`.

- create → read → update round trip: the graph read is the graph the writer
  accepts, one node edited and the rest intact, `enabled` untouched, journaled,
  prior body snapshotted
- `expected_version` required at args; a stale one is the company layer's 409
  with the concurrent edit intact underneath
- seed-backed id reads `editable: false` with no token and refuses both writes
- **a scheduled target is refused by the tools while `update_company_workflow`
  still accepts the same write** — the only test that can show the guard is
  agent-surface-only
- an edit that would drop a node's `requires_approval` is refused, naming node
  and field
- delete removes body + enabled id + history, journals, and names what went
- unknown / bodiless / corrupt-body ids; a corrupt body still returns a version
  so it stays deletable
- **#682 non-weakening pin**: an agent edit with a `field`-less `condition` or a
  slug-less `tool_call` is rejected with the console's own validation error
- policy: only the delete parks (under `supervised` and `auto`) and is denied
  under `readonly`; none of the three is grantable
- description routing pins (`read_workflow`, `expected_version`, permanence)

**Every new gate was proved by reverting the production change it guards and
observing RED**, then restoring: the required token, the scheduled refusal, the
policy-strip refusal (twice — once for the refusal, once to show the unguarded
update really strips the gate), the read projection's round-trip fidelity, the
no-revision-store refusal, the `delete_workflow` reach, and the #682 pin
(by disabling `validate_draft_against_record`'s per-kind branch).

Commands run locally:

- [x] `cargo clippy --no-deps --features openhuman,mcp,telegram --all-targets -- -D warnings`
- [x] `cargo test --locked` — 2240 passed, 0 failed, 1 ignored (+ 12 in the other targets)
- [x] `cargo test --features openhuman,mcp,telegram` — 3379 passed, 0 failed, 3 ignored (+ 24 in the other targets); no CI lane builds this combo
- [x] `cargo build --features openhuman,mcp,telegram`
- [x] frontend `npx tsc --noEmit` + `npm run build`

**Manual verification, and its limit.** A local host was booted from this branch
(`companies/e2e_harness`, console served, dev magic-link) and the full
create → read → update → delete lifecycle driven against it: the graph reads back
in the round-trippable shape with its token, an update with the token lands and
leaves `enabled` alone, a stale token 409s, the workflow list shows the seed as
`editable: false` and the runtime graph as editable, and the journal carries
`WorkflowCreated` / `WorkflowUpdated` / `WorkflowDeleted` in order.

**What I could not drive: the chat-driven pass through the tools themselves.**
This environment has no inference credential, so the brain answers
`"You said: …"` and no agent turn can be taken. The step verified above is the
company-layer + console half, which is the half the tools share; the tool-call
path itself is covered by the unit tests and by the belt-registration test, not
by a live turn. Stated rather than claimed.

## Documentation

- `docs/spec/runtime/ports-state.md` — who writes `overlay_workflows` now
- `docs/spec/runtime/events.md` — `WorkflowUpdated`/`WorkflowDeleted` reach both surfaces
- `docs/spec/runtime/manifest.md` — the trio is orchestrator-only authority
- `src/harness/workflow_admin.rs` module docs carry the design argument in full

## Related

Part of #661 — this closes finding **M7**. The tracker stays open for **M5**
(no in-workflow task-card create) and **L1** (`on_error: continue` reports run
success, upstream in tinyflows).

Related but deliberately not fixed here: #708 (delete+recreate inherits the old
schedule's fire ledger) — the agent-driven half is closed as a side effect of the
scheduled-target refusal; the console path is untouched.

Coordination: no textual overlap with #753 (create-time copilot — does not touch
`src/harness/orchestrator.rs`) or #771 (delegation queue — `orchestrator.rs`
hunks are all in the `DelegationQueue` region, ~1000 lines from anything here).


M3gA-Mind:
Reviewed the authority surface — whether these tools let an agent reach something it could not reach directly. **Not an approval.**

## What holds up

The classification block is the most careful part of this PR, and the residual it records rather than hides — *"`Reach::Nothing` means a `readonly` desk can edit a workflow… this is a pre-existing classification, not a new hole. If it is wrong, create and update are reclassified TOGETHER"* — is the right way to carry that. Splitting `delete_workflow` out as `Consequence` on the #671 argument (the revision history goes with it, so there is nothing to restore from) is well-reasoned, and `delete_workflow` is added to the parking-tools test list so the split stays a decision rather than a drift.

`expected_version`, `refuse_scheduled`, and validation shared with the console are each doing real work, and I checked `refuse_scheduled` applies on both update and delete.

I also checked the obvious inversion first and it is **not** here: `workflow_admin.rs` contains no grant/allow-list reads at all, so this is not a path to widening an agent's own `[tools]` grant the way `add_agent` was in #619. Good.

## The finding: this makes `CallPath::AuthoredWorkflowNode`'s premise false

`judgement.rs` documents that path as:

> /// An operator authored this node into a saved workflow, past the manifest
> /// grant and the authoring refusal.

and `apply_policy_gates` selects it via `for_authored_workflow_nodes()`, which makes `judge` **silent** for any non-templated node. The entire justification for that silence is that *a human saw the call before it was saved*.

After this PR an agent can author and edit those nodes. So a node written by a model is judged on the path reserved for nodes a person reviewed — and that is the one arm `for_authored_workflow_nodes` changes.

Concrete shape: an agent calls `update_workflow` on an existing manual-run workflow the operator wrote and trusts, changing a node's arguments. It then calls `run_workflow`, which parks. The operator sees an approval for running a workflow they recognise; the node then executes judged **silent**, where the identical call made directly by that agent would have gone down `CallPath::Agent` and been judged in full.

Your mitigations are real and I do not think this is a hole so much as a premise that has quietly expired:

* `run_workflow` parking is the load-bearing one — but it is **one approval for the whole graph**, and its card names the workflow, not the node that changed.
* #274 undo is after-the-fact; the run has already happened.
* `refuse_scheduled` closes the unattended path, which is the worst version, and I checked it covers update *and* delete.

**The increment over `create_workflow` is the part I would not wave through.** You are right that create is already `Reach::Nothing` — but an agent-authored *new* workflow is unfamiliar to the operator, who reads its card with fresh eyes. `update_workflow` lets the agent change one the operator already knows, and `expected_version` does not help, since the agent reads the current version immediately before writing. "Pre-existing classification" covers create; it does not quite cover editing something already trusted.

Two things that would settle it, either of which I would find sufficient:

1. **Update the `CallPath::AuthoredWorkflowNode` doc comment.** It is now inaccurate, and it is exactly the comment the next person will reason from when deciding whether some other tool may write nodes. It should say authored by an operator **or by an agent through the M7 tools**, and state why the silence still holds. I closed #681 against this enum today; the next reader deserves the current premise.
2. **Say whether the `run_workflow` card can name a workflow edited since the operator last saw it** — e.g. surfacing "modified by <agent> at <time>" on the card. If it can, this reduces to a UI nicety. If it cannot, the operator's one approval is being asked for a graph whose content may have changed under them.

I could not determine (2) from the diff and did not want to assert it: I did not find the operator-facing label for `run_workflow` in a form that shows what the card renders, and #706 suggests label coverage for `Consequence` tools is itself incomplete. You will know immediately whether the card can carry that.

## Not covered

I read `workflow_admin.rs` for the authority question and the classification block in full; I did **not** review the 998-line test file or the `orchestrator.rs` refactor line by line, and I did not compile or run anything.

PR #780, opened 2026-08-12
board: a card opened from the prompt box enters Planning itself

### LINKED ISSUE #576: spine: decide whether a new card enters Planning automatically, or waits for a person
## The question

A card created from the prompt box lands in `todo` (`src/server/ops/tasks.rs:333`, defaulting to `COLUMN_TODO`). The planning pass edge-fires on the transition **into** `planning` (`CompanyRuntime::upsert_task`, `docs/spec/runtime/planning.md`). Nothing moves a card between the two.

So the spine is automatic from Planning onward — plan, dispatch, run, settle, In Review or Done — and manual for exactly one step at the front: **someone has to drag the card into Planning.**

## Why this needs deciding rather than fixing

Epic #183 presents `prompt box → Todo → Planning → In Progress → … → Done` as the automatic spine, which reads as though a new card should self-promote. But the manual step is also the only place an operator sees the work before the company starts spending on it, and `docs/spec/runtime/planning.md` describes the trigger as "a card dragged into Planning" — deliberate phrasing, not an accident.

Both readings are defensible, and they lead to different products:

| If auto-promote | If manual |
|---|---|
| Prompt in, deliverable out, no clicks | Todo is a real queue you triage |
| A typo costs a planning call | Nothing runs until a person says so |
| Matches the epic's diagram | Matches the current spec text |

## What to decide

1. Does a card created from the prompt box enter Planning automatically?
2. If yes, is there a debounce, a per-company cap, or an opt-out for cards created by an agent rather than a person?
3. If no, the epic's diagram and this issue should say so plainly, so the next reader does not file this again.

## Acceptance

Whichever way it goes, `docs/spec/runtime/planning.md` and epic #183 agree with the code, and the board's To-do column has a documented meaning.

Part of the spine — epic #183.

M3gA-Mind:
## Summary

A card created from the prompt box landed in `todo`, and nothing moved it to `planning`. So the spine was automatic from Planning onward — plan, dispatch, run, settle, In Review, Done — and manual for exactly one step at the front: someone had to drag the first card.

An actionable message typed by **a person** now opens its card directly in `planning`. Epic #183's diagram is the intended product; the drag was the gap.

Closes tinyhumansai/opencompany#576

## The shape: created **directly** in `planning`, not created in `todo` and promoted

This is the decision a reviewer will ask about, and it is not recoverable from the diff.

The planning pass **edge-fires on the transition *into* `planning`**:

```rust
fn task_enters_planning(prev_column: Option<&str>, next_column: &str) -> bool {
    next_column == PLANNING && prev_column != Some(PLANNING)
}
```

A card that does not exist yet has **no previous column**, so `task_enters_planning(None, "planning")` is `true`: one `upsert_task` writing the card straight into Planning is one transition, and therefore exactly one pass. The alternative — create in `todo`, then promote — was rejected because:

- **Two writes, two board events, and a visible intermediate.** Between them the card is in To-do, where it can be seen, dragged, or dispatched by someone else.
- **The half-failure is worse.** If the single write fails, nothing exists. If the *second* of two writes fails, a card is stranded in To-do while the operator believes they asked the company to start it — the failure mode is silent and looks like success.
- **No new recovery path is needed.** `sweep_stranded_planning` already returns a Planning card whose pass died at boot, so a directly-created card is covered by machinery that already exists.

Both shapes fire exactly once, so "exactly once" alone does not choose between them — the atomicity and the failure mode do.

## Scope: the prompt box is `operator.rs`, not `tasks.rs`

The issue cites `src/server/ops/tasks.rs:333`. That is the REST `POST /tasks` handler, which serves the board's `+` button and chat's "Add to board". The **prompt box** — an actionable chat message triaged as `Track` — creates its card at `src/server/operator.rs`, which independently defaulted to `COLUMN_TODO`. That is the line this PR changes.

The other two callers are left inert deliberately: `create_task`'s own comment describes it as "manual entry — the board's `+` button", and `Conversation.tsx` omits the column so that "the human drag [stays] the only thing that spends money". Promoting there would buy a planning call for every card someone jots down. All four existing `todo` assertions are on that REST route and are untouched.

## Agent-created cards are explicitly out of scope

The promotion buys a planning pass, which is a model call. The branch is on **who asked**:

```rust
let opened_by_a_person = matches!(
    by.as_ref().map(|actor| actor.kind),
    Some(ActorKind::User | ActorKind::Operator)
);
```

`by` is `Some` only when a signed-in user is behind the request; a machine credential resolves to `None`. `Agent` and `System` are excluded **by being named**, not by falling through `None`, so a future caller that passes an agent actor is refused by this branch rather than by luck.

The reason, which is why this is a hard line rather than a default: an agent whose card self-promoted would trigger a planning pass, whose output can open further cards, which promote, which plan — **a spend loop with no human in it**. The issue's own framing — "a typo costs a planning call" — is about a *person's* mistake costing one call. **Widening this later is safe; narrowing it after a spend loop is not.**

## API Or Behavior Changes

- **An actionable prompt-box message from a signed-in person now opens its card in `planning` instead of `todo`**, and that entry buys one planning pass. On a build with no planner attached the column is inert, exactly as before.
- The same message from a machine credential, an agent, or the runtime still opens the card in `todo` — today's behaviour, unchanged.
- No API shape changes: no route, request body, or response field is altered.

## Documentation

- **`docs/spec/runtime/planning.md` updated.** Its opening line — "A card dragged into **Planning**…" — was the text the issue named as contradicting the epic, and it is now wrong; the document now names both routes into Planning (the prompt box opening the card there, and the drag from To-do), states that both are the same transition buying one pass, and records that a card opened by anything other than a person still lands in To-do, with the spend-loop reason.
- **The To-do column now has a documented meaning** (acceptance item 4). On `COLUMN_TODO`: it is **where work waits for a person to decide it should start** — hand-entered cards (`+` button, `POST …/tasks`), cards opened by something that is not a person, and **returned** work: a failed pass, a hard prerequisite missing, a rejected workflow proposal, a cancelled run. That last group is the largest in practice and is why the column still needs a clear meaning — a returned card is waiting for a person to read the reason and decide. The meaning is unchanged in kind (not-started work) and narrowed in membership.
- Epic #183 and the code now agree, which was this issue's acceptance criterion.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — exit 0, in both CI's default-feature form and `--no-deps --features openhuman,tinycortex`.
- [x] `cargo build --all-targets` — covered by `cargo check --locked --all-features --all-targets` (exit 0) and the test lanes below, which compile every target.
- [x] `cargo test` — exit 0. Default **2242 passed**; `--features openhuman,tinycortex` **3390 passed, 0 failed, 3 ignored**.

### The pass fires exactly once — counted, not assumed

A test asserting "it planned" passes just as happily when it planned twice, so `a_prompt_box_card_buys_exactly_one_planning_pass` walks the card's real lifecycle and **counts the fires**:

```
(None, "planning")               ← the prompt box opens it
(Some("planning"), "planning")   ← the pass writes its plan back, still in-column
(Some("planning"), "in_progress")← the success settle
```

The middle step is the real double-fire risk: `harness/planning.rs` re-upserts the card *while it is still in Planning*. If resting counted as entering, that write-back would start a second pass, which would write back, billing a call each time.

**Four revert-checks, each leaving the rest of the suite passing:**

| Mutation | What failed |
| --- | --- |
| No promotion at all | both operator tests (78 passed / 2 failed) |
| Promote for **every** actor | `only_a_person_gets_a_self_promoting_card` — `an agent: the card must land in \`todo\`` (79 / 1) |
| Entry requires a previous column (**zero-fire**) | `a_prompt_box_card_buys_exactly_one_planning_pass` (8 / 2) |
| Resting counts as entering (**double-fire**) | the same test (8 / 2) |

The last two are the pair that matters: the edge-fire semantics make both zero and two plausible outcomes of a naive implementation, and only a counting assertion distinguishes them.

`only_a_person_gets_a_self_promoting_card` drives `run_chat` directly rather than the HTTP route, because the route's job is to *resolve* the actor and the test's job is to pin what each resolved actor does — going through HTTP would only ever exercise whichever principal the harness happens to authenticate as.


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Creating directly in `planning` rather than creating in `todo` and promoting is the right call, and the argument that settles it is the failure asymmetry:

> If the single write fails, nothing exists. If the *second* of two writes fails, a card is stranded in To-do while the operator believes they asked the company to start it — the failure mode is silent and looks like success.

That is the correct axis to decide on. The visible-intermediate problem is real too — between two writes the card sits in To-do where someone else can drag or dispatch it — but the half-failure is the one that would have been discovered in production rather than in review.

The edge-fire holds up, and I checked it rather than reading it:

```rust
next_column == PLANNING && prev_column != Some(PLANNING)
```

`prev_column` is derived from an actual load (`runtime.rs:613`), not supplied by the caller, so `None` genuinely means "no such card yet" rather than "the caller did not know". That is what makes `task_enters_planning(None, "planning")` sound rather than convenient — a caller-passed `None` on an existing planning card would have double-fired, and this shape cannot express that. All three cases are pinned, including the one that matters most: resting in the column does not re-fire.

Reusing the shape `task_enters_in_progress` already had, one line above, means this follows the file's existing idiom instead of inventing a second way to express the same idea.

### Question — load-then-write, and the pass is now expensive

`prev_column` is read, then the upsert writes, and I could not see a lock spanning the two. For two *different* cards that is fine — distinct ids, distinct rows. The case I would want ruled out is the same id arriving twice: a client retry, or a double submit from the prompt box, where both requests load `None` and both conclude they are the transition into Planning.

Before this change the cost of that was a duplicate write of the same row. After it, it is **two planning passes for one card** — real model work, and plausibly two plans racing to attach to the same task.

If `upsert_task` already runs under `company_write_lock` — as the team writes do — this is answered and worth a sentence saying so at `task_enters_planning`, since the guard's correctness then rests on the lock rather than on the comparison alone. If it does not, the natural fix is the same one used elsewhere in this repo: serialise the load-and-write for a company, so the transition is decided under the lock that performs it.

Not blocking: the window is narrow and the duplicate is wasteful rather than destructive. But this PR is what turns that window from cheap to expensive, so it is the right moment to know the answer.

**Before merging:** 15 commits behind, and it merges cleanly against #680, which touches the neighbouring question of *which* messages become cards at all. Worth landing in either order — they do not collide — but their combined behaviour ("what becomes a card" plus "where a card lands") is the thing to sanity-check on `main` once both are in.

oxoxDev:
Merged current `main` and fixed a regression this branch introduced (73594ec4, 9381362d). Both lanes green now.

**`chat-to-card.spec.ts` was failing under the live brain, and it was real — not a flake.** I ran it on this branch and on `main` with the same binary and the same fixture: fails here, passes there. The live-brain lane drives a scripted mock rather than a model, so there is no latency to blame it on.

**The mechanism.** `chat_handler_card` in `runtime/delegation.rs` recognises the REST handler's card by its properties, and one clause was `column == COLUMN_TODO`. This branch lands a person's prompt-box card in Planning, so the commonest of the two stopped matching — `spawned_task` fell back to `None`, the operator bubble reported no card, and the "Card opened" chip disappeared. Nothing failed loudly: the card was still created, still planned, still correct on the board. Only the link between the reply and the card went missing, and the console renders that only when the reply carries a task id.

Fixed by naming both landing columns. Kept the clause rather than dropping it — a card resting anywhere else was moved there by somebody, and is no longer the untouched write this seam may adopt. Two unit tests pin both directions (Planning is adopted, moved-on is refused), because the E2E that caught this runs only in CI while the rule lives in `delegation.rs`.

**Second push** was my own doing: main grew `origin_run_id` / `origin_workflow_id` on `TaskRecord` in #791 after I merged, and a struct literal has to name every field, so the fixtures I had just added stopped compiling. `None` on both — the card is a person's prompt, not a workflow run's, which is the state the rule is about.

Ran: `cargo build --locked --features openhuman,tinycortex --all-targets` · `cargo clippy --locked --features openhuman,tinycortex --all-targets --no-deps -- -D warnings` · `cargo fmt --all --check` · `cargo test --features openhuman,tinycortex --lib runtime::delegation` (55 passed) and `--lib runtime::cycle` (68 passed) · `PW_LIVE_BRAIN=1 playwright test chat-to-card board-columns` (5 passed, having reproduced the failure first).

Issue #770, opened 2026-08-12
ci: every merge cancels the previous merge's verification, so main is never actually verified


oxoxDev:
## Summary

`ci.yml` declares `cancel-in-progress: true` on a concurrency group keyed by workflow and ref. On a PR branch that is exactly right — a new push should kill the stale run. On `main` it means **every merge cancels the previous merge's verification**, so when merges land faster than the slowest lane takes, `main` is never actually verified.

## What I observed

Merges today landed faster than the ~26-minute `Rust (openhuman, tinycortex)` lane. The post-merge runs for #730, #737 and #756 were all `cancelled` before finishing. Walking back through `gh run list --branch main`, the most recent run that reached any conclusion at all was #744 (success); before that, #739 (failure).

So for a stretch of the day, the honest state of `main` was: not known-broken, and not known-good either. Nothing was red — the signal simply was not there.

## Why it matters

This is the same defect shape as #592, where the same submodule-init block is copied into several lanes and a missed copy fails exactly one of them: a check that quietly is not doing what everyone assumes it is. Here it is the *post-merge* check, which is the one thing standing between a bad merge and everyone else's branches.

It also interacts badly with a busy queue. The more PRs land, the less likely any post-merge run survives — so verification thins out precisely when the merge rate makes it most valuable.

## Suggested fix

Scope the cancellation so it applies to PR refs but not the default branch. The usual shape:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

Every merge commit on `main` then gets its own run to completion, while PR branches keep the current cancel-the-stale-run behaviour.

Worth checking the same expression in the other workflows while touching this — `release.yml` and `deploy-staging.yml` should be reviewed for the same assumption rather than fixed blind.

## Trade-off, stated

Letting main's runs finish costs runner minutes: several merges in quick succession means several full runs instead of one. That is the point — the cost buys the only proof that main compiles and passes. If the minutes turn out to matter more than the signal, the alternative is a scheduled run on `main` (hourly, say), which is cheaper but tells you later and does not attribute the breakage to a merge.

## Related

- #592 — same defect shape (a lane quietly not doing what the others do)

Issue #759, opened 2026-08-12
workspace: two concurrent publishes can both create the same agent/task folder, breaking every deliverable beneath it


M3gA-Mind:
## The gap

`resolve_folder` in `src/company/artifact_mirror.rs` creates a folder unconditionally when the name is free:

```rust
[] => {
    let id = create_folder(workspace, company, name, Some(parent.to_string()), origin(agent_id)).await?;
    ...
}
```

`children_named` returning `[]` is a statement about the instant it read the tree. Two concurrent publishes that both need `Agents/<agent>/<task>/` both see `[]` and both create. Two folders, one name.

This is the **same defect #697 fixed for files**, in the arm directly above it. #697 closed only the file arm:

```rust
// resolve_file — fixed by #697
None => create_first(workspace, company, &parent, filename, target).await,
```

## Why it does not decay

`resolve_folder`'s own fallthrough refuses a duplicated name permanently:

```rust
many => Err(OpenCompanyError::Conflict(format!(
    "{count} nodes under this folder are named `{name}`, so the path is ambiguous",
)))
```

So once two publishers race, **every future publish beneath that folder is refused, for every agent**, until somebody edits the tree by hand. A race lasting microseconds leaves a permanently broken subtree — and because agent/task folders sit at the top of every publish path, the blast radius is larger than the file case: it breaks *all* deliverables under that task, not one.

## Why it is reachable

The same reasoning #683 and #697 both accepted: workflow runs are `tokio::spawn`ed rather than serialized behind the cycle lock, and `materialize` is reached from the publish drain rather than a `company_write_lock`ed write plane.

## How it was found

While writing the barrier test for #697, `two_first_publishes_of_one_path_have_one_winner_and_no_duplicate` initially **passed with both publishes succeeding**. The cause was this bug: `ensure_agent_folder` / `resolve_folder` run *before* file resolution, so each racer minted its **own** parent folder and the two `report.md` nodes never contended for one path.

That test now seeds the shared folders first, with a comment saying why, specifically so the seeding is not read as incidental — see `src/company/artifact_mirror.rs`. **That seeding is a workaround for this issue, and should be revisited when this is fixed.**

## The fix already exists one arm down

[...]

PR #808, opened 2026-08-12
feat(workspace): fold duplicate folders left by a publish race (#759)

oxoxDev:
## Summary

Recovery half of #759. A publish race can leave two sibling folders sharing a name; once
that happens `resolve_folder`'s `many =>` arm answers `Conflict("the path is ambiguous")`
for **every** future publish beneath that path, for **every** agent, permanently. Stopping
new races does nothing for a tree already broken by one — this is the other half of the
answer: an operator-triggered pass that folds duplicate sibling folders back together, and
says plainly what it could not fold.

Semantics, all conservative by construction:

- Duplicate set = ≥2 sibling **folders** sharing `(parent, name)`.
- Oldest `updated_at_millis` wins; node id breaks the tie.
- Losers' children `rename_move` into the winner. **Node ids are preserved**, so artifact
  chains survive the fold.
- Folder–folder child collisions iterate to a fixpoint via a worklist, not recursion — a
  corrupt tree must not blow the stack.
- **Any collision involving a file is left in place and reported as a residual.** Merging
  files is where data gets lost, so the irreducible remainder is detected and named rather
  than resolved. Three causes: `fileSharesTheName`, `fileInTheWay`, `treeMovedOn`.
- A loser is deleted only when structurally empty **per a fresh read taken after the
  moves**, never per the snapshot it was planned from (#671/#700 discipline). The port's
  `delete` is recursive, so that re-read is the difference between reclaiming an empty
  folder and destroying a note that landed in it mid-run. The count is decremented as this
  run's own deletes land, so a nested duplicate and its parent can both go in one pass.
- Nothing is renamed, nothing non-empty is deleted, nothing is overwritten. A publish
  landing mid-merge degrades to "not merged this run", never to loss.

Operator-triggered, never automatic — the #570/#645/#700 opt-in doctrine.

Two pathological shapes converge over runs rather than in one click: a duplicate that
itself holds duplicates folds one layer per run, and a name duplicated 3+ ways inside a
duplicate resolves in reach order. Loss-free at every step; documented in the module docs.

**Merge order**: the race-fix half of #759 (`fix/759-folder-publish-race`) should land
first. The only overlap is one additive `pub mod` line in `src/company/mod.rs`; this branch
rebases after.

## API Or Behavior Changes

- **New route** — `POST /api/v1/company/workspace/merge-duplicate-folders?dry_run=<bool>`.
  `dry_run=true` previews; without it the pass applies. Both answers carry `residuals`
  **always**, including as `[]` — that field is the half of the answer that says whether
  the tree is actually fixed.
- **New console action** in the workspace toolbar beside the existing sweep, same
  preview → confirm flow. The dialog renders folds *and* residuals, each with its own
  plain-English instruction, so an operator knows what is left to settle by hand.
- No existing route, response shape, or workspace behaviour changes. Nothing runs
  automatically.

## Tests

`cargo test --locked` → **2312 passed, 0 failed, 1 ignored** (13 new: 12 unit + 1 route).
Frontend `npm ci` / `npm run typecheck` / `npx vitest run` → **585 passed / 47 files**
(9 new) / `npm run build` ok. `cargo fmt --all -- --check` clean;
`cargo clippy --locked --all-targets -- -D warnings` clean;
`cargo clippy --locked --no-deps --features openhuman,mcp,telegram --all-targets -- -D warnings`
clean. Counts read rather than exit codes — a filter that selects nothing exits 0.
This repo has no frontend lint script; none was run.

Teeth proved by mutating the source and observing the failure, not asserted:

| Mutation | Observed |
|---|---|
| newest wins instead of oldest | 8 tests fail |
| merge a file collision anyway | `left: ["theirs", "safe"]` — both documents collapsed onto one path |
| drop the fresh-read emptiness check | `the duplicate gained a deliverable and must survive` — the racing note is recursively deleted |
| no fixpoint | nested layer left unmerged, reported as a bogus residual |
| merge a group a file shares | `nothing may move while a file shares the name` |
| route answers in the wrong field | `a preview answers with a list of folds, got {"merged":…}` |
| residuals dropped | `and told exactly which document is still theirs to settle` |
| console `applyRepair` deletes before relocating | relocated notes vanish from the tab's tree |

**Intentionally untested against a real backend, and why.** The `fs` backend cannot hold
the state this repairs: `reject_path_collision` (#665) refuses a duplicate sibling name at
`create` *and* at `rename_move`, so the raced tree is unreachable through `FsOps`. Only
sqlite and mongodb accept it — what hosted tenants run — and both are non-default features,
so a feature-gated test would land in a lane that mostly does not run (#770). The route test
therefore injects a `#[cfg(test)]` permissive in-memory `WorkspaceStore` modelled on those
backends, via `RuntimeBuilder::with_workspace`. It is the one place these tests do not run
against a real backend and is worth a reviewer's eye.

A side effect worth stating: because the merge only ever moves into a name that is already
free, the algorithm is compatible with the fs collision rule rather than fighting it.

## Documentation

`docs/spec/runtime/api-write-plane.md` (the route, its query parameter, and the residual
causes) and `docs/modules/server/README.md`. The convergence-over-runs behaviour is
documented on the module itself.

`RepairPlan::is_empty()` is public API with no in-crate caller — the console does its own
check. Left in deliberately as part of the returned type's surface; happy to drop it.

Part of #759


oxoxDev:
Stacking note: this is the recovery half of #759. The race fix is #810 and should land first — it stops new duplicates; this one folds duplicates a race has already left behind. Rebasing this branch after #810 merges.

The `src/company/mod.rs` overlap we expected between the two turned out not to exist: `create_folder` was `pub(crate)` and never re-exported, so #810 never opened that file.

Issue #512, opened 2026-08-07
auth: hosted console Google/GitHub sign-in cannot complete — redirect target rejected on every tenant


oxoxDev:
## Summary

The hosted console's Google and GitHub sign-in buttons cannot complete sign-in on
any tenant. The console passes an `https` tenant origin as the redirect target;
the platform backend's redirect gate accepts loopback origins only and answers
400 before the provider handshake starts.

The backend-side change is filed as `tinyhumansai/backend#1243`. This issue tracks
the OpenCompany side: verifying the flow end to end once the gate accepts
provisioned tenant origins, and deciding what the console does in the meantime.

## Evidence

Three hosted tenants, both hubs, same result:

| Tenant | Hub | Sign-in button |
|---|---|---|
| `smoke1` | staging | 400 |
| `tinyhumans-marketing` | staging | 400 |
| `marketing` | production | 400 |

The gate rejects any non-loopback `https` redirect target regardless of origin or
hub, so this is not tenant- or environment-specific. The prod→staging hub switch
is orthogonal — both hubs reject identically.

The console-side redirect handoff landed 2026-08-04; the backend gate predates it.
The two have never been compatible, so this path has not worked since the handoff
shipped. Successful sign-ins before now went via email magic link or a build
predating it.

## Decision taken

Fix at the backend by accepting provisioned tenant origins, matched exactly
against the tenant registry — not a suffix pattern on the tenant domain. One
change covers every tenant and keeps the console handoff design already built.
Rationale and acceptance criteria live in `tinyhumansai/backend#1243`.

The alternative considered and not taken: drop the redirect target from the
console and use the no-redirect `redirect=dashboard` handoff the backend already
accepts. It needs no backend change, but it discards the per-tenant return target
the handoff exists to carry.

## Work here

- [ ] Confirm what the console sends as its redirect target, and that it matches
      the tenant's provisioned origin exactly (no trailing-slash or port drift)
- [ ] Once backend#1243 lands, verify Google and GitHub sign-in end to end on a
      live hosted tenant on both hubs — not against a stub
- [ ] Decide whether the buttons should be hidden or show an explanatory state
      while the backend gate is outstanding, rather than failing with a bare 400
- [ ] Confirm the session minted on return is a human session, not the machine
      bearer (see #315 — that split must not collapse)

## Note on #315

#315's "what already exists" section states the console ships magic-link and
password sign-in "with no third-party provider anywhere". That is now stale — the
[...]

PR #818, opened 2026-08-13
auth: hide ecosystem sign-in where the hub will refuse the redirect target

CodeGhost21:
## Summary

The hosted console's Google/GitHub/X buttons cannot complete sign-in on any tenant. The console hands the hub an `https://<slug>.<domain>/?company=<id>` return target; the hub's redirect gate accepts RFC 8252 loopback origins only and answers `400` before the provider handshake starts.

There is no fix to make on this side — the console derives that origin from `OPENCOMPANY_PUBLIC_URL`, so hosted is already a pure configuration path and `hub_identity.rs`'s "nothing here needs to change when it does" holds. The fix is `tinyhumansai/backend#1243`.

What this PR does is the second half of #512's checklist: stop offering a button that cannot complete, and pin the string the gate has to accept.

## What changed

- **`hub_accepts_redirect_uri`** (`src/server/hub_identity.rs`) — a mirror of the backend's `isLoopbackHttpUri`. `hub_providers` returns `{"providers": []}` when the redirect target would be refused.

  This is not new UI. It is the judgement the route already makes one line above for a host with no exchange, whose comment reads *"the difference between a console that says 'sign in with a link' and one that sends someone through Google to be turned away on return."* An empty list is a state the console already handles by rendering the magic-link form alone — which works on a hosted tenant today.

  Deliberately no configuration knob: a switch to disable this would outlive the condition it exists for. One function, one call site, both doc-commented with backend#1243 as the removal condition, and the route test goes red when the gate lands so it cannot silently stay off.

- **Pinned the hosted start URL** — the existing test asserted only the loopback form. The hosted `https` shape is what backend#1243 must accept, and it is **not a bare origin**: the `?company=` rides along, and in shared-single-DB mode the id is namespaced `<tenant>--<id>`. A gate matching the whole string against a registry of origins would reproduce this failure exactly.

- **Docs** — `docs/spec/runtime/users.md` had no entry for the hub routes at all. Added them to the routes table plus an "Ecosystem sign-in" section covering both empty-list cases and the not-a-bare-origin trap.

## Behaviour change

| Host | Before | After |
|---|---|---|
| Local (`http://127.0.0.1:<port>`) | 3 buttons, working | unchanged |
| Self-hosted, no hub | no buttons | unchanged |
| Hosted (`OPENCOMPANY_PUBLIC_URL=https://…`) | 3 buttons → raw JSON `400` on click | magic-link form alone |

## Verified

`cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` — 2311 passed, 0 failed. 6 new tests; the 10 existing hub tests are untouched and still pass.

End to end against two real hosts from the same binary and console build, differing only in `OPENCOMPANY_PUBLIC_URL`, driven in a real browser:

- local `:8091` → 3 buttons, hrefs carry the percent-encoded loopback target
- hosted `:8092` → 0 buttons, no orphaned divider or Terms line (they live in the same block)
- the RCA itself against production `api.tinyhumans.ai`: loopback target → `302` to Google; `https://smoke1.tinyhumans.ai/?company=smoke1` → `400 redirectUri must be an http loopback URI`
- the fallback actually completes: with SMTP wired on the hosted host, requested a link in the browser, read the code from the mail the host sent, followed it, and got `oc_session_agentic-marketing-agency` (httpOnly)

Note for anyone reproducing: the buttons need `--features tinyhumans` and mail needs `--features smtp`; and setting `OPENCOMPANY_PUBLIC_URL` makes `is_local_only()` false, so the host correctly refuses to echo `dev_code`.

## Not done here

Verifying a *successful* hosted OAuth round trip is genuinely blocked on backend#1243 and deliberately not stubbed. I have posted two further traps on that issue: the callback prefers `targetUri` over a non-loopback payload URI (`callbackHandler.ts:319-324`), and `buildLoginRedirectUrl` synthesizes a path that loses the `?company=`. Fixing the start gate alone is not sufficient.

Refs #512, `tinyhumansai/backend#1243`.

## Summary by CodeRabbit

- **New Features**
  - Added ecosystem sign-in documentation covering provider discovery, hub token sign-in, identity verification, and tenant-specific callbacks.
  - Added validation for sign-in redirect targets, accepting only safe local loopback addresses.
- **Bug Fixes**
  - Hosted consoles now show no sign-in providers when their redirect target cannot be supported.
  - Improved rejection of malformed, deceptive, or non-local redirect URLs.

### DIFF
--- a/docs/spec/runtime/users.md
+++ b/docs/spec/runtime/users.md
@@ -155,6 +155,8 @@ addressing forms work: `/api/v1/companies/{id}/…` and `/api/v1/company/…`.
 | `POST …/auth/request` | Mail a magic link. Always `{"sent": true}` |
 | `POST …/auth/verify` | Redeem a link → session cookie |
 | `POST …/auth/login` | Email + password → session cookie |
+| `GET …/auth/hub` | The ecosystem sign-in buttons this host can offer. `{"providers": []}` when it can offer none |
+| `POST …/auth/hub` | A platform token from the hub → session cookie |
 | `POST …/auth/password` | Set/replace your own password (needs a session) |
 | `GET …/auth/me` | Who this session belongs to |
 | `POST …/auth/logout` | Revoke this session |
@@ -180,6 +182,42 @@ the org chart, and every answer is a phishing target. It is also why
 — response *time* would otherwise answer what the body refuses to.
 
 Clients must not undo this. The console renders one vague message.
+
+## Ecosystem sign-in
+
+A host wired to the TinyHumans hub can also offer Google, GitHub and X. The
+browser goes to the hub's OAuth start pointed back at this console; the hub
+returns a platform token in the URL; `POST …/auth/hub` asks the hub whose it is
+and then applies **this company's own roster** — the same
+`eligibility` → `upsert_from_eligibility` → `mint_session` path a magic link
+takes. The hub says who they are; it never says whether they may in, and the
+session minted is an ordinary human session, never the hosting layer's machine
+credential.
+
+`GET …/auth/hub` answers `{"providers": []}` — not a 404 — whenever this host
+cannot complete the flow, so the console has one code path and falls through to
+the magic-link form. Two things produce an empty list:
+
+- **No hub.** A self-hosted host has no exchange, so it could not check a token
+  that came back. Three buttons that send someone through Google to be turned
+  away on return are worse than none.
+- **A redirect target the hub will refuse.** The return URL is
+  `{OPENCOMPANY_PUBLIC_URL}/?company={company_id}`, and the hub's redirect gate
+  currently accepts RFC 8252 **loopback** origins only. A local console on
+  `http://127.0.0.1:<port>` passes; every hosted `https://<slug>.<domain>`
+  origin is answered `400` before the provider handshake begins.
+
+The second is issue #512 and is a hosted-only gap: sign-in there is by magic
+link until `tinyhumansai/backend#1243` teaches that gate to accept provisioned
+tenant origins. Nothing here needs to change when it does — the origin comes
+from `OPENCOMPANY_PUBLIC_URL` — beyond deleting `hub_accepts_redirect_uri` and
+its one call site.
+
+Note the shape the gate has to accept: **not** a bare origin. The `?company=`
+rides along, and in shared-single-DB mode the id is namespaced `<tenant>--<id>`,
+so it varies per tenant and over time. Only the origin component is stable, and
+a gate matching the whole string against a registry of origins would reproduce
+this failure exactly.
 
 ## Passwords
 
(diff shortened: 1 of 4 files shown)


oxoxDev (approved):
Establishing that **there is no fix to make on this side** before writing any code is the right order, and it changes what the PR should be: the console derives its origin from `OPENCOMPANY_PUBLIC_URL`, so hosted is already a pure configuration path and `hub_identity.rs`'s "nothing here needs to change when it does" holds. Stopping short of a workaround that pretends otherwise is the discipline this repo has needed all week.

Reusing the judgement the route already makes one line above — the host-with-no-exchange case, with its comment about *"the difference between a console that says 'sign in with a link' and one that sends someone through Google to be turned away on return"* — means this is not new behaviour, just the same rule applied to a second reason the same thing is true. And an empty provider list is a state the console already renders correctly, so nothing downstream learns a new shape.

**The self-deleting design is the best part.** No configuration knob, because "a switch to disable this would outlive the condition it exists for"; both sites doc-commented with backend#1243 as the removal condition; and the route test goes red **when the gate lands**, so the workaround cannot quietly outlive its cause. That is the correct answer to the usual failure mode of a temporary suppression, and it is rare to see it built in on the first pass.

Pinning the hosted start URL is worth as much as the gate. Naming that it is **not a bare origin** — the `?company=` rides along, and in shared-single-DB mode the id is namespaced `<tenant>--<id>` — gives backend#1243 the actual string to accept, and "a gate matching the whole string against a registry of origins would reproduce this failure exactly" is the sentence that stops the fix being written wrong.

The predicate itself holds up against the ways these break: `http` only via `eq_ignore_ascii_case`, authority taken before path/query/fragment, userinfo dropped with `rsplit('@')` so `http://127.0.0.1@evil.com/` reads as `evil.com`, and IPv6 brackets stripped before the port split.

**0 major. 1 note. Approving.**

### Note — this repo now has two loopback predicates that disagree, and the difference is deliberate

`proxy::may_carry_a_credential`'s helper (from #747) decides loopback by parsing an `IpAddr` and calling `.to_canonical().is_loopback()`, so it admits `127.0.0.2` and `::ffff:127.0.0.1`. This one matches exactly three literals: `127.0.0.1`, `localhost`, `::1`.

That divergence is correct here — this function is a **mirror of an external contract**, so it must track the backend's `isLoopbackHttpUri` rather than this repo's own view of what loopback means, and being stricter than the backend fails safe (a button hidden that would have worked, falling back to the magic-link form that does).

What I would add is a line saying so at `hub_accepts_redirect_uri`. Someone will eventually notice the two and try to harmonise them — that is exactly the kind of tidy-up that looks like a cleanup and silently breaks a mirror. The doc comment already names backend#1243 as the removal condition; naming "do not align this with the proxy's loopback check, it tracks the hub" costs one sentence and answers the question before it is asked.

**Before merging:** 37 commits behind, `MERGEABLE` / `CLEAN`. Worth a rebase given how much has landed, but nothing here looks exposed to it.

Issue #698, opened 2026-08-11
feat: meter MCP tool calls, so per-connection usage is a real number rather than an honest-looking zero


CodeGhost21:
## Problem

#404 asks the provider detail view to show what has actually gone through a connection, and is explicit about the failure mode to avoid: *"Do not invent a number. If per-connection usage is not recorded today, say so on screen and file what it would take, rather than rendering a plausible-looking zero."* This is that filing.

Auditing what is metered today, there is exactly **one** production site in the codebase that emits an `OauthCall` sample:

```
src/harness/composio.rs — ComposioExecuteTool::execute, on success only
```

It records `provider = <toolkit slug>`, which `src/metering/usage.rs` aggregates into `by_provider` and surfaces as `byProvider` / `oauthCalls` on the usage route. So for a **Composio** connection the detail view has a real number available today, and needs no new instrumentation.

For the other two connection systems it has nothing:

- **Remote MCP servers** — no tool call through an MCP server is metered anywhere. A company can route a large share of its real work through MCP and the usage view will attribute none of it. Note this is not only a #404 concern: `oauthCalls` and the `connections` KPI (which is `by_provider`'s row count) both under-report for any company whose agents work primarily through MCP.
- **The native OAuth catalog** — unmetered, though this is moot while #396 stands: the credential is written and read by nothing, so there is no call to meter. Worth stating explicitly so the gap is understood as *inert* rather than *missing*, and so it is revisited if #396's functional fix ever lands.

The consequence for #404 is that a detail view rendering a usage figure would be honest for one of three systems and quietly wrong for the other two — a zero that reads as "nothing has happened here" when it means "nothing has been counted here". Those are different sentences and only one of them is true.

## Proposal

Smallest useful change, in two independent parts.

**1. Meter MCP tool calls.** Emit an `OauthCall` sample per successful MCP tool invocation, mirroring the Composio path in `src/harness/composio.rs` exactly — on success only, zero-token, with the server name as `provider`. The existing aggregation, the `byProvider` chart, and the `connections` KPI then all work with no change downstream.

One decision to make deliberately rather than by default: whether an MCP server should share the `provider` namespace with Composio toolkits, or carry a prefix so `gmail` (Composio) and an MCP server also called `gmail` cannot collide into one row. A prefix is the safer default.

[...]

PR #819, opened 2026-08-13
feat(connections): open a provider and manage it, the way OpenHuman does (#404)

CodeGhost21:
## Summary

The console half of #404. #696 made a Composio connection an object on the wire; this is the surface that opens one.

A connected provider was a row with a button. It is now a thing you open: which accounts are connected, under which status, since when, what has gone through it, a per-account disconnect, and "connect another account". Ported from `vendor/openhuman`'s `ComposioConnectModal` — its information architecture and its wording — diverging in the three places OpenCompany's reality forces, each named below.

**It also fixes a disconnect that released nothing.** Every Disconnect on this page went to `POST …/connections/{provider}/disconnect`, which blanks the host's own `oauth/{provider}` secret. A Composio-connected provider never had one, so the call answered `200`, the toast said "Disconnected Gmail", and Gmail was connected again on the next refresh. `DELETE …/composio/connections/{id}` is the route that releases it, and `disconnectRouteFor` is the one rule that decides which route a tile takes — Composio when the toolkit holds a live account, native when only the host's own secret backs it, nothing at all on a host predating #696, where there is no id a revoke could name.

Composio wins when a provider is connected through both: the native secret is read by no agent (#396), so blanking it releases nothing an agent can feel while the tile stays connected.

**Second fix, found while building "connect another account":** the post-authorize poll waited for `connected`, which is already true when adding a second account to a connected toolkit. The first tick — two seconds in, with the Composio tab still open — announced a sign-in that had not happened. It now waits for an account id it had not seen before, which answers the first connect and the second with one rule.

### Three things the port would have inherited untruthfully

**1. No account is marked as the default.** OpenHuman marks the first of several, and inheriting that was the plan named on the issue. It is not true here: `composio_execute` posts `{tool, arguments}` and **no connection id** (`src/harness/composio.rs:1155`; `execute_tool` in the shared client builds the body), so Composio resolves which account acts and nothing in this codebase selects, orders or defaults one. `list_connections_detailed`'s `(toolkit, id)` sort never reaches the execute path — it orders a read. A "Default" chip would name a decision this product does not make, so the panel says so instead. @oxoxDev — this answers the question you left open on the issue, and it is narrower than a preference: making a default real means sending a connection id on execute, which is a harness change, not a console one, and worth its own issue if we want it.

**2. The connection date.** Only Composio records one. The native store keeps `{token, account}` and journals nothing on connect; MCP has no such concept. So the panel says "connection date not recorded" rather than leaving a blank that reads as "never" — and no timestamp column is added to the native store, per the issue thread.

**3. Usage.** Real for Composio, which meters a successful `composio_execute` per toolkit (#698 / #744), so the number is not invented. But it is counted **per provider, not per account**, and the panel states that: a call through either of two Gmail accounts lands on the same total. A host that cannot answer the usage route says so rather than rendering a zero, because a failed read is not a company that made no calls.

### Scope

A Composio-reachable tile opens whether or not it is connected — "connected **or not**" is the issue's wording, and OpenHuman's modal opens on a disconnected toolkit and connects from inside. Connect therefore moves into the panel and the tile costs one extra click, which is the trade being inherited rather than re-derived. The one provider that does not open is one Composio reports as connected on a host answering without `accounts` (predating #696): the panel would open on a connection it cannot name, date or release.

Composio only. The native catalog gets no detail view — its credential is written and read by nothing (#396), and a page about how healthy a connection is, is where an inert one would look most alive. A provider that also holds a native credential says so as a caveat, not as a second connection. Connect and disconnect are hidden from a member who cannot use them (#403).

**Left for follow-ups, deliberately.** A remote MCP server still cannot be opened — `McpServersSection.tsx` is its own 717-line surface and folding it in means two ports in one PR; the panel names MCP as a system without opening one. And the native catalog stays listed on the page, per the thread's "doesn't have to happen in this slice".

A tile holding two accounts shows neither label: one of two names reads as "this is the account it acts as", which is the claim nothing here can back.

## API Or Behavior Changes

No host change — this consumes routes #696 already shipped.

Console behavior:
- A connected Composio tile opens a detail panel instead of offering an inline Disconnect. Natively-connected tiles keep the inline control and get no panel.
- Disconnect now reaches Composio. It previously reported success and changed nothing there.
- The post-authorize poll's success condition changed from `connected` to "an account id that was not there before".
- `buildGridProviders` takes a sixth, optional `composioAccounts` argument; `GridProvider` gains `accounts`. `canDisconnect` keeps its name and changes meaning — it now covers the Composio route too.

## Tests

Console-only change; the Rust suite is untouched. Commands run locally, all green:

- [x] `npm run typecheck` (`tsc -b`)
- [x] `npm run typecheck:e2e`
- [x] `npm test` — 625 pass, 27 new
- [x] `npm run build`
- [x] `./scripts/ci/assert-design-tokens.sh`

New coverage:
- `test/unit/provider-grid.test.ts` — the routing rule, including the case that produced the bug (Composio-connected must not take the native route), both-connected preference, the pre-#696 host, an expired account, and the two-account tile showing no label.
- `test/unit/connection-detail.test.ts` — the absent date, an unparseable date, and that an MCP server's `mcp:<server>` row can never be read as the Composio toolkit's total (the collision #698's prefix exists to prevent).
- `test/unit/provider-detail-render.test.ts` — jsdom, rendering the real panel through a real grid row: no "Default" anywhere, the not-recorded date, the per-provider usage caveat, no zero on an unmetered host, the disconnect-scope wording, that a member is offered no control the host would refuse, and the disconnected-provider states (opens, offers the connect, hides a usage figure that would say nothing, explains a missing credential where the button is).

**Not verified:** the panel driven against a live host. It renders for a connected Composio provider, which needs a real connected account — no path to one locally. The render tests exercise the actual component rather than a stand-in, which is the closest this can get without that.

## Documentation

`docs/modules/server/README.md` — its connections section still said "No host route releases a Composio connection", which #696 made false. Replaced with the two routes, what each acts on, why they are not interchangeable, and the fact that no account is selected at execute time.

### DIFF
--- a/docs/modules/server/README.md
+++ b/docs/modules/server/README.md
@@ -365,9 +365,37 @@ Two consequences worth knowing:
   probe is a network call with no injection point, and the removed gate lived on
   the far side of it, which is how it survived unasserted.
 
-No host route releases a Composio connection — `…/connections/{provider}/disconnect`
-blanks this host's own secret — so the console offers Disconnect only for a row
-whose `via` includes `native`.
+### Releasing a connection: two routes, not interchangeable (issue #404)
+
+There are two disconnects and they act on different things:
+
+- `POST …/connections/{provider}/disconnect` blanks this host's own
+  `oauth/{provider}` secret and best-effort revokes it upstream. It has never
+  touched Composio.
+- `DELETE …/composio/connections/{connection_id}` revokes **one connected
+  account** at Composio. Addressed by connection id, because a company can hold
+  two accounts for one toolkit and exactly one of them is being released.
+
+Until the second route existed the console sent every Disconnect to the first,
+so a Composio-connected provider answered `200`, reported success, and was still
+connected on the next refresh. The console now routes by what the provider is
+actually connected through (`disconnectRouteFor` in
+`frontend/src/lib/provider-grid.ts`), preferring the Composio account when a
+provider is connected through both — the native secret is inert until #396
+lands, so blanking it would release nothing an agent can feel.
+
+`GET …/composio/connections` carries the accounts a revoke is addressed to:
+`accounts[]` per toolkit, each with Composio's verbatim `status`, its
+`createdAt` when Composio published one, and the account label when the provider
+did. `toolkit` and `connected` keep their previous meaning, so the callers that
+read only those two are untouched.
+
+**Which account an agent acts as is not decided here.** `composio_execute` posts
+`{tool, arguments}` and no connection id (`src/harness/composio.rs`), so
+Composio resolves it for the entity — nothing in this codebase selects, orders
+or defaults an account, and the console's provider detail view says so rather
+than marking one as the default. Changing that means sending a connection id on
+execute, which is a harness change, not a console one.
 
 ## tiny.place A2A inbound + discovery (`tinyplace` feature)
 
--- a/frontend/src/api/composio.ts
+++ b/frontend/src/api/composio.ts
@@ -133,12 +133,60 @@ export interface ComposioAuthorize {
   connectUrl: string;
 }
 
+/**
+ * One connected account inside a {@link ComposioConnection} (issue #404).
+ *
+ * A non-secret projection of the host's `ConnectedAccountDto` — the id is
+ * Composio's connection id, which is safe to hold here because it is the path
+ * segment `DELETE …/composio/connections/{id}` takes and carries no credential.
+ */
+export interface ComposioConnectedAccount {
+  /** Composio's connection id — what a disconnect is addressed to. */
+  id: string;
+  /**
+   * Composio's raw status string (`ACTIVE`, `INITIATED`, `EXPIRED`, …),
+   * forwarded verbatim so the console can tell "never set up" from "set up and
+   * since expired" rather than flattening both to "not connected".
+   */
+  status: string;
+  /** Whether this individual account is usable. */
+  connected: boolean;
+  /**
+   * When Composio recorded the connection. Absent when Composio did not say —
+   * and absent for every other connection system, none of which records one.
+   * Rendered as "not recorded" rather than as a blank that reads like "never".
+   */
+  createdAt?: string;
+  /**
+   * The account label the provider published (`ops@acme.test`). Omitted rather
+   * than guessed: an account the provider did not name has no honest label, and
+   * inventing one from the toolkit is how two Gmail accounts become
+   * indistinguishable at exactly the moment the operator has to pick one.
+   */
+  account?: string;
+}
+
 /** One toolkit's connected state, as returned by `GET …/composio/connections`. */
 export interface ComposioConnection {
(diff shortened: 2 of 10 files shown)


oxoxDev (approved):
This closes a seam I raised on both #696 and #699 and could not point at a fix for, so I want to be specific about what landed:

> the toast said "Disconnected Gmail", and Gmail was connected again on the next refresh

That is the failure I described on #696 as "worse than an absent button, because it is silent and it is about access revocation", and on #699 as a capability that existed on the server and was reachable from nothing. `disconnectRouteFor` is the answer to both, and deriving `canDisconnect` from it (`disconnectRouteFor(grid) !== null`) rather than restating the condition is what stops the two drifting — the shape that produced the original bug.

The doc comment carries the history properly: recording that this was `via.includes("native")` before #696 and *why that was right then* means the next reader sees a transition rather than an inconsistency.

Composio winning when a provider is connected through both is correct on the stated ground: the native secret is read by no agent (#396), so blanking it releases nothing an agent can feel while the tile stays connected. And carrying the account ids on the `composio` arm — "a toolkit with two accounts has two, and exactly one of them is the operator's target" — pushes the ambiguity to the only surface that can ask about it instead of guessing.

**The poll fix is the better catch of the two.** Waiting for `connected` when adding a second account to an already-connected toolkit meant the first tick announced a sign-in that had not happened, two seconds in, with the provider tab still open. Waiting for an account id it had not seen before answers the first connect and the second with one rule rather than branching on which case it is.

**The three untruthful inheritances are the part I would hold up as the standard.** Declining a "Default" chip because `composio_execute` sends no connection id, so nothing in this codebase selects or defaults an account — and saying that a default would be a harness change rather than a console one — is refusing to render a decision the product does not make. Same for "connection date not recorded" over a blank that reads as never, and for stating that usage is per provider rather than per account. A host that cannot answer the usage route saying so rather than rendering a zero is the #404 rule applied where it is least convenient.

### Question — a disconnected provider keeps the host's own secret

When a tile is connected through both, this releases the Composio account and deliberately leaves `oauth/{provider}` in place. The reasoning is sound *today*: that secret is inert because #396 established no agent reads it, so blanking it would release nothing while the tile stayed connected.

But the operator's model of Disconnect is "this company no longer holds my credential", and after this the company does still hold it — a real OAuth token, at rest, for a provider the console now shows as disconnected. Two futures make that matter: #396 landing would make the secret live again behind a provider someone believes they revoked, and a credential the operator does not know is retained is the kind of thing that reads badly in an audit long before it reads badly in an incident.

I do not think it belongs in this PR — the route that blanks it is the one whose success message was the original lie, and chaining both revokes on one click needs its own thinking about partial failure. But it is worth an issue, and worth a sentence in the panel if the answer is "retained deliberately", because right now the panel's silence and the toast's confidence say the same thing the old Disconnect said.

**Before merging:** base is current; the lanes are still reporting. Nothing else outstanding from me.

PR #828, opened 2026-08-13
fix(console): stop offering the inert native OAuth catalog (#822)

### LINKED ISSUE #822: connections: stop offering the inert native OAuth catalog from the provider grid
## Problem

The Connections page still lists the native OAuth catalog as something you can connect, and it is inert: `oauth/{provider}` is written by the Connect button and read by nothing (#396 — zero readers under `src/harness/`, and hosted tenants are injected no `OPENCOMPANY_OAUTH_*` at all).

@oxoxDev called this on #404 while settling how the detail view should treat it:

> **The native catalog comes off the Connections page.** This issue asks for a view honest about an inert system; if it isn't a connection, the honest render is not to list it at all. Doesn't have to happen in this slice — just don't build it a detail view.

#819 held both halves of that: the native path got no detail view, and a provider that also holds a native credential is named as a caveat rather than as a second connection. The listing itself is untouched, which is the remaining half.

So an operator on a self-hosted instance can still click Connect on a native-only tile, complete a real OAuth handshake, see the tile go green, and have given their agents nothing. That is the #396 defect; what this issue covers is the page no longer inviting it.

## Proposal

Stop offering the native route from the provider grid.

Concretely, in `frontend/src/lib/`:

- `connectRoute` no longer returns `{kind: "native"}` — the arm becomes `unavailable`, which the tile already renders without an action.
- `CONNECTION_PROVIDERS` keeps its metadata role (brand colours and the ids the host's `well_known` table knows) but stops being a source of *offerable* tiles, so a provider the backend catalog does not carry no longer appears solely because the console has a logo for it.
- A provider already connected natively keeps saying so — removing the offer must not silently hide a credential the company has stored. `via: ["native"]` and its Disconnect stay until #396 is settled one way or the other.

Worth doing in the same pass: `docs/modules/server/README.md` describes the native route as a live path.

## Alternatives

- **Fix #396 instead** — make the native credential something an agent reads, and the listing becomes honest by itself. That is the larger piece of work and has no owner; this is the cheap half that stops the page making a promise in the meantime.
- **Keep the tiles, disable the button.** Rejected on the same grounds as the detail view: a disabled Connect still says "this is a connection you could have", and it isn't.
- **Leave it.** The status quo, and the reason it is uncomfortable: the page's one grid is otherwise now a single honest answer per provider (#582), and this is the last row in it that isn't.

CodeGhost21:
## Summary

The Connections page still listed the native OAuth catalog as something you can connect. The route it offered was not broken, and that is what made it worth removing: `POST …/connections/{provider}/start` completes a real handshake against a provider application the operator registered themselves, and the callback stores `oauth/{provider}` — which is read by **no agent tool anywhere under `src/harness/`** (#396; the only hit on this checkout is a fixture string in `planning/test.rs`). So an operator on a self-hosted instance could click Connect on a native-only tile, complete a real OAuth handshake, see the tile go green, and have given their agents nothing.

#599 removed the Connect buttons that *failed*. This is the one that succeeded and bought nothing. #819 held the other half of @oxoxDev's call on #404 — the native path got no detail view; the listing is what was left.

## What changed

All of it in `frontend/src/lib/`, as the issue proposes.

- **`connectRoute` loses its `native` arm.** `{ kind: "native" }` is gone from `ConnectRoute` entirely rather than merely deprioritised, so no combination of tier and reach can reach it. A host reporting `static` takes the Composio route where it has one, and reports `unavailable` — which the tile already renders without an action — where it does not.

  Worth naming: the arm used to take precedence over everything, on the reasoning that `static` is a deliberate act by the operator and preferring Composio would take away the hatch they configured. What that missed is what the hatch confers. It preserved their configuration by handing them a green tile and no capability.

- **`CONNECTION_PROVIDERS` stops being a source of offerable tiles.** `buildGridProviders` no longer appends every local tile the backend catalog missed. It keeps its metadata role — brand colours, the ids `well_known` keys, the Composio slug each maps to, the `x`/`twitter` alias fold — so a provider the backend catalog does not carry no longer appears solely because the console has a logo for it.

- **A provider already connected natively keeps saying so.** The tail that used to be "everything we have metadata for" is now "everything the host reports as connected", catalog or no catalog. `via: ["native"]` and Disconnect stay, per the issue. `providerId` gained a middle fallback — local tile, then the host's own spelling from `GET …/connections`, then the slug — so `DELETE …/connections/{provider}` can still name a connected provider the console has no tile for, now that its row is the only reason it has a tile.

- **An honest empty state.** The native fallback used to paper over a host whose catalog is genuinely empty; without it, "No provider in All." would read as a broken filter rather than as a host with nothing to offer.

- `docs/modules/server/README.md` no longer describes the native route as a live console path, and the `static` row of the `credentialSource` table says what the console does with it now.

The host routes are **untouched**. This is the console declining to offer a path, not the path being removed — settling #396 makes the offer honest again, and reinstating it is one arm in `connectRoute`.

### One correction to the issue's picture of the blast radius

The issue (and my first draft of the new spec) assumed a host without Composio serves an empty catalog. It does not: `GET …/composio` answers `200` even in a build carrying no `composio` feature — `inBuild: false`, `catalogSource: "fallback"`, and a built-in starter list of eight slugs. So on a typical self-hosted instance this removes **five** tiles (Dropbox, Stripe, HubSpot, X, LinkedIn — the `CONNECTION_PROVIDERS` entries that list does not carry), not eleven, and the eight the host does offer stay and say "not available here". All eleven go only on a host that answers no catalog at all. The code comments and the new spec say so; this is also why the empty state is rarer than it looks and is exercised through a stubbed `/composio`.

## How it was verified

Beyond the unit and Playwright suites, the defect and the fix were both driven end-to-end against the real binary with the native hatch **genuinely open** — a local stand-in provider behind `OPENCOMPANY_OAUTH_SLACK_AUTHORIZE_URL` / `_TOKEN_URL`, a real `OPENCOMPANY_OAUTH_STATE_SECRET`, and a company declaring `[[connection]] slack` — so the handshake under discussion actually completes rather than being described:

| | pre-fix bundle | post-fix bundle, same data dir |
| --- | --- | --- |
| tiles | 13 | **8** (the backend's own list) |
| Slack | Connect offered → handshake completed → host stored `oauth/slack`, tile green "Acme Workspace" | no Connect; "not available here" |
| Dropbox / Stripe / HubSpot / X / LinkedIn | listed | **gone** |
| the connection the handshake made | — | still listed, still labelled, still releasable |

Then Disconnect from the post-fix bundle: the host went `connected: true, via: ["native"]` → `connected: false, via: []`, the stand-in provider logged the revoke, and Slack fell back to "not available here" with no Connect — while `credentialSource` stayed `static`. Which is the whole point: the route still works, and the page no longer invites it.

## Commands run

- `npm run typecheck` / `npm run typecheck:e2e` — clean
- `npx vitest run` — 603 passed (50 files)
- `PW_BASE_URL=… PW_STORAGE_STATE=… npx playwright test` against an isolated host — **127 passed, 10 skipped** (the live-brain lane), including the four new specs and the three pre-existing connections specs

## Tests

`frontend/test/e2e/connections-native-not-offered.spec.ts` is new: the console-only tiles are gone, no Connect is offered where none can confer anything, a natively connected provider keeps its tile and its Disconnect, and an empty catalog explains itself. Two of the four stub one host answer each — a stored native credential, and a host with no catalog — because this harness can produce neither; both are pinned on the merge itself in `test/unit/provider-grid.test.ts`, and the spec header says which is which and why.

`connection-route.test.ts` swaps its two `{ kind: "native" }` assertions for the routes that replace them, and adds a sweep asserting no tile routes natively under any tier/reach combination. `provider-grid.test.ts` replaces "keeps every local tile when the host offers no Composio catalog" — the assertion that *was* the defect — with the connected-tail rules.

## Notes for review

`armTourResume` (issue #300) loses its only caller: it existed for the full-page navigation the native Connect performed, and the Composio sign-in that remains opens a tab instead. The read half stays wired — `TourController` still honours a marker, and the host still redirects an in-flight handshake back to `/connections?connected=…`, so a marker written by an older bundle is still consumed. I kept the function with a note rather than deleting the machinery, which would have been a second, unrelated change; say the word if you would rather it went.

Closes #822.

## Summary by CodeRabbit

* **New Features**
  * Provider listings now use the backend catalog and display connected providers, including previously unavailable entries.
  * Added a clear empty state when no providers are available.
  * Connected provider aliases are combined into a single listing while retaining disconnect support.

* **Updates**
  * The Connections page now offers Composio connections only.
  * Existing native connections remain visible and can be disconnected, but no new native authorization flows are offered.
  * Providers that cannot be connected are shown as unavailable without a Connect action.

* **Tests**
  * Added coverage for provider visibility, unavailable states, alias handling, and native connection behavior.


oxoxDev (approved):
The framing is what makes this worth doing:

> The route it offered was not broken, and that is what made it worth removing … complete a real OAuth handshake, see the tile go green, and have given their agents nothing.

#599 removed the buttons that failed; this removes the one that succeeded and bought nothing. A working control that confers no capability is harder to notice and worse to hit than a broken one, because nothing ever tells the operator.

Deleting `{ kind: "native" }` from `ConnectRoute` **entirely** rather than deprioritising it is the right move — no combination of tier and reach can reach an arm that does not exist, which is a stronger property than an ordering that a later edit can reshuffle. Same instinct as making an unscoped fetch grant unrepresentable in #739.

Naming why the old precedence existed, and what it missed, is the part I would keep:

> It preserved their configuration by handing them a green tile and no capability.

That is a fair correction of a decision that was defensible on its own terms, and stating both halves means the next person does not restore it from the same reasoning.

Two things you handled that the removal could easily have broken:

- **A provider connected natively still says so**, with `via: ["native"]` and Disconnect intact — so existing native connections are not orphaned by the catalog losing its tile-source role.
- **`providerId`'s middle fallback** — local tile, then the host's own spelling, then the slug — so `DELETE …/connections/{provider}` can still name a connected provider the console has no tile for, which is newly possible now that its row is the only reason it has a tile at all. That is the kind of second-order breakage a removal usually ships with.

The honest empty state matters too: without the native fallback papering over an empty catalog, "No provider in All." had to stop reading as a broken filter.

### Question — the trap is gone from the console, not from the host

`POST …/connections/{provider}/start` is still registered (`src/server/ops/connections.rs:69`). So the handshake that "succeeded and bought nothing" is still reachable — by a script, an older console, a bookmarked URL, or anyone reading `docs/modules/server/README.md`'s route table. It still stores `oauth/{provider}`, and nothing under `src/harness/` still reads it.

Your scope is deliberate and stated — "all of it in `frontend/src/lib/`, as the issue proposes" — so I am not asking you to widen it here. But the console was the only thing that made this discoverable, and removing it means the route is now a live path with no surface warning about it. Worth either a deprecation note at the route itself, or an issue to retire it alongside whatever resolves #396, so the decision is recorded where the next person will be standing.

Related, and narrower than when I raised it on #819: with the console no longer able to create native secrets, the residual set is bounded and shrinking rather than growing. That is a real improvement to the question I asked there — a Composio disconnect still leaves an `oauth/{provider}` secret behind, but after this, fewer of them can exist to be left.

**Before merging:** GitHub still reports `mergeable: UNKNOWN` and the branch is 12 commits behind. Worth a rebase before merge given how much of the connections surface has moved this week — #819 and #827 both land near this.

CodeGhost21:
> Worth either a deprecation note at the route itself, or an issue to retire it alongside whatever resolves #396, so the decision is recorded where the next person will be standing.

Filed as #838, and keeping this PR frontend-only as its scope states.

One correction that came out of writing it up, because it changes what #838 can wait for: **#396 is already closed as completed** (2026-08-08) — this PR is what made its offer honest. So there was nothing left to retire the route "alongside", and without its own issue the write path would have quietly outlived the issue it was closed under. #838 is that issue.

Two things I verified against `upstream/main` while writing it, both of which sharpen your point rather than soften it:

- The routes are **not** feature-gated in any meaningful sense — `oauth` is in `default`, so `POST …/connections/{provider}/start`, its `disconnect`, and the unscoped `GET /api/v1/oauth/callback` are registered in a shipped build.
- The product's own planner already encodes the consequence. `verify_connection` (`src/harness/planning.rs:1189-1194`) reports a natively-connected provider as `missing`, with a note saying the credential is real but no agent can reach it. So the planner has been treating a completed native handshake as *not a connection* for some time, while the route that creates one stayed live — which is a second witness for the thing this PR removes the button for.

#838 proposes the note at the route plus a decision on its fate, and argues `disconnect` should outlive `start`: a company can still hold an `oauth/{provider}` secret written before this PR, and after it no console can create a new one — so retiring `start` first shrinks the residual set to zero without stranding anyone. That is the bounded-and-shrinking property you noted from #819, stated where someone can act on it.

Not touching the merge conflict on this branch — another session is resolving it.

PR #829, opened 2026-08-13
feat(connections): open a remote MCP server the way a Composio provider opens (#821)

### LINKED ISSUE #821: feat(connections): open a remote MCP server the way a Composio provider opens (#404)
## Problem

#404 asked for a provider detail view across the three connection systems OpenCompany has. #819 delivers it for one of them: a Composio provider opens into accounts, statuses, dates, usage and a per-account disconnect. A **remote MCP server** still cannot be opened.

That leaves the page uneven in a way an operator can feel. A company routing most of its real work through MCP servers gets a detail view for the surface it uses least. `McpServersSection.tsx` lists servers with connect/disconnect and nothing behind them: no view of what a server has been used for, no per-server status detail, no statement of what a disconnect there does and does not revoke.

It was left out of #819 deliberately rather than missed — `McpServersSection.tsx` is a separate 717-line surface with its own OAuth flow and console shape, and folding it into a Composio port would have been two ports in one PR. The Composio panel names MCP as a system without opening one.

Two things have changed since #404 was written, and both make this cheaper now than it looks:

- **MCP tool calls are metered** (#698 / #744). Every successful call records an `OauthCall` sample under `mcp:<server>` (`src/metering/oauth.rs`), so the usage half of a detail view has real data — the panel from #819 already reads `byProvider` and would need only the prefixed lookup.
- **The honest-rendering decisions are made.** Date not recorded, usage counted per connection rather than per account, disconnect stating its scope — #819's `ProviderDetail` is where they live, and its shape is reusable.

## Proposal

Open a remote MCP server into its own view, reusing `frontend/src/views/connections/ProviderDetail.tsx` rather than growing a second panel with its own vocabulary.

What it has to state, and what is already known about each:

- **Connected or not, and as what.** From the existing MCP status surface.
- **Usage** — `mcp:<server>` in `byProvider`, a real number. `callsForProvider` in `frontend/src/lib/connection-detail.ts` matches on the bare slug today and would take the prefix.
- **When it was connected** — **not recorded**, and say so. MCP has no such concept at all, which is the same answer the native path gets and for the same reason.
- **Disconnect, with its scope stated.** What removing a server does to a running agent's tool belt, and what it does not do at the server's own end.

The smallest first slice is read-only: open, state, show usage. Connect and disconnect can stay where they are until the panel is proven.

## Alternatives

- **A second, MCP-specific panel.** Rejected on #404's own terms — the point of the port is one vocabulary for "a connection", and two panels drift into two.
- **Leave MCP as a list.** Tolerable today, and it is what ships now. It stops being tolerable for any company whose agents work mostly through MCP, which is the case #698 already found under-reported everywhere else.

CodeGhost21:
Closes #821.

Follows #819, which is already merged — `ProviderDetail.tsx` and
`connection-detail.ts` come from it, and this branch was cut from its head. The
diff here is #821's own: three commits, eight files.

## What this does

#819 gave a Composio provider a detail view and deliberately left MCP as a list
— the uneven half of #404, and the wrong half to leave for a company routing its
real work through MCP servers. This opens a remote MCP server into the **same**
panel, on a `ConnectionSubject` union. Not a second MCP-specific panel, which is
exactly what #404 exists to prevent, and what happened to the MCP screen once
already (#414).

Read-only, as the issue's "smallest first slice" asks. Enable, `Test`, `Tools`
and `Remove` stay on the row; the panel adds what the row cannot say:

| Claim | How it is backed |
|---|---|
| Connected, and as what | `mcpStanding()` — `enabled` and the last probe kept apart |
| Usage | `byProvider` under `mcp:<server>` (`src/metering/oauth.rs`, #698/#744) |
| When it was connected | Stated as not recorded; MCP has no connect step |
| What a disconnect reaches | Tool belt next turn; nothing at the server's own end |

## Two decisions worth review

**`callsForProvider` was not taught the prefix**, though the issue suggests it
"would take the prefix". Its existing test pins the opposite, and correctly:
prefix-matching would re-create by hand the `gmail` / `mcp:gmail` collision the
namespace was chosen to prevent. `mcpProviderSlug()` builds the namespaced key
instead and the lookup stays one exact match on one key.

**The panel renders from `McpServersSection`, not `ConnectionsView`.** The live
health an operator just pressed `Test` for lives in that component's state, so
hoisting it would let the panel and the row it opened from report different
health. It also means Settings, MCP Servers gets the detail view for free.

Two smaller ones. `mcpStanding` keeps "never probed" apart from every probe
status including `unknown` — a server nobody has pressed `Test` on is neither
reachable nor broken, and it is the common case on a host without the
`openhuman` feature. And `probedOn` returns null rather than the epoch for a
zero `checkedAtMillis`, on the same grounds as the issue's rule about zeros.

The row's name got a chevron rather than hover styling alone — caught by looking
at a screenshot of the real page, where the clickable name was indistinguishable
from plain text.

## Commands run

- `npm run typecheck`, `typecheck:unit`, `typecheck:e2e`, `npm run build`
- `npm test` — 645 pass (52 files), including 22 render tests over both arms
- `npx playwright test` against a real `target/debug/opencompany` host —
  **124 pass**, 10 skipped (live-brain), including a new `mcp.spec.ts` case
  asserting the four claims above
- `scripts/ci/assert-design-tokens.sh`, `scripts/ci/assert-md-line-cap.sh`

## Not covered

The Composio arm has no e2e walk. The default-feature host serves no Composio,
so the grid renders no openable tile — its coverage is the vitest render suite,
which is where #819 put it for the same reason.

### DIFF
--- a/docs/modules/mcp.md
+++ b/docs/modules/mcp.md
@@ -148,6 +148,32 @@ Settings page built on it crashed on open (issue #414). The client casts an
 unparsed body to the declared type, so a second surface is never caught by the
 compiler — only by whoever opens the page.
 
+### Opening one server
+
+A row's name opens the server into
+[`ProviderDetail`](../../frontend/src/views/connections/ProviderDetail.tsx) — the
+same panel a Composio provider opens into (issues #404, #821), on a
+`ConnectionSubject` union rather than a second MCP-specific panel. The reason is
+the paragraph above one level up: two surfaces describing the same idea acquire
+two vocabularies and then drift.
+
+The panel is read-only. Enable, `Test`, `Tools` and `Remove` stay on the row;
+what it adds is what the row cannot say —
+
+- **Connected, and as what.** MCP has no connection object, so this is assembled
+  from two facts a single badge would collapse: `enabled` (whether any agent
+  receives the tools at all) and the last probe (whether the endpoint answered
+  when someone last asked). A server nobody has pressed `Test` on has no `health`
+  at all, and "never probed" is neither reachable nor broken. See `mcpStanding`
+  in `frontend/src/lib/connection-detail.ts`.
+- **Usage**, read from `byProvider` under the `mcp:<server>` key this module's
+  metering records (`src/metering/oauth.rs`) — never the bare slug, which is the
+  same-named Composio toolkit's row.
+- **No connection date**, stated rather than left blank. There is no connect step
+  to record one; the probe timestamp the host *does* keep sits beside it.
+- **What a disconnect reaches**: the tool belt on the next turn, and nothing at
+  the server's own end. A manifest server says it cannot be removed at all.
+
 ## Pool-staleness caveat
 
 Agents materialize their MCP registry once, when the
--- a/frontend/src/lib/connection-detail.ts
+++ b/frontend/src/lib/connection-detail.ts
@@ -1,4 +1,4 @@
-// What a provider's detail view is allowed to claim (issue #404).
+// What a connection's detail view is allowed to claim (issues #404, #821).
 //
 // The issue's rule — "do not invent a number … say so on screen rather than
 // rendering a plausible-looking zero" — is not only about usage. Two of the
@@ -10,10 +10,16 @@
 //  - **what has gone through it**: metered per *provider*, never per account,
 //    and an unmetered host is not a host where nothing happened.
 //
-// These are the two decisions, kept out of the component so they can be pinned
+// A remote MCP server (#821) adds a third, which is the one its own surface
+// gets wrong today: it has no connection object at all, so "connected" has to
+// be assembled from two facts that a single badge would collapse — whether the
+// harness hands its tools out, and what the endpoint said the last time anyone
+// asked. See [`mcpStanding`].
+//
+// These are the decisions, kept out of the component so they can be pinned
 // without a DOM.
 
-import type { ProviderCallsDto } from "@/api/types";
+import type { McpHealth, McpServer, ProviderCallsDto } from "@/api/types";
 
 /**
  * "connected 9 Aug 2026", or a statement that no date exists.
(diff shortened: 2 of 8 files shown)


oxoxDev (approved):
Opening MCP into the **same** panel on a `ConnectionSubject` union, rather than building a second MCP-specific one, is the decision #404 exists to force — and citing #414 as the time that already went wrong here is the right way to justify it. A second panel is always the cheaper next step and always the thing that has to be undone later.

The usage lookup is the part I checked hardest, because I asked about this namespace on #744, and it is handled better than I hoped:

> Prefix-matching here would re-create by hand the collision that naming was chosen to prevent.

Matching on the bare slug against `mcp:<server>` — rather than stripping a prefix and comparing — means a company with a Composio `gmail` and an MCP server its operator also called `gmail` cannot read one total as the other's. That is the consumer-side half of #698's namespacing, and getting it wrong here would have quietly undone the reason the prefix exists. The prefix is a key, not a label, which also answers the question I left on #744 about whether `mcp:linear` reaches the operator's eye.

**"Zero is a real answer, not a missing one"** is the same discipline #819 and #827 apply, and it is worth the sentence: because every successful call is metered, a provider with no row has genuinely had none, so the panel can say zero rather than hedging. A claim table that says how each line is backed is the right shape for a read-only slice.

### Question — this is the third cross-language mirror, and drift here makes "zero" lie

`mcpProviderKey` is described as "the mirror of `mcp_provider` in `src/metering/oauth.rs`: trimmed, lowercased…". That makes three mirrored predicates now living in this repo:

- `mayCarryACredential` (TS) mirroring `may_carry_a_credential` (Rust) — #747
- `hub_accepts_redirect_uri` (Rust) mirroring the backend's `isLoopbackHttpUri` (JS) — #818
- this one

The first two have something this one does not. #747's is *advisory*: the core enforces, so a drift yields a confusing UI rather than a wrong result. #818's mirrors an **external** contract and is built to be deleted when backend#1243 lands, with a test that goes red to force it.

This one mirrors an **internal** Rust function that can change at any time, and a drift is silent and wrong in a specific way: if `mcp_provider`'s normalisation moves — a different sanitisation, a trailing-dot rule, anything — the TS key stops matching, `byProvider` returns no row, and the panel reports **zero**. Which this PR's own doc defines as "a real answer, not a missing one". The failure mode is the panel confidently telling an operator their MCP server has had no calls.

Cheap ways to close it, both with precedent here: assert the two agree on a shared fixture list of server names, the way I suggested for `mayCarryACredential`; or generate the constant from the Rust side, which is what #728 proposed for the gated-tool labels and #761 discussed again. Either turns a silent wrong answer into a failing test.

Not blocking — the normalisation is simple today and unlikely to move soon. But "unlikely to move" is what makes a mirror worth pinning rather than watching.

**Before merging:** base is 8 commits behind and the lanes are still reporting. #828 and #827 both land on the same surface, so worth a rebase and a re-check rather than trusting the current run.

Issue #852, opened 2026-08-13
console: Node 25's unflagged WebStorage defeats vitest's populateGlobal, so window.localStorage is an empty object and 57 tests fail


M3gA-Mind:
## What happens

Installing the console's dependencies **within the range `package.json` declares** breaks 57 unit tests across three files:

```
 ❯ test/unit/connection-registry.test.ts (29 tests | 29 failed)
 ❯ test/unit/desktop-bridge.test.ts      (17 tests | 17 failed)
 ❯ test/unit/tour-resume.test.ts         (11 tests | 11 failed)

 Test Files  3 failed | 50 passed (53)
      Tests  57 failed | 601 passed (658)
```

Every one of them is the same error:

```
TypeError: window.localStorage.clear is not a function
 ❯ test/unit/connection-registry.test.ts:31:23
     29| beforeEach(() => {
     30|   resetConnections();
     31|   window.localStorage.clear();
       |                       ^
```

## Cause

`frontend/package.json` declares:

```json
"jsdom": "^27.0.1"
```

`frontend/package-lock.json` pins **27.0.1**. The caret admits **27.4.0**, which is what a fresh resolution picks today, and in 27.4.0 `window.localStorage` no longer presents `clear` the way these tests use it.

So the declared range is **wider than the code tolerates**. Nothing is wrong with the pinned version; what is wrong is that the constraint permits a version that is not.

## Why CI is green and will stay green until it isn't

The `Console` job installs with `npm ci`, which installs the lockfile **exactly** and ignores the range. That is deliberate and correct — the job's own comment says so:

> `npm ci` … which means CI would be typechecking a dependency tree that is not the one committed — and a green run could not be attributed to the diff.

The consequence is that this defect is **invisible to CI by construction**. It surfaces the moment anyone regenerates the lockfile, adds a dependency, or installs with a resolver that is not reading `package-lock.json` — at which point 57 tests fail at once, in three files unrelated to whatever they were changing, and the cause is nowhere near the diff they are looking at.

Found while working #785, on a `pnpm install` in a clean clone. It cost a stash-and-rerun to establish the failures were not mine; the next person will pay that too, or worse, will assume they broke something.

## Fix

Pin the constraint to what the lockfile and the code agree on:

```json
"jsdom": "~27.0.1"
```

or `27.0.1` exactly. Then either fix the three suites against 27.4.x and widen the range again deliberately, or leave it narrow with the reason recorded — but the range should not silently admit a version that fails the suite.

[...]

Issue #673, opened 2026-08-11
approvals: an outward fetch parks on every call, because a grant can only name the tool and not the host


M3gA-Mind:
## What this is

`web_fetch`, `http_request` and `curl` park on every call, in every tier that parks
anything. A research agent fetching twenty URLs raises twenty approval cards, and
`web_fetch` mutates nothing at all.

This is the one remaining population of repeated approvals after #559 (Composio reads stop
parking) and #560 (sandbox writes stop parking under `auto`). It surfaced while working out
whether #563 still had a subject — see
[that analysis](https://github.com/tinyhumansai/opencompany/issues/563#issuecomment-5255799584);
this is the narrow remainder it identified, deliberately filed on its own rather than
folded back in.

## Why they are per-call today, and why that reason does not close the question

The declaration is deliberate and the rationale is stated at `src/policy/consequence.rs:310`:

> **Arbitrary code, arbitrary addresses** — The three shapes issue #444 names, plus the two
> web tools that share `http_request`'s shape. A standing grant on any of these is a
> standing grant on "anything the sandbox permits", which is not a sentence an operator can
> consent to.

That is correct and should not be reversed. But read precisely, it is an argument against
consenting **by tool name**. It is not an argument against consenting to *"fetch from
`docs.rs`"*, which is a sentence an operator can absolutely consent to — and which names
the thing that actually varies.

The codebase already has the concept for exactly this. #457 added `StandingGrant.scope` so a
grant could be confined to a **slice** of a tool rather than the whole of it, because
`composio_execute` carries every action under one name. An outward fetch has the same
shape: one tool name, an unbounded space of destinations, and a consent sentence that only
makes sense once the destination is named.

## The proposal

Let a standing grant on the web trio be scoped to a **host**, the way a Composio grant is
scoped to a toolkit. "Read from `docs.rs`" becomes grantable; "make any HTTP request"
stays refused.

`standing_scope_of` (`src/policy/consequence.rs`) is the existing seam — it returns the
toolkit for `composio_execute` and `None` for everything else. A host derived from the
request URL is the same kind of answer for these three tools.

## Deliberately out of scope

Stated explicitly, because this is the narrow remainder of a much larger issue and the
narrowness is the point:

* **No rule engine.** This is the existing standing-grant mechanism given a scope, not a
  new persisted allow-rule system.
* **No listing screen or revocation UI.** `StandingGrantMinted` / `StandingGrantRevoked` /
[...]

PR #853, opened 2026-08-13
fix(approvals): render a host-scoped grant's URL verbatim

### LINKED ISSUE #785: approvals: a host-scoped standing permission renders as "Https://docs.rs" — a toolkit label function applied to a URL
## What happens

A standing permission scoped to a host renders its scheme capitalised:

> Fetch a web page — **Https://docs.rs** only

Found by driving the console in a browser against a real host — CI is green and the PR was reviewed; this is only visible on screen.

## Reproduce

1. Run a company where an agent holds `web`.
2. Have the agent call `web_fetch` against `https://docs.rs/...` — it parks (unbounded reach, #338).
3. Approve with **"Let &lt;agent&gt; use this tool for 1 hour"** (`scope: "tool"`).
4. Open **Approvals → Standing permissions**.

## Cause

`providerLabel` in `frontend/src/views/ApprovalsView.tsx:427` was written for **Composio toolkit identifiers**, and its doc says so:

> A toolkit identifier as a person would write it: `microsoft_teams` → "Microsoft Teams". Mechanical on purpose.

```ts
function providerLabel(scope: string): string {
  return scope
    .split("_")
    .filter(Boolean)
    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
    .join(" ");
}
```

**#673/#739 introduced a second kind of scope** — a URL origin, `scheme://host[:port]` from `standing_scope_of` — and `grantHeadline` (line 416) sends both kinds through the one function:

```ts
return g.scope ? `${action} — ${providerLabel(g.scope)} only` : action;
```

`"https://docs.rs"` contains no `_`, so it survives as a single "word" and only its first character is upper-cased: `Https://docs.rs`.

## Why it is worth fixing rather than ignoring

The row exists so an operator can **read a permission in order to decide whether to revoke it** — that is the stated reason `providerLabel` was written at all. A mangled scheme undermines exactly that: `Https://` is not a thing, and an operator checking whether a grant covers the host they think it does should not have to wonder whether the display is lying about the rest of the string too.

It is cosmetic in consequence and precise in cause, so it is cheap.

## Suggested fix

`grantHeadline` should branch on the **kind** of scope rather than assuming toolkit. A host scope is already distinguishable — it contains `://`:

```ts
const label = g.scope.includes("://") ? g.scope : providerLabel(g.scope);
```

Better still, have the host say which kind it is rather than having the console infer it from punctuation — the two scope kinds come from different code paths (`standing_scope_of` for toolkits, the URL origin for `web_fetch`) and the console currently cannot tell them apart except by sniffing.

**Whichever is chosen, pin it with a test** that asserts a host-scoped grant renders its scheme verbatim. A snapshot of the toolkit case alone would not have caught this.

M3gA-Mind:
## Summary

A standing permission scoped to a host rendered its scheme capitalised — **`Fetch a web page — Https://docs.rs only`**. Confirmed on staging today, on three live grants (`Https://www.theguardian.com`, `Https://www.bbc.com`, `Https://www.espn.com`).

`grantHeadline` sent **both kinds** of `StandingGrant.scope` through `providerLabel`, which was written for Composio toolkit slugs and upper-cases the first letter of each `_`-separated word. A URL origin has no `_`, so it survived as a single "word" and only its scheme was touched.

The two kinds are now told apart in one named place. `grantHeadline` moves from `ApprovalsView.tsx` into `lib/language.ts` — next to the `toolAction` it composes with (which its own doc already pointed at), and where the unit runner can reach it: it was module-private in a `.tsx`, and `vitest.config.ts` is explicit that the runner is for pure functions with no React. It gains `scopeLabel`, the single function a third kind of scope would have to teach.

### Which fix, and why

The issue offered two options. I took the value-side branch, and **rejected the typed host-side discriminator** — with reasons, since the issue is right that sniffing punctuation can re-create this class of bug:

- `scope` is the **enforcement key**. `StandingGrant::admits_scope` (`src/runtime/grants.rs`) matches a live call against it with **exact string equality**, and `standing_scope_of`'s own doc insists the mint side and the live-call side must read with identical code *"or a grant could be minted that never matches its own tool"*.
- It is `Serialize, Deserialize` and **journal-replayed** — the field docs repeatedly describe grants *"replayed from a journal line written before this field existed"*.

So retyping it is a persisted-format change to a security-relevant comparison, not a label fix, and it does not belong in a cosmetic label PR.

But the part of that suggestion that *was* free has been taken. Both the Rust field doc and the TS `StandingGrant.scope` doc still described a Composio toolkit **only** — three releases after #673/#739 added the URL origin. That stale documentation is precisely why the console's assumption read as true to everyone who looked at it. Both now name the two kinds, and say that a *displaying* reader must tell them apart while a *matching* one must not care.

If a third kind ever arrives, `scopeLabel` is the one place that has to learn about it, and the type doc on both sides now says so.

## API Or Behavior Changes

None. No wire, storage, or matching change — `admits_scope` and the journal format are untouched. The only behaviour change is the rendered string for a host-scoped grant:

| scope | before | after |
| --- | --- | --- |
| `https://docs.rs` | `Fetch a web page — Https://docs.rs only` | `Fetch a web page — https://docs.rs only` |
| `microsoft_teams` | `… — Microsoft Teams only` | unchanged |
| absent | `Write a file in its workspace` | unchanged |

## Tests

Four cases in `frontend/test/unit/language.test.ts`, covering both scope kinds and the unscoped row. The host cases assert the origin **verbatim**, including a port (`http://localhost:8080`), an explicit `:443`, and a multi-label domain.

**Revert proof.** With `scopeLabel` reverted to the old speller and nothing else changed:

```
Tests  2 failed | 29 passed (31)

Expected: "Fetch a web page — https://www.bbc.com only"
Received: "Fetch a web page — Https://www.bbc.com only"
```

The two host cases fail; the toolkit case and the no-scope case stay **green**. That asymmetry is the whole reason this shipped past CI and a review, and it is why the issue asked for more than a snapshot of the toolkit case.

**Console checks run locally** — all four steps the `Console` CI job runs, plus the targeted file:

| command | result |
| --- | --- |
| `npm run typecheck` (`tsc -b`) | exit 0 |
| `npm run typecheck:e2e` | exit 0 |
| `npm run typecheck:unit` | exit 0 |
| `npx vitest run test/unit/language.test.ts` | **31 passed** |

The full unit suite also reports 57 failures in `connection-registry`, `desktop-bridge` and `tour-resume` — **pre-existing and unrelated**: identical with these changes stashed, all `window.localStorage.clear is not a function`, caused by a local `pnpm` install resolving `jsdom` 27.4.0 under the declared `^27.0.1` range where `package-lock.json` pins 27.0.1. CI installs with `npm ci`, so CI is unaffected. Filed separately as #852 rather than widened into this PR.

- [x] `cargo fmt --all -- --check` — not run locally (fleet policy: no local cargo builds on the shared disk). The Rust change is a **doc comment only** — no executable line changes — so CI's Rust lanes are the check.
- [x] `cargo clippy --all-targets -- -D warnings` — as above; doc-comment-only change, covered by CI.
- [x] `cargo build --all-targets` — as above; doc-comment-only change, covered by CI.
- [x] `cargo test` — as above; doc-comment-only change, covered by CI. No Rust behaviour is touched, so no Rust test could distinguish this diff.

## Documentation

The documentation fix *is* half of this change, not an afterthought:

- `src/runtime/grants.rs` — `StandingGrant::scope` now documents the `Some(origin)` kind from #673/#739 alongside the toolkit kind, and states the display/match split.
- `frontend/src/api/types.ts` — the mirrored TS doc does the same, and points readers at `grantHeadline` rather than letting them spell a scope out at a call site.

No `docs/` page describes this string, so none needed updating.

Closes #785

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -416,7 +416,20 @@ export interface StandingGrant {
   expires_at_millis: number;
   /**
    * The slice of the tool it is confined to, when the tool's name is not the
-   * whole of what it can do (#457) — a Composio toolkit like `github`.
+   * whole of what it can do (#457).
+   *
+   * **Two kinds of value, in one untyped string** — both minted by the host's
+   * `standing_scope_of`, and a reader that assumes either one is the bug #785
+   * was:
+   *
+   * * a **Composio toolkit** identifier like `github` — a slug, which has to be
+   *   spelled out before an operator can read it;
+   * * a **URL origin** like `https://docs.rs`, added for `web_fetch` by
+   *   #673/#739 — already exactly what the operator approved, and to be shown
+   *   untouched.
+   *
+   * Render it through `grantHeadline` in `lib/language`, which is the one place
+   * that tells them apart. Do not spell a scope out at a call site.
    *
    * Absent for every tool whose name already says everything, and absent from
    * an older host that predates the field. Both mean "nothing to narrow", so
--- a/frontend/src/lib/language.ts
+++ b/frontend/src/lib/language.ts
@@ -2,7 +2,7 @@
 // never exposes runtime internals ("agent graph", "tier", "dispatch", "cycle",
 // "checkpoint", "A2A"). Everything a person sees goes through this layer.
 
-import type { ApprovalSummary, FeedbackCategory } from "../api/types";
+import type { ApprovalSummary, FeedbackCategory, StandingGrant } from "../api/types";
 
 /**
  * A company's lifecycle state, in plain language, with a status tone.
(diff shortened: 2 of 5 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Fixing this on the **value side** and rejecting the typed host-side discriminator is the right call, and the reasons you give are the ones that matter rather than the convenient ones:

- `scope` is the **enforcement key** — `admits_scope` matches a live call against it with exact string equality, and `standing_scope_of`'s own doc insists mint and match must read with identical code "or a grant could be minted that never matches its own tool";
- it is `Serialize, Deserialize` and **journal-replayed**, so grants written before any new field existed have to keep resolving.

So a rendering bug does not get to reshape the stored key. That containment is the whole review: even if `scopeLabel` mis-classified a scope tomorrow, the blast radius is a wrong sentence on a card, not a grant that stops matching its own tool. I verified the scope-matching path on #739 and would have pushed back hard on touching it for a capitalisation defect.

Distinguishing the two kinds in **one named place** is right for the same reason `disconnectRouteFor` was on #819 — the bug existed because two kinds of value shared one speller, and the fix is a function that has to be taught about a third rather than a branch that has to be found again. Moving `grantHeadline` out of `ApprovalsView.tsx` into `lib/language.ts` also makes it reachable by the unit runner, which is what stops this recurring silently: it was module-private in a `.tsx` and `vitest.config.ts` is explicit that the runner covers pure functions.

Confirming it on staging against three live grants, rather than reasoning from the code, is what makes "this is what operators are seeing right now" a fact.

**0 major. Approving.**

**Before merging:** GitHub still reports `mergeable: UNKNOWN`. Worth a rebase and a green run given how much has landed near `language.ts` this week — #728 and #761 both work on the same file's label maps.

Issue #850, opened 2026-08-13
workflows: a continuation still repeats a shell / curl / git_operations call that already reached a counterparty


M3gA-Mind:
## The gap

Issue #846's fix (`src/workflows/replay.rs`) stops a continuation repeating an outward call an
earlier run in its lineage already made. It covers two node classes:

* `http_request` with a mutating method (anything but `GET`/`HEAD`)
* `tool_call` whose `consequence_of` group is an outward one

**It deliberately does not cover `shell`, `curl` or `git_operations`** — and all three can reach a
counterparty:

| tool | how it leaves the building |
|---|---|
| `shell` | anything at all — `curl -X POST`, `git push`, an emailing CLI |
| `curl` | downloads from an address, and can be pointed at a mutating endpoint |
| `git_operations` | pushes to a remote |

All three are declared `EffectGroup::Other` in `src/policy/consequence.rs`, which is correct for
what that declaration is *for* (naming a consequence class on an approval card) and useless as a
replay predicate: the host cannot tell a `git log` from a `git push`, because the difference is
inside an argument string it does not parse.

## Why #846 did not just fold it in

Replaying these on the strength of a guess would change the behaviour of the most common effectful
node in a workflow. A `shell` node upstream of a gate would stop re-running and start replaying its
recorded stdout on every continuation — correct for the one node that pushed, wrong for the many
that build, lint or read. That is a bigger behavioural change than the bug being fixed, and it
belongs behind its own decision rather than inside a fix for a different one.

## Shape of a fix, for whoever picks this up

Three routes, none free, listed in the order I would try them:

1. **An authored per-node declaration** — `idempotent = true` / `repeatable = false` on the node,
   so the author states what the host cannot infer. Cheap, honest, and it puts the knowledge where
   it exists. Costs an authoring-surface change and does nothing for existing graphs.
2. **Argument-level classification for the three** — parse `git_operations.operation` (already an
   enum, so this one is nearly free), `curl`'s method, and refuse to classify `shell` at all.
   Partial by construction, and it re-introduces the hand-maintained carve-out list that
   `src/policy/consequence.rs`'s module docs were written to delete.
3. **Checkpointed resume**, which makes the whole question moot. Still blocked on the same engine
   work #438 and #496 both recorded: `run_resumable` installs a no-op observer and a process-local
   in-memory checkpointer and takes no cancellation token, so it composes with neither the per-node
[...]

Issue #842, opened 2026-08-13
approvals: consolidate a turn's gated calls into one approval — research currently costs one interruption per domain


M3gA-Mind:
## What happens

The inline approval card offers a scope control:

- `( ) Just this once`
- `( ) Let <agent> use **this tool** for [1 hour]`

Selecting the second option on a `web_fetch` to `https://www.espn.com` does **not** let the agent fetch other hosts. `https://www.bbc.com/sport` and `https://www.theguardian.com/sport` each parked again with their own card.

The grant is **host-scoped**; the label promises **tool-scoped**.

## Evidence (staging, 13 Aug 2026)

Company `marketing--agentic-marketing-agency`, policy tier **Full**.

One chat turn asking an agent to research a topic across sources produced **three separate approval cards** — one per domain. After granting `bbc.com` for 1 hour, two different BBC paths ran unattended in the same turn:

```
Web Fetch — url=https://www.bbc.com/sport/athletics   316ms   8.9k characters
Web Fetch — url=https://www.bbc.com/sport/football    194ms  14.7k characters
```

So the host-scoping itself works correctly and as designed (#739) — a blanket web grant would be a real security regression, and this is not a request to widen it.

## Why it matters

The operator expects one sign-off and receives one per domain, with no way to predict how many are coming. On the **Full** tier — where the stated contract is "agents act without asking" — a three-source research task becomes three interruptions. The scoping is right; the wording is what misleads.

## Suggested fix

Name the host the grant will actually cover, resolved at render time from the pending call:

> Let this agent fetch from `bbc.com` for [1 hour]

That makes the security model visible instead of surprising, and sets the expectation that a second domain will ask again.

Related but distinct: #785 is about how an already-granted host-scope *string* renders (`Https://docs.rs`). This is about the pre-approval radio label misdescribing what the grant will do.

M3gA-Mind:
## Preferred fix (from the operator): consolidate the parks and ask once

Relabelling the control (below) only makes the cost *predictable*. The better fix is to remove the cost: **when a turn needs several gated calls, surface them as one approval, not one per call.**

### The shape

A research turn plans to reach `espn.com`, `bbc.com`, `theguardian.com`. Instead of three sequential cards, one card:

> **Fetch web pages** — SEO Specialist needs to reach 3 sites
> - ☑ `espn.com`
> - ☑ `bbc.com`
> - ☑ `theguardian.com`
>
[...]

PR #854, opened 2026-08-13
fix(workflows): an approval must not repeat an outward call, finish a run that is waiting, or hide what it is approving (#846)

### LINKED ISSUE #435: approvals: an approval raised inside a thread resumes in the channel, not the thread

### LINKED ISSUE #846: workflows: approving a parked node re-executes every upstream node (#438 regression), the run reports Finished with its approval outstanding, and the card never says what it is approving
## A workflow node's parked approval is unreachable, and the run finishes without it

Reproduced on staging just now, on a workflow whose first step is a `web_fetch` tool_call node.

**The run knows it parked.** Run history:

```
● manual   13/08/2026, 19:43:01     [1 pending approval]
  Finished — this run routed no reports.
```

Live run panel while executing: `Run result — 1 pending approval · Waiting on: fetch_bbc`.

**The Approvals page does not.** Checked twice within a minute of the run, and again immediately after re-confirming the run state:

> **All clear** — Nothing is waiting on you. Your company will park anything that needs a sign-off here.

**And the run does not wait.** It reports **Finished** while the approval is still pending, having routed no reports. The gated call never ran, nobody was ever asked, and the workflow reports completion.

## Why this matters more than a missing list entry

The operator has **no reachable surface to grant it**. A chat-originated park renders an inline card in the conversation *and* lists on the Approvals page (verified working today — that path is fine). A **workflow-originated** park has no conversation to render into, so the Approvals page is the only possible surface, and it is empty.

For a **scheduled** workflow this is terminal: at its trigger time it will park, finish, route nothing, and no human can intervene even in principle. The company reports a successful run that did none of its work.

## Relationship to #395

#395 — *"a workflow run never parks an approval, so the Approvals page stays 'All clear' whatever its nodes do"* — was closed as completed by **#437**.

Half of that has clearly shipped: the node **does** now consult the policy, and the run **does** record `1 pending approval` (#460 / #614 behaviour is present and working). The other half — the park reaching the Approvals page — does not hold on staging today. Either #437's fix regressed, or it wired the run-level state without wiring the park into the approvals queue.

## Reproduction

1. Create a workflow whose first node is `tool_call` with slug `web_fetch` and any external URL.
2. **Run** it (not Test run — a dry run does not execute nodes).
3. Run panel shows `1 pending approval · Waiting on: <node>`.
4. Open `#/approvals` → **All clear**.
5. Run history shows `Finished — this run routed no reports`, with the approval still pending.

## Expected

A parked call from a workflow node appears on the Approvals page with its workflow, run and node named, and the run stays open awaiting the decision rather than reporting Finished.

## Environment

Staging, company `marketing--agentic-marketing-agency`, policy tier **Full**, 13 Aug 2026. Workflow `daily-sports-news-blog`.

M3gA-Mind:
## Summary

Three defects on the workflow approval path, in the order they matter.

### 1. Approving repeated an outward call the run had already made

**This is a gap in a narrower fix, not a regression — and the citation in the issue was wrong.**
#846 says #438 was closed by #624. The timeline says otherwise: **#438 was closed by #496**
("approving a paused run does not re-send what it already delivered"); #624 closed #435 (threaded
approvals). That is not pedantry, because it changes the diagnosis:

**#496 deliberately kept re-run-on-approve.** A paused tinyflows run is *settled*, not suspended —
the engine finishes and reports the gates it stopped at — so resuming is necessarily a re-run from
the trigger. #438's written requirement was never "stop re-executing"; it was **"an approval must
never cause a message to be sent to a person twice."** #496 met that by adding a delivery ledger.
Nothing here calls #496 broken, and a reviewer reading this as "the ledger regressed" should read
it as "the ledger's reach was narrower than the exposure."

The reach: #496 guards what the **host** routes with its own hands — an `output` node's report,
dispatched by `deliver_outputs` after the engine settles — because that is the half the host can
decline after the fact. A call made by a **node**, mid-run, through a capability has no post-hoc
dispatch to decline. Put one upstream of a gate and the first run calls, the operator approves, and
the continuation calls again. Same lineage, same exposure, the other mechanism.

**The severity claim in the issue needed correcting too, and the correction widens the work rather
than shrinking it.** The issue's example — "put a `send`, `publish` or `repo_publish` node after two
gated steps and this delivers it twice" — **is not reachable today.** A workflow `tool_call` can only
invoke the four families the invoker wires (shell, code, web, search), and every slug in them is
declared `EffectGroup::Other` except `web_search`, a billed read. `send`/`publish`/`repo_publish` are
agent-turn families workflows do not wire at all.

What **is** reachable today is an **`http_request` node with a mutating method** — arbitrary address,
every company, right now. So this guards both, and neither half is speculative:

| class | rule | reachable today |
|---|---|---|
| `http_request` | method is not `GET`/`HEAD` | **yes** — this is the live duplicate |
| `tool_call` | `consequence_of` group is outward, less `Reach::Money` | no — this is the guard that must exist *before* a send-capable namespace is wired, or wiring one silently re-opens #438 on the day it lands |

**Mechanism.** An outward-call ledger, built on #496's terms: recorded from the settled run's own
output, unioned down the lineage, carried on the approval card, threaded into the continuation input,
and stripped from the gate's dedupe identity. On the continuation the node is rewritten **before the
run starts** to invoke a host-private sentinel slug carrying its recorded result; the invoker answers
that slug from its arguments and touches no toolbelt.

**Why the translation and not the invoker.** `ToolInvoker::invoke` receives `(slug, args, conn)` — no
node id, no run state — which `src/workflows/gate.rs` already records as the reason the *approval*
gate could not live there. The same fact rules out a node-keyed skip inside it. And node identity is
not negotiable: keying on `(tool, args)` instead would miss every send whose body an upstream agent
node regenerates, which is most of them. The translation is where the host holds both the node ids
and the trigger input, and where `apply_policy_gates` already writes per-node decisions.

The `http_request` arm rewrites the node's **kind** to `tool_call`. Safe because the two kinds already
agree about the only thing that leaves a node — every capability node wraps its result in the same
`{ json, text, raw }` envelope, and `envelope::wrap` is a pure function of the recorded value, so
replaying reconstructs the node's output rather than approximating it. **Flagging it explicitly as the
one part of this diff I would most like a second pair of eyes on.**

The recorded result is **JSON-encoded into a single string**. `tinyflows::expr::resolve` walks node
config before the node runs and evaluates every leaf beginning with `=`; a recorded result is
arbitrary data from a counterparty. `serde_json::to_string` never yields a document starting with `=`,
so it is inert by construction rather than by escaping rules. Pinned by a test.

### 2. A run reported `Finished` with its approval outstanding

19 of the 26 runs in the reproduction ended `Finished — this run routed no reports` while their gate
sat undecided.

Root cause: `runTone` and the run row read only parked **deliveries**. A run stopped at a gate never
reaches an output node, so it has no deliveries, no error, no cancellation, and is not `running` —
the engine settled it. Every branch reads "fine" on its own, so it fell through to the green "ok" and
the "Finished" line. `pendingApprovals` has been on the wire since #395 and nothing read it for the
run's *state*.

The run object really is settled; what has not happened is the work, and the work is what the operator
is asking about. So this reports the waiting state rather than claiming the run holds open — and names
the nodes, because a scheduled run that silently did nothing is the failure that makes this matter and
the fix is a click.

### 3. The card never said what it was approving

`GatedCall` was produced only by `policy_gates`, so it existed only for a node the **company's policy**
stopped. An **authored** `requires_approval` produced none — and on a `full`-tier company, where the
policy stops nothing, that is *every* workflow card. The operator was asked to authorise `fetch_bbc`
and shown `{"items":[{"json":{}}],"port":null}`.

#372 made this exact complaint about the chat surface and #375 fixed it there by carrying the call's
own arguments. The information was already on the host in that case and it is already on the host in
this one: `call_of` has read the slug and args since #460, and only the **reason** was ever
policy-specific. So an authored gate is now described from the graph and gains everything except the
sentence nobody wrote. `tool_call` nodes also gain a destination (host only — #614 gave `http_request`
one and left `tool_call` without, which is why a parked `web_fetch` named no host).

Console-side, the card is labelled by its tool ("Fetch a web page") and the payload block shows the
call instead of the resume record. The dropped keys are dropped for stated reasons — notably `input`,
which is what read as a description and was not one, and which carries the accumulating
`{"approvals":[...]}` list an operator reasonably misreads as one card covering several gates.

**Coordination with #842 / #848 (batching):** deliberately untouched. All card changes are in
`frontend/src/lib/language.ts` (pure functions) and **not** in `approval-card.tsx`, specifically so
this does not collide with W2's PR. Nothing here groups or consolidates cards.

## API Or Behavior Changes

**No HTTP/GraphQL API changes.** No route, no wire type gained a required field.

**Wire (additive, all optional).** The `workflow.approve` effect payload gains `performed` (the
outward-call ledger) and `args` (the gated call's arguments); `tool` and `target` are now written for
authored gates as well as policy-raised ones. `args` is credential-redacted host-side by the **existing**
`display_payload` projection that already redacts a chat card's payload — this adds no new redaction
rule and cannot bypass the current one. An older console ignores every one of them; this console treats
absence as "old host" and renders exactly the pre-#846 card, which is pinned by a test.

**Journal / replay.** The trigger input gains a reserved `__opencompany_delivered`-style sibling,
`__opencompany_performed`, written only when there is something to suppress — so a first run's payload
keeps byte-identical shape. Both reserved keys are stripped before two parked gates are compared, or
every continuation gate would read as a new decision and stack a duplicate card.

**Behaviour.**
1. A continuation no longer repeats a mutating `http_request`, or an outward-classified `tool_call`,
   that an earlier run in its lineage already made — it replays the recorded result.
2. A settled run with gates outstanding reads as waiting, not finished, in the run row, the last-run
   chip and the status dot.
3. A workflow approval card names its tool, arguments and destination.
4. The approval card's cost note gains a clause about replayed calls.
5. Everything with no ledger on its input — every first run, every existing test — is unchanged.

## Limits, stated rather than hidden

* **`shell`, `curl`, `git_operations` are not guarded.** All three are `EffectGroup::Other` and all
  three can reach a counterparty. The host cannot tell a `git log` from a `git push`, and replaying
  every shell node's recorded stdout would change the behaviour of the most common effectful node in a
  workflow on the strength of a guess. **Filed as #850**, with three candidate routes, rather than
  folded in here.
* **A per-item fan-out is not replayed.** The invoker sees no item index, so one recorded result cannot
  answer N invocations without inventing which. Refused rather than approximated — and surfaced as a
  run notice, so the operator learns *before* they approve. #496 took the same side of the same trade
  for a partial delivery fan-out.
* **A result too large for the card is not replayed.** A duplicate send is bad; feeding downstream a
  silently-clipped receipt as though it were whole is worse. Also surfaced as a notice.
* **`web_fetch` / `web_search` still re-run.** The three re-executed fetches in the reproduction still
  re-execute, which is the issue's own reading of them: idempotent reads whose cost is latency and
  tokens. `web_search` is excluded by the `Reach::Money` carve-out rather than by its group — it is
  declared `EffectGroup::Spend` for billing and is a read.
* **Checkpointed resume remains the complete answer, and remains unavailable.** `run_resumable`
  installs a no-op observer and a process-local in-memory checkpointer and takes no cancellation token,
  so it composes with neither the per-node progress trail (#371) nor the stop signal (#398), and an
  approval arriving after a restart would find no checkpoint. Engine work in a vendored crate, exactly
  as #438 and #496 both recorded. This guard is forward-compatible: keyed on the node, and it withdraws
  to nothing when the ledger is empty.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — ran as CI does, both lanes: `--locked --no-deps`
      default, and `--locked --no-deps --features openhuman,tinycortex`. Both clean.
- [x] `cargo build --all-targets` — ran as `cargo check --locked --all-features --all-targets` plus
      `cargo check --locked --features openhuman,tinycortex --all-targets`. Both clean. (`check`, not
      `build`: nothing here is a linker-visible change, and the gated lane's test binaries link in CI.)
- [x] `cargo test` — `cargo test --locked --features openhuman,tinycortex --lib`: **3610 passed, 0
      failed, 3 ignored.** Needs `RUST_MIN_STACK=16777216` locally, or `harness::brain::tests::a_cancelled_hand_off_returns_the_card_to_todo_as_a_cancellation`
      overflows the default thread stack — an **untouched** module, and a local artifact rather than a
      code defect.
- [x] `npm ci`, `npm run typecheck`, `npm run build` — all clean. `npx vitest run`: 11 new tests pass.
      57 failures in `tour-resume` / `connection-registry` / `desktop-bridge` are **pre-existing** —
      verified identical with this branch stashed.

### Every new test was proven to fail with its fix reverted — 11 of 11

Each fix was reverted individually and the named test re-run:

|---|---|
| `performed_ledger` drops the lineage union | `an_outward_call_is_made_once_across_two_gates` |
| `continuation_input` stops threading the ledger | `an_outward_call_is_made_once_across_two_gates` |
| `without_ledger` stops stripping the reserved key | `the_outward_ledger_is_not_part_of_a_gates_identity` |
| `replay_performed` becomes a no-op | `a_recorded_node_is_rewritten_to_replay`, `a_recorded_expression_string_is_not_evaluated` |
| the invoker loses its replay arm | `the_replay_sentinel_is_answered_without_a_grant_and_reaches_nothing` |
| the `Money` carve-out and the `GET` arm | `reads_are_not_recorded_whatever_they_cost` |
| the truncation guard | `an_oversized_result_is_refused_rather_than_clipped` |
| the fan-out guard | `a_fan_out_is_refused_and_surfaced` |
| `approvalAction` loses the workflow-tool rung | `names the tool rather than the mechanism` |
| `payloadLines` dumps the raw resume payload | `shows the call's arguments…`, `hides the engine's resume payload…` |
| `awaitingCount` forgets the gates | `counts a parked gate as awaiting…`, `counts parked gates and parked reports together` |

**One test initially failed to fail, and was rewritten.** The first version of
`an_outward_call_is_made_once_across_two_gates` **passed** with the lineage union reverted: the ledger
was surviving anyway, because a card copies the paused run's whole trigger input and that input already
carried the key. The union only decides anything when a later run performs a *new* outward call, since
a non-empty ledger **replaces** rather than merges. The test now does exactly that, and fails on the
first node's entry when the union is reverted. Recording it because a test that cannot fail is worse
than no test.

**No test drives a real outward call end-to-end through a whole lineage.** Two independent blockers,
both pre-existing: the SSRF `url_guard` refuses loopback for `http_request` regardless of the company's
allowlist (already recorded in `runner.rs`'s cancel test, which had to use an agent node for the same
reason), and no wired `tool_call` slug is outward, so there is nothing to count at a transport. It is
covered at each seam instead — including the **real** `WorkflowToolInvoker`, constructed with zero
grants, answering the sentinel and still refusing an ungranted `shell` in the same test, which is what
proves the arm sits above the fail-closed check without widening anything.

**The run-history JSX branch has no component test.** Its logic (`awaitingCount`, `runTone`) is covered
and revert-proven; the markup around it is not. This repo has no component-test harness for that view.

## Documentation

No spec doc changes. The reasoning lives at the code it governs: `src/workflows/replay.rs`'s module docs
carry the seam argument, the two rules and every limit above, beside the tests that pin them.

Closes #846

### DIFF
--- a/frontend/src/lib/language.ts
+++ b/frontend/src/lib/language.ts
@@ -266,11 +266,48 @@ const TOOL_LABELS: Readonly<Record<string, string>> = {
  * left alone so no other surface shifts under this change.
  */
 export function approvalAction(a: ApprovalSummary): string {
+  // Issue #846: a paused workflow gate is named by the CALL it is stopping, not
+  // by the mechanism that stopped it. "Continue a paused workflow" is true of
+  // every one of these cards and therefore tells an operator nothing — it is
+  // #372's "Glob" complaint, one surface over: the chat path was fixed by #375
+  // and this path was not, because its label came from `EFFECT_LABELS` (which
+  // keys on the effect kind, and the kind here is always `workflow.approve`)
+  // rather than from the tool.
+  //
+  // The tool is on the wire whenever the host could classify the node's call.
+  // When it could not — an authored gate on a step that calls nothing — the
+  // glossary entry below is still the honest answer, so this promotes and never
+  // hides.
+  const workflowTool = workflowGateTool(a);
+  if (workflowTool) return toolAction(workflowTool);
   return (
     labelFor(EFFECT_LABELS, a.kind) ??
     labelFor(TOOL_LABELS, a.kind) ??
     (a.agent ? "Use one of its tools" : "Do something that needs your sign-off")
   );
+}
+
+/** The effect kind a paused workflow gate parks as — mirrors
+ * `WORKFLOW_APPROVE_KIND` in `src/runtime/workflow_resume.rs`. */
+const WORKFLOW_APPROVE_KIND = "workflow.approve";
+
+/**
+ * The tool a paused workflow gate is stopping, when the host named one
+ * (issue #846) — `null` for every other kind, and for a gate whose node makes
+ * no classifiable call.
+ *
+ * Reads `payload.tool`, which the host writes for a policy-raised gate (#460)
+ * and, since #846, for an authored one too. Absence is meaningful and is
+ * preserved as such: an **older host** omits the key entirely, and this then
+ * returns `null` so the card renders exactly the pre-#846 line rather than an
+ * empty label.
+ */
+function workflowGateTool(a: ApprovalSummary): string | null {
+  if (a.kind !== WORKFLOW_APPROVE_KIND) return null;
+  const payload = a.payload;
+  if (payload == null || typeof payload !== "object" || Array.isArray(payload)) return null;
+  const tool = (payload as Record<string, unknown>).tool;
+  return typeof tool === "string" && tool !== "" ? tool : null;
 }
 
 /**
@@ -324,7 +361,10 @@ export function payloadLines(a: ApprovalSummary): PayloadLine[] {
   if (typeof payload !== "object") return [{ label: "value", value: renderValue(payload) }];
   if (Array.isArray(payload)) return [{ label: "items", value: renderValue(payload) }];
 
-  const entries = Object.entries(payload as Record<string, unknown>);
+  const entries =
+    a.kind === WORKFLOW_APPROVE_KIND
+      ? workflowGateEntries(payload as Record<string, unknown>)
+      : Object.entries(payload as Record<string, unknown>);
   if (entries.length === 0) return [];
 
   // Preferred ordering for the kinds whose argument names we know. Unlisted
@@ -339,6 +379,66 @@ export function payloadLines(a: ApprovalSummary): PayloadLine[] {
     .sort(([a1], [b1]) => rank(a1) - rank(b1))
     .map(([label, value]) => ({ label, value: renderValue(value) }));
 }
+
+/**
+ * The payload of a paused workflow gate, as the lines an operator needs
+ * (issue #846).
+ *
+ * A `workflow.approve` payload is not a tool call's arguments — it is the host's
+ * **resume record**, and every other card's rule ("show the payload, it is what
+ * you are consenting to") produces exactly the wrong thing here. What an
+ * operator saw was `input: {"items":[{"json":{}}],"port":null}`: the engine's
+ * seed payload, verbatim, as the sole description of a decision.
+ *
+ * So this promotes the call's own arguments to the top level, where every other
+ * card carries them, and drops the machinery. The dropped keys are dropped for
+ * stated reasons, not for tidiness:
+ *
+ * * `input` — the resume payload. It is engine seed data, it is the thing that
+ *   read as a description and was not one, and it also carries an `approvals`
+ *   list that accumulates down a lineage: an operator seeing
+ *   `{"approvals":["fetch_bbc","fetch_espn"]}` reads it as a card covering all
+ *   of them, which it is not. (Consolidating several gates into one card is
+ *   real and is issue #842 — not something to imply by accident here.)
+ * * `delivered`, `performed` — the #438/#846 ledgers. They say what will NOT
+ *   happen again, which is a property of the mechanism and is already stated in
+ *   prose by `note`.
+ * * `content` — rendered in full by its own block (`WorkflowContentReview`,
+ *   #596); repeating it as one clamped JSON line would be strictly worse.
+ * * `note` — prose, rendered as prose elsewhere on the card, not as a
+ *   `key: value` pair in a monospace block.
+ *
+ * Everything else survives, including any key a **newer host** adds that this
+ * console has never heard of. Dropping unknown keys would make an old console
+ * silently hide new information, which is the failure mode this whole file is
+ * written against; the denylist is closed and the allowlist is not.
+ */
+function workflowGateEntries(payload: Record<string, unknown>): [string, unknown][] {
+  const args = payload.args;
+  const argEntries: [string, unknown][] =
+    args != null && typeof args === "object" && !Array.isArray(args)
+      ? Object.entries(args as Record<string, unknown>)
+      : [];
+  const rest = Object.entries(payload).filter(([key]) => !WORKFLOW_GATE_HIDDEN.has(key));
+  // The call first, then where and what stopped it. An operator decides on the
+  // call; the node id is how they find it afterwards. This order survives
+  // `payloadLines`' sort because no `PAYLOAD_KEY_ORDER` entry exists for this
+  // kind — every rank is equal, and the sort is stable.
+  return [...argEntries, ...rest];
+}
+
+/** Payload keys of a `workflow.approve` card that are machinery, not the
+ * decision — see {@link workflowGateEntries} for why each one is here. */
+const WORKFLOW_GATE_HIDDEN: ReadonlySet<string> = new Set([
+  "input",
+  "delivered",
+  "performed",
+  "content",
+  "note",
+  // Unwrapped one level above, so keeping it would print the same arguments
+  // twice — once readably and once as a JSON blob.
+  "args",
+]);
 
 /** Which arguments lead the preview, per tool. Presentation only. */
 const PAYLOAD_KEY_ORDER: Readonly<Record<string, string[]>> = {
--- a/frontend/src/views/workflows/RunHistoryPanel.tsx
+++ b/frontend/src/views/workflows/RunHistoryPanel.tsx
@@ -16,7 +16,7 @@ import type {
 } from "@/api/workflows";
 
 import { failedNodeOf } from "./graph";
-import { pendingCount, relativeTime, runTone, undeliveredCount } from "./run-health";
+import { awaitingCount, relativeTime, runTone, undeliveredCount } from "./run-health";
 
 /** Badge styling per delivery outcome. A report that did NOT go out must not
  * look like one that did — `denied` and `failed` are the two an operator has to
@@ -82,7 +82,9 @@ export function DeliveryRows({ deliveries }: { deliveries: DeliveryReport[] }) {
 export function LastRunChip({ run }: { run: WorkflowRunOutcome }) {
   const tone = runTone(run);
   const undelivered = undeliveredCount(run.deliveries);
-  const pending = pendingCount(run.deliveries);
+  // Issue #846: gates and parked reports together. The chip said "Manual run"
+  // and a green dot for a run whose first node was still waiting on a person.
+  const awaiting = awaitingCount(run);
   return (
     <Badge
       variant="outline"
@@ -111,8 +113,8 @@ export function LastRunChip({ run }: { run: WorkflowRunOutcome }) {
             ? " stopped"
             : undelivered > 0
             ? ` · ${undelivered} not delivered`
-            : pending > 0
-              ? ` · ${pending} awaiting approval`
+            : awaiting > 0
+              ? ` · ${awaiting} awaiting approval`
               : ""}
       <span className="text-muted-foreground">· {relativeTime(run.atMillis)}</span>
     </Badge>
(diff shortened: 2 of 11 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Pushed `30747c26` — **test fixture only, no source change.**

The `Console` lane went red on `npm run typecheck:unit`, which I had not run locally (I ran `typecheck`, which covers `src` but not `test/unit`). My new fixture built its `DeliveryReport` rows from an invented shape — `target: null` where the wire says `target?: string`, plus a `reason` field that does not exist — and `WorkflowRunOutcome` was force-cast with `as` rather than satisfied.

Vitest strips types rather than checking them, so it went green locally and red in CI. That is precisely the blind spot `typecheck:unit` was added for (#434), so the step did its job on me.

Now typed against the real `DeliveryReport` / `WorkflowRunOutcome` with no casts. I have since run the whole Console lane locally — `typecheck`, `typecheck:e2e`, `typecheck:unit`, `build`, `assert-design-tokens.sh`, `assert-md-line-cap.sh` — all clean, and the 11 tests still pass.

No Rust changed, and none of the revert-proofs in the PR body are affected.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
**CI is green, including `Rust (openhuman, tinycortex)`** (27m46s) — the lane that matters here, since everything in `src/workflows/` is behind the `openhuman` feature and the default lanes compile none of it. `Rust`, `Rust (mongodb)`, `Console`, `Console E2E`, `Console E2E (live brain)` and `Desktop` all pass.

**CodeRabbit has not reviewed this PR.** Requested three times over ~50 minutes; every response was a rate-limit notice on the shared account, with no review content:

- 16:22 — `Review limit reached`, next review in 35 minutes
- 16:22 — `Review rate limited` (reply to my `@coderabbitai review`)
- 17:11 — `Review rate limited` again, after the stated window had passed

So there are **no CodeRabbit findings outstanding** — not "none found", but *none produced*. Flagging that explicitly so a green checks list is not mistaken for a clean automated review. Worth a re-trigger by whoever picks this up if you want that pass before merging.

Human review is the one that matters for this diff anyway, and the part I would point at first is the `http_request` → `tool_call` **node-kind rewrite** in `replay_performed`. It is safe on the argument that both kinds share the `{ json, text, raw }` envelope and `envelope::wrap` is a pure function of the recorded value — but that is the load-bearing assumption in the change, and the one I would want a second reader on.

Not merging or approving.

oxoxDev (approved):
Correcting the issue in **both** directions, and having the correction widen the work, is the part that makes this trustworthy:

- the citation was wrong (#438 was closed by #496, not #624), which changes the diagnosis from "the ledger regressed" to "the ledger's reach was narrower than the exposure";
- the issue's own example — a `send`/`publish`/`repo_publish` node after two gated steps — **is not reachable today**, because a workflow `tool_call` reaches only the four families the invoker wires and every slug in them is `EffectGroup::Other` bar a billed read;
- and the thing that **is** reachable is one the issue did not name: an `http_request` node with a mutating method, arbitrary address, every company, right now.

Shipping the guard for the unreachable class anyway — "the guard that must exist *before* a send-capable namespace is wired, or wiring one silently re-opens #438 on the day it lands" — is the right call. That is the failure mode this repo has hit repeatedly: a constraint that holds today, is not written down as a guard, and quietly stops holding when something adjacent changes.

The root cause is stated precisely and matches what #777 established one layer over: a paused tinyflows run is **settled, not suspended**, so resuming is necessarily a re-run from the trigger. Given that, replaying a recorded result is the only correct answer — you cannot decline a call a node already made, so the continuation must not make it again.

**Why the translation and not the invoker** is the strongest paragraph. `ToolInvoker::invoke` receives `(slug, args, conn)` — no node id, no run state — which `gate.rs` already records as why the approval gate could not live there, and the same fact rules out a node-keyed skip inside it. And the alternative is worse for a reason easy to miss: keying on `(tool, args)` "would miss every send whose body an upstream agent node regenerates, which is most of them". Node identity is the only stable key, and the translation is the only place that holds both node ids and the trigger input.

I went looking for whether the replay sentinel is forgeable, since a slug that answers from its own arguments is the obvious thing to abuse. `the_replay_sentinel_is_answered_without_a_grant_and_reaches_nothing` pins the property that matters: it touches no toolbelt, so the worst an author — or an agent using `create_workflow` — can do by invoking it directly is fabricate a value they could have written as a literal anyway. Not an escalation, and the test names it rather than leaving it to be re-derived.

### Question — what does the gate make of the rewritten node?

The node is rewritten to the sentinel **before the run starts**, and the ledger is stripped from the gate's dedupe identity. What I could not establish is what `apply_policy_gates` then classifies that node as on the continuation.

The sentinel is not in the declaration table, so `consequence_of` should fall to the undeclared path — and #660's judgement arm stops an undeclared tool as `StopReason::Undeclared` on the harness side. If the workflow gate reaches the same conclusion, the continuation would park the *replay* of a call the operator has already approved, which is a loop rather than a leak but an unpleasant one.

I suspect this is handled — stripping the ledger from the dedupe identity suggests the gate's view of the node was thought about carefully — but it is the one interaction where the fix meets the guard I reviewed on #660, and it deserves an explicit sentence or a test rather than being inferred from the absence of a failure.

**Before merging:** `MERGEABLE` / `CLEAN`. Worth confirming the gated lane is green before landing given this touches the workflow gate, which is the lane that compiles those tests.

PR #859, opened 2026-08-13
fix(workflows): resolve agent workflow tools against the workflow registry, and let "Build me the workflow" reach the builder

### LINKED ISSUE #845: workflows: the chat agent is never told a builder owns the request, so it refuses a workflow the builder is concurrently creating — and list_workflows returns skills
## 1. The "Build me the workflow" composer mode cannot build a workflow

The channel composer offers a mode toggle: **`Do it once`** / **`Build me the workflow`**. Selecting the second and asking for a workflow produces a refusal, because **no agent behind it has an authoring tool**.

Verbatim, from `#strategy-desk` on staging with `Build me the workflow` selected:

> **I can't build the workflow.** There are exactly 4 installed workflows, all pre-packaged skills (Brand Positioning, Email Campaign, Landing Page, SEO Audit), and my toolset has no capability to create or save a new one. `weekly-aeo-audit` does not exist and I cannot make it exist. I'm not going to claim I saved it.

A DM to a desk agent gives the same result, with the toolset named explicitly:

> The only workflow tools I have are read-only (`list_workflows`, `describe_workflow`, `run_workflow`).

The agents behave **correctly** — both refused to fabricate a saved workflow, and both produced a complete spec instead. The defect is the affordance: a control whose entire purpose is authoring a workflow is wired to agents that cannot author one. It is worse than an absent feature, because the button advertises the capability by name.

**Expected:** either the mode reaches an orchestrator that holds create/save workflow tools, or the control is not offered where nothing can service it.

## 2. `list_workflows` returns *skills*, not workflows

The same reply reveals the agent is reading a different registry from the console.

| Agent's "installed workflows" | Workflows page (6) |
|---|---|
| Brand Positioning | Campaign pipeline |
| Email Campaign | Claude News Blog |
| Landing Page | Content Workflow |
| SEO Audit | Daily Task Digest |
| | Weekly Competitor Analysis |
| | Daily Sports News Blog |

The left column is **exactly** the four entries under `Settings → Skills` ("Installed (4)"). The overlap between the two lists is **zero**.

So an agent asked to run `Campaign pipeline` — a real workflow, visible on the Workflows page, runnable from the UI — will report that it does not exist. `run_workflow` is in the agent toolset, which makes this reachable rather than theoretical.

Independently corroborated by the Copywriter in an unrelated conversation:

> The four installed workflows (Brand Positioning, Email Campaign, Landing Page, SEO Audit) are available to run, but none have been run by me.

**Expected:** `list_workflows` enumerates the workflow registry the Workflows page renders. Skills are a separate concept with their own surface (`Settings → Skills`) and should not be returned under the name "workflows" — the shared word is what makes this silent.

## Environment

Staging, company `marketing--agentic-marketing-agency`, policy tier **Full**, 13 Aug 2026.

M3gA-Mind:
## Summary

Fixes tinyhumansai/opencompany#845 — two defects on the workflow surface. Three commits, each droppable on its own.

### 1. `list_workflows` returned *skills* — `fix(skills)`

Upstream OpenHuman **calls a skill a "workflow"**. Its three skill read tools are named `list_workflows` / `describe_workflow` / `read_workflow_resource`, take a `workflow_id`, and answer with a payload keyed `workflows`. OpenCompany also has workflows — the saved graphs the Workflows page renders and `run_workflow` runs — and `EffectiveSkills::read_tools()` wired upstream's tools in unrenamed, putting both concepts on one belt under one word.

So an agent asked what workflows the company had answered with the four installed *skills* — a set with **zero** overlap with the Workflows page — and an agent asked to run `Campaign pipeline` reported it did not exist. Nothing errored; the shared word is the whole of why it stayed silent. Upstream's own `list_workflows` description ends *"…to inspect (`describe_workflow`) or run (`run_workflow`)"*, i.e. it instructs the model to take a skill out of the skill tree and run it as a workflow.

Fixed by wrapping the three tools at the OpenCompany boundary (`src/harness/skills/naming.rs`), **not** in `vendor/` — "workflow" is upstream's settled word for a skill and is correct everywhere except a host that also has workflows.

The rename covers all four agent-visible surfaces, because leaving any one leaves the bug: the **name**; the **description** (written here rather than delegated); the **argument** (`skill_id`, mapped back before the inner tool runs — a legacy `workflow_id` still resolves, so a model on a stale schema is not stranded); and the **result payload**.

> **The payload rename is the part that actually fixes the reported symptom.** Renaming only the tool leaves the agent reading `{"count":4,"workflows":[…]}` and still reporting "there are exactly 4 installed workflows" — the sentence quoted in the issue. Only top-level keys are re-keyed, so a skill's own description and a bundled file's content are never touched.

**This grants nothing.** Same inner tools, same workspace scoping, same permission level, arguments only re-keyed. An agent that could not reach the workflow registry still cannot — it now says so instead of answering from the wrong list.

> **`policy::consequence` is a fix, not a tidy-up.** The `DECLARED` rows had to follow the rename. An unlisted tool falls through to `undeclared`, whose read-only prefixes include `list` and `read` but **not** `describe` — so an undeclared `describe_skill` would be `Reach::Consequence`: it parks, and interrupts an operator to approve a *local file read*. That is the exact regression `consequence.rs`'s own comment records `describe_workflow` causing before that table existed. `skill_read_tools_are_declared_reads` pins it, and fails if the rows are reverted.

### 2. "Build me the workflow" — `fix(chat)` + `feat(console)`

**The capability exists and works.** The builder pass (#580) does not dispatch a `workflow`-deliverable card to its assignee, because authoring the graph *is* its In-Progress work; it lands a `TaskWorkflowProposal` In Review for a person to apply. On staging it had already produced a proposal for `weekly-aeo-audit` while a desk agent was answering the same message with *"I can't build the workflow … it does not exist and I cannot make it exist."* Both halves behaved correctly and the operator was told the opposite of what was happening.

Three distinct failure modes, three fixes:

**(a) The chat cycle runs on that message in parallel, and was never told.** `CompanyEvent::OperatorMessage` carried no deliverable, so the turn answered blind — and the desk agent answering it holds no authoring tool, correctly refused to pretend otherwise, and said so. Thread the choice onto the event and fold a briefing into the message the brain sees, on exactly `OPEN_WORK_ANNOTATION`'s terms: in-memory only, never journaled, stripped by `operator_words` before anything reasons about what the operator asked for. Missing that strip would be the "thanks!" bug in a new costume — the briefing is several lines of imperative prose containing "workflow", "create" and "build", and `looks_like_work` scores length and work verbs.

**(b) The choice was consulted only on the card-opening branch, which is gated on the triage.** A `workflow` request the classifier read as a question or as chatter dropped the choice silently: no card → no builder pass → nothing built, and no error either. Now the card opens whatever the triage said. The operator reached for a control *named* "Build me the workflow" and pressed it; that is better evidence about this message than a lexical classifier's guess. Titled through `to_title`, so a bypassed card is titled byte-for-byte as a tracked one, and it stays inside the `!confined` guard — a copilot thread still opens nothing.

**(c) DMs had no toggle at all** (`deliverableChoice={active.kind === "channel"}`). Nothing downstream was ever scoped to channels, so the asymmetry lived entirely in that one caller: a DM asking for a workflow went as a `once` card, was dispatched to a desk agent with no authoring tool, and came back a refusal — the DM case in the issue. The rule now lives in `offersDeliverableChoice`, total over `ChannelKind`, so a new kind is a decision someone makes rather than a control that silently fails to appear.

### Security: `CallPath::AuthoredWorkflowNode`'s premise survives

Raised on #776 and worth checking for explicitly, since a reviewer who remembers that concern will look for it.

**None of this grants an agent workflow-authoring power.** `create_workflow` and the #661 admin tools stay **orchestrator-only**; nothing here touches `orchestrator_tools`. The builder pass only ever *proposes* — it courtesy-validates a graph and stamps it on a card, and `create_company_workflow` runs solely from `POST …/workflow-proposal/apply`, which a **person** calls. So a human still authors every workflow node, and the premise that makes `AuthoredWorkflowNode` a silent path is intact. The expensive path — giving agents authoring tools, which *would* invalidate it — is the one this change makes unnecessary.

## API Or Behavior Changes

- **Agent-facing tool rename** (behaviour change, intended): `list_workflows`→`list_skills`, `describe_workflow`→`describe_skill`, `read_workflow_resource`→`read_skill_resource`; arg `workflow_id`→`skill_id`; result keys `workflows`→`skills`, `workflow_id`→`skill_id`. A legacy `workflow_id` argument is still accepted, so an in-flight conversation is not stranded.
- **`CompanyEvent::OperatorMessage` gains `deliverable`**, additive on exactly the `by`/`chat`/`parent` terms: `#[serde(default, skip_serializing_if = "Option::is_none")]`, so every stored event loads and an unset one serializes byte-for-byte as before. No migration.
- A `workflow`-deliverable chat message now (i) always opens a card and (ii) reaches the brain with a briefing appended in memory. `once` and unset messages are byte-for-byte unchanged — pinned by tests on both sides.
- DM composers now show the once-vs-workflow control.
- No HTTP route, GraphQL or storage schema changes.

## Tests

Run from the repo root. The gated lane is the one that matters here — default features skip the whole `harness` module.

- [x] `cargo clippy --all-targets -- -D warnings` — ran as CI does, `--no-deps`, **both lanes**: default, and `--features openhuman,tinycortex`. Clean.
- [x] `cargo build --all-targets` — ran with `--features openhuman,tinycortex`.
- [x] `cargo test` — ran as `cargo test --features openhuman,tinycortex --tests`: **3642 + 10 + 1 + 11 + 2 passed, 0 failed**. Default lane also green (2360 + 10 + 1). Also `scripts/ci/assert-feature-lanes.sh` (0 owed a lane).
- [x] Frontend, mirroring the console job: `npm ci`, `npm run typecheck`, `npm run typecheck:unit`, `npm test`. New test passes.

**Revert checks.** Fleet standard is to prove each new test fails with its fix reverted.

- Part 1 (skills naming): reverted all three sites independently — **10 of 13 fail**. The 3 that do not are unit tests of helper functions that do not exist pre-fix (`replace_whole_words`, `rewrite_prose`, `rename_top_level_keys`); they cannot fail on a revert and are not claimed as regression coverage.
- Part 2: **all 4 fail** with their fixes reverted, checked one at a time — the triage bypass, the `operator_words` strip, and the injection. `the_builder_briefing_reaches_the_brain_through_run_cycle` covers the **wiring** specifically and fails when only the call site is deleted, which the injection's own unit test cannot catch.

**Known pre-existing failures, reproduced on a clean `origin/main` with none of my changes applied:**

- `harness::brain::tests::a_cancelled_delegated_card_records_no_artifact` — debug-build stack overflow. Passes under `RUST_MIN_STACK=16777216`, which is what the runs above used.
- `frontend`: 57 failures in `tour-resume`, `desktop-bridge`, `connection-registry` — `TypeError: window.localStorage.clear is not a function` (jsdom environment). Verified identical on stashed-clean main.
- `frontend/pnpm-lock.yaml` is out of sync with `package.json` (`@fontsource-variable/geist-mono` is declared but absent from the lock), so `pnpm install --frozen-lockfile` fails on main today. **Not touched here** — CI installs with `npm ci` against `package-lock.json`, which is fine. Flagging it rather than bundling an unrelated lockfile regen.

## Documentation

Module-level docs carry the reasoning, which is where this belongs — the failure was one of naming and wiring, not of a documented contract:

- `src/harness/skills/naming.rs` — a full module header on the collision, why the rename happens at this boundary and not in `vendor/`, why it covers four surfaces, and the `consequence.rs` hazard.
- `BUILDER_ANNOTATION` in `src/runtime/cycle.rs` — what the briefing is for and the staging evidence, alongside `OPEN_WORK_ANNOTATION`, which it parallels.
- The `deliverable` field on `CompanyEvent::OperatorMessage`, and `offersDeliverableChoice` in `frontend/src/views/chat/model.ts`.

No `docs/` page described either behaviour, so none needed updating.

### DIFF
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -55,6 +55,7 @@ import {
   dmChannelId,
   findChannel,
   firstChannel,
+  offersDeliverableChoice,
   resolveDmChannelId,
   toggleReaction,
   type DecidedApproval,
@@ -855,10 +856,11 @@ export function ChatView({
               placeholder={`Message ${channelTitle(channel)}`}
               disabled={sending}
               onSend={(text, deliverable) => void send(text, deliverable)}
-              // Only the company-channel composer offers "do it once" vs "build
-              // me the workflow" (issue #580): a channel line can open a board
-              // card, so the once-vs-workflow choice belongs at that prompt box.
-              deliverableChoice={active.kind === "channel"}
+              // Channel *and* DM composers offer "do it once" vs "build me the
+              // workflow" (issues #580, #845) — see `offersDeliverableChoice`,
+              // which owns the rule. Only the thread and copilot composers below
+              // go without.
+              deliverableChoice={offersDeliverableChoice(active.kind)}
             />
           </div>
 
--- a/frontend/src/views/chat/MessageComposer.tsx
+++ b/frontend/src/views/chat/MessageComposer.tsx
@@ -24,10 +24,16 @@ interface Props {
   /**
    * Show the once-vs-workflow control (issue #580), opt-in per composer.
    *
-   * Only the company-channel composer asks for it — a channel line can open a
-   * board card, so "do it once" versus "build me the workflow" belongs there.
+   * The channel and DM composers ask for it — either can open a board card, so
+   * "do it once" versus "build me the workflow" belongs at both prompt boxes.
    * The thread and copilot composers never carry it, so their `onSend` stays a
    * plain `(text)` and their wire shape is unchanged.
+   *
+   * DMs were omitted when #580 landed (issue #845). Nothing downstream was
+   * scoped to channels — the chat route reads `deliverable` off the payload
+   * whatever thread it came from — so a DM asking for a workflow was sent as a
+   * `once` card, dispatched to a desk agent holding no authoring tool, and came
+   * back as a refusal. The control was the only part missing.
    */
   deliverableChoice?: boolean;
 }
--- a/frontend/src/views/chat/model.ts
+++ b/frontend/src/views/chat/model.ts
@@ -242,6 +242,35 @@ export function channelTitle(channel: Channel): string {
   return channel.kind === "dm" ? channel.name : `#${channel.name}`;
 }
 
+/**
+ * Whether this chat target's composer offers "Do it once" / "Build me the
+ * workflow" (issues #580, #845).
+ *
+ * True for every real chat target — a channel line and a DM can each open a
+ * board card, and the card is what routes a `workflow` request to the builder
+ * pass. It is deliberately a total function over [`ChannelKind`] rather than an
+ * inline `kind === "channel"`, so adding a kind is a decision someone makes here
+ * instead of a control that silently fails to appear.
+ *
+ * Not every composer asks: the thread and copilot composers pass nothing and
+ * keep their plain `(text)` `onSend`. A thread reply continues a message that
+ * already made this choice, and a copilot line is about one graph rather than a
+ * request to the company.
+ *
+ * #580 shipped the control on channels only. Nothing downstream was ever scoped
+ * to channels — the chat route reads `deliverable` off the payload whatever
+ * thread it came from — so the asymmetry lived entirely in the caller, and a DM
+ * asking for a workflow had no way to say so. It went as a `once` card, was
+ * dispatched to a desk agent holding no authoring tool, and came back a refusal.
+ */
+export function offersDeliverableChoice(kind: ChannelKind): boolean {
+  switch (kind) {
+    case "channel":
+    case "dm":
+      return true;
+  }
+}
+
 /* ---- senders ---- */
 
 export type SenderKind = "you" | "company" | "agent" | "system";
--- a/frontend/test/unit/chat-deliverable-choice.test.ts
+++ b/frontend/test/unit/chat-deliverable-choice.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from "vitest";
+
+import type { ChannelKind } from "@/views/chat/model";
+import { offersDeliverableChoice } from "@/views/chat/model";
+
+/**
+ * Issue #845: which composers offer "Do it once" / "Build me the workflow".
+ *
+ * #580 shipped the control on channel composers only. Nothing downstream was
+ * ever scoped to channels — `client.chat` carries `deliverable` off the payload
+ * whatever thread it came from, and the chat route reads it the same way — so
+ * the asymmetry lived entirely in the caller. A DM asking for a workflow had no
+ * way to say so: it went as a `once` card, was dispatched to a desk agent
+ * holding no workflow-authoring tool, and came back as a refusal. Reported
+ * verbatim on staging as "The only workflow tools I have are read-only".
+ */
+describe("offersDeliverableChoice", () => {
+  it("offers the choice on a DM — the gap this closes", () => {
+    expect(offersDeliverableChoice("dm")).toBe(true);
+  });
+
+  it("still offers it on a channel, unchanged from #580", () => {
+    expect(offersDeliverableChoice("channel")).toBe(true);
+  });
+
+  /**
+   * The rule is total over `ChannelKind` rather than an inline
+   * `kind === "channel"`, so a new kind is a decision someone makes in the
+   * function instead of a control that silently fails to appear. This pins that
+   * every kind the type admits has an answer.
+   */
+  it("answers for every channel kind", () => {
+    const kinds: ChannelKind[] = ["channel", "dm"];
+    for (const kind of kinds) {
+      expect(typeof offersDeliverableChoice(kind)).toBe("boolean");
+    }
+  });
+});
--- a/src/brain/echo.rs
+++ b/src/brain/echo.rs
@@ -187,6 +187,7 @@ mod test {
                     text: "hi".into(),
                     by: None,
                     chat: None,
+                    deliverable: None,
                 }]),
                 &host,
             )
--- a/src/brain/hosted/test.rs
+++ b/src/brain/hosted/test.rs
@@ -112,6 +112,7 @@ fn operator_request() -> CycleRequest {
             text: "hi".into(),
             by: None,
             chat: None,
+            deliverable: None,
         }],
         event_seqs: Vec::new(),
         compressed_history: Vec::new(),
@@ -537,6 +538,7 @@ async fn e2e_operator_message_drives_tool_call_and_gated_send_dm() {
             text: "how are we doing".into(),
             by: None,
             chat: None,
+            deliverable: None,
         }])
         .await
         .unwrap();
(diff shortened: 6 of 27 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Both defects are diagnosed to the mechanism rather than the symptom, and the second framing is the one I would keep:

> Both halves behaved correctly and the operator was told the opposite of what was happening.

Nothing was broken. The builder pass landed a `TaskWorkflowProposal` In Review exactly as #580 intends, and a desk agent said *"I can't build the workflow … it does not exist and I cannot make it exist."* A composition that lies while every component is right is the hardest class to find, because each half passes its own review.

**The namespace collision is the same shape one layer down.** Upstream calls a skill a "workflow", OpenCompany has workflows, and `read_tools()` wired both onto one belt under one word — so an agent asked what workflows the company had answered with four installed skills, a set with zero overlap with the Workflows page, and nothing errored. Upstream's own description telling the model to take a skill out of the tree and *run* it as a workflow is the detail that turns a naming clash into a behavioural one.

Fixing it at the **OpenCompany boundary rather than in `vendor/`** is right: "workflow" is upstream's settled word for a skill and is correct everywhere except a host that also has workflows. Patching vendored code for a local collision is how a submodule bump becomes a merge conflict forever.

**The payload rename being called out as the part that actually fixes it** is the insight. Renaming only the tool leaves the agent reading `{"count":4,"workflows":[…]}` and still reporting "there are exactly 4 installed workflows" — the model reads the key, not the signature. Re-keying only top level, so a skill's own description and a bundled file's content are untouched, is the right boundary for that.

Stating **"this grants nothing"** explicitly — same inner tools, same workspace scoping, same permission level, arguments only re-keyed — is worth having in the description for a change that renames things an agent invokes.

**I verified the `consequence.rs` claim, and it is a fix rather than a tidy-up.** `READ_ONLY_PREFIXES` is `read`, `list`, `get`, `search`, `recall`, `query`, `peek`, `inspect`, `view`, `memory_recall`, `memory_search` — **`describe` is not there**. So an undeclared `describe_skill` really would land `Reach::Consequence` and park, interrupting an operator to approve a local file read. Following the rename through `DECLARED`, and pinning it with `skill_read_tools_are_declared_reads`, is what stops the rename shipping a gate regression alongside a naming fix.

**0 major. 1 note. Approving.**

### Note — the instance is fixed, the class is not

`describe` missing from `READ_ONLY_PREFIXES` is exactly why this PR had to touch the declaration table. Declaring the three tools closes it for them, and the test keeps it closed. But the next tool named `describe_*` that ships **undeclared** hits the identical trap — parking on a read, for the same reason, with nothing to catch it.

I do not think the fix is to add `describe` to the list. That list is a guess applied to tools nobody declared, and #660's judgement arm deliberately treats undeclared tools as stop-worthy *because* the guess is unreliable — widening it makes the guess more confident in a place the codebase has decided not to trust it. The real answer is the one this PR takes: declare your tools.

What would help is a sentence at `READ_ONLY_PREFIXES` recording that `describe` is deliberately absent and why, with a pointer to this PR. Right now the omission reads as an oversight, and the next person to hit it will either add it — quietly loosening a fail-closed default — or rediscover this whole argument from scratch.

**Before merging:** `MERGEABLE`, lanes still reporting. Three commits each droppable on its own is a good property to have stated; worth keeping them separate through the merge rather than squashing, since the `consequence.rs` row change is the one a future bisect would want to find on its own.

Issue #905, opened 2026-08-16
main is red: the 7cc3f451 vendor bump pins openhuman/tinyagents to a pair that does not compile


oxoxDev:
## main does not compile

`upstream/main` at `7cc3f451` fails to build with the gated feature set:

```
$ cargo check --features openhuman,tinycortex
error[E0432]: unresolved import `crate::core::event_bus`
  --> vendor/openhuman/src/openhuman/memory/guard/audit.rs:33:18
error[E0433]: cannot find `event_bus` in `core`
   --> vendor/openhuman/src/openhuman/memory/binding.rs:398:26
error[E0433]: cannot find `event_bus` in `core`
   --> vendor/openhuman/src/openhuman/memory/binding.rs:399:30
error: could not compile `openhuman` (lib) due to 4 previous errors
```

Reproduced on a **pristine checkout** of `7cc3f451` with submodules initialised — no local changes, no branch of mine involved. CI agrees: the `CI` and `Deploy tenant image to staging` runs on `7cc3f451` both failed at 15:18.

## Cause

`7cc3f451` — "chore(vendor): update openhuman and tinyagents submodules", committed directly to `main` (message says `Auto-committed-on: macbook`, "to include local changes").

```
vendor/openhuman   a221052e → 8774fe4a
vendor/tinyagents  c6a5f24b → 5e026cd8
```

Two things about the new pins:

1. **`vendor/openhuman` moved to `8774fe4a`, which is dated 2026-08-09** — a merge commit on openhuman's `main` that predates the code depending on it. At that pin, `event_bus` exists only at `src/openhuman/web_chat/event_bus.rs`; there is no `src/core/event_bus`, which is what `memory/guard/audit.rs` and `memory/binding.rs` import.
2. **`vendor/tinyagents` moved *backwards*** — `5e026cd8` is the pin main carried yesterday, before `c6a5f24b`.

So the bump appears to have pinned to whatever the committer's local checkouts happened to be on, rather than forward to a coherent pair.

Worth noting the previous pin `a221052e` is **not resolvable** in a fresh clone of the openhuman repo, so the tree before this commit may also have been pinned to an unpushed local commit. Either way, the pair on `main` now does not build.

## Impact

Every open PR that merges current `main` inherits a red gated lane, and the staging deploy is failing. This is not diagnosable from a PR's own diff, so it will read as "my branch broke it" to whoever looks next.

## Suggested fix

[...]

PR #907, opened 2026-08-16
fix(console): make restart, lifecycle, and tour controls take effect immediately

senamakel:
## Why

Three controls in the console did not take effect when you pressed them. Each
had a different cause, so this fixes three distinct things rather than one.

## 1. "Restart required" was a dead end

`InferenceSection` showed a persistent notice saying the company had to be
restarted before agents would stop echoing — with no way to restart it. That is
worst for the operator most likely to see it: a hosted tenant's unit of restart
is its container, and the control plane has no button for it.

The in-place rebuild from #290 already existed and already ran automatically
inside `PUT …/inference`. What was missing was any way to ask for it on its own,
so `rebuild_company` had **no HTTP surface** — it was reachable only as a side
effect of saving the form.

- **New route** `POST …/inference/restart`, admin-scoped like the save it
  mirrors, since rebuilding swaps the brain every agent thinks with.
- **New button** "Restart now" on the notice, admin-only so a member is not
  shown a control that can only 403.

It deliberately does not gate on `restart_required` first. Rebuilding a healthy
company is a no-op from the operator's point of view — same journal, same parked
approvals, same grants — so gating would buy nothing, would introduce a race
between the check and the rebuild, and would refuse the case most worth
allowing: recovering a company whose state the console is reading wrongly.

The resulting status is read off the **successor** runtime, so a rebuild that
lands on the same brain still reports honestly. A host with no rebuilder wired
fails with the message naming the process restart that would work, and the
console shows that rather than claiming a success that changed nothing.

## 2. Lifecycle buttons lagged the click by two round trips

`run()` awaited the action, then fired `feed.refresh()` un-awaited while
clearing `busy` immediately. Between those, the operator saw re-enabled buttons
still offering "Pause" on a company they had just paused — which reads as a
click that did not register, and invites a second one.

Now a `pending` state is set *before* the await, so the buttons move on the
click, and the refresh is awaited so `pending` is not dropped while the
authoritative read is still in flight. Cleared in `finally`, so a rejected
action reverts to the host's answer instead of leaving the console asserting a
state that was refused.

Note `resume` → `running`, not `"resumed"`: the success toast and the lifecycle
value are different vocabularies, and collapsing them would leave the Resume
button on screen forever. There is a test for exactly that.

## 3. "Replay tour" waited on a network download

`Joyride` is `lazy()`, so the chunk only began downloading *after* the click.
The toast said the tour was starting and then nothing visible happened until it
arrived — long enough on a cold cache to read as a broken button.

The fetch now starts on pointer-enter/focus of the button, so by the time
`Suspense` asks for the module it has usually resolved. `import()` caches its
module promise, so repeated calls are lookups rather than requests. Lazy loading
is kept — most operators never run the tour.

## Commands run

- `cargo test` — 2418 passed, 0 failed, 1 ignored
- `cargo fmt -p opencompany -- --check` — clean
- `cd frontend && pnpm typecheck` — clean
- `cd frontend && pnpm test` — 725 passed across 64 files

(`cargo fmt --all` reports pre-existing diffs in the untouched `vendor/openhuman`
submodule, hence `-p opencompany`.)

## Tests added

11 new. Backend: the restart swaps the registered runtime and leaves it
accepting work rather than parked in the quiesce window; calling it twice is not
a refusal; a host that cannot rebuild says so **and is still serving**, since a
failed rebuild that left a company quiesced would turn a cosmetic dead end into
an outage. Frontend: the route the client calls including the single-company
alias, that the console follows the returned status, and the lifecycle mapping
with `resume` → `running` pinned explicitly.

## Behaviour changes

- New route `POST …/inference/restart` (admin authority required).
- Lifecycle buttons update optimistically and revert on failure.
- No API change for the tour — a prefetch only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -408,7 +408,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
 dependencies = [
  "axum-core",
- "axum-macros",
  "base64 0.22.1",
  "bytes",
  "form_urlencoded",
@@ -456,17 +455,6 @@ dependencies = [
  "tower-layer",
  "tower-service",
  "tracing",
-]
-
-[[package]]
-name = "axum-macros"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7aa268c23bfbbd2c4363b9cd302a4f504fb2a9dfe7e3451d66f35dd392e20aca"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.119",
 ]
 
 [[package]]
@@ -1334,6 +1322,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs"
+version = "6.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3e8aa94d75141228480295a7d0e7feb620b1a5ad9f12bc40be62411e38cce4e"
+dependencies = [
+ "dirs-sys 0.5.0",
+]
+
+[[package]]
 name = "dirs-sys"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1782,6 +1779,18 @@ checksum = "f0d8a4362ccb29cb0b265253fb0a2728f592895ee6854fd9bc13f2ffda266ff1"
 dependencies = [
  "opaque-debug",
  "polyval",
+]
+
+[[package]]
+name = "git2"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ddddbf932745a6be37109b6112d3ee09696106f848449069d3a57bba937ab82e"
+dependencies = [
+ "bitflags 2.13.1",
+ "libc",
+ "libgit2-sys",
+ "log",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


senamakel:
## PR babysitter status

**Inspected head:** `3be7326d84ce06ab82dbe32478031e5933cfbf1b`

**Local validation (all pass on this head):**
- `cargo test` — 2418+ passed, 0 failed, 1 ignored
- `cargo fmt -p opencompany -- --check` — clean
- `cd frontend && pnpm typecheck` — clean
- `cd frontend && pnpm test` — 725/725 passed across 64 files

**Review:** tinysweeper approved (no blocking findings). CodeRabbit review is rate-limited (no findings posted). No unresolved review threads.

**CI status — blocked by pre-existing infra breakage, not by this PR's diff:**
`Rust`, `Rust (mongodb)`, `Rust (openhuman, tinycortex)`, and `Desktop` all fail during the `actions/checkout` submodule step itself, before any of this PR's code builds:

```
fatal: No url found for submodule path 'vendor/openhuman/app/src-tauri/vendor/tauri-cef' in .gitmodules
fatal: run_command returned non-zero status while recursing in the nested submodules of vendor/openhuman
```

This is inherited from `main`'s vendor bump at `7cc3f451` (`vendor/openhuman` pinned to `8774fe4a1`), which is red on `main` itself — confirmed via `gh run list --branch main`, both runs on `7cc3f451` failed at 15:18 UTC, before this branch existed. Already tracked in **#905** ("main is red: the 7cc3f451 vendor bump pins openhuman/tinyagents to a pair that does not compile"), which documents the same pin also failing to compile the gated `openhuman,tinycortex` feature set once past the checkout step. Fixing it requires repinning `vendor/openhuman`/`vendor/tinyagents` to a coherent, buildable pair on `main` — currently ~30 commits ahead on openhuman's `main` with API changes, out of scope for this PR to absorb. `Console`, `Actions pinned`, and the tinysweeper checks are green.

**Next action:** land the vendor-pin fix for #905 on `main` (separately), then re-run CI on this branch — no further changes needed on this PR's own diff.

Issue #881, opened 2026-08-15
workflows: a node whose publish parked still reports ok, and passes its own apology downstream


oxoxDev:
## What happens

When a workflow node's agent has its `publish_artifact` parked for approval, the node still records `status: "ok"` and stores the agent's *apology* as the node output. Verbatim, from a `feature_pipeline` run's `spec` node:

> "The spec is complete and sitting in my sandbox at `specs/notes-export-spec.md`, but the hand-off is blocked again. **What's blocked:** `publish_artifact` — it requires approval under policy `opencompany-approval` … The tool did not run; the file is not yet a deliverable you can open."

The graph then proceeds to the next node with that text as its input, and the run completes with every node green.

## Why it matters

"Green" stops meaning "produced its output". The downstream node consumes prose about a blockage instead of the artifact it was supposed to build on, and the run's terminal state claims success for a pipeline that delivered nothing.

## Expected

A node whose deliverable did not publish is not `ok`. Either park the run (surfacing the approval, see the companion issue on run-level approval visibility) or mark the node blocked and stop the branch, rather than passing the apology downstream.

## Repro

Run any multi-agent workflow on a company whose policy parks `publish_artifact`, then read `GET /api/v1/companies/<id>/workflows/runs/<rid>/output`.

---
*Found in a parity audit of the five target product use cases against `main` @ `7009e829`, run locally and on the `smoke1` + `marketing` staging tenants with real inference and real tool calls.*

oxoxDev:
Evidence for the `gate`/`security` question raised during planning — I pulled the completed run's output.

The `gate` condition node's input items are the **frontend engineer's apology prose**, verbatim:

> "Both publish calls are blocked. I'll report honestly and completely. **Client changes implemented; all handover blocked pending approval.** … All three files are written in my sandbox, but **none could be published** — `publish_artifact` is blocked by `opencompany-approval` on this auto desk."

So the condition node was branching on prose about a blockage rather than on the structured result it expects, and `security` was skipped downstream of that. That does not prove the gate is broken — but it does mean its verdict on this run carries no information, which is a concrete argument for stopping the branch at the blocked node rather than letting the apology travel.

PR #918, opened 2026-08-17
fix(build): set the fields main's own literals are missing

oxoxDev:
## Summary

`main` has not built under `openhuman,tinycortex` since #881/#880 and the desk-tools overlay landed. Two literals were left behind by shapes that changed under them:

```
error[E0063]: missing fields `approvals` and `blocked_nodes` in initializer of `ports::types::CompanyEvent`
   --> src/server/ops/workflows.rs:3699

error[E0063]: missing field `overlay_desk_tools` in initializer of `ports::types::CompanyRecord`
   --> src/workflows/blocked_node_test.rs:123
```

Two lines. Empty values throughout — the first is a legacy-error journal append and the second a test fixture, and neither has blocked nodes, approvals or per-desk tool overrides to record.

## Problem

The fourth union break in this repository in about a day, and the same shape every time: a struct gains a field on one branch while another branch adds or keeps a literal of it. The two never conflict textually, so GitHub reports `MERGEABLE`, both are green individually, and the merge is red.

What makes this one worse than the earlier three is where it landed. Neither file is reachable from the default feature set, so the `Rust` lane stays green and only `Rust (openhuman, tinycortex)` reports it. `main` therefore merged red and **stayed** red — [run 32001078448](https://github.com/tinyhumansai/opencompany/actions/runs/32001078448) on tip `d681a665` — while looking healthy to anything reading the default lane.

Every open pull request that merges `main` inherits it. It is currently the only red lane on #851, in files that PR never touches.

## Solution

Set the fields.

## Submission Checklist

- [x] `cargo check --locked --features openhuman,tinycortex --all-targets` — the lane that was failing
- [ ] Full gated test run — deferred to CI here; the failure was a build error, and the build is what this fixes.

## Impact

Unblocks the gated lane for `main` and for every open pull request. No runtime behaviour changes: one journal append gains two empty collections, one test fixture gains an empty map.

## Related

Surfaced while resolving conflicts on #851, whose gated lane fails on exactly these two errors in files it does not touch.

Worth considering separately: four breaks of this shape in a day is a process gap rather than four coincidences. A merge queue, or requiring the gated lane on `main`, would catch the class — `MERGEABLE` demonstrably does not mean "compiles".

### DIFF
--- a/src/server/ops/workflows.rs
+++ b/src/server/ops/workflows.rs
@@ -3706,6 +3706,8 @@ mod tests {
                         cancelled: false,
                         notices: Vec::new(),
                         board: Vec::new(),
+                        blocked_nodes: Vec::new(),
+                        approvals: Vec::new(),
                     },
                 )
                 .await
--- a/src/workflows/blocked_node_test.rs
+++ b/src/workflows/blocked_node_test.rs
@@ -127,6 +127,7 @@ fn record() -> CompanyRecord {
         lifecycle: "running".to_string(),
         overlay_agents: Vec::new(),
         overlay_desk_members: Vec::new(),
+        overlay_desk_tools: std::collections::BTreeMap::new(),
         overlay_desk_order: Vec::new(),
         overlay_desks: Vec::new(),
         overlay_workflows: Vec::new(),


oxoxDev:
Superseded by #917, which landed the identical fix a few minutes earlier — same two literals, same empty values, same reasoning. That is why this now shows CONFLICTING: the lines it edits no longer exist in the form it expects.

Same root cause diagnosed independently, so nothing here is wrong; it is just a parallel-session collision. Closing to keep the queue honest rather than rebasing it to a no-op.

For the record, both diagnoses agreed on the important part: `cargo build` passes on the broken main and only `cargo test` fails, which is why the deploy lane stayed green while CI went red.

PR #940, opened 2026-08-17
fix(console): wait for a channel's history before calling it empty (#934)

### LINKED ISSUE #934: chat: reloading a DM shows the "start of your direct message" empty state before history arrives
## Summary
Hard-reloading a DM thread that has history renders the *empty* state — "This is the start of your direct message with Product Manager" — with no messages, for long enough to be mistaken for data loss. Switching to another DM and back repopulates the full thread, so nothing is actually lost.

## Repro
1. Open a DM with existing history.
2. Reload the page.
3. The thread renders as brand new.

## Expected
Show a loading state while history is being fetched. The "start of your direct message" copy is a positive claim that the thread is empty, and it should only render once we know that is true.

## Severity
Cosmetic in effect, but alarming in practice — I read it as lost conversation history until I checked, and a user has no way to check.

## Environment

CodeGhost21:
## Summary

Fixes #934 — reloading a DM with history rendered "This is the start of your direct message with …" and no messages until `chat/history` answered. Nothing was lost, but the copy is a positive claim that the thread is empty, and an operator has no way to tell a false one from real data loss.

## Root cause

The console could not tell the two states apart. `Transcripts` is `Record<string, ChatMessage[]>`, and the timeline reads it as:

```ts
const messages = channel ? (transcripts[channel.id] ?? []) : [];
```

That `??` collapses *nobody has asked the host yet* into *the host says this channel is empty*. `ChannelIntro` then renders the claim off `items.length === 0`, so it printed the second while the first was true.

The window is wide enough to notice because DM history is three round-trips deep on a cold load — `listDesks` → `await listTeam` → one `getChatHistory` per channel — while `ChatView` gates its render on `desks` only, and knows nothing about hydration happening in the shell above it. Switching away and back "fixed" it only because hydration had completed in the background; the shell never unmounts, so the return trip renders state that had already arrived.

An empty answer is a legitimate final state, so this cannot be derived from the transcripts alone. It needs a fact that was not being recorded.

## The fix

`HistoryHydration` + `historyReady()` in `views/chat/model.ts` — per-channel status, owned by `AppShell` because `AppShell` owns the fetches. A channel is marked `loading` **before** its request goes out, and `ready` on every settle path: a normal answer, the `entries.length === 0` early return, and the `.catch`.

| State | May the intro claim "this is the start"? |
|---|---|
| `loading` | No — the request is in flight. |
| `ready` | Yes. Settled, including an empty answer or an outright failure. |
| No entry, pass unfinished | No. `ChatView` resolves its own desk list independently of the shell's, so it can paint a channel a moment before the shell marks it. |
| No entry, pass finished | Yes. Nothing will hydrate it — a console-only teammate, or a host with no `chat/history`. A permanent spinner would be a worse lie than the one this prevents. |

While a channel is pending with no rows, `MessageTimeline` renders skeleton rows under `role="status"` and drops the claim sentence. The avatar and title stay — those say *where you are*, which is not an assertion about history. Locally-sent rows still render immediately; only the claim of emptiness waits.

The `hydration` prop is optional and defaults to "untracked", so a `ChatView` mounted without a shell behind it renders exactly as before rather than spinning on a pass that is never coming.

## Commands run locally

- `npm run typecheck` / `typecheck:unit` / `typecheck:e2e` — clean
- `npx vitest run` — 804 passed
- `npx playwright test` against a host built from this branch — **104 passed, 14 skipped, 0 failed** (the whole suite, not only the chat specs)

**The new e2e specs were confirmed load-bearing**: with the fix disabled, both fail; restored, both pass. A timing bug observed by luck passes whether or not it is fixed, so the DM's `chat/history` is held open on a hand-released latch rather than merely being slow — the loading state is asserted while the request is provably in flight, and the settled state after it provably is not. Every other channel's history is left alone, which also proves the hold is per channel and not a pane-wide freeze.

## Tests

- `frontend/test/unit/chat-history-hydration.test.ts` — 7 cases over `historyReady()`, one per state the console actually reaches.
- `frontend/test/e2e/chat-history-loading.spec.ts` — the race itself, which only exists in a browser with a real host behind it. The second spec covers the inverse: a genuinely empty DM **does** get the copy once the host answers, so the fix cannot regress into a spinner that never resolves without a test noticing.

## One regression this caught, and fixed

The full run surfaced a real break in `chat-channel-membership.spec.ts`: the skeleton is a second `role="status"` region, so the spec's unqualified `getByRole("status")` started resolving to two elements and failed on strict mode.

That locator meant *the unknown-channel notice* and only worked because the notice happened to be the sole live region on the page. It now says so — including the closing `toHaveCount(0)`, which asserted the notice was gone and would otherwise have started asserting the skeleton was gone too. Second commit, kept separate from the fix.

## Behaviour changes

None to any API. One deliberate rendering change: the intro's claim sentence is replaced by the channel's plain purpose line while history is pending. The intro still renders above a full thread as it always has — standing over rendered messages it reads as "you are at the beginning", which is true; what made it a lie was standing alone over nothing.

## Notes

Docs updated in `frontend/src/views/chat/README.md` (new "Empty, or not answered yet" section).

One adjacent bug with the same root — `ChatView` gates on `desks` but not on its own roster load, so a `#/chat/dm:…` deep link can briefly compute `unknownChannel` truthy before `members` lands — is deliberately **not** included here, since it is a different symptom from what #934 reports. Happy to file it separately.

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -71,7 +71,10 @@ import {
   channelIdForThread,
   deskFromDto,
   dmChannelId,
+  HISTORY_UNSTARTED,
   type DecidedApproval,
+  type HistoryHydration,
+  type HistoryStatus,
   type Transcripts,
 } from "@/views/chat/model";
 import { Conversation } from "@/views/Conversation";
@@ -292,6 +295,11 @@ export function AppShell({
   // mounts and unmounts `ChatView` per route, so component-local state there
   // would be discarded on every trip away from Chat and back.
   const [transcripts, setTranscripts] = useState<Transcripts>({});
+  // How far each channel's history rehydration has got. Kept beside
+  // `transcripts` rather than inside it because an empty transcript is a
+  // legitimate final answer, and the timeline has to tell that apart from not
+  // having asked yet before it prints "this is the start of…" (issue #934).
+  const [hydration, setHydration] = useState<HistoryHydration>(HISTORY_UNSTARTED);
   // Host thread id → chat channel id, for every channel this company has.
   // Resolved by the desks/roster effect below, which already works the pairing
   // out to hydrate each channel and used to throw it away — leaving the shell
@@ -481,6 +489,10 @@ export function AppShell({
     // channels that no longer exist, and start the unread floor again so the
     // incoming company's rehydrated history isn't counted as news.
     setChatChannelByThread({});
+    // Another company's channels are another namespace here too, and a status
+    // carried over would let the incoming company's channels claim to be
+    // settled before anything has asked about them.
+    setHydration(HISTORY_UNSTARTED);
     setFirstDeskChannelId(null);
     setLastViewedChannel({});
     setUnreadSince(Date.now());
@@ -536,20 +548,35 @@ export function AppShell({
     // untouched), but not for a DM: the channel id is the console-local
     // `dmChannelId`, while the thread id `getChatHistory`/`chat` read is the
     // roster agent id (see `ChatView`'s `send`) — so this takes both.
+    const markHistory = (channelId: string, status: HistoryStatus) =>
+      setHydration((h) => ({ ...h, byChannel: { ...h.byChannel, [channelId]: status } }));
+
     const hydrateChannel = (channelId: string, threadId: string) => {
+      // Marked before the request, not after: the gap between "this channel
+      // exists" and "its history is in flight" is precisely the window the
+      // timeline used to fill with the empty-channel copy.
+      markHistory(channelId, "loading");
       client
         .getChatHistory(threadId, company)
         .then((entries) => {
-          if (cancelled || entries.length === 0) return;
+          if (cancelled) return;
+          if (entries.length === 0) {
+            // An empty answer is still an answer, and the only thing that ever
+            // makes the "start of your direct message" copy true.
+            markHistory(channelId, "ready");
+            return;
+          }
           const hydrated = fromHistory(entries);
           setTranscripts((t) => {
             const known = new Set((t[channelId] ?? []).map((m) => m.id));
             const fresh = hydrated.filter((m) => !known.has(m.id));
             return fresh.length === 0 ? t : { ...t, [channelId]: [...fresh, ...(t[channelId] ?? [])] };
           });
+          markHistory(channelId, "ready");
         })
         .catch(() => {
           /* host without `/chat/history`, or offline — channel stays empty */
+          if (!cancelled) markHistory(channelId, "ready");
         });
     };
 
@@ -586,6 +613,9 @@ export function AppShell({
         setFirstDeskChannelId(chatDesks[0]?.id ?? null);
         chatDesks.forEach((d) => hydrateChannel(d.id, d.id));
         roster.forEach((m) => hydrateChannel(dmChannelId(m), m.id));
+        // Every channel this pass will hydrate now has a status, so a channel
+        // with none is one nothing is coming for.
+        setHydration((h) => ({ ...h, discovered: true }));
       })
       .catch(() => {
         // Host without `/desks`, or offline — keep the static default
@@ -596,6 +626,7 @@ export function AppShell({
         setChatChannelByThread(channelMap(fallbackDesks, []));
         setFirstDeskChannelId(fallbackDesks[0]?.id ?? null);
         fallbackDesks.forEach((d) => hydrateChannel(d.id, d.id));
+        if (!cancelled) setHydration((h) => ({ ...h, discovered: true }));
       });
 
     return () => {
@@ -1198,6 +1229,7 @@ export function AppShell({
               onReply={() => void feed.refresh()}
               transcripts={transcripts}
               setTranscripts={setTranscripts}
+              hydration={hydration}
               onSendStart={onSendStart}
               onSendEnd={onSendEnd}
               liveStepsByThread={liveStepsByThread}
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -55,10 +55,13 @@ import {
   dmChannelId,
   findChannel,
   firstChannel,
+  historyReady,
+  HISTORY_UNTRACKED,
   offersDeliverableChoice,
   resolveDmChannelId,
   toggleReaction,
   type DecidedApproval,
+  type HistoryHydration,
   type Transcripts,
 } from "./chat/model";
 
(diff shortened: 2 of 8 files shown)


CodeGhost21:
## E2E test report — console against a live backend

Run against a real OpenCompany host built from this branch, with the console bundle built from this branch and served by it (`OPENCOMPANY_CONSOLE_DIR`). No mocked backend: every spec drives the actual HTTP API, and the two `#934` specs additionally intercept `GET .../chat/history` in the browser to control its timing.

### Environment

| | |
|---|---|
| Host | `target/debug/opencompany serve` — default features, offline echo brain |
| Company | `companies/e2e_harness` |
| Console | `frontend/dist` built from this branch |
| Base | `http://127.0.0.1:8099` (loopback, no `OPENCOMPANY_PUBLIC_URL`, so the magic-link `dev_code` is echoed and the suite can sign in) |
| Auth | Real magic-link flow — `POST auth/request` → `POST auth/verify`, session cookie shared across specs |

### Results

**Full Playwright suite — 104 passed, 14 skipped, 0 failed** (2.0m)

The 14 skips are the live-brain-gated specs (issue #467), which need a host built with `--features openhuman,tinycortex,mcp` plus the mock-brain fixture. Not related to this change.

**Chat specs in full — 27 passed, 3 skipped** (29.3s)

```
✓ chat-approval-line.spec.ts:109  the line recording a decision is visible in a real channel
✓ chat-approval-line.spec.ts:132  a decision the host refuses says so in the same channel
✓ chat-approval-line.spec.ts:150  with no channel ever opened, the decision still reaches the first desk
✓ chat-channel-membership.spec.ts:140  #369 each desk counts and lists its own members, not the company
✓ chat-channel-membership.spec.ts:170  #369 a DM reads as two people, not as the whole company
✓ chat-channel-membership.spec.ts:184  #369 a host with no desks surface still shows the whole roster
✓ chat-channel-membership.spec.ts:197  #370 a deep link never flashes a channel the company doesn't have
✓ chat-channel-membership.spec.ts:225  #370 an unknown channel opens the first one and says so
✓ chat-channel-membership.spec.ts:245  #370 a broken /desks is a retryable error, not invented channels
✓ chat-channel-membership.spec.ts:268  #485 a desk channel links out to its own desk on the org chart
✓ chat-channel-membership.spec.ts:291  #485 a DM has no desk to manage
✓ chat-channel-membership.spec.ts:308  #485 a fallback desk offers no link to a desk the host doesn't have
✓ chat-channel-membership.spec.ts:321  the send path and the thread id it addresses are undisturbed
✓ chat-dispatch-marker.spec.ts:100   a settled dispatch marks the channel its card was raised in — and only that one
✓ chat-history-loading.spec.ts:108   a reloaded DM waits for its history instead of calling itself new     ← #934
✓ chat-history-loading.spec.ts:142   a DM the host reports empty does say it is the start                  ← #934
✓ chat-inline-approval.spec.ts:94    a request raised in a channel appears in that channel, and only there
✓ chat-inline-approval.spec.ts:115   the Approvals page still shows everything, including the inline one
✓ chat-inline-approval.spec.ts:124   deciding inline detaches, and the continuation lands in the same channel
✓ chat-inline-approval.spec.ts:180   declining inline says so in the thread rather than leaving it stalled
✓ chat-inline-approval.spec.ts:201   a decision made on the Approvals page settles the inline card, with no reload
✓ chat-inline-approval.spec.ts:235   a parked approval raises its card live, without a reload
✓ chat-live-events.spec.ts:148       a reply the console never asked for lands in the open channel, with no reload
✓ chat-live-events.spec.ts:167       a reply to a channel you are not on leaves an unread badge, and opening it clears it
✓ chat-live-events.spec.ts:189       a turn sent from the composer renders exactly one company bubble
✓ chat-live-events.spec.ts:215       a running turn shows its tool rows in the channel
✓ chat-to-card.spec.ts:52            any message on a desk thread can be added to the board
```

**Unit — 804 passed across 76 files.** **Typecheck — `typecheck`, `typecheck:unit`, `typecheck:e2e` all clean.**

### The two `#934` specs are load-bearing

A timing bug observed by luck passes whether or not it is fixed, so this was verified both ways rather than asserted:

| `historyPending` | `chat-history-loading.spec.ts` |
|---|---|
| disabled (forced `false`) | **both specs fail** — `getByText(/This is the start of your direct message/)` resolves to 1 element while the request is provably still in flight |
| restored | **both pass** |

The DM's `chat/history` is held open on a latch the test releases by hand, so the loading state is asserted while the request is demonstrably unanswered and the settled state after it demonstrably is not. Every other channel's history is left untouched, which also demonstrates the hold is per channel rather than a pane-wide freeze.

The second spec covers the inverse — a genuinely empty DM **does** get the "start of your direct message" copy once the host answers — so the fix cannot silently regress into a spinner that never resolves.

### A regression this run caught

`chat-channel-membership.spec.ts:225` failed on the first full run. The skeleton introduces a second `role="status"` region, so the spec's unqualified `getByRole("status")` began resolving to two elements and failed strict mode:

```
strict mode violation: getByRole('status') resolved to 2 elements:
  1) <p role="status" …> aka …filter({ hasText: "#does-not-exist isn't a" })
  2) <div role="status" aria-busy="true" …> aka …filter({ hasText: 'Loading messages…' })
```

That locator meant *the unknown-channel notice* and only worked because the notice happened to be the sole live region on the page. Fixed in `a3c2c8a1`, kept as its own commit, including the closing `toHaveCount(0)` which asserted the notice was gone and would otherwise have started asserting the skeleton was gone too.

### Scope of this report — read before trusting it

- These are **Playwright-driven** browser runs, not hand-clicked ones. Real Chromium, real host, real API; the interaction is scripted rather than manual.
- I attempted a manual pass driving Chrome directly against this host and **could not complete it**, so nothing here comes from that. It was blocked on sign-in, in a way unrelated to this change (details below). No manual result is claimed.
- The 14 skipped specs were not exercised; they need the live-brain feature build.

### Unrelated observation, possibly worth its own issue

While setting up the manual browser pass, browser-issued requests authenticated inconsistently against the single-company alias. With an identical, freshly minted session cookie:

| Request | curl | Chrome |
|---|---|---|
| `POST /api/v1/company/auth/verify` | 200 | 200 |
| `GET /api/v1/company/auth/me` | 200 | 200 |
| `GET /api/v1/companies/e2e-harness-co` | 200 | 200 |
| `GET /api/v1/company/desks` | 200 | **401** |
| `GET /api/v1/companies` | 200 | **401** |

Routes with the company explicit in the path authenticate; routes on the `/api/v1/company/*` alias do not — but only for browser-issued requests. The console gates its signed-in state on `GET /api/v1/companies`, so it renders the sign-in screen indefinitely.

Ruled out: stale or duplicate cookies (server-side logout, then one clean sign-in), cookie-jar state (repeated on `localhost`, a distinct origin with an empty jar), a proxy in the path (identical directly against the host), `Origin` / `Referer` / `Sec-Fetch-Site` (curl with each still 200), and host state (a fresh curl session succeeds concurrently). The UI's own "Use it now" dev affordance redeems successfully and still bounces back to the form.

Untouched by this PR either way — flagging it because it makes the console unusable in a plain browser against a loopback single-company host, which is the ordinary local-development path.

Issue #874, opened 2026-08-15
bug: tool-slugs advertises granted-but-unwired tools, so the copilot authors nodes that cannot run


oxoxDev:
## Summary

`GET …/workflows/tool-slugs` answers from the company's grants alone and ignores whether the deployment actually wires the tool. So on a deployment where a granted tool cannot run, the endpoint still lists it, the create-time copilot grounds on it, and the graph it authors fails at the first node.

Seen on staging (`marketing` tenant), which reports:

```
/workflows/tool-slugs → {"slugs":[ … ,"web_search"]}
/capabilities         → searchGranted: true, searchInBuild: true, searchCredentialConfigured: false
```

`web_search` is offered, and a saved workflow using it failed at run time:

```
error: "harness error: capability error: tool_call 'web_search' is not available in company workflows"
nodes: fetch_news=error
```

The wiring-aware list already exists. `workflow_effective_tool_slugs` (`src/company/workflow_create.rs:909`) intersects grants with the deployment's wiring and its doc-comment says it is "the tools the create-time copilot may ground on: catalogue, company grant, and deployment wiring all agree". The endpoint (`src/server/ops/workflows.rs:1710-1728`) still calls the grant-only `workflow_callable_tool_slugs` (`:886`).

Worth keeping the distinction the code already draws: **create validation stays permissive** on purpose, so an operator can author now and wire the provider later. This is about what the console and the copilot are *told is available*, which is the narrower set.

## Reproduction

1. Run a company that grants `search` on a deployment with no managed search credential (`searchCredentialConfigured: false`).
2. `GET /api/v1/company/workflows/tool-slugs` → `web_search` is in the list.
3. Ask the create-time copilot for a workflow that needs a web search; it authors a `tool_call` node on `web_search`, because that is what it was grounded on.
4. Run the workflow → the first node fails with `tool_call 'web_search' is not available in company workflows`.

Observed on `marketing.staging.opencompany.work` (`claude_news_blog`, run `019ffb1ee6c3-00000000001e`, 2026-08-13).

## Expected behavior

- `tool-slugs` serves the effective set — granted **and** wired on this deployment — so the copilot cannot ground on a tool that has no chance of running here.
- Where a tool is granted but unwired, the answer says so rather than silently dropping it, so an operator reading the console can tell "not allowed here" from "allowed, not configured yet" — the same distinction `refusal_for` already makes at run time (`no managed search backend is configured` vs `capability tier filtered`).
[...]

PR #943, opened 2026-08-17
Report a missing path as not-found, not as a missing app

### LINKED ISSUE #924: tools: grep and read_skill_resource are offered to agents but fail as "App unavailable" in a hosted tenant
## Summary
Two tools are advertised in the agent's catalogue but fail immediately when called on a hosted tenant, and the failure is classified as `MissingApp` — so the agent is told *"The app or program needed for this action isn't available. Install or open the app, then try again."* There is no app to install; this is a server tenant.

## Observed
Asked the Product Manager for a short feature spec. The turn ran 20 steps, **2 failed**:

```
Grep                 App unavailable   pattern=[Uu]sage · path=Agents/... · max_matches=50    1ms
Read Skill Resource  App unavailable   skill_id=feature-spec · relative_path=references/spec.md  2ms
```

Both fail in 1–2ms, i.e. they are rejected before doing any work.

## Why it matters
`read_skill_resource` failing means **an agent running the Feature Spec skill cannot read that skill's own reference material.** It fell back to imitating the structure of an unrelated existing note. Skills are the mechanism we tell users to author playbooks with, so a skill that cannot read its own `references/` is a hole in that feature, not a cosmetic error.

`grep` failing costs the agent a step and pushes it toward slower `workspace_search` loops.

Both are present in the catalogue and the policy table:
- `src/policy/consequence.rs` — `d("grep", …)`, `d("read_skill_resource", …)`
- `src/harness/skills/naming.rs` — `READ_SKILL_RESOURCE_TOOL`

## Expected
Either wire the tools for the hosted tenant, or stop advertising them there. If a tool genuinely cannot run in this deployment, the failure copy should say that instead of `ToolFailureClass::MissingApp`'s desktop-oriented "install or open the app" remediation, which is unactionable on a server.

## Environment

M3gA-Mind:
## Summary

**The title invites the wrong fix.** Nothing here is unwired, and nothing needed wiring.

`grep` and `read_skill_resource` are both **on the belt**, both **declared in
`policy::consequence`**, and both **refused correctly**. Only the *verdict* on their refusal was
wrong. A reviewer arriving expecting a wiring change will not find one.

### What actually happens

`ENOENT` is the same operating-system error whether a binary is not installed or a file is not
there. `src/openhuman/tools/status/ops.rs`'s `classify()` keys on message text alone, and **rule 5
carries a bare `"no such file or directory (os error 2)"` needle alongside six needles that name a
program** (`"command not found"`, `"executable not found"`, …). So every `ENOENT` — file or binary —
lands on `ToolFailureClass::MissingApp`, and `describe()` renders it:

> The app or program needed for this action isn't available. **Install or open the app, then try again.**

On a server tenant there is nothing to install. That misclassification, at
`ops.rs:137-151`, is the defect.

The two reported calls reach it from different directions and both were behaving correctly:

1. **`grep` with `path=Agents/…`.** `Agents/` is the **company note tree** — a *different surface*,
   served by `workspace_read` / `workspace_search`, scoped by the store rather than the filesystem.
   The `files`/`docs` tools are sandboxed `workspace_only` to the agent's **own** workspace
   (`build.rs:1131` `file_tools`, `build.rs:1118` `workspace_security`), so they genuinely cannot see
   it. `validate_path` joins the sub-path onto that sandbox, `canonicalize` fails, and the tool
   returns `Failed to resolve path 'Agents/…': No such file or directory (os error 2)`. **Refusing a
   path it cannot see is the correct behaviour** — the sandbox is the point.

2. **`read_skill_resource(feature-spec, references/spec.md)`.** Its `symlink_metadata` pre-check
   returns the same `ENOENT`. `companies/agentic_software_company/skills/feature-spec/` ships
   **only `SKILL.md`**; no bundled skill in this repo has a `references/` directory at all. The file
   does not exist, so refusing is correct here too.

### Two corrections to the issue's premises

- **Not hosted-specific.** It reproduces wherever the path is absent, which is why the regression
  test below is a plain unit test and needs no tenant.
- **"An agent running the Feature Spec skill cannot read that skill's own reference material"**
  overstates it: that skill has no reference material. The tool was not denied a file it should have
  had.

### On the read-only-prefix hypothesis

Raised while triaging, and it **does not apply** — recording it because the premise looked right.
The concern was that `consequence.rs`'s read-only prefixes cover `list`/`read` but not
`describe`/`grep`, so an undeclared local read can park for approval. Both tools here are
**explicitly declared** — `d("grep", …)` at `consequence.rs:415`, `d("read_skill_resource", …)` at
`:587` — and an explicit declaration means the prefix rule is never consulted. The prefix comments at
`:407` and `:571-578` are earlier fixes for exactly that hazard. Neither tool parks; both fail in
1–2 ms.

## API Or Behavior Changes

**New `TurnStepFailure` variant: `NotFound`** (`"not_found"` on the wire), rendered
**"Not found"** in the console.

A failed step from a tool that resolves a path *in this process* and cannot spawn a program now
reports `not_found` with the cause *"The file or folder this action asked for does not exist."*,
where it previously reported `missing_app` with *"The app or program needed for this action isn't
available."* No other verdict changes.

`refine_missing_app` returns the class and the cause **together**, so the chip and the sentence beside
it cannot disagree.

**Keyed on the tool, not on the message — deliberately.** Sniffing for the six program-specific
needles cannot separate the two cases, because `Command::new(program)` on a missing binary yields the
*bare* `ENOENT` with none of them (`install_tool.rs:198`; `workspace_state.rs:203` spawning `git`).
A text-based rule would relabel a genuinely missing `git` as a missing file. Which tool ran is the
signal that actually distinguishes them.

**`PATH_ONLY_TOOLS` is under-inclusive by design.** A path tool absent from it keeps today's
behaviour rather than gaining a new wrong one, so the failure mode of drift is a stale label, never a
false `not_found` on a real missing program.

## Does this belong on #874, and would a coverage test have caught it?

**No, and no — plainly:**

**A test asserting "every advertised slug resolves" would have passed today and would not have caught
this.** Both tools resolve. `every_registered_tool_is_declared`
(`src/harness/mod.rs:6160`) already enforces declaration coverage exhaustively against the live belt,
and both tools satisfy it. Neither an advertisement check nor a declaration check can see a bug whose
entire content is *which failure class a correct refusal is reported as*.

What would have caught it is an invariant one level down, and this PR adds it:
`every_path_tool_on_the_belt_is_listed` derives the set from the same constructor the belt uses
(`build::file_tools`) rather than transcribing names, and fails when the belt grows a path tool
`PATH_ONLY_TOOLS` does not name — the "exhaustive by construction rather than by memory" shape
`every_registered_tool_is_declared` documents.

**This partly undercuts #874's premise as applied here.** #874 is about grants-vs-deployment-wiring
on `GET …/workflows/tool-slugs`, where a granted-but-unwired tool really is advertised and really
cannot run. That is a genuine bug and this PR does not address it. But #924 was filed under the same
"advertised then fails" heading and is **not** an instance of it: nothing was unwired, so the
wiring-aware set #874 proposes would have listed both of these tools too. #924 is narrower and shares
no code with it — different surface, different fix.

## Tests

Run locally, on the lane that actually builds this code:

- [x] `cargo fmt --all -- --check` — PASS
      PASS. `--no-deps` mirrors `ci.yml:679`; an unscoped run drowns in vendored `tinyagents` lints CI
      never sees.
- [x] `cargo build --all-targets` — covered by the clippy and test runs above at
      `--features openhuman,tinycortex --all-targets`.
- [x] `cargo test` — **4084 passed / 0 failed** at `--features openhuman,tinycortex --lib`, and
      **2721 passed / 0 failed** at default features. Needs
      `RUST_MIN_STACK=16777216` (pre-existing; a `harness::brain` test overflows without it, which is
      why `ci.yml:626` sets it).

Frontend: `npm run typecheck` — PASS.

### Each new test proven to fail with the fix reverted

- **`a_path_tools_missing_file_is_not_reported_as_a_missing_app`** — the two reported failures,
  verbatim in the shape their producers emit, driven end to end through the fold. Reverting
  `refine_missing_app` to a pass-through fails it, reproducing the bug exactly:
  `result: Some("The app or program needed for this action isn't available."), failure: Some(MissingApp)`
  for `grep`. It also asserts upstream still calls this `MissingApp`, so the test reports an obsolete
  re-read rather than silently passing if upstream is fixed.
- **`every_path_tool_on_the_belt_is_listed`** — dropping `"grep"` from `PATH_ONLY_TOOLS` fails it:
  `these sandboxed file tools … are not in PATH_ONLY_TOOLS …: ["grep"]`.
- **`not_found_serializes_as_the_snake_case_the_console_indexes_on`** — pins the wire string the
  console keys on.
- Frontend: removing `not_found` from `STEP_FAILURE_LABEL` fails the build with
  `error TS2741: Property 'not_found' is missing … but required in type 'Record<TurnStepFailure, string>'`.

**One test is a guard, not coverage — stating it rather than counting it:**
`a_missing_program_is_still_a_missing_app` **passes with the fix reverted**. It exists to prove the
change does not over-reach — that `shell` / `git_operations` / `apply_patch` keep `missing_app` for
the same `ENOENT` — so it is a non-regression assertion, not evidence the fix works.

### Known gap, not addressed here

**Nothing ties the Rust `TurnStepFailure` to the TypeScript union.** `Record<TurnStepFailure, string>`
forces a *label* to exist, and the serde test pins this variant's string, but no check asserts the two
lists hold the same members. Both sides are updated here; the seam itself stays unguarded.

## Documentation

`src/harness/steps.rs`'s module header gains a bullet for the one re-read verdict, alongside the
existing statements about what a step may carry — the taxonomy is documented there, so a class this
crate reinterprets belongs in the same list. `PATH_ONLY_TOOLS` and `refine_missing_app` carry the
reasoning inline, including why the discriminator is the tool and not the message. Both
`TurnStepFailure` variants document how they differ and why.

No `docs/` change: this alters no route, no launcher behavior, and no `tiny*` integration.

Closes tinyhumansai/opencompany#924

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -67,6 +67,7 @@ export type TurnStepFailure =
   | "unauthorized"
   | "missing_permission"
   | "missing_app"
+  | "not_found"
   | "timeout"
   | "unavailable"
   | "failed";
@@ -115,6 +116,7 @@ export const STEP_FAILURE_LABEL: Record<TurnStepFailure, string> = {
   unauthorized: "Unauthorized",
   missing_permission: "Missing permission",
   missing_app: "App unavailable",
+  not_found: "Not found",
   timeout: "Timed out",
   unavailable: "Service unavailable",
   failed: "Failed",
--- a/src/harness/steps.rs
+++ b/src/harness/steps.rs
@@ -65,6 +65,14 @@
 //! * **Result on failure** is the classifier's plain-language
 //!   [`cause_plain`](oh::tools::status::ClassifiedFailure::cause_plain), or an
 //!   intrinsic tool's own message — never the remote error text.
+//! * **One upstream verdict is re-read** (issue #924). `ENOENT` is the same
+//!   operating-system error whether a binary is not installed or a file is not
+//!   there, and upstream classifies on message text alone, so it calls both
+//!   [`MissingApp`](oh::tools::status::ToolFailureClass::MissingApp) — telling a
+//!   server operator to "install or open the app" when a note is simply absent.
+//!   For a tool that resolves a path in this process there is no app to install,
+//!   so [`PATH_ONLY_TOOLS`] names those and [`refine_missing_app`] reports
+//!   [`TurnStepFailure::NotFound`] instead. Nothing else is re-classified.
 //!
 //! The unit test `planted_secret_never_reaches_serialized_steps` proves it end
 //! to end: a secret planted in a tool's output, its nested arguments, and its
@@ -667,11 +675,13 @@ fn complete(
         None => oh::tools::status::classify(output, false),
     };
 
+    let (failure, cause_override) = refine_missing_app(tool_name, &classified);
+
     Completed {
         status: TurnStepStatus::Error,
         detail,
-        result: failure_result(tool_name, output, &classified),
-        failure: Some(failure_of(classified.class)),
+        result: failure_result(tool_name, output, &classified, cause_override),
+        failure: Some(failure),
         truncated: false,
     }
 }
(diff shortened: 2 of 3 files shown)


M3gA-Mind:
@coderabbitai review

Issue #899, opened 2026-08-16
workflows: an approved agent-internal call has no replay path, so a blocked run can only be re-run


oxoxDev:
## What is missing

When a workflow agent node's internal tool call is parked for approval, approving it mints a grant that **no run re-dispatches**. The operator's only move is to re-run the workflow — and the re-run makes a *fresh* call rather than replaying the one they approved.

## Why #846's ledger does not cover it

`PAYLOAD_PERFORMED` keys replay on a `tool_call` **node's** static slug and arguments, which the graph knows ahead of time. An agent node's internal calls are not in the graph: the model chooses them mid-turn, so there is no static payload to key against and nothing to match a grant to on resume.

This is the missing sibling of #846, on the agent-node side.

## Why it cannot simply resume today

The vendored engine's pause primitive, `NodeControl::Interrupt`, **discards the activation's state update — the node re-runs from the top on resume**. An agent node is not re-enterable: resuming means a fresh 33–89s LLM turn that calls the tool again and parks a *new* card. Approve → re-run → re-park, without converging. The engine already refuses this itself — `finish_agent_run` maps `StopReason::Paused` to "resuming a paused agent is not supported yet".

`#469`'s `ContinuationQueue` is also not the tool: it is keyed on the parking *cycle* and continues one agent turn, and a workflow run has no `CycleHost` on this path.

## What a fix probably needs

Some record of the approved call's payload at the point it was parked, plus a replay path on re-entry that consumes the grant against that exact payload instead of re-asking the model — i.e. the agent-internal analogue of the `PAYLOAD_PERFORMED` ledger, with an at-most-once guarantee.

## Consequence today

Documented as an honest limit of #881, not hidden: a blocked run reports blocked truthfully, and clearing it is a manual re-run. Filed as its own issue rather than a TODO, per the #438 → #846 precedent.

## Related

- #881 — makes a blocked node report blocked instead of `ok` (introduces this limit's visibility)
- #880 — makes a run report the approvals it parked
- #846 — the `tool_call`-node replay ledger this needs a sibling to

PR #945, opened 2026-08-17
fix(console): a paused card names the approvals blocking it, and Resume stops re-running them (#883)

### LINKED ISSUE #883: console: a paused card never says which approvals block it, and Resume re-parks it
## What happens

A paused card shows a **Resume** button and nothing about *why* it is paused. The card is blocked until every approval its turn parked has been decided (the #469 continuation rule: a turn is continued once, when the last decision it was blocked on lands). Neither the count nor the identity of those approvals appears on the card.

The consequence is a loop that looks like a hang:

1. QA's triage turn parks 5 approvals (`shell` ×4, `publish_artifact`).
2. Operator approves the publish — nothing happens. The card sits in Paused.
3. Operator presses **Resume** — the card re-dispatches, the agent re-runs from scratch, parks again, and the queue grows.
4. Only once *all* outstanding approvals for the turn are decided does the card move — in this case straight to `in_review` the instant the last `shell` was denied.

Nothing in the UI distinguishes step 2 (decided some, still waiting) from a stuck card, so Resume is the natural next click and it makes things worse.

## Expected

- The paused card names what it is waiting on: "blocked on 4 approvals" with a link to them.
- Resume is disabled, or warns, while approvals for that turn are still undecided — pressing it re-runs work that will park again.
- Ideally the approvals list can be filtered to one card.

The engine is correct here; this is entirely about the operator being able to see the state it is in. The task detail API already carries the approvals array — the board card does not use it.

---

M3gA-Mind:
## Summary

A paused card carried a **Resume** button and nothing about *why* it was paused. The card is blocked
until every approval its turn parked has been decided — the #469 continuation rule: a turn is
continued once, when the last decision it was blocked on lands. Neither the count nor the identity of
those approvals reached the card.

That makes two different states look identical:

| the operator has | the card shows |
|---|---|
| decided 1 of 5, still waiting on 4 | a Resume button |
| a genuinely wedged card | a Resume button |

Resume is the natural next click from both, and it is the wrong click from the first: it
re-dispatches, the agent re-runs from the top, parks the same calls again, and the queue grows. The
issue's own reproduction is that loop.

> **Branched off `9dc9971a`** — the #941 ledger-board merge, which landed the same day. This is
> written against the **new** board (`LedgerBoard` + `TasksView`'s `renderCard`), not the one the
> issue was filed against. A reviewer comparing to the issue text will find the card component has
> moved underneath it.

### 1. The card names the blocked call — through the function every other surface already uses

This is the part worth reading closely, because it is a **cross-surface consistency** change rather
than "the card now shows some text".

`approvalAction` in `frontend/src/lib/language.ts` is the one place that answers *"what is this
approval asking for"*. It is what #372/#375 fixed the **chat card** with when it was showing raw
runtime identifiers ("Glob", "Shell"), and what **PR #854** extended when the **workflow card** was
labelled `Continue a paused workflow` — true of every one of those cards and therefore an answer to
nothing. The board card was the third surface with the same defect and no fix.

So the blocked row calls the same function. All three now say **"Fetch a web page"** about the same
approval, instead of three different things — and a glossary entry added for one of them lands on all
three. Nothing new was written here; the wording layer is reused as-is.

- **One** blocked call is quoted in full.
- **Several** are counted (`Blocked on 4 approvals`), because five tool names is not something to
  read on a Kanban card. The count plus a **Review** link is, and the page it links to lists them.

### 2. Resume is disabled, not hidden — with the reason on it

Stated explicitly because it is a deliberate choice a reviewer will otherwise ask about.

**Hiding the button would re-create the bug.** The ambiguity being fixed is a card that gives the
operator no way to tell "blocked, and here is on what" from "stuck". A card whose only action
silently disappeared answers neither — it just moves the confusion from *which click* to *where did
the button go*. The operator has to be able to see that Resume exists and is the wrong click **right
now**.

So the button stays, `disabled`, carrying its reason in `title`
(*"Blocked — decide its approvals first; resuming re-runs the work from the start"*) for a pointer,
with the blocked row above it carrying the same reason for everyone else.

The rule lives in **exactly one place** — `taskApprovalBlock` — read by the button's `disabled` and
by the row that explains it, so the card cannot say "blocked on 4 approvals" beside a button that
dispatches anyway. There is deliberately **no second guard inside `resume()`**: `onResume` is wired
only to that button, so a copy of the rule there would be a branch nothing can execute and therefore
a branch nothing keeps true.

### 3. The queue can be narrowed to one card

`#/approvals/<taskId>`, using the unvalidated second hash segment every other view already takes
(`#/workflows/<id>`, `#/team/<agentId>`). Linkable, survives a refresh, honours Back.

A filtered queue with nothing left in it renders **"This card is clear"**, deliberately *not* the
existing "Nothing is waiting on you" — that one is a claim about the **whole company**, and it would
be said here while other cards' approvals sit one click away unread. It is also the wrong answer to
the question the operator arrived with.

### 4. The detail row names the call too

`AwaitingApprovalRow` already said *that* the card was waiting and for how long (#468). It now also
says **what**, which is what the MVP spec asks for: *"clicking the task shows what approval is
needed."*

Its **source of truth is unchanged**: `detail.approvals`, the host's own read, still decides *whether*
the card is waiting — because `approval_owner` resolves ownership with an attempt-level `run_id` key
this side cannot see. The feed only supplies the words, matched in **by approval id**. An approval the
host counts and the feed has not caught up on is still counted; it goes unnamed for one poll rather
than disappearing. It also only names the call when there is exactly one *and* every pending row was
matched — naming "the" blocked call while a second sits beside it would say one decision clears the
card when two do, which is the precise mistake this issue is about.

### Where the data comes from — and why there is no new wire field

The board reads `…/tasks`, whose card projection carries no approvals, and opening every card to find
out would be N reads per 4s poll — the cost `TaskCard::output`'s own docs record as the reason it
rides the board read.

Neither was needed. **The shell already polls `GET …/approvals` every 5s** for the sidebar badge, and
since #333 every entry carries `task: {link: "task", id}`. `CycleHostImpl::park` stamps that link on
**every** park path, so a card-dispatched approval always has it. The board joins against the feed it
was already paying for.

The consequence worth noting: the board and the Approvals page now read **one list**, not two that can
disagree.

Two states are skipped rather than guessed at, and both are deliberate:

- `{link: "unlinked"}` — a workflow delivery, an operator-chat turn, a scheduler tick. Counting one
  would put "blocked on 1 approval" on a card that is not blocked, then disable its Resume forever,
  since deciding that approval is not something the operator would ever connect to this card.
- **absent** — a park predating #333. The host keeps a run-window heuristic for exactly this case;
  the board has no window to apply it against, so it skips.

`approval_owner`'s first rule — the attempt-level `run_id` — is unreachable from here. It separates
two *runs of the same card*, which does not change the answer to "is this card blocked", the only
question the board asks.

## What this does **not** fix — stated rather than left implicit

**Resume still re-runs.** This stops the operator walking into the loop; it does not make a resume
resumable. That is engine work in the vendored pause primitive, and it is already filed twice:

| issue | the gap | why it is not here |
|---|---|---|
| **#899** | an agent node's internal call has no replay path | `NodeControl::Interrupt` discards the activation's state update, so the node re-runs from the top; `finish_agent_run` already refuses with *"resuming a paused agent is not supported yet"*. #846's `PAYLOAD_PERFORMED` ledger keys on a `tool_call` node's **static** slug and args, and an agent's internal calls are chosen mid-turn, so there is nothing to key against. |
| **#850** | `shell` / `curl` / `git_operations` still repeat | all three are `EffectGroup::Other`, which is useless as a replay predicate — the host cannot tell a `git log` from a `git push`. #846 refused to fold it in on the grounds that replaying on a guess is a bigger behavioural change than the bug. |

So: the issue's two halves land in different places. **"A paused card never says which approvals block
it"** is fixed here, completely. **"Resume re-parks it"** is *contained* here — the operator can no
longer start it from a blocked card — and *fixed* in #899 / #850, neither of which this touches.

A card sitting in Paused with **no** approvals outstanding — one moved there by hand, or a
`RunStatus::Paused` "resumed, not approved" card whose approvals have since been decided — keeps its
enabled Resume and its existing behaviour exactly.

Also untouched by design: batching (#842 / PR #848) — nothing here groups or consolidates cards — and
`approval-card.tsx`, which PR #854 deliberately stayed out of for the same reason.

## API Or Behavior Changes

**None on the wire.** No route, no HTTP/GraphQL type, no Rust file, no `Cargo.lock` line. Every new
prop is optional and defaults to the pre-#883 render, so a console built from this against an older
host behaves exactly as before.

**Behaviour:**

1. A paused card with approvals outstanding shows what it is waiting on, how many, and how long.
2. Its Resume button is disabled while they are undecided, and says why.
3. **Review** opens `#/approvals/<taskId>`, the queue narrowed to that card, with **Show all** back.
4. The task-detail awaiting row names the call and links to the same filtered queue.
5. A paused card with nothing outstanding is unchanged.

## Tests

- [x] `cargo fmt --all -- --check` — N/A: no Rust file is touched. `git diff --name-only
      upstream/main` is seven paths, all under `frontend/`.
- [x] `cargo clippy --all-targets -- -D warnings` — N/A: no Rust file is touched, and no feature
      gate is involved, so neither the default nor the `openhuman,tinycortex` lane has anything to
      lint here.
- [x] `cargo build --all-targets` — N/A: no Rust file is touched; `git diff --stat Cargo.lock` is
      empty.
- [x] `cargo test` — N/A as above. **Ran instead, as the console's CI lane does:**
  - `npm run typecheck` (`tsc -b --noEmit`) — clean
  - `npm run build` — clean
  - `npx vitest run` — **828 passed, 79 files, 0 failed.** No pre-existing failures in this repo;
    13 of those tests are new.

  Not run: `prettier`. CI does not run it, and untouched files (`src/views/InboxView.tsx`,
  `src/hooks/use-company.ts`) fail `prettier --check` on `main` — so running it would have produced
  reformatting noise across files this change has nothing to do with.

### Every new test was proven to fail with its fix reverted — 12 of 13

Each fix reverted individually, the suite re-run, the named test observed failing:

|---|---|
| `since` takes the newest park instead of the oldest | `anchors the wait to the oldest park, whatever order the feed is in` |
| the oldest-first `sort` is dropped | `anchors the wait to the oldest park…`, `orders the approvals oldest park first` |
| `approvalsForTask` stops filtering on the task link | all 3 `approvalsForTask` cases, both `taskApprovalBlock` ownership cases, `is not blocked by another card's approvals` (6 in total) |
| `BlockedRow` is not rendered | `names the one call it is blocked on…`, `counts them when there is more than one…` |
| the row names the mechanism instead of the call | the same two |
| `disabled={block !== null}` → `disabled={false}` | `disables Resume, so the re-dispatch loop cannot be started from here` |
| the `title` reason is removed | `still shows the Resume button, with the reason on it` |

**The 13th is a control case** — `renders no blocked row and an enabled Resume`, the unblocked path —
and has no meaningful revert: any change that makes it fail also fails one of the tests above.
Recorded rather than counted as coverage.

**One test file earns an exception to `vitest.config.ts`'s pure-functions rule**, on the same grounds
`approval-batch-card.test.ts` earns it: the claim under test is *"Resume is disabled"*, which exists
only at the rendered button. `taskApprovalBlock` is unit-tested next door and decides *whether* the
card is blocked; it cannot reach whether the button the operator's hand lands on is actually stopped,
and that is the half that breaks the loop. `TaskItem` is exported for it, with the reason recorded at
the export.

- **No live check.** Browser automation was unavailable for this session, so nothing here was driven
  against the `marketing` staging tenant. The claims are from the code and the suite.
- **The Approvals filter's markup has no component test.** Its predicate — `approvalsForTask`, the
  same one the board uses, deliberately not a second copy — is unit-tested and revert-proven, but the
  filter banner, the **Show all** link and the `ClearedForTask` state are JSX with no component-test
  harness in this repo for that view. Same gap PR #854 recorded for `RunHistoryPanel`'s branch.
- **No test drives an approval end-to-end from park to blocked card.** That needs the e2e lane
  (a live host plus the mock brain), which this change did not add a spec to.

## Documentation

No spec doc changes. The reasoning lives at the code it governs: `frontend/src/lib/task-approvals.ts`
carries the ownership rule, why the two skipped link states are skipped, and why the host's
attempt-level key is out of reach; `TasksView`'s `BlockedRow` carries the disabled-not-hidden
argument and the `approvalAction` reuse.

Closes #883

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -1250,6 +1250,12 @@ export function AppShell({
               // counter the chat's in-flight strip reads, so a card opened from
               // chat lands on the board without a reload.
               taskEventTick={taskEventTick}
+              // Issue #883: a paused card is blocked until every approval its
+              // turn parked is decided, and the board's own read carries none
+              // of them. This is the feed the sidebar badge already polls, so
+              // the card says what it is waiting on without a second request.
+              approvals={feed.approvals}
+              now={feed.now}
               // Issue #246: the card → chat half of the round trip. A card
               // opened from a conversation remembers which one, so its detail
               // screen can put the operator back in that thread.
@@ -1257,6 +1263,11 @@ export function AppShell({
                 setActiveThreadId(threadId);
                 setView("conversation");
               }}
+              // Issue #883: "Review" on a blocked card opens the queue narrowed
+              // to that card. Through `navigate` rather than `setView` so the
+              // filter lands in the hash and survives a refresh and the Back
+              // button, like every other sub-page.
+              onReviewApprovals={(taskId) => navigate("approvals", encodeURIComponent(taskId))}
             />
           )}
           {view === "ledgers" && (
@@ -1317,6 +1328,13 @@ export function AppShell({
               client={client}
               company={company}
               feed={feed}
+              // Issue #883: `#/approvals/<taskId>` narrows the queue to one
+              // card, so "Review" on a blocked card lands on its approvals
+              // rather than on a page the operator has to search. Same
+              // unvalidated second segment every other sub-page gets — only
+              // this view knows whether the id matches anything parked, so it
+              // does that check itself and says so when it does not.
+              sub={sub}
               onResolved={noteSystem}
               onGoToConversation={() => setView("chat")}
             />
--- a/frontend/src/lib/task-approvals.ts
+++ b/frontend/src/lib/task-approvals.ts
@@ -1,4 +1,5 @@
 import type { TaskApproval } from "@/api/tasks";
+import type { ApprovalSummary } from "@/api/types";
 
 /**
  * What the task card says about approvals (issue #468).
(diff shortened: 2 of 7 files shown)


M3gA-Mind:
@coderabbitai review

Issue #380, opened 2026-08-05
approvals: a slow approval times out at the proxy, then reports the decision as failed when it was recorded


oxoxDev:
## What the operator sees

Approve something from the Approvals page on a hosted tenant. Wait. Then get an error that dumps a raw nginx `504 Gateway Time-out` HTML page — `<html><head><title>…` and all — into the console as the failure reason, prefixed with **"Couldn't record your decision"**.

Return to the conversation and it does not continue.

### 1. The message is false. The decision *was* recorded.

`resolve_approval` (`src/runtime/cycle.rs:449-474`) does four things in order:

1. `resolve_outcome` — records the verdict in the approval gate
2. `journal.record_resolved(id)` — **durably journals it**
3. `settle_approved_effect` — mints the single-use grant (#243)
4. `self.run(…)` — the follow-up cycle so the agent learns the verdict and re-issues its call

Steps 1–3 are fast. Step 4 is a full agent turn. The proxy times out during step 4, so by the time the operator sees "Couldn't record your decision", the decision is recorded, journalled and granted. Only the continuation was lost.

One saving grace worth stating: re-approving is safe. A second attempt resolves to `ResolveOutcome::NotParked` and returns `already_resolved_report()`, so no second grant is minted. But the operator has no way to know that, and the natural response to "couldn't record" is to try again.

**Needs verification during the fix:** whether the follow-up cycle survives the proxy giving up, or whether the dropped connection cancels the future mid-turn. If it is cancelled, the grant is minted and the re-dispatch never completes — which matches the reported symptom that the conversation does not continue.

### 2. A raw error body is rendered verbatim, on every route

`safeJson` (`frontend/src/api/client.ts:570-576`) fails **open**: an unparseable body becomes `{ error: text, code: "unparseable" }`. `request()` then takes `envelope?.error` as the `ApiError` message (`client.ts:93`), so the entire nginx HTML page — padding comments included — becomes the error string the UI renders.

This is not specific to approvals. Any non-JSON error response from any route does it: a proxy error page, an HTML 502, a plain-text upstream failure. The client should treat an unparseable body as "no envelope" and fall back to the status text, and should never render an arbitrary upstream body as prose.

### 3. A synchronous resolve cannot survive a reverse proxy

[...]

PR #946, opened 2026-08-17
fix(chat): run the operator chat turn off the request future

### LINKED ISSUE #882: chat: a long turn returns a raw 504 and the turn's work is never journaled
## What happens

A long chat turn returns a raw nginx error page instead of a reply:

```html
<html><head><title>504 Gateway Time-out</title></head>…nginx/1.24.0 (Ubuntu)
```

Worse than the lost response: **the work is lost too**. After the 504, that agent's DM history holds the operator's message and no reply — nothing was persisted, so the turn cannot be recovered or resumed.

Workflow runs are unaffected by comparison: a `feature_pipeline` run also 504'd, and the run kept executing server-side, nodes completing normally. Only the chat lane loses the turn.

## Repro

Ask an agent for something that takes longer than the proxy's read timeout — an SEO audit that fetches a page, or any turn doing several tool calls — on a hosted tenant. Observed on `marketing` (SEO audit) and `smoke1` (workflow run).

## Expected

Either the turn's result survives the dropped connection (journaled and readable from history on reload), or the proxy timeout is raised past the realistic turn ceiling. A multi-tool agent turn routinely exceeds a default 60s gateway timeout, so this is the normal path, not an edge case.

---

M3gA-Mind:
## Summary

A long chat turn returned a raw nginx `504` **and lost the turn's work**. After the timeout the
operator's DM history held their question and no reply: nothing was persisted, so the turn could
neither be read back on reload nor resumed. That is the half that corrupts state — the board and the
journal disagree with reality.

The cause is not a timeout value. It is **cancellation**. The chat turn ran inline inside the request
handler future (`run_chat` → `runtime.run_cycle(...).await`), and hyper drops a handler future the
moment the peer closes — which is exactly what a reverse proxy does when it decides the upstream is
too slow. Tokens spent, side effects half-applied, nothing journaled.

The issue's own evidence points straight at this: a `feature_pipeline` run **also** 504'd and kept
executing server-side, nodes completing normally. Workflow runs are `tokio::spawn`ed
(`WorkflowSpawn::spawn_admitted`); the operator chat turn was the remaining inline path. Same proxy,
same 504, opposite outcome — a cancellation signature, not a timeout signature.

**No timeout is touched here.** Raising one moves the cliff without removing it.

### Why this diff is +232 and not +20

The non-obvious half: **the journal write had to cross the spawn boundary too.**

Spawning only the cycle would still have lost the answer. The `AgentReply` append did not live in the
cycle — it sat back in the handler, *after* `run_chat` returned (`operator.rs:1373-1412` before this
change). So a spawned cycle would have completed correctly and the append that makes a reply readable
would still have died with the dropped handler future. That same append is what the `agent_reply` SSE
frame is derived from, so the live stream would have stayed silent as well. The work is not recorded
until it is journaled, so both halves move together — which is what the new `spawn_chat_turn` /
`journal_chat_replies` split is for.

### Precedent, not a new pattern

This is the shape the repo already prescribes for exactly this failure, applied to the one lane that
had not received it. `CompanyRuntime::resolve_approval_spawned` (`src/company/runtime.rs:1130-1152`)
documents the reasoning verbatim:

> hyper drops a handler's future the moment the peer closes the connection, and a reverse proxy in
> front of a hosted tenant closes it the moment it decides the upstream is too slow […] Awaiting a
> `JoinHandle` is drop-safe: dropping the handle abandons the *waiting*, not the work. So every caller
> — including the one that just awaits it and answers exactly as before — survives a disconnect, with
> no wire change required to get that.

That is precisely what this does for chat.

### Context: the unfinished half of #380

#380 ("a slow approval times out at the proxy, then reports the decision as failed when it was
recorded") named four defects. PR #394 closed it, **frontend-only** — its checklist reads
`cargo — N/A: no Rust file is touched`. Defects 3 and 4 were never done:

> **3.** Decouple the response from the follow-up cycle […] Raising the proxy timeout is a mitigation,
> not a fix.
> **4.** Confirm and fix whether a dropped client connection cancels the in-flight cycle.

The approval lane later got that treatment. Chat did not. This is that work, in the chat lane.

## API Or Behavior Changes

**No wire change.** The handler awaits the spawned turn's `JoinHandle`, so `POST {scope}/chat` answers
with the same `ChatResponse`, the same status, and the same `message_id`/`responses` bytes as before.
No route, DTO, schema or timeout is altered.

The behaviour that changes is what survives a dropped connection: the turn now runs to completion and
its replies are journaled, so the answer is readable from `chat/history` on reload and is emitted on
the `agent_reply` SSE frame. Previously both were lost.

A task that panics or is aborted now surfaces as `OpenCompanyError::BackgroundTask` rather than a
silent empty reply, mirroring `join_follow_up`.

**Ordering note:** replies are now journaled *before* `emit_cycle_webhooks` rather than after. Durable
first is the right order for this bug, and no reader depends on the previous interleaving.

### Known limitation, stated deliberately

`emit_cycle_webhooks` still runs in the handler, so a dropped connection still skips the webhook
fan-out. This mirrors the approval route's **synchronous** arm, which has the same property; only its
`detach` arm spawns the fan-out. Journal state is now drop-safe, outbound notification is not. Left
out of scope rather than silently widened — worth its own follow-up.

### The 504 itself is not fixed here

Reported rather than changed, because the premise it would rest on does not hold:

* **It is a proxy limit.** `frontend/nginx/default.conf.template` sets **no `proxy_read_timeout`** for
  `location /api/`, so nginx's default **60s** applies. `grep` finds no `proxy_read_timeout` anywhere
  in the repo.
* **There is no server deadline.** The router carries one `.layer()` (auth); no `TimeoutLayer`
  anywhere. The only turn deadline in the tree is `BUILD_TIMEOUT = 120s`
  (`src/harness/workflow_build.rs:96`), scoped to the workflow-copilot build turn.
* **The turn is unbounded server-side.**
* **There is no existing 600s deadline to align with.** No such constant exists; the only 600s is the
  unrelated `OAUTH_PENDING_TTL` (`src/app/types.rs:878`).

So a 504 fix would *create* a policy rather than align with one — either a `202` + SSE shape for chat
(the detach arm approvals already has) or a server-side turn deadline returning a real error envelope
instead of an nginx page. Both are larger and both are product calls, so they belong in their own
issue referencing #380 defect 3. This PR fixes the half that loses data.

*UNVERIFIED:* the reported 504 came from a hosted tenant whose ingress nginx is not in this repo. Only
the in-repo template was inspected; no live staging check was made (browser automation unavailable).

## Tests

  exit 0. `--no-deps` mirrors the `Clippy (openhuman, tinycortex)` lane, where it is load-bearing;
  an unscoped run hard-fails on vendored lints CI never sees.
- [x] `cargo build --all-targets` — `cargo build --locked --features openhuman,tinycortex
  --all-targets`, exit 0.
- [x] `cargo test` — two lanes:
  * `cargo test --locked --features openhuman,tinycortex --tests` → **4105 passed, 0 failed, 3
    ignored** across 5 targets (the gated lane; default features skip roughly a third of the code).
  * `cargo test --lib` → **2722 passed, 0 failed, 1 ignored**.

### The new test was verified to gate

`a_dropped_connection_does_not_lose_the_chat_turns_work` (`src/server/operator.rs`) is the chat-lane
twin of the existing `a_dropped_connection_does_not_cancel_the_follow_up_cycle`. A `StalledChatBrain`
holds the turn open; the test drops the request future mid-turn, releases the brain, and asserts the
reply reached the durable journal.

`Router::oneshot` reproduces the cancellation by the same mechanism hyper uses rather than by analogy:
the handler future is owned by the future the caller polls, so dropping the latter drops the former.
This is the method the existing keystone test already relies on.

**Reverting the fix fails it.** Restoring the pre-#882 inline shape (cycle and journal both in the
handler) and re-running produces the field symptom exactly:

```
test server::operator::test::a_dropped_connection_does_not_lose_the_chat_turns_work ... FAILED
panicked at src/server/operator.rs:
  the chat turn died with the dropped connection: the operator's message is journaled,
  the answer is not, and the turn can neither be read back nor resumed (issue #882)
```

Restoring the fix returns it to green. The test also asserts that **nothing** is journaled before the
brain is released, so it cannot pass vacuously against a fix that journals eagerly.

## Documentation

No spec doc changes. `docs/spec/runtime/api.md:48-52` already states the drop-safety contract for the
resolve/run paths ("it is no longer cancelled when a client or a reverse proxy gives up mid-turn");
this brings chat into line with that documented behaviour rather than changing it. The reasoning
specific to the chat lane — and why the journal write must sit on the spawned side — is recorded in
the doc comments on `spawn_chat_turn` and `journal_chat_replies`.

Closes #882

### DIFF
--- a/src/server/operator.rs
+++ b/src/server/operator.rs
@@ -25,6 +25,7 @@ use axum::{Json, Router};
 use futures::StreamExt;
 use futures::stream::Stream;
 use serde::{Deserialize, Serialize};
+use tokio::task::JoinHandle;
 
 use crate::AppState;
 use crate::company::runtime::CompanyRuntime;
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

Issue #947, opened 2026-08-17
workflows: all 21 seeded company templates ship an output node with no destination, so no template run can deliver


M3gA-Mind:
## Summary

**21 seeded workflows across the company templates contain an `output` node. None of them declare a `destination`.**

An output node with no destination is not delivered from — `src/workflows/delivery.rs` has nothing to route to — so every run of a seeded workflow, on every template, delivers nothing. This is the underlying cause of #925, which was filed against one tenant's Feature pipeline but reproduces on all of them.

## Evidence

```
$ rg -A8 'kind = "output"' companies/agentic_software_company/workflows/feature_pipeline.toml
76:kind = "output"
77:name = "Ship and report"
78:summary = "Release behind approval and report back."     ← no destination key
```

Counted across `companies/`: **21 workflows with an output node, 0 with a `destination`.**

That includes `feature_pipeline` (the one #925 observed), `month_end_close`, `acquisition_pipeline`, `discovery_pipeline`, `engagement_pipeline`, `research_loop`, `cohort_pipeline`, `opportunity_pipeline`, `game_build_pipeline`, and the rest.

## Why this is separate from #925

#925's fix (in flight) makes the gap **visible**: a reached output node with no destination now emits a `Skipped` / `NoDestinationConfigured` delivery row instead of silently contributing nothing, so a run says *why* it delivered nothing. That is a behaviour fix and it is small.

This issue is **content**, and it needs a judgement per template that a delivery-path change should not smuggle in: *where should each company's pipeline actually report?* Folding 21 mechanical TOML stanzas into the delivery fix would bury the reviewable part of it under diff.

## What needs deciding

Two candidate defaults, and they are not equivalent:

- **`kind = "channel"`, targeting a desk.** Every template defines `[[group_chat]]` desks, and #837 made desks addressable as `channel` destinations by desk id. This **actually delivers** — the report lands in the desk chat, a surface the console already reads — which is what makes the MVP demo path (*"Done with a link to the artifact"*) work end to end. Needs one judgement per template: which desk is the right audience for that pipeline's report.
- **`kind = "owner"`.** Needs no target and no grant. But per #837 an `owner` delivery on a company with no mailbox now reports **`Failed` / `OwnerFallbackFailed`**, and that is every provisioned tenant by default — so this would turn all 21 templates' runs red rather than delivering anything.

Recommendation: **desk channel.** `owner` is honest but ships a red run to every template, and it does not put the report in front of anyone.

## Acceptance

[...]

Issue #813, opened 2026-08-12
EPIC workflows: a first-time author cannot get a working workflow, and the copilot is the only path most will take


oxoxDev:
## Why this exists

Authored a workflow end-to-end on staging as a first-time user would, using the
create-time copilot, and ran it four times. It took **four runs and three edits**
to deliver one report, and every failure was an authoring-surface problem, not an
engine problem. The engine was correct every single time.

That is the headline: **the runtime works; the authoring experience does not
lead anyone to it.** Most users will only ever touch the copilot, so the copilot
being approximately right is the same as the product being wrong.

## What actually happened, in order

Prompt: *"Every weekday, have the QA Engineer read the Release checklist from the
shared workspace, check the Engineering standards note for anything that changed,
and post a short readiness summary to the engineering channel."*

The copilot produced a complete, wired, plausible graph — id, name, description,
cron `0 9 * * 1-5`, four nodes, three edges. It looked finished. It could not work.

| # | Run | Outcome | Cause |
|---|-----|---------|-------|
| 1 | 29.0s | no delivery | agent tried to post via Slack/composio, parked on approval |
| 2 | 30.6s | no delivery | reworded to forbid external tools — agent correctly reported **no internal chat tool exists** |
| 3 | 5.6s | `1 not delivered` | added an `output` node → **"`engineering` is not a wired channel — this runtime has: operator"** |
| 4 | 6.1s | **sent** | channel id changed to `operator` |

Runs 1 and 2 were ~60s of inference spent proving a dead end the authoring
surface should never have offered.

### 1. The copilot expresses delivery as an agent instruction, when delivery is a node kind

It wrote *"post a readiness summary to the engineering channel"* into an **agent
node's** summary. Delivery is not something an agent can do — `deliver_outputs`
routes an **`output` node's `destination`**, host-side and post-engine, and the
module says why: *"No path here may let a workflow email an arbitrary external
address without an explicit company grant."* An agent has no chat tool by
design — `namespace_of` covers shell/code/web/media/composio/search/repo and
nothing else.

So the copilot's graph asked an agent to do the one thing agents structurally
cannot do, and never emitted the node kind that does it. **A prompt that names a
destination must produce an `output` node.**

### 2. The copilot assigned a teammate that does not exist

Asked for "the QA Engineer"; it set the agent node to `019fcbc3cb55-000000000000`
— a raw id whose workspace folder is one of the provably-empty ones the tidy
[...]

Issue #170, opened 2026-07-29
feat(workflows): Output node has no destination selector — report can't be routed anywhere


oxoxDev:
## Summary
The terminal `output` node ("owner summary" report-back) has no way to configure a destination. There is no email/channel/owner selector in the model or the create form, so an authored output node can only produce a value that surfaces in the transient run-result drawer; it can never be routed to a person or channel.

## Reproduction
1. Workflows → "New workflow".
2. Add an `output` node ("Output — reports the result back"). The only editable fields are id, name, and free-text summary.
3. There is no field to choose who/where the result is sent.

## Root cause
The output node type has no destination field, and neither the create form nor the inspector offers one.
- Model: `WorkflowNodeKind::Output` is documented as "A terminal report-back node" (`src/company/workflow_file.rs:46-48`), but `WorkflowNodeDef` (`src/company/workflow_file.rs:119-152`) has no destination field; an output node is just id/name/summary. Its only kind-specific bucket is the generic `config` blob, which the form never populates.
- Create form: the output palette entry is "Output — reports the result back" (`frontend/src/api/workflows.ts:129`), and `NodeRow` renders no destination input for `kind === "output"` (`frontend/src/views/WorkflowCreateDialog.tsx:344-425`) — only the agent picker is kind-specific.
- Inspector: `NodeDetailPanel` (`frontend/src/views/WorkflowsView.tsx:318-406`) shows summary/agent/config/onError/retry but nothing for a destination.
- Effect: the runner returns `WorkflowRun { output, pending_approvals }` (`src/workflows/runner.rs:125-128`) and the only consumer is the ephemeral `RunResultPanel` (`frontend/src/views/WorkflowsView.tsx:294-301,423`). No delivery ever happens.

## Proposed fix
Add a first-class destination to the output node and honor it at run time.
- Model: add a `destination` field to output nodes (e.g. `{ kind: "email" | "channel" | "owner", target: string }`) in `WorkflowNodeDef` / `RawWorkflow` (`src/company/workflow_file.rs`), validated in `validate`.
- Wire + form: thread it through `CreateNode`/`WorkflowNode` (`src/server/ops/workflows.rs`), the TS model (`frontend/src/api/workflows.ts`), and add a conditional destination selector to `NodeRow` for `kind === "output"` (`frontend/src/views/WorkflowCreateDialog.tsx`).
[...]

PR #948, opened 2026-08-17
fix(workflows): report an output node that has no destination (#925)

### LINKED ISSUE #925: workflows: every run reports "routed no reports" — an Output node has nowhere to send anything
## Summary
Every Feature pipeline run in history — all 13, going back to 04/08 — ends with `Finished — this run routed no reports.` The pipeline runs to completion and delivers nothing anywhere.

There is no destination selector on the Output node, so there is no way to configure where an owner summary or report should go. This has been in the QA list since 7/29 and has never had an issue behind it.

## Observed
Run history for Feature pipeline, every entry:
```
Finished — this run routed no reports.
```
Confirmed again on a fresh run today (`01a00f04a907-000000000000`): all 8 nodes `ok`, `deliveries: []`.

## Expected
An Output node should let the author choose a destination — a chat channel, an email address, a workspace note, a task card — and a run that routes nothing should be distinguishable from a run whose destination was never configured.

## Why it matters
This is the difference between a workflow that runs and a workflow that delivers. Right now the engine is doing all the work and dropping the result on the floor; the only way to see output is to open the run on the canvas.

Note the run-summary line is at least honest about it now (post-#900/#888), which is how this was spotted.

## Environment

M3gA-Mind:
## Summary

Every Feature pipeline run reports `Finished — this run routed no reports.` The cause is not the one #925 was filed against — a destination selector **does** exist, in the create/edit dialog, since #813/PR #825 (details and correction posted [on the issue](https://github.com/tinyhumansai/opencompany/issues/925#issuecomment-5317878487)). The live defect is that `deliver_outputs` skipped a destination-less output node in complete silence:

```rust
let Some(destination) = &node.destination else {
    continue;          // no row, no reason, no log
};
```

So `deliveries` came back empty, and `RunHistoryPanel` fell through every branch to the "routed no reports" line — a sentence that reads identically whether the author routed nothing on purpose or never configured a destination at all. That is exactly what #925's **Expected** section asks to be told apart.

Two changes, both in the routing half:

- **The reached check moves ahead of the destination check.** A node with no destination still has to be *reached* before its absence is worth reporting; an unreached node contributes nothing either way, exactly as before.
- **A reached output node with no destination emits a row.** `DeliveryStatus::Skipped` with a new `DeliveryReason::NoDestinationConfigured`, naming what is missing and how to fix it. The dry-run router takes the same rule, because a test run is where an author most wants to discover this before scheduling it.

`Skipped` is deliberate rather than `Failed`: nothing broke and nothing was attempted, so this is the same class as the existing `AlreadyDelivered` — a report that was never owed to an address, stated with its reason instead of by omission. No new status, no new type.

### The larger finding, filed separately

**21 seeded workflows across the company templates contain an output node; 0 declare a destination.** The Feature pipeline was seeded and never hand-edited, which is why all 13 of its runs since 04/08 delivered nothing — and it means #925 reproduces on every company template, not just the tenant it was filed against. Seeding them is a content decision per template (which desk should each report to?) and would bury this diff, so it is **#947**. This PR is what makes that gap visible to an author instead of silent.

## API Or Behavior Changes

- New `DeliveryReason::NoDestinationConfigured` variant. Additive; the enum has a `#[default] Unspecified` arm and every construction site matches exhaustively, so nothing deserializing an older event changes.
- **A run whose output node has no destination now returns one delivery row where it previously returned none.** The console therefore shows a "Report delivery" block with a `skipped` row instead of "Finished — this run routed no reports", and the existing "not delivered" counter includes it. That is the point of the change: the row is what distinguishes a misconfiguration from a deliberate no-op. Runs whose output nodes *do* carry a destination are byte-for-byte unaffected.
- No change to what leaves the process. A skipped row is never journaled and never touches a transport.

## Tests

Three new tests in `src/workflows/delivery.rs`:

- `a_reached_output_node_with_no_destination_says_so` — the regression. Deps are `None` on purpose so the check must land *before* anything touches a transport; if the arms were ever reordered this would fail with a `NotWired` row instead.
- `deliver_outputs_dry_reports_a_node_with_no_destination` — the same rule on the dry router.
- `an_unreached_output_node_with_no_destination_stays_silent` — the other half: "never configured" is only worth reporting about a node the run actually reached, or every untaken branch would file a complaint.

**Teeth proved by reverting, not asserted.** With both arms reverted to the bare `continue`, **2 of the 3 fail** (`a_reached_output_node_with_no_destination_says_so`, `deliver_outputs_dry_reports_a_node_with_no_destination`). The third **passes reverted** — it asserts unchanged behaviour and exists to guard against over-reach, not to prove the fix. Saying so rather than claiming three.

⚠️ **`workflows::delivery` is behind `#[cfg(feature = "openhuman")]`.** `cargo test --lib workflows::delivery` runs **0 tests and exits 0** — I hit this and briefly believed a vacuous pass. Anything touching this module must use the gated lane below; the default lane cannot run it at all.

- [x] `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` — clean. Mirrors CI; a local `--all-targets` without `--no-deps` drowns in vendored lints CI never sees.
- [x] `cargo build --all-targets` — N/A as written; covered by the `--all-targets` clippy and test compiles above, which build the same targets under the feature set that actually compiles this module.
- [x] `cargo test --locked --features openhuman,tinycortex --lib` — full lib suite green. The default-feature lane is **not** sufficient here (see the warning above).

## Documentation

- `src/workflows/delivery.rs` module header now states that `destination` is optional, that a node without one is the pre-#170 shape rather than a deliberate "route nothing", and what the new row means.
- `deliver_outputs_dry`'s doc updated for the new row.
- `DeliveryReason::NoDestinationConfigured` carries the reasoning for `Skipped`-not-`Failed` at the variant.

No `docs/spec/` change: the delivery module's spec describes destinations in terms of wired adapters, which stays accurate.

Closes #925

### DIFF
--- a/src/ports/workflow_runner.rs
+++ b/src/ports/workflow_runner.rs
@@ -529,6 +529,24 @@ pub enum DeliveryReason {
     /// The destination kind is not one this runtime knows how to deliver to
     /// (unreachable through `parse_workflow`, which rejects unknown kinds).
     UnknownDestinationKind,
+    /// The run reached this `output` node and the node names no destination, so
+    /// there was nowhere to send its report (issue #925).
+    ///
+    /// `destination` is optional on the model because it postdates the node kind
+    /// — an `output` node without one is the shape every graph had before
+    /// [`WorkflowDestinationDef`](crate::company::WorkflowDestinationDef)
+    /// existed, and its report surfaced only in the console's run drawer. That
+    /// made "the author routed nothing on purpose" and "the author never
+    /// configured a destination" the *same* observation: an empty `deliveries`
+    /// list and a run summary reading `Finished — this run routed no reports.`
+    ///
+    /// This row is what tells them apart. It is deliberately a
+    /// [`Skipped`](DeliveryStatus::Skipped) rather than a
+    /// [`Failed`](DeliveryStatus::Failed): nothing broke and nothing was
+    /// attempted, so this is the same class as
+    /// [`AlreadyDelivered`](Self::AlreadyDelivered) — a report that was never
+    /// owed to an address, stated with its reason instead of by omission.
+    NoDestinationConfigured,
     /// This was a **dry run** (issue #542): the report was routed as far as its
     /// destination but deliberately not dispatched, so an operator can see
     /// *where* a report would have gone without anything actually leaving the
@@ -584,6 +602,9 @@ impl std::fmt::Display for DeliveryReason {
             Self::ChannelRefused => "the channel refused the message",
             Self::UnknownDestinationKind => {
                 "the destination kind is not one this runtime can deliver to"
+            }
+            Self::NoDestinationConfigured => {
+                "this output node has no destination, so there was nowhere to send its report"
             }
             Self::DryRun => {
                 "this was a test run, so the report was not sent — its destination is shown so you \
--- a/src/workflows/delivery.rs
+++ b/src/workflows/delivery.rs
@@ -6,6 +6,14 @@
 //! way to send it anywhere. A node's
 //! [`destination`](crate::company::WorkflowDestinationDef) closes that gap.
 //!
+//! `destination` is **optional**, and a node without one is the pre-#170 shape
+//! rather than a deliberate "route nothing": every graph authored before it
+//! existed still has one, including all 21 seeded company templates. Such a node
+//! is not delivered from, but it does now produce a `Skipped` /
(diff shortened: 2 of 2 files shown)


M3gA-Mind:
@coderabbitai review

Issue #339, opened 2026-08-04
spine: a card in Done carries a link to what it produced


oxoxDev:
Part of the spine epic **#183**, §6 — the last station. **Blocked on #242 and #244.**

## What is missing

A task reaching Done today ends as a chat message. The work is real, but what the operator gets is prose in a conversation, with no durable thing to open, share, or hand to someone else.

The spine's whole promise is *prompt in, deliverable out*. Right now the deliverable is a transcript.

## What this should do

A card in Done carries a link. Clicking it opens the thing the task produced.

What "the thing" is depends on what the task did, and the three cases are already distinguishable:

| The task produced | Done links to |
|---|---|
| A file, document, or asset | the artifact, at the version the run produced |
| A workflow it built or ran | that workflow, opened on its canvas |
| An action with no object (a message sent, a record updated) | the attempt's trace, showing what was done |

The third row matters more than it looks. Plenty of real work has no file at the end of it, and "no artifact" must not degrade into "no link" — the trace *is* the deliverable in that case, and #242 makes it a real, addressable object rather than a folded chat message.

## Why it blocks on both

**#242** gives the link something to point at. An attempt becomes addressable — id, status, cost, step-by-step trace — so the third row above has a destination, and so the first two can say *which run* produced the thing rather than just *which task*.

**#244** is the harder blocker. Today an "artifact" is the agent's chat reply, not a file the agent wrote. Until agent-authored files are real artifacts, the first row of that table has nothing honest to link to — the link would open a message and call it a document.

## Design questions to settle first

- **One link or several?** A task that wrote three files and ran a workflow has more than one output. A single primary link is cleaner and probably wrong; a list is honest and clutters the card. Likely: one primary, with the rest reachable from the task detail.
- **Which attempt owns the output?** A task retried three times has three attempts. The output belongs to the one that succeeded — which #242 makes expressible for the first time, and which is the reason this issue waits for it rather than guessing.
- **What does a link do after the thing changes?** An artifact edited by a human after the run should still be reachable, but the card should not silently start pointing at something the agent never produced. Version-pinned link, with a "newer version exists" affordance, is the likely shape.

## Acceptance

- Every card in Done has a link, including tasks that produced no file.
- The link opens the artifact / workflow / trace, not the chat.
- The link is pinned to the attempt that produced it, and a task with multiple attempts links to the successful one.
- A task whose output was later edited by a human still resolves, and says so.

## Related

- #183 — the spine epic (§6)
[...]

PR #949, opened 2026-08-17
fix(tasks): let a card settled by a chat turn carry an output link (#806)

### LINKED ISSUE #806: tasks: a card settled by an operator chat turn can carry no output link, because TaskOutput requires a run_id
## What is missing

`TaskOutput.run_id` is `String` — required, documented as *"always present, and
the link of last resort"* (`src/ports/tasks.rs`). An **operator chat turn has no
run row**: run records are reserved for actual work attempts, so a turn that
produces something from chat has nothing to address a `TaskOutput` with.

The consequence is a card that settles honestly but cannot link to what it
produced.

## Where it bites today

PR #805 (part of #678) settles the #463 handler card when a chat turn authors a
workflow inline with `create_workflow`. Before it, that card sat in To-do
forever. After it, the card moves and **names its workflows in the note** — but
carries no `TaskOutput`, so the console shows no output link on a card whose
work is demonstrably finished.

The note is a real record. It is not a link, and the board's own contract
(#339: *"Done carries a link to what it produced"*) is written in terms of links.

## Why this has no home right now

#678 states this is "#183's territory". That is stale: **#183 closed with its
checklist complete and did not add any notion of output without a run.** So the
gap is real, unowned, and only visible from the two call sites that hit it.

## What to decide

The shape is a product/schema decision, not a patch. Three candidates, none
free:

1. **Make `run_id` optional.** Honest, and the widest blast radius — every
   reader of `TaskOutput` currently relies on it being present, and the type's
   own doc calls it the link of last resort.
2. **Synthesise a run row for an operator turn.** Keeps `TaskOutput` intact, at
   the cost of run records no longer meaning "a work attempt" — which #183 §4
   settled deliberately (*"Run records stay reserved for actual work attempts,
   so the Attempts list shows turns that did something"*). Directly reopens a
   settled decision.
3. **A second, link-only output shape** that does not carry a run. Narrower than
   1, but adds a type the console must learn.

## Acceptance

A card settled by an operator chat turn carries a link to what that turn
produced, and whichever option lands, `docs/spec/` and the #183 §4 statement
about what a run record means still agree with the code.

## Related

- #805 — the settle that exposes this; its comment points here rather than at a
  closed issue
- #678 — where it was first written down
- #183 — the epic that owned it, now closed
- #339 — "Done carries a link to what it produced"

M3gA-Mind:
## Summary

A card settled by an operator chat turn could carry no output link at all. `TaskOutput.run_id` was a required `String`, an operator chat turn has no run row, and PR #805 therefore settles the #463 handler card with its workflows named in the **note** and no `TaskOutput` — so the console shows no output link on a card whose work is demonstrably finished. The board's own contract (#339) is written in terms of links, and prose is not one.

### The distinction this rests on

`ports/tasks.rs` argued the case for a mandatory `run_id` under the heading *"`run_id` is unconditional, and that is the whole design"*. **That conflates the guarantee with the single mechanism that happened to provide it.** The invariant the type protects is *"there is always something to open"* — `run_id` is one way to satisfy it, not the thing itself. A run is *an* addressable producer; it is not the only one. `TaskRecord::origin_chat_id` has existed since #151, and `TaskDetailView` already renders "Opened from a conversation".

A reviewer reading that heading will otherwise think this PR broke a deliberate design. It did not — it separates the two and keeps the guarantee, now enforced by the type rather than by a comment.

So `TaskOutput` names **what produced it** as a closed set:

```rust
#[serde(untagged, rename_all_fields = "camelCase")]
pub enum TaskOutputSource {
    Run { run_id: String, attempt: Option<u32> },
    ChatTurn { chat_id: String },
}
```

`settle_authored_workflow_card` now stamps the card with the workflows it was **already holding** — `authored: &[TaskOutputWorkflow]` was in scope the whole time; `run_id` was the only thing stopping it.

### The two rejected alternatives, and why

- **Synthesise a run row for the turn.** Rejected: it reopens #183 §4 on purpose — *"run records stay reserved for actual work attempts, so the Attempts list shows turns that did something"* — and puts an id that addresses nothing real behind a type that promises it does. `docs/spec/runtime/ports-runs.md` already says old `RunRecord`s are never synthesised because *"fabricating identity for attempts nobody recorded would be worse"*; this is the same rule one case further on.
- **Make `run_id` optional.** Rejected: honest about the absence, silent about what *is* there. Every reader would re-derive "then what do I link to?", and the console's trace arm would degrade to the card even for a turn that demonstrably produced a workflow — a wrong link traded for no link.

## API Or Behavior Changes

- `TaskOutput.run_id` / `.attempt` are replaced by `TaskOutput.source`. Readers that only wanted "which run, if any" use `source.run_id() -> Option<&str>` and `source.attempt() -> Option<u32>` rather than each growing a `match`.
- **No wire change and no migration.** `#[serde(untagged)]` + `#[serde(flatten)]` means a `Run` source reads and writes exactly the `runId` / `attempt` keys the struct always had. Stored cards deserialize into `Run`, which is what they always were. A `ChatTurn` is new keys on new records only.
- A card settled by an operator chat turn now carries a `TaskOutput` where it previously carried none, so the console shows its workflow link instead of "Open this task". A card with **no** `origin_chat_id` keeps the note-only behaviour — there is no conversation to address, and a stamp pointing nowhere would be worse than none.
- Console: `primaryLink`'s precedence (artifact → workflow → last resort) is unchanged. The last-resort arm branches on the source; a chat turn resolves to the card, never to `#/tasks/<id>?run=undefined`.

## Tests

Backend (`src/ports/tasks.rs`):

- `a_stamp_written_before_the_source_union_still_reads_as_a_run` — **the migration guarantee.** The legacy JSON is a hand-written literal, deliberately *not* produced by serializing the current type: a symmetric round-trip would still pass if both halves drifted together, which is precisely the failure it exists to catch. It did catch one — see below.
- `a_run_source_serializes_to_the_keys_it_always_did` — the other direction, and it pins that the union is flattened rather than nested or tagged.
- `a_chat_turn_stamp_carries_a_conversation_and_no_run`.

`src/runtime/delegation.rs`:

- `a_workflow_authored_in_a_chat_turn_gives_its_card_an_output_link` — the regression: the settled card carries the link, sourced from the conversation, with `run_id()` `None`.
- The existing sibling test gained an assertion that a card with no `origin_chat_id` still gets no stamp.

Console (`frontend/test/unit/task-links.test.ts`): 15 pass. **Honestly, 1 of the 2 new ones has teeth** — `never labels a chat turn as a run trace` fails when the last-resort arm is reverted; `opens the workflow a chat turn authored` passes reverted, because workflow precedence never reaches the trace arm, so it is a type-level guard rather than a runtime one. Stating that rather than claiming two.

> **A real bug this caught.** The first cut used `#[serde(rename_all = "camelCase")]` on the enum. On an **untagged** enum that renames the *variants* — which are never written — and does nothing to their fields, so it emitted `run_id` / `chat_id` in snake_case and **every stored card's output would have failed to deserialize.** `rename_all_fields` is the attribute that reaches them. Only the hand-authored legacy literal caught it.

⚠️ **`runtime::delegation` is behind `#[cfg(feature = "openhuman")]`** — `cargo test --lib runtime::delegation` runs 0 tests and exits 0. The gated lane below is the only one that runs it.

- [x] `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` — clean. Mirrors CI; a local `--all-targets` without `--no-deps` drowns in vendored lints CI never sees.
- [x] `cargo build --all-targets` — N/A as written; the `--all-targets` clippy and test compiles above build the same targets under the feature set that actually compiles these modules.
- [x] `cargo test --locked --features openhuman,tinycortex --lib` (with `RUST_MIN_STACK=16777216`, as CI sets) — green. Plus `npx vitest run test/unit/task-links.test.ts` — 15 passed.

## Documentation

- `TaskOutput`'s doc heading rewritten from *"`run_id` is unconditional"* to the invariant it actually protects, with the mechanism/guarantee distinction spelled out.
- `TaskOutputSource` carries the rejected alternatives and the wire-compatibility reasoning at the type.
- `docs/spec/runtime/ports-runs.md` gains the chat-turn case beside the existing "never synthesised" rule, so #183 §4's statement about what a run record means still agrees with the code — the acceptance criterion in the issue.
- `src/runtime/delegation.rs`'s two stale comments ("there is no `TaskOutput` without a run row") corrected.

## Known gap, deliberately excluded

The chat last-resort link points at the **card**, not the conversation. `conversation` is a view in the hash router, but the active thread is component state (`activeThreadId`) and nothing reads a thread id out of the hash — so `#/conversation/<id>` resolves nowhere today. Minting that href would look right and silently land on the wrong thread. The card is where the conversation is already reachable from ("Opened from a conversation", #246). Deep-linking a thread is its own change.

This does not weaken the fix: `settle_authored_workflow_card` is called under `if !authored.is_empty()`, so a chat-settled card always has at least one workflow and `primaryLink` returns the **workflow** link. The chat arm is defensive, not the main path.

Closes #806

### DIFF
--- a/docs/spec/runtime/ports-runs.md
+++ b/docs/spec/runtime/ports-runs.md
@@ -133,6 +133,17 @@ Old `RunRecord`s are never synthesised from historical `AgentReply` events:
 fabricating identity for attempts nobody recorded would be worse than a
 pre-existing card honestly showing zero of them.
 
+Nor is one synthesised for an **operator chat turn** (issue #806). A turn that
+produces something from chat — authoring a workflow inline with
+`create_workflow`, say — has no attempt behind it, and minting a run row so its
+card could carry a `TaskOutput` would make the Attempts tab claim work was
+attempted. Run records stay reserved for actual work attempts, so the Attempts
+list shows turns that did something (epic #183 §4). `TaskOutput` instead names
+*what produced it* as a closed set — `TaskOutputSource::Run` or
+`TaskOutputSource::ChatTurn` — which keeps the board's "every card in Done links
+to what it produced" (#339) true without weakening what a run means. See
+[ports-state.md](ports-state.md) for the task record itself.
+
 ## Reading runs back
 
 Three surfaces, all in `src/server/ops/runs.rs`, under both scope forms:
--- a/frontend/src/api/tasks.ts
+++ b/frontend/src/api/tasks.ts
@@ -48,29 +48,54 @@ export interface TaskOutputWorkflow {
 }
 
 /**
+ * What produced a {@link TaskOutput} — and what its link of last resort points
+ * at (issue #806).
+ *
+ * A run is *an* addressable producer, not the only one. An operator chat turn
+ * produces things too, and run records stay reserved for actual work attempts,
+ * so such a turn has no run to name. This is a union rather than an optional
+ * `runId` so the console can label what is at the other end instead of calling
+ * a conversation an attempt.
+ *
+ * Discriminate with `"runId" in source` — the host flattens these keys onto the
+ * output, so a run-sourced record is byte-identical to what this type carried
+ * before the union existed.
+ */
+export type TaskOutputSource =
+  | {
+      /** The attempt that produced it. */
+      runId: string;
+      /** Its 1-based ordinal, for the label. Absent if unreadable. */
+      attempt?: number;
+    }
+  | {
+      /** The conversation whose turn produced it. */
+      chatId: string;
+    };
+
+/**
  * What a card's latest **successful** attempt produced (issue #339, epic #183
  * §6) — the link that turns a finished card into something you can open.
  *
- * `runId` is unconditional, and that is the whole design: an output with no
- * artifacts and no workflows is not an absence, it is the *trace case*. Plenty
- * of tasks produce no file — a message sent, a record updated, a question
- * answered — and for those the attempt's trace **is** the deliverable. It is
- * also the fallback when an artifact is later deleted.
+ * **A last-resort link is unconditional, and that is the whole design**: an
+ * output with no artifacts and no workflows is not an absence, it is the *trace
+ * case*. Plenty of tasks produce no file — a message sent, a record updated, a
+ * question answered — and for those the producer's own record **is** the
+ * deliverable. It is also the fallback when an artifact is later deleted.
+ *
+ * The guarantee is "there is always something to open", not "there is always a
+ * `runId`" (issue #806) — see {@link TaskOutputSource}.
  *
  * Absent entirely on a card that has never succeeded, one moved to Done by
  * hand, and every card settled before this shipped. Those link to the card
  * itself rather than to a synthesized output.
  */
-export interface TaskOutput {
-  /** The attempt that produced it — always present. */
-  runId: string;
-  /** That attempt's 1-based ordinal, for the label. Absent if unreadable. */
-  attempt?: number;
+export type TaskOutput = TaskOutputSource & {
   /** Epoch-millis the stamp was written. */
   atMillis: number;
   artifacts?: TaskOutputArtifact[];
   workflows?: TaskOutputWorkflow[];
-}
+};
 
 /**
  * What surface a plan's prerequisite is checked against (issue #337).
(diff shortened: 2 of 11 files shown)


M3gA-Mind:
@coderabbitai review

PR #953, opened 2026-08-17
fix(console): stop a toast being pinned open forever (#933)

### LINKED ISSUE #933: console: the "Starting the product tour." toast never dismisses
## Summary
The success toast raised by Settings → Replay tour stays on screen indefinitely. It survived nine minutes and four full page navigations (Settings → Overview → MCP Servers → Connections → Approvals), sitting over the bottom-right of every view.

## Repro
1. Settings → Product tour → Replay tour.
2. Navigate anywhere else. The toast follows and never fades.

Its close button works; nothing else clears it.

## Expected
Auto-dismiss on the usual timeout, and in any case clear on route change.

## Environment

CodeGhost21:
## Summary

The "Starting the product tour." toast could stay on screen indefinitely (#933).
`toast.success` relies on sonner's 4s default, and that default is not a deadline
— it is a timer sonner *pauses* on three pieces of `Toaster` state: `expanded`,
`interacting`, and a hidden document. Two of them latch with no path back:

- `expanded` is set by `mouseenter`/`mousemove` **and by sonner's built-in Alt+T
  hotkey**, which also focuses the toast list. It clears only on a `mouseleave`,
  on Escape while focus is inside the list, or when the toast array changes.
  Expand it with the hotkey while the pointer is elsewhere and no `mouseleave`
  can arrive; with one toast up nothing changes the array either.
- `interacting` is set on `pointerdown` and cleared on `pointerup` *on the
  toaster*. Pointer capture covers the ordinary drag, but sonner skips the
  capture for a toast that opted out of dismissal.

The single `<Toaster>` is mounted at the app root, outside the routed tree, so a
latched toast then rides over every view until someone clicks its ×. That is the
report: nine minutes, four navigations, only the × working.

sonner 2.0.8 (latest) has identical logic, so there is nothing to upgrade to.

## What this does

Adds the ceiling sonner does not give, at the console's one choke point:
`src/lib/toast-lifetime.ts` accumulates each toast's *visible* time and names the
ones whose duration plus a 2s grace is spent; `components/ui/sonner.tsx` sweeps
every 500ms and dismisses them. It reads the *real* hover state
(`[data-sonner-toaster]:matches(":hover")`) rather than sonner's belief about it.

All three intentional pauses still hold, so no caller has to change:

- hovering to read holds a toast for as long as the pointer is on it;
- a hidden tab stops the clock, so an unseen toast keeps its full life;
- an explicit `duration: Infinity` is never touched.

The sweep only runs while a toast is on screen — the console is a long-lived
dashboard and a timer ticking all day to find nothing keeps the tab awake.

Deliberately **not** included: clearing toasts on route change (the other half of
the issue's Expected). It is in direct tension with this very toast — the tour
navigates to Overview immediately after raising it, so a blanket
dismiss-on-view-change would flash it away, and it would silently affect every
other raise-then-navigate pair in the console. The ceiling gives the guarantee
the report actually wanted without that cost.

## Behaviour changes

No API change. Any toast is now guaranteed to leave on its own; previously none
were. A toast held open by hover is unaffected.

## Commands run locally

```
npm run typecheck            # tsc -b --noEmit
npm run typecheck:e2e
npm test                     # 807 unit tests, incl. 10 new
npm run e2e                  # 146 passed, 14 skipped, 0 failed
```

The two latch specs in `test/e2e/toast-dismissal.spec.ts` were **seen failing**
with the sweep disabled, while the two that pin existing behaviour (plain
dismissal, hover-to-read) kept passing — so the new specs assert something.

## Untested edge cases

`interacting` latching via a non-dismissible toast is described in
`toast-lifetime.ts` but not covered by a spec: no console toast sets
`dismissible: false` or uses `toast.loading`, so there is no way to reach it from
the app today. The ceiling covers it regardless, since it does not consult
sonner's state.

Closes #933

### DIFF
--- a/frontend/ARCHITECTURE.md
+++ b/frontend/ARCHITECTURE.md
@@ -285,6 +285,15 @@ feature-gated off.
 - **Graceful 404:** until an endpoint exists it should 404; the console already
   treats that as "not wired yet" and shows the sample/notice — so partial
   rollout is safe.
+- **Toast lifetime:** the one `<Toaster>` is mounted at the app root, outside the
+  routed tree, so a toast outlives the view that raised it and nothing about
+  changing view clears one. `sonner`'s auto-dismiss is a timer it *pauses* — on
+  hover, on a pointer interaction, on a hidden tab — and two of those latch with
+  no way back (issue #933). `src/lib/toast-lifetime.ts` is the ceiling that makes
+  a stuck toast impossible: it accumulates each toast's *visible* time and
+  dismisses one whose duration plus a grace period is spent. Hovering to read, a
+  backgrounded tab, and an explicit `duration: Infinity` are all still honoured,
+  so callers keep raising plain `toast.*` calls and need not think about it.
 
 ## Implementation order (delivered)
 
--- a/frontend/src/components/ui/sonner.tsx
+++ b/frontend/src/components/ui/sonner.tsx
@@ -1,9 +1,66 @@
+import { useEffect, useRef } from "react"
 import { useTheme } from "next-themes"
-import { Toaster as Sonner, type ToasterProps } from "sonner"
+import { Toaster as Sonner, toast, useSonner, type ToasterProps } from "sonner"
 import { CircleCheckIcon, InfoIcon, TriangleAlertIcon, OctagonXIcon, Loader2Icon } from "lucide-react"
+
+import { reconcileTracked, sweepToasts, type TrackedToast } from "@/lib/toast-lifetime"
+
+/** How often the dismissal guard re-checks the toasts that are up. */
+const SWEEP_INTERVAL_MS = 500
+
+/** Is the pointer genuinely over the toaster, whatever sonner's state believes? */
+function toasterHovered(): boolean {
+  return Array.from(document.querySelectorAll("[data-sonner-toaster]")).some((el) =>
+    el.matches(":hover"),
+  )
+}
+
+/**
+ * The ceiling on a toast's life that sonner does not provide (issue #933).
+ *
+ * sonner's auto-dismiss is a *pausable* timer, and two of the things that pause
+ * it can latch with no way back — see `lib/toast-lifetime.ts` for which, and
+ * why. The reported symptom was the "Starting the product tour." toast sitting
+ * over every view for nine minutes with a working × and nothing else that
+ * cleared it. This sweep is what makes that impossible: a toast nobody is
+ * hovering, in a tab the operator is looking at, goes away once its own duration
+ * plus a grace period has passed, whatever sonner thinks its timer is doing.
+ *
+ * Deliberately additive. Every intentional pause still holds: hover to read,
+ * hidden tab, and an explicit `duration: Infinity` are all left alone.
+ */
+function useToastDismissalCeiling(): void {
+  const { toasts } = useSonner()
+  const tracked = useRef<TrackedToast[]>([])
+
+  useEffect(() => {
+    tracked.current = reconcileTracked(
+      tracked.current,
+      toasts.map((t) => ({ id: t.id, duration: t.duration })),
+    )
+  }, [toasts])
+
+  // Only while something is on screen: the console is a long-lived dashboard, and
+  // a timer that ticks all day to find nothing is a timer that keeps the tab
+  // awake for no reason.
+  const anyToasts = toasts.length > 0
+  useEffect(() => {
+    if (!anyToasts) return
+    const timer = window.setInterval(() => {
+      const { next, overdue } = sweepToasts(tracked.current, SWEEP_INTERVAL_MS, {
+        hovered: toasterHovered(),
+        documentHidden: document.hidden,
+      })
+      tracked.current = next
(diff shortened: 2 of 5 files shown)


CodeGhost21:
## Manual E2E report — real backend API, driven through Chrome

Re-tested this change end to end in a **real Chrome browser** (Claude in Chrome
extension) against a **real host serving the real API** — not the mocked
Playwright lane. Two hosts were built from this branch and run side by side so
the before/after is a like-for-like comparison:

| host | bundle | `matches(":hover")` fingerprint |
| --- | --- | --- |
| `:8092` | ceiling **disabled** (pre-fix baseline) | 1 |
| `:8091` / `:8093` | ceiling **enabled** (this PR) | 2 |

Both served `frontend/dist` through `opencompany serve --company
companies/e2e_harness` with a real MongoDB-less store, real magic-link auth, and
a live SSE stream (`[events] connected` in the console).

### Backend integration verified first

Everything below ran against genuine API traffic, not fixtures:

- `POST /api/v1/company/auth/request` → `202` with a `dev_code`
- `POST /api/v1/company/auth/verify` → `200`, session cookie set
- `GET /api/v1/companies` → `200`
- Settings → General rendered live host data: company `e2e-harness-co`, template
  `e2e_harness`, mode `Single-company`, host `same origin`
- Overview rendered the real knowledge graph (Engineering desk, Content desk,
  agents), and Workflows rendered the real `Committed flow` from the company
  source tree

### Pre-fix baseline (`:8092`) — the bug, reproduced in a real browser

Clicked **Settings → Product tour → Replay tour** with the mouse. The tour opened
at "Step 1 of 8" and the toast appeared bottom-right. Pressed **Alt+T** (sonner's
built-in expand hotkey) and parked the pointer at the far left of the viewport.

The live DOM confirmed the latch: `[data-sonner-toast][data-expanded="true"]`.

```
1.8s PRESENT expanded=true      13.8s PRESENT expanded=true
3.8s PRESENT expanded=true      15.8s PRESENT expanded=true
5.8s PRESENT expanded=true      17.8s PRESENT expanded=true
7.8s PRESENT expanded=true      19.8s PRESENT expanded=true
9.8s PRESENT expanded=true      21.8s PRESENT expanded=true
11.8s PRESENT expanded=true     23.8s PRESENT expanded=true
```

Then navigated **Overview → Approvals → Workflows** by clicking the sidebar. The
toast rode along onto `#/workflows/committed` and kept going for another 11.8s of
samples. Clicking its **×** removed it immediately.

That is issue #933 exactly: never auto-dismisses, survives route changes, only
the close button works.

### With this PR (`:8091` / `:8093`) — the ceiling fires

Identical steps, identical latch (`data-expanded="true"`). An instrumented build
of the same code logged what the guard sees:

```
[g] reconcile live=1 tracked=1
[g] effect anyToasts=true
[g] tick n=1 ms=4500 hov=false hid=false due=0
[g] tick n=1 ms=5000 hov=false hid=false due=0
[g] tick n=1 ms=5500 hov=false hid=false due=0
[g] tick n=1 ms=6000 hov=false hid=false due=1   <-- ceiling reached, dismiss called
```

`due=1` lands at exactly `ms=6000` — the toast's 4000ms duration plus the 2000ms
grace — with the pointer nowhere near the toaster and the document visible. The
pre-fix build never reaches that line at all.

The hidden-tab exemption was also observed working, unprompted: while the browser
window was genuinely occluded, every tick read

```
[g] tick n=1 ms=0 hov=false hid=true due=0
```

— the clock stays at zero, so a toast raised against a console nobody is looking
at keeps its full life rather than expiring unseen.

### Two honest caveats about this environment

1. **The extension's tab is occluded, so `document.hidden === true`.** That
   legitimately pauses both sonner's timer and this fix's clock, which would make
   any timing comparison meaningless. For the timing runs above, visibility was
   pinned in-page (`Object.defineProperty(Document.prototype, 'hidden', …)` plus a
   `visibilitychange` dispatch) purely to compensate for the automation window.
   Everything else — the clicks, the Alt+T latch, the route changes, the API
   traffic — is untouched.
2. **DOM removal cannot complete in an occluded tab.** sonner's removal path runs
   inside `requestAnimationFrame`, and this tab schedules **0 animation frames in
   3.5 s** (measured). So the node stays even after the guard has correctly called
   `toast.dismiss()` — visible as `due=1` repeating on later ticks. This cannot
   occur in production: the guard only ever asks for dismissal when
   `document.hidden === false`, and a visible tab gets frames. Confirmation that
   the node actually leaves therefore comes from the Playwright lane in a
   non-occluded browser, where `test/e2e/toast-dismissal.spec.ts` passes 4/4 and
   the same instrumented build shows the toast gone one tick after `due=1`.

### Automated suites, for completeness

Run on this branch against a host built from it:

```
npm run typecheck        # tsc -b --noEmit — clean
npm run typecheck:e2e    # clean
npm test                 # 807 unit tests pass (10 new)
npm run e2e              # 146 passed, 14 skipped, 0 failed
```

The two latch specs were seen failing with the sweep disabled while the two that
pin existing behaviour (plain dismissal, hover-to-read) kept passing, so the new
specs assert something real.

PR #955, opened 2026-08-17
fix(workflows): serve the effective tool set from tool-slugs (#874)

### LINKED ISSUE #874: bug: tool-slugs advertises granted-but-unwired tools, so the copilot authors nodes that cannot run

CodeGhost21:
## Summary

Fixes #874. `GET …/workflows/tool-slugs` answered the company's grants alone and
ignored whether this deployment wires the tool, so a granted-but-unwired
`web_search` was advertised, the browser-side copilot grounded on it, and the
authored graph died at its first node with
`tool_call 'web_search' is not available in company workflows`.

### Root cause

Two copilot tool-grounding surfaces existed and only one was made wiring-aware.
`53e58389` added `workflow_effective_tool_slugs` / `workflow_granted_but_unwired_tool_slugs`
and wired them into the **in-process** create/fix copilot
(`src/harness/workflow_build.rs`), but the **HTTP route** the console's
`CopilotPanel` reads still called the grant-only `workflow_callable_tool_slugs`.

That helper filters `WORKFLOW_TOOL_CATALOG` by `[tools].allow` only — deliberately,
as the companion of `validate_tool_call_node`, which stays permissive so an
operator can author before wiring. Serving that *validation* answer to a caller
asking a *grounding* question is the bug.

### The fix

- The route serves `workflow_effective_tool_slugs` — catalogue, grant and
  deployment wiring all agreeing — the same set the in-process copilot already
  grounds on, so the two surfaces cannot drift.
- Granted-but-unwired tools are **reported, not dropped**: a new `unwired` list
  carries each slug with a `reason` (`searchBackendNotConfigured` /
  `capabilityTierFiltered`) and prose `detail`, taken from the same
  `WorkflowToolWiring` the run-time gate reads — the distinction `refusal_for`
  already makes at run time, moved forward to when the console asks.
- `CompanyRuntime::workflow_tool_wiring` exposes the full wiring including
  reasons; `wired_workflow_namespaces` now derives from it.
- The console passes `unwired` into the copilot prompt as a "do NOT author these"
  advisory, mirroring the create-time copilot's `list_effective_tools`.

### Deliberate non-changes

- **Create/save validation stays permissive.** `validate_tool_call_node` still
  checks grants alone; authoring ahead of wiring remains legal. This narrows what
  a caller is *told is available*, not what the host will *accept*.
- **Unknowable wiring is not "unwired".** With no harness deps the deployment
  cannot be asked, so `slugs` falls back to the grant-only set and `unwired` is
  empty — the pre-#874 answer.

### API change

```jsonc
{ "slugs": ["shell", "read_workspace_state"],
  "unwired": [ { "slug": "web_search",
                 "reason": "searchBackendNotConfigured",
                 "detail": "granted, but no managed search backend is configured on this deployment; …" } ] }
```

`unwired` is additive. A console reading a host that predates it defaults to `[]`,
which reads identically to a fully wired deployment.

`workflow_callable_tool_slugs` lost its only production caller and is removed; its
grant rule lives on as the shared `grants_workflow_tool` predicate, replacing what
had been three copies of the same filter.

## Tests

- `tool_slugs_omits_a_granted_but_unwired_tool_and_says_why` reproduces the
  staging scenario through the route. Verified it **fails** against the pre-fix
  behavior, emitting exactly the reported bug:
  `"slugs":["shell","read_workspace_state","web_search"]`.
- `workflow_wiring_names_why_each_namespace_is_unwired`,
  `a_granted_but_unwired_tool_is_reported_not_offered`, and
  `unknowable_wiring_offers_the_grant_only_set_and_reports_nothing_unwired`
  cover the reason mapping and the "cannot say" fallback.
- Frontend: new `workflow-tool-slugs.test.ts` (both lists, back-compat default)
  plus copilot-prompt coverage for the advisory and its absence.
- The `HarnessDeps` wiring fixture moved to `harness::workflow_wiring_deps` so the
  runtime and route tests describe one deployment shape instead of two copies.

## Commands run locally

```
cargo +1.96.1 fmt --all -- --check
cargo +1.96.1 clippy --features openhuman --all-targets -- -D warnings
cargo +1.96.1 clippy --all-targets -- -D warnings
RUST_MIN_STACK=33554432 cargo +1.96.1 test --features openhuman --tests   # 4104 + all integration targets pass
RUST_MIN_STACK=33554432 cargo +1.96.1 test                                # 2728 pass (default build)
cd frontend && npx vitest run                                             # 847 pass
cd frontend && npm run typecheck && npm run typecheck:unit
```

Two local-only notes, both verified pre-existing and unrelated to this diff:
`libsqlite3-sys 0.38.1` needs a newer rustc than a stale local `stable`
(hence `+1.96.1`), and several `harness::brain` delegation tests overflow the
debug-build stack without `RUST_MIN_STACK`.

## Summary by CodeRabbit

* **New Features**
  * Workflow tool discovery now distinguishes effective, deployment-ready tools from granted but unavailable tools.
  * The workflow copilot explains when a granted tool cannot be used and provides the reason.
  * Tool availability responses include structured details about unavailable tools, with compatibility support for older responses.

* **Bug Fixes**
  * Copilot suggestions no longer propose tools unavailable in the current deployment.
  * Empty tool sets are now communicated accurately as having no runnable workflow tools.

* **Documentation**
  * Added documentation for workflow tool availability responses and fallback behavior.


M3gA-Mind (commented):
**Review: comment only — not an approval.** Approval on this repo is the maintainer's call.

I reviewed this against what #874 actually reports, not as a tidiness pass. **It closes the failure**, and the closure looks complete rather than local to the one endpoint:

- `workflow_callable_tool_slugs` now has **zero callers** — the grant-only set is no longer reachable as a grounding source anywhere in `src/`. I checked every `WORKFLOW_TOOL_CATALOG` consumer, and both grounding surfaces (this route and the in-process create/fix copilot at `src/harness/workflow_build.rs:1763`, which feeds `list_effective_tools`) read `workflow_effective_tool_slugs`. That is the part that makes this a fix rather than a patch on one path.
- **Create/save validation stays permissive**, which #874 explicitly asked for. `validate_tool_call_node` is untouched; the grant rule was extracted to `grants_workflow_tool` and shared, so author-now-wire-later remains legal.
- The `None`-wiring case is handled the right way round: unknowable wiring degrades to the grant-only set instead of declaring everything broken.
- The frontend gets this right where it would have been easy to get wrong — `unwired` is injected under an explicit *"do NOT author these"* heading, and only `slugs` is offered as groundable. Passing both into the prompt as one list would have quietly restored the bug.

On the size: **768 lines is justified, not bundled.** Roughly 73 of them are the `HarnessDeps` fixture moved to `src/harness/mod.rs` (`#[cfg(test)]`-gated) so the runtime tests and the route test describe one deployment shape instead of two; the rest is tests, docs, and the frontend half. I did not find unrelated change riding along.

Two findings, both minor. Neither is a live defect.

---

### 1. `src/server/ops/workflows.rs:2104` — the `_` arm will mislabel a future `MissingReason`

```rust
Some(MissingReason::SearchBackendNotConfigured) => ("searchBackendNotConfigured", …),
_ => ("capabilityTierFiltered", "…ask the platform operator to raise the capability tier"),
```

`MissingReason` has exactly two variants today (`src/workflows/caps/tools.rs:62`), so this is correct as written. But the catch-all means adding a third variant compiles silently and reports it to operators as `capabilityTierFiltered`, with prose telling them to *raise the capability tier* — advice that would be actively wrong for a reason that isn't tier filtering. Since the whole point of this field is letting an operator tell one cause from another, this is the one place worth making the compiler enforce:

```rust
Some(MissingReason::SearchBackendNotConfigured) => (…),
Some(MissingReason::CapabilityTierFiltered) => (…),
None => (…),   // the defensive branch the comment already describes
```

That also separates the genuinely-defensive `None` case from the tier case, which the current `_` conflates.

### 2. `frontend/test/unit/workflow-tool-slugs.test.ts:63` — this test passes with the fix reverted

`keeps an unwired tool out of the groundable set` feeds a canned body of `slugs: ["shell"], unwired: [web_search]` and asserts `slugs` doesn't contain `web_search`. But `listWorkflowToolSlugs` is a pure pass-through — `r.slugs` out, `r.unwired ?? []` — so the assertion holds because the *fixture* was written that way, not because of anything the client does. The filtering it claims to pin happens server-side. Revert the entire #874 change and this test still goes green.

The docstring above it makes the strong claim ("*that is issue #874 — `web_search` was offered, the copilot authored a node on it*"), which is what makes it worth flagging: it reads as coverage of the core regression and isn't. The two neighbouring tests already cover what this file can genuinely test (pass-through of both lists, and the `unwired` default). I'd either delete it or retarget it at `composeCopilotMessage` — asserting a slug in `unwiredTools` never lands in the groundable list of the composed prompt would be a real assertion.

Worth saying explicitly: **the Rust tests do not have this problem.** `a_granted_but_unwired_tool_is_reported_not_offered` and the route-level `tool_slugs_omits_a_granted_but_unwired_tool_and_says_why` both assert the absence *and* that a granted-and-wired tool (`shell`) survives, so neither can pass by narrowing the list to nothing — and both would fail against the grant-only set. Those are the ones carrying the regression, and they carry it properly.

oxoxDev (changes requested):
The source-of-truth migration is **complete** — verified, and that's the crux of #874. `workflow_callable_tool_slugs` is gone along with its `company/mod.rs` re-export, both pre-PR callers (`server/ops/workflows.rs`, `harness/workflow_build/test.rs`) are migrated, and a grep at head finds no survivor. At head the only `WORKFLOW_TOOL_CATALOG` consumers are `workflow_effective_tool_slugs`, `workflow_granted_but_unwired_tool_slugs` (both through the shared `grants_workflow_tool`) and `validate_tool_call_node`'s required-args check — no third grounding path. `wired_workflow_namespaces` surviving as a wrapper over `workflow_tool_wiring` is delegation, not a stale path.

Fail-closed behaviour also checks out: `missing` is a total partition of `WORKFLOW_TOOL_NAMESPACES` (`workflows/caps/tools.rs:80-96`), so every granted tool is either offered or reported with a reason — nothing is silently dropped. The Rust route test is genuinely load-bearing: `tool_slugs_omits_a_granted_but_unwired_tool_and_says_why` drives the real route with `search: None` and asserts `web_search` absent **and** `shell` present, so it can't pass by emptying the list — it would fail against the old grant-only path.

Requesting changes for one item only; the rest are optional.

**Should fix before merge**

- `frontend/src/views/workflows/CopilotPanel.tsx:254-258` — stale `unwiredTools` across a company switch. The effect resets `roster` / `toolSlugs` / `toolSlugsKnown` but not `unwiredTools`, so during the in-flight window the prompt says "The granted tools could not be listed here" while *also* carrying the previous company's "do NOT author these" list — self-contradictory, and it bleeds one company's wiring into another's grounding. `setUnwiredTools(undefined)` alongside the others fixes it; additionally gating the block at `workflow-copilot.ts:327` on `toolSlugsKnown` closes the same hole structurally rather than by reset discipline. (Same as CodeRabbit's Major — agreed, it's real.)

**Optional / follow-up**

- `src/server/ops/workflows.rs:2121-2131` — the `_` catch-all mislabels any future `MissingReason`. Correct with today's two variants, but a third would compile silently and be reported as `capabilityTierFiltered`, telling the operator to raise a tier that isn't the problem. Since distinguishing causes is the whole point of the field, prefer an exhaustive match; it also separates the defensive `None` from the tier case, which `_` currently conflates. (= M3gA-Mind.)
- `src/company/workflow_create.rs:882-886` — the grant rule is still duplicated, and the PR body undercounts: it collapsed two of three copies. `validate_tool_call_node` keeps its own `if namespace == "search" { … } else { grants_cover }`, and so does `refusal_for` at `workflows/caps/tools.rs:107`. Unifying needs a namespace-level signature (`grants_workflow_namespace(grants, ns)`) with `grants_workflow_tool` as a thin wrapper, since the current fn takes `&WorkflowToolInfo`. (= CodeRabbit.)
- `frontend/test/unit/workflow-tool-slugs.test.ts:63` — `keeps an unwired tool out of the groundable set` asserts a property of its own fixture. `listWorkflowToolSlugs` is a pass-through, so this passes with the entire #874 change reverted, while its docstring claims it pins the core regression. Worth retargeting at `composeCopilotMessage` or dropping. (= M3gA-Mind.)
- Residual #874 exposure when harness deps are absent: `workflow_tool_wiring` returns `None` when `workflow_harness_deps` is unset (`runtime/builder.rs:2656` only attaches it when the openhuman arm resolves an inference source), and the route then serves the grant-only set — the pre-fix answer. It's the honest degradation and staging has deps, so #874 is closed there; worth a line in the merge note rather than a code change.
- Enforcement stays advisory: create/save validation is permissive by design, so a model that ignores the "do NOT author these" heading can still author a node that saves and fails at run. The primary bug (offering the slug in the groundable list) is fixed; the `unwired` half depends on model compliance, so the fix isn't fail-closed end to end.

**Nits**

- `frontend/src/api/workflow-copilot.ts:290` — the empty arm now says "No tools can run", but the unknown arm still says "The **granted** tools could not be listed", which is the wrong noun for an effective-set route.
- `unwired[].reason` is typed and tested but never branched on client-side (only `detail` is consumed). Fine as an API contract, though the "new tokens may appear" doc comment is unreachable while the `_` fallback always emits one of two tokens.
- `docs/modules/server/workflow-routes.md` matches the implemented shape exactly — `slugs` + `unwired[{slug,reason,detail}]`, camelCase rename is a no-op on these single-word fields, back-compat default documented. No drift.

Scoping is properly end-to-end (route + client + panel + prompt + docs + tests both sides), and `src/harness/mod.rs:+74` is a `#[cfg(test)]` `HarnessDeps` fixture lifted out of `runtime.rs` and shared with the route test — de-duplication, not dead code. The API change is purely additive and the client's `?? []` makes old-host and fully-wired hosts indistinguishable, which is the right equivalence.

oxoxDev:
Pushed the one required item from my review straight to this branch (hope that's alright — shout if you'd rather revert and do it yourself):

- `CopilotPanel.tsx` — the company-change effect now also does `setUnwiredTools(undefined)`, alongside the roster/slug resets.
- `workflow-copilot.ts` — the "do NOT author these" section is gated on `toolSlugsKnown`, so an unwired list can't be emitted while the effective list is unreadable. Closes the same hole by construction rather than by reset discipline.
- Tests: a pure one in `workflow-copilot.test.ts` for the gate, plus `workflow-copilot-company-switch.test.ts`, which mounts the panel, switches company with the tool-slug read still in flight, and asserts the composed message doesn't carry the previous company's tools. Both fail on the pre-fix code.

Commits: 5913daa0 and f1ce32c1. Gates green locally — `typecheck`, `typecheck:unit`, `typecheck:e2e`, `build`, and the full unit suite (894 passing).

Everything else in my review was optional/nit — left entirely to you, I haven't touched any of it.

PR #962, opened 2026-08-18
fix(harness): settle a not-automatable verdict as an answer, not a failure

### LINKED ISSUE #873: bug: a "better done once" verdict is recorded as a failed run and traps the card in a builder loop
## Summary

When the workflow builder answers "this is better done once", that verdict is recorded as a **failed run**, and the card it came from keeps its `workflow` deliverable. Both together trap the card: it bounces builder → To-do → builder forever, shows a red error to the operator on every pass, and can never reach the assignee who could simply do the work.

Found on staging (`marketing` tenant) on a card asking for a Claude-news blog workflow. Four attempts, all `failed`, card still in To-do, never dispatched. Its most recent run:

```
status: failed
error: "this is better done once than built into a workflow: A workflow named 'Claude News Blog'
        already exists for this exact recurring task, and the plan depends on a composio search
        tool that is not available in the roster or environment, so it cannot be built as a
        reliable repeatable graph with the current teammates."
```

The builder's judgement is **correct** — that workflow really does already exist, and the search tool really is unwired on that deployment. The defect is entirely in what happens to the answer.

Two halves, one settle path:

**1. A decision is written as a fault.** `settle_to_todo` (`src/harness/workflow_build.rs:612`) ends with `finish_run(..., RunStatus::Failed, Some(&reason), usage)` at `:650`. `BuildOutcome::NotAutomatable` reaches it through the same door as a model timeout or a state-read failure (`:397-408`), so an honest "don't automate this" is indistinguishable from a crash — to the console, to the attempt history, and to anything counting failures. The card's note reads sensibly; the run row next to it says the run failed.

**2. Nothing converts the card, so it cannot proceed.** The settle clears the proposal and returns the card to To-do (`:637-641`) but leaves `deliverable: workflow`. Dispatch routes any `workflow`-deliverable card to the builder rather than to its assignee (`src/company/runtime.rs:716-737`), so the next dispatch re-enters the builder, gets the same verdict, and fails the same way. The verdict says the work should be done once; the board has no path to doing it once.

## Reproduction

1. On a company with a workflow already saved for some recurring task, send a chat message asking for that same workflow with the composer set to **"Build me the workflow"** (or open a card and set its deliverable to `workflow`).
2. Let the card reach In Progress (the planning pass does this on its own).
3. The builder settles it back to To-do with a "better done once" note.
4. Open the card: the attempt shows as a failed run with the verdict as its error.
5. Move it to In Progress again (or press Retry). Steps 3–4 repeat, indefinitely.

Observed on `marketing.staging.opencompany.work`, card `019ff5e8a724-000000000002`, attempts 1–4 between 2026-08-12 and 2026-08-15.

## Expected behavior

- A `NotAutomatable` verdict settles its attempt as a **completed** run, not a failed one. A build that could not be *attempted* (timeout, unreadable state, model error) stays `Failed`; a build that ran and concluded "don't automate this" is a result. The reason belongs on the card either way.
- The card comes out of the verdict able to move. Either flip its deliverable to `once` so the next dispatch reaches the assignee named on it, or move it somewhere that requires a person's decision — but not back into the only column whose next transition re-runs the builder that just declined.
- Re-running the builder on a card it has already declined, with no change to the card or the company's workflows, should not be the default outcome of a retry.

M3gA-Mind:
## Summary

A `NotAutomatable` verdict from the workflow builder was recorded as a **failed run**,
and the card kept its `workflow` deliverable. Together those trapped the card:
`dispatch_task` routes a `workflow`-deliverable card to the builder pass rather than to
its assignee (`src/company/runtime.rs:755`), so a declined card returned to To-do
re-entered the builder on its next dispatch, drew the same verdict, and failed the same
way — builder → To-do → builder, a red error to the operator every pass, and no path to
the person who could simply do the work.

The builder's judgement in the reported case was **correct** — the workflow really did
already exist and the search tool really was unwired. So this is a legitimate, terminal,
*successful* outcome of the builder: a decision, not a crash. The defect was entirely in
what happened to the answer.

This fixes all three halves named in #873:

1. **Not recorded as a failed run.** A new `settle_not_automatable` settles the attempt
   `Succeeded` with **no** `error` on the run row; the reason goes on the card note.
   `settle_to_todo` becomes the fault-only door. This matches the convention the
   sibling draft-from-description endpoint already follows — `src/server/ops/workflows.rs:1608`
   answers **200** for this same verdict, documented as "neither is an error the operator
   fixes by retrying differently".
2. **The `workflow` deliverable is cleared.** The verdict converts the card to
   `TaskDeliverable::Once`. This is the loop-breaker: the write is ordered after
   `claim_settle`, which guards on `deliverable == Workflow`, so the guard still sees the
   deliverable the pass was started for and only this write changes it.
3. **The card reaches someone who can act.** As `once`, the next dispatch falls through
   to an ordinary dispatch on the card's `assignee`, which the conversion preserves — the
   card becomes plain one-off work instead of a build request nobody can satisfy.

### Why `BuildOutcome::NoAnswer` exists — a regression this change had to avoid

This is the least obvious part of the diff, so stating it plainly.

`BuildOutcome::NotAutomatable` was absorbing **two different things**: a reasoned refusal
("don't automate this, because…"), and a draft that parsed but decided nothing at all —
no graph, no reason, no refusal, e.g. the model returning `{"automatable":true}`.

Collapsing those two was **harmless while both settled `Failed`**: the card went back to
To-do with a reason either way, and nothing downstream distinguished them. It stops being
harmless the moment a verdict starts *converting the card and filing a success*. Without
a split, an empty or malformed model response would flip a card to `once`, hand it to a
human as though the builder had deliberated, and record the attempt as successful — a
worse bug than the one being fixed, and a silent one.

So `NoAnswer` is split out and keeps the old behaviour exactly: settles `Failed`, and the
card stays `deliverable: workflow`, because retrying the **build** is the right next move
for a fault. Only a genuine decision converts the card. `BuildOutcome` is private to
`workflow_build.rs`, so the new variant has no callers outside this file.

### How the routing claim was verified

The three fixes were each checked against **current `main`**, not inferred from the
original working notes:

- `src/company/runtime.rs:755` dispatch-routes on exactly `task.deliverable ==
  TaskDeliverable::Workflow`, and everything else falls through to an ordinary dispatch on
  `task.assignee`. **That is what makes the `once` flip the loop-breaker** rather than a
  cosmetic field change — it is the single condition standing between the card and its
  assignee.
- `TaskDeliverable::Once` (`src/ports/tasks.rs:790`) is documented as "the card dispatches
  to its assignee", so the converted card lands where the verdict says it should.
- `claim_settle` (`src/harness/workflow_build.rs:601`) guards on `deliverable == Workflow`,
  and the conversion is ordered *after* it — so the guard still sees the deliverable the
  pass was started for, and only the settle changes it.

## API Or Behavior Changes

- A not-automatable verdict now settles its attempt **Succeeded** (was `Failed`) and files
  **no** `error` on the run row. Anything counting builder failures will see fewer.
- Such a card's `deliverable` flips `workflow` → `once`. It still lands in To-do; that
  column was only a trap while the deliverable said `workflow`.
- The card note now also states what became of the card, not only the verdict.
- Unchanged: the human `workflow-proposal/reject` endpoint still keeps the `workflow`
  deliverable, so an operator dragging a rejected card back into In Progress rebuilds
  deliberately. That is an explicit human action, not a loop.

## Tests

Scoped to the gated lane, since default features skip this code.

- [x] `cargo fmt --all -- --check` — clean; changed nothing in the touched files.
- [x] `cargo clippy --no-deps -p opencompany --features openhuman,tinycortex --all-targets -- -D warnings`
      — mirrors CI's `--no-deps`; no findings in the touched files.
- [x] `cargo check -p opencompany --features openhuman,tinycortex --tests` — clean.
      Ran `check --tests` rather than `build --all-targets`; it type-checks the same targets.
- [x] `cargo test -p opencompany --features openhuman,tinycortex workflow_build` — 54/54 pass.
      Scoped to this module rather than the whole suite. Needs `RUST_MIN_STACK=16777216`:
      the unrelated `a_cap_hit_folds_to_not_automatable` overflows its stack at the default
      size, which I confirmed reproduces identically on unmodified `main` — pre-existing and
      independent of this change.

### Every new test proven to fail with its fix reverted

Four separate reverts, each run on its own:

| Reverted | Test that fails |
|---|---|
| the `deliverable = Once` write | `a_not_automatable_verdict_converts_the_card_so_it_stops_re_entering_the_builder` |
| the `Succeeded`/no-error settle | `a_not_automatable_verdict_files_no_error_and_says_what_happened_to_the_card`, `a_not_automatable_answer_returns_the_card_to_todo` |
| the `NoAnswer` split | `a_genuine_build_failure_still_fails_and_stays_builder_routed`, `the_outcome_resolves_graph_versus_not_automatable` |
| the note's "what happened to the card" clause | `a_not_automatable_verdict_files_no_error_and_says_what_happened_to_the_card` |

## Documentation

`docs/spec/runtime/workflow-build.md` — the exit table documented the old behaviour
(both non-graph outcomes as `Failed`). Updated to four exits, with the attempt status and
deliverable per outcome, and why a fault deliberately stays builder-routed.

Closes tinyhumansai/opencompany#873

### DIFF
--- a/docs/spec/runtime/workflow-build.md
+++ b/docs/spec/runtime/workflow-build.md
@@ -39,15 +39,34 @@ and it stays off the wire so a one-off card is byte-identical to a pre-#580 card
 | **Cost** | one `SampleKind::Inference` sample, charged to the card's **assignee**, carrying the attempt's `run_id` |
 | **Run row** | one — building the workflow *is* the card's In-Progress work |
 | **Locks** | none |
-| **Exit** | automatic, three-way (below) — the card never stays in `in_progress` |
+| **Exit** | automatic, four-way (below) — the card never stays in `in_progress` |
 
-### The three exits
+### The four exits
 
 | Outcome | Landing | Card carries | Attempt |
 |---|---|---|---|
 | a graph that could be created | `in_review` | the `TaskWorkflowProposal` awaiting approval | Succeeded |
-| the plan is not automatable | `todo` | the model's reason, no proposal (decision D2c) | Failed |
-| the pass itself failed | `todo` | the reason only, no proposal | Failed |
+| the plan is not automatable | `todo` | the model's reason, no proposal (decision D2c), and `deliverable: once` | Succeeded |
+| the model decided nothing | `todo` | the reason only, no proposal; stays `deliverable: workflow` | Failed |
+| the pass itself failed | `todo` | the reason only, no proposal; stays `deliverable: workflow` | Failed |
+
+A **not-automatable verdict is an answer, not a fault** (issue #873). The builder
+was asked whether the work should be automated and it decided; that settles
+**Succeeded**, with the reason on the card note and **no** `error` on the run row
+— the same convention the [draft-from-description endpoint](../../../src/server/ops/workflows.rs)
+already follows in answering `200` for the identical verdict.
+
+The verdict also **converts the card to `deliverable: once`**. This is what lets
+the card proceed: dispatch routes a `workflow`-deliverable card to this pass
+rather than to its assignee, so a declined card left carrying `workflow` would
+re-enter the builder on its next dispatch, draw the same verdict, and fail again
+— builder → To-do → builder, with no path to the person who could simply do the
+work. As `once`, the next dispatch reaches the card's assignee.
+
+A build that could not be *attempted* — an unreadable company state, a model
+timeout or error, or a draft that parsed but decided nothing — keeps
+`deliverable: workflow` on purpose, because retrying the **build** is the right
+next move for a fault.
 
 An operator who moves the card out from under the pass wins: the pass discards
 its result and the attempt settles **Cancelled** — the tokens stay metered
--- a/src/harness/workflow_build.rs
+++ b/src/harness/workflow_build.rs
@@ -429,18 +429,37 @@ pub async fn run_workflow_build_pass(
     };
     record_usage(&runtime, &builder, &agent, &run_id, &usage).await;
 
-    // The model's two answers: a graph, or "this is not automatable".
+    // The model's answers: a graph, a reasoned "this is not automatable", or
+    // — split out by #873 — a draft that decided nothing at all.
     let spec = match draft.into_outcome() {
         BuildOutcome::NotAutomatable(reason) => {
-            settle_to_todo(
+            // Issue #873: a verdict, not a fault. `settle_not_automatable`
+            // settles the attempt Succeeded and converts the card to a `once`
+            // deliverable so the next dispatch reaches its assignee instead of
+            // re-entering this pass to draw the same conclusion again.
+            settle_not_automatable(
                 &runtime,
                 &task_id,
                 token,
                 &run_id,
-                &format!("this is better done once than built into a workflow: {reason}"),
+                // The operator-facing half of #873: the note has to say what
+                // happened to the card, not only what the builder concluded.
+                // Before this, the card read as a verdict while the run row next
+                // to it read as a failure, and nothing said the card had been
+                // converted or where it goes next.
+                &format!(
+                    "this is better done once than built into a workflow: {reason} — this card is \
+                     now a one-off for its assignee to do by hand, not a workflow to build"
+                ),
                 usage,
             )
             .await;
+            return;
+        }
+        // Parsed, but decided nothing — a fault, so it keeps the Failed settle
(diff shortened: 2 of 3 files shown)


M3gA-Mind:
@coderabbitai review

PR #964, opened 2026-08-18
fix(store): give sqlite a sibling-name guard on workspace create (#894)

### LINKED ISSUE #894: store(sqlite): no sibling-name uniqueness guard, so a publish race can poison every descendant path
Spun out of #887, which is scoped to Phase A + Phase C and does not touch this.

## What is missing

Two nodes can share `(company_id, parent_id, name)` on the sqlite backend. The other two backends both stop that:

- **fs** — `reject_path_collision` runs under the same index lock every write to that index takes.
- **mongodb** — two partial unique indexes on `(company_id, file_path_key)` and the folder equivalent (`src/store/mongodb.rs`), so the server refuses the loser.
- **sqlite** — `workspace_nodes` is `PRIMARY KEY (company_id, id)` and nothing else. There is no equivalent of `UNIQUE(company_id, parent_id, name)`.

## Why it bites

`workspace_create` is check-then-act: it reads the path index, confirms the path is free, then calls `WorkspaceStore::create`. On fs and mongodb the store refuses the loser of a race. On sqlite both callers pass the check and both insert.

From then on `render_path` yields the same path for both, so:

- `workspace_read` on that path fails with `ResolveError::Ambiguous` — correctly, since silently picking one is how the wrong operator-owned note gets quoted; while
- `workspace_list` / `entries_under` still lists them, because listing does not have to choose.

One duplicated **ancestor folder** poisons every descendant path the same way, not just the duplicated node.

That read-fails-while-list-succeeds asymmetry is one of the two candidate causes #887 enumerated for a live turn. #887 makes the cause legible (the refusal now reaches the operator instead of being replaced by a generic sentence) and fixes the other candidate (the fs torn read); it deliberately leaves this one alone.

The existence of `src/company/workspace_repair.rs` — "fold the duplicate sibling folders a publish race already left behind" — is evidence the condition is reached in practice rather than only in theory.

## Suggested direction

1. Add the sqlite-side uniqueness guard equivalent to mongodb's two partial indexes (files and folders keyed separately, matching `path_key`'s `{parent}\0{name}` shape and its NUL-separator argument).
2. Route `workspace_create` off check-then-act onto a primitive the store decides, the way `adopt_or_create_folder` was routed for #759.
3. Extend the port-conformance suite with a contention case so all three backends are held to it, rather than fs and mongodb having it and sqlite not.

Migration needs care: existing sqlite databases may already hold duplicates, so the index cannot simply be added — `workspace_repair` is the fold-first step.

M3gA-Mind:
## Summary

Two nodes could share `(company_id, parent_id, name)` on **sqlite**. `workspace_nodes` is
`PRIMARY KEY (company_id, id)` and nothing else, so `create` checked only that the node **id** was
fresh. fs refuses the loser under its index lock and MongoDB against a partial unique index; sqlite
refused nothing. `workspace_create` is check-then-act, so two callers both pass its check and both
insert — and from then on `render_path` yields one path for two nodes: `read` answers `Ambiguous`
while `list` still shows both, and one duplicated **ancestor folder** poisons every descendant path.

The damage outlives the race, which is what makes this data integrity rather than a UI defect.

## Why there is no index, no migration, and nothing that can fail on boot

**Taking this first, because it is the reviewable decision and "why not just add
`UNIQUE(company_id, parent_id, name)`?" is the obvious first question.**

`CREATE UNIQUE INDEX` over a table that **already holds duplicates fails**, and it fails *during
index setup* — so store startup goes down for that tenant. That is not hypothetical: PR #733
documented exactly this failure mode for MongoDB and is why its guard is a **partial** index, keyed
on a field only new writes carry, deliberately ignoring history it cannot repair.

The sqlite equivalent of that partial index needs a `file_path_key` column maintained at **four**
sites — `create`, `create_binary`, `rename_move`, `swap_files`. A key left stale by a rename would
keep guarding the path the file *left*, refusing that path to every later write forever. That is a
worse and more permanent failure than the race being fixed, and MongoDB's own `rename_move` comment
says so about the folder key for the same reason.

**And `workspace_repair` cannot pre-fold the way in.** A "fold first, then add the index" migration
is unavailable: that pass folds duplicate **folders** and explicitly refuses to resolve *two files at
one path* — which is precisely the shape a file-uniqueness index would trip over.

So this PR adds **no schema change and no index**. A database already holding duplicates opens and
serves exactly as it does today (`read` still `Ambiguous`, `list` still shows both, `workspace_repair`
still the folder-only fold). The guard refuses **new** collisions only.

**Residual, stated plainly:** a *future* `INSERT` path added outside `create` would not be caught by
a transaction the way it would by a table constraint. That is the cost of this trade and it is
deliberate.

## Why the transaction, and why this is reuse rather than invention

sqlite did not "miss" #733's treatment — **its constraint model made an index awkward, and #733/#759
already worked around that.** Name and parent live inside the `node_json` text blob, so there is no
column to constrain. Both issues gave sqlite the **transaction** device
(`TransactionBehavior::Immediate`) and gave MongoDB the **index** device. Plain `create` is simply
the one entry point neither issue routed — `adopt_or_create_folder`'s own header says it outright:

> plain `create` checks only that the node **id** is fresh and has never had a sibling-name guard at all.

So `create` and `create_binary` now take the write reservation **before** the read, exactly as
`adopt_or_create_folder` and `swap_files` already do in the same file.

**One `SqliteStore` cannot reproduce the race**: `conn()` holds a `StdMutex` across the read and the
`INSERT`, so two tasks on one store serialize and the old code passes. The reachable shape is **two
stores over one database file** — named in `adopt_or_create_folder`'s header and in the
`busy_timeout` note in `apply_pragmas`. `IMMEDIATE` makes the loser wait for the winner's commit and
then observe it; `busy_timeout=5000` makes that a bounded block rather than an immediate
`SQLITE_BUSY`.

`create_binary` needs the same treatment: a binary node is a file and contends for a name
identically — MongoDB stamps the same path key there for that reason.

## Files only, and the divergence this surfaced

The rule enforced is **file-vs-file**. Folder-vs-folder stays `adopt_or_create_folder`'s to *adopt*
rather than refuse (#759).

File-vs-folder is deliberately **left unasserted**, and finding out why was worth the round trip: the
conformance case first asserted that a folder may take a file's name, and **fs failed it**. fs's
`reject_path_collision` compares the rendered path and refuses it; MongoDB keys files and folders
under separate partial indexes and permits it. They genuinely disagree and always have. That
assertion was stating a *permission* rather than a *rule* — pinning either answer would have forced
one backend to change behaviour, either loosening the strictest one or widening a race fix into a new
tree rule. The suite now states the rule all three must keep and is explicit about staying silent on
the rest.

The port contract for `create` previously mentioned no sibling rule at all, so sqlite was
*contract-conformant*. It now states the rule, that it is a **store** guarantee — a caller-side check
narrows the window, never closes it — and that pre-existing duplicates are kept rather than repaired.

## API Or Behavior Changes

- **`WorkspaceStore::create` / `create_binary` on sqlite** refuse a **file** whose `(parent, name)` a
  file already holds, with `Conflict`. Previously accepted silently.
- **MongoDB** maps its duplicate-key violation on those two inserts to `Conflict` instead of
  `Store("mongodb error: E11000 …")` — a contract outcome that was reaching callers as a 500 carrying
  a driver string, where fs and sqlite both answer `Conflict`. The users-email index has been mapped
  this way since it was added.
- New shared helpers `file_name_taken` / `duplicate_file_refusal` in `ports::workspace`, so the three
  backends cannot drift on what "taken" means or how they word the refusal.
- No schema change, no migration, no index. Existing data is untouched.

## Tests

- **`two_stores_racing_one_name_have_one_winner`** (`store::sqlite`) — the race itself: two
  `SqliteStore`s over one temp file, barrier-synchronised so neither enters `create` until both have
  arrived. Asserts one `Ok`, a `Conflict` loser, and — the real assertion — **one** node in the tree
  with that name. A passing error code with two rows would still be the bug.
- **`assert_workspace_sibling_names`** (`store::conformance`) — the contract, wired into **all three**
  backends. Sequential on purpose: it holds every backend to the rule without depending on scheduling.

**Teeth proved by reverting, not asserted.** With both guard blocks removed,
**2 of 2 new sqlite tests fail**, the race test with exactly the defect:

```
assertion `left == right` failed: exactly one create may take the name, got [Ok(()), Ok(())]
  left: 2
 right: 1
```

Both new tests were confirmed to *actually run* by name — a filter that selects nothing exits 0, and
`store::sqlite` is behind `--features sqlite` (`scripts/ci/feature-lanes.txt:56`), so the default
lane cannot execute them.

- [x] `cargo clippy --all-targets -- -D warnings` — clean. Run in **both** forms: CI's default-feature
      form verbatim, and `--locked --no-deps --features sqlite --all-targets`, which is the one that
      actually compiles `store::sqlite`.
- [x] `cargo build --all-targets` — N/A as a separate step: the `--all-targets` clippy and the test
      compiles below build the same targets, including under `--features sqlite`, which a default
      `build` cannot reach.
- [x] `cargo test` — default lane green. Plus the lanes that reach this code:
      `--features sqlite --lib store::` **162 passed / 0 failed**; `--lib store::fs_ops` 24 passed;
      and the MongoDB arm run against a **live local mongod** (`OPENCOMPANY_TEST_MONGODB_URI`),
      **39 passed** — genuinely executed rather than skipped, which matters because that lane
      no-ops silently without a server.

## Documentation

- `WorkspaceStore::create`'s port doc now carries the sibling rule, the store-vs-caller distinction,
  the folder exemption, and the pre-existing-duplicates statement.
- `file_name_taken` documents why the rule is files-only.
- `assert_workspace_sibling_names` documents the fs/MongoDB divergence and why the suite stays silent
  on file-vs-folder.
- No `docs/spec/` change: the storage spec describes backends in terms of the port contract, which
  this tightens rather than reshapes.

Closes tinyhumansai/opencompany#894

### DIFF
--- a/src/ports/workspace.rs
+++ b/src/ports/workspace.rs
@@ -248,6 +248,37 @@ pub fn folder_claim_ambiguous_refusal(name: &str, count: usize) -> String {
     format!("{count} nodes under this folder are named `{name}`, so the path is ambiguous")
 }
 
+/// Whether a **file** already occupies `(parent, name)` (issue #894).
+///
+/// The sibling-uniqueness predicate every backend's [`WorkspaceStore::create`]
+/// applies, shared so the three cannot drift on what "taken" means.
+///
+/// # Files only, and that is the rule rather than an omission
+///
+/// A folder is not consulted and does not block, so a folder and a file may
+/// still share a name in one parent exactly as they always could. Folder-vs-
+/// folder claims are [`adopt_or_create_folder`]'s job (issue #759), which
+/// adopts rather than refuses. Widening this to every kind would be a **new
+/// tree rule** rather than the race fix issue #894 asks for, and it would break
+/// MongoDB, whose guard is a partial index over file documents alone.
+///
+/// [`adopt_or_create_folder`]: WorkspaceStore::adopt_or_create_folder
+pub fn file_name_taken<'a>(
+    nodes: impl Iterator<Item = &'a WorkspaceNode>,
+    parent: Option<&str>,
+    name: &str,
+) -> bool {
+    nodes.into_iter().any(|node| {
+        node.kind == NodeKind::File && node.parent_id.as_deref() == parent && node.name == name
+    })
+}
+
+/// The refusal every backend gives when a file already carries the name a
+/// [`create`](WorkspaceStore::create) asked for.
+pub fn duplicate_file_refusal(name: &str) -> String {
+    format!("workspace already contains a note named `{name}` in that folder")
+}
+
 /// One node in the workspace tree. `id` is a stable ULID; `parent_id` is `None`
 /// at the workspace root.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
@@ -460,6 +491,28 @@ pub trait WorkspaceStore: Send + Sync {
     /// Creates a node (folder or file). The node's `id` must be fresh; the
     /// `parent_id`, when set, must name an existing folder. `content` seeds a
     /// file body.
+    ///
+    /// # A file's name must be free among its siblings (issue #894)
+    ///
+    /// Creating a **file** whose `(parent_id, name)` a file already occupies is
+    /// an [`OpenCompanyError::Conflict`](crate::error::OpenCompanyError). The
+    /// predicate is [`file_name_taken`] and the refusal is
+    /// [`duplicate_file_refusal`], shared so the backends cannot disagree about
+    /// either.
+    ///
+    /// **This is a store guarantee, not a caller convention, and the difference
+    /// is the whole of issue #894.** `workspace_create` checks the path index
+    /// and then calls this — check-then-act, which two concurrent callers both
+    /// pass. Only the store can decide the loser: fs under the index lock,
+    /// SQLite inside an `IMMEDIATE` transaction, MongoDB against a partial
+    /// unique index. A caller-side check is a narrowing of the window, never a
+    /// closing of it, so nothing above this line may be trusted to enforce it.
+    ///
+    /// Folders are deliberately exempt — see [`file_name_taken`]. A tree that
+    /// *already* holds duplicates keeps them: this refuses new collisions and
+    /// repairs no history, so a store carrying pre-existing duplicates opens
+    /// and serves exactly as before (`read` still answers that path
+    /// `Ambiguous`, `list` still shows both).
     ///
     /// No `author` argument: the node arrives fully formed, so the caller sets
     /// [`WorkspaceNode::created_by`] and [`WorkspaceNode::updated_by`] on it
--- a/src/store/conformance.rs
+++ b/src/store/conformance.rs
@@ -3459,6 +3459,128 @@ pub async fn assert_workspace_folder_claims(ws: Arc<dyn WorkspaceStore>) {
 ///   a document, the agent grounds an answer in it, and nothing anywhere says
 ///   so.
 ///
+/// Sibling-name uniqueness for files, which every backend must enforce **in the
+/// store** (issue #894).
+///
+/// SQLite had no equivalent of fs's `reject_path_collision` or MongoDB's partial
+/// unique index: `workspace_nodes` is `PRIMARY KEY (company_id, id)` and nothing
+/// else, so `create` checked only that the *id* was fresh. Two nodes could share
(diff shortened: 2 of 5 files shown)


M3gA-Mind:
@coderabbitai review

PR #965, opened 2026-08-18
fix(chat): journal a desk reply against its author, not the channel it went to (#885)

### LINKED ISSUE #885: chat: desk history attributes agent replies to `operator` and operator messages to the tenant
## What happens

In a desk transcript, the company's replies are attributed to `operator`, and the operator's own messages are attributed to the tenant slug.

`GET /api/v1/companies/<id>/chat/history?desk=engineering`:

```
nikhil   | operator | Security Engineer: please code-review this diff…
operator | operator | I have the full picture now. …
```

and on a DM thread, the operator's message came back authored `marketing` (the tenant name).

## Why it matters

The transcript never records which agent spoke. On a desk with four members that is the single most useful fact in the log — it is what makes "the QA Engineer said this, the Security Engineer disagreed" reconstructable after the fact, and it is what a rehydrated transcript needs to render the right avatar and name.

Some replies do carry the agent (`channel: "product_manager"` on an approval follow-up), so the field exists and is populated on some paths and not others.

## Expected

A company reply is authored by the agent that produced it; an operator message is authored by the person who sent it.

---

M3gA-Mind:
## Summary

A desk transcript attributed the company's own replies to `operator`. On a desk with four
members the author is the single most useful fact in the log — it is what makes "the QA
Engineer said this, the Security Engineer disagreed" reconstructable — and the transcript
was recording the wrong one.

### The fault is stored, not rendered

`MessageView::project` (`src/server/chat_history.rs`) is correct on every arm and
faithfully renders what it is given. The bad data is on disk.

`CompanyEvent::AgentReply.agent_id` is documented as **"the agent that produced the
reply"**. `OutboundMessage.channel` is documented as **"the channel to emit on"** — a
*destination*. All three writers copied one into the other:

| writer | stored | correct? |
|---|---|---|
| `src/server/operator.rs` (the `/chat` POST — the main path) | `agent_id: response.channel.clone()` | ❌ |
| `src/company/runtime.rs` (approval continuation) | `agent_id: response.channel.clone()` | ❌ |
| `src/harness/brain.rs` (task-dispatch reply) | `agent_id: responder.to_string()` | ✅ |

Every operator-facing bubble is built with a hardcoded `channel: "operator"`, so the
journal permanently recorded the company's reply as authored by the operator.

That third row is the issue's own clue — *"some replies do carry the agent
(`channel: "product_manager"`)"* — explained: the one writer that used the responder
instead of the channel is the one that got it right.

**The identity was available and simply dropped.** `responder_for(chat)` is computed in
`Brain::cycle` and the operator bubble is pushed ~130 lines later without it.

### Why no test caught it

The projection tests feed `agent_id: "ceo"` — a *correct* fixture — into a *correct*
projection. Nothing asserted what the write path stores. The bug lived exactly in the gap
between them.

## The fix

`OutboundMessage` gains `agent: Option<String>` — the author, distinct from `channel`,
which is now documented as a destination and nothing else. The brain fills it from the
`responder_for` result it already had; the two broken writers prefer it and **fall back to
`channel`**, so no producer that has not been taught to set it changes behaviour.

Additive on the wire (`#[serde(default, skip_serializing_if = "Option::is_none")]`): the
POST `/chat` body and every stored record round-trip byte-identically for a producer that
sets nothing. No stored record migrates.

## Historical rows carry an unrecoverable attribution — this fix is forward-only

**The true author of a pre-fix reply cannot be recovered.** It was overwritten in place and
nothing else on disk records it. Every candidate was checked and every one fails:

| candidate | why it cannot answer |
|---|---|
| `agent_id` | the loss itself |
| `steps` | `TurnStep` has no agent field |
| `task_id` | `None` on exactly these rows — it is set on the dispatch path, the writer that was already correct |
| `parent` | names the question, never the answerer |
| `chat_id` | the desk → *today's* lead. Membership is mutable (manifest ∪ operator-added overlay), so a re-derivation against current state, silently wrong for any desk whose lead changed |
| runs | no `RunRecord` exists for a chat turn — runs are task-scoped |
| `CompressedTrace` | `{cycle_id, summary, at_millis}` — no agent |
| metering | bucketed per calendar **day** with per-agent aggregates; cannot name the author of one message |

So there is **no backfill and no migration to file**: a backfill would synthesise an author
rather than restore one. A confident wrong name in a transcript is worse than an admitted
gap — the same class of mistake as a fabricated citation. Nothing here invents attribution
for a row whose truth is gone, and the reasoning is recorded in the code beside the audit so
it is not re-litigated.

**How those rows should be *presented* is filed separately** rather than widening this PR.

## The detection query

"We do not know the blast radius" is not an acceptable end state for a data-integrity bug,
and the count needs a journal — which no test fixture and no source checkout has. So:

`GET /api/v1/companies/{id}/chat/attribution-audit` (and the single-company alias) returns

```json
{ "replies": 0, "affected": 0, "by_agent_id": {} }
```

Registered beside `chat/history`, behind the same reader gate, and returning strictly less
than it: counts and agent-id strings, never message text.

**The rule is "an `agent_id` naming no roster teammate"** — deliberately *not*
`== "operator"`. The same writer defect on another channel produces a different wrong string
(a Telegram chat id, a desk slug) and has to be counted too; a rule keyed on the literal
would report a clean company.

**One deliberate false positive, stated rather than hidden.**
`CompanyRuntime::announce_continuation_failure` journals a *system* notice as
`agent_id: "operator"` on purpose — the runtime telling the operator a continuation failed,
not an agent speaking. It is indistinguishable from a #885 row on disk, so it is counted.
The number is therefore an upper bound.

### The actual count is not in this PR, and that is a gap

The query needed a company journal and I had none — a source clone carries no data, and I
have no staging credentials. **I am not going to invent a number for a data-integrity bug.**
Against a deployed host it is one request:

```
curl -s "$HOST/api/v1/companies/$COMPANY/chat/attribution-audit" -H "$AUTH"
```

The blast radius is therefore **unmeasured, not zero**, and this says so rather than
implying otherwise.

## API Or Behavior Changes

**Wire (additive, optional).** `OutboundMessage` gains `agent`; omitted when absent. One new
read-only route, `GET {scope}/chat/attribution-audit`, on both scope forms.

**Behaviour.**
1. A desk reply is journaled against the agent that produced it, not the channel it went to.
2. Replies journaled before this keep their stored value; nothing is rewritten.
3. Every producer that sets no `agent` keeps its exact previous behaviour, via the fallback.

## Tests

- [x] `cargo fmt --all -- --check` — clean (rustfmt applied).
- [x] `cargo clippy --all-targets -- -D warnings` — run as CI does, `--locked --no-deps`, both
      the default lane and `--features openhuman,tinycortex`.
- [x] `cargo build --all-targets` — run as `cargo check --locked --all-targets` and
      `cargo check --locked --features openhuman,tinycortex --all-targets`. (`check`, not
      `build`: nothing here is a linker-visible change.)
- [x] `cargo test` — `RUST_MIN_STACK=16777216 cargo test --locked --features openhuman,tinycortex --lib`:
      **4162 passed, 0 failed, 3 ignored.** (The stack bump is a known local artifact of an
      untouched module, not a code defect.)

### Every new test was proven to fail with its fix reverted — 4 of 4

Each fix was reverted individually and the suite re-run. The four reverts isolate cleanly —
each one leaves a different test standing:

|---|---|
| the roster filter is disabled, so nothing is counted | `a_reply_authored_by_the_operator_channel_is_counted`, `any_non_roster_author_is_counted…`, `a_non_reply_event_is_neither_scanned_nor_counted` |
| every reply is counted, roster ignored | `a_reply_authored_by_a_real_teammate_is_not_counted`, `a_reply_authored_by_the_operator_channel_is_counted` |
| the rule is narrowed to `agent_id == "operator"` — the tempting wrong fix | `any_non_roster_author_is_counted_not_just_the_operator_channel` |
| `fold` scans every event, not only `AgentReply` | `a_non_reply_event_is_neither_scanned_nor_counted` |

The third row is the one worth noting: narrowing the rule to the literal channel name passes
three of the four tests, which is exactly why the fourth exists.

- **No test asserts the write path end-to-end.** The tests pin the audit's classification
  rule; the writer change is a two-line substitution guarded by the compiler and the
  `channel` fallback. A test that stood up a brain, ran a cycle and read the journal back is
  the one that would have caught the original bug, and this PR does not add it — that is an
  honest gap, not coverage.
- **The new route has no request test.** It is a thin composition of `history_viewer`,
  `store().load` and `channel_attributed_replies`, each already exercised, but the wiring
  itself is unasserted.
- **No count against real data**, for the reason given above.

## Documentation

No spec doc changes. The reasoning lives at the code it governs: `OutboundMessage`'s field
docs carry the author-vs-destination distinction that caused this, and
`channel_attributed_replies` carries the full recoverability argument beside the query that
measures the damage.

Closes tinyhumansai/opencompany#885


M3gA-Mind:
Follow-up filed as #966 — how the console should present the pre-fix replies whose author is unrecoverable. Deliberately not widened into this PR.

M3gA-Mind:
@coderabbitai review

Issue #1017, opened 2026-08-18
workflows: silent disarm on schedule edit, stale pause after re-create, and a 500 on an unparseable graph


graycyrus:
Three smaller edit-path defects that share a theme: the host changes state and the console does not say so.

## An edit that adds a cron silently pauses the workflow

`update_company_workflow` disarms the schedule (`src/company/workflow_create.rs:1244, 1280`), but the console toasts only "Workflow saved." (`frontend/src/views/WorkflowsView.tsx:995`). A small "Schedule paused" badge is the only explanation on screen.

**Fix:** when `saved.enabled === false` and the edit added a schedule, say so in the toast, with a Resume action.

## A stale pause survives delete and re-create

Delete clears the overlay and `[workflows].enabled` (`workflow_create.rs:1618-1621`) but never clears `disabled_workflows`. A workflow re-created under the same id lands disabled and never fires (`src/runtime/workflow_scheduler.rs:303`). The revision cascade on delete is best-effort and log-only (`:1643-1651`).

**Fix:** clear `disabled_workflows` on delete.

## An unparseable stored graph 500s on open, so it cannot be repaired

`GET …/workflows/{wid}` surfaces `DataInvalid`/`DataParse`, which have no arm in `ApiError::status` (`src/server/error.rs:81`), producing a 500 (`src/server/ops/workflows.rs:529-531`). The list route skips the same body without complaint (`src/company/workflow_file.rs:869-872`), so the workflow appears in the picker and cannot be opened — or repaired. `render_workflow` failure (`workflow_create.rs:1213`) is a 500 for the same reason.

**Fix:** map `DataInvalid`/`DataParse` to 400/422 and return the parse problem, so the editor can show what is wrong.

oxoxDev:
Fresh confirmation of the "stale pause after re-create" half, `smoke1`, 2026-08-18:

1. `PUT …/workflows/qa_probe_wf/enabled {"enabled": false}` → paused.
2. `DELETE …/workflows/qa_probe_wf` → 204.
3. `POST …/workflows` with a clean spec for the same id (no schedule, nothing about enablement) → 200.
4. `GET …/workflows` reports the new workflow as **`enabled: false`**, and the console header renders **"Schedule paused"** on it. The previous workflow's run history attaches to it too.

So a re-created id comes back disarmed with nothing in its own spec saying so — a scheduled workflow that silently never fires.

Issue #1007, opened 2026-08-18
workflows: a failed run leaves nothing on screen but a vanishing toast


graycyrus:
When a run fails, the console returns to its resting state. There is no drawer, no row, no error text — only a toast that disappears.

## Cause

The sync POST throws, so `setResult(res)` (`frontend/src/views/WorkflowsView.tsx:802`) never runs and `RunResultPanel` — the only surface that renders run detail — never mounts. The failure *is* journaled and refetched (`:846-849`), but the History drawer is closed by default (`historyOpen`, `:257`) and nothing opens it. All the operator gets is `toast.error(e.message)` at `:844`. Same for `connection-lost`.

## Made worse by a stale drawer

`run()` clears `overlayRun` and `optimistic` (`:771-772`) but never `result`, and `ranWith` is only set on success (`:793`). Run twice with the second failing and the drawer still shows run #1's nodes, output and "Requested:" line — presented as the new run's detail.

## Nothing acknowledges the click either

There is only a button spinner and an optimistic canvas frontier: no toast, no optimistic history row, no elapsed timer. `startedAtMillis` is already on the wire (`frontend/src/api/workflows.ts:403`) and unused by every workflow surface. The success toast arrives minutes later with nothing in between.

## Proposed fix

- On the failure path, open the History drawer and select the run just journaled (match `awaitingRunId`), or render a persistent failure panel built from the caught `ApiError`.
- `setResult(null)` beside `setOverlayRun(null)`.
- Dispatch-time `toast.info`, an optimistic running row, and elapsed time from `startedAtMillis`.
- Print the node *name* on the failure line (`RunHistoryPanel.tsx:356` prints the raw id while `NodeTimeline` maps it), and show run duration.

Issue #1005, opened 2026-08-18
workflows: the create dialog never shows it is working, and a failed create is never announced


graycyrus:
The create dialog is the only write surface in the console that never shows it is working, and a create that fails server-side is never announced.

## What the operator sees

Press **Create workflow** and the only feedback is the button label swapping to "Creating…". The copilot box, the node rows, **Add node**, **Add edge** and every input stay live, so the dialog looks idle while a request is in flight. If the host rejects the draft, a banner appears somewhere in a long scrolling form and nothing scrolls to it or announces it.

## Evidence

- `frontend/src/views/WorkflowCreateDialog.tsx:1338-1355` — only Cancel, Create and Restore consult `submitting`.
- `grep -c "animate-spin\|Loader2" frontend/src/views/WorkflowCreateDialog.tsx` → **0**. Every sibling surface has one: `WorkflowsView.tsx:1516,1551,1631,1668`, `CreateTaskDialog.tsx:207`.
- `WorkflowCreateDialog.tsx:991-998` — the server-error branch sets `error` with no `scrollIntoView`/`focus`. That treatment exists only on the client-validation branch (`:913-916`).
- `submit()` (`:905`) has no re-entrancy guard; double-submit is prevented by DOM `disabled` alone.
- `error` is cleared only on open, on a successful draft, and at submit start, so a stale banner sits there while the operator fixes the problem.

## Proposed fix

- `Loader2 … animate-spin` on Create while `submitting`, and on "Drafting…"/"Restoring…".
- Lift the `requestAnimationFrame(scroll + focus)` into a `showError(message)` helper called from both `:908` and `:992`; give the banner `role="alert"` / `aria-live="assertive"`.
- `aria-busy={submitting}` on `DialogContent`; disable the copilot block and Add node/Add edge while submitting.
- `if (submitting) return;` at the top of `submit()`.
- Clear `error` in `updateNode`/`updateEdge`/`setName`.

## Tests

No component test renders `WorkflowCreateDialog` at all — the seven workflow unit specs (114 passing) only cover pure helpers. Add an RTL spec: pending state during create; a failed create keeps the dialog open with the draft intact; a double-click fires one POST.

PR #1020, opened 2026-08-18
fix(console): stop the workflow editor discarding unsaved graph edits

### LINKED ISSUE #1006: workflows: unsaved graph edits are destroyed silently, and a serialisation throw locks the dialog
Half an hour of graph edits can vanish with no prompt, through three separate paths, and a serialisation throw can lock the dialog so the only exit also loses the work.

## Paths that destroy unsaved edits

1. **Esc / outside click / Cancel** close unconditionally — `frontend/src/views/WorkflowCreateDialog.tsx:1012`, `:1339`. `isDraftDirty()` exists at `:835` but is used only for the copilot overwrite confirm.
2. **Browser Back re-hydrates the open dialog with a different workflow.** The hydration effect depends on `workflow` (`:473-557`), so `hashchange` (`WorkflowsView.tsx:394`) → new `selectedId` → refetch → `setGraph` (`:605`) swaps the graph underneath the operator. The dialog title never names which workflow it is editing (`:1015`).
3. **A failed background refetch unmounts the dialog outright** — `graph=null` (`WorkflowsView.tsx:612`) with `open={editOpen && graph !== null}` (`:1979`).

There is no `beforeunload` guard either.

## The lock

`setSubmitting(true)` happens at `WorkflowCreateDialog.tsx:919`, but the graph is assembled at `:921-968` **outside** the `try` at `:969` — and `configFromDraft` does an unguarded `JSON.parse` (`frontend/src/lib/workflow-node-config.ts:321`). Any validate/serialise disagreement throws, so `finally` never runs and `submitting` stays `true` forever. `onOpenChange` is gated on `!submitting` (`:1012`) and Cancel is `disabled={submitting}`, so the dialog cannot be closed. Recovery is a page reload, which loses the edits.

## Proposed fix

- Track dirty state properly and confirm on close, unmount and hash change; add a `beforeunload` guard.
- Freeze `selectedId` while the dialog is open, or re-open against the pinned graph.
- Move graph assembly inside the `try`; make `configFromDraft` return a typed error instead of throwing.

## Tests

`frontend/test/e2e/workflow-edit-delete.spec.ts` covers 8 cases but none of these. Add: unsaved-edit confirm on close, and a serialisation failure leaving the dialog closable.

graycyrus:
## Summary

Unsaved graph edits could vanish through three separate paths in the workflow edit dialog, and a serialisation throw could lock the dialog shut over them.

- **Dirty tracking + confirms.** The draft carries a fingerprint of everything the form can change, taken from the exact values each open hydrates with, so "has this been edited?" is one comparison against that baseline (and a re-hydrate — Reload, a History restore — resets it rather than leaving the dialog permanently dirty). Esc, an outside click and Cancel now route through one confirm; `beforeunload` guards a reload/tab close; a `hashchange` guard covers Back and the sidebar, since this console routes on the hash — that is where in-app navigation away from an open dialog actually arrives, and declining puts the address back.
- **The dialog no longer re-hydrates with a different workflow.** It binds to the graph pinned when it opened. A Back press used to move `selectedId`, and the refetch swapped a *different* workflow's graph in underneath the operator; a refetch that *failed* landed `null` and unmounted the dialog outright. A re-read of the *same* workflow still gets through, which is what keeps the conflict banner's Reload and the History restore re-hydrating as documented. The title now names the workflow being edited, so a swap would be visible at all.
- **`submit()` can no longer lock the dialog.** Graph assembly moved inside the `try` — it used to run above it, so a throw escaped past the `finally` that clears `submitting`, which disables Cancel and gates the dialog's own close handler; recovery was a page reload, i.e. losing the edit. `configFromDraft` now returns `{ ok: true; config } | { ok: false; error }` instead of doing an unguarded `JSON.parse`, so a failure names the node that caused it in the error banner and leaves the dialog closable with the draft intact.

## Tests

New `frontend/test/unit/workflow-editor-unsaved-work.test.ts` (jsdom + `react-dom/client`, the same earned exception as `setup-wizard-finish-gate`) covers: the title naming the workflow; Cancel closing silently when nothing was edited; Cancel and Esc confirming when there are edits, staying open and keeping the draft when declined; the `beforeunload` guard arming and disarming with dirtiness; and a serialisation failure reporting itself, sending nothing to the host, keeping the draft, and leaving Cancel enabled and working.

`frontend/test/unit/workflow-node-config.test.ts` gains the `configFromDraft` result-shape cases and unwraps the existing ones through a local helper.

## Commands run (from `frontend/`)

| Command | Result |
| --- | --- |
| `npm ci` | ok |
| `npm run typecheck` | pass (no output) |
| `npm run typecheck:unit` | pass (no output) |
| `npm run typecheck:e2e` | pass (no output) |
| `npm test` | 88 files, 899 tests passed |

`npm test` was re-run after the final wrapping edits; the new file is 6 tests, `workflow-node-config` 49.

No Rust touched — this is frontend only, so no cargo commands were run.

## Behaviour changes

- Closing the workflow edit dialog with unsaved changes now prompts. So does navigating away in-app or reloading the tab while it is open and dirty.
- The edit dialog title reads `Edit “<workflow name>”` rather than `Edit workflow`.
- `configFromDraft` is now a `ConfigFromDraft` result rather than `Record<string, unknown> | undefined`; the one production caller unwraps it.

## Deliberately out of scope

The other workflow defects noted alongside this one (#1005, #1007–#1017) are untouched. No Playwright spec was added — `workflow-edit-delete.spec.ts` needs a live host, and both new claims are about a mounted dialog reacting to a click, which the jsdom suite pins directly and in milliseconds.

Closes #1006


graycyrus:
Pushed two commits since the last review pass:

- `219c2960` — **fixes the failing Console lane.** All 899 tests passed, but the run failed on one unhandled error: `TypeError: errorRef.current?.scrollIntoView is not a function`. jsdom implements no scrolling at all, so `Element.prototype.scrollIntoView` does not exist; `WorkflowCreateDialog` scrolls its error banner into view from a `requestAnimationFrame` after a failed save, which puts the throw outside every test body — green tests, red run. The new serialisation test is the first thing in the suite to reach that branch. Supplying the missing method in the test file is the honest fix: the production call is correct in a browser, and guarding it in the component would only be guarding against jsdom.
- `53a1b728` — merge of `upstream/main` (which now contains #1021). Clean, no conflicts: `WorkflowsView.tsx` carries **both** the `editGraph` pinning from this PR and #1021's `RunFailurePanel`. Suite is now 952 tests across 96 files, all green.

Three design positions in this PR are deliberate, in case they come up:

1. **No literal "confirm on unmount".** React cannot cancel an unmount — a cleanup function runs after the decision is already made, so a confirm there can only ask a question whose answer it must ignore. Pinning the dialog to `editGraph` removes the spurious unmount path that was destroying the draft, and `hashchange` + `beforeunload` cover the two ways an operator actually leaves a dirty editor.
2. **The serialisation-failure test drives a mocked `configFromDraft`.** That branch means the form's validation and the serializer disagreed, which by construction no keystroke can produce. Making them disagree is the only honest way to pin "and the operator can still get out".
3. **No Playwright spec.** Both claims are about a mounted dialog reacting to a click; the jsdom suite pins them directly and a browser walk would only reach the same assertions through a live host.

Scope is #1006 only — neighbouring workflow-editor suggestions belong to #1005 and #1007-#1017.

graycyrus:
`b7dba9ef` — **fixes the three Console E2E failures**, which were real and were mine.

`test/e2e/workflow-edit-delete.spec.ts`'s `openEditDialog` helper waited for the exact text "Edit workflow". This PR deliberately replaced that with `Edit “<name>”` (the whole point of the naming change: the dialog is reachable from a canvas, a card and a run, and said the same anonymous thing from all three). So the dialog was opening correctly and the helper was watching for a string that no longer exists — every test that goes through it timed out at the first assertion.

The helper now takes the name the caller selected and asserts `Edit “<name>”`, which turns a cosmetic wait into a real check that the dialog opened on the right graph. Also added `acceptDiscard(page)` before the two Cancels that leave a *touched* form: those now raise the discard confirmation, and Playwright dismisses an unhandled `confirm`, which reads as "no, keep editing" and would have left the dialog open. Both lanes (`Console E2E` and `Console E2E (live brain)`) failed on exactly these three tests, so one fix covers both.

Local, from `frontend/`: `typecheck`, `typecheck:unit`, `typecheck:e2e`, `build` and `npm test` (952 tests / 96 files) all green.

graycyrus:
@coderabbitai review

Context for the request: this PR's CodeRabbit check has reported `pass` throughout, but the body was a rate-limit notice rather than a review — so no CodeRabbit findings exist for it. CI is otherwise 11/11 green on head `b7dba9ef` (verified bound to that SHA, not a stale rollup), and tinysweeper approved.

Worth a look at three deliberate choices, in case you disagree on the merits:
1. No literal confirm-on-unmount — React cannot cancel an unmount, so the fix pins the dialog to the graph it opened with (`editGraph`) to remove the spurious unmount path, and covers real exits via `hashchange` + `beforeunload`.
2. The serialisation-failure test drives a mocked `configFromDraft`, because by construction no keystroke can make the validator and the serializer disagree.
3. `Element.prototype.scrollIntoView` is supplied in the test file rather than guarded in the component — jsdom implements no scrolling, and the production call is correct in a browser.

graycyrus:
`3004a707` — merged `main` after #1019 (issue #1005) landed. Five conflict hunks in `WorkflowCreateDialog.tsx`, all resolved keeping **both** sides:

1. **Imports** — `useCallback` (mine, for `requestClose`) and `Loader2` (#1005's spinners) both kept.
2. **The `dirty` fingerprint + `requestClose` + `beforeunload`/`hashchange` guards** (mine) and **`clearSubmitError()`** (#1005) landed adjacent in the same block — both kept verbatim, neither is a rewrite of the other.
3. **`raise()` vs `showError()`** — the same function, renamed and re-documented by #1005 to cover the host-rejection branch as well as the client-validation one. Took #1005's version and repointed my serialisation-failure call site (`configFromDraft` returning `{ok:false;error}`) at it, so the defect message gets the same scroll-and-focus treatment as every other banner.
4. **`submit()`'s validation branch** — same rename, same resolution. #1005's `submittingRef` re-entrancy guard and its `finally` are untouched, and my "assemble the graph INSIDE the try" change still sits under them: a serialisation failure clears `submitting` on the way out, so the dialog cannot lock itself shut.
5. **The `<Dialog>` element** — kept my `onOpenChange` routing (Esc and outside-click go through the same discard confirm as Cancel) *and* #1005's `aria-busy={submitting}` on `DialogContent`.

On the collision you flagged: the `<fieldset disabled={submitting} className="contents">` still opens before the Id field and closes after the edges grid, enclosing every draft control including the ones this PR touches. The error banner stays outside it, which is right — it carries `errorRef` and must remain focusable while a save is in flight. The dirty guard is unaffected by the disable: `dirty` is computed from state, and the `beforeunload`/`hashchange` listeners are window-level, so they still fire mid-submit. `Cancel` and Esc are deliberately inert during the write (#1005's rule), which is the only exit #1006 does not need to intercept.

I read the two overlapping unit files rather than trusting the clean compile: `workflow-create-feedback.test.ts` (#1005) and `workflow-editor-unsaved-work.test.ts` (#1006) make no contradictory claims — neither cancels a *touched* form in the other's territory, and each stubs `scrollIntoView` in its own jsdom environment.

Local gate from `frontend/`: `typecheck`, `typecheck:unit`, `typecheck:e2e`, `build`, `assert-design-tokens.sh`, and `npm test` — **972 tests across 100 files, zero failures, zero unhandled errors.`

senamakel:
Pass 2 — merged the rest of `main` and pushed `eb85c42b`.

- `293bb6fa` — merge of `upstream/main` through #1093 (#1091, #1073, #1081, #1037, #1068). One conflict, in `WorkflowCreateDialog.tsx`, resolved keeping **both** sides: #1005/#1052 converge with #1006. The dirty fingerprint + `requestClose` + `beforeunload`/`hashchange` guards, the `raise()`→`showError()` rename, `submittingRef`/`clearSubmitError`, `draftEpochRef`/`draftLanding`, and the `aria-busy` spinner all land adjacent — no side is a rewrite of the other. The serialisation-failure call site (`configFromDraft` returning `{ok:false}`) repoints at `showError`, so the defect banner gets the same scroll-and-focus treatment as every other. `grep` confirms zero leftover conflict markers.
- `eb85c42b` — forward the `vendor/openhuman` pin by the single merge (`#5580`) beyond `main`'s current `0abb318b`.

Local gate from `frontend/` against the merged tree: `typecheck`, `typecheck:unit`, `typecheck:e2e`, `build`, and the full vitest suite — **105 files / 1044 tests, zero failures, zero unhandled errors**. GitHub now reports `MERGEABLE`; the fresh CI run on `eb85c42b` is under way.

graycyrus:
`5fb214a3` — unblocks the Rust lanes, which were failing on a **merge artifact, not on this PR's work**.

`src/server/ops/workflows.rs:5767` is a test double (`tests::hosted_mode::FinishesDuringTheRead`, from #1081). Upstream separately changed `EventLog::subscribe` to yield `ports::events::EventStreamItem` instead of `ports::types::StoredEvent`. `main` carries both changes and compiles; the branch's last upstream merge kept the stale return type on the double, so `cargo clippy` died with E0053 + E0308 before `Test` ever ran — which is why `Console E2E` and `Console E2E (live brain)` showed as *skipped* rather than failing: both are gated on the Rust job.

The fix is the exact text `upstream/main` already has — `git diff upstream/main HEAD -- src/server/ops/workflows.rs` is now empty, so the branch carries no Rust deviation from main at all. I did not run cargo locally (per the handover); the lane will verify.

Also re-verified the console side on the current head after that upstream batch landed. Both PRs' behaviours are intact in `WorkflowCreateDialog.tsx`: #1005's `submittingRef` guard, `clearSubmitError()` routing (18 call sites), `aria-busy`, and `<fieldset disabled={submitting} className="contents">` at 1473-1599; #1006's `dirty` fingerprint, `requestClose`, `beforeunload`/`hashchange` guards, the `Edit “<name>”` title, the result-typed `configFromDraft`, and `editGraph` pinning in `WorkflowsView.tsx`.

On the fieldset collision specifically: it opens after the copilot block and closes at 1599, enclosing the whole draft region (id, name, description, node rows, edge rows) — so #1005's lock still covers everything, including the rows this PR touches. **Cancel sits in `DialogFooter` at 1696, outside the fieldset**, gated only by its own `disabled={submitting}`. That is the arrangement both issues need: the exit is inert only while a write is genuinely in flight, and #1006's change to assemble the graph inside the `try` is what guarantees that window always ends — a serialisation failure now returns through the `finally`, so `submitting` cannot latch and lock the operator in. `workflow-editor-unsaved-work.test.ts` asserts exactly that (`expect(cancel.disabled).toBe(false)` after the failure, then Cancel closes).

I also read the two overlapping unit files rather than trusting the clean compile. They are complementary, not contradictory: #1005 pins "every draft control is dead *during* a live request"; #1006 pins "after a failure the guard is released and Cancel works". Neither cancels a touched form in the other's territory.

Local gate on this tree: `typecheck`, `typecheck:unit`, `typecheck:e2e`, `build` — clean; `npm test` — **1044 tests across 105 files, zero failures, zero unhandled errors**.

PR #1022, opened 2026-08-18
fix(harness): say so when a turn stops at its step budget

### LINKED ISSUE #926: chat: an agent that exhausts its step budget stops mid-task with a plan, and never says it stopped
## Summary
An agent turn that runs out of steps ends by printing what it *was going to do next*, with no indication that it hit a limit and no follow-up. To the user this is indistinguishable from the agent losing interest — and it is what QA has been reporting as "agents get permanently stuck mid-task; replies stop at 'Let me check X' and never follow up".

## Repro
DM the Product Manager: "Write a short feature spec for adding CSV export to the usage page."

## Observed
The turn ran to `20 steps · 2 failed` and ended with:

> **2. Next steps**
> Retrieve the full client.ts … Attempt to locate the actual Usage page component … **Write the spec to a sandbox file and deliver it with `publish_artifact`.**

No spec was written. No artifact was published. No further message arrived. The turn simply ended at step 20 — a round number that strongly suggests a step cap — while the agent still believed it had three things left to do.

The work up to that point was good: it read `Standards/Engineering standards.md`, the release checklist, and the closest prior export spec, and correctly flagged that a truncated 12KB-of-30KB workspace read had hidden the fields it needed. It was making progress when it was cut off.

## Expected
- When a turn ends because it hit its step budget, say so plainly in the reply.
- Prefer delivering a partial artifact over delivering a plan to write one.
- Consider whether the budget should extend, or the turn should continue, when the agent is mid-plan rather than idle.

## Contributing factor
Two of the twenty steps were burned on tools that cannot run here at all — see the `grep` / `read_skill_resource` issue filed alongside this one.

## Environment

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#926.

A turn that runs out of steps ends by printing **what it was about to do next**, with no
sign it hit a limit and no follow-up. QA has been reporting this as *"agents get
permanently stuck mid-task; replies stop at 'Let me check X' and never follow up."*

### The cap is 10 model calls — not the 20 in the repro

The repro ended at `20 steps · 2 failed`, and the round number is a red herring worth
naming, because it is the first thing anyone will reach for.

A console "step" is one tool call or one coalesced thinking run — `fold_steps`
(`src/harness/steps.rs`) explicitly discards iteration and cost events. **Steps are not
model calls.** The real limit is `AgentConfig::max_tool_iterations`, openhuman's default
of **10** (`vendor/openhuman/src/openhuman/config/schema/agent.rs:413`). It is not a
deadline and not a sub-agent budget. `build_agent` never calls an iteration setter, so it
inherits the default.

That is confirmed by running it, not by arithmetic: the new end-to-end test drives a real
capped turn and the harness reports `ExhaustedStepBudget { cap: 10 }`, with openhuman's
own fallback text reading *"I reached the tool-call limit for this turn (10 steps)"*.

The cap is also **not a single constant** — `harness/workflow_build/agent.rs:128` already
calls `set_max_tool_iterations(7)`. That is why the notice reads the cap off the agent that
enforced it rather than hard-coding a number on this side.

### Why it is silent, precisely

openhuman does not error on a cap hit. At `session/turn/core.rs:1366` it asks the model for
a resumable checkpoint with tools disabled and returns that prose through **the same
`Ok(String)` a finished turn returns**. Upstream's own field doc is explicit: `run_single`
"only returns the checkpoint/final text, **with no other signal** for which case occurred."

The decisive detail is the instruction itself (`turn_checkpoint.rs:45`). It asks the model
for "**Done so far**" and "**Next steps**" and **never asks it to tell the user a limit was
hit**. That is verbatim the repro's *"2. Next steps → …"*. Only openhuman's *deterministic
fallback* checkpoint names the cap, and that runs solely when the model's wrap-up call
fails or returns empty. So the silence is conditional on which checkpoint path ran — and in
the repro the model path ran. That is exactly why this has to be a system-emitted fact
rather than a prompt tweak asking the model to mention it.

### What this does

`Agent::last_turn_hit_cap()` is public and purpose-built (openhuman consumes it itself at
`flows/ops.rs:7111`). It is read in `CompanyAgent::run_with_steer` while the agent lock is
still held — same place and same reason as `last_turn_usage`, since both are per-turn state
the next turn overwrites — and becomes
`TurnOutcome::exhausted_budget: Option<ExhaustedStepBudget>`, threaded through
`OperatorTurn` and rendered by `harness::brain`.

**The cap is not raised.** Raising it moves the cliff and keeps the silence.

`ExhaustedStepBudget` is a type rather than a `bool` for the reason `DrainedRequests`
carries its own `cap` (#561): the notice quotes a number, and a caller holding a bare
`true` would have to find that number somewhere else — a hard-coded copy of openhuman's
default, or the round figure off a step timeline. Either hands the operator a
confidently-worded, wrong sentence about a limit the turn was not held to.

The wording reuses `DrainedRequests::overflow_notice` (#625/#561) deliberately — same
"Heads up:" opening, same habit of quoting the limit that actually applied, same closing
move of telling the operator the one thing they can do. Two systems reporting
silently-dropped work in two different voices is how one of them stops being believed.

It is emitted as **its own operator bubble**, not appended, following the rationale already
written at the approvals-overflow site: the reply is the agent's answer, and this is the
system saying the agent was cut off. Wired on the main operator path and on the confined
copilot path, which bypasses the delegation seam.

Non-loop `TurnOutcome` sites (budget refusals, ACP) set `None` explicitly — an ACP turn runs
on an external agent's loop and must not claim a cap it was never held to.

## API Or Behavior Changes

New public type `harness::ExhaustedStepBudget`; `TurnOutcome` gains
`exhausted_budget: Option<…>` (both behind the existing `openhuman` feature).

Behaviour: one additional operator bubble when a turn stops at its cap. No turn is gated,
refused, or run differently; an ordinary turn is byte-identical and silent.

## Tests

      exit 0. `--no-deps` mirrors `ci.yml`; a bare `--all-targets` drowns in vendored lints
      CI never compiles.
- [x] `cargo build --all-targets` — covered by the two test lanes below, which build all
      targets for both feature sets.
- [x] `cargo test` — both halves. Default features: `cargo test --locked`, exit 0. Gated
      lane: `RUST_MIN_STACK=16777216 cargo test --locked --features openhuman,tinycortex --tests`
      → **4188 passed / 0 failed**. `RUST_MIN_STACK` matches `ci.yml:626`; without it a
      pre-existing `harness::brain` test overflows the stack locally.

`scripts/ci/assert-feature-lanes.sh` and `scripts/ci/assert-integration-targets-run.sh` both
pass. No new `feature-lanes.txt` row is owed: the tests sit behind `openhuman`, already
classified `tested` on `openhuman,tinycortex`.

### Revert proof — three, each isolating a different piece

| Reverted | Result |
| --- | --- |
| harness read → hard-wired `None` | `a_turn_that_runs_out_of_model_calls_reports_the_cap_it_hit` **FAILED** |
| `OperatorTurn.exhausted_budget` → `None` | 3 delegation tests **FAILED** |
| relay `.or(responder)` → plain replacement | `a_relay_that_finished_does_not_erase_an_exhausted_responder_turn` **FAILED** |

The negative test (`…finished_normally_reports_no_exhausted_budget`) correctly keeps
passing under all three — it asserts `None`, which is what makes the notice believable.

### Two things worth stating plainly

**The end-to-end test exercises the deterministic fallback checkpoint, not the
model-written one**, because the loopback stub is not a real model. It proves the plumbing
— cap reached → `last_turn_hit_cap` → `TurnOutcome` → cap value — which is what the fix is.
The model-written variant differs only in the prose openhuman returns, which this change
does not read.

**Writing that test surfaced a real trap.** The first version repeated one tool call and
tripped openhuman's no-progress breaker after three identical batches ("the run is stuck
repeating one action"), which halts with its own good root-cause summary and *deliberately*
reports `hit_cap = false`. It would have passed while testing the breaker and never
reaching the cap. The script now alternates two distinct successful calls so the budget is
the only thing left to stop the loop, and the reason is written into the test.

## Documentation

`docs/modules/openhuman/README.md` gains a "The per-turn step budget" section: what the cap
is and is not, why steps ≠ model calls, why the reply cannot carry the fact, where it is
read instead, and why the cap is not raised.

## Summary by CodeRabbit

* **New Features**
  * Added per-turn model-call budget tracking, with a seven-call cap for OpenHuman workflows.
  * Turns that reach the cap now return a resumable checkpoint instead of an error.
  * Operators receive a separate budget-exhaustion notification while the original agent reply remains unchanged.
  * Budget status and the applied cap are preserved across delegated and relayed turns.

* **Bug Fixes**
  * Corrected budget-exhaustion reporting so normally completed and budget-refused turns are not incorrectly marked as exhausted.

* **Documentation**
  * Documented model-call budgets, resumable checkpoints, and operator notifications.

### DIFF
--- a/docs/modules/openhuman/README.md
+++ b/docs/modules/openhuman/README.md
@@ -55,6 +55,45 @@ runtime override > manifest `[inference]` > managed env default — on every tur
 so a provider switch or key rotation reaches agents on the next turn with no
 rebuild at all.
 
+## The per-turn step budget (issue #926)
+
+An agent turn is bounded by openhuman's `AgentConfig::max_tool_iterations` — the
+number of **model calls** one turn may make, defaulting to 10. It is not a
+deadline and not a sub-agent budget, and it is not the step count the console
+shows: a console "step" is one tool call or one coalesced thinking run
+(`harness::steps::fold_steps`), so a turn stopped at 10 model calls routinely
+renders as ~20 steps. `harness::workflow_build` is the one place OpenCompany
+overrides the cap (`set_max_tool_iterations(7)`); everything else inherits the
+default.
+
+**Hitting the cap is invisible in the reply, by construction.** openhuman does
+not error. It asks the model for a resumable checkpoint with tools disabled
+(`MAX_ITER_CHECKPOINT_INSTRUCTION`) and returns that prose through the same
+`Ok(String)` a finished turn returns. The instruction asks for "Done so far" and
+"Next steps" and never asks the model to mention the limit, so a capped turn
+typically ends on a confident plan — indistinguishable from an agent that
+finished and described what it would do. Only openhuman's deterministic
+*fallback* checkpoint (used when the model's wrap-up call fails or comes back
+empty) names the cap.
+
+So the fact is carried out of band. `Agent::last_turn_hit_cap()` is read in
+`CompanyAgent::run_with_steer` while the agent lock is still held — the same
+place and for the same reason as `last_turn_usage` — and becomes
+`TurnOutcome::exhausted_budget: Option<ExhaustedStepBudget>`. The cap it quotes
+comes off the agent that enforced it, never a constant on this side, because
+callers may override it per agent.
+
+`harness::brain` renders it as **its own operator bubble**, not appended to the
+reply: the reply is the agent's answer and this is the system saying the agent
+was cut off. The wording deliberately mirrors
+`DrainedRequests::overflow_notice` (issue #561) — same "Heads up:" opening, same
+habit of quoting the limit that actually applied, same closing move of telling
+the operator the one thing they can do. Two systems reporting silently-dropped
+work in two different voices is how one of them stops being believed.
+
+The cap itself is **not** raised. Raising it moves the cliff and keeps the
+silence, which is the failure this fixes.
+
 ## Approval parking
 
 openhuman resolves a `ToolPolicyDecision::RequireApproval` **inline**: it blocks
--- a/src/harness/acp_run_turn.rs
+++ b/src/harness/acp_run_turn.rs
@@ -199,7 +199,14 @@ pub fn fold(turn: AcpTurn) -> TurnOutcome {
         }
     }
 
-    TurnOutcome { reply, steps }
+    TurnOutcome {
+        reply,
+        steps,
+        // ACP turns run on the external agent's own loop, which reports no cap
+        // to us. Claiming one would be inventing a limit this turn was never
+        // held to.
+        exhausted_budget: None,
+    }
 }
 
 #[async_trait]
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -2540,6 +2540,7 @@ impl HarnessBrain {
                                 &confinement,
                             )
                             .await?;
+                        let exhausted = outcome.exhausted_budget.clone();
                         channel_responses.push(OutboundMessage {
                             message_id: None,
                             // No card, by construction: a confined turn has no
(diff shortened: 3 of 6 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Closing as a **duplicate of #996** (merged `08c4a9b5`, 11:20 UTC today), which landed the same fix at the same seam before this PR was opened. Our fault for not checking; nothing here needs reviewing.

Having read both, **#996 is the better implementation** and I want to be specific rather than gracious about it:

1. **It deliberately does not quote the cap**, and its reason is stronger than ours. From `mod.rs`: *"one notice can cover a responder turn, a desk turn and a relay turn, and naming one of their caps would be a number the operator cannot map back to anything."* Ours carried the cap, modelled on `DrainedRequests` — but a drain is one queue against one cap, whereas a chat turn can span three turns with different caps. That analogy was weaker than we thought.
2. **It reads the cap from the live `IterationStarted` progress event, not from config** — the number the turn actually ran under, rather than the one the crate believes it configured.
3. **It unions across all three turns** (`hit_iteration_cap |= desk`, `|= relay`). Ours used `.or()` across responder and relay and **missed the desk turn entirely** — a real hole, and our own relay test would not have caught it.

It also declines to promise resumption, because `HarnessPool::ensure` rebuilds the pooled agent on a roster/skill/MCP fingerprint change — a durability caveat we had not considered.

**One gap survives #996**, and we will send it as a small follow-up using #996's own `ITERATION_CAP_PAUSE_NOTICE` constant and design: the confined workflow-copilot path (`run_confined`, `brain.rs:2648-2674`) pushes its reply and `continue`s with no cap check, so a copilot turn that pauses is still silent. Roughly ten lines against current `main` — much better than rebasing this branch 214 commits and reintroducing a competing type.

Issue #811, opened 2026-08-12
console: every select popup is pinned to its trigger's width, so option labels are clipped mid-word


oxoxDev:
## Summary

`SelectContent` sets the popup to `w-(--anchor-width)` — an exact match to the
trigger's width. The triggers are `w-fit`, so a trigger showing a short value
(`trigger`, `none`, `agent`) produces a popup that is too narrow for its own
options, and every label is cut mid-word.

## Where it bites hardest

The workflow **create** and **edit** panes, where the option labels are the
documentation — each one explains what the node kind does, and the explanation
is the half that gets cut:

```
Trigger — starts the w…        Agent — a teammate…
Condition — branches…          Merge — combines s…
Transform — reshape…           Output — reports the…
```

The schedule picker on a trigger node has the same problem:

```
No schedule (run man…          Weekly — Monday 09…
```

An operator picking a node kind for the first time is reading exactly the text
that is missing.

## Not confined to workflows

The rule lives in the shared primitive (`frontend/src/components/ui/select.tsx`),
so it applies to all 25 `<SelectContent>` sites — tasks, people, skills, memory,
inbox, artifacts, usage, connections, domain settings, feedback. Workflows is
where it is most visible because its labels are the longest, not because it is
the only one affected.

## Cause

`frontend/src/components/ui/select.tsx`, the `SelectContent` popup className:

```
w-(--anchor-width)
```

`--anchor-width` is the trigger's width. Binding `width` to it makes the popup
exactly as wide as the trigger and no wider, whatever the options need. The
convention this component follows elsewhere is a *minimum*, not an exact match:
the popup should never be narrower than its trigger, and should be free to grow
to fit its content.

## Fix

Make it a floor with a ceiling, rather than an exact width:

```
min-w-(--anchor-width)   max-w-[min(28rem,var(--available-width))]
```

Verified live against `smoke1` on `staging-d0f908bb` by injecting exactly this
rule: all eight node kinds and all five schedule options render in full, and the
popup still cannot exceed the viewport.

Nothing gets narrower under this change — the popup's floor is what its width
used to be — so no existing dropdown can start clipping as a result.

## Verification

- [ ] Node-kind picker shows all eight labels in full, create **and** edit pane
- [ ] Schedule picker shows all five options in full
- [ ] A long teammate name does not push the popup past the viewport
- [ ] Popups elsewhere (tasks, people, skills) are unchanged or wider, never narrower

PR #1025, opened 2026-08-18
fix(console): show the select popup's truncation at the cut edge (#975)

### LINKED ISSUE #975: workflows: the selector truncates with no sign there are more, so workflows below the fold go unseen
## Summary

The workflow selector shows only ~4–5 entries before it needs scrolling, with **no affordance saying more exist**. Found during live QA on `smoke1.staging.opencompany.work` (17 Aug), where the company has 8 workflows: **Feature Pipeline**, **AJ** and **TESTINGG** were below the fold and one of them — Feature Pipeline, the healthiest workflow on the tenant, 13 runs — was missed on first pass.

## Why this is worth fixing rather than shrugging at

The dropdown is the only route to a workflow that is not currently selected. If a user cannot see that the list continues, an unscrolled dropdown is indistinguishable from a complete one. The QA tester recorded two workflows as *"Not inspected due to dropdown truncation"* — the control silently under-reported the company's own contents to someone whose job was to enumerate them.

## Repro

1. Open a company with 8+ workflows.
2. Go to **Workflows** and open the selector.
3. Only the first ~4–5 render; nothing indicates the rest.

Verified **not** reproducing at 6 workflows (all six fit) — so this is count-dependent, and a fix should be tested against a company with enough workflows to overflow.

## Suggested direction

A count next to the control (`Workflows 8`) would help but is not sufficient on its own — it tells you the total without telling you the list is cut. Prefer a visible scroll affordance on the menu itself, so the truncation is apparent at the point of truncation.

## Not to be confused with

- **#824** — the workflows *toolbar* overflowing a clipped ancestor (closed)
- **#812** — the *add-node* dropdown's legibility (closed)

Both are different controls. This one is unfiled.

Source: live QA report, `smoke1.staging.opencompany.work`, 17 Aug 2026 (BUG-07).

M3gA-Mind:
## Summary

The workflow selector showed ~4–5 of a company's 8 workflows with nothing saying more existed. Live QA
recorded two as *"not inspected due to dropdown truncation"* — including Feature Pipeline, the healthiest
workflow on the tenant, 13 runs. The control under-reported the company's own contents to someone whose job
was to enumerate them.

**The issue's diagnosis — and my own first hypothesis — were both wrong, and measuring is what caught it.**
The issue says there is no affordance. There is one, it mounts at the right moment, and it is visible. My
first guess was that the arrows never mount at all; a Chromium measurement killed that too.

## The two measured facts this rests on

Taking these first, because without them the obvious fix is "add a count badge, add a scrollbar" — which
layers a third signal on top of two working ones and leaves the actual cause untouched.

**1. Base UI trades the native scrollbar away *because* the arrow is meant to be the signal.**

`SelectList` gets `base-ui-disable-scrollbar` whenever scroll arrows are mounted. Measured on the live
component:

```
listClasses:      "p-1 base-ui-disable-scrollbar"
scrollbarWidthPx: 0            // offsetWidth - clientWidth
```

So the arrow is not *a* signal, it is the *only* remaining one. Anything wrong with it is total.

**2. The arrow was an opaque strip painted over the half-clipped next row.**

The arrow was a low-contrast chevron on a flat `bg-popover` band pinned to the bottom edge. Underneath it,
the list was already doing the right thing. Hiding the arrow and screenshotting what it covered:

```
listBottom: 217
rowsNearEdge: [
  { n: 6, t: 166, b: 194, clippedAtListBottom: false },
  { n: 7, t: 194, b: 222, clippedAtListBottom: true   },   <-- cut mid-row at the boundary
  ...
]
```

**"Workflow 7" was there all along, clipped mid-row** — a textbook "content continues" cue. The opaque strip
was painting over exactly it, so the list ended on a clean, uncut row. A list that ends tidily reads as a
list that ended.

## The fix

Stop covering it. Both arrows fade to transparent instead of filling:

```
bottom-0 items-end   bg-gradient-to-t from-popover from-30% to-transparent
top-0    items-start bg-gradient-to-b from-popover from-30% to-transparent
```

The clipped row shows through, faded, so the boundary reads as a cut rather than an end. `from-30%` keeps the
outer 30% solid and the chevron is pinned there (`items-end` / `items-start`), so making the strip
see-through costs the icon no contrast.

One change to the shared primitive, so all ~20 selects in the console get it rather than the workflow picker
alone.

**A count beside the trigger was the issue's other suggestion and is deliberately not what this does.** A
total tells you how many exist; it does not tell you the list in front of you is cut. It would also have to
be added to each select one at a time, and it would sit at the trigger rather than at the point of
truncation, which is where the information was missing.

## API Or Behavior Changes

- Both select scroll arrows fade to transparent instead of painting an opaque band. Visible everywhere
  `SelectContent` is used.
- `SELECT_SCROLL_UP_ARROW_CLASSES` / `SELECT_SCROLL_DOWN_ARROW_CLASSES` are exported for the regression test
  (see Tests for why they have to be).
- No change to when an arrow appears — that is Base UI's, and it was already correct.

## Tests

`frontend/test/unit/select-scroll-affordance.test.ts`, 8 cases pinning: each arrow fades to transparent,
neither carries a flat `bg-popover`, each fades away from its own edge (swapping them would fade the popup's
outer edge into the page and leave the row still covered), and the chevron sits on the solid end.

**Teeth proved by reverting: 7 of the 8 fail** against the old opaque strip. The eighth is the
guard-the-guard case (`top-0` / `bottom-0`), which the old strip also satisfied — reporting 7, not 8.

**Why it asserts strings, stated plainly.** jsdom does no layout, so Base UI never measures the list as
scrollable and **the arrows never mount there**; rendering the parts standalone throws for want of
`SelectRootContext`. There is no rendered element to interrogate. The rule is therefore pinned at its source,
following the same precedent as `select-popup-width.test.ts` (#811) in this file, and the behavioural proof
is the browser work below.

**Browser verification — 12 entries, real Chromium, viewport constrained to force the cut.** It does *not*
reproduce at full viewport height (the popup simply fits), which matches the issue's "not at 6 workflows".

| | before | after |
|---|---|---|
| light | 6 rows, clean end, faint chevron | 6 rows **+ faded clipped "Workflow 7"** under the chevron |
| dark | same | same — `from-popover` and `text-muted-foreground` are theme tokens, so both adapt; chevron measured at `oklch(0.682 …)` on the dark popover |

Up-arrow checked symmetrically by scrolling to the end: it appears, and the down arrow correctly disappears.

- [x] `cargo fmt --all -- --check` — N/A: console-only change, no Rust touched.
- [x] `cargo clippy --all-targets -- -D warnings` — N/A: console-only change, no Rust touched.
- [x] `cargo build --all-targets` — N/A: console-only change, no Rust touched.
- [x] `cargo test` — N/A for Rust. The console gates were run instead: `pnpm typecheck` (`tsc -b --noEmit`)
      clean, and `pnpm test` **886 passed / 86 files**. There is no eslint config under `frontend/`, so
      `typecheck` is the lint gate.

## Documentation

The reasoning lives at the code: `SELECT_SCROLL_ARROW_CLASSES` carries why a gradient rather than a chevron
alone, why the scrollbar is not the answer here, and why a count was rejected. The test header records the
two measured facts so the next reader does not have to re-derive them.

Closes tinyhumansai/opencompany#975

### DIFF
--- a/frontend/src/components/ui/select.tsx
+++ b/frontend/src/components/ui/select.tsx
@@ -88,6 +88,57 @@ function SelectTrigger({
 const SELECT_POPUP_CLASSES =
   "relative isolate z-50 max-h-(--available-height) max-w-[min(28rem,var(--available-width))] min-w-[min(max(9rem,var(--anchor-width)),28rem,var(--available-width))] origin-(--transform-origin) overflow-x-hidden overflow-y-auto rounded-lg bg-popover text-popover-foreground shadow-md ring-1 ring-foreground/10"
 
+/**
+ * The shared look of both scroll arrows (issue #975).
+ *
+ * # Why a gradient and not just a chevron
+ *
+ * An arrow was already here and already worked: Base UI mounts it exactly when
+ * the list can scroll further, and a browser measurement confirms it is present
+ * and visible at the moment of truncation. It was still missed in live QA, where
+ * a tester recorded two of a company's eight workflows as *"not inspected due to
+ * dropdown truncation"* — including the healthiest one on the tenant.
+ *
+ * The reason is that it was the **only** signal. Base UI adds
+ * `base-ui-disable-scrollbar` to the list whenever scroll arrows are mounted, so
+ * the native scrollbar is deliberately traded away for the arrow — and the arrow
+ * was a small low-contrast chevron on a flat `bg-popover` strip, below a list
+ * that ended on a clean, uncut row. A list that ends tidily reads as a list that
+ * ended. There was nothing at the cut edge saying otherwise.
+ *
+ * The gradient is what makes the edge itself the affordance: the row under the
+ * arrow fades out instead of stopping square, so the content visibly continues
+ * past the boundary rather than finishing at it. `from-55%` keeps the chevron
+ * sitting on solid `popover` so it loses no contrast, and only the half nearest
+ * the content fades.
+ *
+ * A count beside the trigger (`Workflows 8`) was the issue's other suggestion
+ * and is deliberately **not** what this does. A total tells you how many exist;
+ * it does not tell you that the list in front of you is cut, and it would have
+ * to be added to each of the twenty-odd selects in the console one at a time.
+ * This is one change to the shared primitive and it lands at the point of
+ * truncation, which is where the information was missing.
+ *
+ * `h-7` over the old `py-1`: the gradient needs vertical room to read as a fade
+ * rather than a band, and a fixed height keeps both arrows the same size whether
+ * or not their icon renders.
+ */
+const SELECT_SCROLL_ARROW_CLASSES =
+  "z-10 flex h-7 w-full cursor-default justify-center text-muted-foreground [&_svg:not([class*='size-'])]:size-4"
+
+/**
+ * Exported for `select-scroll-affordance.test.ts`, which cannot reach them any
+ * other way: Base UI mounts a scroll arrow only once it measures the list as
+ * scrollable, and jsdom does no layout — so the arrows never render there, and
+ * rendering the parts standalone throws for want of `SelectRootContext`. The
+ * rule is therefore pinned at its source, and the behavioural proof (a real
+ * Chromium, twelve entries, a constrained viewport) lives in the PR.
+ */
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
@coderabbitai review

Issue #983, opened 2026-08-18
chat: an agent turn that does real work has no response path — the operator gets a 504 while the work runs on invisibly


oxoxDev:
## Summary

`POST /api/v1/company/chat` is contractually defined to answer with the *finished* turn. A turn's duration is unbounded, so a single synchronous shape is broken by construction — no timeout value fixes it. #882 already moved the work onto its own task, which is why the work survives; but nothing is *recorded* when a turn is accepted, so a lost response leaves nothing to read back.

## Evidence (smoke1, 2026-08-18)

Five real persona tasks POSTed together. **All five returned HTTP 504 with an empty body at 120s** (120, 120, 120, 121, 121). The work was not lost — fifteen minutes later four of the five cards were still `in_progress` and one deliverable had been published. A sixth, single request afterwards stayed open past five minutes and never settled.

Trivial turns are fast: operator round trip 6.0s, three desk probes 4.1 / 5.4 / 4.1s.

The concurrency shape matters: the per-company cycle lock is a mutex held for a whole turn with unbounded waiters and no admission control, so five concurrent messages became a serial train and the sixth inherited the train's full latency.

## Why the turn is invisible — three independent causes, all needing fixing

1. The operator's message is journaled **inside** the cycle lock, so `chat/history` shows nothing until the cycle wins the lock.
2. The SSE projection deliberately drops the operator's message as "the operator's own echo".
3. Nothing durable records that a turn is working.

The task card that did appear is an accident of ordering — it is written before the cycle runs. Messages that open no card have no indicator at all.

## Where the 120s comes from — open

Not in either repo. The manager's proxy client sets no timeout, the ingress carries no annotations, and no read-timeout config exists anywhere in the control plane. Corroborating: a manager-side cut maps to a masked 500, so it could not have produced the observed 504. It is the host nginx in front of Traefik on the k3s node, configured outside both repos. **Nothing in the fix depends on the answer.**

Raising it is rejected as a fix and is mildly harmful: it cannot bound an unbounded operation (the sixth request outlived it), it costs held-open sockets, and it makes a lost response *more* similar to lost work — the exact axis that must stay separable. The legitimate move is the inverse: once a response path exists, lower it deliberately so the synchronous mode fails fast.

## Fix — staged

[...]

Issue #981, opened 2026-08-18
workflows: the destination picker offers a target delivery refuses by name, and a dropped report still reads as a successful run


oxoxDev:
## Summary

Two defects on the same path, found together in the 2026-08-18 QA pass.

**1. Three surfaces disagree about which channels may receive a report.** `CompanyRuntime::wired_channel_ids()` returns every adapter and documents itself as "exactly what an `output` node's `channel` destination may target… `operator` is always present". `RuntimeBuilder` builds the delivery deps from that same vec **minus** `operator`, and `post_to_channel` refuses `operator` by name. So the picker endpoint and the console's destination editor offer authors the one target guaranteed to fail.

**2. A run's verdict exists only in the console's TypeScript.** No server-side surface carries one, so `summarize_run` never reads `run.deliveries` and closes a failed-delivery run with "The run reached its terminal node(s) without pausing for approval". Any API client folding `nodes[].status` scores a dropped report as green — including the QA harness that found this.

## Evidence (smoke1, 2026-08-18)

`daily-release-readiness` finished with every node `ok`:

```json
{"kind":"channel","node":"report","target":"operator","status":"failed",
 "reason":"channel-not-wired",
 "detail":"`operator` is not a workflow delivery channel — this runtime has: engineering, product_design, go_to_market"}
```

An older run of the **same** workflow failed the other way round — `` `engineering` is not a wired channel — this runtime has: operator `` — because desk adapters postdate the graph. Across 20 runs in history, 7 attempted a delivery and 3 failed.

## Fix

Split into two PRs.

**PR-1** — name the exclusion rule once, next to the operator-channel constant, with its reason attached; have the builder and the picker both read it; reject a channel destination outside that set when a workflow is **saved**, while the author can still act on it. Deliberately not validated at TOML parse time — that would refuse to boot a template on a runtime whose desks are not yet resolved.

**PR-2** — a server-side `WorkflowRunVerdict` (`running | failed | stopped | blocked | undelivered | awaiting-approval | ok`) in the console's existing precedence order, always serialized on both run DTOs; the console reads it instead of owning the definition. **Derived, never stored**, so history re-scores on deploy with no migration.

## Constraints the fix must honour

- A failed delivery must **not** populate the run's `error` or flip `nodes[].status`. That would mislead the copilot's fix-from-run into repairing a graph whose nodes were fine, and collapse three terminal readings #383 kept distinct. Existing consumers of node status see exactly what they see today.
[...]

Issue #979, opened 2026-08-18
console: index.html is served cacheable, so every returning user white-screens after a deploy


oxoxDev:
## Summary

The host serves the built console with `ServeDir` and sets no `Cache-Control` on anything. `index.html` therefore gets heuristic caching in the browser. After a deploy, a returning user's browser keeps **yesterday's `index.html`**, which names yesterday's entry bundle. That entry bundle's lazy chunks are gone from the new image, the SPA fallback answers them with `index.html`, the dynamic import throws, React unmounts, and the page is blank with nothing shown to the user.

## Reproduced on smoke1 (2026-08-18, after rolling to `fa51ee99`)

The tab had a cached `index.html` from the previous image. Every route rendered an empty `#root`. Console:

```
TypeError: Failed to fetch dynamically imported module:
  https://smoke1.staging.opencompany.work/assets/KnowledgeGraph-BnTAWS1T.js
```

Fetching the chunks the stale entry asks for:

```
/assets/KnowledgeGraph-BnTAWS1T.js  -> 200 text/html   (SPA fallback, not JS)
/assets/FinancesView-CEYbSrau.js    -> 200 text/html
/assets/UsageView-Ceh7s4Gs.js       -> 200 text/html
/assets/chart-ArNM2hl6.js           -> 200 text/html
```

A freshly-fetched `/` names a different entry, which is the one actually on disk:

```
$ curl -s '/?cb=...' | grep -o 'assets/[^"]*'
assets/index-Ch7qxK26.js
assets/index-BnjpXGv1.css
```

Response headers on `/` carry `etag` and `last-modified` and **no `cache-control`**.

A cache-busting reload fixed it instantly and the console worked normally, which is what confirms the diagnosis: nothing was wrong with the build, only with which `index.html` the browser kept.

## Why it matters

This fires for every user who has visited before, on every deploy, and presents as a blank page with no error — indistinguishable from the app being down. It is the highest-impact defect found in the 2026-08-18 QA pass.

## Fix

Set cache headers on the console fallback in `src/server/routes.rs`:

- HTML responses (the shell, and every SPA-route fallback) -> `Cache-Control: no-cache`, so the browser revalidates and always learns the current entry name.
- Hashed assets under `/assets/` -> `Cache-Control: public, max-age=31536000, immutable`, which is safe precisely because the name changes when the bytes do.

## Verification

[...]

PR #1026, opened 2026-08-18
qa: land the console parity harness in the repo (#987)

### LINKED ISSUE #987: QA: land the console parity harness in the repo so a release can be checked in one pass
## Summary

The 2026-08-18 QA parity pass was driven by a throwaway console script and a checklist that currently live outside the repository. Both proved their worth — two of the checks exist *because* they caught bugs nothing else did, and neither is visible from the UI — so they should be versioned alongside the thing they test.

## What to land

**`oc-qa.js`** — a zero-dependency script pasted into the browser console on a signed-in tenant. It rides the session cookie, so there is no token to distribute. Two entry points:

- `OCQA.read()` — 22 read-only checks in ~10s: host, **console cache headers**, company lifecycle, roster, desks, task board, workspace, approvals backlog and its age, approval tier, manifest connections, Composio account health, MCP servers and their auth, inference provider, repo binding, workflows, run history, **workflow deliveries**, data hygiene, skills, usage/finances, chat history, tool catalog. Each row reports PASS/WARN/FAIL **next to the value it judged**, so a verdict can be checked rather than trusted.
- `OCQA.probe()` — five live checks that need the company to actually work: chat round trip with latency, one probe per desk, card create → board → delete, a real workflow run to a terminal state, and whether the probes parked anything new at the approvals gate.

**`MASTER-QA.md`** — the seven UI checks a script cannot judge, the five persona tasks with a scoring rubric (answered / grounded / honest / delivered), and a single prompt that hands the whole pass to an agent with browser access.

## The two checks worth keeping regardless of the rest

- **Console cache headers** — caught #979. A cacheable `index.html` white-screens every returning user after a deploy, and there is nothing in the UI to see, because the UI is the thing that failed to load.
- **Workflow deliveries** — caught #981. A run whose nodes are all `ok` can still have delivered nothing.

## Also worth fixing in the harness itself

The probe scored the delivery-failure run as PASS, because at the time it judged a run by node status alone — the same mistake the product makes in #981. That check was added afterwards; the lesson is that the harness needs to be reviewed against each bug it misses, not just extended.

## Known gaps in the pass, to close when they become testable

Three checks were **untested, not passed**: board drag, approvals-resolve (the queue was empty, so there was nothing to resolve without changing the tenant's approval tier), and persona quality — only partly scored, because four of five turns were still running when the pass ended. `#983` makes that last one properly measurable for the first time.

## Suggested placement

Somewhere a release checklist can point at without hunting — a `qa/` directory or under `scripts/`. Docs should state plainly that the tenant must be rolled to the commit under test first: a tenant on an old image reports bugs main has already fixed, which is what made the first attempt at this pass misleading.

CodeGhost21:
Closes #987.

## Summary

Lands the QA parity harness in the repo, so a release can be checked in one pass without hunting for the script first.

- **`qa/oc-qa.js`** — zero-dependency console script. `OCQA.read()` (22 read-only checks, ~10s), `OCQA.probe()` (5 live checks), `OCQA.report()` (Markdown for an issue). Rides the session cookie, so there is no token to distribute.
- **`qa/MASTER-QA.md`** — the seven UI checks a script cannot judge, the five persona tasks with an answered/grounded/honest/delivered rubric, and the single prompt that hands the pass to an agent.
- **`qa/README.md`** — how to run it, and the roll-the-tenant-first prerequisite.
- Pointers from `README.md` and a new *Checking a Release* section in `CONTRIBUTING.md`.

## Three rules enforced in code, not just described

1. **A verdict prints beside the value it judged**, so a PASS can be checked rather than trusted.
2. **Unreadable is never PASS.** A 404, an error, a feature absent from the build, or a missing precondition is `SKIP` — counted apart from PASS in the summary. A host answering `{"code":"not_wired"}` is included: a build with no workflow runner cannot run a workflow, and scoring that red sends somebody chasing a graph that is fine. Matched on the typed `code`, never the prose (#248).
3. **A check that missed a bug gets changed, not just supplemented.** `runVerdict` reads `deliveries`, `pendingApprovals` and `blockedNodes` in the console's own precedence order (`running → failed → stopped → blocked → undelivered → awaiting-approval → ok`) rather than folding `nodes[].status` — which is the mistake the harness itself made on the very run that found #981.

Those seven words are exactly the `WorkflowRunVerdict` #981 proposes server-side. When it lands, this transcription should be deleted and the host's verdict read instead; the file says so.

## The two checks worth keeping regardless of the rest

- **`console-cache-headers`** — caught #979. Now a regression guard: the fix is in `src/server/routes.rs`, so this reads PASS on a current build and a FAIL means a deploy reintroduced it.
- **`workflow-deliveries`** — caught #981. A run whose nodes are all `ok` can still have delivered nothing.

## Tests

`frontend/test/unit/qa-harness.test.ts` (31 tests) evaluates the script in a `vm` sandbox — it has no `export`, because an `export` would make it unpasteable, which is the one property it must not lose. It:

- pins `runVerdict` against `run-health.ts` case by case, **asserting both halves** (the console still reads it this way, AND the harness agrees) so the two cannot drift together silently;
- covers the `blocked` (#881) and gate-folding (#846) arms, whose absence scores a run that delivered nothing as green;
- pins the `#979` cache reading, including that an *absent* header is FAIL rather than WARN;
- drives the real `read()` over a 404ing host, asserting exactly 22 checks and that nothing unreadable reports PASS.

## Verification

- `npm run typecheck`, `npm run typecheck:unit`, `npm test` (922 tests, 88 files) — all green.
- Driven against a **live host** (`test/e2e/host.sh`, e2e_harness company) in both modes, signing in the way the e2e suite does. That run found three defects in the harness itself, all now fixed and pinned: a `not_wired` build scored FAIL, every duration printed as `0.0s`, and a doubled unit suffix.

## Known gaps, reported as SKIP rather than passed

- **Approval tier.** `[policy].mode` has no read surface on REST or GraphQL, so no API client can verify the setting that decides whether effects park at all. Filed separately.
- **Board drag** and **approvals-resolve** are browser-only; the latter needs a queued approval, which the 2026-08-18 pass could not create without changing the tenant's tier. `MASTER-QA.md` gives two ways to get one.
- **Persona quality** — #983 makes it properly measurable for the first time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,3 +30,14 @@ cargo check --features tiny
 
 Keep changes focused. Include a short summary, any API or behavior changes, and
 the local verification commands you ran.
+
+## Checking a Release
+
+Local checks answer "does the code work". A release also needs "does the thing
+we deployed work for the operator", which no in-repo test can reach: a stale
+`index.html`, an unwired delivery channel and a missing credential are all
+failures of the deployment rather than of the code.
+
+That pass lives in [`qa/`](qa/README.md) — a console script (`qa/oc-qa.js`) and
+a checklist (`qa/MASTER-QA.md`). Roll the tenant to the commit under test
+first; a tenant on an older image reports bugs `main` has already fixed.
--- a/README.md
+++ b/README.md
@@ -302,6 +302,9 @@ and the [tiny.place economy](gitbooks/overview/tiny-place.md). Builders should
 start with the [developer section](gitbooks/developers/README.md) — build,
 CLI, architecture, authoring companies, deployment, and configuration.
 
+Checking a release against a deployed tenant is [`qa/`](qa/README.md): a
+zero-dependency console script and the checklist that goes with it.
+
 ## Community
 
 [Discussions](https://github.com/tinyhumansai/opencompany/discussions) is where
--- a/frontend/test/unit/qa-harness.test.ts
+++ b/frontend/test/unit/qa-harness.test.ts
@@ -1,0 +1,405 @@
+/**
+ * The QA harness's judgements, pinned (issue #987).
+ *
+ * `qa/oc-qa.js` is pasted into a browser console, so nothing imports it and
+ * nothing type-checks it. Two things about it are worth a gate anyway:
+ *
+ * 1. **It parses.** A syntax error is discovered by an operator mid-incident,
+ *    which is the worst possible moment and the only moment it is ever run.
+ *
+ * 2. **Its run verdict agrees with the console's.** The harness owns a
+ *    transcription of `run-health.ts`, and a second definition of "did this run
+ *    succeed" is precisely the defect issue #981 filed against the product —
+ *    where the console's TypeScript held the only verdict and every API client
+ *    folding `nodes[].status` scored a dropped report as green. The harness made
+ *    exactly that mistake and scored a delivery-failure run as PASS. Pinning the
+ *    two together is what stops a change to the console's reading from silently
+ *    re-greening a bad run in the harness.
+ *
+ * The script is evaluated in a `vm` sandbox rather than imported: it is an IIFE
+ * that assigns `globalThis.OCQA`, with no `export`, because an `export` would
+ * make it unpasteable — which is the one property it must not lose.
+ */
+
+import { readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { createContext, runInNewContext } from "node:vm";
+import { describe, expect, it } from "vitest";
+
+import { runTone } from "@/views/workflows/run-health";
+import type {
+  DeliveryReport,
+  WorkflowBlockedNode,
+  WorkflowRunOutcome,
+} from "@/api/workflows";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const SCRIPT = resolve(here, "../../../qa/oc-qa.js");
+
+/** One reported check, as the script emits it. */
+interface Row {
+  check: string;
+  verdict: "PASS" | "WARN" | "FAIL" | "SKIP";
+  value: string;
+  note: string;
+}
+
+/** A minimal `Response` stand-in — the four things `http()` reads. */
+function response(status: number, body: unknown, headers: Record<string, string> = {}) {
+  return {
+    ok: status >= 200 && status < 300,
+    status,
+    text: async () => (body === undefined ? "" : JSON.stringify(body)),
+    headers: new Headers(headers),
+  };
+}
+
+/** Loads `oc-qa.js` into a sandbox and hands back its `_internals`. */
+function loadHarness(fetchImpl?: (path: string) => Promise<unknown>) {
+  const source = readFileSync(SCRIPT, "utf8");
+  const sandbox: Record<string, unknown> = {
+    console: { log: () => {}, table: () => {} },
+    fetch:
+      fetchImpl ??
+      (async () => {
+        throw new Error("no network in this test");
+      }),
+    setTimeout,
+    clearTimeout,
+    AbortController,
+    Headers: globalThis.Headers,
+    location: { host: "test" },
+  };
+  createContext(sandbox);
+  runInNewContext(source, sandbox);
+  const ocqa = sandbox.OCQA as {
+    version: string;
+    read: (options?: { company?: string }) => Promise<Row[]>;
+    probe: unknown;
+    report: unknown;
+    _internals: {
+      runVerdict: (run: unknown) => string;
+      undeliveredCount: (d: DeliveryReport[]) => number;
+      pendingCount: (d: DeliveryReport[]) => number;
+      awaitingCount: (run: unknown) => number;
+      isBlocked: (run: unknown) => boolean;
+      judgeCacheHeader: (kind: string, header: string | null) => { verdict: string; note: string };
+      age: (atMillis: number, now?: number) => string;
+      notWired: (res: { body: unknown }) => boolean;
+      secs: (ms: number) => string;
+    };
+  };
+  return ocqa;
+}
+
+/** A run outcome with only the fields the verdict reads. */
+function run(overrides: Partial<WorkflowRunOutcome>): WorkflowRunOutcome {
+  return {
+    seq: 1,
+    atMillis: 1_700_000_000_000,
+    workflowId: "daily-release-readiness",
+    scheduled: false,
+    deliveries: [],
+    pendingApprovals: [],
+    ...overrides,
+  } as WorkflowRunOutcome;
+}
+
+function delivery(status: DeliveryReport["status"]): DeliveryReport {
+  return { node: "report", kind: "channel", target: "operator", status, detail: "" };
+}
+
+/** A node the run stopped short at, waiting on a person (issue #881). */
+function blocked(nodeId: string): WorkflowBlockedNode {
+  return { nodeId, tools: ["send_email"] };
+}
+
+describe("oc-qa.js loads", () => {
+  it("parses and exposes both entry points", () => {
+    const ocqa = loadHarness();
+    expect(typeof ocqa.read).toBe("function");
+    expect(typeof ocqa.probe).toBe("function");
+    expect(typeof ocqa.report).toBe("function");
+    expect(ocqa.version).toMatch(/^\d+\.\d+\.\d+$/);
+  });
+});
+
+describe("read() against a host whose surfaces do not answer", () => {
+  /**
+   * Rule 2 of the harness, end to end: **unreadable is never PASS.**
+   *
+   * The 2026-08-18 pass wrote three checks up as green that had never run, and
+   * a pure-function test cannot catch that — the mistake lives in the plumbing,
+   * where a 404 body decays to `{}` and every field read off it comes back
+   * `undefined`. So this drives the real `read()` over a host that answers
+   * `/healthz`, `/spec` and the company status and 404s everything else, and
+   * asserts nothing downstream claims to have passed.
+   */
+  const reachable: Record<string, unknown> = {
+    "/healthz": { status: "ok" },
+    "/spec": {
+      name: "opencompany",
+      version: "0.1.0",
+      capabilities: ["rest", "graphql"],
+      storage: "memory",
+      instance_id: "abcdef0123456789",
+    },
+    "/api/v1/company": { id: "acme", name: "Acme", lifecycle: "running", pending_approvals: 0 },
+  };
+
+  async function runRead(): Promise<Row[]> {
+    const ocqa = loadHarness(async (path: string) =>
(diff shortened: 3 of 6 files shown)


CodeGhost21:
**The red `Console E2E` check is pre-existing on `main`, not from this PR.**

One spec fails, and it is the same one on both:

| | Failing spec | Result |
| --- | --- | --- |
| This PR (`278299e0`) | `test/e2e/toast-dismissal.spec.ts:83 › hovering a toast still holds it open to be read` | 1 failed, 152 passed |
| `main` at this PR's merge base (`bbf55770`, [run](https://github.com/tinyhumansai/opencompany/actions/runs/32133630290)) | same spec, same line | 1 failed |

It is unrelated to this change by construction: this PR adds no e2e spec and touches nothing under `frontend/src`. Its whole diff is `qa/` (a console script and two docs, none of it shipped in the bundle), one vitest unit test, and two README pointers. The failing spec exercises toast lifetime, which is [#933](https://github.com/tinyhumansai/opencompany/issues/933)'s territory.

For the parts this PR does add, CI is green — `Console` (typecheck + `vitest run`, 922 tests) and both Rust lanes all pass.

oxoxDev (dismissed):
Genuinely good QA tooling, and the unit test is the best part — I mutation-tested it and it killed 5/5 mutants, including `unread() → PASS`. No secrets anywhere, and `runVerdict` is a faithful transcription of `run-health.ts:137-166` cross-pinned against the console's real `runTone` import, so the two can't drift together. Two things block, both small.

**Major — `qa/oc-qa.js:837-873`: `read()` has no per-check exception boundary, so one bad body loses all 22 rows.**

The transport docstring at L69-72 justifies the non-throwing `http()` precisely because "a check that throws takes the other twenty-one with it" — but that defence stops at the transport layer. No check is wrapped and `read()` has no `try`/`catch`. The concrete instance is L782: `f.spentUsd.toFixed(2)` / `f.budgetUsd` / `f.balanceUsd` are dereferenced unguarded whenever `finances.ok` is true, while L777 already treats those same fields as possibly-absent. Driving the real `read()` in a `vm` sandbox against a host answering `/finances` `200 {}`:

```
read() THREW -> TypeError: Cannot read properties of undefined (reading 'toFixed')
   => ALL rows lost, no summary printed.
```

`/finances` is documented as "Phase 1" in `src/server/ops/finances.rs:1`, so shape drift on an older tenant is exactly the deployment-era condition this tool exists to catch — and it's the case where you get nothing at all. About five lines: wrap each `await check…(…)` in `read()`/`probe()` and emit a SKIP row on throw. L782 is the only sibling I found; every other `.join`/`.toFixed` is guarded by `|| []` / `|| 0` / `&&`.

**Major — `qa/oc-qa.js:973,978-979`: bare `OCQA.probe()` executes an arbitrary real workflow, and no doc says so.**

`opts.workflow ? … : flows[0]` picks whatever happens to be first, and `if (opts.dryRun) … else body.detach = true` makes the real run the default. Together with `probeChat` and `probeDesks` that's up to 6 real agent turns plus one real workflow execution with real deliveries (channel/email), against `location.host` — no host allowlist, no `confirm()`, no echo of the target. `qa/README.md` documents only the `{workflow, dryRun:true}` form, and `qa/MASTER-QA.md:23` likewise; neither states what the bare call does. The sharpest instance is `MASTER-QA.md` step 3, which instructs an autonomous agent with browser access to run `OCQA.probe()` un-parameterized against a `<TENANT URL>` supplied at runtime.

Real-by-default is defensible — #987 does ask for "a real workflow run to a terminal state". The arbitrary target and the silent docs aren't. Smallest fix: require `workflow:` to be named for the non-dry path (SKIP otherwise), or default `dryRun: true` unless `{dryRun:false}` is passed, plus one sentence in README/MASTER-QA naming what the bare call fires.

**Worth fixing before merge, not blocking**

- Commit `278299e` carries `Co-Authored-By: Claude Opus 5 (1M context)`. That's against the convention on tinyhumans repos — needs an amend and force-push. (It's also unsigned, but `a56ed8c3` on main from the same contributor is too, so that's precedent rather than a new problem.)
- `OCQA.report()` (L902, L1120-1131) emits tenant content into Markdown "for pasting straight into an issue" — `probe-chat`'s `value` carries 120 chars of a real agent reply, and other rows carry company name/id, roster ids, desk ids, workflow ids, delivery targets, `instance_id[0:8]` and spend figures. `MASTER-QA.md` then tells you to paste that into a GitHub issue. The probe prompt is benign by construction, but this collides with our rules on redacting ids and not putting real message text in public artifacts. Dropping the reply excerpt (or gating it behind an opt-in) would settle it.
- Nit: L340 uses bare `location.host` while `finish()` at L1093 guards it as `global.location ? … : "unknown"`. Harmless in a browser, just inconsistent.

**Answering the things I'd want to know before merging this**

*Secrets:* none. Zero hardcoded hosts, tokens, keys, user ids or `/Users/…` paths — the design is credential-free by riding the session cookie of the tab it's pasted into, so there's nothing to mint or distribute. Nice property.

*Prod safety:* it can be pointed at production and nothing guards it, but severity is bounded — nothing deletes user data. `probeBoardCard` creates and then deletes only its own card (L943-950). The exposure is uncontained side effects rather than destruction, which is what Major 2 is about. Mitigations that do exist are decent: `dryRun` opt-in, a WARN when a host ignores the flag (L995-996), and `probe-approval-delta` (L1044-1059) reporting what the probes parked. `read()` is genuinely read-only — verified, GETs only.

*False green:* no, and this is the harness's strongest property. `unread()` (L286-290) returns SKIP unconditionally, `finish()` (L1088-1110) tallies SKIP separately and prints "SKIP means untested, not passed", and the unit test drives the real `read()` over a 404ing host asserting only `host` and `company-lifecycle` pass. The one caveat is Major 1's inverse — it can report *nothing*, which is loud rather than falsely green, but total.

*CI wiring:* better than it looks. The live probes are manual-only by design (correctly — they need a signed-in tenant), but the judgement layer is genuinely exercised: `qa-harness.test.ts` loads the real file in a `vm`, drives the real `read()`, and cross-pins `runVerdict` against the console's actual import, and that runs in the `Console` job. So the verdict logic won't rot; the HTTP/probe layer is untested by CI and will.

*Overlap:* minimal, and the PR argues the distinction correctly. `frontend/test/e2e/` drives a locally-built host via `test/e2e/host.sh`, so it structurally cannot catch a deployment failure — no e2e spec touches cache headers, and the #979 fix lives in `src/server/routes.rs:68` where only a real deploy exercises it. Not a parallel harness.

One thing to be aware of on the check row: **this PR has had no substantive bot review.** tinysweeper approved with zero inline comments on an embedding-only pass, and CodeRabbit never actually reviewed — it hit `Review limit reached` at 12:31:58Z, and its check reports `pass` because rate-limiting is treated as non-blocking. The green looks like two approvals and isn't.

Gates I ran on a temp checkout: `npm ci`, all three typecheck projects (`tsc -b --noEmit`, `tsconfig.unit.json`, `tsconfig.e2e.json`) all exit 0, and `npx vitest run test/unit/qa-harness.test.ts` → 31/31. All 16 PR checks are currently green; the `Console E2E` toast-dismissal flake you documented at 13:05 has since gone green on its own, so your pre-existing-on-main read was right.

M3gA-Mind:
## `approval-tier` can be a real PASS — the read surface it says is missing does exist

Flagging this from #1027 rather than pushing a branch, since `qa/oc-qa.js` lives only on this PR and the edit is yours to make.

`qa/oc-qa.js:527-534` hardcodes the row to `SKIP`:

```js
rows.push(
  row(
    "approval-tier",
    SKIP,
    observed.join(" · ") || "no observable gate activity",
    "[policy].mode has no read surface, so the tier this tenant runs is unverified by any API client",
  ),
);
```

The surrounding function reads `/capabilities` and `/approvals` only. **`{scope}/policy` is never probed, and that is the endpoint the note says does not exist.**

It has been there since #562 (`src/server/ops/policy.rs`, mounted in `src/server/ops/mod.rs`). `read_policy` takes only `ScopedCompany`, so **no admin token is needed for the read** — the harness's existing credentials will do. Verified live on `marketing.staging`: `{scope}/policy` answers **401** unauthenticated while a bogus path under the same scope answers **404**, so the route is mounted and served.

`PolicyDto` gives the harness more than the SKIP note wanted:

- `mode` — the tier **actually in force**, from `record.effective_policy()`, which is the same resolver the approval gate itself reads
- `manifestMode` — the declared tier
- `overridden` / `setBy` / `setAtMillis` — whether a console override is producing that tier, and who set it

That last group is worth having in a QA row: "supervised (overridden by alice@…, manifest says full)" is a materially different tenant state from "supervised" and the harness can now say which.

It also upgrades the row this issue cares most about. An empty approvals queue currently reads the same whether the company is `supervised` with nothing pending or `full` where nothing will ever park — with the tier in hand the harness can tell those apart instead of reporting `no observable gate activity`.

I have **not** touched the file and have opened no branch against it. Raised because that comment is the origin of #1027, which asserted an API gap on its authority; I have posted the correction there and proposed closing it.

senamakel:
Addressing the review feedback.

**oxoxDev changes-requested (both Majors + non-blocking items) — resolved in `32a79cfe`, already pushed:**

- *Major 1 — per-check exception boundary.* Every check in `read()`/`probe()` now runs through `attempt`, so a body whose shape drifted costs only its own SKIP row instead of taking the other twenty-one with it. The cited `/finances` case is fixed at the source too: `spentUsd`/`budgetUsd`/`balanceUsd` are only dereferenced once confirmed numeric, otherwise the row says which fields the host omitted rather than printing absent figures as `$0.00`.
- *Major 2 — bare `probe()` choosing a workflow.* `probe()` no longer picks `flows[0]`. An un-named target is `SKIP` saying how to name one (`{workflow: "<id>"}` for real, or with `dryRun: true` to rehearse), and `probe()` prints the host and company it will act on before it acts. `qa/README.md` and `qa/MASTER-QA.md` both now state this.
- *Report redaction.* `probe-chat` judges the reply by shape in its value and holds the text in `detail`, which `report()` withholds — `report({ raw: true })` opts back in; the table says when text was held back.
- *`location.host` nit.* `hostname()` now guards with `global.location`.
- *Co-Authored-By: Claude trailer.* Gone — the flagged commit `278299e` was amended away in the earlier force-push; no commit on the branch carries it now.

**M3gA-Mind approval-tier — addressed in `5fada4db` (this push).** See the reply above: `approval-tier` now reads `{scope}/policy` and is a real PASS that distinguishes a `supervised` tenant with an empty queue from a `full` one, and reports overrides. The "no read surface" claims in code and `qa/README.md` are removed, and the behaviour is pinned by new unit tests.

Validation: `frontend/test/unit/qa-harness.test.ts` → 47/47 pass; `npm run typecheck:unit` green. CI: `Console` (typecheck + vitest) and both Rust lanes were green before this push; the `Console E2E` toast-dismissal failure is pre-existing on `main` (see the comment above) and unrelated to this PR's `qa/`-only diff.

PR #1028, opened 2026-08-18
fix(workflows): refuse to arm a schedule the graph cannot keep

### LINKED ISSUE #976: workflows: a graph with only a Start node can be saved and scheduled, so a schedule can promise work that cannot run
## Summary

A workflow containing **only a Start node** can be saved *and carry a schedule*. On `smoke1.staging.opencompany.work` two such workflows exist:

- **`campaign`** — tagged "test", **only a Start node**, and it holds a **schedule (currently paused)**. If someone resumes that schedule it will fire on time, run nothing, and report nothing.
- **`QA Test Pipeline`** — only a Start node, **6 recorded runs**.

That second one is the part that bothers me: a stage-less graph has already been *run six times*. Six runs that could not have done anything, recorded as runs.

## Why this is a defect and not just untidy data

A schedule is a promise that something happens. A stage-less workflow accepts that promise and cannot keep it. The failure is invisible in both directions — nothing errors at authoring time, and nothing errors at run time, because there is no stage to fail. It reads as "scheduled and healthy" from every surface.

This is the same *shape* as **#661** (*"authoring & validation gaps let broken graphs (incl. shipped seeds) save clean and fail silently"*, closed) — a graph that saves clean and fails silently. Either that fix does not cover the degenerate single-node case, or this is a regression. Worth establishing which before writing code.

## What needs deciding

1. Should a graph with no runnable node be **refusable at save**, or allowed to save but **refused a schedule**? Saving a stub mid-authoring is legitimate; *scheduling* one is not.
2. What should a run of such a graph record? Today it apparently records a run. `Skipped` with a reason would at least be honest — compare the `NoDestinationConfigured` treatment in **#925**.

I have not found a node-count guard in `src/workflows/`, but my search was not exhaustive — confirm before assuming there is none.

## Related

- **#661** authoring/validation gaps (closed) — check coverage of the single-node case
- **#925 / #947** the same "runs but delivers nothing, and says nothing" family

Source: live QA report, `smoke1.staging.opencompany.work`, 17 Aug 2026 (BUG-08, which filed this only as data hygiene — the schedulability is the real defect).

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#976.

A graph whose only node is its trigger has nothing to execute. The engine runs it happily —
no stage fails, because there is no stage — so it saves clean, accepts a schedule, and
settles as an ordinary finished run. On staging that left `campaign` one resume away from a
schedule it cannot keep, and `QA Test Pipeline` with **six recorded runs** that could not
have done anything.

### First: #661 is a coverage gap, not a regression

Worth settling before anything else, because the next person will otherwise assume #661
covered this and go looking for a regression.

The two commits that touched the draft validator under #661 are `1c5863cf` *"enforce
per-kind required config at author time"* and `c3449ab9` *"enforce the #661 rules strictly
on the draft path"*. What they added: a `condition` needs a `field`; branch labels must be
`yes`/`no`; `tool_call` needs a slug; `agent` must name a real roster teammate; collect
every missing-config problem per node.

**Every one of those rules presupposes a node of that kind exists.** A graph of one
`trigger` node has zero conditions, zero tool_calls, zero agents — so all of #661 is
satisfied **vacuously**. It validated the *quality of the nodes present*; this is the
*absence of any node to run*. Nothing regressed; the case was never in scope.

Confirming the issue's other suspicion: there is no node-count guard.
`validate_draft_shape` bounds nodes and edges from **above** (`MAX_WORKFLOW_NODES`,
`MAX_WORKFLOW_EDGES`) and requires **exactly one** trigger — so a lone trigger *passes*.
The only `len() < 2` in the module is an unrelated cycle-detection component check.

### The guard is at arming, not at save

A reviewer will ask why not refuse at save, so: four reasons, and the first is the
important one.

1. **A stub must stay saveable mid-authoring, and the module is already built that way.**
   `parse_workflow` was made lenient **on purpose** — `c3449ab9`'s own commit message moved
   author-time strictness onto the draft path precisely so the console can drop a Start node
   and add stages later. Refusing at save fights that design.
2. Refusing at save would reject **every existing seed and legacy body on its next edit**,
   including what is on staging now.
3. **Saving promises nothing; arming a schedule promises that something happens.** That
   sentence is the issue's whole thesis, and it lands on arming.
4. #276 already routes scheduled graphs through that human gate **disarmed**, and
   `set_company_workflow_enabled` **already loads the parsed graph** for the journal display
   name — so the check costs no extra I/O.

**Pausing is never refused.** Only `enabled = true` is guarded, and only when a trigger
schedule exists. An operator must always be able to stop a thing — the same call the
existing "a body that no longer parses still toggles" rule makes. A guard that trapped a
workflow *armed* would be worse than the silence it fixes. A manual (unscheduled) stub is
left alone too: running one by hand is the author's own business, and the run says so
itself.

### What a run records: a notice, not an error

No new run status. `WorkflowRunFinished.notices` already exists for exactly this shape
(#638), and its own doc reads like it was written for this issue:

> A run has no conversation to speak on… **Separate from `error` on purpose: a run that
> overflowed the cap succeeded, and putting this there would mark it failed and inflate the
> failure count.**

A stage-less run is the same class. **Nothing broke and nothing was attempted**, so marking
it failed would put a half-authored stub into the failure count beside runs that genuinely
went wrong — the same call #925 makes one level down with `NoDestinationConfigured`, which
exists so "routed nothing on purpose" and "never configured a destination" stop being the
same observation. Here the pair is "this run had nothing to do" versus "this run did
nothing". That is what makes `QA Test Pipeline`'s six runs read honestly without
retroactively marking them failed.

### One predicate

`WorkflowFile::has_runnable_node()` sits beside `trigger_schedule()` and inherits its
reasoning: two copies of a predicate that disagreed would let a graph be refused a schedule
and still run silently, or the reverse.

It asks "is there a non-trigger node", **not** a node count. The count is the tempting
shortcut and is wrong twice: a graph of three triggers has three nodes and still does
nothing, while a legitimate one-stage graph (trigger → agent) has only two. What matters is
whether any node kind *executes*, and every kind except `Trigger` does.

## API Or Behavior Changes

New: `WorkflowFile::has_runnable_node()`, and two operator-facing string constants.

Behaviour: `PUT …/workflows/{wid}/enabled` with `{"enabled": true}` now returns 400 for a
scheduled graph with no runnable node. Pausing, saving, editing, and manual runs are
unchanged. A stage-less run gains one `notices` row; every other run is byte-identical.

**Existing staging data is untouched by this PR.** `campaign` stays paused and this guard
would refuse resuming it, which is the intended outcome; cleaning up those rows and the six
recorded runs is a separate human decision.

## Tests

      exit 0. `--no-deps` mirrors `ci.yml`; a bare `--all-targets` drowns in vendored lints CI
      never compiles.
- [x] `cargo build --all-targets` — covered by the two test lanes below, which build all
      targets for both feature sets.
      `RUST_MIN_STACK=16777216 cargo test --locked --features openhuman,tinycortex --tests`
      → **4230 passed / 0 failed**. `RUST_MIN_STACK` matches `ci.yml:626`; without it a
      pre-existing `harness::brain` test overflows the stack locally.

`scripts/ci/assert-feature-lanes.sh` passes; no new row is owed.

### Revert proof — two, each isolating a half

| Reverted | Result |
| --- | --- |
| the arming refusal block | `arming_a_stageless_schedule_is_refused` **FAILED** |
| the run-notice condition → `if false` | `a_run_of_a_stageless_graph_says_it_had_nothing_to_run` **FAILED** |

**Being straight about the other five.** `a_stageless_graph_still_saves`,
`a_stageless_schedule_can_still_be_switched_off`,
`arming_a_scheduled_graph_with_a_stage_still_works`,
`a_graph_with_a_stage_raises_no_stageless_notice` and
`a_stageless_run_is_not_recorded_as_a_failure` survive both reverts. They are guard tests —
they exist so the fix cannot *over*-fire (refusing every schedule, noticing on every run,
trapping a workflow armed) — not discriminating ones, and I am not counting them as
coverage for the fix itself.

## Documentation

`docs/modules/server/pausing-workflows.md` — the doc that already owns the arming gate and
the #276 disarm rule — gains the refusal, the save-versus-arm reasoning, why pausing is
never refused, and the run-notice decision.

### DIFF
--- a/docs/modules/server/pausing-workflows.md
+++ b/docs/modules/server/pausing-workflows.md
@@ -34,6 +34,37 @@ all", and putting a re-enable click behind every typo fix is how an operator
 learns to click through it. `a_switched_off_workflow_does_not_fire` and
 `an_edit_that_adds_a_schedule_switches_the_workflow_off` pin the two halves.
 
+**A schedule that cannot run is refused arming (issue #976).** `set_company_workflow_enabled`
+rejects `enabled = true` when the graph carries a trigger schedule and has **no
+node other than its trigger**. Such a graph fires on time, runs nothing and
+reports nothing: on staging `campaign` was one resume away from exactly that.
+
+**Refused at arming, not at save**, and the distinction is the fix. Saving a stub
+mid-authoring is legitimate — the console drops a Start node first and adds
+stages after, `parse_workflow` was made lenient on purpose (issue #661) to allow
+it, and refusing at save would also refuse every existing seed and legacy body on
+its next edit. Saving promises nothing; switching a schedule on promises that
+something happens. Arming is also already the human gate the disarm rule above
+forces a scheduled graph through, and it has the parsed graph in hand for the
+journal name — so the check costs no extra load.
+
+Switching such a workflow **off** stays allowed, on the same principle as the
+unparseable-body case: an operator must always be able to stop a thing. A manual
+(unscheduled) stub is left alone — running one by hand is the author's business,
+and the run says so itself (below).
+
+**A run of a stage-less graph records a notice.** The engine runs such a graph
+happily — no stage fails, because there is no stage — so before this it settled
+as an ordinary finished run, and `QA Test Pipeline` banked six of them. It now
+carries `WorkflowRun.notices` (the issue #638 channel) saying it had nothing to
+run. Deliberately a notice and **not** an error: nothing broke and nothing was
+attempted, so marking it failed would put a half-authored stub into the failure
+count beside runs that genuinely went wrong — the same call issue #925 makes one
+level down with `NoDestinationConfigured`. `WorkflowFile::has_runnable_node` is
+the single predicate behind both halves, for the reason `trigger_schedule` is
+single: two copies that disagreed would let a graph be refused a schedule and
+still run silently, or the reverse.
+
 **In the console.** Each row carries a switch, rendered from `enabled` on the
 list read. Like `editable`, only an explicit `false` counts — a host predating
 issue `#276` sends no field, and `undefined` must not read as paused. A row the
--- a/src/company/mod.rs
+++ b/src/company/mod.rs
@@ -131,9 +131,10 @@ pub use types::{
     grants_search_explicit, grants_workspace_write_explicit, orchestrator_id,
 };
 pub use workflow_file::{
-    WORKFLOW_DESTINATION_KINDS, WORKFLOW_NODE_KINDS, WorkflowDestinationDef, WorkflowEdgeDef,
-    WorkflowFile, WorkflowNodeDef, WorkflowNodeKind, WorkflowRetryDef, list_source_workflows,
-    list_workflows_union, list_workflows_with_globals, load_company_workflows, load_workflow_union,
+    STAGELESS_SCHEDULE_REFUSAL, STAGELESS_WORKFLOW_NOTICE, WORKFLOW_DESTINATION_KINDS,
+    WORKFLOW_NODE_KINDS, WorkflowDestinationDef, WorkflowEdgeDef, WorkflowFile, WorkflowNodeDef,
+    WorkflowNodeKind, WorkflowRetryDef, list_source_workflows, list_workflows_union,
+    list_workflows_with_globals, load_company_workflows, load_workflow_union,
     load_workflow_with_globals, parse_workflow,
 };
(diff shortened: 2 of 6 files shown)


M3gA-Mind:
@coderabbitai review

PR #1036, opened 2026-08-18
Journal every task attempt status change

### LINKED ISSUE #1015: runtime: task attempt status changes emit no event, so run rows are poll-only
A task attempt sits at `pending`/`running` on the detail screen for up to 4 seconds after it has really moved, and does not move at all while the tab is hidden.

## Cause

`RunRecord.status` transitions (`pending → running → waiting_approval|paused → succeeded|failed|cancelled`, `src/ports/runs.rs`) are written by `src/runtime/cycle.rs:301` (`begin_run`), `:515` (`finish_run`) and `:3050` (`create_run`) — and none of them append a `CompanyEvent`. So `project_event` (`src/server/operator.rs:640-1055`) has no arm to project, and attempt status is **poll-only**: `frontend/src/views/TaskDetailView.tsx:341-380` polls every 4s, and `frontend/src/views/TasksView.tsx:312` passes it no event tick at all.

Of the four workflow frames that do exist, none covers attempt status.

## Also: `approval_parked` carries no run correlation

The frame carries `{approvalId, kind, chatId?}` only (`src/server/operator.rs`, ~line 800), so a run that parks a gate — and `RunStatus::WaitingApproval` generally — cannot be applied to the right row live.

## Proposed fix

- Add `CompanyEvent::RunStatusChanged { run_id, task_id, attempt, from, to, error }`, appended inside `begin_run`/`finish_run` (one place, under the same `serial` guard).
- Project it as `run_status_changed`; add `onRunEvent` to `frontend/src/hooks/use-events.ts`; thread a tick into `TaskDetailView`.
- Add optional `runId`/`taskId` to `approval_parked`.

M3gA-Mind:
## Summary

Task attempt status was written by three paths and journalled by none, so the task detail screen could
only **poll**: it sat at `pending`/`running` for up to four seconds after an attempt had really moved,
and did not move at all while the tab was hidden. That is the surface #581 spent its effort removing
when it replaced the five-second whole-company refetch with Snapshot + Refresh — a poll-only screen is
swimming against the direction the console is going.

## The seam: `put_run`, and why that makes this correct by construction

**`put_run` is the single write primitive for every status change in the tree.** Three facts, each
checkable:

1. `begin_run` and `finish_run` are **trait defaults on `RunStore`** whose bodies call `self.put_run`
   (`src/ports/runs.rs:562`, `:593`).
2. **No backend overrides either.** sqlite, MongoDB and the filesystem store implement `put_run` and
   `create_run` and nothing else of the transition surface.
3. **Nothing else in the tree calls `put_run`** — the only two callers are those two defaults.

So decorating `put_run` covers every status change **by construction**, rather than by anyone
remembering six call sites. That is a stronger claim than "events were added at the write paths", and
it is what makes the next section safe rather than lucky: because the defaults dispatch through
`self`, a decorator that overrides only `put_run` sees the transitions too.

### Why not the cycle's call sites — the reaper

The obvious seam is `cycle.rs`, which is what #1015 proposes (`:301`, `:515`, `:3050`). It is wrong.

`reap_orphaned_runs` (`src/ports/runs.rs:684`) settles crash-killed runs by calling `finish_run`
**directly on the store**, never through the cycle. Emitting from the cycle would leave exactly those
runs — the ones whose visibility the reaper exists to provide — transitioning in silence. That is the
partial event surface that is worse than none, because a consumer assumes the frames are complete, and
it would fail in the one case where silence is most harmful.

`the_boot_reaper_journals_the_runs_it_settles` pins it.

### `put_run` stays a documented escape hatch, and it emits

A reviewer's instinct here is to seal it. Two reasons not to.

It is documented as *"the backend seam behind the transition methods — **not** the API … calling it
directly is how a run ends up in a state the rest of the system does not believe in."* It stays that,
and it emits. **A hatch that can skip the legality check but cannot skip being observed is the useful
shape**: the illegal state still lands, and the log still says it did. Sealing it would trade a
visible bad state for an invisible one.

And sealing it would have meant reimplementing `begin_run`/`finish_run` inside the decorator — which
is precisely how a decorator and its port drift apart. Overriding the one primitive they are both
built on keeps a single definition of what a transition is.

## Why not `TurnStarted` / `TurnFailed`

Reuse is the natural suggestion, because `TurnStarted.turn_id` is documented as *"the same id its
`RunRecord` is keyed on"* — the same object, joined by the same key. It still does not fit:

- they are appended **only on the chat HTTP path** (`src/server/operator.rs:1511`), so they do not
  exist for a task attempt at all;
- they **bracket an `OperatorMessage`** — that is their meaning. Firing one for a task attempt would
  put **a turn in the chat transcript that nobody took**;
- they cover accepted/failed, not the status machine (`waiting_approval`, `paused`, `cancelled`).

The `WorkflowRun*` / `WorkflowNode*` frames are a different object — workflow runs, not task attempts.

So this is a new variant, and the surface it joins was **empty rather than partial**: no transition
journalled anything before this, so there was no half-built contract for a new frame to contradict.

## Ordering

The frame is appended **after** the inner write resolves `Ok`, never before. A consumer reacting to it
therefore cannot read a row that has not landed. The reverse order is a worse bug than the polling
this replaces: a reader that re-fetches on the frame would see the **old** status, cache it as current,
and nothing would fire again to correct it.

`the_row_is_durable_before_its_frame_is_appended` proves it by reading the store *from inside the
event log's own `append`* — at the moment the frame exists, the row must already carry the status it
names. It is the only test that fails when the two lines are swapped.

An append that fails is **logged and swallowed**: the status write already happened and is the durable
fact, and turning a journal hiccup into a failed transition would let an observability concern strand
a run. That is exactly why the consumer keeps its poll (below) rather than replacing it.

## API Or Behavior Changes

**New `CompanyEvent::RunStatusChanged { run_id, task_id, attempt, from, to, error }`**, projected to
the operator stream as **`run_status_changed`** with `runId`, `attempt`, `status`, and optional
`taskId` / `from`.

`error` is **not** projected. A failure reason is our own prose about the host and is tenant-scoped, so
the console learns *that* an attempt moved from the frame and reads *why* from the run row — the same
discipline `turn_settled` already applies.

`from` rides along so a consumer holding a row can tell a live frame from a replayed or out-of-order
one, which a bare `to` cannot. It is omitted rather than null on the mint, where there is no prior
state — the presence-check discipline `turn_started`'s `parentId` uses.

Two classifications the compiler forced, both decided rather than swept in:

- **Retention: `Prunable`**, against the three questions the doc comment asks. Not evidence (the
  `RunRecord` is the record of status; the frame carries nothing the row does not). Nothing points at
  it (joined by `run_id`, nothing addressed by its sequence). Nothing reads it back (no boot fold
  consults it — `reap_orphaned_runs` reads the *rows*, by status, which is why it can settle runs this
  frame never covered). And it is high-volume exhaust: several frames per attempt, on every card and
  every chat turn.
- **Neutral in the cycle's trigger scanners**, and this one is load-bearing. The store appends the
  frame *after* a status write made by that very cycle, so treating it as a stimulus would let **a run
  re-trigger itself on every transition it makes**. Same argument as `TaskCardChanged` beside it, and
  sharper.

### Console

`onRunEvent` on `useEvents`, its own counter in the shell, threaded through `TasksView` to
`TaskDetailView`. A counter rather than a payload, like `taskEventTick`: the screen re-reads its own
detail, so two frames collapsing in one React batch still mean "re-read".

Separate from `onTaskEvent` deliberately — an attempt moving does not move a card, and folding it in
would refetch the whole board on every transition of every run, which is the timer load #581 removed.
**No toast**, for the reason the board frames raise none: this fires several times per attempt, and
training the operator to dismiss toasts is how the ones that matter get dismissed.

**The poll stays.** The push half is its own effect (the split `TasksView` uses, for the reason #464
gives: folding it into the polling effect would restart the fallback timer on every frame, so a busy
company would lose the fallback exactly where it is most needed). Visibility-gated like the poll, per
#581. A frame can be lost — the stream can drop, and a failed append is swallowed — so the timer is the
floor under a liveness the frame alone does not guarantee.

## Not in this PR

The issue's secondary ask — optional `runId`/`taskId` on `approval_parked` — is **not** included, and
is largely moot now. Parks settle through `finish_run`, so a run parking for a gate already emits
`RunStatusChanged { to: "waiting_approval", run_id, task_id, .. }`. That correlation was wanted
because there was no other way to apply a park to the right row live; there now is. Happy to add the
fields if a consumer still wants them off the approval frame itself.

## Tests

      PASS. `--no-deps` mirrors CI; an unscoped run drowns in vendored lints CI never sees.
- [x] `cargo build --all-targets` — covered by the clippy and test runs at
      `--features openhuman,tinycortex --all-targets`.
- [x] `cargo test` — **4258 passed / 0 failed** at `--features openhuman,tinycortex --lib`.

Console: `npm run typecheck`, `typecheck:e2e`, `typecheck:unit` all PASS; unit suite **895 passed / 88
files**.

- Stripping the emission from the decorator fails **3 of 4** Rust tests, including
  `the_boot_reaper_journals_the_runs_it_settles` — the case the issue's own proposed seam drops.
- **Swapping the append and the write fails `the_row_is_durable_before_its_frame_is_appended`, and
  only that test** — so it isolates the ordering hazard rather than incidentally covering it.
- Removing the `run_status_changed` dispatch arm fails **2 of 4** console tests.

**Four assertions are negative guards and pass either way** — no toast, no board refetch, and no frame
for a write that revises a settled row's cost without moving its status. Stating that rather than
counting them as coverage.

## Documentation

`src/runtime/run_events.rs` carries the reasoning as module docs: why the store and not the cycle, why
`put_run` is the primitive, why it stays an escape hatch, and the ordering contract. The variant, the
retention arm and the two cycle scanners each state their own decision at the point it is made.

Closes tinyhumansai/opencompany#1015

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -342,6 +342,17 @@ export function AppShell({
   // React batch still means "re-read" — the frame-loss the workflow canvas had
   // to fold an event window to avoid cannot happen to a tick.
   const [taskEventTick, setTaskEventTick] = useState(0);
+  // Issue #1015: bumped on every `run_status_changed`, so the task detail screen
+  // sees an attempt move rather than waiting up to four seconds for its poll —
+  // and sees it at all while the tab is hidden, which the poll deliberately does
+  // not do. Its own counter rather than a share of `taskEventTick`: this fires
+  // several times per attempt, and folding it in would make the whole board
+  // refetch on every transition of every run.
+  //
+  // A counter, not the payload, for the same reason the tick above is one: the
+  // screen re-reads its own detail, so two frames collapsing inside one React
+  // batch still mean "re-read".
+  const [attemptEventTick, setAttemptEventTick] = useState(0);
   // Issue #327: the latest workspace write, as the Workspace view needs it.
   //
   // The payload-carrying variant of the `taskEventTick` pattern above, and the
@@ -1094,6 +1105,7 @@ export function AppShell({
     pendingApprovals: pending,
     onAgentReply: injectAgentReply,
     onTaskEvent: useCallback(() => setTaskEventTick((n) => n + 1), []),
+    onRunEvent: useCallback(() => setAttemptEventTick((n) => n + 1), []),
     // Issue #377. Beside the board tick above, not instead of it: a settle both
     // moves a card between columns and needs saying in the conversation the
     // card came from.
@@ -1290,6 +1302,7 @@ export function AppShell({
               // counter the chat's in-flight strip reads, so a card opened from
               // chat lands on the board without a reload.
               taskEventTick={taskEventTick}
+              attemptEventTick={attemptEventTick}
               // Issue #883: a paused card is blocked until every approval its
               // turn parked is decided, and the board's own read carries none
               // of them. This is the feed the sidebar badge already polls, so
--- a/frontend/src/hooks/use-events.ts
+++ b/frontend/src/hooks/use-events.ts
@@ -53,6 +53,32 @@ export type CompanyStreamEvent =
   // the card sits. There is **no title and no note** on purpose — the card's
   // text lives on `GET …/tasks`, and a console reacts to this frame by
   // re-reading that, so the board's content has exactly one source.
+  // One task attempt moved (issue #1015). Before this the task detail screen
+  // could only learn an attempt's status by polling every four seconds — and not
+  // at all while the tab was hidden — which is the surface #581 removed
+  // elsewhere when it replaced the whole-company refetch with Snapshot +
+  // Refresh.
+  //
+  // Deliberately thin, like `task_card_changed` beside it: ids, the status, and
+  // where it came from. There is **no error text** on purpose — a failure reason
+  // is tenant-scoped, so a console reacts to this frame by re-reading the run
+  // row, which is the one place that answers *why*.
+  | {
+      type: "run_status_changed";
+      seq: number;
+      atMillis: number;
+      runId: string;
+      /** Absent for a chat turn, which is an attempt at work with no card. */
+      taskId?: string;
+      /** 1-based attempt ordinal at that card. */
+      attempt: number;
+      /** The status moved to, widened so a word from a newer host is not a type
+       *  error. */
+      status: string;
+      /** The status moved from. Absent on the mint, which has no prior state —
+       *  a presence check, matching `turn_started`'s `parentId`. */
+      from?: string;
+    }
   | {
       type: "task_card_changed";
       seq: number;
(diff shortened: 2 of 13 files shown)


M3gA-Mind:
@coderabbitai review

Issue #963, opened 2026-08-18
workflows: 18 company templates have an output node and no desk to deliver it to


M3gA-Mind:
## Summary

18 company templates ship a workflow with an `output` node and define **no `[[group_chat]]` desk**, so there is nothing their pipeline report can be delivered to. Split out of #947, which fixes the 3 templates that do have desks.

## The 18

`agentic_accounting_firm` · `agentic_consultation_firm` · `agentic_customer_support` · `agentic_design_studio` · `agentic_enterprise_sales` · `agentic_game_business` · `agentic_game_studio` · `agentic_influencer_business` · `agentic_law_firm` · `agentic_media_company` · `agentic_pharma_startup` · `agentic_realestate_company` · `agentic_recruiting_company` · `agentic_research_lab` · `agentic_venture_capital` · `agentic_venture_studio` · `signals_opportunity_studio` · `startup_accelerator`

Each has exactly one workflow with an output node (`month_end_close`, `engagement_pipeline`, `ticket_pipeline`, `design_pipeline`, `sales_pipeline`, `liveops_pipeline`, `game_build_pipeline`, `content_pipeline`, `matter_pipeline`, `newsroom_pipeline`, `discovery_pipeline`, `acquisition_pipeline`, `hiring_pipeline`, `research_loop`, `diligence_pipeline`, `venture_launch_pipeline`, `opportunity_pipeline`, `cohort_pipeline`).

## Why none of the existing destination kinds solves it

Established while working #947, against `main` @ `0b332148`. All three closed kinds (`WORKFLOW_DESTINATION_KINDS`, `src/company/workflow_file.rs:52`) fail here:

- **`owner`** — fails on a freshly provisioned tenant. With no mailbox the delivery reports `Failed` / `OwnerFallbackFailed`. Note the harness in `delivery.rs`'s own test sets `with_channel = true` and it *still* fails, so having a channel adapter registered does not rescue it. Shipping this would turn every one of these templates' runs red.
- **`channel`** — needs a target, and these manifests declare no desk, so there is nothing addressable. `DeskChannel`s are registered per manifest desk at `src/runtime/builder.rs:1820`.
- **`email`** — would mean hardcoding a recipient address into a shipped template.

The operator channel is not an escape hatch either. It is explicitly disqualified at `src/workflows/delivery.rs:684`:

> *"try the operator adapter **only to produce the explicit failure row**. Its buffer is **not a workflow delivery surface**, so this cannot report a successful discard."*

[...]

PR #1038, opened 2026-08-18
Deliver the three desk-bearing templates' pipeline reports

### LINKED ISSUE #947: workflows: all 21 seeded company templates ship an output node with no destination, so no template run can deliver

M3gA-Mind:
## Summary

21 seeded workflows carry an `output` node and **none declared a destination**, so no template run
could deliver anything. This fixes the three templates that can be fixed, and says plainly why the
other 18 cannot be.

**The issue's central premise does not hold.** #947 states *"Every template defines `[[group_chat]]`
desks"*, and the desk-channel recommendation rests on it. Against `main`, only **3 of the 21**
companies with an output node declare any desk:

```
$ for f in companies/*/company.toml; do echo "$(grep -c '^\[\[group_chat\]\]' $f)  $(basename $(dirname $f))"; done | sort -rn
3  agentic_software_company     engineering, product_design, go_to_market
3  agentic_marketing_agency     strategy, creative, growth
2  e2e_harness                  engineering, content
2  openhuman_demo               (no output-node workflow — not one of the 21)
0  × 18 others
```

The 18 are tracked in #963 with the full reasoning. In short: `owner` reports
`Failed`/`OwnerFallbackFailed` on a tenant with no mailbox (`delivery.rs`'s own test sets
`with_channel = true` and it still fails), `channel` has nothing to target, `email` would hardcode a
recipient into a shipped template, and the operator adapter is explicitly disqualified at
`src/workflows/delivery.rs` — *"its buffer is **not** a workflow delivery surface, so this cannot
report a successful discard."* Giving 18 businesses desks is a content judgement per company, not a
stanza, and folding it in here would restructure them under a delivery heading — the "buried under
diff" problem #947 was itself split from #925 to avoid.

`agentic_research_lab` is desk-less **deliberately** — its manifest says so — so it likely needs a
different answer from the other 17, which is noted on #963.

## The three judgements

One per template, each a decision about *who the report is for*:

| Template | Desk | Why |
|---|---|---|
| `agentic_software_company` / `feature_pipeline` | `go_to_market` | Engineering built the feature and does not need a report of its own work. Docs, devrel and support are the people whose next action depends on knowing what shipped — and the pipeline already ends on `docs_writer`, who sits on this desk. |
| `agentic_marketing_agency` / `campaign_pipeline` | `strategy` | The report closes the loop the brief opened: the strategist turned the client's brief into the angle, and "what shipped" is only meaningful read against it. Creative made the post and already saw it; growth runs acquisition on a campaign this pipeline does not measure. |
| `e2e_harness` / `committed` | `content` | The one node before the output is `writer` producing a written draft, and this harness's only other desk is engineering. The draft is the deliverable. |

## Verified for real, not asserted from the TOML

Built the host, ran it on `e2e_harness` against a **fresh data dir** — a genuinely freshly provisioned
tenant, no mailbox — signed in through the dev-code flow, and ran the workflow:

```
deliveries: [{"node":"done","kind":"channel","target":"content","status":"sent",
              "detail":"posted to the channel","reason":"channel-posted"}]
```

and the report is readable at `GET /api/v1/company/chat/history?desk=content`, the endpoint the
console renders — no canvas involved:

> `[E2E Harness Co] Committed flow — Hand it to the operator` …

**The before/after was run too.** With the destination reverted, on a fresh tenant, the same run gives
`{"kind":"none","status":"skipped","reason":"no-destination-configured"}` and the content desk chat is
`[]`. So both halves of the acceptance criterion are demonstrated rather than argued.

## Relationship to #995

#995 refuses an undeliverable channel destination **at save**, and it does not constrain this PR —
deliberately. Its own guard says why:

> *at TOML parse time: seed templates are parsed with no runtime in hand, and checking there would
> refuse to boot a template on a company whose desks are not resolved yet.*

So seeded templates never pass through that guard. The test below is its counterpart for the surface
it cannot cover, and the three targets here are real desk ids, so they are inside
`deliverable_channel_ids` at run time.

## API Or Behavior Changes

Three seeded templates' `output` nodes gain a `[node.destination]` of `kind = "channel"`. A run of
those workflows now posts its report to a desk chat instead of delivering nothing. No code path
changes; no route, schema or default changes. The other 18 templates are untouched and still deliver
nothing — which #925's `Skipped` / `NoDestinationConfigured` row now reports honestly.

## Tests

`every_seeded_output_destination_resolves_against_its_own_manifest` walks all 21 seeded workflows and
checks each destination against **its own manifest's desks**, because a `channel` target is a
`ChannelAdapter` id and a company's adapters are one per desk. A target naming a desk its manifest
does not declare parses fine, ships, and fails at run time on a fresh tenant — the #947 class of bug
one step further along.

The 18 are an explicit **allowlist**, not an `is_empty()` skip, so it fails in both directions: a
desk-bearing template that loses its destination fails, and a listed template that gains desks fails
until it gets one.

- [x] `cargo build --all-targets` — covered by the clippy and test runs above; the host binary was
      also built to drive the live verification.
- [x] `cargo test` — **4279 passed / 0 failed** at `--features openhuman,tinycortex --lib`.

- Removing `feature_pipeline`'s destination → *"…has an output node with no destination, so a run of
  it delivers nothing."*
- Changing its target to `gtm` → *"…delivers to channel `gtm`, which is not a desk
  agentic_software_company's manifest declares. Declared: ["engineering", "product_design",
  "go_to_market"]"*
- End to end, reverted on a fresh tenant: `status: "skipped"`, `reason: "no-destination-configured"`,
  and an empty desk chat.

## Documentation

Each destination carries its judgement as a comment at the node, so the "why this desk" is read where
the choice is made. The test states why the 18 are an allowlist and points at #963.

Closes tinyhumansai/opencompany#947

### DIFF
--- a/companies/agentic_marketing_agency/workflows/campaign_pipeline.toml
+++ b/companies/agentic_marketing_agency/workflows/campaign_pipeline.toml
@@ -84,6 +84,14 @@ kind = "output"
 name = "Report back"
 summary = "Summarize what shipped."
 
+# Strategy, because this report closes the loop the brief opened: the strategist
+# turned the client's brief into the angle, and "what shipped" is only meaningful
+# read against it. Creative made the post and already saw it; growth runs
+# acquisition on a campaign this pipeline does not measure.
+[node.destination]
+kind = "channel"
+target = "strategy"
+
 [[edge]]
 from = "brief"
 to = "strategist"
--- a/companies/agentic_software_company/workflows/feature_pipeline.toml
+++ b/companies/agentic_software_company/workflows/feature_pipeline.toml
@@ -77,6 +77,14 @@ kind = "output"
 name = "Ship and report"
 summary = "Release behind approval and report back."
 
+# Go-to-market, not engineering: engineering built the thing and does not need a
+# report of its own work, while docs, devrel and support are the people whose
+# next action depends on knowing what shipped. The pipeline already ends on
+# `docs_writer`, who sits on this desk.
+[node.destination]
+kind = "channel"
+target = "go_to_market"
+
 [[edge]]
 from = "request"
 to = "spec"
--- a/companies/e2e_harness/workflows/committed.toml
+++ b/companies/e2e_harness/workflows/committed.toml
@@ -40,6 +40,13 @@ kind = "output"
 name = "Hand it to the operator"
 summary = "Park the draft for review."
 
+# Content: the one node before this is `writer` producing a written draft, and
+# this harness's other desk is engineering. The draft is the deliverable, so the
+# desk that reviews written work is the audience.
+[node.destination]
+kind = "channel"
+target = "content"
+
 [[edge]]
 from = "start"
 to = "draft"
--- a/src/company/content_test.rs
+++ b/src/company/content_test.rs
@@ -6,6 +6,7 @@
 
 use std::path::{Path, PathBuf};
 
+use super::workflow_file::WorkflowNodeKind;
 use super::{
     CompanyManifest, Tools, grants_chargebee_explicit, grants_composio_explicit,
     grants_media_explicit, grants_paypal_explicit, grants_search_explicit, load_dir_skills,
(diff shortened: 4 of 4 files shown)


M3gA-Mind:
@coderabbitai review

Issue #966, opened 2026-08-18
console: desk history renders pre-#885 replies with a confident but unrecoverable author


M3gA-Mind:
## What is needed

Desk transcripts contain messages whose author is **permanently unknown**, and the console currently renders them as though the stored value were an author.

`tinyhumansai/opencompany#885` fixed the write path: `CompanyEvent::AgentReply.agent_id` no longer receives `OutboundMessage.channel` (a destination) in place of the author. That fix is **forward-only**. Every reply journaled before it carries a destination string — usually the literal `operator` — in a field whose contract is *"the agent that produced the reply"*.

Those rows cannot be repaired. The author was overwritten in place and nothing else on disk records it:

- `steps` — `TurnStep` has no agent field
- `task_id` — `None` on exactly these rows (it is set on the dispatch path, the one writer that was already correct)
- `parent` — names the question, never the answerer
- `chat_id` — the desk, which yields *today's* desk lead; membership is mutable (manifest members unioned with operator-added overlay members), so that is a re-derivation against current state, silently wrong for any desk whose lead has changed
- no `RunRecord` exists for chat turns; `CompressedTrace` is `{cycle_id, summary, at_millis}`
- metering is bucketed per calendar **day** with per-agent aggregates, so it cannot name the author of one message

## Why this needs a decision rather than a default

Today `MessageView::project` maps `AgentReply.agent_id` straight to `author`, so a pre-fix row renders as a confident, specific, wrong name — "the operator said this" about something the company said. On a desk with four members, the author is the single most useful fact in the log, and a wrong one is worse than an absent one. It is the same class of mistake as a fabricated citation: the surface asserts something it does not know.

**An admitted "unknown" beats a confident wrong name.** What that should look like is a product decision, not something to settle inside a bug fix, which is why #885 deliberately did not widen to cover it.

## Options

1. **Render unknown.** A pre-fix reply shows "author unknown" (or an unattributed company bubble) instead of a name. Truthful; costs the reader a name on old rows that never had one.
2. **Mark as legacy.** Keep the stored value but visibly flag it as unverified attribution, so a reader knows not to trust it.
3. **Do nothing.** Old rows keep lying. Recorded as an option only so the choice is deliberate.

Option 1 is the one that matches how the rest of the console treats what it does not know, but the call is the product's.

## How to identify the affected rows

[...]

PR #1040, opened 2026-08-18
fix(chat): author a workflow-copilot reply to the copilot, not the operator channel

M3gA-Mind:
## Summary

A workflow-copilot reply was journaled as though the **operator** had written it.

`#885` established the rule: `CompanyEvent::AgentReply.agent_id` is *"the agent that
produced the reply"*, and `OutboundMessage.channel` is *"the channel to emit on"* — a
destination. It taught the main operator bubble to carry its responder and gave the
journal writers a `channel` fallback for producers that name no author.

The confined copilot turn (#416) is one of those producers. It pushes its bubble on the
operator channel with no author, so the fallback stamps `agent_id: "operator"` on a reply
an agent genuinely produced.

**This is the #885 defect still happening, not history needing a label.** It matters
because #966 is about presenting rows whose author is unrecoverable, and this path was
still *adding* to that set on every copilot turn. There is no point captioning a bucket
that is still filling.

## Why `Refs #966` and not `Closes`

**#966 stays open after this merges.** That issue asks how the console should present desk rows
whose author is permanently unrecoverable; this stops one path from creating more of them.
Marking it resolved here would claim the actual question was answered when only its prerequisite was —
and a cross-referenced PR already shows on the issue timeline as progress, so nothing is
lost by being accurate about which half landed. The presentation half follows as its own
PR, and that one carries the closing keyword.

### 1. The copilot turn carries its author

`confined_bubble` names the construction rather than inlining it, for a specific reason:
its **author** is the load-bearing field, it was wrong, and the branch is otherwise
reachable only by standing up a scripted model endpoint and a whole harness pool. A named
function is what lets the fix be asserted at all — see the test note below.

### 2. `CONFINED_AGENT_ID` moves to `ports::ids`

`harness` is `#[cfg(feature = "openhuman")]`. The attribution audit from #965 lives in
`server::chat_history` and compiles in the **default** build, where the harness does not —
and it has to recognise this id. Reading `harness::confine::CONFINED_AGENT_ID` from there
does not compile on the default lane; copying the literal instead would leave two copies
free to drift, which is the hazard the console's `mapComposioCategory` carries a written
warning about.

So the constant lives in the ungated id module beside `generate_id` and `agent_slug`, and
`confine` re-exports it. **Every existing `confine::CONFINED_AGENT_ID` call site is
unchanged.**

### 3. `is_known_author`, and the false positive it prevents

This is the part that would have gone wrong quietly.

`CONFINED_AGENT_ID` is **deliberately not a roster id** — `confine` documents it as naming
no teammate, carrying no manifest grants, and being unaddressable. The audit's rule is *"an
`agent_id` naming no roster teammate"*. So writing the copilot's real id would have made
**every future copilot reply register as affected**, and the blast-radius figure would climb
on a company doing nothing wrong. That number is the one thing in #965 that has to be
trustworthy.

`is_known_author(agent_id, record)` — the roster **plus** the confined copilot — is that
recognition, extracted as the single predicate the audit and any future presentation of its
result must share. Two copies would let the count and the rendering disagree about which
rows are unknown, which is precisely the bug class this whole thread is about.

## API Or Behavior Changes

**No wire change.** No route, no HTTP/GraphQL type, no new field. `Cargo.lock` untouched.

**Behaviour.** A workflow-copilot reply is journaled against `workflow-copilot` instead of
`operator`. Replies journaled before this are not rewritten — their author is unrecoverable
(see #965) and nothing here synthesises one.

**Audit.** The attribution count no longer treats the confined copilot as an affected row.
Existing pre-fix rows are unaffected by that: they store `operator`, not the copilot id.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — run as CI does, `--locked --no-deps`,
      both the default lane and `--features openhuman,tinycortex`. Both clean.
- [x] `cargo build --all-targets` — run as `cargo check --locked --all-targets` and the
      gated equivalent. (`check`, not `build`: nothing here is a linker-visible change.)
- [x] `cargo test` — `cargo test --locked --features openhuman,tinycortex --lib`:
      **4281 passed, 0 failed, 3 ignored.**

**Verified after rebasing onto current `main`**, not against the base the branch was cut
from. `src/harness/brain.rs` had moved underneath it in the meantime, so the pre-rebase run
(4257 passing) was against a tree `main` no longer had. The 24-test difference is `main`'s
newer tests arriving with the rebase.

### Every new test was proven to fail with its fix reverted — 3 of 3

|---|---|
| `confined_bubble` stops setting the author | `a_copilot_turn_is_authored_by_the_copilot_not_the_operator_channel` |
| `is_known_author` drops the confined arm | `the_confined_copilot_is_a_known_author_not_an_affected_row` |
| …the same revert, on the roster half | `the_confined_copilot_is_not_reachable_through_the_roster_alone` |

**Two of these initially proved nothing, and are recorded because a test that cannot fail is
worse than no test.**

1. The two predicate tests first re-implemented `is_known_author` as a local helper in the
   test module. Reverting the production function changed nothing — they proved only that
   the test agreed with itself. They now build a real `CompanyRecord` and call the actual
   function.
2. The `brain.rs` fix — the live defect, the whole point of this PR — had **no test at all**.
   Reverting it failed nothing. That is the same gap that let #885 happen: nothing tested
   what the write path stores. `confined_bubble` exists so that gap could be closed without
   a full turn harness.

The writer test asserts the author is *not* the channel, not merely that it equals the
constant: the two being conflated is the entire defect, and an equality-only assertion would
still pass if `CONFINED_AGENT_ID` were ever redefined to `"operator"`.

- **No test drives a copilot turn end to end.** `confined_bubble` is asserted directly; that
  it is *called* on the copilot branch is guarded by the compiler and by reading, not by a
  test. A real one needs the scripted-endpoint harness the `*_turn_test.rs` files use, and
  this PR does not add one.
- **No count against real data.** #965's audit still has no measured blast radius; that needs
  a deployed host, as recorded there.

## Documentation

No spec doc changes. The reasoning lives at the code it governs: `confined_bubble` carries
why it is a named function, `ports::ids` carries why the constant moved out of the harness,
and `is_known_author` carries the false positive it exists to prevent.

Refs tinyhumansai/opencompany#966

### DIFF
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -191,6 +191,35 @@ pub struct HarnessBrain {
     runs: Option<Arc<dyn crate::ports::RunStore>>,
 }
 
+/// The single bubble a workflow-copilot turn returns (issues #416, #966).
+///
+/// Named rather than inlined because its **author** is the load-bearing field
+/// and it was wrong. #885 taught the main operator bubble to carry its
+/// responder and left this branch on `None`, so a genuine copilot reply kept
+/// journaling as `agent_id: "operator"` — the #885 defect still happening, not
+/// history needing a label. A function is what lets that be asserted without
+/// standing up a scripted model endpoint and a whole harness pool.
+///
+/// `CONFINED_AGENT_ID` is deliberately not a roster id (see [`confine`]): it
+/// names no teammate and cannot be addressed. That makes it a **truthful**
+/// author rather than a resolvable one, which is why
+/// `chat_history::is_known_author` has to know it — otherwise this row trades
+/// one wrong answer for a permanent false positive in the attribution audit.
+///
+/// No card, by construction: a confined turn has no `spawn_task` to call, and
+/// the chat handler does not open one from a copilot message either.
+fn confined_bubble(outcome: crate::harness::TurnOutcome) -> OutboundMessage {
+    OutboundMessage {
+        message_id: None,
+        task_id: None,
+        channel: "operator".to_string(),
+        agent: Some(confine::CONFINED_AGENT_ID.to_string()),
+        text: outcome.reply,
+        reply_to: None,
+        steps: outcome.steps,
+    }
+}
+
 impl HarnessBrain {
     /// Builds a harness brain for `record`, answering unaddressed operator
     /// messages with the company orchestrator (the `tier = "orchestrator"` agent,
@@ -2659,18 +2688,7 @@ impl HarnessBrain {
                                 &confinement,
                             )
                             .await?;
-                        channel_responses.push(OutboundMessage {
-                            message_id: None,
-                            // No card, by construction: a confined turn has no
-                            // `spawn_task` to call, and the chat handler does
-                            // not open one from a copilot message either.
-                            task_id: None,
-                            channel: "operator".to_string(),
-                            agent: None,
-                            text: outcome.reply,
-                            reply_to: None,
-                            steps: outcome.steps,
-                        });
+                        channel_responses.push(confined_bubble(outcome));
                         continue;
                     }
                     // Route to the addressed desk's lead, else the orchestrator.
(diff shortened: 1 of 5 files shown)


M3gA-Mind:
@coderabbitai review

Issue #877, opened 2026-08-15
approvals: the gate asks what a tool is called, not what the call is doing


oxoxDev:
## Summary

**OpenHuman asks what this call is doing. OpenCompany asks what this tool is called.** That one difference is where most of our approval noise comes from, and it explains why the same agent, running the same toolbelt, interrupts an operator far more often here than it does on the desktop product.

The tool name is a coarse proxy for consequence. `shell` can be `grep` or `rm -rf /`. `git_operations` can be `status` or `push`. `mcp_call_tool` can be a list or a wire transfer. Classifying the name means the *worst thing a tool could do* is charged to *every call it makes* — so an agent doing something harmless pays the interruption for something it did not do.

This issue is the general statement. #875 is the first instance (shell), fixed in #876; this one exists so the remaining ones are not each rediscovered from a screenshot.

## Where each side asks its question

| | OpenHuman (`vendor/openhuman`) | OpenCompany |
|---|---|---|
| Input to the decision | the call's **arguments**, per tool: `Tool::external_effect_with_args(args)` | the **tool name**, via the `DECLARED` table in `src/policy/consequence.rs` |
| Who answers | the tool itself — `shell` classifies the command, `file_write` checks the path, `git_operations` checks the operation, `schedule` checks the action, `composio` gates only `action="execute"` | one static table, with two argument-aware exceptions: `composio_execute` by action slug (#441) and `web_fetch` by URL host (#673) |
| The gradient | `classify_command` → `Read` / `Write` / `Network` / `Install` / `Destructive`, maximum across unquoted segments, redirects and `tee` lift to `Write`, unrecognised is `Write` | a fixed `Reach` per tool name |

The seam on our side is already argument-aware: `ToolPolicyRequest` carries `arguments`, and `consequence_of(tool, args)` takes them. Every fix of this shape is a classifier, not a re-plumbing — #876 is about forty lines and adds no new list, because it calls the vendored classifier the desktop product already gates with.

## What is left

In the order they cost an operator today:

1. **`git_operations`** — upstream grades it by the `operation` argument (`requires_write_access(operation)`); we treat every operation as a push. `status`, `log`, `diff` and `show` are reads of the agent's own workspace. This was named in #560's deferred scope alongside `shell` and is the obvious next one.
[...]

Issue #875, opened 2026-08-15
approvals: shell parks on read-only commands because the gate reads the tool name, not the command


oxoxDev:
## Summary

An agent that investigates by grepping its own workspace asks the operator for permission on every command. `shell` is declared `Reach::Consequence` by **tool name** (`src/policy/consequence.rs:420`), and nothing on the approval path ever looks at the command, so `grep -c foo *.jsonl` and `rm -rf /` are the same input to the classifier. They park identically, in `supervised` **and** in `auto`.

This is the scope #560 named and deferred:

> **This issue stays open** for the `shell` / `git_operations` argv-grading scope added above, which #604 does not deliver.

#560 was later closed as shipped — the closing comment covers the tier only — and nothing carried the grading forward. This issue is that scope, restated against what is on `main` now.

Observed on staging (`smoke1`, tier `auto`, `always_approve` empty) on 2026-08-15: five pending approvals, four of them `shell`, all from `qa_engineer` in one batch, all reads inside its own workspace:

```
cd …/qa_engineer/workspace && grep -l -i "resets\|forgot" session_raw/*.jsonl
cd …/qa_engineer/workspace && grep -c -i "plus\|+b@\|plus-addressing" session_raw/*.jsonl | grep -v ":0"
cd …/backend_engineer && find . -maxdepth 4 -type d | grep -vE "node_modules|\.git" | head
```

`auto` tells the operator: *"the agents work on their own and stop before anything that leaves the company or spends money."* None of those commands does either.

## The upstream runtime already grades this, and we vendor it

OpenHuman gates the same tool per call. `ShellTool::external_effect_with_args` (`vendor/openhuman/…/tools/impl/system/shell.rs:255`) classifies the actual command and asks the tier what to do with that class:

```rust
let mut class = self.security.classify_command(command);
if let Some(declared) = args.get("category")… { class = class.max(declared); }  // escalate-only
self.security.gate_decision(class) == GateDecision::Prompt
```

`SecurityPolicy::classify_command` (`security/policy/command_checks.rs:126`) splits the command into unquoted segments, classifies each, and takes the **max** — so `grep x && rm -rf /` is `Destructive`, not `Read`. Redirects and `tee` lift to `Write`. Anything unrecognised is `Write`, fail-closed. The classes are `Read` / `Write` / `Network` / `Install` / `Destructive`, and only `Read` runs unprompted in an acting tier.

[...]

PR #1043, opened 2026-08-18
fix(policy): grade git_operations by its operation, not its name

M3gA-Mind:
## Summary

Part of tinyhumansai/opencompany#877 — the `git_operations` instance.

`git status` and `git push` cost an operator the same interruption, because the gate classified
the **tool name**. `status`, `diff`, `log`, `show`, `branch` and `rev-parse` only read the
repository, and an agent orienting in its own workspace should not have to ask.

This is the fourth argument-graded tool, after `composio_execute` (#441), `web_fetch` (#673) and
`shell` (#875/#876).

## ⚠️ Read this before merging: the exposure this downgrade accepts

Stated up front rather than buried, because it is a deliberate scope decision and whoever merges
this is accepting it.

`code_tools` builds `GitOperationsTool::new(security, workspace)` against **the agent's own
workspace** — the same directory `file_write` writes into. Upstream's `run_git_command_in` is a
bare `Command::new("git").args(args).current_dir(cwd)` with **no `GIT_CONFIG_NOSYSTEM`, no `-c`
overrides and no environment scrub** (I grepped the pinned file for `GIT_CONFIG`, `env_clear` and
`.env(` — there is none). Several git config keys name a command to run — `core.fsmonitor`,
`core.pager`, `diff.external`, `core.sshCommand` — and the repository config lives where the agent
can write it.

That is the identical primitive `read_workspace_state` is gated for. Its note in `DECLARED` says to
revert that stopgap *"once a hardened `run_git` is vendored, **and not before**"*, tracking the work
at `tinyhumansai/openhuman#5494`.

So after this lands, `git_operations status` runs unattended while `read_workspace_state` — which
runs `git status` in the same directory — still parks. **When #5494 lands the two should be
reconciled**, either both downgraded or both gated, because they run the same command against the
same directory. The trade-off is recorded in a `⚠️ The exposure this downgrade accepts` block on the
classifier itself, so it is discoverable from the code and not only from here.

## Why the upstream helper is NOT called from the gate

A reviewer's first instinct will be *"why not just reuse `external_effect_with_args`?"*, and the
answer matters enough that it is written into the classifier's doc comment as well as here —
without it, someone will "simplify" this into a security hole later.

`Tool::external_effect_with_args` is the only **public** route into upstream's judgement, and it is
**tier-coupled**:

```rust
self.requires_write_access(operation)
    && self.security.gate_decision(CommandClass::Write) == GateDecision::Prompt
```

Calling that here would import OpenHuman's desktop tier into a gate that answers the tier question
one layer up — and **under a policy whose `gate_decision` is not `Prompt` it returns `false` for a
genuine write. It fails open.** That is the worst possible direction for a helper whose whole job is
deciding what runs unattended.

There is no untainted route to borrow instead: `requires_write_access` and `is_read_only` are
**private inherent methods** on the tool, not part of any trait.

So the read set is a local copy — and a copy that can drift silently is exactly what #877 warns
against. The vendored hook is therefore used as a **test oracle** rather than a runtime dependency:
`the_read_only_set_matches_the_vendored_classifier` drives it at `SecurityPolicy::default()`, which
is `AutonomyLevel::Supervised` — the one tier where `gate_decision(Write)` *is* `Prompt` and the
conjunction reduces to the operation test alone. Upstream reclassifying any of these fails the build
here. The test also asserts `commit` comes back `true` through the same call, so it cannot pass
vacuously.

That gets upstream's judgement as a cross-check without taking a runtime dependency that can fail
open.

## Fail-closed, and durably so

All five `shell_consequence` mechanisms:

1. The gated verdict is built first and every early return uses it.
2. A missing or non-string `operation` gates — the tool's schema marks it required, so such a call
   could not have run.
3. Only **affirmative** membership of `GIT_READ_ONLY_OPERATIONS` downgrades.
4. Comparison is exact and case-sensitive, matching upstream's `matches!`.
5. There is no self-declared hint here, so the escalate-only rule `shell` needs is satisfied
   vacuously — nothing exists that could lower a verdict.

**`push`, `pull`, `fetch`, `merge`, `rebase` and `clone` are in neither upstream list** —
`requires_write_access` does not name them and `is_read_only` does not either. They gate **by
construction**, not by a rule anyone has to remember to maintain. That is a durability property, not
an implementation detail: an operation added upstream tomorrow gates by default, and `push` — the one
that reaches an address this layer never sees — is covered by the same mechanism as a typo.

## Scope: why only `git_operations`

From the survey on #877, so nobody re-proposes the ones already settled.

**Grading by argument today — three tools**

| Tool | Since | Keyed on |
| --- | --- | --- |
| `composio_execute` | #441 | action slug (`tool` arg) |
| `web_fetch` | #673 | URL host |
| `shell` | #875 / #876 | the command, via vendored `SecurityPolicy::classify_command` |

Everything else — 60-odd `DECLARED` entries — answers from the name.

**Varies by argument, no grader yet**

| Tool | Varies by | Vendored grader | Assessment |
| --- | --- | --- | --- |
| `git_operations` | `operation` | yes (private; see above) | **this PR** |
| `mcp_call_tool`, `mcp_registry_tool_call` | server + tool | none | needs per-server advertised metadata — **a design, not a classifier**; its own issue |
| `http_request`, `curl` | method + host | none | ruled out, below |
| `workspace_write` / `_create` / `_delete` / `_rename` | whose note | none | #877's own lowest rank |
| `file_write`, `edit`, `apply_patch` | path | upstream checks path | already `Grantable`; lower value |

**Ruled out, with reasons**

* **`http_request` / `curl` — settled, do not relitigate.** `consequence.rs:322` already decided it:
  only `web_fetch` gets host scoping because the other two "can mutate, so a host-scoped grant on
  them would consent to *writing* to that host — a different act from the read this issue is about,
  and one nobody asked for."
* **⚠️ #614 is not a precedent for argument grading.** It is *"workflows: an `http_request` node never
  consults `ApprovalPolicy`"* — a **plumbing** fix that made the node reach the gate at all. It has
  already been cited twice as precedent; it is not one.
* **`read_workspace_state` — deliberately gated, and it explains itself** via `denial_reason`
  (`consequence.rs:805`). See the exposure section above.
* **The ~55 remaining name-only tools are not a backlog.** `send_email` is a send whatever its body
  says. Grading them would add parsers whose failure mode — mis-reading an argument as safe — is
  strictly worse than the gate they replace.

## API Or Behavior Changes

New `pub const GIT_OPERATIONS`; `consequence_of` gains a `git_operations` arm.

Behaviour: a `git_operations` call whose `operation` is one of the six read subcommands no longer
parks for approval under `supervised`/`auto`, and is no longer refused under `readonly`. Every other
operation — and any call whose `operation` cannot be read — behaves exactly as before.

## Tests

      compiles.
- [x] `cargo build --all-targets` — covered by the two test lanes below, which build all targets for
      both feature sets.
      **4284 passed / 0 failed**. `RUST_MIN_STACK` matches `ci.yml:626`; without it a pre-existing
      `harness::brain` test overflows the stack locally.

`scripts/ci/assert-feature-lanes.sh` passes.

### Revert proof — three, each isolating a different property

| Reverted | Result |
| --- | --- |
| the `consequence_of` dispatcher arm | `a_git_read_operation_does_not_park` **FAILED** |
| allowlist → denylist, missing `operation` defaults to a read | **4 FAILED**, incl. `an_unrecognised_git_operation_still_parks` and `a_git_call_with_no_readable_operation_parks` |
| added `commit` to `GIT_READ_ONLY_OPERATIONS` | `the_read_only_set_matches_the_vendored_classifier` and `a_git_write_operation_still_parks` **FAILED** |

The second is the fail-closed requirement proven directly rather than asserted. It also tripped the
pre-existing `the_auto_tier_line_is_pinned_tool_by_tool` guard, which is that guard working.

## Documentation

`N/A` — no module doc or spec describes the per-tool consequence table; the reasoning that would go
in one lives on the classifier itself, which is where the three existing argument-graded tools keep
theirs. The two blocks worth reading are `⚠️ The exposure this downgrade accepts` and
`Why a local list rather than the vendored hook`, both on `git_operations_consequence`.

### DIFF
--- a/src/policy/consequence.rs
+++ b/src/policy/consequence.rs
@@ -305,6 +305,16 @@ pub(crate) const COMPOSIO_ACTION_KEY: &str = "tool";
 /// name (issue #875).
 pub const SHELL: &str = "shell";
 
+/// The git tool, classified by the `operation` it was handed rather than by
+/// this name (issue #877).
+pub const GIT_OPERATIONS: &str = "git_operations";
+
+/// The argument `git_operations` names its subcommand in. Shared with the
+/// fixtures for the reason [`COMPOSIO_ACTION_KEY`] is: a key hard-coded
+/// separately in a test is a test that stops reaching the classifier without
+/// saying so (issue #470).
+pub(crate) const GIT_OPERATION_KEY: &str = "operation";
+
 /// The argument key [`SHELL`] carries the command line under.
 ///
 /// A required parameter of the vendored tool's schema, so a call that omits it
@@ -779,6 +789,12 @@ pub fn consequence_of(tool: &str, args: &serde_json::Value) -> Consequence {
     if name == SHELL {
         return shell_consequence(args);
     }
+    // Issue #877: the fourth. `git_operations` is how an agent orients in its own
+    // workspace, and classifying the NAME charged a `git status` the same
+    // interruption as a `git push` to a configured remote.
+    if name == GIT_OPERATIONS {
+        return git_operations_consequence(args);
+    }
     match DECLARED.iter().find(|d| d.tool == name) {
         Some(found) => Consequence {
             group: found.group,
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

PR #1044, opened 2026-08-18
Implement the brand guide's colour system and screen layouts

graycyrus:
Implements the `opencompany-prod` brand guide: the colour system, and the chat
and board layouts its mockup frames draw.

## Colour system

The single `--gray-*` ramp is retired for the guide's two surface ladders and
five ink levels. Light and dark are drawn as independent ladders whose rungs
carry the same *roles* rather than the same lightnesses, so the rung is the
primitive and the theme picks a set.

- **Brand** re-anchored on `#7153F0`. The other nine steps hold the ramp's
  original lightness cadence and hue drift with chroma scaled so 500 lands
  exactly on it; every step verified in sRGB gamut. White-on-brand improves
  4.70:1 → 5.00:1.
- **Surfaces** — all 12 values are the guide's, unchanged.
- **Ink** — five levels per theme, exposed as `text-ink-*`.

### Two things worth review attention

**The light-surface captions in Figma are wrong.** s2/s3/s4/active on the light
panel carry the dark hexes, copy-pasted. Values here were read off the rendered
swatches instead, with the method validated against the dark panel first (all
six samples matched their labels exactly).

**Six of the ten ink levels as drawn do not clear 4.5:1** against the surfaces
the guide itself assigns them — dark `tertiary`, marked *body text*, measured
2.91:1; light `muted` measured 1.45:1. Lifting each by the minimum that clears
the floor collapses the ramp onto one colour, so it is redistributed: the
weakest level sits on the floor and the rest step up in even increments of L —
the same perceptual-uniformity argument the brand ramp is built on. Dark `hint`
and `muted` also move off ~122° green to the ~285° the guide's own subtitle
specifies. Full before/after table in `docs/design-system/color.md`.

## Layout

Chat gains the action-cards row (240×132, per the frame), rules either side of
the day divider, and a members-count pill. The composer gains the guide's icon
row and circular send button; its formatting toolbar moves behind the `Aa`
toggle — which is what the frame's "AA" glyph is — rather than being deleted.
The cards are wired: "Create agent" opens the org chart, "Add people" opens the
members pane.

Board gains a rule under each column header, the count as a pill, a visible
drop target in place of the bare "No cards" line, and a filled "Add task".

**Deliberately not copied**, because matching literally would delete working
product: the composer keeps the "Do it once" / "Build me the workflow" chips
(the frame's Slack-style composer has no concept of agent dispatch), and the
rail keeps its 8 routes against the frame's 4.

## Verification

Driven against a real host (`test/e2e/host.sh`, harness company), signed in
through the magic-link flow, sampling computed colour through a canvas:

- Every surface and ink value matches the guide exactly across chat, tasks and
  overview in both themes — 18/18 assertions.
- Action cards measure 240×132; `Aa` toggles formatting; send is disabled empty
  and circular when enabled; "Add people" opens the pane. No console errors.
- 887/887 unit tests, 15/15 applicable e2e specs, `assert-design-tokens.sh`
  clean.

**One bug this caught:** the dark popover and `--border` both resolved to rung 4,
so the task dialog rendered its own edge invisible — exactly what the previous
translucent-border comment warned about. Fixed by moving the popover to rung 3.
Static analysis would not have found it; both tokens were individually correct.

## Follow-ups not in this PR

- `--tone-1` now sits ~7° from the brand rather than ~14°. Form still separates
  identity from interaction, but it is the first token to retune if they start
  reading alike.
- `MessageRow.tsx:102` puts `text-primary` on `hover:bg-accent` at 4.20:1. It
  measured 4.22:1 before this change, so it is pre-existing rather than a
  regression here.
- The Figma file's own defects are unfixed: the wrong light-surface captions,
  two "Ligth Theme" frame names, and no bound variables anywhere on the page.

### DIFF
--- a/docs/brand/README.md
+++ b/docs/brand/README.md
@@ -89,13 +89,17 @@ Full values, ramps and measured contrast: [`design-system/color.md`](../design-s
 
 ### The one hue we own
 
-**Indigo `#635BFF`.** It means *interactive or ours*: buttons, links, focus
+**Violet `#7153F0`.** It means *interactive or ours*: buttons, links, focus
 rings, the active nav row, the leading chart series, the logo. Nothing else.
 
-The discipline that makes it work is what indigo is **forbidden** from doing:
+A cool-warm hybrid, deliberately carrying more warmth than a generic indigo
+such as `#6366F1`.
 
-- It is never a status. A run is not indigo.
-- It is not a background wash. There are no indigo gradients, no glows, no
+The discipline that makes it work is what the brand hue is **forbidden** from
+doing:
+
+- It is never a status. A run is not violet.
+- It is not a background wash. There are no violet gradients, no glows, no
   tinted hero panels.
 - It is not used to make something look important. Hierarchy comes from size,
   weight and position first.
@@ -141,8 +145,8 @@ This is not hypothetical tidiness. Identity used to be drawn from the same
 palette as status, so a desk was tinted the exact green that means "done" and
 a category the red that means "failed".
 
-Where the two palettes do come close — identity violet against brand indigo,
-identity blue against running cyan — **form separates them**:
+Where the two palettes do come close — identity violet against the brand
+violet, identity blue against running cyan — **form separates them**:
 
 | | Shape |
 | --- | --- |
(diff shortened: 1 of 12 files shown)


graycyrus:
@coderabbitai review

Issue #1009, opened 2026-08-18
workflows: a run can hang in running forever when its finish is never journaled


graycyrus:
A node pulses indefinitely, the header reads "running", the Stop button stays on screen, and the 2s history poll never stops — until the host is restarted.

## Cause

Two ways a finish never gets journaled:

- `record_run_finished` swallows a failed append — `src/runtime/workflow_outcome.rs:174-183`.
- A panicking run task journals nothing at all — `src/runtime/run_supervisor.rs:31-35` (documented).

`list_runs` then folds start-without-finish as `running: true` (`src/server/ops/workflows.rs:2320`). Only boot-time `sweep_interrupted_runs` repairs it, so a long-lived process never recovers. On the console side the id never reaches `settledRunIds` (`frontend/src/views/WorkflowsView.tsx:1124-1131`), so `foldLiveRun`'s #921 backstop never fires.

## Compounding: the recovery poll is not visibility-gated

`WorkflowsView.tsx:1216` uses a bare `setInterval`, unlike every other poll in the console (`frontend/src/lib/visible-poll.ts`). Combined with the hang, it polls forever in a background tab.

## Proposed fix

- In `list_runs`, cross-check each `running: true` row against `RunSupervisor::live()` (`src/runtime/run_supervisor.rs:201`). Absent means dead — append the synthetic `INTERRUPTED_BY_RESTART` finish.
- Add a `JoinHandle` watchdog on `spawn_workflow_run` (`src/server/ops/workflows.rs:1248`) that journals a finish when the task panics or is aborted.
- Route the recovery poll through `startVisiblePolling`.

Issue #1008, opened 2026-08-18
workflows: failed and blocked runs persist no per-node output, so detail is missing exactly when it matters


graycyrus:
Open a failed or blocked run on the canvas, click the node that broke, and the inspector says *"No output for this node — this run predates output capture."* That is false, and it is missing exactly when the detail matters most.

## Cause

`persist_run_output` is called only on the success arm (`src/workflows/runner.rs:792`) and the clean-cancel arm (`:609`). Both failure paths return first:

- `runner.rs:568` — `return Err(map_engine_error(err))`
- `runner.rs:578` — `return Ok(blocked_run(...))`

The blocked arm is the sharper loss: it is holding a real partial `outcome.output` when it returns.

## Partial rows are zeroed at settle too

`Settled::from`'s Err arm (`src/runtime/workflow_outcome.rs:~125-140`) zeroes `deliveries`, `board`, `blocked_nodes` and `approvals`. A run that delivered two reports and *then* failed shows zero delivery rows. A failed journal append is a `warn!` only (`:181`), which also leaves the row stuck `running` forever (see the hang issue).

## Proposed fix

- Persist run output before both early returns; the `Err` arm persists whatever the engine returned or the observer collected, flagged partial.
- Carry the partial `deliveries`/`board`/`blocked_nodes`/`approvals` on the failure arm instead of zeroing them.
- Surface a `persist_run_output` failure as a run `notice` rather than a silent log line.

PR #1051, opened 2026-08-18
workflows: carry a failed run's partial rows, and fix a blocked-node regression from #1041 (#1008)

graycyrus:
> [!IMPORTANT]
> **This is not only a follow-up — it fixes a regression in #1041, which merged today.**
>
> A node refused inside its model's tool loop ends its turn by writing *prose about being blocked*, and the engine records that prose as the step's `output`. #1041's collector inserts **every** finished step's output into the capture unfiltered, so on `main` right now that apology is filed under the **blocked node's own id** and the run inspector renders it as that node's product.
>
> Issue #881 exists precisely because that apology used to travel to the *next* node as if it were the artifact. This is the same failure one surface over. This PR excludes blocked nodes' entries from the capture, on both the durable snapshot and the run body.
>
> Related, same area: `blocked_run` still returns `output: Value::Null` (`runner.rs:1080`), so the **live run drawer shows nothing** for a blocked run even though the durable snapshot now holds the data. This PR carries the same capture there too.

## What changed since this PR opened

This PR opened at 14:24 against issue #1008. **PR #1041 merged at 14:49** with an independent fix for the same issue, covering the persist-on-failed/blocked arms, the `partial` flag on `WorkflowRunOutputRecord`, the persist-failure notice, and the console badge.

Rather than leave a duplicate open, this branch has been **rebuilt from scratch on top of #1041** and narrowed to what #1041 did not do. #1041's implementation is taken wholesale — nothing already on `main` is re-landed — and its two tests are *extended* rather than duplicated. Everything below is additive to what is already on `main`.

### 1. The issue's second bullet, still unfixed on `main`

From #1008:

> `Settled::from`'s Err arm zeroes `deliveries`, `board`, `blocked_nodes` and `approvals`. A run that delivered two reports and *then* failed shows zero delivery rows.

`Settled::from` on `main` still hard-codes all five empty on the failure arm; #1041 never opened that file. The argument in its doc — that a failure "returns no `WorkflowRun` to read rows off" — was true of the signature, not of the world: every one of those rows records something the run **already did**, and all of them are durable by the time it breaks. A run that mailed two owner summaries and then died at a later node journals `deliveries: []`, which reads as "it sent nothing" while the reports sit in people's inboxes. A run that opened a board card and then failed leaves the card on the board with no run admitting to opening it.

- `record_run_finished` now takes `FailedRun { error, partial }` (with `From<&str>`, so every call site that only has a message reads as it did).
- The runner threads the partial run up on a new `OpenCompanyError::WorkflowRunFailed`, which **wraps** its cause: `Display`, `code()` and the HTTP status are byte-identical, so a caller that never asks for the partial cannot tell the wrapper apart from the error inside it. `ApiError::status` classifies `self.0.unwrapped()`.
- `pending_approvals` stays empty on purpose — a failed run is not waiting on anybody. Any gate it parked is still decidable from the Approvals page, and `approvals` is the receipt for that.

### 2. The failure arm's persist notice, which #1041 had to leave log-only

#1041's own comment says it plainly:

> Log-only on a write error (Part 3): this branch returns `Err`, so there is no `WorkflowRun` to hang a notice on.

It has one now, so a lost snapshot reaches the operator on the failure arm exactly as it does on the blocked arm.

### 3. The blocked arm — the regression fix in the callout above

See the note at the top. Both halves: the blocked node's refusal prose is excluded from the capture, and `WorkflowRun.output` carries the capture instead of `null`.

## Commands run locally

| Command | Result |
| --- | --- |
| `cargo clippy --all-targets --features openhuman,tinycortex -- -D warnings` | clean |
| `cargo test --features openhuman,tinycortex --tests` | `4324 passed; 0 failed; 3 ignored` (lib) + `10` / `1` / `11` / `2` across the four integration targets, all `0 failed` |
| `cargo check --all-targets` (default features) | clean |

The full gated suite, not a filtered subset.

## Tests

- `runtime::workflow_outcome::test::a_run_that_delivered_and_then_failed_keeps_its_rows` — the headline: two deliveries, a board card, a parked approval and a notice all survive the failure arm; `pending_approvals` stays empty.
- `runtime::workflow_outcome::test::a_failure_with_no_partial_run_journals_empty_rows` — the other half: a failure with nothing known still journals honestly empty rows (the boot-sweep shape).
- `workflows::runner::tests::a_failed_run_persists_the_partial_output_it_reached` — **extends #1041's test**: the error also carries the partial run, with the reached node's rows and the same capture that was persisted.
- `workflows::runner::tests::without_nodes_removes_only_the_blocked_entries` — unit-level, including the "leaves a value it does not recognise alone" branch that no real run reaches.
- `workflows::blocked_node_test::a_blocked_run_persists_the_partial_output_it_reached` — **extends #1041's test**: the blocked node has no entry in the snapshot, and the settled body carries the same capture rather than `null`. Both assertions fail against `main`.

`an_ordinary_run_and_a_failed_run_carry_no_notices` is renamed to `…_and_a_failure_with_nothing_known_…`, because its second claim is now conditional on the caller.

## API / behavior changes

- New `OpenCompanyError::WorkflowRunFailed { source, partial }`, plus `partial_run()` and `unwrapped()`. Additive by construction — see above.
- `record_run_finished`'s error arm takes `FailedRun` instead of `&str`.
- A blocked run's `output` is no longer `null` on the HTTP response, and excludes blocked nodes. `docs/modules/server/workflow-routes.md` is updated, since it documented the old value.

## Not in scope

The console rendering of `partial` shipped with #1041. The other issues on this surface (#1009–#1017) are their own.

Refs #1008 — the persist half is already closed by #1041; this closes the `Settled::from` half and fixes the blocked-node capture.

### DIFF
--- a/docs/modules/server/workflow-routes.md
+++ b/docs/modules/server/workflow-routes.md
@@ -219,7 +219,7 @@ step waiting on a person is not a failure — and says so structurally:
 
 ```jsonc
 {
-  "output": null,
+  "output": { "nodes": { "draft": { "items": [ { "json": { "text": "…" } } ] } } },
   "pendingApprovals": ["spec"],
   "deliveries": [],
   "nodes": [ { "nodeId": "spec", "status": "blocked", "elapsedMs": 42000 } ],
@@ -232,6 +232,17 @@ step waiting on a person is not a failure — and says so structurally:
   ]
 }
 ```
+
+`output` carries what the nodes **upstream** of the block produced (issue
+#1008). It used to be `null`, so the run drawer showed nothing for a step that
+had just written a draft.
+
+The **blocked node itself has no entry**, here or in the durable snapshot behind
+`GET …/workflows/runs/{runId}/output`. A node refused inside its model's tool
+loop ends its turn by writing prose about being blocked, and filing that prose
+as the node's product would re-open, one surface over, exactly the confusion
+issue #881 fixed by stopping it reaching the next node. The node's `blocked`
+chip and the run's notice are what say what happened.
 
 `blockedNodes` and `approvals` are omitted entirely when empty, so a run that
 blocked on nobody is byte-unchanged. Both also ride the `WorkflowRunFinished`
--- a/src/error.rs
+++ b/src/error.rs
@@ -144,6 +144,36 @@ pub enum OpenCompanyError {
         limit: usize,
     },
 
+    /// A workflow run failed after some of its nodes had already done durable
+    /// work (issue #1008).
+    ///
+    /// # Why an error variant, rather than a message
+    ///
+    /// A run's outcome is journaled by the **caller**, off what
+    /// [`WorkflowRunner::run`](crate::ports::WorkflowRunner::run) returned — and
+    /// the error arm returned nothing but a string. So a run that opened two
+    /// board cards, parked an approval and then broke at a later node journaled
+    /// a `WorkflowRunFinished` with every one of those lists empty: the cards and
+    /// the approval were sitting in front of the operator while no run admitted
+    /// to opening them.
+    ///
+    /// This carries the partial run so the caller can list what did happen. It
+    /// **wraps** the underlying failure rather than replacing it, so
+    /// [`Display`](std::fmt::Display), [`code`](Self::code) and the HTTP status
+    /// are exactly what they were — a caller that never asks for the partial
+    /// cannot tell this variant apart from the error inside it, which is the
+    /// property that keeps it additive.
+    #[error("{source}")]
+    WorkflowRunFailed {
+        /// The failure as it would have been reported before this wrapper.
+        #[source]
+        source: Box<OpenCompanyError>,
+        /// What the run had done by the time it broke: its per-node rows and
+        /// output, and the durable board / approval / notice rows its nodes
+        /// already produced.
+        partial: Box<crate::ports::WorkflowRun>,
+    },
+
     /// An operation conflicts with the company's lifecycle state (e.g. the
     /// company is paused or archived).
     #[error("company is {0}")]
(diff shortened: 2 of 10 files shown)


graycyrus:
@coderabbitai review

graycyrus:
Heads-up for reviewers: this PR opened at 14:24 and #1041 merged at 14:49 with an independent fix for the same issue. I have force-pushed this branch rebuilt on top of #1041 and narrowed it to the part #1008 asks for that #1041 did not do — `Settled::from`'s Err arm still zeroes `deliveries`/`board`/`blocked_nodes`/`approvals` on `main`, so a run that delivered two reports and then failed still journals zero delivery rows.

The diff also picks up two things that follow from having a `WorkflowRun` on the failure arm: the persist-failure notice that #1041 explicitly had to leave log-only there, and the blocked arm's `WorkflowRun.output` (still `null` on `main`, so the live run drawer shows nothing even though the durable snapshot now has the data).

One of those is a fix to what just landed rather than an addition: a blocked node's step output is the prose the model writes *after being refused inside its tool loop*, and `main` currently files that under the blocked node's id in the durable snapshot. Issue #881 exists because that apology used to travel to the next node as the artifact; showing it in the inspector as the node's product re-opens the same confusion one surface over. This excludes blocked nodes' entries from the capture on both surfaces.

@coderabbitai review

graycyrus:
@coderabbitai review

Context for the review: this branch was force-pushed at 15:38 UTC, rebuilt from scratch on top of `main` after #1041 merged with an independent fix for the same issue. The current head is `9d520df8`. Please review that diff — none of #1041's implementation is re-landed here.

Flagging for reviewer attention: the change to `src/workflows/runner.rs` that excludes blocked nodes from the per-node capture is a fix to behaviour #1041 introduced today, not a new feature. A blocked node's step output is the prose its model writes after being refused inside its tool loop, and #1041 files that under the blocked node's id.

Issue #1024, opened 2026-08-18
approvals: a parked outbound send fires on approval with no freshness check, so a days-old payload mails as if current


oxoxDev:
## Summary
A `GMAIL_SEND_EMAIL` gate sat parked for days. The moment the operator cleared an approvals backlog, it sent — mailing a weekly digest built from **2026-08-13** data under the heading **"Weekly Digest — 18 Aug 2026"**, to a real recipient.

Nothing re-validated whether the payload was still current before sending, and the approval card gave no indication the content was five days old. From the operator's side it looked like a routine send.

## Why this is not just #971
#971 is "the queue accumulates stale approvals" — an internal accuracy problem. This is the outward-facing version: the stale item **leaves the company**. A stale internal approval wastes attention; a stale outbound send puts wrong information in front of someone outside, over our name, with a current date on it.

## Mechanism
Per #243, approving does not resume a suspended call — it refuses the gated call, mints a single-use grant and re-dispatches. So the payload was composed days earlier and executed verbatim on approval. There is no revalidation step between compose-time and send-time, and no TTL on an outbound effect.

The digest's own body advertised that it was already unreliable at compose time:

> Team-output report is PARTIAL: 17 of 32 issues enumerated; PR list count-only.

and two agents disagreed on the totals in the same document (75 vs 76 items). It shipped anyway.

## Expected
At minimum, an approval card for an outbound effect should show **how old the payload is**, so an operator clearing a backlog can see they are about to mail a five-day-old report.

Better: an outbound send whose payload is older than some threshold should re-validate or refuse rather than fire, and say why.

Worth considering separately: whether a report that declares itself partial should be sendable at all without acknowledgement.

## Repro
1. Let a workflow or task compose an outbound send that gates for approval.
2. Leave it parked for several days.
3. Approve it.
4. Observe: it sends the original payload, with no age shown at approval time and no freshness check at send time.

Observed on the smoke1 tenant, 2026-08-18.

## Related
#971 (approvals never age out — the internal-facing sibling), #243 (approve re-dispatches rather than resumes), #558 (epic).

oxoxDev:
@M3gA-Mind handing this to you. Context beyond what's in the description:

**How it was found.** Not a synthetic repro — the digest actually reached a real inbox on 18 Aug carrying 13 Aug data, after I cleared a 17-item approvals backlog on smoke1 that had been accumulating for three days. I approved what looked like routine sends; one of them mailed out.

[...]

Issue #971, opened 2026-08-18
approvals: nothing ages out, so abandoned approvals accumulate in the badge forever


oxoxDev:
## Summary
Pending approvals have no expiry. An approval whose task was long since paused or abandoned stays pending, keeps its place in the queue, and keeps counting toward the sidebar badge indefinitely.

## Observed on staging
A snapshot of the queue this week, by age:

| Age | Count |
|---|---|
| ~68h | 13 |
| ~21h | 4 |
| ~30m | 2 |

The 68-hour entries belong to tasks now sitting in the **Paused** column. **16 of the 17 belonged to a single task** — one weekly-digest run that parked, was never decided, and re-asked.

Because a parked call is refused and re-dispatched rather than suspended (#243/#469), approving a three-day-old entry does not usefully resume anything — the turn it belonged to is long gone. So these are unactionable *and* permanent.

## Why it matters
The badge stops describing current state. When we needed to judge whether a just-shipped policy fix (#876) had reduced approval volume, the number was dominated by three-day-old test residue and told us nothing — the only usable signal was filtering the queue by age by hand.

An operator who sees a badge that never goes down learns to ignore it, which is the failure mode the whole approvals surface exists to avoid.

## Suggested direction
- Expire pending approvals whose originating turn is no longer live, and say so on the card ("this turn ended; approving will not resume it").
- Or scope the badge to actionable approvals and keep the rest in a separate, collapsed history.
- Either way, an approval tied to a paused or settled task should not be counted as work waiting on the operator.

## Related
- #558 (epic), #469, #243

## Environment
smoke1.staging.opencompany.work · observed 2026-08-17/18

PR #1054, opened 2026-08-18
fix(approvals): say what a parked payload's age is an age of

M3gA-Mind:
## Summary

A `GMAIL_SEND_EMAIL` gate sat parked for days. The moment an operator cleared an approvals backlog it
sent — mailing a digest built from **13 Aug** data under the heading **"Weekly Digest — 18 Aug 2026"**,
to a real recipient, over the company's name.

**The age was already on the card.** That is the finding, and it is why "add the age" would have been a
no-op with a comment claiming to fix a live bug.

It rendered as a bare `5d ago`, in muted 12px footer text, between *"Asked by Maya"* and *"Open the
card"* — routing metadata. In that position the number reads as **queue latency**: how long the
operator's backlog has held this item. What decides an outbound send is that the **payload** is five
days old. Same integer, two different facts, and only the second should stop a send.

That is exactly what the report describes: *"From the operator's side it looked like a routine send."*
The signal was not missing. It was **unlabelled**, and dressed as metadata.

So this says what the age is **of**, and gives it a weight the surrounding metadata does not have:

> Asked by Maya · Open the card · **Composed 5d ago** · declined in 3h

## Why not re-validate the payload instead

The obvious question a reviewer asks — *"why not just re-run it and recompose?"* — has an answer in the
code, and it is a decision rather than an omission.

Per #243 approving does not resume a suspended call: it refuses the gated call, mints a **single-use
grant**, and re-dispatches. The payload is frozen in `GrantedCall.args`, and `redispatch_granted_call`
(`src/harness/brain.rs`) builds the re-dispatch instruction from the **stored** arguments:

```rust
let args = serde_json::to_string(&grant.args)…;
instruction: format!(
    "Operator approved your `{tool}` call. Re-issue it now with EXACTLY these \
     arguments: {args}. Do not modify them.", …)
```

with the code's own note: *"Re-issue verbatim. The grant admits ONE call matching these arguments
exactly, so any drift the model introduces will simply re-park."*

**Two independent locks, both deliberate.** The instruction tells the model not to recompose, and
`GrantSet::consume` matches `(agent, tool, args)` by whole-value equality — so a recomposed payload
would not match the grant and would **re-park rather than send**.

Recomposing on approval would therefore authorise content the operator never saw. That is a worse
defect than staleness: a stale send is at least the thing the human read. **Showing the age is the fix
here, not a consolation prize.**

## What the age actually measures

`at_millis` is stamped by `CycleRunner::park` in the **same turn that composed the arguments**, and both
journal fold sites read it back off the record rather than re-stamping — so it survives replay instead
of resetting to boot time. It dates the payload, not the queue.

Strictly the field is *park* time; compose and park differ by the remainder of one turn, i.e. seconds.
**"Composed" is the deliberate word.** "Parked" would be closer to the field name and would reintroduce
the precise reading this fixes — parking is a queue event, and the queue reading is what let a
five-day-old digest look routine.

## Outbound only, and why the host has to say so

Labelled only for effects that leave the company. Elsewhere the age genuinely *is* queue latency, and
labelling everything would spend the emphasis where it does not matter.

**The console cannot work out which those are.** `effect_for` sets `kind` to the **tool name** and
`group` from `classify_group(tool, args)`, so the reported send arrives as
`kind: "composio_execute"` — not `email.send`. A console rule keyed on `kind` (or on the icon table)
would match native effects and **miss exactly the send this was reported for**. A classifier that misses
the reported case is worse than none.

So `ApprovalSummary` now carries the parked effect's `group`, read straight off `Effect::group` at the
single projection point. Outbound is `group != "other"` — the same line the host already draws for
`broadly_grantable`. The judgement stays in one place.

## API Or Behavior Changes

**Additive.** `ApprovalSummary` gains `group`; the console DTO types it optional, so an older host
omitting it renders exactly the pre-#1024 card rather than labelling something the console cannot
classify.

The label lands on **both** approval surfaces, because it is in the shared `ApprovalMeta` — the
Approvals page and the chat approval row.

**No TTL, deliberately.** An outbound send that silently expires is a *new* failure mode: the operator
approves, nothing happens, nobody is told. A stale send is at least visible to its recipient; a silent
non-send is visible to nobody. Refusing loudly needs a synchronous channel back to the operator who just
clicked approve, and the resolve path answers before the follow-up completes — that surface is its own
change, and half of it is worse than none.

**Rebased onto #971 (#1028), not around it.** That sibling landed mid-work and touches the same struct,
the same projection map and the same footer line. `ApprovalSummary` now carries both `expires_at_millis`
(#971) and `group` (#1024); the footer composes as `Composed 5d ago · declined in 3h` — how old the
payload is, and how long is left to decide. Worth stating because a payload-age diff adjacent to that
code should be read as a deliberate merge rather than a collision.

**Not in scope:** #971 itself (the internal-facing sibling — a stale internal approval wastes attention;
this one leaves the company), and whether a report that declares itself partial should be sendable at
all — the digest advertised *"PARTIAL: 17 of 32 issues enumerated"* and shipped anyway. That is filed
separately rather than folded in here.

## Tests

- [x] `cargo build --all-targets` — `--locked --features openhuman,tinycortex --all-targets`.
- [x] `cargo test` — `cargo test --locked --features openhuman,tinycortex --tests`:
  **4302 passed, 0 failed, 3 ignored**. Default features are not sufficient here; they skip roughly a
  third of the code.

Plus the console's own lane, since that is where this change lives: `npm run typecheck` (exit 0),
`vitest` — **944 passed across 95 files** — and `npm run build` (exit 0).

Every number above was measured on **this** tree: after dropping an unrelated commit that had landed
underneath this branch, and after rebasing onto current `main`. An earlier green on the contaminated
stack is not evidence about this one, so it is not quoted.

### Both halves verified to gate

A rendering test that asserts only the presence of a time string passes against the pre-fix card too, so
the wording is extracted into a pure `payloadAge()` and the test asserts the **string**:

* **Console** — reverting to the bare age fails with
  `expected '5d ago' to be 'Composed 5d ago'` (2 failed / 3 passed).
* **Rust** — hard-coding the summary's group to `Other` fails
  `a_parked_effect_carries_its_group_to_the_card` with `left: Other, right: Send`:
  *"the card must carry the parked effect's own group; anything constant renders an outbound send as
  internal."*

Six tests total, including one pinning that outbound-ness comes from `group` and **never** from `kind`,
and one asserting an absent `group` claims nothing.

## Documentation

No spec doc changes. The reasoning lives where the decisions do: why "Composed" and not "Parked", and
why `group` cannot be derived console-side, are recorded on `payloadAge`/`leavesTheCompany` and on the
`ApprovalSummary` field itself.

Part of tinyhumansai/opencompany#1024 — the card now shows the payload's age. The threshold behaviour
that issue also asks for is deliberately not built here; see the TTL note above.

## Summary by CodeRabbit

* **New Features**
  * Approval cards now distinguish externally directed effects with “Composed” age labels and visual emphasis.
  * Approval summaries now include consequence classifications such as spending, sending, signing, publishing, hiring, and identity-related actions.
  * Effect timestamps are clearly identified as payload composition times.

* **Bug Fixes**
  * Preserved consequence classifications for parked approvals, including approvals triggered by external integrations.

* **Tests**
  * Added coverage for payload-age labels and classification handling across supported effect types.

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -298,6 +298,10 @@ export interface ApprovalSummary {
   /** The parked effect's dotted kind, e.g. "payment.send". */
   kind: string;
   amount_usd: number | null;
+  /**
+   * Epoch-millis the effect was parked — stamped in the same turn that composed
+   * its arguments, so it dates the **payload**, not the queue (#1024).
+   */
   at_millis: number;
   /**
    * Epoch-millis this approval default-denies if nobody decides it (#971) —
@@ -314,6 +318,19 @@ export interface ApprovalSummary {
    * than guessing one.
    */
   expires_at_millis?: number | null;
+  /**
+   * The host's consequence group for the parked effect (#1024).
+   *
+   * Derived server-side from the tool **and its arguments**, so a
+   * `composio_execute` carrying `GMAIL_SEND_EMAIL` arrives as `"send"` rather
+   * than as the catch-all its tool name alone implies. It cannot be computed
+   * here: for a harness tool call `kind` is the tool name, so a console keying
+   * on `kind` would miss exactly the outbound sends this marks.
+   *
+   * Optional, and that is how an old host degrades: no field, no age label,
+   * exactly the pre-#1024 card.
+   */
+  group?: "spend" | "send" | "sign" | "publish" | "hire" | "identity" | "other";
   /**
    * Which board task this approval was parked for (#333). Mirrors `TaskLink` in
    * `src/runtime/journal.rs`.
--- a/frontend/src/components/approval-card.tsx
+++ b/frontend/src/components/approval-card.tsx
@@ -38,7 +38,7 @@ import {
 
 import type { OpenCompanyClient } from "@/api/client";
 import { GRANT_DURATIONS, type ApprovalSummary, type GrantScope } from "@/api/types";
-import { approvalAction, money, payloadLines, timeAgo, untilLabel } from "@/lib/language";
+import { approvalAction, money, payloadAge, payloadLines, untilLabel } from "@/lib/language";
 import { cn } from "@/lib/utils";
 
 const KIND_ICONS: Record<string, LucideIcon> = {
@@ -126,6 +126,8 @@ export function ApprovalMeta({
   // An id the roster does not know still beats no attribution at all — the
   // operator can at least tell two askers apart.
   const asker = a.agent ? (askerNames.get(a.agent) ?? a.agent) : null;
+  // #1024, computed once: the age, and whether this card should say it loudly.
+  const age = payloadAge(a, now);
 
   return (
     <div className="flex flex-wrap items-center gap-x-2 gap-y-1 text-xs text-muted-foreground">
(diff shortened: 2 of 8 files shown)


M3gA-Mind:
Referenced as `Part of tinyhumansai/opencompany#1024` with **no closing keyword**, deliberately: this ships the issue's "at minimum" (show the payload's age) and not its "better" (re-validate or refuse past a threshold). `closingIssuesReferences` is empty and #1024 stays open for the threshold work.

@coderabbitai review

PR #1056, opened 2026-08-18
fix(workflows): show the node a joined run is executing

### LINKED ISSUE #1010: workflows: history cannot show the node currently executing, and the live canvas loses runs on switch
A console that joins a run already in progress — a reload, a cron fire, a reconnect, or simply switching workflow and back — shows finished nodes and a dead-looking graph.

## The history fold has no `WorkflowNodeStarted` arm

`list_runs` folds `WorkflowRunStarted`, `WorkflowNodeFinished` and `WorkflowRunFinished` only (`src/server/ops/workflows.rs:2295-2440`). Because nothing records which node *started*, `statesFromRun` (`frontend/src/views/workflows/graph.ts:266`) can never produce `"running"`, and `inFlightSeed` (`frontend/src/views/WorkflowsView.tsx:1104-1116`) seeds finished nodes only.

**Fix:** fold `WorkflowNodeStarted` into a `startedNodes: Vec<String>` on `WorkflowRunOutcome` and mark started-but-unfinished as `running`.

## The canvas blanks after switching workflow away and back

`liveRanRef` (`WorkflowsView.tsx:369`) is never cleared, while `adoptedFromHistoryRef` **is** cleared on selection change (`:1240`). `inFlightSeed` then returns `null` at `:1107`; if the 300-frame window (`components/app-shell.tsx:180`) has evicted the start frame, `foldLiveRun` returns `null`, `watchingRun` goes false, and the 2s recovery poll (`:1216`) never arms.

**Fix:** replace the `liveRanRef` guard with "does the window actually hold frames for this runId", and clear both refs on company/workflow switch.

## Run frames bleed across companies

`workflowRunEvents` (`components/app-shell.tsx:384`) is never reset when `company` changes, and `foldLiveRun` matches on `workflowId`/`runId` only. Provisioned companies share manifest workflow ids, so switching company can paint the previous company's run onto an identically-named workflow.

**Fix:** `useEffect(() => setWorkflowRunEvents([]), [company])`.

graycyrus:
## Summary

A console that joins a run already in progress — a reload, a cron fire, an `EventSource` reconnect, or simply switching workflow and back — showed finished nodes and a dead-looking graph. Three separate holes met in that one symptom; this fixes all three.

**1. The history fold had no `WorkflowNodeStarted` arm.** `list_runs` folded `WorkflowRunStarted` / `WorkflowNodeFinished` / `WorkflowRunFinished` only, so the only per-node fact the journal carried was *finished*. `statesFromRun` could therefore never produce `"running"`, and a run in flight came back listing only what was already over — the graph showed a hole exactly where the work was happening.

The engine has emitted the opening bracket since #382; nothing read it. It now folds to `startedNodes: Vec<String>` on `WorkflowRunOutcome`, in start order. Started-minus-finished is the node executing right now, and the subtraction is left to the reader — the only side that knows whether it is drawing a live canvas or a settled overlay.

`startedNodes` deliberately **survives the finish**: an id there with no matching `nodes` row on a settled run is the node the run was standing on when it was cancelled or lost, which neither list says on its own. So `statesFromRun` pairs it with `running` before painting anything in flight — that guard is load-bearing, not defensive, and is what stops a settled run overlaying a spinner nothing can ever clear. Omitted when empty, so a run journaled before #382 is byte-unchanged.

**2. The seed's guard asked the wrong question.** `liveRanRef` accumulated every run this console had ever seen a frame for and was never cleared, while the frame window is a rolling 300 that evicts and the workflow/company switch cleared its sibling `adoptedFromHistoryRef` but not it. On the way back from a switch both clauses of the guard held, so `inFlightSeed` returned `null` for a run still going — and with the start frame already aged out, `foldLiveRun` had nothing left to fold and the canvas blanked.

It now asks the window itself, via `windowHasRunStart` — which is precisely the condition `foldLiveRun` branches on. Node frames alone are deliberately *not* enough: `foldFromHistory` applies them on top of the seed, so seed + surviving node frames is strictly better than either half. Both refs are now cleared together on switch.

**3. Run frames bled across companies.** `workflowRunEvents` was never reset when `company` changed, and every fold matches frames on `workflowId`/`runId` alone. Provisioned companies are built from the same manifests and routinely hold a workflow of the *same id*, so switching company painted the previous company's run onto an identically-named workflow — live node and all. Now `useEffect(() => setWorkflowRunEvents([]), [company])`.

Closes #1010

## Tests

**Backend** (`src/server/ops/workflows.rs`) — four new fold tests:
- a running run names the node it is executing (`startedNodes` = both brackets, `nodes` = the finished one only);
- a started node of a `?workflow=`-filtered-out run is dropped, not turned into a headless run;
- a run with no start rows omits the field entirely rather than sending `[]`;
- a settled run keeps the node it was standing on.

**Frontend** (`frontend/test/unit/workflow-live-node-state.test.ts`, 8 tests) — pure coverage of `statesFromRun` (start read, finish-beats-start, the settled-run guard, a host that sends no trail) and `windowHasRunStart`, plus two jsdom renders of `WorkflowsView` for the lifetime claim: joining from history alone, and keeping the paint through a switch away-and-back that evicts the start frame.

The lifetime test was checked against the old code and fails there with `expected null to be 'running'` — the blank canvas, reproduced.

**Not covered by a test:** the company reset in `app-shell.tsx`. It is a three-line `useEffect` on the shell's own state, and nothing in the unit suite mounts `AppShell` (there is no precedent harness for it — the only shell coverage is e2e). Covered by argument above rather than by a test; flagged here per the repo's untested-edge rule.

## Commands run locally

```
cargo fmt --all -- --check                                              # clean
cargo clippy --all-targets --features openhuman,tinycortex -- -D warnings  # clean
cargo test --lib server::ops                                            # 493 passed
frontend: npm run typecheck                                             # clean
frontend: npm test                                                      # 96 files, 952 tests passed
```

## API / behaviour changes

`GET …/workflows/runs` gains an optional `startedNodes: string[]` per row, omitted when empty. Additive: no existing field changes shape, and a console predating this reads exactly what it read before. Documented in `docs/modules/server/workflow-routes.md`.

## Summary by CodeRabbit

- **New Features**
  - Workflow run history now records nodes in the order they started.
  - Active workflow views identify started-but-unfinished nodes as running.
  - Started-node information remains available after interrupted or incomplete runs.

- **Bug Fixes**
  - Live workflow state stays accurate when switching companies or losing older event data.
  - Existing run records without started-node information remain supported.

- **Documentation**
  - Updated workflow API documentation to describe run-history tracking and node-state interpretation.


graycyrus:
@coderabbitai review

PR #1057, opened 2026-08-18
feat: one first-run setup — three questions, then a real company

graycyrus:
## Summary

Merges the two first-run flows into one, and turns three questions about a
business into a real company with a real team.

Before this, `SetupWizard` (pre-login) configured the **instance** — sign-in
mode, brain, host, and which of 19 templates to seed — and a separate dialog
(post-login) staffed the **company**. They never met: every shipped template
declares a roster, so the wizard handed over an already-staffed company and the
empty-roster gate never fired. Company setup was unreachable from the default
path on all 24 shipped companies.

One flow now, entirely pre-login, ending at sign-in:

| # | Step | |
|---|---|---|
| 1 | Business | what you do, what you want automated, who you want — one screen |
| 2 | You | the address that becomes the company's admin |
| 3 | Model | the credential, or skip it |
| 4 | Advanced | sign-in, brain, host, tools — present and skippable |
| 5 | Review | the team, editable, before it exists |

Then it builds, and hands you straight in.

## Two dead ends this closes

**No product template invites anybody.** Only the two e2e fixtures carry
`[users].admins`. Pick email sign-in on a laptop and — unless
`OPENCOMPANY_ADMIN_EMAIL` is in the environment — the apply succeeded and nobody
could sign in. The flow now asks for the address and writes it into the manifest.

**The hand-off forgot the operator.** They typed an address on step two and the
console then handed them an *empty* email box, waiting for a link that a host
with no mail transport was never going to send. The wizard now arranges sign-in
itself: it hands over the link directly on a loopback host, says which address it
mailed on a host that can mail, and opens the console where there is no sign-in
at all.

## How a company gets designed

`POST /api/v1/setup/roster` is the one new host route. The company-scoped twin
cannot serve the wizard — it resolves a `CompanyRuntime`, and during first-run
there is none. `RosterBuilder::for_setup` builds the pass with no company behind
it, taking the credential the operator just typed, used and discarded so the
apply stays one atomic step. Deliberately unmetered: there is no company to
charge yet, and inventing an attribution would put a row in a Usage view for a
period the company did not exist.

`manifest_from_setup` turns the reviewed roster into a `CompanyManifest`, and
`seed_generated_company` registers it. Three decisions worth review:

- `[policy].mode` reads `PROVISIONED_POLICY_MODE` rather than a literal, so a
  company this flow creates is indistinguishable from one `POST /api/v1/companies`
  provisions. That is `auto` per #605 — **not** the `supervised` the Architect
  spec asks for. Following the shipped decision rather than reopening it here.
- A designed company carries **no** `TemplateProvenance`. It did not come from a
  template, and provenance is exposed verbatim on the API — an operator reading
  `ecommerce` would reasonably conclude their company *is* that template.
- The reviewed roster is revalidated on the way in. It crossed the wire after an
  operator edited it, so bounds and de-duplication cannot be assumed to have
  survived.

## Relationship to the Architect

`docs/spec/agentic/setup.md` specs a conversational Architect and is unbuilt
(`git grep Architect -- src/` is empty). This is its first slice, arrived at
independently — including the design it already calls for: *"replaces
pick-one-of-18 as the only path while keeping templates as priors and as the
offline fallback."* Two gaps against that spec: it wants a **rationale** per
decision and full provenance; this has provenance-lite and no rationale.

Renaming this half to the Architect's vocabulary is deliberately **not** in this
PR — it is a rename worth doing on its own, and it also settles the "which
setup?" question the file collision only half-answered
(`api/setup.ts` is the instance wizard; `api/company-setup.ts` is this).

## Also here

- `starterTeam()` is deleted. It rendered twelve invented agents whenever the
  host answered with nobody — in Team, in Chat's DM list, and in the Overview
  graph — so budgets and inboxes were offered for teammates the host had never
  heard of, and DMs whose first message went nowhere.
- The roster is no longer padded from a template when a model returns a short
  one. A yoga studio was handed a Content Strategist it never asked for,
  rendered identically to the three it had. A short roster stays short; too thin
  falls back to the curated team **whole**.
- `source: "model" | "fallback"` replaces a `generated` boolean, and Review says
  which in a sentence.

## Commands run

```
cargo test --lib                                    2838 passed
RUST_MIN_STACK=16777216 cargo test --features openhuman --lib   4271 passed
cargo fmt --all -- --check                          clean
cargo clippy --all-targets -- -D warnings           clean
npm test          (frontend)                        940 passed
npm run typecheck && npm run typecheck:e2e          clean
npx playwright test                                 153 passed, 21 skipped
```

Plus the whole flow walked end to end in Brave against a live host, with the
company it produced checked on disk.

## What was theatre, and what replaced it

Reviewing the flow end to end turned up four places where it claimed more than it
did. Three were honesty problems. One was a live authority bug.

**Designed agents inherited everything.** `manifest_from_setup` builds from a
name-only manifest, so `[tools]` took the globals default
`["*", "media", "composio"]` — and an agent whose `tools` list is empty inherits
the company belt whole. Every teammate a first-run operator created held shell,
code, web, subagent, files, docs, **media** (real money) and **composio**
(per-tenant credentials), for a company described in three sentences. The globals
teammates beside them already do the opposite and say why in
`globals/agents/researcher.toml`: a request is intersected with `[tools].allow`,
so naming one can only ever narrow.

Each agent now carries an `AgentFocus` — `research` / `writing` / `operations` /
`analysis` — and the host maps it to a belt. The four lists are the globals
agents' own strings, so the vocabulary is already exercised in every company
rather than invented here. **The model names a job shape, never a tool**: grants
are a permission boundary, and letting free text a stranger typed reach `[tools]`
would put that boundary inside the prompt's blast radius. A test quantifies over
the whole enum, so a fifth focus added later cannot quietly reach `media`,
`composio`, `search`, `repo` or `shell`. Unknown resolves to `None`, which
inherits — the old behaviour, worse but never broken. Curated templates declare a
focus too: an operator with no credential must not end up with the *wider*
company.

**Nothing checked the roster covered the ask.** "Every job they mention should
have an obvious owner" was in the prompt, and a prompt is advice. The host now
splits the automation answer into a numbered checklist, each agent claims the
numbers it owns, and the host computes what is unowned by set maths over *its
own* list — if the model both listed the jobs and reported covering them, it
would be marking its own homework. Claims are collected from the roster that
survived validation, so a dropped duplicate cannot take a job's only owner with
it. A gap buys exactly one re-ask; what survives is shown to the operator rather
than hidden.

**The chips were decoration.** `inferSignals` rendered "e-commerce · physical
products" under the first question from a regex over a hand-copied duplicate of
the host's template keywords — never sent anywhere, never in the prompt, never
touching the roster, and already drifted from the host's list once. Deleted. What
replaces it splits their own words the way the host does, and Review reports what
the host actually checked.

**The spinner claimed design that never happened.** "Designing your team…" showed
on a laptop with no credential, where the curated team comes back in milliseconds
and nothing designed anything.

`tests/fixtures/setup-jobs.json` holds the splitting rule for both
implementations. Verified it binds: drifting one case fails the Rust test and the
vitest test together.

Two decisions worth a reviewer's attention:

- **`code` is not in any belt.** The plan called for it on `analysis`, but
  `grant_matches` and `grants_cover` read a glob from opposite ends and no
  shipped manifest uses `code` as a grant string, so I could not verify
  tool-level behaviour. A designed Software Engineer therefore gets the
  `operations` belt and no code execution. Narrower than today either way, and a
  company that wants it grants it.
- **`writing` and `operations` map to the same belt today.** Not an oversight —
  they differ in mandate and routing, not in tools. Keeping them distinct is what
  lets the belts diverge later without re-deciding which agents are which.

## Since the first draft

The flow was walked end to end against a real model, and eight business scenarios
were run through the design pass and judged against their inputs. That found five
defects, three of them in code I had just written.

**The model step moved to the front, and now tests itself.** It sat third, after
the questions. The design pass is silent when a credential is missing or wrong —
it falls back to a curated team — so a mistyped key produced a *plausible* company
and the operator found out two screens later, if at all.
`POST /api/v1/setup/inference/test` is a live one-turn probe built through the
same `resolve_endpoint` the runtime uses, persisting nothing. It is a gate with an
explicit escape, and on a hosted tenant it arrives already answered: `GET
/api/v1/setup` reports whether the host holds a credential, so an operator with no
key of their own never has to find one.

**An invalid tool focus was wider than any valid one.** `tools_for_focus(None)`
returned an empty list, which `agent_effective_grants` reads as *inherit the
company belt* — `["*", "media", "composio"]`. So an unrecognised focus produced an
agent holding real-money media and per-tenant credentials while every recognised
one narrowed, and the operator's free text reaches the model that writes that
string. My own comment shipped it: "degrade to the pre-focus behaviour — worse,
but never broken." Wrong default for a permission boundary. Unknown now takes the
narrowest working belt.

**A second business went unstaffed.** "A yoga studio, plus I sell mats online"
staffed the studio and nobody owned the shop, despite a "staff both" line in the
prompt. Now stated concretely; the mixed case scores coverage 5, distinctness 5.

**Three fallbacks claimed one cause.** The review screen said "we couldn't reach a
model" even where a model had answered and its answer was unusable — and pointed
the operator at adding a key when they needed to say more. `FallbackReason` now
carries `no_model` or `not_designable`.

**A bad admin address failed on the last screen.** `as` passed the email step, the
roster was designed, and the apply died with "`[users].admins` has an invalid
entry". The host's rule had no name; it is now `is_usable_admin_email`, mirrored
by the console and pinned by a shared fixture. Deliberately loose — a stricter
console rule would refuse addresses the host accepts everywhere else.

**Reference-team prose stopped leaking.** Three of six mandates in a real run were
the template's, one verbatim. The prompt now asks for the operator's words, and a
roster whose every role is the reference team's is reported as *curated* whatever
produced it. Verbatim borrowing is 0 across every designed roster since.

**Removed:** the example chips, the Brain group in Advanced (step one owns the
model now), and the read-only build inventory.

Two shared fixtures — `setup-jobs.json` and `setup-admin-email.json` — pin rules
that exist in both Rust and TypeScript. Both verified binding by deliberately
drifting them and watching both suites fail.

## Not done yet

- **No CI lane runs the setup e2e.** It needs `PW_HOST_COMPANY=companies/e2e_setup`,
  because setup opens only on an empty roster and every other shipped company has
  one. It skips cleanly without it, which is exactly the coverage trap
  `CLAUDE.md` describes for Rust integration targets.
- **Globals interact with the reviewed roster.** Since `src/globals/` landed, a
  company also receives the baseline agents, so a reviewed team of four arrives
  as seven. That is intended behaviour, but Review does not mention it.
- Phase 2 — workflows — is next, and the answers are already stored for it.


graycyrus:
@coderabbitai review

PR #1058, opened 2026-08-18
Prove the offline path with a lane that runs it

### LINKED ISSUE #579: runtime: prove and document a fully local deployment, with a CI lane that runs it
## Where this already stands

Running OpenCompany with no cloud dependency is mostly true and nowhere proven:

| Piece | State |
|---|---|
| Local inference | `ollama` and `openai_compatible` are manifest providers (`src/company/types.rs`, `INFERENCE_PROVIDERS`); a keyless Ollama sends no bearer |
| Storage | the fs store is the default; sqlite needs no server |
| Agent runtime | openhuman is embedded as a library, not a service |
| Tools | MCP servers can be local processes |
| Desktop | `src/desktop.rs` binds loopback only and ships 19 company presets |

What is **not** local: the Medulla hosted brain, Composio, and the hub identity exchange.

## The gap

No lane proves it. There is no test that boots a company with `provider = "ollama"`, an fs store and no network, and drives a task to Done — so the offline path is a claim, and it will rot the first time a cloud call is added to a shared path without anyone noticing.

## What this needs

1. **A documented offline configuration** — a worked `company.toml`, the exact environment, and what is unavailable when you run this way, in `docs/`.
2. **A CI lane** that builds with the offline feature set and runs an end-to-end task with no network egress. Given #475 and #477 — where feature-gated suites were type-checked and never executed — this lane must assert a non-zero test count, the way `scripts/ci/assert-integration-targets-run.sh` does.
3. **An honest capability table**: which surfaces degrade, and how they say so. Composio-backed tools should be absent rather than present-and-failing.

## Why it earns a place

Two reasons beyond privacy. It is the foundation for the local server in the multi-server picture, and it is the cheapest continuous proof that the runtime has no hidden dependency on a service somebody has to pay for.

## Acceptance

- A documented config boots a company, runs a task end to end, and produces a deliverable with no network egress.
- CI runs that path on every change and fails if the test count is zero.
- The docs state plainly what does not work offline.

Post-MVP, P2.

M3gA-Mind:
## Summary

Running OpenCompany with no cloud dependency was mostly true and nowhere proven. This adds the proof
first and the document second, because a documented offline path that nothing executes **rots the
first time a cloud call is added to a shared code path — and rots silently**, since nothing fails and
the page keeps asserting with authority.

`tests/offline_e2e.rs` boots a company from the documented manifest (`provider = "ollama"`, fs store),
signs in over the loopback magic-link flow, creates a card, drags it into In Progress, and an agent
works it. Verified against a real run:

```
timeline = [dispatched, "Reply from writer", "Finished on writer → in_review"]
script calls = 1
```

Every model call goes to a scripted OpenAI-compatible endpoint on loopback, and the test asserts that
endpoint **was actually called** — so a run that never reached a model cannot pass as one that did.

## How "no network" is enforced, and why that is the whole point

CI runs the compiled binary under:

```sh
sudo unshare --net -- sh -c "ip link set lo up && exec <binary>"
```

A fresh network namespace has no routes, no DNS and no interfaces beyond the loopback that brings up.
An accidental cloud call therefore **fails** rather than passing unnoticed. A lane that merely declines
to configure cloud providers proves nothing: it passes just as happily when something dials out on a
code path the test did not reach.

Compiled **before** the namespace is entered, deliberately — compiling fetches crates, so sandboxing
the build would fail on the registry and say nothing about the runtime.

Alternatives rejected, with reasons, so nobody re-litigates them: `iptables -P OUTPUT DROP` would also
cut the Actions runner's own control connection to the service; a container with `--network none`
needs an image built for the purpose and buys no extra isolation over a namespace, with the binary
already on disk.

### The sentinel — proof versus theatre

`a_deliberate_outbound_call_fails_inside_the_namespace` attempts a real outbound connection and
asserts it **fails**. If it succeeds:

> egress SUCCEEDED inside the offline lane, so the network namespace was not applied and every other
> assertion in this file proves nothing about being offline

**If `unshare` misbehaves on `ubuntu-latest`, the lane goes red on its own sentinel rather than passing
green.** That is the difference between proof and theatre. Without it, a sandbox that silently failed
to apply produces a run indistinguishable from a working one, and "nothing dialled out" cannot be told
from "the jail was never locked".

**Do not delete this test to make a red build go away** — a red here means the isolation is not real,
which invalidates every other assertion in the file.

It asserts only when the lane sets `OPENCOMPANY_OFFLINE_LANE=1`, so the same file still runs on a
developer's networked laptop without failing for having a network.

### What I could and could not verify locally

Stating this plainly rather than leaving it implied.

**I could not run `unshare` on macOS**, so the sandbox mechanism itself is unverified until this runs
in CI. What I *did* verify locally is the part that decides whether the lane is meaningful: running
the binary with `OPENCOMPANY_OFFLINE_LANE=1` **and a real network present** — exactly what an
unapplied namespace looks like — fails with the message above. So the failure mode that would make
this lane worthless is caught, and if the mechanics are wrong the first CI run says so instead of
going quietly green.

## Vacuity: the #475 / #477 trap

The target carries `#![cfg(feature = "openhuman")]`, so it reports **0 tests at default features and 2
in the gated lane**. That is precisely the pathology `scripts/ci/assert-integration-targets-run.sh`
exists to catch, and it is satisfied rather than worked around:

```
  offline_e2e: 2 tests
Every integration target executed at least one test.
```

## A footgun found on the way — filed separately, not fixed here

`RuntimeBuilder` does not attach the agent pool by default. Without `.with_harness(...)`,
`CompanyRuntime::dispatch_task` is a **documented no-op**: a card dragged into In Progress simply sits
there, no dispatch, no error, no log. `serve` wires it (`src/bin/opencompany.rs:589`); nothing warns
anyone who does not.

That cost real time here — the first version of this test had a card inert in In Progress with an
empty timeline and zero inference calls — and it will catch the next person writing a test or a tool
against the builder. Filed as tinyhumansai/opencompany#1059 rather than widened into this PR.

## Judgement calls, all stated in the doc as well

**A scripted endpoint rather than a real Ollama.** `provider = "ollama"` resolves through the same
code as any OpenAI-compatible endpoint — one `base_url`, no bearer when keyless (already pinned by
`request_plan_omits_bearer_for_keyless_ollama`), same request shape — and the URL is overridable. So
the script exercises the **real** `ollama` provider path while keeping the lane fast and
deterministic. What it does **not** prove is that a real Ollama server is wire-compatible with this
client; that needs a model pull, which is slow and fails for reasons unrelated to this repository.

**Why the operator finishes the card.** #579 asks for a card driven "to Done". Since the operator
decision of **2026-08-05** (recorded in `harness::lifecycle`), `done` is reached **only by a person**
through an approving verdict — every card settles in `in_review` first. So the agent half drives the
card as far as an agent can take it, offline and unaided, and the harness then plays the operator who
approves it, also offline.

That is the claim stated accurately, not weakened. A version reporting `in_review` as "Done" would
read as proof that an agent finishes work alone, which the product does not permit. Noted on #579,
whose acceptance predates the decision and cannot be met literally.

## Out of scope by design

**Medulla, Composio and the hub identity exchange are hosted services and stay hosted.** The doc leads
with that as a table. A green lane says **the offline path works** — never that everything works
offline. With the documented manifest those surfaces are *absent* rather than present-and-failing: no
`composio` grant means those tools are never wired onto a belt, no managed credential means the
cognition brain is not the hosted one, and nothing contacts the hub.

## API Or Behavior Changes

None. This adds a test target, a CI step and documentation; no runtime code changes.

## Tests

- [x] `cargo build --all-targets` — covered by the clippy and test runs above.
- [x] `cargo test` — **4282 passed / 0 failed** at `--features openhuman,tinycortex --lib`; the new
      target is **2 passed** and `assert-integration-targets-run.sh openhuman,tinycortex` reports
      `offline_e2e: 2 tests`.

Also: `scripts/ci/assert-md-line-cap.sh` PASS.

### Proven to fail when the thing it guards is broken

- **The sentinel**: run with `OPENCOMPANY_OFFLINE_LANE=1` and a network present (an unapplied
  namespace) → fails with the message quoted above.
- **The end-to-end test**: without `.with_harness(...)` the card never leaves In Progress and the test
  fails on `card never reached ["in_review", ...]; it is sitting in 'in_progress'` — which is how the
  footgun above was found.
- **Vacuity**: dropping the crate-level `cfg` would make the target report 0 in the gated lane and
  `assert-integration-targets-run.sh` fails on the count.

**Not verified locally, and cannot be**: the `unshare` sandbox itself — see the section above. The
first CI run is its first execution.

## Documentation

`docs/spec/runtime/offline.md` — the configuration, what is not local, what the lane does, how the
isolation is enforced, why the sentinel exists, and what to do when a new cloud call turns it red
("do not make the lane pass by giving the namespace a route"). Indexed in `docs/spec/README.md` and
`docs/spec/runtime/README.md`.

Closes tinyhumansai/opencompany#579

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -764,6 +764,47 @@ jobs:
       - name: Assert every integration target actually runs
         run: scripts/ci/assert-integration-targets-run.sh openhuman,tinycortex
 
+      # Issue #579: the offline claim, executed rather than asserted.
+      #
+      # `sudo unshare -n` puts the test binary in a fresh network namespace: no
+      # routes, no DNS, no interfaces but the loopback this brings up. An
+      # accidental cloud call on a shared code path is therefore a RED BUILD
+      # here, which is the whole point — a lane that merely declines to
+      # configure cloud providers passes just as happily when something dials
+      # out on a path the test did not reach.
+      #
+      # Why not the alternatives. `iptables -P OUTPUT DROP` would also cut the
+      # Actions runner's own control connection to the service, and a container
+      # with `--network none` needs an image built for the purpose; the
+      # namespace needs neither and the binary is already on disk.
+      #
+      # BUILT FIRST, DELIBERATELY. `--no-run` compiles (and therefore fetches)
+      # while the network still exists; only the execution is sandboxed. A
+      # namespace around `cargo test` would fail on the registry, not on the
+      # code, and would say nothing about the runtime.
+      #
+      # `OPENCOMPANY_OFFLINE_LANE=1` is what arms the guard's own self-check —
+      # `a_deliberate_outbound_call_fails_inside_the_namespace` asserts that
+      # egress FAILS. Without it a sandbox that silently failed to apply would
+      # produce a green lane indistinguishable from a working one, and "nothing
+      # dialled out" could not be told from "the jail was never locked".
+      - name: Build the offline end-to-end target
+        run: cargo test --locked --features openhuman,tinycortex --test offline_e2e --no-run
+
+      - name: Run it with no network at all
+        run: |
+          set -eu
+          binary="$(cargo test --locked --features openhuman,tinycortex --test offline_e2e \
+            --no-run --message-format=json 2>/dev/null \
+            | jq -r 'select(.executable != null and (.target.name == "offline_e2e")) | .executable' \
+            | tail -n 1)"
+          test -n "$binary" || { echo "could not locate the offline_e2e test binary" >&2; exit 1; }
+          echo "running $binary inside a network namespace"
+          sudo --preserve-env=RUST_MIN_STACK,OPENCOMPANY_OFFLINE_LANE \
+            unshare --net -- sh -c "ip link set lo up && exec '$binary' --test-threads=1 --nocapture"
+        env:
+          OPENCOMPANY_OFFLINE_LANE: "1"
+
       # Catches the combination traps no single feature set can: most often a
       # mandatory struct field constructed differently across feature arms.
       #
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -107,6 +107,7 @@ L0  Substrate       api.tinyhumans.ai, openhuman-core, tiny.place, filesystem
 | [runtime/repos.md](runtime/repos.md) | Bound repositories: mirror cache, credential handling, quota |
 | [runtime/data-root.md](runtime/data-root.md) | Data-root resolution, the single-writer lock, instance identity |
 | [runtime/desktop.md](runtime/desktop.md) | The desktop client: connections, transport seam, embedded host |
+| [runtime/offline.md](runtime/offline.md) | Running with no network: the configuration, what is not local, and the CI lane that proves it |
 | [runtime/hub-console.md](runtime/hub-console.md) | One console deployment operating many hosts on other origins |
 | [security/agent-isolation.md](security/agent-isolation.md) | What confines an agent and what does not — enforced controls, the gaps, and the capability that survives every planned control |
 | [company-as-agent/README.md](company-as-agent/README.md) | Companies as economy citizens |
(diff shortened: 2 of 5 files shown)


M3gA-Mind:
@coderabbitai review

PR #1060, opened 2026-08-18
fix(chat): give host-authored notices an author that is honestly not an agent

M3gA-Mind:
> ## ⚠️ Do not land before #1040
>
> This is stacked on #1040 and **cannot compile on `main` today**. `is_known_author` is
> created by that PR (commit `b1cea77b`); this one only adds the `SYSTEM_AUTHOR` arm to it.
> Merging this first fails on a missing symbol. The base branch is
> `fix/966-confined-turn-author` for that reason — GitHub retargets to `main` automatically
> once #1040 merges.

## Summary

Three sites emit prose the **runtime** wrote itself:

| site | what it says |
|---|---|
| `src/harness/brain.rs` | the approval-overflow notice |
| `src/harness/brain.rs` | the `"Acknowledged."` cycle fallback |
| `src/company/runtime.rs` | `announce_continuation_failure`'s report |

Each landed as an ordinary `AgentReply` carrying `"operator"` in the author field — making a
**correct system row byte-identical on disk** to a reply whose author the pre-#885 defect had
overwritten.

## Why this had to be a writer-side fix

The plan on tinyhumansai/opencompany#966 was to caption unattributable rows with *"we do not
know who spoke"*. That is true of a damaged reply and **false** of a system notice, whose
author is known and correct. So the marker needed a way to tell them apart. There isn't one.

`CompanyEvent::AgentReply` stores exactly these fields, and nothing else about a row is
available to a reader:

| field | a system notice | a reply whose author was overwritten | discriminates? |
|---|---|---|---|
| `chat_id` | the desk | the desk | no |
| `agent_id` | `"operator"` | `"operator"` | **no — byte-identical** |
| `steps` | `Vec::new()` | *also* empty for a memory-served or tool-less turn — `TurnOutcome::steps` calls this "the zero-steps tell" | no |
| `task_id` | `None` | `None` for an ordinary chat reply | no |
| `parent` | written by `journal_chat_replies` | written by the same call, same value | no |
| `message_seq` | a log position | a log position | no |
| `text` | host-authored, fixed | model output | only this — and it is prose |

Message text is the single separator, and keying a data-integrity treatment on host-authored
prose is too brittle to ship: the strings are free to be reworded, and a reworded notice would
silently start reading as an unattributable agent reply.

**So a reader-side marker is not buildable against current data**, and any marker applied
today would caption three correct rows as unknown — teaching operators to distrust legitimate
system messages, which is strictly worse than the bug it set out to fix. The finding is
recorded in full on the issue so nobody re-attempts it.

The treatment therefore goes where #1040's did: to the **producer**, teaching it to record its
true author rather than teaching a reader to guess.

## The change

`SYSTEM_AUTHOR` is a reserved id in the ungated `ports::ids`, carried on the
`OutboundMessage.agent` field #885 added. **No new event variant, no new field, no wire
change.**

Its value matches the literal `MessageView::project` already uses for the `DeskTaskCompleted`
marker, so it reaches the console's existing system rendering with **no reader-side change at
all** — and the projection's copies of that literal now point at the constant instead of
repeating it, so the two cannot drift.

`chat_history::is_known_author` accepts it. Without that, the attribution audit from #965
would count three *correct* rows as damage and inflate the one figure there that has to be
trustworthy.

`system_notice` and `continuation_failure_notice` name the two constructions. That is not
decoration — see the test note below.

## What this buys, and what it does not

**Buys:** every notice journaled from here is distinguishable from a genuine reply by a
**stored field** rather than by prose. A future marker for unrecoverable authors then has a
clean set to apply to and cannot slander a system row.

**Does not buy:** it is **forward-only**. Every notice already in a journal keeps `"operator"`
and stays byte-identical to a damaged reply, permanently. Rows from before this remain a mixed
bucket that **no marker can honestly caption** — the same conclusion #885 reached about
unrecoverable authors, for the same reason: the distinguishing information was never written
down, and nothing recovers it after the fact.

## API Or Behavior Changes

**No wire change.** No route, no HTTP/GraphQL type, no new field. `Cargo.lock` untouched.

**Behaviour — one visible change, called out rather than buried.** Those three notices move
from rendering as a company bubble to rendering as a centred **system pill**. That is arguably
what they always should have been (they are the runtime speaking, not an agent), but it is a
visible change to a shipped surface, not a silent correctness fix.

**Audit.** The attribution count no longer treats a host-authored notice as an affected row.
Rows journaled before this are unaffected — they store `"operator"`, not the reserved id.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — as CI runs it, `--locked --no-deps`, both
      the default lane and `--features openhuman,tinycortex`. Both clean.
- [x] `cargo build --all-targets` — run as `cargo check --locked --all-targets`. (`check`, not
      `build`: nothing here is a linker-visible change.)
- [x] `cargo test` — `cargo test --locked --features openhuman,tinycortex --lib`:
      **4285 passed, 0 failed, 3 ignored** on the branch.

### Verified against the merged tree, not just the branch

A green branch does not prove `main` still builds — this repo has had two individually clean
merges leave `main` not compiling. `main` also moved to `9881aea7` mid-review with commits
touching **both** `src/company/runtime.rs` and `src/harness/brain.rs`, the two files this
edits.

So the stack was merged into current `main` on a scratch branch and rebuilt there:
**`cargo check` clean, 4308 passed, 0 failed.** The 4308-vs-4285 gap is `main`'s own newer
tests arriving in the merge and passing alongside these.

### Every new test was proven to fail with its fix reverted — 4 of 4

|---|---|
| `is_known_author` drops the `SYSTEM_AUTHOR` arm | `a_host_authored_notice_is_a_known_author_not_an_affected_row`, `a_notice_and_an_overwritten_reply_are_no_longer_the_same_author` |
| all three writer sites stop naming a system author | `a_host_authored_notice_is_not_authored_by_the_operator_channel`, `a_failed_continuation_report_is_authored_by_the_runtime_not_the_operator` |

**The first version of these tests proved nothing about the writers, and that is why the two
constructions are named functions.** The initial pair covered only whether the audit
*recognised* the reserved id; reverting all three writer sites failed **nothing**. That is the
same gap that let #885 exist — nothing tested what the write path stores — so
`system_notice` and `continuation_failure_notice` exist to make the author assertable without
standing up a cycle.

Both writer tests assert the author is **not** `"operator"`, not merely that it equals the
constant: the two values being conflated is the entire defect, and an equality-only assertion
would still pass if `SYSTEM_AUTHOR` were ever redefined to the channel name.

- **No test drives a cycle end to end.** The constructions are asserted directly; that they
  are *called* at the three sites is guarded by the compiler and by reading, not by a test.
- **The premise was verified from source, not from a live tenant.** `GET {scope}/chat/history`
  on staging returns `401` and no operator token was available, so the misattribution is
  established by reading the writers rather than by observing a real transcript. This is the
  weakest evidence in the change and is stated rather than glossed.
- **No count against real data.** #965's audit still has no measured blast radius; that needs
  a deployed host, as recorded there.

## Documentation

No spec doc changes. The reasoning lives at the code it governs: `SYSTEM_AUTHOR` carries why it
exists and that it is forward-only, `is_known_author` carries the false positive it prevents,
and both constructions carry why they are named functions.

Refs tinyhumansai/opencompany#966 — the presentation half of that issue remains open; this
records the finding that no reader-side marker is buildable and fixes the producers instead.


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Rebased onto `main` (`02743fda`). CI builds the merge, and `de533ba2 feat(metering): persist planning attempt usage` added a required `planning_attempts` field to `TaskRecord`, so branches predating it stopped compiling against current main.

This branch constructs no `TaskRecord`, so it needed no code change — the rebase alone clears it. Nothing else touched.

Verified on the rebased tip: `cargo test --locked --features openhuman,tinycortex --tests` → **4390 passed, 0 failed**.

Issue #1047, opened 2026-08-18
metering: inference spend is posted positive, so Finances reports every model cost as revenue


graycyrus:
Every model charge is posted to the ledger as a **positive** `amount_usd`, and the Finances projection reads a positive amount as **income**. So a company that has only ever spent money reports growing revenue and a growing balance, while `spentUsd` and `byCategory` stay at zero.

## The two halves disagree

`src/metering/finances.rs:10-11` states the convention:

> Sign convention (set by the economy adapter and cost hook): outflows are negative `amount_usd`, inflows positive.

and `finances_from` implements exactly that (`:44-52`): `amount_usd < 0` accumulates into `spent_usd` and `by_category_map`; `amount_usd > 0` accumulates into `revenue_usd`.

But the inference writer posts the cost unnegated — `src/metering/inference.rs:68-79`:

```rust
Some(LedgerEntry {
    at_millis: now_millis(),
    kind: INFERENCE_SPEND_KIND.to_string(),
    amount_usd: usage.cost_usd,   // positive
    memo: agent.to_string(),
})
```

and `src/harness/cost.rs:214` asserts that value is positive.

## Verified at every call site

`inference_ledger_entry` is the single constructor for `inference.spend`, and every path appends its result without negating:

- `src/harness/cost.rs:79` → `:117` (`store.append_ledger`)
- `src/metering/planning.rs:118`
- `src/metering/triage.rs:87`
- `src/metering/workflow_build.rs:120`

`grep -rn "amount_usd: *-" src --include='*.rs'` returns three hits, and only one is production code:

- `src/economy/adapter.rs:153` — the x402 outflow, correctly negated
- `src/server/ops/finances.rs:181` and `src/server/graphql/test.rs:1037` — test fixtures, both using `-12.0` / `-2.0`

So the fixtures encode the documented negative convention while the live writers contradict it, which is why no test catches this.

## Impact

Every operator-facing money surface built on `finances_from` is wrong in the same direction: `GET /api/v1/company/finances` and its GraphQL equivalent report inference spend as revenue, inflating `revenueUsd` and `balanceUsd` and leaving `spentUsd` and the `byCategory` breakdown empty. The `[budget].monthly_usd` comparison built on the same projection is therefore also wrong.

Note this does **not** affect the per-agent daily budget gate, which reads the usage meter rather than this projection.

## Proposed fix

[...]

PR #1061, opened 2026-08-18
fix(metering): post inference spend as a negative amount (#1047)

M3gA-Mind:
## Summary

The ledger's sign convention is stated in `src/metering/finances.rs:10-11` — outflows negative, inflows positive — and `finances_from` implements exactly that. The inference writer posted `usage.cost_usd` **unnegated**, so the projection read every model charge as income.

A company that has only ever spent money therefore reports **growing `revenueUsd` and a growing `balanceUsd`**, while `spentUsd` and the `byCategory` breakdown stay at zero.

Fixed at `inference_ledger_entry` — the single constructor all four writers go through (`harness::cost`, `metering::planning`, `metering::triage`, `metering::workflow_build`).

**The reader was deliberately not touched.** The x402 outflow in `economy::adapter:153` already posts negative, and both server fixtures (`server/ops/finances.rs:181`, `server/graphql/test.rs:1037`) already encode the negative convention. Inverting `finances_from` would have meant changing all three to match a convention the docs already state correctly.

## It is a one-character fix with a nine-site audit

Worth stating plainly, because the diff is larger than a reviewer expecting one line will predict:

**Grep found five assertions. The compiler found four more. Nine total.**

My initial sweep for `INFERENCE_SPEND_KIND` / `"inference.spend"` missed `cost.rs:239`, `cost.rs:286`, `inference.rs:417` and `inference.rs:437` — each asserts the amount inline without naming the kind, so no text search for the kind would surface them. That is the argument for auditing every consumer rather than negating and moving on.

Each site was decided on its own merits — signed ledger value, or absolute magnitude:

| Consumer | Wants | Outcome |
| --- | --- | --- |
| `finances_from` spent/revenue split | signed | the target of the fix — now correct |
| `bookkeeping_net` → `balance_usd` | signed | was inflating the balance; now decreases correctly |
| `Transaction.amount_usd` | **magnitude** | already `.abs()`, sign carried separately — see below |
| per-agent daily budget gate | — | **unaffected**: reads the usage meter, not this projection (verified, not assumed) |
| 9 test assertions | — | corrected to negative |

### The display half needs no change, and is now pinned

This is the part a sign fix could plausibly break, so it was checked rather than assumed.

`finances.rs:85-90` builds each `Transaction` with `amount_usd: e.amount_usd.abs()` and expresses the sign as a separate `direction: Direction::Out | In`. **The display never sees the sign.** After this change a charge flips `In` → `Out` carrying the same positive figure, which is the correct rendering — it does not start showing `-$0.42`.

`a_charge_renders_as_an_outgoing_transaction_with_a_positive_amount` pins both halves of that.

## Forward-only — and what that leaves behind

**There is no migration, deliberately.** An `inference.spend` ledger is an append-only financial record; rewriting historical rows to correct a sign is not a fix but a falsification — anyone who exported or reconciled against those figures would silently disagree with the system and have no way to tell why.

So every row already written keeps its positive amount and will keep reading as revenue. What that means for a company that has run costed cycles, stated precisely:

- **The revenue and category distortion ages out.** `finances_from` filters `spent_usd` / `revenue_usd` / `by_category` to the **current month**, so at the next month boundary those three surfaces become correct on their own.
- **The balance distortion is permanent.** `balance_usd` is `bookkeeping_net`, summed across **all time** with no month filter. Every pre-fix charge stays in it as a positive contribution, so the reported balance remains inflated by the total of all inference spend posted before this lands — indefinitely.

That distinction is the one someone reconciling a company's books needs, and it is why "it fixes itself next month" is only two-thirds true.

The accountancy-correct remedy for a wrong posting is not an edit but a **compensating entry** that names itself as a correction, leaving both the original and the adjustment visible. That is a product and finance decision rather than something to settle inside a sign fix, so it is filed as a follow-up rather than folded in here — see the linked issue.

*(A company with an economy wallet is unaffected on this axis: `economy_balance` overrides `bookkeeping_net` when present.)*

## API Or Behavior Changes

- `inference.spend` ledger entries are now written with a **negative** `amount_usd`. Same kind, same memo, same gating on non-zero cost.
- `GET /api/v1/company/finances` and the GraphQL equivalent now report inference spend under `spentUsd` and `byCategory` instead of `revenueUsd`; `balanceUsd` decreases with spending instead of increasing.
- The `[budget].monthly_usd` comparison, built on the same projection, becomes correct as a consequence.
- **Unchanged:** the per-agent daily budget gate (reads the usage meter), the Usage surface, the rendered `Transaction` figure, and every pre-existing ledger row.

## Tests

Gated lane — default features skip this code. No prettier run; this change is Rust only.

- [x] `cargo clippy --no-deps -p opencompany --features openhuman,tinycortex --all-targets -- -D warnings` — mirrors CI's `--no-deps`; no findings.
- [x] `cargo check -p opencompany --features openhuman,tinycortex --tests` — clean, via the test run below.
- [x] `cargo test -p opencompany --features openhuman,tinycortex` — **4303 passed, 0 failed, 3 ignored.**

### Both new tests fail with the fix reverted — on direction, not magnitude

A sign test that passes with the sign flipped either way is worthless, so both assert the **direction**. With `amount_usd: -usage.cost_usd` reverted to `usage.cost_usd`:

| Test | Fails with |
| --- | --- |
| `a_costed_cycle_is_spending_not_revenue` | `inference is an outflow, so the ledger amount is negative: 0.42` |
| `a_charge_renders_as_an_outgoing_transaction_with_a_positive_amount` | `a charge is money going out` |

Neither failure is a magnitude mismatch. The first also asserts `revenue_usd == 0.0` exactly and `balance_usd < 0.0` — a company that only spent money must have **no** revenue — and runs the reader over the entry the writer actually produced, so it pins the writer/reader pairing that was the whole defect. Each half was self-consistent; only their combination was wrong, which is why no existing test caught this.

### One flake, investigated rather than waved off

A full-suite run failed on `server::operator::test::a_failed_continuation_tells_the_operator`. It passed 3/3 in isolation, and the **same tree** then passed the full suite 4303/0 in **31s** against **154s** for the failing run. Same code, different result under machine contention — a load-sensitive flake in an async continuation test, unrelated to ledger amounts.

## Documentation

None needed. `metering::finances` already documents the convention correctly — this makes the writer obey it. The reasoning is recorded on `inference_ledger_entry` itself, including why the fix is at the writer rather than the reader.

Refs tinyhumansai/opencompany#1047

### DIFF
--- a/src/brain/hosted/test.rs
+++ b/src/brain/hosted/test.rs
@@ -679,7 +679,7 @@ async fn e2e_reported_usage_lands_on_the_usage_meter() {
         record
             .ledger
             .iter()
-            .any(|e| e.kind == crate::metering::INFERENCE_SPEND_KIND && e.amount_usd == 0.042)
+            .any(|e| e.kind == crate::metering::INFERENCE_SPEND_KIND && e.amount_usd == -0.042)
     );
 }
 
--- a/src/harness/cost.rs
+++ b/src/harness/cost.rs
@@ -212,7 +212,8 @@ mod tests {
         let turn = turn_with(0.42);
         let entry = ledger_entry_for(&turn, "ceo").unwrap();
         assert_eq!(entry.kind, "inference.spend");
-        assert_eq!(entry.amount_usd, 0.42);
+        // Negative: an outflow, per the ledger convention (issue #1047).
+        assert_eq!(entry.amount_usd, -0.42);
         assert_eq!(entry.memo, "ceo");
     }
 
@@ -235,7 +236,8 @@ mod tests {
 
         let ledger = store.ledger.lock().unwrap();
         assert_eq!(ledger.len(), 1);
-        assert_eq!(ledger[0].amount_usd, 1.5);
+        // Negative: an outflow, per the ledger convention (issue #1047).
+        assert_eq!(ledger[0].amount_usd, -1.5);
         let samples = meter.samples.lock().unwrap();
         assert_eq!(samples.len(), 1);
         assert_eq!(samples[0].kind, SampleKind::Inference);
@@ -282,7 +284,7 @@ mod tests {
         // Attribution only — the spend itself is unchanged.
         let ledger = store.ledger.lock().unwrap();
         assert_eq!(ledger.len(), 2);
-        assert!(ledger.iter().all(|e| e.amount_usd == 0.5));
+        assert!(ledger.iter().all(|e| e.amount_usd == -0.5));
 
         // The field is additive on the wire: an unattributed sample serializes
         // exactly as it did before #242.
--- a/src/metering/inference.rs
+++ b/src/metering/inference.rs
@@ -65,6 +65,25 @@ pub const INFERENCE_SPEND_KIND: &str = "inference.spend";
 /// tokens but bills backend-side and echoes no USD, so a token-bearing
 /// zero-cost cycle must not post a meaningless `$0.00` spend line to Finances.
 /// Its tokens still land on the Usage surface through [`inference_sample`].
+///
+/// # Sign (issue #1047)
+///
+/// The amount is posted **negative**. Inference is an outflow, and the ledger's
+/// convention — stated in [`finances`](super::finances) and implemented by
+/// `finances_from` — is that outflows are negative and inflows positive. This
+/// posted `usage.cost_usd` unnegated, so the projection read every model charge
+/// as income: `revenueUsd` and `balanceUsd` grew with spending while `spentUsd`
+/// and the `byCategory` breakdown stayed empty.
+///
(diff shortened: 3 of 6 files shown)


M3gA-Mind:
The forward-only residue described under **Forward-only — and what that leaves behind** is filed as #1063 (assigned to @oxoxDev), stating the two options plainly: accept the permanent balance inflation, or post a dated compensating entry per company.

Deliberately not folded in here — the accountancy-correct remedy for a wrong posting is a correction entry that names itself as one, not an edit to the original row, and choosing between those is a product and finance decision rather than a detail of a sign fix.

@coderabbitai review

Issue #887, opened 2026-08-15
runtime: workspace_read's real error is discarded and rendered as "Something went wrong with this action."


oxoxDev:
## What happens

In one observed live turn the console step list showed 10 steps, **3 failed**. Each failure was a `Workspace Read` against a path under `Agents/<agent>/<id>/…`, rendering only as:

```
Workspace Read  Failed  — path=Agents/frontend_engineer/019ff5e11b51-0000000000…
51ms   Something went wrong with this action.
```

Sibling reads of the same shape, and a `Workspace List` over the same prefix, succeeded in the same turn.

## Root cause of the *visible* symptom (certain)

`workspace_read` writes a specific sentence for each of its five failure exits, and the step renderer throws it away. `failure_result` (`src/harness/steps.rs:685`) surfaces a tool's own message only for members of `INTRINSIC_TOOLS` (`:485-508`) — no `workspace_*` tool is in that list, so the message is replaced by `classified.cause_plain`, which for the unmatched `Unknown` class is literally `"Something went wrong with this action."` (`vendor/openhuman/src/openhuman/tools/status/ops.rs:238-241`). `steps.rs:609-611` already names this exact string as a known gap.

## Root cause of the *underlying* failure (not yet knowable — and that is the point)

It cannot be diagnosed from the evidence precisely because the message is discarded. Two candidates survive, and the storage backend selects between them:

- **fs** (`StorageKind::Fs` is `#[default]`): node content is written with bare `tokio::fs::write` (`src/store/fs_ops.rs:1328`, `:1384`) while the index is written atomically (`:1802`), and `read` (`:1276`) takes no lock. A reader inside the `O_TRUNC` window gets `read_to_string` → `InvalidData`. **This explains the observed asymmetry**: `workspace_list` touches the index only and succeeds; `workspace_read` touches index + content and fails.
- **sqlite/mongodb**: no torn-read window, but duplicate sibling nodes are possible — sqlite has no sibling-name guard (`src/store/sqlite.rs:224-231`) and `workspace_create` is check-then-act (`workspace_tools.rs:1750-1761`). One duplicated ancestor folder poisons every descendant path via `render_path`, yielding `ResolveError::Ambiguous` on read while `entries_under` still lists them. The existence of `src/company/workspace_repair.rs` ("fold the duplicate sibling folders a publish race already left behind") is evidence this condition is real in production.

## Suggested direction

[...]

PR #1062, opened 2026-08-18
fix(store): make write_atomic durable, not only atomic

### LINKED ISSUE #1049: store: write_atomic never fsyncs, so a rename-ordered save can still be lost on power failure
`write_atomic` (`src/store/fs.rs:565-601`) does `create_dir_all` → `tokio::fs::write(tmp)` → `rename`, with no `sync_data()` on the temp file and no `sync_all()` on the parent directory.

Its own doc block (`:571-587`) claims only that a reader never sees a torn file, and it delivers that. But rename is not a durability barrier: on ext4 or APFS, power loss between the `rename` and the kernel's writeback flush can leave the file empty or holding the previous contents, even though the rename itself is ordered.

## Why this reads as an inconsistency rather than an oversight

The append path in the same file **does** fsync properly — `src/store/fs.rs:176` and `:265` sync the data and the parent directory (`append_line_durable`, used for the journal's `EffectExecuted`). So the codebase clearly knows the distinction; `write_atomic` just does not carry it.

## Failure scenario

Host loses power shortly after a company manifest, `meta.json`, or an overlay save is written. On restart the file is empty or stale, and the last save is silently gone — no corruption, no error, just a lost update on a write the caller was told had succeeded.

## Severity

Medium on desktop and self-hosted deployments, which use the fs backend. **Nil for hosted tenants**, which run mongodb.

## Proposed fix

Mirror the append path: `File::sync_data()` on the temp handle before the rename, then `sync_all()` on the parent directory, as at `src/store/fs.rs:249-265`.

If the cost is judged too high for hot paths, make it selective and say which writes are durable and which are not — an unqualified "atomic" in the name reads as a durability promise it does not make.

Found while verifying a code-reading audit against source at `14fe9527`. No open issue matches `fsync`.

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#1049.

`write_atomic` did `create_dir_all` → `write(tmp)` → `rename` with no flush. The rename orders
the **publish**; it does not make the bytes **survive**. Power loss between the rename and the
kernel's writeback leaves the new name pointing at an inode whose data never landed, so the file
comes back empty or holding the previous contents — no corruption, no error, just a save the
caller was told succeeded.

Fixed by mirroring the durable append path in the same file: `sync_data` on the temp file
**before** the rename, `sync_parent_dir` **after** it, reusing the existing helper rather than
inventing a second convention.

## ⚠️ Copying the append path would get this wrong

Leading with this, because a reviewer's instinct is *"why not just mirror `append_line_inner`?"*
— and the answer is that mirroring it literally reintroduces the bug for every overwrite.

`append_line_inner` guards its parent-directory flush:

```rust
if sync {
    file.sync_data()?;
    if creating {          // ← correct there
        sync_parent_dir(&owned_path)?;
    }
}
```

That guard is right for an append: appending to a file that already exists changes **no directory
entry**, so there is nothing in the parent to flush.

**It is wrong here.** A `rename(2)` repoints an *existing* name at a *different* inode — that is a
directory-block change on **every** call, whether or not the destination already existed. Flush the
file but not the directory and a crash can leave the name still resolving to the old inode, with
the new data durable under a name nothing refers to.

So `sync_parent_dir` is unconditional here, and that difference is deliberate rather than an
oversight in either direction. It matters concretely: **overwrite is the common case for all 32
callers**, because these are whole-collection rewrites (`write_atomic(&path, &to_string(&tasks)?)`)
against a file that exists after the first save. With an `if creating` guard, the first save of each
file would be durable and every update after it still losable — which is most of the bug, back.

`overwriting_an_existing_file_still_flushes_the_directory` exists for exactly this, and it is the
test the naive guard fails. Please do not "harmonise" the two paths without it.

## Cost — measured, and larger than I assumed

Stated rather than buried: a durability guarantee whose cost is hidden is one somebody rips out
later on seeing only the cost.

Back-to-back on one machine (macOS/APFS, SSD), same lane, `git stash` / `git stash pop`:

| | store test suite |
| --- | --- |
| without the fix | **2.58s** (157 tests) |
| with the fix | **5.47s** (162 tests) |

**≈ 2.1x on a write-heavy workload.** Two device flushes per save, and they are the expensive
syscalls in a function whose others are cheap — not a rounding error on `create_dir_all` + write +
rename. I recommended unconditional flushing *before* measuring; the number does not change the
recommendation, but it makes it a real trade rather than a free one, so here is the argument rather
than an assumption:

* **Every one of the 32 callers is a state save whose caller was told it succeeded** — the task
  list, agent specs, users, invites, sessions, auth codes, skill states, notification read markers,
  workspace notes and their index. Not one is a cache or a re-derivable artifact, so there is no
  clean line to cut along, and the cost of drawing that line wrong is a silent lost update, which is
  the bug this fixes.
* `append_line` / `append_line_durable` split because *there* the line is clean: routine journal
  chatter versus `EffectExecuted`, which records real money. No equivalent split exists here.
* The test suite is an unusually write-dense proxy — several saves per test against temp dirs. A
  desktop user saves at human speed.
* The exposure is bounded: the fs backend is the **desktop and self-hosted** deployment. Hosted
  tenants run mongodb and never reach this function.

**If a different trade is wanted, the honest lever is a `sync: bool` on `write_atomic_bytes`**
mirroring `append_line_inner` — but it should follow a *measured* hot caller, not a guess. Happy to
add it as a follow-up if a profile turns one up.

## The tests pin the calls, not the physics

Said plainly because the alternative is worse than no test: **CI cannot pull the power.** A flushed
file and an unflushed one are byte-identical on disk, so nothing here observes a lost update. What
is asserted is that `sync_data` and the parent `sync_all` are *requested at the points the recipe
requires* — the part a refactor can silently drop. That those calls make a save survive a power cut
rests on the filesystem contract, not on this test.

That stance is the repo's own, not a new one: `append_probe`'s doc already says *"a flushed
directory and an unflushed one are identical on disk, so the honest check is to count the request
where it is made."* This extends that module with an `atomic_syncs` tally rather than adding a second
mechanism.

**Worth recording: the first version of this test was vacuous.** The probe was recorded at the end of
the `spawn_blocking` block, so deleting `sync_data` outright left all 30 store tests passing — it was
proving the function ran, not that the sync happened. The revert proof caught it. The record now sits
immediately after `sync_data`, and the placement is commented as load-bearing so it is not
"tidied" back down later.

## Live verification

`N/A` by judgement, not omission: the fs backend is desktop/self-hosted, the issue itself scores this
**nil for hosted tenants** (they run mongodb), and `marketing.staging` therefore exercises none of
this code. Source is the ground truth for this one. No staging state was read or changed.

## API Or Behavior Changes

None to any signature. `write_atomic` / `write_atomic_bytes` keep their contract and gain durability;
the body moves from four `tokio::fs` calls to one `spawn_blocking` doing create/write/sync/rename/sync,
mirroring `append_line_inner` — each `tokio::fs` call is its own hop to the blocking pool, and the two
flushes are exactly the operations that hold a pool thread longest.

Every caller becomes slower and durable. Nothing becomes fallible that was not already: the same
`OpenCompanyError::StoreIo` surfaces, now also for a failed flush.

## Tests

      compiles.
- [x] `cargo build --all-targets` — covered by the two test lanes below, which build all targets for
      both feature sets.
      **4317 passed / 0 failed**. `RUST_MIN_STACK` matches `ci.yml:626`.

Five new tests, `#[cfg(unix)]` where they assert the directory flush — matching `sync_parent_dir`'s
own POSIX-only construction.

### Revert proof — two, one per half

| Reverted | Result |
| --- | --- |
| `sync_data` on the temp file | **3 FAILED** — `an_atomic_write_flushes_the_file_and_the_directory_entry`, `overwriting_an_existing_file_still_flushes_the_directory`, `the_bytes_entry_point_is_durable_too` |
| `sync_parent_dir` → naive `if creating` guard | `overwriting_an_existing_file_still_flushes_the_directory` **FAILED** |

The second is the finding above proven directly rather than asserted: the naive mirror of the append
path fails exactly the test written for it, and nothing else.

`an_atomic_write_leaves_no_temp_file_behind` and
`an_atomic_write_onto_an_unusable_parent_reports_the_error` survive both reverts — they are guards on
what the rewrite must *not* have cost (the #887 atomicity property, and errors still surfacing), not
proofs of the durability change, and I am not counting them as such.

## Documentation

`N/A` — no module doc or spec describes the store's flush discipline; the reasoning lives on
`write_atomic_bytes` and on `append_line_durable`, which is where this file already keeps it. The two
blocks worth reading are `Durable, not only atomic` and `Why step 3 is unconditional, unlike the
append path`.

### DIFF
--- a/src/store/fs.rs
+++ b/src/store/fs.rs
@@ -404,6 +404,36 @@ pub(crate) mod append_probe {
             .copied()
             .unwrap_or(0)
     }
+
+    /// How many times [`super::write_atomic_bytes`] flushed a temp file's data
+    /// before publishing it, keyed on the **final** path rather than the temp
+    /// one — the temp name carries a fresh id per call, so a test could never
+    /// ask about it.
+    ///
+    /// Counted here for the reason [`dir_syncs`] is: a flushed file and an
+    /// unflushed one are identical on disk, so the honest check is to count the
+    /// request at the point it is made. This proves the call happens; it does
+    /// not — and cannot — prove what a power cut would leave behind.
+    static ATOMIC_SYNCS: LazyLock<Mutex<HashMap<PathBuf, usize>>> =
+        LazyLock::new(|| Mutex::new(HashMap::new()));
+
+    pub(crate) fn record_atomic_sync(path: &Path) {
+        *ATOMIC_SYNCS
+            .lock()
+            .expect("append-probe poisoned")
+            .entry(key(path))
+            .or_insert(0) += 1;
+    }
+
+    /// How many times `path` was flushed before being published by a rename.
+    pub(crate) fn atomic_syncs(path: &Path) -> usize {
+        ATOMIC_SYNCS
+            .lock()
+            .expect("append-probe poisoned")
+            .get(&key(path))
+            .copied()
+            .unwrap_or(0)
+    }
 }
 
 /// Reads a file to a string, returning an empty string if it does not exist.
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Rebased onto `main` (`02743fda`). CI builds the merge, and `de533ba2 feat(metering): persist planning attempt usage` added a required `planning_attempts` field to `TaskRecord`, so branches predating it stopped compiling against current main.

This branch constructs no `TaskRecord`, so it needed no code change — the rebase alone clears it. Nothing else touched.

Verified on the rebased tip: `cargo test --locked --features openhuman,tinycortex --tests` → **4391 passed, 0 failed**.

Issue #1064, opened 2026-08-18
toasts: hovering defers the dismissal verdict but not the clock, so a toast read for 9s vanishes the instant you look away


M3gA-Mind:
## Summary

Hovering a toast does **not** preserve reading time — it defers the verdict while the debt keeps
accruing. So a toast you read for nine seconds is dismissed the **instant** you glance away, with none
of the grace it would have had if you had never hovered it at all.

Found while diagnosing #969 (the flaky hover-hold e2e test). The test is being fixed separately; this is
the product behaviour underneath it, which may well be intended — but it does not look intended, and
#953 would want to know about its own fix.

## The mechanism

`sweepToasts` (`frontend/src/lib/toast-lifetime.ts`) advances `visibleMs` on **every** tick, then
suppresses only the overdue *decision* while hovered:

```rust
const next = tracked.map((t) => ({ ...t, visibleMs: t.visibleMs + tickMs }));

// Hovering is the one pause the console keeps …
if (env.hovered) return { next, overdue: [] };
```

Note `next` is computed **before** the hover check and is returned by it. The accumulation is not paused;
only the verdict is.

`hovered` is the browser's live `:hover` state, re-read every 500 ms
(`toasterHovered()` in `components/ui/sonner.tsx`), so the moment the pointer genuinely leaves, the very
next tick evaluates a `visibleMs` that has been climbing the whole time the toast was being read.

## What an operator gets

With the shipped constants — `DEFAULT_TOAST_DURATION_MS = 4000`, `DISMISSAL_GRACE_MS = 2000`, sweep every
500 ms:

| What the operator does | When the toast goes |
| --- | --- |
| never hovers | ~6 s (4 s duration + 2 s grace) — as designed |
| hovers at 1 s, reads for 9 s, looks away at 10 s | **within 500 ms of looking away** — `visibleMs` is already ~10 s |
| hovers for a minute | **within 500 ms of looking away** |

So the longer somebody reads a toast, the less time they have to finish reading it after they move the
pointer. The grace exists precisely to avoid yanking something mid-read, and hovering is what spends it.

The docstring says *"a toast under the pointer is being read, and yanking it mid-read is the opposite of
the fix"* and *"The ceiling applies the moment they move away."* The second sentence is implemented
exactly; the interaction between them is what looks unintended.

## Why it is not obviously a bug

It is arguable that ten seconds under the pointer IS the reading time and nothing more is owed. If that
[...]

Issue #969, opened 2026-08-18
tests(e2e): the toast hover-hold test is flaky in the live-brain lane and has reddened two unrelated PRs


M3gA-Mind:
## Summary

`frontend/test/e2e/toast-dismissal.spec.ts:83` — **"hovering a toast still holds it open to be read"** — is
flaky in the **Console E2E (live brain)** lane. It has already turned two unrelated PRs red today:

- **#962** `fix(harness): settle a not-automatable verdict as an answer` — failed on **two** runs, then
  **passed on re-run with no code change**.
- **#954** `ci(console): constrain and verify the contributor's Node version` — same test, same line.

Neither PR touches the console's toast code. #962 is a Rust harness change.

## The failure

```
✘ 127 test/e2e/toast-dismissal.spec.ts:83:1 › hovering a toast still holds it open to be read (14.8s)
   Error: expect(locator).toBeVisible() failed
   Error: element(s) not found
     at frontend/test/e2e/toast-dismissal.spec.ts:93:23
1 failed, 163 passed
```

The toast is **gone** by the time the assertion runs — the hover did not hold it.

## Why it is on a threshold

```ts
await toast.hover();
await page.waitForTimeout(9_000);
await expect(toast).toBeVisible();   // ← line 93
```

The test asserts a hovered toast survives a **fixed 9-second** wait. `toast-lifetime.ts` also carries an
absolute **backstop** ("a backstop, and firing it early would race the exit animation"). If that backstop
can elapse inside the 9s window — or if the synthetic hover is lost when the toast stack reflows — the
toast dismisses and the assertion fails. That it passes on some runs and fails on others is what you would
expect from an assertion sitting right on the boundary, and the **live-brain** lane is exactly where extra
toasts and variable timing appear.

Also worth a look: `const TOAST = "[data-sonner-toast]"` is an unqualified locator matching **every** toast.
Nothing scopes it to the one under the pointer, so a second toast arriving mid-test changes what is being
hovered and asserted.

## Why it matters

It is not a harmless red. It cost a re-run on #962 and is currently blocking #954, which belongs to another
contributor. A test that fails on unrelated PRs trains people to re-run reds without reading them — which is
how a real failure gets waved through.

## Suggested direction (not prescriptive)

Assert the behaviour rather than a wall-clock margin: hold the hover and poll that the toast *stays* present
while hovered, rather than sleeping 9s and checking once. Scope the locator to the specific toast. If the
[...]

PR #1067, opened 2026-08-18
test(e2e): hold the hover and poll, instead of sleeping past the ceiling

M3gA-Mind:
## Summary

`toast-dismissal.spec.ts` — *"hovering a toast still holds it open to be read"* — has reddened **six**
unrelated PRs (#962, #954, #967, #1036, #1038 and one more), every one cleared by a re-run with no code
change. That is roughly 40 minutes of CI, and a more expensive half: a human reading each red and
deciding it is not real. A test that cries wolf trains people to re-run without looking, which is how a
genuine failure gets waved through.

## First: is the test right and the product wrong?

Asked and answered before touching anything, because "fixing" the test would otherwise bury a hole in
#953's fix.

**It cannot be the backstop.** `sweepToasts` (`frontend/src/lib/toast-lifetime.ts`) returns
unconditionally before any overdue computation:

```ts
if (env.hovered) return { next, overdue: [] };
```

and `hovered` is not sonner's belief about the pointer — it is the browser's own state, re-read on every
500 ms tick (`toasterHovered()` in `components/ui/sonner.tsx`):

```ts
el.matches(":hover")   // on [data-sonner-toaster]
```

So the ceiling can only dismiss a hovered toast when the pointer is **genuinely not over the toaster** —
which, for a real operator, means they moved away, and dismissing is exactly what should happen.
**Editing this test buries no defect.**

## What actually made it knife-edge

Not the wall clock racing the backstop, which is what the issue supposed. `visibleMs` accumulates
**during** the hover — `next` is computed before the hover check and returned by it, so only the
*verdict* is suppressed, never the accumulation.

By the end of a nine-second hold the toast is already at `visibleMs ≈ 9000`, well past
`duration + grace = 4000 + 2000 = 6000`. From that point a **single tick that reads `:hover` false
dismisses it instantly, with no grace left**.

So `await page.waitForTimeout(9_000)` followed by one assertion passed only if **~18 consecutive reads of
a live CSS state came back true**, with a stationary synthetic pointer, on a loaded CI runner. That is
fragile by construction, whatever the individual transient turns out to be.

## The change

**Hold the hover in rounds and assert between them** — 12 × 750 ms, re-hovering each round — so a stack
that reflows underneath a stationary synthetic pointer cannot silently end the run.

**The total held time still exceeds the ~6 s ceiling**, so the behaviour under test is unchanged: this
weakens the *flakiness*, not the *assertion*. A toast that genuinely stopped being held would still fail
this test, and would fail it sooner.

**The failure message names the round.** A future red now reads

> the hovered toast was dismissed after ~3000ms of hold

instead of a bare `element(s) not found`. That turns the next occurrence into evidence about *when* the
hold broke, rather than another re-run — which is worth as much as the change itself.

**Scoped the locator.** `const TOAST = "[data-sonner-toast]"` matched every toast with no `.first()`, so
a second toast arriving mid-test would change what is being hovered and asserted — and `toBeVisible()` on
a two-match locator fails as a *strict-mode violation*, not as the thing under test. A `firstToast()`
helper names the subject; the same hazard in the `replayTour` helper is fixed too. The `toHaveCount(0)`
assertions stay **unqualified on purpose** — they mean *every* toast, and scoping them would have
silently weakened them.

## What this does NOT claim

**It was not reproduced locally**, and this PR is not a red-to-green demonstration:

| Run | Result |
| --- | --- |
| **Before**, `--repeat-each=10`, 1 worker | 10 passed — no failure |
| **Before**, `--repeat-each=12`, 4 workers under CPU load | 12 passed — no failure |
| **After**, whole spec file, `--repeat-each=8` | **32 passed** |
| **After**, this test, `--repeat-each=20`, 4 workers under CPU load | **20 passed** |

52 green runs after the change is **not** proof that the flake is gone — the failure was never observed
here, so nothing can be shown to have stopped. The justification is the mechanism above: an assertion
depending on 18 consecutive true reads of a live CSS state is fragile whether or not this machine can
lose the race. An instrumented probe (sampling `matches(":hover")` and `elementFromPoint` every 250 ms
through the window) held true for all 36 samples, so the react-joyride overlay theory was tested and
**discarded** rather than shipped as a premise.

## API Or Behavior Changes

None. One test file; no product code is touched.

## Related product question, deliberately not fixed here

Filed as **#1064** (assigned @oxoxDev): because hovering defers the verdict without pausing the clock, an
operator who reads a toast for nine seconds and then glances away loses it **within 500 ms, with none of
the grace they would have had if they had never hovered it**. That may be intended — it is a real
trade-off against #933's ceiling, since pausing the clock lets a toast latched under a stationary pointer
age forever — so it is a question for the owner rather than a change smuggled into a test fix.

## Tests

- [x] `cargo fmt --all -- --check` — N/A: no Rust file is touched. `Cargo.lock` and submodule pins
  unchanged.
- [x] `cargo clippy --all-targets -- -D warnings` — N/A: no Rust file is touched.
- [x] `cargo build --all-targets` — N/A: no Rust file is touched.
- [x] `cargo test` — N/A for Rust. The console lane is what exercises this: `npm run typecheck:e2e`
  (exit 0), and the spec itself run 52 times as tabled above, including 20 iterations across 4 workers
  under deliberate CPU contention.

No prettier was run: this repo has no prettier config and no prettier CI gate, so it would reformat
untouched code and bury the change.

## Documentation

No doc changes. The reasoning — why the old shape was knife-edge, why the rounds still exceed the
ceiling, and why `toHaveCount(0)` stays unqualified — is recorded in the test's own comments, next to the
assertions it explains.

Refs tinyhumansai/opencompany#969

### DIFF
--- a/frontend/test/e2e/toast-dismissal.spec.ts
+++ b/frontend/test/e2e/toast-dismissal.spec.ts
@@ -24,6 +24,19 @@ type Page = import("@playwright/test").Page;
 const TOAST = "[data-sonner-toast]";
 
 /**
+ * The one toast a test is about (issue #969).
+ *
+ * `TOAST` is unqualified and matches *every* toast on screen. These tests raise
+ * one, so it resolves — but a second arriving mid-test would silently change
+ * what is being hovered and asserted, and `toBeVisible()` on a two-match
+ * locator fails as a strict-mode violation rather than as the thing under test.
+ * Naming the first match makes the subject explicit instead of incidental.
+ */
+function firstToast(page: Page) {
+  return page.locator(TOAST).first();
+}
+
+/**
  * Land on Settings with no tour running, and no welcome dialog in the way.
  *
  * The dialog is not cosmetic here: it is a Radix dialog, so while it is open
@@ -41,7 +54,10 @@ async function settings(page: Page): Promise<void> {
 /** Raise the toast the issue is about, and park the pointer far from it. */
 async function replayTour(page: Page): Promise<void> {
   await page.getByRole("button", { name: "Replay tour" }).click();
-  await expect(page.locator(TOAST)).toBeVisible();
+  // Scoped for the same reason as `firstToast` (#969): a `toBeVisible()` on the
+  // unqualified locator would fail as a strict-mode violation, not as the
+  // assertion it looks like, the moment a second toast is up.
+  await expect(firstToast(page)).toBeVisible();
   await page.mouse.move(20, 20);
 }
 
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
Referenced as `Refs tinyhumansai/opencompany#969` with **no closing keyword**, deliberately: the failure was never reproduced locally, so this cannot be shown red-to-green and #969 should stay open until the flake is observed to have stopped in CI. `closingIssuesReferences` is empty.

The product question found underneath it — hovering defers the dismissal verdict but does not pause the clock — is filed separately as #1064 and assigned to @oxoxDev, deliberately unfixed here.

@coderabbitai review

M3gA-Mind:
## Note for anyone tempted to simplify this back to a fixed sleep

I filed #1064 arguing that the instant-dismissal-on-unhover this test dances around was a product defect. **It is not** — it is a specified contract, pinned by `frontend/test/unit/toast-lifetime.test.ts:85`, whose name is literally *"still ages a hovered toast, so it goes the moment the pointer leaves"*. #1064 is now closed as invalid.

That settles the shape of this test rather than leaving it provisional. The 12×750ms rounds are **not** belt-and-braces awaiting a product fix that would make them redundant — they are the correct way to exercise a contract that says:

* a toast under the pointer is held for as long as the pointer is there (`toast-lifetime.test.ts:77`), **and**
* it is overdue on the first unhovered tick once `visibleMs` has passed `duration + grace` (`:85`).

Because `visibleMs` accumulates *during* the hold by design, a single tick reading `:hover` false is genuinely fatal — so a fixed `waitForTimeout(9_000)` followed by one assertion is not a stricter test, it is the same test with 18 chances to lose a race it cannot detect. Re-hovering each round keeps the pointer's position true against a stack that may reflow underneath it, and the per-round message localises a failure to when the hold broke instead of reporting a bare `element(s) not found`.

So: please do not replace the rounds with a sleep. The behaviour they accommodate is intended, and it is not going to change under them.

Issue #1052, opened 2026-08-18
workflows: a late copilot draft silently overwrites the graph the operator was editing


graycyrus:
Click **Draft it**, wait, start filling the form in by hand while it thinks — and when the draft lands it replaces the graph with no warning. Or: cancel mid-draft, reopen the dialog, and the form fills itself in with the draft you abandoned.

## Cause

`runDraft` (`frontend/src/views/WorkflowCreateDialog.tsx:854-903`) takes its `window.confirm` at `:859`, **before** the request, then hydrates at `:875-891` with no re-check and no live/abort guard — unlike every other async path in the file.

Cancel is `disabled={submitting}` only (`:1339`), not `disabled={drafting}`, so the dialog closes while the draft promise is still alive. The reset effect (`:473`) then runs on the next open and is overwritten by the late response.

## Proposed fix

- Capture a draft token in a `useRef`; bump it in the reset effect and on close, and drop the response if it changed while the request was in flight.
- Re-run `isDraftDirty()` *after* the await and confirm again if the operator made it dirty in the meantime — the pre-flight confirm answers a question that was asked too early.
- Disable Cancel while `drafting`, or abort the request through an `AbortSignal`.

Found during a workflow-surface audit. Related creation defects: #1005 (in-flight feedback).

M3gA-Mind:
## Research notes

**Verdict:** REDUNDANT — RECOMMEND CLOSE
**Still reproduces on main:** no. Checked against `559fb491` (current `upstream/main`). Fixed by **PR #1068**, merged (`ab984570`, merge `9c7a43ee`, branch `fix/1052-late-draft-overwrite`). The commit reads `Refs tinyhumansai/opencompany#1052`, not `Closes`, which is why this is still open.

### Proof, against the three things this issue asked for

**1. Consent moved to after the await.** `frontend/src/views/WorkflowCreateDialog.tsx:1156-1167` — `runDraft` no longer confirms before the request; the comment there names this issue. The confirm now happens at `:1193-1200`, after `draftWorkflowFromDescription` resolves, and only when the response would actually replace something.

The fix went one step past the proposal: rather than keeping the pre-flight confirm *and* adding a post-await re-check, the consent was **moved**. Asking twice would have asked about two different form states, and the pre-flight one also fired for a draft that turns out not automatable, which leaves the form untouched and needed no permission at all.

[...]

PR #1068, opened 2026-08-18
fix(workflows): take the draft-overwrite consent when the draft lands, not when it is requested

M3gA-Mind:
## Summary

Click **Draft it**, keep filling the form in while it thinks, and the draft replaced your work
with no warning. Or cancel mid-draft, reopen, and the dialog filled itself in with the draft
you walked away from.

`runDraft` took its `window.confirm` **before** the request. A copilot draft takes seconds and
the operator is explicitly invited to keep working through it — so the consent was granted over
a form that no longer existed when the answer arrived, and it authorised replacing work started
*after* it was given.

**That is the bug: a permission question asked too early, not a missing guard.** A guard added
around the same early confirm would still be asking about the wrong state.

## Why the early confirm is *removed*, not kept alongside a late one

The obvious shape is "confirm before, then re-check after". It is worse, and a reviewer's
instinct to restore the early confirm as a belt-and-braces safety would reintroduce this bug in
a politer form — so, explicitly:

Two confirms ask about **two different forms**. The first is about the form as it was when the
request went out; the second about the form as it is when the draft lands. Only the second one
is about the thing actually being replaced. Keeping both means interrupting the operator twice
to answer one question, and the first answer is meaningless by the time it matters.

So the confirm moved. It is taken once, against the state being overwritten.

**A side effect worth stating, because it looks like a dropped prompt and is not:** the dialog
no longer asks at all when the draft comes back **not automatable**. That branch leaves the
form completely untouched — the old code asked permission to replace work it then didn't
replace. A confirm that appears only when something is genuinely about to be overwritten is the
correct behaviour.

## Why Cancel stays enabled

Disabling Cancel while `drafting` is the one-line fix and a worse product: it traps the
operator in a dialog until a slow model call returns, to prevent a problem they are trying to
walk away from.

**The draft becomes irrelevant instead of un-cancellable.** `draftEpochRef` identifies the
dialog's current contents; the reset effect bumps it on every open and re-hydrate. `runDraft`
captures it before the request and compares after, so a response whose dialog has moved on is
**dropped silently** — nothing was asked of the operator, so nothing needs explaining, and its
banners belong to a form that is gone.

## The detail that decides whether this fix is real

**`isDraftDirty()` closes over its render's state.** Re-calling it after the `await` — the
literal shape the issue proposes — re-reads the values captured *when the request was issued*.
It looks like a re-check, passes review, and does nothing: the operator's new typing is
invisible to it and the draft still clobbers.

Dirtiness therefore rides `draftDirtyRef`, refreshed by an effect on every render, so the
post-await read sees what is on screen now. This is the load-bearing line of the change; the
rest is plumbing around it.

## Reusing the file's convention rather than adding a second one

Every other async path here uses an effect-scoped `let live = true` flipped by the effect's
cleanup. `runDraft` is an **event handler** — there is no cleanup to flip — so the same idea is
carried on a ref instead. That is recorded at `draftEpochRef` so it reads as the existing
convention adapted, not a rival one.

Two smaller consequences of the same ownership rule:

* the `finally` clears `drafting` only when the epoch still matches — a stale request switching
  off the spinner would kill the indicator for a newer draft the operator *is* waiting on;
* the reset effect clears `drafting`, because cancelling mid-draft previously left **Draft it**
  inert on reopen until an abandoned promise settled.

## API Or Behavior Changes

**No API change.** Console-only; no route, no wire type, no host code.

**Behaviour.**
1. The overwrite confirm is asked when the draft lands, against the form being replaced.
2. It is not asked when nothing will be overwritten (clean form, or a non-automatable result).
3. A draft whose dialog was closed, reopened or re-hydrated is discarded instead of applied.
4. Cancel remains enabled throughout — unchanged.

## Tests

- [x] `cargo fmt --all -- --check` — N/A: no Rust file is touched. The diff is three paths, all
      under `frontend/`.
- [x] `cargo clippy --all-targets -- -D warnings` — N/A: no Rust file is touched, so neither the
      default nor the `openhuman,tinycortex` lane has anything to lint here.
- [x] `cargo build --all-targets` — N/A as above; `git diff --stat Cargo.lock` is empty.
- [x] `cargo test` — N/A as above. **Ran instead, as the console's CI lane does:**
  - `npm run typecheck`, `typecheck:unit`, `typecheck:e2e` — all clean
  - `npm run build` — clean
  - `npx vitest run` — **960 passed, 98 files, 0 failed** (5 of them new)

  Not run: `prettier`. CI does not run it, and files untouched by this change already fail
  `prettier --check` on `main`, so running it would produce reformatting noise unrelated to this.

### Every new test was proven to fail with its fix reverted — 3 reverts, 5 tests

|---|---|
| the staleness check is removed | `drops a response whose dialog has moved on`, `drops rather than asking when the dialog moved on AND the form is dirty`, `treats any epoch change as stale, in either direction` |
| dirtiness is ignored (the silent-clobber bug itself) | `asks when the operator has made the form dirty since the request` |
| staleness is checked *after* dirtiness | `drops rather than asking when the dialog moved on AND the form is dirty` |

That third revert is the one worth keeping: checking dirtiness first still passes four of the
five tests, and produces a confirm about a form the operator is no longer looking at. A question
nobody can place is worse than silence.

The decision lives in `@/lib/workflow-draft` beside `draftBanners`, which is where this file
already puts logic it wants proved without a render.

**The wiring is untested.** The tests pin the `draftLanding` decision; that `runDraft` calls it,
that the epoch is bumped on open, and — most importantly — that dirtiness comes from the **ref**
rather than a direct `isDraftDirty()` call are guarded by the compiler and by reading, not by a
test. If someone simplifies that ref away, every test here still passes and the bug returns.

Proving it needs a render harness with a stubbed slow request and a stubbed `window.confirm`;
this repo has no component-test harness for this dialog, and its e2e lane needs a live host.
Recorded as a real gap rather than papered over — it is the same gap `#1005` touches on the same
surface.

**Not reproduced on a live tenant.** The cause is precisely located in source and the fix is
reasoned from it; no staging observation was made.

## Documentation

No spec doc changes. The reasoning lives at the code it governs: `draftLanding` carries why
consent must be taken at landing, `draftEpochRef` carries why an event handler cannot use the
file's effect-scoped `live` flag, and `draftDirtyRef` carries why re-calling `isDraftDirty()`
after an await cannot work.

Refs tinyhumansai/opencompany#1052

### DIFF
--- a/frontend/src/lib/workflow-draft.ts
+++ b/frontend/src/lib/workflow-draft.ts
@@ -43,3 +43,49 @@ export function draftBanners(drafted: WorkflowDraftFromDescription): DraftBanner
     reason: drafted.reason ?? "This is better done once than built into a workflow.",
   };
 }
+
+/**
+ * What a landed copilot draft may do to the form it came back to (issue #1052).
+ *
+ * - `drop` — the dialog moved on (closed, reopened, re-hydrated) while the
+ *   request was in flight. The answer is about a form that no longer exists, so
+ *   it is discarded silently: nothing was asked, so nothing needs explaining.
+ * - `confirm` — the form holds work the operator did, so replacing it is a
+ *   question, not a side effect.
+ * - `apply` — nothing would be lost; hydrate.
+ */
+export type DraftLanding = "drop" | "confirm" | "apply";
+
+/**
+ * Decides what a draft response is allowed to do **at the moment it lands**.
+ *
+ * The bug this exists for: `runDraft` used to take its `window.confirm` *before*
+ * the request. A model call takes seconds, and the operator is invited to keep
+ * typing while it runs — so the consent was granted over a form that no longer
+ * existed by the time the answer arrived, and a draft could replace work the
+ * operator started *after* agreeing. Consent has to be taken against the state
+ * being replaced, which means after the await, which means here.
+ *
+ * `epoch` is the identity of the dialog's contents. The reset effect bumps it on
+ * every open, so a draft abandoned by cancelling cannot hydrate the *next* open —
+ * the failure mode where reopening the dialog filled itself in with a draft the
+ * operator had walked away from.
+ *
+ * Staleness is checked **before** dirtiness on purpose: a response for a form
+ * that is gone must not raise a confirm about a form the operator is no longer
+ * looking at. Asking a question nobody can place is worse than silence.
+ *
+ * Pure so the three outcomes can be proved without a slow request and a rendered
+ * dialog — the same reason `draftBanners` above lives here rather than inline.
+ */
+export function draftLanding(args: {
+  /** The dialog's epoch when the request was issued. */
+  requestedEpoch: number;
+  /** The dialog's epoch now that the response has landed. */
+  currentEpoch: number;
+  /** Whether the form holds operator work **right now**, not at request time. */
+  dirtyNow: boolean;
+}): DraftLanding {
+  if (args.requestedEpoch !== args.currentEpoch) return "drop";
+  return args.dirtyNow ? "confirm" : "apply";
+}
--- a/frontend/src/views/WorkflowCreateDialog.tsx
+++ b/frontend/src/views/WorkflowCreateDialog.tsx
@@ -46,7 +46,7 @@ import {
   configFromDraft,
   hasConfigForm,
 } from "@/lib/workflow-node-config";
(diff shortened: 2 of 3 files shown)


M3gA-Mind:
@coderabbitai review

Issue #1050, opened 2026-08-18
tenancy: a failed set_owner is swallowed, orphaning a provisioned company from its tenant


graycyrus:
`ownership.set_owner` failures are logged and swallowed in both places that record a company's tenant:

- `src/server/provision.rs:299-303`
- `src/bin/opencompany.rs:250-253`

The boot path self-heals — `bin:249` re-runs `set_owner` on every boot, so a company named on the command line recovers on restart.

**Provision-route companies do not.** In shared-single-DB mode the in-memory registry has the tenant, the `owners` collection does not, and boot hydration filters on exactly that row (`src/bin/opencompany.rs:1211-1228`). So after a restart the company is unattributable to its tenant: it is not hydrated for that tenant, and a tenant-scoped purge or export will miss it.

## Failure scenario

1. Tenant `acme` provisions a company through `POST /api/v1/companies`.
2. The `owners` write fails transiently (mongo blip, election, timeout). The request still returns **success** — the company exists.
3. On the next restart, hydration finds no `owners` row and does not load it for `acme`.
4. The company is orphaned: present in the database, owned by nobody, invisible to the tenant that paid for it and missed by tenant-scoped operations.

This is a tenancy/data-retention correctness problem, not a cross-tenant leak — nobody else gains access.

## Severity

Medium, and confined to shared-single-DB deployments using the provisioning route.

## Proposed fix

Fail the provision request when `ownership.set_owner` errors, or queue a retry that must succeed before the company is reported as created. A company that exists with no owner row is a worse outcome than a provision that failed cleanly and can be retried by the caller.

Found while verifying a code-reading audit against source at `14fe9527`.

M3gA-Mind:
## Research notes

**Verdict:** REDUNDANT — RECOMMEND CLOSE
**Still reproduces on main:** no. Checked against `559fb491` (current `upstream/main`). Fixed by **PR #1073**, merged 2026-08-18 (`2dd408e7`, merge `e1a6968a`, branch `fix/1050-ownership-write-before-provision`). The commit says `Refs tinyhumansai/opencompany#1050` rather than `Closes`, which is why this is still open.

### Proof

The swallow at the old site is gone, and the fix went further than this issue proposed: the durable owner row is now written **before** the runtime is built, not after.

- `src/server/provision.rs:292-307` — the write is now a hard gate. On failure the handler returns `503 ownership_not_persisted` ("could not record which tenant owns this company, so it was not created") and nothing is provisioned.
[...]

PR #1073, opened 2026-08-18
fix(tenancy): record company ownership before the company exists (#1050)

M3gA-Mind:
## An orphaned company is missed by a tenant-scoped purge or export

Its data survives a deletion someone believes they performed.

That is the consequence, and it is why this is more than a swallowed log line. `ownership.set_owner` failures were logged and discarded, so a transient write failure (a mongo blip, an election, a timeout) returned **`201 Created`** for a company with no `owners` row. Boot hydration filters on exactly that row (`bin/opencompany.rs:1242`), so after a restart the company is unattributable to its tenant.

Traced through, an orphan today is worse than "untidy":

- **Not hydrated** for its tenant at boot.
- **Addressable by nobody.** `owner_of` returns `None`, so `authorize_address` (`platform_auth.rs:489`) and the GraphQL twin answer **403** — the tenant is locked out of the company it paid for.
- **Missed by tenant-scoped operations**, including a purge. The data stays in the shared database, attributable to no one.

It fails *closed*, so this is not a cross-tenant leak — nobody else gains access. It is a tenancy and data-retention correctness problem, and the purge-invisibility is the part that matters.

Confined to shared-single-DB deployments using the provisioning route. The boot path already self-heals (`bin:249` re-runs `set_owner` every boot); provision-route companies never did.

## Which option, and why not the other three

Issue #1050 proposes failing the request. **At the site it names, that cannot be atomic** — and this is the crux of the change.

By `provision.rs:299` the runtime is already built and `registry().insert()` has already run: files and rows exist. Returning an error there leaves **exactly the orphan it means to prevent**, plus a caller who believes nothing happened. True atomicity would need a compensating delete, and there is no company-deletion path to roll back with — `archive` is a lifecycle transition plus registry removal, not a purge.

But `id` and `tenant` are both fully resolved at `provision.rs:210`, well before the build. So:

**The durable owner row is now written *before* the company is built, with a bounded retry, and the request is refused if it cannot be written.**

That makes failing honest, because at that point there is nothing to roll back. It also inverts the failure direction, deliberately:

> **An `owners` row with no company is harmless. A company with no `owners` row hides data from a purge.**

A stale row hydrates a benign in-memory entry and can be removed; a failed build also rolls it back best-effort. The residual failure is the one that leaves data **visible** rather than the one that hides it. That is the trade this change makes on purpose.

Self-healing (the other suggested option) does not work here: boot re-runs `set_owner` only for the company named on the command line, so a provision-route company never heals on any boot.

## What window remains

- **A refused provision creates nothing.** The caller gets `503 ownership_not_persisted` and can retry. No orphan.
- **A transient failure is retried** (3 attempts, no backoff — the retry is for a failed round-trip, not a busy one, and a caller is waiting on the request). Most would-be refusals stay successes.
- **A crash between the owner write and a successful build** leaves a row naming a company that was never created — benign, and rolled back on a clean build failure.
- **The boot path keeps its swallow, deliberately.** `bin:249` re-runs `set_owner` on every boot, so a boot-company failure self-heals on the next restart: its window is one boot cycle, not permanent. Changing it would trade a self-healing warning for a boot abort.
- **Companies already orphaned are not healed by this.** Nothing here repairs existing state — see the detection section below, which matters more to anyone who has one today.

What makes it observable in the meantime: the failed write is no longer silent at the point it happens. Each failed attempt logs at `warn` with the company and tenant, a final failure logs at `error`, and — the part that actually reaches a human — **the API call fails** instead of reporting success. The defect was never really the failed write; it was that nothing anywhere said it happened.

## Existing orphans *can* be found by a query — but nothing computes it

Established, not assumed:

- `MongoStore::list()` is `find(doc! {})` — **unfiltered across every tenant** in the shared database (`store/mongodb.rs:602`).
- `OwnershipStore::owners()` returns every row (`store/select.rs:260`).
- So the orphan set is exactly **`list()` ids − `owners()` ids**, computable from ports that exist today.

But `owners()` has **one caller in the entire tree** — boot hydration at `bin/opencompany.rs:1242` — and there is no orphan/unowned surface anywhere in `src/` or `docs/`. An operator cannot run that query without writing code.

**An operator with orphans today needs that detection more than this fix**, which only prevents new ones. It is filed separately rather than widened into here.

## API Or Behavior Changes

- `POST /api/v1/companies` now answers **`503 ownership_not_persisted`** when the company → tenant row cannot be written, and **creates nothing**. Previously it answered `201 Created` and left an orphan.
- Only reachable on a backend that persists ownership (MongoDB). Deployments with no ownership store are unaffected — the branch is skipped exactly as before.
- A failed build now also removes the ownership row it wrote, best-effort.
- **Unchanged:** the boot path, the in-memory owner map, hydration, `archive`, and every success path.

## Tests

Gated lane — default features skip this code.

- [x] `cargo clippy --no-deps -p opencompany --features openhuman,tinycortex --all-targets -- -D warnings` — mirrors CI's `--no-deps`; no findings.
- [x] `cargo check -p opencompany --features openhuman,tinycortex --tests` — clean, via the test run below.
- [x] `cargo test -p opencompany --features openhuman,tinycortex` — **4324 passed, 0 failed, 3 ignored.**

### Reverted-fix proof

Reverting `persist_owner_with_retry` to its pre-#1050 behaviour (one attempt, failure swallowed):

| Test | On revert |
| --- | --- |
| `a_transient_ownership_failure_is_retried_and_succeeds` | **FAILS** |
| `a_persistent_ownership_failure_gives_up_and_reports_it` | **FAILS** — the swallow returns `Ok`, which is the bug |
| `a_successful_ownership_write_is_attempted_once` | passes — see below |

The third passes either way **by design**: the happy path costs one write before and after. It is a guard that the retry does not multiply the normal case, not a regression test, and it is not counted as coverage of the fix.

### A coverage gap, stated rather than papered over

The **route-level** behaviour — that provisioning returns `503` and creates nothing — is not covered by a test. Reaching it requires `state.stores()` to be `Some`, which means constructing a `StorageHandles` with ~30 port implementations; **no test in this repository constructs one**, and the single-store backend that builds one (`sqlite`) is not in this lane.

What is covered is `persist_owner_with_retry`, where the new decision lives, through a fake `OwnershipStore` that fails a chosen number of times. The existing provision tests continue to pass unchanged because they run with no ownership store, which also pins that the new branch is inert on such a host.

## Documentation

None needed. The reasoning — why the write moved ahead of the build, why the inverted failure direction is the safer one, and why the boot path keeps its swallow — is recorded at the call site and on `persist_owner_with_retry`.

Refs tinyhumansai/opencompany#1050

### DIFF
--- a/src/server/provision.rs
+++ b/src/server/provision.rs
@@ -267,9 +267,63 @@ async fn provision(
     if let Some(overlay) = state.memory_overlay() {
         builder = builder.with_memory_overlay(overlay);
     }
+    // Issue #1050: the durable owner row is written BEFORE the company exists.
+    //
+    // It used to be written after, best-effort, and a failure was logged and
+    // swallowed — so a transient write failure returned `201 Created` for a
+    // company with no `owners` row. Boot hydration filters on exactly that row,
+    // so after a restart the company was unattributable to its tenant: not
+    // hydrated, addressable by nobody (`owner_of` → `None` → 403), and **missed
+    // by a tenant-scoped purge or export**. Data surviving a deletion someone
+    // believes they performed is the part that makes this more than untidiness.
+    //
+    // Ordering it first is what makes failing honest. Failing at the old site
+    // could not be atomic: by then the runtime is built and registered, so
+    // returning an error left exactly the orphan it meant to prevent, and there
+    // is no company-deletion path to roll back with (`archive` is a lifecycle
+    // transition plus registry removal, not a purge).
+    //
+    // This inverts the failure direction, deliberately: a crash between here
+    // and a successful build leaves an `owners` row naming a company that was
+    // never created. That is the strictly safer orphan — it hydrates a harmless
+    // in-memory entry and can be removed, whereas a company with no owner row
+    // hides data from a purge. Prefer the failure that leaves data visible over
+    // the one that hides it.
+    if let Some(ownership) = state.stores().and_then(|s| s.ownership.clone())
+        && let Err(err) = persist_owner_with_retry(ownership.as_ref(), &id, &tenant).await
+    {
+        tracing::error!(
+            company = %id,
+            tenant = %tenant,
+            error = %err,
+            "refusing to provision: company ownership could not be persisted"
+        );
+        return envelope(
+            StatusCode::SERVICE_UNAVAILABLE,
+            "ownership_not_persisted",
+            "could not record which tenant owns this company, so it was not created. \
+             Nothing was provisioned — retry the request.",
+        );
+    }
+
     let runtime = match builder.build().await {
         Ok(runtime) => runtime,
-        Err(err) => return ApiError(err).into_response(),
+        Err(err) => {
+            // The owner row above now names a company that does not exist.
+            // Best-effort removal keeps the reversed orphan from lingering; if
+            // it fails the row is still the benign direction.
+            if let Some(ownership) = state.stores().and_then(|s| s.ownership.clone())
+                && let Err(cleanup) = ownership.remove_owner(&id).await
+            {
+                tracing::warn!(
+                    company = %id,
+                    error = %cleanup,
+                    "provision failed and its ownership row could not be rolled back; the row \
+                     names a company that was never created"
+                );
+            }
+            return ApiError(err).into_response();
+        }
     };
 
     // The same refusal boot applies to `serve --company`: a company with no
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
Filed the detection gap as #1077 (assigned to @oxoxDev).

This PR only prevents **new** orphans. A company already in that state stays there, and an operator has no way to find one — the query is computable from ports that exist today (`company.list()` is unfiltered across tenants, `owners()` returns every row), but `owners()` has exactly one caller in the tree and there is no orphan surface anywhere. Anyone who has an orphan today needs that more than they need this.

@coderabbitai review

Issue #1075, opened 2026-08-18
workflows: opencompany copies upstream's SSRF predicates because url_guard is private


M3gA-Mind:
## Summary

OpenCompany now carries a **copy** of the upstream SSRF/allowlist predicates, because the originals are unreachable. Exporting them upstream would let the copy be deleted rather than maintained.

## Why the copy exists

#1048: a workflow dry run reported `ok` for an `http_request` node aimed at a host the capability layer blocks, and the real run refused it immediately. The fix evaluates the part of the guard's verdict that is decidable without performing anything — scheme, private/loopback/link-local literal, and the company's `web_allowed_domains`.

Making that verdict requires the same rules the real client applies. In the vendored runtime those live in `src/openhuman/tools/impl/network/url_guard.rs`, which is declared `mod url_guard;` — **private** — inside `pub mod network`, and every predicate is `pub(super)`:

- `validate_url`
- `is_private_or_local_host`
- `host_matches_allowlist`
- `normalize_allowed_domains` / `normalize_domain`
- `extract_host`

So nothing in opencompany can call them, and `src/workflows/caps/http.rs::preflight_refusal` reimplements the subset it needs.

## Why that is a real risk

Two implementations of an SSRF verdict drift. The dangerous direction is specific: if upstream **loosens** a rule, the copy keeps refusing, and a dry run then blocks a workflow that would really run — a false refusal an operator cannot distinguish from a genuine one without arming the graph anyway.

It is currently mitigated, not solved. `dry_run_refusal_matches_the_real_client` drives a table of URLs through the real `GuardedHttpClient` and asserts the two verdicts agree, so drift fails a test rather than shipping. But a table only covers the cases someone thought to enumerate; the normalization edge cases (a malformed allowlist entry, the fail-closed sentinel when every entry is malformed) are exactly the ones a copy gets subtly wrong, and the copy deliberately declines to reproduce them at all.

## The ask

Make the pure predicates callable from a downstream crate — `pub` on the module, or a small `pub` façade exposing a validation that performs no I/O. `validate_url` is already the DNS-free half; `validate_url_with_dns_check` is the one that resolves and should stay unavailable to a dry run, since resolving is itself a network effect.

With that, `preflight_refusal` collapses to a call and the parity test becomes unnecessary.

## Related

- tinyhumansai/opencompany#1048 — the dry-run false green this was needed for
- `src/workflows/caps/http.rs` — where the copy lives, with its reasoning

[...]

PR #1076, opened 2026-08-18
Refuse a dry-run target the real run would refuse

### LINKED ISSUE #1048: workflows: a dry run reports ok for a node the real run refuses, so Test run gives a false green
A dry run reports success for a node that a real run cannot even start, so the one control an operator has for checking a graph before arming it gives a false green.

## Observed

Against a host at `c30b07b3` (features `openhuman,tinycortex,mcp`), driving the real console:

- A workflow whose `http_request` node targets a host the capability layer blocks.
- **Test run** → that node reports **`ok`** and the run shows a green result.
- **Run** (the real one) → the same node fails immediately with a capability refusal, and the run is journaled as failed.

## Why

`dry_run.rs` stubs every effectful capability slot by design, so the SSRF/allow-list refusal that the real `GuardedHttpClient` raises is never exercised. The grant check is deliberately retained in dry runs, but a *target* refusal is not a grant refusal — it happens inside the capability, which the dry run has replaced.

## Why it matters

Test run is the affordance an operator uses to decide whether a graph is safe to arm on a schedule. A green dry run followed by a real run that cannot start is worse than no dry run: it converts "I checked it" into a false belief. It also interacts badly with #976 — a stage-less or misconfigured graph can be armed on a schedule after appearing to pass.

## Proposed fix

Make the dry run evaluate the checks that do not require performing the effect, rather than stubbing the whole capability:

- resolve and validate `http_request.url` against the same SSRF/allow-list guard the real client uses, and report a refusal as a dry-run failure
- likewise refuse a `tool_call` whose slug the company's `[tools].allow` does not cover (this part may already hold — verify)

Alternatively, if a dry run genuinely cannot know, it must report that node as **unchecked** rather than `ok`. A truthful "not verified" is fine; a false `ok` is not.

Found while empirically verifying #1005-#1017 against a live host.

M3gA-Mind:
## Summary

A dry run reported **`ok`** for an `http_request` node aimed at a host the capability layer blocks, and
the real run failed immediately. Test run is the single control an operator has before arming a graph
on a schedule, so a green there is worse than no dry run at all: it converts *"I checked it"* into a
false belief.

Reproduced before fixing, on the same graph as the existing
`t5_http_request_to_loopback_is_ssrf_denied`:

```
dry:  nodes: [WorkflowRunNodeRow { node_id: "fetch", status: Ok }]
      body:  "[dry run] http_request was not sent"
real: http_request: Blocked local/private host: 127.0.0.1
```

**Reproduced locally rather than on staging, deliberately.** Reproducing needs *creating a workflow*,
which is a write to a live company, and this is deterministic code with no tenant state in it — the
verdict is a pure function of the URL and the manifest. A live probe would have left a workflow behind
and told us nothing the runner does not.

## The stubs stay — this is not "stop stubbing"

`dry_run.rs` stubs every effectful capability **by design**, and it must: a Test run that performed
real effects would be a far worse bug than this one. **No request is issued in either branch**, before
or after this change.

What changes is that the verdicts a dry run can reach *without performing anything* are now reached.
The distinction, slot by slot:

| Stubbed slot | Category | Verdict |
|---|---|---|
| `tool_call` grant (`[tools].allow`) | decidable pre-flight | **already correct** — `DryRunTools` keeps the fail-closed grant gate deliberately. The issue's "verify this may already hold" item does hold. |
| `http_request` URL vs **private / loopback / link-local literal** | decidable pre-flight — pure string→IP, no DNS | **now refuses** ← the reported bug |
| `http_request` URL vs **company `web_allowed_domains`** | decidable pre-flight — pure host-vs-list | **now refuses**, conservatively (below) |
| `http_request` **DNS resolution** | requires the network | **must not**, see below |
| host down / TLS / stale credential / non-2xx / malformed body | requires execution | **must not** |
| agent inference, state persistence, delivery | requires execution | **must not** |

### Why DNS is excluded

The real path calls `validate_url_with_dns_check`, which **resolves** the host to catch a public name
pointing at a private address. **Resolving is itself a network effect, and it fails offline.** A dry
run must not attempt it.

The consequence is stated rather than hidden: a hostname that resolves to a private address still
passes this check and is refused by the real run. That is a **missing** refusal, which is honest. An
invented one would block a working graph, and an operator cannot distinguish that from a real refusal
without arming the thing anyway — which is the failure that erodes trust fastest.

### Conservative where a copy could invent a refusal

Upstream substitutes a fail-closed sentinel when *every* allowlist entry is malformed, refusing all
traffic. This deliberately does **not** reproduce that: an allowlist `preflight_refusal` cannot read
cleanly returns `None` and checks nothing. Reproducing that subtlety is the easiest place for a copy
to refuse something the real run would allow.

### An allowed target now says "not checked", not "ok"

Passing pre-flight yields `"[dry run] http_request was not sent — target allowed, delivery not
checked"` plus a `checked` field naming the limit, rather than a bare stub that reads as a successful
request. A dry run cannot know whether the host is up, the credential is current, or the response
parses, and saying so is more useful than a green that implies it does.

## The copy, and how drift is caught

The authoritative rules live upstream in `mod url_guard;` — a **private** module whose predicates are
`pub(super)`, so nothing in opencompany can call them. `preflight_refusal` is therefore a copy, which
is a real drift risk and is named as one.

`dry_run_refusal_matches_the_real_client` is what makes that manageable. It does **not** compare
against upstream's source; it drives the same URLs through the real
`GuardedHttpClient` and asserts the two verdicts agree. So it keeps holding when upstream changes its
internals, and **fails loudly the day upstream loosens a rule**. Six cases: loopback, `localhost`,
RFC1918, link-local `169.254.169.254`, IPv6 `[::1]`, and an off-allowlist public host.

That comparison needs no network and performs no effect, because the real guard refuses these
**before it dials anything** — which is also why only refusals are compared. Confirming an *allowed*
URL would mean actually issuing the request. The direction covered is the one that matters.

Filed **tinyhumansai/opencompany#1075** asking upstream to export the predicates, so this copy can be
deleted rather than maintained.

## A trap worth more than the fix

**`WorkflowRunContext::new(true)` is `scheduled`, not `dry_run`.** The dry flag defaults to `false`
and is set after construction, exactly as the run route does it.

My first reproduction used `new(true)` believing it selected a dry run. It ran a **real** run, which
refused the loopback target correctly, so the test asserting "a dry run refuses this" went **green —
for entirely the wrong reason**. I found it only by printing the actual error instead of trusting the
assertion.

A test that passes for the wrong reason is indistinguishable from one that works, right up until it
matters. The suite now has a `dry_context()` helper whose doc says which argument is which, so the
next person writing a dry-run test does not have to rediscover it.

## Two existing tests had pinned the bug in place

`dry_http_echoes_without_sending` and `a_dry_bundle_wires_stubs_and_noop_state` both used
**`http://127.0.0.1:9/` and asserted success** — the exact URL class the real guard refuses. So the
false green was not merely untested; it was *asserted*, twice. Both now use a public URL, since what
they are really about is the slot being the stub, not what any particular target does.

## API Or Behavior Changes

Dry runs only. An `http_request` node whose target the real run would refuse now **fails the dry run**
with the same `EngineError::Capability` shape and message the live client produces, so the node's
`on_error`/retry policy governs it identically. An allowed target's stub body is reworded as above.

Live runs are untouched. No route, schema or default changes.

## Tests

- [x] `cargo build --all-targets` — covered by the clippy and test runs above.
- [x] `cargo test` — **4285 passed / 0 failed** gated (`--features openhuman,tinycortex --lib`) and
      **2809 passed / 0 failed** at default features.

Removing the `preflight_refusal` call from `DryRunHttp::request` fails exactly three, tests untouched:

```
workflows::caps::dry_run::tests::dry_http_refuses_a_blocked_target
workflows::caps::dry_run::tests::dry_run_refusal_matches_the_real_client
workflows::runner::tests::a_dry_run_refuses_a_target_the_real_run_would_refuse
```

All pass with it restored. The first is the unit verdict, the second is the parity guard, the third is
end to end through `run_workflow` — the shape the operator actually meets.

## Not in this PR

A node that *passes* pre-flight still reports `status: Ok` in the run's node rows. The body now says
"not checked", but the row-level status still reads as success, which is the residual half of the same
false green. Closing that properly needs a distinct not-executed status on `WorkflowRunNodeRow` —
a wider change than this fix, and better argued on its own.

## Documentation

`preflight_refusal` carries the reasoning at the point of decision: what it decides, what it
deliberately leaves to the real run, why it is never stricter than the real guard, and what `None`
means (**"not refused by the decidable subset"**, never "this would work"). `DryRunHttp::request`
states why both branches issue nothing.

Closes tinyhumansai/opencompany#1048

### DIFF
--- a/src/workflows/caps/dry_run.rs
+++ b/src/workflows/caps/dry_run.rs
@@ -137,15 +137,53 @@ impl ToolInvoker for DryRunTools {
 /// [`GuardedHttpClient`](super::http::GuardedHttpClient): it echoes the request
 /// descriptor back without issuing anything, so no outbound request (guarded or
 /// not) leaves the process.
-pub(super) struct DryRunHttp;
+pub(super) struct DryRunHttp {
+    /// The company's `[tools].web_allowed_domains`, so the target check a dry
+    /// run *can* make is made against the same list the live client uses.
+    allowed_domains: Vec<String>,
+}
+
+impl DryRunHttp {
+    pub(super) fn new(allowed_domains: Vec<String>) -> Self {
+        Self { allowed_domains }
+    }
+}
 
 #[async_trait]
 impl HttpClient for DryRunHttp {
+    /// Refuses a target the real run would refuse; otherwise reports the request
+    /// as **not checked**, never as a success.
+    ///
+    /// Issue #1048: this used to return the same cheerful stub for every URL, so
+    /// a node aimed at a host the capability layer blocks reported `ok` on Test
+    /// run and failed immediately on the real one. Test run is the single
+    /// control an operator has before arming a graph on a schedule, so a green
+    /// there followed by a refusal on the real run is worse than no dry run at
+    /// all — it converts "I checked it" into a false belief.
+    ///
+    /// The refusal message and `EngineError::Capability` shape match
+    /// [`GuardedHttpClient`](super::http::GuardedHttpClient)'s, so a node fails
+    /// under its own `on_error`/retry policy exactly as it would live.
+    ///
+    /// **Still no request is issued, in either branch.** The verdict below is a
+    /// pure function of the URL and the company's config — see
+    /// [`preflight_refusal`](super::http::preflight_refusal) for what it decides
+    /// and, more importantly, what it deliberately leaves to the real run.
     async fn request(&self, request: Value, _conn: Option<&str>) -> TfResult<Value> {
+        if let Some(reason) = super::http::preflight_refusal(&request, &self.allowed_domains) {
+            tracing::debug!(%reason, "workflow dry run: refusing http_request target");
+            return Err(EngineError::Capability(format!("http_request: {reason}")));
+        }
         tracing::debug!("workflow dry run: stubbing http_request (not sent)");
         Ok(json!({
             "status": Value::Null,
-            "body": "[dry run] http_request was not sent",
+            // Deliberately not phrased as a result. A dry run cannot know
+            // whether the host is up, the credential is current or the response
+            // parses, and saying so is more honest than a green that implies it
+            // does — the opposite error to the one #1048 fixed, and the one that
+            // erodes trust fastest because an operator cannot tell it is wrong.
+            "body": "[dry run] http_request was not sent — target allowed, delivery not checked",
+            "checked": "target only: a real run may still fail to reach this host",
             "request": request,
             DRY_RUN_MARKER: true,
         }))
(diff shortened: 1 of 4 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:

PR #1080, opened 2026-08-18
fix(workflows): refuse to arm a schedule with nowhere to deliver

### LINKED ISSUE #1046: bug: workflows: a schedule is armed on a graph whose report has nowhere to land, so every scheduled run drops its output unseen
## Summary

Saving a graph with a cron trigger checks the cron, the node kinds, the teammate names and — since #981's server-side gate — a named `channel` target. It does not check that the graph's terminal `output` node can deliver at all. A workflow whose only destination is `owner` on a company with no mailbox configured saves clean, arms clean, and then drops its report on every fire.

Manually this is survivable: the run banner and the history row both say **"1 report not delivered"** with the reason. On a schedule nobody is watching either surface. The workflow runs every N minutes, burns a metered agent turn each time, and produces nothing anyone receives.

This is the same family as #976 / PR #1028 ("refuse to arm a schedule the graph cannot keep") — that one covers a graph with nothing to execute; this is a graph with nowhere to deliver.

## Reproduction

On a company whose admins have no mailbox configured (`smoke1` today), save a scheduled graph whose output goes to `owner`:

```js
await fetch('/api/v1/company/workflows/qa_probe_wf', {
  method: 'PUT', credentials: 'include',
  headers: { 'content-type': 'application/json' },
  body: JSON.stringify({
    id: 'qa_probe_wf', name: 'QA probe wf', description: 'QA test workflow',
    nodes: [
      { id: 'start', kind: 'trigger', name: 'Manual start', schedule: '*/5 * * * *' },
      { id: 'note', kind: 'agent', name: 'Write a line', summary: 'Write one short sentence saying hello.', agent: 'product_manager' },
      { id: 'done', kind: 'output', name: 'Report', destination: { kind: 'owner' } },
    ],
    edges: [{ from: 'start', to: 'note' }, { from: 'note', to: 'done' }],
  }),
});
// -> 200, schedule armed
```

Run it once and the delivery outcome is:

```
failed  done -> owner operator — no mailbox is configured for this company, and the
operator channel fallback failed: `operator` is not a workflow delivery channel —
this runtime has: engineering, product_design, go_to_market
```

Every `*/5` fire repeats that. Reproduced on two passes, 2026-08-18.

Note the fallback chain is what closes the trap: `owner` with no mailbox falls back to `operator`, and `operator` is exactly the target delivery refuses by name (#981). So on a company in this shape, `owner` is not a working destination at all, while being the default the editor offers.

## Expected behavior

Pick whichever fits the arming model:

- Refuse to arm (400 with the reason) a schedule whose terminal `output` node has no reachable destination — matching the shape #1028 established for a graph with no stage to run; or
- Accept the save but arm it disarmed with a stated reason, the way the disarm rule already reports itself over `workflow_enabled_changed`.

Either way, the operator learns at author time rather than never. Manual runs should stay allowed — a hand-run's result is on screen, which is a real reader.

## Where to look

- `src/company/workflow_create.rs` — the save-time validation stack; the natural home for the check, beside the cron and channel-target gates.
- `src/workflows/delivery.rs` — `post_to_channel` and the `owner` → mailbox → `operator` fallback that makes `owner` undeliverable on a company with no mailbox.
- `src/company/mod.rs` / `src/workflows/runner.rs` — where #1028 put the equivalent stage-less refusal.

## Related

- #976 / PR #1028 — refuse to arm a schedule the graph cannot keep (stage-less graph).
- #981 — the destination picker offers a target delivery refuses by name, and a dropped report still reads as a successful run.
- #947 / #963 — seeded templates whose output nodes declare no destination at all.

M3gA-Mind:
## Summary

Closes tinyhumansai/opencompany#1046.

The sibling of #1028: that one refuses to arm a schedule on a graph that **cannot run**, this one a
graph that **cannot deliver**. Same principle behind both — *saving promises nothing; arming a
schedule promises that something happens* — and the same failure without it: the workflow fires every
N minutes, burns a metered agent turn, and produces nothing anyone receives.

The trap is a fallback, not a typo. `owner` with no mailbox falls back to the `operator` channel, and
`operator` is precisely the target delivery refuses by name (#981). So on a company in that shape
`owner` is not a working destination at all — while being the default the editor offers.

## What #981 already covers, and what was left

Worth stating so nobody re-guards what is guarded. `reject_undeliverable_channel_destinations`
(#981, `server/ops/workflows.rs`) checks `destination.kind == "channel"` at save, against
`deliverable_channel_ids()`. That is all it checks.

The remaining hole is therefore narrower than "output nodes are unvalidated":

* `kind: "owner"` on a company with no mailbox — the #1046 repro.
* `kind: "email"` on a company with no mailbox — `NoMailboxConfigured`; nothing to send *from*.
* `kind: "channel"` on a graph saved **before** #981's gate existed, or whose desk was deleted after
  the graph named it.

## Why this is not an extension of the #1028 guard

My #1028 refusal lives in `set_company_workflow_enabled`, and a reviewer will reasonably ask why this
did not go beside it.

`set_company_workflow_enabled` takes a store and an event log — **not a runtime**. Runnability is a
property of the **graph** (`has_runnable_node()`), so it can be answered there from the parsed file
alone. Deliverability is a property of the **runtime**: which channels this instance wired, and
whether this company has a mailbox. That is reachable only at the server layer, where `ScopedCompany`
carries the runtime handle.

So this sits next to `reject_undeliverable_channel_destinations`, which needs the same handle for the
same reason, rather than threading a new dependency into a core function so it can ask a question it
has no business holding the answer to.

## Knowable at arm time — and the half deliberately left alone

The check refuses only on conditions that are **structural at arm time**: an operator cannot resolve
them without changing how the instance is deployed. A missing mailbox is one. An undeliverable channel
id is another.

**Not refused: `owner` on a company that has a mailbox but no admin address.** Delivery fails there
too (`OwnerFellBackNoAdminAddress`), and it is left alone on purpose — an admin can be added, or a
standing invite can gain an address, at any moment. That answer changes between arm and fire, so
refusing on it would block a legitimate arm on a company that is one invite away from working. **A
guard that blocks a legitimate arm is worse than the silence it replaces.**

### If it becomes deliverable later

The refusal is a `400` on one request. It **stores nothing, disarms nothing, and is not sticky** —
configure a mailbox or fix the destination, send the same `PUT …/enabled` again, and it succeeds.
There is no state to un-stick and no support path to document beyond "try again", which is what the
refusal message says: *"Fix the destination, or configure a mailbox, then switch it on."*

`UNVERIFIED`: whether a newly-configured mailbox becomes visible to a running process without a
runtime rebuild is deployment-dependent (`docs/spec/runtime/rebuild.md` covers the rebuild routes), and
this PR does not change it either way. The guard reads whatever the runtime currently reports.

## Failure direction

Three of the five tests exist for what the guard must **not** do:

* **Pausing is never refused.** Only `enabled = true` is guarded. An operator must always be able to
  stop a thing — the same call the existing unparseable-body rule makes one layer down. A guard that
  trapped a workflow *armed* would be worse than the bug.
* **Manual runs stay possible.** Untouched by construction: this is reached only from the arm path. The
  issue is explicit that a hand-run is survivable, because the banner and the history row both say
  *"1 report not delivered"* — an operator is watching. Nobody is watching a schedule.
* **A graph with no `output` node arms normally.** It promises no report, and may still publish an
  artifact or open a card. "Runs and delivers nothing" is #1028's territory.

Plus: **any deliverable output is enough.** The refusal needs *every* `output` node to be
undeliverable — a graph that mails the owner *and* posts to a desk still delivers when the mailbox is
missing. And a node with no `destination` at all is left to `parse_workflow`, which says something more
specific; reporting the wrong problem first is worse than reporting it second.

## API Or Behavior Changes

`PUT …/workflows/{wid}/enabled` with `{"enabled": true}` now returns **400** for a scheduled graph
whose every `output` node is structurally undeliverable, naming the node and the reason. Saving,
editing, pausing, and manual runs are unchanged, as is arming any graph with a reachable destination.

The check is read-only and runs **before** the write, so a refused arm leaves the record exactly as it
was.

## Tests

      compiles.
- [x] `cargo build --all-targets` — covered by the two test lanes below.
- [x] `cargo test` — both lanes. Default features: `cargo test --locked`, exit 0. Gated:
      **4335 passed / 0 failed**. `RUST_MIN_STACK` matches `ci.yml:626`.

### Revert proof — two

| Reverted | Result |
| --- | --- |
| the `reject_undeliverable_schedule` call | `arming_a_schedule_with_nowhere_to_deliver_is_refused` **FAILED** |
| made it over-fire: guard both toggle directions, drop the schedule check, drop the no-output exemption | **3 FAILED** — `an_unscheduled_graph_with_the_same_destination_still_arms`, `an_undeliverable_schedule_can_still_be_switched_off`, and the pre-existing `the_enabled_route_toggles_a_global_only_workflow` |

The second is the failure-direction proof: each exemption is load-bearing, and removing them together
breaks an existing test as well as two of mine.

`a_scheduled_graph_with_no_output_node_still_arms` survives both reverts — it is a guard against
over-firing in a direction neither revert exercises, and I am not counting it as coverage for the fix.

## Live verification

`N/A` by judgement, not omission. The reported tenant is `smoke1`, which this fleet does not sign in
to. The reachable tenant, `marketing.staging`, is a different company with a different mailbox and
channel configuration, so its state neither confirms nor refutes the report — and the behaviour here is
decided by `runtime.mail()` and `deliverable_channel_ids()`, which are read from the deployment rather
than from any tenant's data. Source is the ground truth for this one. No staging state was read or
changed.

## Documentation

`N/A` — `docs/modules/server/pausing-workflows.md` gained the arming-gate reasoning in #1028 and this
extends the same gate; the per-destination rules live on `reject_undeliverable_schedule`'s doc comment,
beside `reject_undeliverable_channel_destinations` where this file already keeps them.

### DIFF
--- a/src/server/ops/workflows.rs
+++ b/src/server/ops/workflows.rs
@@ -730,6 +730,133 @@ fn reject_undeliverable_channel_destinations(
     Ok(())
 }
 
+/// Refuses to **arm** a schedule whose report has nowhere to land (issue #1046).
+///
+/// The sibling of the stage-less refusal in
+/// [`set_company_workflow_enabled`](crate::company::set_company_workflow_enabled)
+/// (issue #976): that one covers a graph with nothing to *run*, this one a graph
+/// with nowhere to *deliver*. Same principle behind both — saving promises
+/// nothing, arming a schedule promises that something happens — and the same
+/// failure without it: the workflow fires every N minutes, burns a metered agent
+/// turn, and produces nothing anyone receives.
+///
+/// # Why this lives here and not beside the #976 guard
+///
+/// `set_company_workflow_enabled` takes a store and an event log, not a runtime.
+/// Runnability is a property of the **graph**, so it could be answered there from
+/// the parsed file alone. Deliverability is a property of the **runtime** — which
+/// channels this instance wired, whether this company has a mailbox — and that is
+/// reachable only from the server layer, where [`ScopedCompany`] carries the
+/// runtime. So this sits next to
+/// [`reject_undeliverable_channel_destinations`], which needs the same handle for
+/// the same reason, rather than being threaded into a core function that would
+/// have to grow a dependency to ask the question.
+///
+/// # What counts as undeliverable, and what deliberately does not
+///
+/// Refused only for reasons that are **structural at arm time** — conditions an
+/// operator cannot resolve without changing how the instance is deployed:
+///
+/// * **`owner` with no mailbox.** This is the trap in #1046, and it closes
+///   because of a fallback: `owner` with no mailbox falls back to the `operator`
+///   channel, and `operator` is precisely the target delivery refuses by name
+///   (issue #981). So on a company in this shape `owner` is not a working
+///   destination at all — while being the default the editor offers.
+/// * **`email` with no mailbox.** `NoMailboxConfigured` — there is nothing to
+///   send *from*, whatever the address says.
+/// * **`channel` naming something outside
+///   [`deliverable_channel_ids`](crate::company::CompanyRuntime::deliverable_channel_ids).**
+///   Already refused at save by [`reject_undeliverable_channel_destinations`],
+///   but a graph saved before that gate existed still holds one, and a desk can
+///   be deleted after a graph names it.
+///
+/// **Not refused: `owner` on a company that has a mailbox but no admin address.**
+/// Delivery fails there too (`OwnerFellBackNoAdminAddress`), and it is
+/// deliberately left alone: an admin can be added or an invite can gain an
+/// address at any moment, so that answer changes between arm and fire. Refusing
+/// on it would block a legitimate arm on a company that is one invite away from
+/// working — and a guard that blocks a legitimate arm is worse than the silence
+/// it replaces. The mailbox itself is different in kind: it comes from the
+/// deployment's own configuration and cannot appear under a running process.
+///
+/// # Any deliverable output is enough
+///
+/// The refusal needs **every** `output` node to be undeliverable. A graph that
+/// mails the owner *and* posts to a desk still delivers when the mailbox is
+/// missing, and refusing it would be wrong.
+///
+/// A graph with **no** `output` node at all is left alone: it promises no report,
+/// and it may still do real work — publish an artifact, write a workspace note,
+/// open a card. "Runs and delivers nothing" is #976's territory, not this one.
+///
+/// Manual runs are untouched by construction: this is reached only from the arm
+/// path. The issue is explicit that a hand-run is survivable because the banner
+/// and the history row both say "1 report not delivered" — an operator is
+/// watching. Nobody is watching a schedule.
+fn reject_undeliverable_schedule(
+    company: &ScopedCompany,
+    file: &crate::company::WorkflowFile,
+) -> Result<(), ApiError> {
+    if file.trigger_schedule().is_none() {
+        return Ok(());
+    }
+    let outputs: Vec<_> = file
+        .nodes
+        .iter()
+        .filter(|node| node.kind == crate::company::WorkflowNodeKind::Output)
+        .collect();
+    if outputs.is_empty() {
+        return Ok(());
+    }
+    let has_mailbox = company.runtime.mail().is_some();
+    let deliverable = company.runtime.deliverable_channel_ids();
+
+    let mut why: Option<String> = None;
+    for node in &outputs {
+        let Some(destination) = &node.destination else {
+            // No destination named. `parse_workflow` has its own, more specific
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:

oxoxDev:
Superseded by #1083, which merged the same guard — a schedule is refused when its report has nowhere to land. That version threads the check through `company/workflow_create.rs` and `company/workflow_file.rs` as well as `server/ops/workflows.rs`, so it covers the create-time path too.

Closing this as duplicate. Thanks for catching the same gap.

PR #1081, opened 2026-08-18
fix(workflows): don't bury a run that finishes while list_runs is reading

### LINKED ISSUE #1009: workflows: a run can hang in running forever when its finish is never journaled

graycyrus:
Follow-up to #1039, which closed #1009. That PR is the fix; this repairs two defects in the read-side settle it added. Credit to #1039 for the design — the `live()` cross-check, the synthetic `INTERRUPTED_BY_RESTART`, the watchdog and the visibility-gated poll are all its work and are untouched here.

Both were found while reviewing a duplicate attempt at #1009 that I abandoned once #1039 landed. Only these two carry over.

## 1. A run that settles inside the request is treated as dead

`live()` is consulted *after* the journal snapshot is taken. A run that appends its finish too late for the snapshot and drops its guard in time to be missing from `live()` matches the dead predicate exactly — start with no finish, id absent from the live set.

**Order is what makes this harmful rather than a harmless duplicate.** The rebuild false positive the cross-check knowingly accepts is self-correcting, and the existing comment says why: that run's truthful finish lands *after* the synthetic one, and the fold settles an entry from the last finish it sees. Here the truthful finish lands **first** and loses. So:

- the run reads `INTERRUPTED_BY_RESTART` permanently, though it succeeded;
- and because the fold overwrites `deliveries` from whichever finish settles last, the record of what it actually sent is replaced by the synthetic row's empty list.

A successful run, and the evidence that it delivered, silently gone — and nothing about it is discoverable by reading the history afterwards.

The settle now re-reads the journal from where the snapshot stopped and drops any candidate whose finish turns up there. This is exact rather than a heuristic: a run whose start was in the snapshot was registered before it (`begin` precedes both the spawn and the runner's `WorkflowRunStarted`), so a candidate missing from `live()` has already been deregistered — and a deregistered run journaled its finish first, if it was ever going to. Anything appended by that point sits at a higher sequence than the entire snapshot, so the second read cannot miss it. It runs only when there are candidates at all — after the first settle, nothing — and reads only the tail.

I also qualified the existing comment, whose supersession argument is correct for the rebuild case it was written about but inverts here.

## 2. A settled row changes identity between reads

The settle flipped `running` and `error` but left the row carrying the **start's** `seq` and `at_millis`, while the next read's fold takes both from the appended **finish**. So one poll served seq N and the next served seq N+1 for the same run.

Not cosmetic: `RunHistoryPanel.tsx:213` keys rows on `key={run.seq}`, and lines 215/218/220 compare that field against `selectedRunSeq`, `fixingRunSeq` and `fixReason.seq`. The row remounts and loses its selection one poll after being settled — in the window where it is most likely to be noticed, since the 2s recovery poll runs precisely because somebody is watching that run.

The appended rows are now read back and their durable `seq` / `at_millis` stamped on. Read back rather than returned from `record_run_finished`, which reports only whether the append happened; the values served are then the durable ones rather than a second construction of them.

## Deliberately not included

A third difference between #1039's approach and the duplicate I abandoned: #1039 uses `catch_unwind` inside the run task, which covers a panic but not an abort, whereas a `JoinHandle` watchdog would cover both. **I judged that little practical value and left it out** rather than smuggling it in — nothing in-crate aborts a workflow run handle (the run task's handle has no external holder, and an operator's Stop goes through `RunSupervisor::cancel`, which fires the run's own signal and lets it wind down and journal normally). Raising it here for the record, not proposing it.

## Tests

Both are regression tests, and I verified each fails without its fix rather than assuming:

- `a_run_that_settles_during_the_read_is_not_buried_by_a_synthetic_finish` — uses an `EventLog` that appends the run's real finish *after* handing back the snapshot, which is the only way to hit the interleaving deterministically. Without the guard: **2 finishes journaled**, and the run reads `"error": "this run was interrupted by a host restart…"` with `"deliveries": []`. With it: one finish, and the next read shows the run succeeded with its delivery intact.
  - The interposer is armed *after* `build()` on purpose. Arming it earlier lets the one-shot fire inside `RuntimeBuilder::build`, whose own journal read runs `sweep_interrupted_runs` — the two finishes that produces are the boot sweep's doing, not this route's. My first draft had exactly that bug and passed for the wrong reason.
- `a_settled_dead_run_keeps_its_seq_and_time_across_reads` — two consecutive GETs must agree on `seq` and `atMillis`. Without the fix: `left: Number(3), right: Number(4)`.

## Commands run

```
cargo fmt --all -- --check                                          # clean
cargo clippy --all-targets --features openhuman,tinycortex -- -D warnings   # no warnings
cargo test --features openhuman,tinycortex --tests                  # 4323 + 10 + 1 + 11 + 2 passed, 0 failed
scripts/ci/assert-feature-lanes.sh                                  # 29 features, 0 owed a lane
scripts/ci/assert-md-line-cap.sh                                    # all files under the cap
```

One file changed. No frontend change — the console behaviour this restores is the existing `RunHistoryPanel` keying, which is now honoured rather than altered.

### DIFF
--- a/src/server/ops/workflows.rs
+++ b/src/server/ops/workflows.rs
@@ -2446,10 +2446,16 @@ async fn list_runs(
     // not "whichever of the last N happen to match".
     let wanted = query.workflow.as_deref();
     let matches = |workflow_id: &str| wanted.is_none_or(|w| w == workflow_id);
+    // The high-water mark of the snapshot above, kept over EVERY row rather than
+    // only the matched ones. It is where the settle below resumes reading, which
+    // is what tells a run that died apart from one that merely finished while
+    // this request was folding.
+    let mut read_through = 0u64;
 
     for stored in stored {
         let seq = stored.seq.value();
         let at_millis = stored.at_millis;
+        read_through = read_through.max(seq);
         match stored.event {
             CompanyEvent::WorkflowRunStarted {
                 workflow_id,
@@ -2620,7 +2626,12 @@ async fn list_runs(
     // order and wins the read's last-writer-wins display, and (iii) it is the
     // same class the boot sweep already accepts — which is why that sweep gates
     // on the handover being absent (see the runtime builder call site). It never
-    // corrupts the journal: a second, truthful finish simply supersedes it.
+    // corrupts the journal: that run's second, truthful finish lands *after* the
+    // synthetic one and so supersedes it.
+    //
+    // That last argument turns on ORDER, and it does not carry to a run which
+    // settles inside this request — there the truthful finish lands first and
+    // loses. See the window handled below; it is closed rather than accepted.
     let live_ids: HashSet<String> = company
         .runtime
         .run_supervisor()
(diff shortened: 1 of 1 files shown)


graycyrus:
The CodeRabbit check shows `pass` but its body is a rate-limit notice ("Next review available in: 12 minutes") — no review has run on this PR, so that green tick is not a signal. Re-triggering once the window resets.

For the record, CI here is bound to the actual head SHA `d918d924` (verified via the `head_sha` API rather than the PR check rollup — on the superseded #1055, GitHub silently failed to fire `pull_request: synchronize` for two pushes and the rollup kept showing green results belonging to a two-commit-stale SHA).

graycyrus:
@coderabbitai review

Issue #371, opened 2026-08-05
workflows: a running workflow shows no progress until it finishes


oxoxDev:
## What the operator sees

Press **Run** on a workflow and the console goes opaque. The button spins, nothing else changes, and some seconds or minutes later a drawer appears with delivery rows and a nested `output` blob. Between those two moments there is no signal at all: which node is executing, which have finished, whether the run is progressing or wedged.

A long run is indistinguishable from a hung one. A run that fails at the fourth of six nodes reports only that it failed — the operator cannot see how far it got without reading the output blob, and a non-technical operator cannot read it at all.

This also means a workflow run cannot be demonstrated. There is nothing on screen to show.

## Why it happens

The engine reports once, at the end.

- The run is a single blocking POST (`runWorkflow`, `frontend/src/api/workflows.ts:224`), and the console awaits it with a boolean `running` flag (`frontend/src/views/WorkflowsView.tsx:95,242-269`).
- The journal receives exactly one event for the whole run: `CompanyEvent::WorkflowRunFinished` (`src/ports/types.rs:582`). There is no node-level start or finish event, and `grep` finds one producer.
- `WorkflowRunOutcome` (`frontend/src/api/workflows.ts:189-205`) accordingly carries `deliveries`, `pendingApprovals` and an optional `error` — no per-node rows, no timings.

This is a deliberate boundary rather than an oversight: `src/ports/types.rs:569-570` records that the workflow runner "has no task and no attempt", placing it outside the run records introduced by #242. Task dispatch got durable per-attempt step traces; workflow runs did not.

## What the fix should do

Give a run the same legibility a task attempt already has:

1. Emit a node-level progress signal as the engine walks the graph — at minimum node started and node finished, with the outcome and elapsed time. The existing company event stream (SSE) is the natural carrier, since the console already consumes it.
2. Render it on the canvas: mark the executing node, mark completed ones, and leave the finished states visible after the run so the drawer explains *where* a failure happened rather than only *that* it did.
3. Persist enough of it that a run nobody watched — a scheduled one — can still be inspected afterwards. Today `WorkflowRunFinished` is all that survives, and for a cron run that reaches only the host's stdout, which on a hosted tenant is the platform team rather than the operator (the same gap #228 closed for run outcomes).

[...]

Issue #1110, opened 2026-08-19
workflows: the tab opens inside a workflow instead of showing the list first


graycyrus:
Opening the Workflows tab drops you *inside* a workflow you did not choose. The list of all workflows is a secondary thing behind a **Browse** toggle, and the toolbar above it — a workflow selector, Run, Test run, Copilot, History, Edit, Delete — belongs to whichever workflow got auto-selected.

That inverts the natural order. The tab should answer "what workflows do I have?" first, and "what about this one?" only once you have picked one.

## Proposed flow

1. **`#/workflows`** — the index. Every workflow, with the existing Cards/List toggle, the count, and **New workflow**. No workflow is selected, and none of the per-workflow controls are on screen, because none of them have a subject yet.
2. **Click a row/card → `#/workflows/<id>`** — the detail view for that workflow: its canvas, and its own controls (Run, Test run, Copilot, History, Edit, Delete), plus a clear way back to the index.

## Why this is mostly re-wiring, not new machinery

The pieces already exist:

- `#/workflows/<id>` hash routes are **already written** — `frontend/src/views/WorkflowsView.tsx:586` pushes `#/workflows/${encodeURIComponent(selectedId)}` whenever the selection changes, and `:536` reads a requested id back out.
- `WorkflowIndex` (`frontend/src/views/workflows/WorkflowIndex.tsx:34`) already renders the full list with Cards/List modes and an `onSelect(id)` callback.
- The index is currently gated behind `indexOpen` state (`WorkflowsView.tsx:353`), toggled by the Browse button, rather than being the default view.

What changes is which view is the default and what happens when nothing is selected:

- **No id in the hash → render the index**, and do not auto-select anything. Today the view auto-selects (`WorkflowsView.tsx:462`, `:536`), which is what makes the tab open inside a workflow.
- **An id in the hash → render the detail view** for it.
- The **Browse** button becomes redundant — it is now just "go back to the index". Decide whether it becomes a Back affordance or disappears.

## Decisions to make deliberately

- **Back navigation.** The console routes on the hash, so browser Back must move index ↔ detail correctly, and a detail URL must be shareable and survive a reload.
- **An id in the hash that no longer exists** (deleted workflow, wrong company) should land on the index with a plain explanation, not an empty detail shell.
- **After creating a workflow**, does the operator land on the new workflow's detail page or back on the index? Detail is probably right — they made it in order to work on it.
[...]

PR #1111, opened 2026-08-19
workflows: run history as a left rail beside the canvas

### LINKED ISSUE #1107: workflows: run history is a bottom strip where a left vertical rail fits the content better
Run history opens as a horizontal strip under the canvas. It should be a vertical rail down the left instead.

## Why the current shape fights the content

- The panel is a bottom band — root is `border-t bg-card/60` with a `max-h-72 overflow-auto` list (`frontend/src/views/workflows/RunHistoryPanel.tsx:188`, `:203`) — mounted as a sibling *below* the canvas (`frontend/src/views/WorkflowsView.tsx:1942`).
- Run rows are tall and stack vertically, so a horizontal strip shows very few at once and wastes the width it takes.
- It eats canvas height, which is the dimension a workflow graph needs most: opening history to inspect a run shrinks the very graph the run is being overlaid onto (selecting a row overlays that run's per-node states on the canvas, issue #371).
- Screens are wider than they are tall. A vertical rail costs width, which the canvas has to spare.

## Proposed change

Move the panel to a fixed-width vertical rail on the **left**, full height of the workflows view:

- Panel chrome: `border-r` instead of `border-t`, a fixed width (~`w-80`), and the list growing to fill the rail (`flex-1 overflow-auto`) rather than capped at `max-h-72`.
- Parent layout: canvas and rail side by side in a row, rather than stacked.
- Keep every existing behaviour: row click toggles the canvas overlay, Fix-with-copilot with its `anyFixInFlight` guard, the header count badge, and Dismiss.

## Worth deciding while implementing

- **Narrow viewports.** A fixed rail plus a canvas is bad below roughly tablet width. Either collapse the rail to the existing bottom-strip behaviour under a breakpoint, or overlay it — but do not let it squeeze the canvas to nothing.
- **`RunResultPanel`** currently mounts as a sibling in the same stack (`WorkflowsView.tsx:1933`). Decide whether it stays where it is or joins the rail, and say why; two competing panels in one column would be worse than what we have now.

graycyrus:
Closes #1107

Run history opened as a horizontal strip under the canvas. Run rows are tall, so the strip showed **two runs at a time** while spending the full width of the screen — and it ate the canvas height a workflow graph needs most, including the very canvas a selected run is overlaid onto (#371).

It is now a **320px in-flow rail down the left**, full height of the canvas region, with the list growing to fill it instead of capping at `max-h-72`. At 1440×900 that is four runs visible instead of two, in the same screen.

## What changed

- `RunHistoryPanel` root: `border-t bg-card/60` → `flex h-full flex-col border-t bg-card/60 xl:border-t-0 xl:border-r`; list `max-h-72 overflow-auto` → `max-h-72 overflow-auto xl:min-h-0 xl:max-h-none xl:flex-1`.
- Header gained a `border-b` and `flex-wrap`, so the workflow name drops to its own line at 320px and stays inline where there is room. No breakpoint needed for it.
- The root is now an `<aside aria-label="Run history">`: at `xl` the rail is painted *left* of a canvas it *follows* in the DOM, so it is reachable as a named complementary landmark rather than only by tabbing past the graph. Keeping the canvas first in the DOM is what leaves the narrow reading order exactly as it was.
- New `frontend/src/views/workflows/CanvasShell.tsx` — see "the seam" below.
- New `frontend/test/e2e/workflow-run-history-rail.spec.ts` — pins the geometry, not the class names.

Behaviour is unchanged: row click toggles the canvas overlay (#371), Fix with copilot keeps its `anyFixInFlight` guard disabling *every* row's button while one fix is in flight, the header count badge and Dismiss are where they were.

## Judgement call 1 — narrow viewports: fall back to the strip, at `xl`

Below `xl` (1280px viewport) the rail renders exactly as it does today: a full-width strip under the canvas, capped at `max-h-72`. Nothing new to learn on a tablet, and no risk of the canvas being squeezed.

`xl` rather than `lg` because of the arithmetic below. The app's own nav sidebar is 216px, so a 1024px viewport leaves the view 808px — a 320px rail there would leave 488px of canvas *before* a right-hand overlay covers 396 more of it. At 1280 the numbers work; at 1024 they do not.

## Judgement call 2 — the arithmetic, measured on the real app

Measured on the running host, `history + copilot` both open:

| viewport | view width | rail | canvas | right overlay | canvas still visible |
|---|---|---|---|---|---|
| 1440×900 | 1224 | 320 | **904** | 384 @ x=1044 | **508px** |
| 1280×800 | 1064 | 320 | **744** | 384 @ x=884 | **348px** |
| 1024×800 | 808 | — (strip) | 808 | — | 808px |

The asymmetry is deliberate, and it is the reason those numbers are acceptable: **the left rail is in-flow and costs the canvas real width; the right panels float and only cover it.** That maps onto how they are used. History is a browsing context you open on purpose and keep open while you work, so it should sit *beside* the graph. `CopilotPanel` and `NodeDetailPanel` are a transient focus you dismiss, so the graph keeps its full width underneath and closing the panel restores it instantly — at 1280 the canvas is still 744px, it is just partly occluded.

**No new collapse control**, deliberately. The obvious third option was a collapsible rail, but the control already exists: the toolbar's `History` toggle, one click away, carrying the run count. Closing history does **not** clear `overlayRun` — the canvas overlay and its banner survive — so dismissing the rail is lossless. A second collapse affordance would be the same button twice.

## Judgement call 3 — `RunResultPanel` stays at the bottom

It answers a different question with a different lifetime. `RunResultPanel` (and `RunFailurePanel`, its mutually-exclusive sibling) is the receipt for **the run you just started**: it appears unbidden, you read it, you dismiss it. The rail is **what has run before**: you open it and keep it. Moving the result into the rail would put two panels in one column — the failure mode #1107 explicitly rules out — and would make the rail's meaning ambiguous. Full width under *both* canvas and rail keeps them separate.

## The seam for what comes next

`CanvasShell` is a ~20-line layout component whose header comment is the actual deliverable: it names three slots and says what decides which one a new panel belongs in — **lifetime, not where there happens to be room**.

- **left** — in-flow rail, single occupancy, `xl`+. Shrinks the canvas because it is meant to be read alongside it.
- **right** — floating overlay, `absolute right-3 top-3 bottom-3 z-10` inside the canvas. Single occupancy, arbitrated by `WorkflowsView`.
- **bottom** — full-width strip, a sibling of `CanvasShell`.

A new left rail passes `rail`; a new right panel mounts absolutely inside the canvas and inherits the anchoring for free (`CanvasShell` keeps `relative` on the canvas wrapper, so the overlays now anchor to the canvas's right edge, which has moved — verified at 1440 and 1280 above). Not a panel framework: one component and a documented convention.

**On the "right slot is double-booked" concern:** checked, and it is not. `WorkflowsView` renders them as a ternary — `copilotOpen && graph ? <CopilotPanel/> : selectedNode && <NodeDetailPanel/>` — and opening the copilot clears `selectedNodeId`. They are mutually exclusive by construction. Nothing here changes that, and the new layout does not depend on it.

## Verification

Run locally, real output:

- `npm run typecheck` — clean
- `npm test` — **1044 passed, 105 files**
- `scripts/ci/assert-design-tokens.sh` — `✓ design tokens: no arbitrary sizes, no palette colours, no stray hex`
- Playwright, against a host built from this branch (`workflow-run-history-rail`, `workflow-run-history`, `workflow-selection-persists`, `workflow-run-result`, `workflow-toolbar-reachable`) — **9 passed, 2 skipped** (the two need a live brain)
- Playwright, the rest of the workflow specs (`workflow-canvas-live`, `workflow-create-affordance`, `workflow-dialog-feedback`, `workflow-edit-delete`, `workflow-live-list`, `workflow-node-config`, `workflow-inline-approval`) — **31 passed, 2 skipped**

Screenshotted at 1440×900 (light and dark, with and without the copilot), 1280×800 with the copilot, 1024×800 and 820×900. Dark mode is token-only, so `.dark` picks it up with no `dark:` pairs added.

## Summary by CodeRabbit

* **New Features**
  * Workflow details now open consistently when selecting or creating a workflow, with an “All workflows” control to return to the index.
  * Added Cards/List switching on the workflow index.
  * Invalid or missing workflow links now return safely to the index with a dismissible alert.
  * Workflow run history appears in a dedicated side rail on wide screens and below the canvas on smaller screens.
  * Blocked runs now display related approval information.

* **Tests**
  * Added end-to-end coverage for responsive history layouts and index/detail navigation.

### DIFF
--- a/frontend/src/views/WorkflowsView.tsx
+++ b/frontend/src/views/WorkflowsView.tsx
@@ -101,6 +101,7 @@ import { classifyRunError } from "@/views/workflows/run-error";
 import { runFailureFrom, type RunFailure } from "@/views/workflows/run-failure";
 import { RunFailurePanel } from "@/views/workflows/RunFailurePanel";
 import { RunResultPanel } from "@/views/workflows/RunResultPanel";
+import { CanvasShell } from "@/views/workflows/CanvasShell";
 import { approvalsForRun } from "@/views/workflows/run-approvals";
 import { NodeDetailPanel } from "@/views/workflows/NodeDetailPanel";
 import { type NodeOutputView, nodeOutputFor } from "@/views/workflows/run-output";
@@ -2133,7 +2134,30 @@ export function WorkflowsView({
         </div>
       )}
 
-      <div className="relative flex-1">
+      {/* Issue #1107: canvas and left rail side by side. `CanvasShell` owns the
+          view's column layout and documents which slot a new panel belongs in —
+          left rail, right overlay, or bottom strip. */}
+      <CanvasShell
+        rail={
+          historyOpen && historySupported ? (
+            <RunHistoryPanel
+              runs={historyRows}
+              graph={graph}
+              workflowName={selected?.name ?? selectedId ?? ""}
+              onClose={() => setHistoryOpen(false)}
+              selectedRunSeq={overlayRun?.seq ?? null}
+              onSelectRun={(picked) =>
+                // Clicking the row already shown clears it, so the control is a
+                // toggle rather than a one-way trip into overlay mode.
+                setOverlayRun((prev) => (prev?.seq === picked.seq ? null : picked))
+              }
+              onFixWithCopilot={handleFixWithCopilot}
+              fixingRunSeq={fixingRunSeq}
+              fixReason={fixReason}
+            />
+          ) : null
+        }
+      >
         {/* Issue #303: browsing REPLACES the canvas rather than squeezing in
             above it. A card grid needs the width, and the canvas is meaningless
             while the operator is deciding which workflow they want. Picking one
@@ -2257,7 +2281,7 @@ export function WorkflowsView({
             )}
           </>
         )}
-      </div>
+      </CanvasShell>
 
       {result && (
         <RunResultPanel
@@ -2286,24 +2310,6 @@ export function WorkflowsView({
         <RunFailurePanel
           failure={runFailure}
           onClose={() => setRunFailure(null)}
-        />
-      )}
-
-      {historyOpen && historySupported && (
-        <RunHistoryPanel
-          runs={historyRows}
-          graph={graph}
-          workflowName={selected?.name ?? selectedId ?? ""}
-          onClose={() => setHistoryOpen(false)}
-          selectedRunSeq={overlayRun?.seq ?? null}
-          onSelectRun={(picked) =>
-            // Clicking the row already shown clears it, so the control is a
-            // toggle rather than a one-way trip into overlay mode.
-            setOverlayRun((prev) => (prev?.seq === picked.seq ? null : picked))
-          }
-          onFixWithCopilot={handleFixWithCopilot}
-          fixingRunSeq={fixingRunSeq}
-          fixReason={fixReason}
         />
       )}
 
--- a/frontend/src/views/workflows/CanvasShell.tsx
+++ b/frontend/src/views/workflows/CanvasShell.tsx
@@ -1,0 +1,64 @@
+// The workflows view's column layout — and the convention for where a panel
+// mounts (issue #1107).
+//
+// Before this, every panel the view grew was appended to one vertical stack
+// under the canvas, and each one ate the dimension a workflow graph needs most.
+// Run history was the worst case: tall rows in a `max-h-72` horizontal strip,
+// showing two runs at a time while spending the full width of the screen.
+//
+// The view now has three slots, and each one is a decision about the panel's
+// LIFETIME, not about where there happens to be room:
+//
+//   left   — an in-flow rail. A browsing context the operator opened on
+//            purpose and keeps open while they work: it shrinks the canvas,
+//            because it is meant to be read alongside it. Run history is the
+//            only occupant today. Single occupancy — a second rail must
+//            replace this one rather than stack under it, the same way the
+//            right slot already arbitrates between its two panels.
+//
+//   right  — a floating overlay, mounted `absolute right-3 top-3 bottom-3
+//            z-10` INSIDE the canvas (see `CopilotPanel` and
+//            `NodeDetailPanel`). A transient focus that covers the graph and
+//            is dismissed: the canvas keeps its full width underneath, so
+//            closing the panel restores it instantly. Also single occupancy,
+//            and `WorkflowsView` arbitrates it with a ternary — opening the
+//            copilot clears the selected node.
+//
+//   bottom — a full-width strip below BOTH canvas and rail, mounted as a
+//            sibling of this component. The receipt for something that just
+//            happened and is dismissed when read: `RunResultPanel` and
+//            `RunFailurePanel`. Full width because it spans the whole view,
+//            and below the rail because the rail answers "what has run
+//            before" while the strip answers "what did the run I just started
+//            do" — two questions, two places.
+//
+// The arithmetic that fixes the breakpoint: the rail is in-flow, so it costs
+// the canvas real width, while the right overlays only cover it. A 320px rail
+// plus a 384px overlay is 704px of chrome, which is most of a laptop window
+// once the app's own nav sidebar is subtracted. So the rail is in-flow only at
+// `xl` (≥1280px viewport, ≥~1024px of view) — below that it falls back to the
+// bottom strip it has always been, which never competes with the right slot.
+
+import type { ReactNode } from "react";
+
+export function CanvasShell({
+  rail,
+  children,
+}: {
+  /**
+   * The left rail, or nothing. Rendered AFTER the canvas in the DOM and moved
+   * left with `order` at `xl`: keeping the canvas first means the narrow
+   * layout's reading order is unchanged, and the rail names itself as a
+   * landmark so assistive tech can reach it directly either way.
+   */
+  rail?: ReactNode;
+  /** The canvas. Positioned, because the right slot mounts inside it. */
+  children: ReactNode;
+}) {
+  return (
+    <div className="flex min-h-0 flex-1 flex-col xl:flex-row">
+      <div className="relative min-h-0 flex-1">{children}</div>
+      {rail && <div className="shrink-0 xl:order-first xl:w-80">{rail}</div>}
+    </div>
+  );
+}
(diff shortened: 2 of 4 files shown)


graycyrus:
## Merged `main` — and the approval above predates it

**The approval on this PR was given on the pre-merge diff.** `68c9995c` is a merge of `upstream/main` that changes where the rail mounts, so what is on the branch now is not what was reviewed. Flagging it rather than treating the green tick as still standing — a re-look at the merge commit alone is enough, the rest of the diff is untouched.

I **merged** rather than rebased: the branch is one commit, the conflict is one hunk of `WorkflowsView.tsx`, and a merge commit leaves the approved commit addressable so the delta since the review is exactly `68c9995c`.

### What conflicted, and how the two designs join up

#1116 (#1110) landed while this sat: `#/workflows` is now the index and `#/workflows/<id>` the detail view. It had already given the run-history drawer a `detailOpen &&` guard in the bottom stack.

That guard is the whole answer to where the rail goes. Run history is per-workflow chrome — a list of workflows has no single run to show history for — so **`CanvasShell` now wraps the DETAIL branch only**, not the view body:

```jsx
{!detailOpen ? (
  <div className="relative flex-1">…index…</div>
) : (
  <CanvasShell rail={historyOpen && historySupported ? <RunHistoryPanel …/> : null}>
    …canvas + right overlays…
  </CanvasShell>
)}
```

The index keeps the plain column it has on `main`. It cannot grow run chrome by accident, because the slot a rail would mount in does not exist on that side of the branch — structural rather than a rule each panel has to remember. `CanvasShell`'s header comment now says so in its first paragraph, since "which slot does my panel go in" is the thing that file exists to answer.

### The arithmetic, re-measured on the merged tree

Re-checked rather than assumed, because #1116 reshaped the surrounding layout. Measured on the running host at `68c9995c`, rail **and** `CopilotPanel` both open:

| viewport | view width | rail | canvas | right overlay | canvas still visible |
|---|---|---|---|---|---|
| 1440×900 | 1224 | 320 @ x=216 | **904** @ x=536 | 384 @ x=1044 | **508px** |
| 1280×900 | 1064 | 320 @ x=216 | **744** @ x=536 | 384 @ x=884 | **348px** |

**Unchanged from the pre-merge numbers, and that is the finding.** #1116's changes to the detail view are all vertical — the back-button heading and the one-workflow toolbar are rows above the canvas, and neither the nav sidebar (216px) nor the overlays (`w-96` = 384px at `sm`+) moved. So no adjustment was needed and none was made; had I changed the breakpoint or the rail width off the back of a merge, that would have been a second design decision smuggled into a conflict fix.

The asymmetry the original PR argued still holds and is what makes 348px acceptable: the rail is in-flow and **costs** the canvas width, the right panels float and only **cover** it. At 1280 the graph is still 744px wide underneath and closing the copilot restores it instantly.

### Tests, on the merged tree

`workflow-run-history-rail.spec.ts` was written before #1116 and went straight to `/#/workflows` expecting a canvas. It now opens a workflow first through #1116's own `openFirstWorkflow` helper, and gained a third case for the other half of the design:

> `the index carries no run chrome, and leaving a workflow takes the rail with it` — no History toggle and no `workflow-run-history` on the index; open a workflow, raise the rail, click back, and neither is there again.

Real output, all against a host built from `68c9995c` on a dedicated port (not the default bind, and not Playwright's `webServer` — it reuses an existing server, which would have tested somebody else's build):

- `npm run typecheck` — clean
- `npm test` — **1120 passed, 115 files**
- `scripts/ci/assert-design-tokens.sh` — `✓ design tokens: no arbitrary sizes, no palette colours, no stray hex`
- Playwright, the specs #1116 touched plus this one (`workflow-run-history-rail`, `workflow-run-history`, `workflow-selection-persists`, `workflow-run-result`, `workflow-toolbar-reachable`) — **10 passed, 2 skipped** (the two need a live brain)
- Playwright, the rest of the workflow suite (`workflow-canvas-live`, `workflow-create-affordance`, `workflow-dialog-feedback`, `workflow-edit-delete`, `workflow-live-list`, `workflow-node-config`, `workflow-inline-approval`, `workflow-run-inference`) — **32 passed, 2 skipped**

Screenshotted on the merged build in light and dark at 1440 and 1280: the detail view with the rail up, the same with the copilot over it, and the index — which shows the `Workflows` count, Cards/List and `New workflow`, and nothing about runs.

graycyrus:
### CI on `68c9995c`: everything this PR can reach is green; the one red lane is #1127

Checked by head SHA rather than through `gh pr checks`, which has been reporting green for **cancelled** runs today:

```
$ gh api "repos/tinyhumansai/opencompany/actions/runs?head_sha=68c9995c…" \
    --jq '[.workflow_runs[]|select(.name=="CI")][0].conclusion'
failure
```

Per job:

| job | result |
|---|---|
| Changed areas | ✅ success |
| Actions pinned | ✅ success |
| Console | ✅ success |
| Console (current Node, advisory) | ✅ success |
| Gated host binary | ✅ success |
| Rust | ✅ success |
| **Console E2E** | ✅ **success** |
| **Console E2E (live brain)** | ✅ **success** |
| Rust (mongodb) | – skipped |
| Rust (openhuman, tinycortex) | – skipped |
| **Desktop** | ❌ **failure** |

`Desktop` is **#1127**, and it is not reachable from this diff:

```
test commands::test::an_instance_row_answers_in_the_keys_the_console_reads ... FAILED
  left: ["id", "label", "dataDir", "running", "baseUrl", "instanceId", "operatorEmail", "companies"]
 right: ["baseUrl", "companies", "dataDir", "id", "instanceId", "label", "operatorEmail", "running"]
test result: FAILED. 93 passed; 1 failed
```

That is the key-order assertion #1127 describes — struct order against alphabetical, from `preserve_order` (#985) and the assertion (#1096) merging 19 seconds apart. #1127 records it as red **on `main` and on every PR since 11:26Z**, so it arrived here with the merge.

This branch contains no Rust and no Cargo manifest at all:

```
$ git diff --name-only upstream/main…HEAD
frontend/src/views/WorkflowsView.tsx
frontend/src/views/workflows/CanvasShell.tsx
frontend/src/views/workflows/RunHistoryPanel.tsx
frontend/test/e2e/workflow-run-history-rail.spec.ts
```

Deliberately **not** fixing it here. The fix belongs on `main` where it is tracked — folding it into a layout PR would hide a repo-wide breakage inside an unrelated diff and make this branch's green tick mean something it should not.

Both `Console E2E` lanes passing is the result worth reading: #1116 rewrote most of the workflow specs and this PR moved where the rail mounts, so those two lanes are the ones that would have caught a bad merge. `Not merging` — over to a reviewer, with the caveat in the comment above that the approval predates `68c9995c`.

PR #1112, opened 2026-08-19
console: stop the typography plugin drawing backticks around inline code

### LINKED ISSUE #1108: console: inline code renders with literal backticks, because the typography plugin injects them
Inline code in chat renders as **`` `UsageDto` ``** — with the backtick characters visible — instead of `UsageDto`. It reads as though markdown was never parsed, but it is: the markdown IS rendering, and then literal backticks are being drawn around every inline code span.

## Cause

`frontend/src/components/markdown.tsx:46` wraps output in `prose prose-sm max-w-none dark:prose-invert`, and `frontend/src/index.css:10` loads `@plugin "@tailwindcss/typography"`.

`@tailwindcss/typography` ships this by default:

```css
.prose :where(code):not(:where([class~="not-prose"] *))::before { content: "`" }
.prose :where(code):not(:where([class~="not-prose"] *))::after  { content: "`" }
```

Nothing in `index.css` overrides it — `grep` for `code::before` finds no override. So every inline code span gets a literal backtick before and after, which is exactly what the screenshot shows: backticks visible, content in mono.

## Scope — this is not a chat bug

Every surface using the shared `Markdown` component is affected: `views/chat/MessageRow.tsx`, `views/chat/ThreadPanel.tsx`, `views/Conversation.tsx`, `views/MemoryView.tsx`, `views/LedgersView.tsx`, `views/ArtifactsTab.tsx`, `views/workflows/CopilotPanel.tsx`, `views/workflows/NodeDetailPanel.tsx`, `views/workflows/RunResultPanel.tsx`, plus `ApprovalsView` and `WorkspaceView`.

Agent messages quote type names, fields and file paths constantly, so chat is simply where it is most visible.

## Fix

Null the pseudo-element content in `index.css`, alongside the existing prose customisations:

```css
.prose :where(code):not(:where([class~="not-prose"] *))::before,
.prose :where(code):not(:where([class~="not-prose"] *))::after { content: none; }
```

While there: inline `code` inside `prose` has no background or padding of its own in this theme, so it is distinguished only by font. Giving it a token-based background and small padding would make it read as code rather than as oddly-fonted prose. Use existing semantic tokens — `assert-design-tokens.sh` forbids raw hex and raw Tailwind palette classes.

Worth a test: a rendering assertion that inline code produces no literal backtick, so the plugin's default cannot creep back on upgrade.

graycyrus:
## Summary

Inline code rendered as `` `UsageDto` `` — backticks visible — on every surface
that uses the shared `Markdown` component. The markdown *was* parsed; the
backticks came back afterwards, from CSS.

`@tailwindcss/typography` is written for hand-authored HTML, where nothing else
marks a `code` span, so it ships:

```css
.prose :where(code)…::before { content: "`" }
.prose :where(code)…::after  { content: "`" }
```

`frontend/src/components/markdown.tsx` feeds the plugin the *output* of
`react-markdown`, where the fences were already consumed — so the plugin drew
them back on. Chat is where it showed most (agents quote type names, flags and
paths constantly), but it hit Conversation, MemoryView, LedgersView,
ArtifactsTab, CopilotPanel, NodeDetailPanel, RunResultPanel, ApprovalsView and
WorkspaceView equally.

## What changed

**`frontend/src/index.css`** — one new section, at the end of the file, next to
the other prose-adjacent rules:

- `content: none` on `code::before`/`::after`.
- Inline code gets a ground of its own: `--muted`, a hairline `--border`,
  `calc(var(--radius) * 0.6)` and `0.125em 0.375em` of padding. Without the
  fences, at `0.875em`, the mono face alone was a weak signal in a chat line.
  Tokens only — `.dark` re-declares both, so dark mode comes for free.

Two things the rules depend on, both explained in comments and both covered by
the test:

- **They stay unlayered.** `prose` is a utility, so the plugin's output lands in
  `@layer utilities`, and an unlayered rule outranks every layer. Moving these
  into `@layer base` to tidy the file hands the backticks straight back.
- **The chip rule excludes `pre *`.** The plugin clears the ground inside a
  fenced block (`pre code { background-color: transparent; padding: 0 }`);
  being unlayered, this rule would otherwise outrank that and paint a second,
  bordered ground inside every code block.

**`frontend/test/unit/markdown-inline-code.test.ts`** — new, 7 tests.

**`docs/design-system/typography.md`** — a "Markdown prose" section recording
both overrides and both traps.

## Why the test compiles CSS

The backticks are pseudo-element content: they are never nodes. jsdom does not
implement `getComputedStyle(el, "::before")` at all, and the `<code>` element's
`textContent` is identical before and after the fix — so a DOM-only assertion
would have passed against the bug.

The test instead compiles the real `src/index.css` with the real plugin
(`tailwindcss`'s own `compile`, so it adds no dependency), flattens the result,
and resolves the cascade against a document rendered by the real `Markdown`
component — jsdom + `react-dom/client` + `act` + `createElement`, per the
suite's idiom. It runs in ~115ms.

It covers inline code in a paragraph, a list item and a `remark-gfm` table
cell, and asserts fenced blocks keep the plugin's own ground and padding.

Both halves were checked by mutation:

| Mutation | Result |
| --- | --- |
| Revert `index.css` to `main` | `draws nothing before or after an inline span` fails with `expected '"`"' to be 'none'` |
| Drop `pre *` from the exclusion | `keeps the block's own ground` fails with `expected 'var(--muted)' to be 'transparent'` |

One test asserts the plugin *still ships* the default being overridden — if a
future upgrade drops the backticks itself, that fails, and the override and the
test can go with it. A failure there is good news.

## Commands run locally

```
npm run typecheck      # clean
npm run typecheck:unit # clean
npm test               # 106 files, 1051 tests passed
npm run build          # built in 3.29s
./scripts/ci/assert-design-tokens.sh   # ✓
./scripts/ci/assert-md-line-cap.sh     # ✓
```

## Screenshots

Rendered against the compiled stylesheet from this branch and from `main`, in a
chat-row layout with inline code in a sentence, a list, a table and a fenced
block, both themes.

- **Before, light/dark**: backticks visible around every span, in the sentence,
  the list, and both table columns.
- **After, light/dark**: no backticks; each span sits on a small `--muted` chip
  with a hairline. Fenced blocks are pixel-identical to before in both themes —
  the `pre *` exclusion doing its job.

## Behaviour changes

Visual only, no API change. Every surface using `Markdown` loses its backticks
and gains the chip.

Closes #1108

### DIFF
--- a/docs/design-system/typography.md
+++ b/docs/design-system/typography.md
@@ -75,6 +75,36 @@ carrying `data-numeric`, set once in the base layer rather than per component.
 
 ---
 
+## Markdown prose
+
+`components/markdown.tsx` is the one renderer behind every surface that shows
+agent-written text — chat, memory, ledgers, artifacts, workflows — and it wraps
+its output in `prose prose-sm dark:prose-invert`, styled by
+`@tailwindcss/typography`.
+
+The plugin is written for hand-authored HTML, so two of its defaults are wrong
+here and `index.css` overrides them:
+
+| Default | Override | Why |
+| --- | --- | --- |
+| `code::before`/`::after` draw a literal `` ` `` | `content: none` | The markdown parser already consumed the fences; the plugin drew them back on, so every type name rendered as `` `UsageDto` `` (issue #1108) |
+| Inline `code` gets no ground of its own | `--muted` + hairline `--border`, `0.125em 0.375em` padding | At `0.875em` the mono face alone is a weak signal in a chat line |
+
+Two things about those overrides that are easy to undo by accident:
+
+- **They must stay unlayered.** `prose` is a utility, so everything the plugin
+  emits lands in `@layer utilities`, and an unlayered rule outranks every
+  layer. Tidying them into `@layer base` hands the backticks back.
+- **The chip rule excludes `pre *`.** The plugin clears the ground inside a
+  fenced block; being unlayered, this rule would otherwise outrank that too and
+  paint a second, bordered ground inside every code block.
+
+`frontend/test/unit/markdown-inline-code.test.ts` compiles `index.css` with the
+real plugin and resolves the cascade against a rendered document, so both of
+those fail a test rather than shipping.
+
+---
+
 ## Sizes below the scale
 
 **None remain.** Below 10px is not a size, it is a bug, and the console no
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -737,6 +737,56 @@
     --kg-tool: var(--kg-brain-2);
 }
 
+/* ============================================================================
+ * Markdown prose — inline code.
+ *
+ * `@plugin "@tailwindcss/typography"` at the top of this file ships two
+ * defaults nobody here asked for:
+ *
+ *     .prose :where(code)…::before { content: "`" }
+ *     .prose :where(code)…::after  { content: "`" }
+ *
+ * They exist because the plugin styles hand-written HTML, where a reader has
+ * no other cue that a span is code. What `components/markdown.tsx` feeds it is
+ * the *output* of a markdown parser, so the fences have already been consumed:
+ * the plugin drew them back on, and every surface using the shared `Markdown`
+ * component — chat, memory, ledgers, artifacts, workflows — rendered
+ * `` `UsageDto` `` where it meant `UsageDto` (issue #1108).
+ *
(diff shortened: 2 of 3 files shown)


graycyrus:
@coderabbitai review

PR #1120, opened 2026-08-19
perf(ci): halve the runs, split the binary build, group the feature sets, filter the lanes (#1118)

### LINKED ISSUE #1118: ci: a PR waits ~40 minutes for its first compile — no cache on the main Rust job, no shared cache key, no linker or debug-info tuning, and every lane runs on every PR
## Summary

A PR takes around **40 minutes** before it has even finished compiling. Measured on run `32223481743` (PR #1093, a one-file change):

| Job | Wall clock |
|---|---|
| **Rust (openhuman, tinycortex)** | **28m** |
| Rust | 12m |
| Desktop | 8m |
| Console E2E (live brain) | 7m |
| Console E2E | 5m |
| Rust (mongodb) | 2m |
| Console / Console (advisory) / Actions pinned | ~1m each |

Inside the 28-minute lane, the top steps are:

```
306s  Build (openhuman, tinycortex)
216s  Test tool-belt contract (openhuman, mcp, telegram, media)
196s  Build the gated host binary for the live-brain e2e lane
140s  Clippy (openhuman, tinycortex)
132s  Check (--all-features)
 94s  Test the runner lane
 80s  Install system build dependencies
```

The compile-shaped work is the majority, and almost none of the standard mitigations are in place.

## What is missing, against what OpenHuman already does

`openhuman`'s `ci-lite.yml` / `ci-full.yml` are the reference — every item below is in use there and absent here.

**1. The `rust` job has no build cache at all.** This repo's own comment says so (`ci.yml`, above `rust-gated`): *"`rust` also has no `Swatinem/rust-cache`, so a second feature set built there recompiles cold on every run, in front of everything."* 12 minutes, cold, every PR.

**2. Where the cache does exist, it is not shared.** `rust-gated` and `rust-mongodb` use `Swatinem/rust-cache@v2.9.2` with no `shared-key`, so the key is suffixed per job and no two lanes reuse each other's `target/`. OpenHuman uses `shared-key: pr-rust-*` deliberately, with a comment explaining that a stable name is the point, plus `cache-on-failure: true` so a red run still warms the next one.

**3. No linker tuning.** OpenHuman sets `RUSTFLAGS: "-C link-arg=-fuse-ld=mold"` on every Rust lane. Linking dominates the tail of each of the many `cargo test --features …` steps here, and there are ~20 of them in one job.

**4. No debug-info tuning.** OpenHuman sets `CARGO_PROFILE_DEV_DEBUG: line-tables-only` (and `"0"` on lanes that need no backtraces). Full debug info is the single largest multiplier on both compile time and link time for a workspace this size.

**5. `CARGO_INCREMENTAL` is unset.** OpenHuman pins `CARGO_INCREMENTAL: "0"`. Incremental compilation is useless on an ephemeral runner and costs time and cache footprint.

**6. Every job installs system packages from scratch** (`apt-get update && apt-get install …`, 80s a lane, repeated across lanes). OpenHuman runs its Rust jobs inside a prebuilt image, `ghcr.io/tinyhumansai/openhuman_ci:latest`, which ships the toolchain, mold, node and the system deps. That is the single biggest structural difference.

**7. Every lane runs on every PR.** There is no path filter. OpenHuman gates lanes on a `changes` job (`if: needs.changes.outputs['rust-core'] == 'true'`), so a frontend-only or docs-only PR does not pay for a Rust rebuild. A console-only PR here still pays all 28 minutes.

## Suggested order of work

Cheapest and safest first, each independently measurable:

1. **Env knobs on the Rust lanes** — `CARGO_INCREMENTAL: "0"`, `CARGO_PROFILE_DEV_DEBUG: line-tables-only`. One diff, no new infrastructure, affects every compile step.
2. **Cache the `rust` job**, and give every Rust lane a `shared-key` + `cache-on-failure: true`.
3. **mold**, once the lanes run somewhere it is installed (either apt in the deps step or the image in step 4).
4. **A prebuilt CI image** mirroring `openhuman_ci` — removes the repeated `apt-get`, and is where mold and the toolchain should live.
5. **Path-filtered lanes**, so a frontend or docs PR skips the Rust matrix entirely.
6. **Only then** look at splitting or trimming the 28-minute lane's ~20 sequential `cargo test --features …` steps, which is the riskiest change because each one exists to stop a specific silent-no-coverage regression (the comments in `ci.yml` are explicit about that history).

## Acceptance

- A one-file Rust PR reaches a compile verdict in well under 15 minutes.
- A frontend-only PR runs no Rust lane.
- No lane loses coverage: the feature-scoped test steps still run the same selections, and `Check (--all-features)` still compiles everything.

oxoxDev:
## Summary

Closes #1118.

A pull request on this repository takes ~45 minutes to answer. Measured on run `32243809850` (this branch, one CI-file change):

| Job | Wall clock |
|---|---|
| **Rust (openhuman, tinycortex)** | **42.3m** |
| Desktop | 21.9m |
| Rust | 11.8m |
| Console E2E | 5.2m |
| Rust (mongodb) | 4.0m |
| Console | 1.5m |

`Console E2E (live brain)` runs *after* the 42-minute lane, so the critical path is roughly 47 minutes. Five changes, each attacking a different part of it.

### 1. The `push` lane runs on `main` only

Unscoped `on: push` fired alongside `pull_request` on every push to every branch, so **each push to an open PR started two complete runs of this workflow**. Their `github.ref` values differ, so they land in different concurrency groups and neither cancels the other — `fix/1059-…`, `feat/1070-…` and `feat/1113-…` each show a `push` and a `pull_request` run created within two seconds, both carrying the 42-minute lane.

Kept: pushes to `main` (the post-merge verification the concurrency block exists to protect) and every pull request, forks included. Dropped: runs on branches with no PR open, which run in full the moment one is.

### 2. The gated host binary gets its own job

`Console E2E (live brain)` needs one artifact from the gated lane: a binary built with `openhuman,tinycortex,mcp,composio`. #467 built it as that lane's last step to avoid a second submodule checkout, apt install and compile — a sound argument about total CPU, and the wrong quantity to optimise. On the baseline run the lane reached that step at minute 36 of 42, having spent the intervening time on feature-scoped suites the e2e job does not consume, and `needs: rust-gated` made it wait for all of them.

Split out, the e2e job starts when the binary exists (~8 min) instead of when an unrelated test matrix finishes (~42 min). The cost is one more runner and one more 308s compile; the length of the feedback loop is the expensive resource here, not CPU minutes.

### 3. Twelve feature sets become ten

Twelve distinct `--features` strings are twelve distinct units of compilation, and the lane pays for each. From the baseline step timings — `paypal` 119s, `runner` 117s, `chargebee` 117s, `acp` 113s, `composio` 110s, `tinymemory` 90s — against **51 seconds** of actual test execution in `Test (openhuman, tinycortex)`. Nearly all of that is cargo rebuilding the crate because the feature string changed.

Two groups replace six sets:

| Group | Absorbs | Filters it runs |
|---|---|---|
| `openhuman,tinycortex,chargebee,paypal,composio` | 3 compiles → 1 | `chargebee`, `paypal`, 4× composio |
| `acp,runner,tinymemory` | 3 compiles → 1 | `runner`, 2× acp, 2× tinymemory |

Cargo features are additive, so a union set runs a **superset** of what the members ran apart. What that can delete is coverage of a feature's OFF state. Two such tests exist here — `#[cfg(not(feature = "tinyplace"))]` in `runtime::builder` and `#[cfg(not(feature = "tinymemory"))]` in `store::select` — and no filter in either group reaches them. **`tinyplace` is deliberately not grouped**: its step runs the whole `--lib`, where the `not(tinyplace)` arm is live coverage a union set would silently drop.

The `runner` and `acp` steps move from a bare `cargo test` onto `run-scoped-suite.sh`. A wider feature set makes "this filter now selects nothing" more reachable, and that failure exits 0; the script asserts a non-zero count per filter.

### 4. Each lane runs only when the change can reach it

A `changes` job resolves three areas — rust, frontend, desktop — and each lane declares which it needs. A PR editing only `gitbooks/` compiled the vendored openhuman tree three times; one editing only `frontend/` ran the 42-minute feature matrix.

The console keeps its Rust: `Console E2E` and `Console E2E (live brain)` drive the host binaries, so a frontend-only change still runs `rust` and `Gated host binary` to produce them. It drops the test matrix around them, not the coverage of what it exercises.

The `rust` filter is wide on purpose. `companies/**` is in it because `build.rs` embeds every bundle's `agents/` directory and declares a `rerun-if-changed` on it; `skills/**` because `app::types` and `server::ops::write_test` read that directory from `CARGO_MANIFEST_DIR` at test time. Neither is a Rust file and both change what the Rust lanes assert. A false positive costs one lane that need not have run; a false negative is a green PR that broke `main`.

Every filter names `ci.yml`, so a change to CI runs everything — which is what makes this PR testable by the rule it introduces.

### 5. Compiler knobs and caching (landed first, measured)

`CARGO_INCREMENTAL: 0` and `CARGO_PROFILE_DEV_DEBUG: line-tables-only` at workflow level; a cache on the `rust` job, which had none; `cache-on-failure: true` on all three Rust lanes, because the common case on a PR is iterating on a red run and the default (`false`) leaves the cache coldest exactly when it is needed most.

**Deliberately no `shared-key`.** I had this wrong in the issue and corrected it there: `Swatinem/rust-cache`'s default key already carries the job id and is stable run-to-run, so the caching lanes were never recompiling cold. `shared-key` buys sharing *between* jobs — and `rust` builds the default feature set while `rust-gated` builds `openhuman,tinycortex`, so one key would have them overwrite each other's `target/` every run.

### Expected result

| | Baseline | Projected |
|---|---|---|
| Gated lane | 42.3m | ~30m |
| Critical path (gated → live-brain e2e) | ~47m | ~30m |
| Runs per push to a PR | 2 | 1 |
| Rust lanes on a docs-only PR | 3 | 0 |

## API Or Behavior Changes

None in the product — CI configuration only.

Three changes to what CI *does*, stated plainly:

- **Coverage is a superset, not a subset.** The two grouped feature sets run every filter they ran before, under more features. The off-state tests those unions could have suppressed are named above and are not reached by any filter in either group; `tinyplace` is excluded for that exact reason.
- **Lanes can now be skipped.** `main` carries no required status checks (`repos/…/rules/branches/main` returns `[]`), so a skipped lane blocks no merge. If that changes, revisit the `if:` conditions — a required check that never runs is a PR that never merges.
- **A panic's backtrace resolves to file and line but carries no variable or type detail.** Nothing in this workflow reads the latter.

## Tests

CI on this PR is the test, and every filter names `ci.yml`, so this PR runs every lane.

What to compare against the baseline table above:

- **`Rust (openhuman, tinycortex)`** should lose the 308s binary build and ~400s of redundant feature-set compiles.
- **`Gated host binary`** is new and cold on this run — `Swatinem/rust-cache` keys on the job, so its first run pays full price and later runs do not. Judge it on the *second* run.
- **`Console E2E (live brain)`** should start minutes into the run rather than at minute 42.

- [x] `scripts/ci/assert-feature-lanes.sh` passes locally against the regrouped sets — `29 features — 19 with a lane, 10 compile-only and verified to have no gated tests, 0 owed a lane`
- [x] YAML parses; job graph verified (`changes` → 9 lanes, `console-e2e-live-brain` now needs `rust-gated-binary`)
- [x] All 43 `uses:` references are SHA-pinned, including the new `dorny/paths-filter` — the `Actions pinned` gate enforces it
- [ ] `cargo fmt` / `clippy` / `cargo test` — N/A: no Rust source changes, only the feature sets CI selects

## Documentation

No docs change. The reasoning lives in `ci.yml` beside each change, with the measured step timings quoted inline, which is where the next person to touch this workflow will be reading.

## Related

Closes #1118. Modelled on OpenHuman's CI, which solves the same problem with a `changes` path-filter job and a single product feature set (`scripts/ci/product-features.txt`) rather than a per-feature matrix.

Two follow-ups this PR deliberately does not take:

- **mold, and a prebuilt CI image** mirroring `ghcr.io/tinyhumansai/openhuman_ci`. Both cut the remaining compile-and-link time; the image also removes the ~80s `apt-get` from two lanes now that a third has been added.
- **`cargo-nextest`**, which parallelises test binaries — worth measuring once the compiles stop dominating.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,32 @@ concurrency:
 permissions:
   contents: read
 
+# Issue #1118. Three compiler knobs, set once for every job below, because the
+# critical path through this workflow is compile-shaped: on run 32223481743 the
+# `Rust (openhuman, tinycortex)` lane spent 306s in one `cargo build`, 140s in
+# clippy and 132s in `Check (--all-features)` before a single feature-scoped
+# test ran.
+#
+# `CARGO_INCREMENTAL: 0` — incremental compilation exists to make the SECOND
+# build in a working directory fast. A runner is ephemeral, so there is no
+# second build; what incremental buys here is a larger `target/` for
+# `Swatinem/rust-cache` to save and restore, paid on every lane. Cargo's own
+# guidance for CI is to turn it off.
+#
+# `CARGO_PROFILE_DEV_DEBUG: line-tables-only` — full `debug = true` emits
+# complete DWARF for every crate in the graph, and this workspace's graph is
+# large. Line tables keep a panic's file-and-line (which is what a failing test
+# is read for) and drop the variable and type records a debugger would want and
+# CI never opens. It shrinks both the compile and the link, and — because the
+# artifacts are smaller — the cache save/restore either side of them.
+#
+# These are workflow-level rather than per-job on purpose: every future Rust
+# lane inherits them without anyone remembering to. The console and desktop jobs
+# read `CARGO_*` only where they invoke cargo, so the value is inert elsewhere.
+env:
+  CARGO_INCREMENTAL: "0"
+  CARGO_PROFILE_DEV_DEBUG: line-tables-only
+
 # Issue #251. Every cargo invocation below that resolves a dependency graph
 # passes `--locked` (all of them except `cargo fmt`, which resolves nothing and
 # does not accept the flag): CI must build the graph recorded in the committed
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1100,34 +1100,74 @@ jobs:
       - name: Test (tinyplace)
         run: cargo test --locked --features tinyplace --lib
 
-      # Issue #467. The `Console E2E (live brain)` job below drives a host that
-      # has an agent harness, so it needs a FEATURE-GATED binary — the default
-      # one the `rust` job uploads answers a chat message with nothing, which is
-      # why four of the suite's best specs were skipped rather than run.
+  # Issue #1118. The gated host binary the `Console E2E (live brain)` job needs
+  # is built HERE rather than as the last step of `Rust (openhuman, tinycortex)`
+  # above, and the move is a wall-clock change, not a tidy-up.
+  #
+  # Issue #467 put it in that lane to avoid a second submodule checkout, a
+  # second apt install and a second compile of the vendored openhuman tree — a
+  # sound argument about total CPU, and the wrong one to optimise. The binary
+  # was ready roughly ten minutes into a lane that then spent another twenty
+  # running feature-scoped test suites that the e2e job does not consume, and
+  # `needs: rust-gated` made the e2e job wait for all of it. On run
+  # 32243809850 that lane reached its binary step at minute 31 of 33. The e2e
+  # suite is the LAST thing on the critical path, so every one of those minutes
+  # was added to the time a PR takes to answer.
+  #
+  # What the split costs: one more runner, one more compile of this feature set,
+  # and a cache entry of its own (`Swatinem/rust-cache` keys on the job, so this
+  # job is cold on its first run and warm after). What it buys: the e2e job
+  # starts when the binary exists instead of when an unrelated test matrix
+  # finishes. CPU minutes are the cheap resource here; the length of the
+  # feedback loop is the expensive one.
+  rust-gated-binary:
+    name: Gated host binary
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          submodules: true
+
+      # Same reason as the lane above: `submodules: true` is not recursive, and
+      # `--features openhuman` compiles the nested vendored graph rather than
+      # only resolving it.
+      - name: Init vendored openhuman crate submodules
+        run: scripts/ci/init-vendored-submodules.sh
+
+      # Same system libraries the gated lane installs, and for the same reason:
+      # `--features openhuman` links OpenSSL and the X11 dev libs that
+      # openhuman's unconditional `rdev`/`arboard` deps pull in even headless.
+      - name: Install system build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            pkg-config libssl-dev \
+            libx11-dev libxi-dev libxtst-dev libxrandr-dev libxcb1-dev libxkbcommon-dev
+
+      - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
+        with:
+          toolchain: stable
+
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
+        with:
+          cache-on-failure: true
+
+      # Issue #467. The `Console E2E (live brain)` job drives a host that has an
+      # agent harness, so it needs a FEATURE-GATED binary — the default one the
+      # `rust` job uploads answers a chat message with nothing, which is why four
+      # of the suite's best specs were skipped rather than run.
       #
-      # Here rather than in that job, for the same reason #428 put the default
-      # binary in `rust`: building it over there would mean a second submodule
-      # checkout, a second apt install and a second cold compile of the vendored
-      # openhuman tree, minutes of wall clock for a thing this job has already
-      # very nearly built.
-      #
-      # `mcp` on top of the job's own `openhuman,tinycortex`, because
-      # `mcp-agent.spec.ts` needs the `mcp_call_tool` belt entry, which is gated
-      # on it. That is another feature resolution, but a cheap one: the
-      # belt-contract step above already compiled `openhuman_core` with
-      # `openhuman_core/mcp` enabled, so what recompiles here is this crate plus
-      # its bin, not the vendored tree.
-      #
-      # `composio` on the same grounds as `mcp`, and for one spec:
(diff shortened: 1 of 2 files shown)


oxoxDev:
Measured, run `32247439903`, all eleven jobs green.

| | Baseline `32243809850` | This run |
|---|---|---|
| **Critical path** | **~47m** | **23.2m** |
| Rust (openhuman, tinycortex) | 42.3m | 22.2m |
| Desktop | 21.9m | 14.9m |
| Rust | 11.8m | 9.3m |
| Rust (mongodb) | 4.0m | 2.6m |
| Gated host binary | — (inside the lane) | 7.7m, cold |
| Console E2E (live brain) | started min 42 | **done at min 15** |
| Changed areas | — | 6s |

**Attribution, split honestly.** Of the 20 minutes off the gated lane, ~13 are structural and ~7 are cache warmth that varies run to run:

- *Binary split* — 308s of binary build gone from the lane, and the e2e suite finished 27 minutes earlier. The e2e pass also confirms the standalone job's artifact is a valid substitute: the `mcp` and `composio` specs that need those gates ran and passed.
- *Feature grouping* — 666s → 186s across the six steps it touches. The zeros are the evidence:

  | step | baseline | now |
  |---|---|---|
  | `runner` | 117s | 93s |
  | `acp` | 113s | **0s** |
  | `chargebee` | 117s | 91s |
  | `paypal` | 119s | **0s** |
  | composio ×4 | 110s | **1s** |
  | tinymemory ×2 | 90s | **0s** |

  One compile per group; every later filter in the group is free.
- *Cache warmth, not this PR* — `Build` 480s→314s, `Clippy` 269s→182s, `Check (--all-features)` 260s→176s, `tool-belt` 321s→254s. Real, but it is the previous commit's cache maturing and it will move on its own.

The duplicate-run fix is visible in the run list rather than in a timing: this push produced one run, not the `push`/`pull_request` pair every other branch still shows. That halves total CI minutes without appearing anywhere in the table above.

Every filter names `ci.yml`, so this run exercised all nine lanes — the path-filtering only pays off on PRs that do not touch CI, and it is untested for the skip case until one lands. That is the one claim here still resting on reading rather than measurement.

Issue #1124, opened 2026-08-19
approvals: mcp_call_tool is one card for every remote tool on every server


M3gA-Mind:
Split out of #877 ("approvals: the gate asks what a tool is called, not what the call is doing"), item 2. #877 keeps the standing rule and the roster; this is the classifier, because it is a **manifest-schema change with a console surface** rather than a pure classifier and should be sized as one.

## The problem

`mcp_call_tool` and `mcp_registry_tool_call` are declared flatly, by name:

- `src/policy/consequence.rs:636` — `d("mcp_call_tool", EffectGroup::Other, Reach::Consequence)`
- `src/policy/consequence.rs:674` — `d("mcp_registry_tool_call", EffectGroup::Other, Reach::Consequence)`

So every remote MCP call costs the operator the same card, whether it reads a Jira ticket or files one. That is the exact complaint #877 raises, and MCP is now the widest of the coarse names: one tool name stands in for every tool on every registered server.

## The arguments are already there to key on

`OcMcpCallTool::parameters_schema` (`src/harness/mcp.rs:425-443`) marks **`server`, `tool` and `arguments` all required**. `consequence_of` therefore already sees *which server* and *which remote tool* — the exact pair any declaration would be written against. No re-plumbing; `ToolPolicyRequest` already carries the arguments.

## The source of truth has to be an operator declaration, not discovery

⚠️ Worth stating up front, because #877's body proposes reading it from the server: **there is no client-side annotation to read.** The only `ToolAnnotations` / `readOnlyHint` in the tree is `vendor/openhuman/src/openhuman/mcp/server/` — that is openhuman acting as an MCP **server**, advertising *its own* catalogue. Nothing on the **client** path parses annotations off a remote server's `tools/list`. Reading a remote's own hints would need MCP-client work upstream first, and is a separate piece of work; it is not available to build against today.

So: an operator declaration.

## Proposed shape

**A third per-server list.** `McpServerDecl` (`src/company/mcp.rs:227`) already carries `allowed_tools` and `disallowed_tools`. A third list — remote tools this server exposes that only *read* — sits beside them with no new plumbing: it merges through `effective_mcp_servers` like the other two, and is console-editable through the existing runtime-override layer, so an operator can express it without hand-editing the manifest.

[...]

PR #1125, opened 2026-08-19
approvals: make the argument-graded tool set data, not control flow

M3gA-Mind:
## Summary

`consequence_of` answers from a call's **arguments** for four tools and from the `DECLARED` table for the other ~60. Until now, which four was a fact about **control flow** — four hand-written `if` arms — and `declared_tools()`, which every coverage test in the crate walks, chained exactly one name by hand:

```rust
DECLARED.iter().map(|d| d.tool).chain(std::iter::once(COMPOSIO_EXECUTE))
```

That is the gap #877 names as a done-criterion in as many words:

> the coverage test keeps saying which tools answer from arguments and which from the table, so a new tool cannot quietly join the coarse side

Nothing tied the two together, so a **fifth** classifier could have been added, dispatched correctly, and still been invisible to every test that walks `declared_tools()` — because those tests only ever saw the table plus one hard-coded exception.

This makes the argument-graded set **data**:

```rust
type Grader = fn(&serde_json::Value) -> Consequence;

const ARGUMENT_GRADED: &[(&str, Grader)] = &[
    (COMPOSIO_EXECUTE, composio_execute_consequence),  // #441, keyed on the action slug
    (WEB_FETCH,        web_fetch_consequence),         // #673, keyed on the URL host
    (SHELL,            shell_consequence),             // #875, keyed on the command
    (GIT_OPERATIONS,   git_operations_consequence),    // #877, keyed on the operation
];
```

`consequence_of` dispatches through it, and `declared_tools()` is **derived** from it — the union of the two mechanisms, with a roster entry that also holds a table row counted once. Adding a classifier is now one row, and joining the roster is the same act as joining the coverage test.

**No verdict changes.** The roster is walked in the same order the `if` arms were, and each entry calls the same function that arm called. This is a refactor whose whole product is that the set became enumerable.

## API Or Behavior Changes

**Behaviour: none.** Every `(tool, args)` pair maps to the same `Consequence` it did before, and the existing suite in `consequence.rs`, `judgement.rs`, `always_approve.rs` and `harness/policy.rs` — all of which walk `declared_tools()` — passes unchanged.

**Public API: none.** `declared_tools()` keeps its signature and yields the same set (the same 4 argument-graded names, the same table names, no duplicates). `ARGUMENT_GRADED`, `argument_grader` and `tool_names` are private; `Grader` is a private type alias.

## What this does NOT do — #877 stays open

This closes the roster criterion, not the issue. Deliberately out of scope, and recorded on #877 in [this comment](https://github.com/tinyhumansai/opencompany/issues/877#issuecomment-5341760015):

- **Item 1 (`git_operations`) — already shipped** in #1043, `consequence.rs:1419`. Nothing to do.
- **Item 2 (`mcp_call_tool` / `mcp_registry_tool_call` per-server read declaration) — split out to #1124**, assigned to @YellowSnnowmann. It is a manifest-schema change on `McpServerDecl` (`src/company/mcp.rs:227`) plus a console surface plus a new `ApprovalPolicy` builder dependency — a different size of thing from a classifier. #1124 also records a correction: `readOnlyHint`/`ToolAnnotations` **do** exist in the tree, but only in `vendor/openhuman/src/openhuman/mcp/server/` — openhuman acting as an MCP *server*. Nothing on the **client** path reads a remote's annotations, so the declaration still has to be the operator's, not discovered.
- **Item 3 (`http_request` / `curl` host scoping) — ruled out** by the survey on #877 and not relitigated. `consequence.rs:322` already decided it: those two can mutate, so a host-scoped grant would consent to *writing* to that host.
- **Item 4 (`workspace_write` authorship) — left**, still ranked last on #877.

`Refs #877` rather than `Closes`, because item 4 and the standing rule remain.

## Tests

Four new tests in `src/policy/consequence.rs`. **Being honest about which are revert-proof and which are forward guards** — this cluster has a history of vacuous tests, so each is labelled:

| Test | Proven how |
|---|---|
| `a_roster_entry_that_shadows_a_table_row_is_enumerated_once` | **Mutation-proven.** Dropping the `.filter(...)` from `tool_names` (i.e. concatenating instead of unioning) makes it fail — `shell` is yielded twice. It also takes the existing `declared_tools_covers_the_table_and_the_argument_classified_tool` down with it. |
| `the_roster_is_lower_case_and_has_no_duplicates` | **Mutation-proven.** Changing the roster entry to `("Shell", shell_consequence)` makes it fail. That mutation is not hypothetical: `consequence_of` matches a lower-cased name, so a mixed-case entry is a classifier silently replaced by its table row — the fail-open shape this cluster exists to prevent. |
| `a_roster_entry_with_no_table_row_is_still_enumerated` | **Fails on revert, as a compile error.** It drives `tool_names` with a synthetic roster; before this change there is no `tool_names` to drive, only a hard-coded `once(COMPOSIO_EXECUTE)`. That *is* the point — the property could not be expressed against the old shape. Not a behavioural failure, and I am not claiming it as one. |
| `the_roster_and_the_table_partition_the_known_tool_names` | **Fails on revert, as a compile error** (`ARGUMENT_GRADED` and `argument_grader` do not exist before). Its standing value is forward: it is the assertion #877 asks for, and it fails if either mechanism grows a name `declared_tools()` cannot see. |

The two compile-error entries are not dressed up as behavioural regressions. A refactor that turns control flow into data has no behavioural delta to revert against — that is what makes it a refactor. The behavioural safety net is the **existing** suite, which is unchanged and green.

Commands run locally on this branch (all with the real feature set — default features skip roughly a third of this code):

```
cargo fmt --all -- --check
cargo check  --locked --features openhuman,tinycortex --lib
cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings
cargo test   --locked --features openhuman,tinycortex --tests
```

- [x] `cargo clippy --all-targets -- -D warnings` — ran as `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, mirroring the CI lane (`--no-deps`, because a local `--all-targets` without it fails on vendored `tinyagents` lints CI never compiles).
- [x] `cargo build --all-targets` — ran as `cargo check --locked --features openhuman,tinycortex --lib` plus the clippy line above, which type-checks every target including the tests. No separate `build` run: it produces the same diagnostics at a much larger disk cost on a shared machine.
- [x] `cargo test` — ran as `cargo test --locked --features openhuman,tinycortex --tests`, plus the two mutation runs in the table above.

## Documentation

No user-facing docs. The mechanism is documented where it lives: `ARGUMENT_GRADED` carries the rationale, the shadowing rule, and the lower-case constraint; `tool_names` says why the derivation is split out from `declared_tools`; and each roster entry names the issue that added it. The follow-up split is recorded on #877 and #1124.


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
## Mutation evidence for the two revert-provable tests

Recording the actual output, not the claim. Both runs are `cargo test --locked --features openhuman,tinycortex --lib policy::consequence::tests` on this branch with one line changed, then reverted.

### Mutation 1 — `tool_names` concatenates instead of unioning

Dropping the `.filter(move |tool| !declared.iter().any(...))`:

```
thread '...::a_roster_entry_that_shadows_a_table_row_is_enumerated_once' panicked at src/policy/consequence.rs:2991:13:
assertion `left == right` failed: `shell` holds both a roster entry and a `DECLARED` row and must be enumerated once
  left: 2
 right: 1

failures:
    policy::consequence::tests::a_roster_entry_that_shadows_a_table_row_is_enumerated_once
    policy::consequence::tests::declared_tools_covers_the_table_and_the_argument_classified_tool
    policy::consequence::tests::every_consequence_tool_has_a_console_label

test result: FAILED. 59 passed; 3 failed
```

It takes down the pre-existing `declared_tools_covers_...` (78 names where 75 are expected) and `every_consequence_tool_has_a_console_label` with it, which is the point: the union is what keeps the existing coverage tests meaningful.

### Mutation 2 — a mixed-case roster entry

Changing `(SHELL, shell_consequence)` to `("Shell", shell_consequence)`:

```
thread '...::the_roster_is_lower_case_and_has_no_duplicates' panicked at src/policy/consequence.rs:3013:13:
assertion `left == right` failed: `Shell` is matched against a lower-cased name and would never fire

failures:
    policy::consequence::tests::a_read_of_the_agents_own_workspace_runs_unattended
    policy::consequence::tests::argv_sensitive_workspace_reads_run_without_admitting_writes
    policy::consequence::tests::declared_tools_covers_the_table_and_the_argument_classified_tool
    policy::consequence::tests::every_consequence_tool_has_a_console_label
    policy::consequence::tests::the_roster_is_lower_case_and_has_no_duplicates

test result: FAILED. 57 passed; 5 failed
```

Worth reading the other four failures: `a_read_of_the_agents_own_workspace_runs_unattended` and `argv_sensitive_workspace_reads_run_without_admitting_writes` are #875's behavioural tests, and they fail because the `shell` classifier **silently stopped firing** and its coarse `DECLARED` row answered instead. That is the exact fail-open shape the lower-case guard exists to catch, demonstrated rather than asserted.

### The other two tests, stated honestly

`a_roster_entry_with_no_table_row_is_still_enumerated` and `the_roster_and_the_table_partition_the_known_tool_names` reference `tool_names`, `ARGUMENT_GRADED` and `argument_grader`, none of which exist before this change — so on revert they fail to **compile**, not to assert. That is not dressed up as a behavioural regression: the property they check could not be expressed against the old shape, which is why the old shape needed changing. Their value is forward.

### Full local run on the pushed commit (`f45900f`)

```
cargo fmt --all -- --check                                                          # clean
cargo check  --locked --features openhuman,tinycortex --lib                         # EXIT 0
cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings   # EXIT 0
cargo test   --locked --features openhuman,tinycortex --tests                       # EXIT 0
  4441 passed; 0 failed; 3 ignored   (lib)
  10 / 1 / 2 / 11 / 2 passed; 0 failed  (integration binaries)
```

Working tree restored to `f45900f` after both mutations; `git status` clean.

M3gA-Mind:
## The `Desktop` lane failure is pre-existing, not this PR

Checked before assuming. `Desktop` → `Test` fails here with:

```
---- commands::test::an_instance_row_answers_in_the_keys_the_console_reads stdout ----
panicked at src/commands.rs:454:9: assertion `left == right` failed
  left:  ["id", "label", "dataDir", "running", "baseUrl", "instanceId", "operatorEmail", "companies"]
 right:  ["baseUrl", "companies", "dataDir", "id", "instanceId", "label", "operatorEmail", "running"]
```

That is in `src-tauri/src/commands.rs`, a crate this PR's one-file diff (`src/policy/consequence.rs`) does not touch, and it is **byte-identical on #1119, #1123 and #1126**. No key is missing or renamed on either side — the eight names match; only the *order* differs, because `preserve_order` is unified into the desktop graph via the host path dependency and the test asserts `BTreeMap` sorting.

Filed with the full diagnosis and a fix as **#1129**. Not fixing it here — it is an unrelated crate and would make this PR's diff two things at once.

M3gA-Mind:
Rebased onto current `upstream/main` (`025d97e1`) — `f45900f6` → `4d9f4d7e`. Clean rebase; the diff is still the same one file.

The 11 commits picked up touch **no `src/` file** (CI-lane restructuring for #1118, plus two console changes), so there was no conflict surface and no `CompanyEvent`-style exhaustive-match breakage to absorb. Re-verified on the rebased tree rather than assuming:

```
cargo fmt --all -- --check                                    # EXIT 0
cargo check --locked --features openhuman,tinycortex --lib    # EXIT 0  (5.59s — nothing in the Rust graph moved)
```

The clippy and test runs reported in the comment above were on `f45900f6`, whose tree for `src/` is byte-identical to this one; CI re-runs both here on the new lane layout.

`Desktop` will still be red: it is inherited from `main` (#1129), not from this branch.

M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Approving. I verified the equivalence by construction rather than taking the body's word for it, and it holds arm-for-arm.

All four roster constants are lower-case (`consequence.rs:292,306,310,335`), so `argument_grader`'s exact `*tool == name` against a `to_ascii_lowercase()` name reproduces the old `if name == X` chain exactly; the names are distinct, so dispatch order is immaterial. The `declared_tools()` union claim also checks out — `git_operations` (`:444`), `shell` (`:450`) and `web_fetch` (`:496`) each hold a `DECLARED` row while `composio_execute` holds none, so `DECLARED ∪ {roster entries with no row}` is exactly the old `DECLARED.chain(once(COMPOSIO_EXECUTE))`, and the `assert_eq!(all.len(), DECLARED.len() + 1)` guard is real. No verdict can flip: the graders are untouched, only dispatch moved.

Fail-closed audit is clean too. Unknown tool goes to `undeclared()` (`:1819`), not an allow. Malformed or missing args in `shell_consequence` (`:1398`) and `git_operations_consequence` (`:1485`) both return the `gated` value via `let Some(..) = args.get(..).and_then(as_str) else { return gated }`, so a non-string, array or nested shape yields `None` and gates. No `unwrap`/`expect`/indexing in any grader, no panic path.

**The one thing worth raising is a sibling dispatch this PR leaves as control flow.** `standing_scope_of` (`consequence.rs:1740-1753`) is a second dispatch keyed on the same tool identities, still hand-written, not derived from `ARGUMENT_GRADED`. The coupling is load-bearing: `web_fetch_consequence` (`:1328`) returns `Standing::ScopedGrantable` only when a scope exists, and the mint site (`runtime/cycle.rs:1061`) records `scope: standing_scope_of(...)`. So a future roster entry whose grader can return `ScopedGrantable` but which nobody adds to `standing_scope_of` mints `scope: None` — and `StandingGrant::admits_scope` (`runtime/grants.rs:382-387`) treats `None` as admitting everything, i.e. a blanket standing grant for that tool.

Not reachable today, since `shell` and `git_operations` never return `ScopedGrantable` — so latent rather than live. But it's precisely #877's "a new tool cannot quietly join the coarse side" one layer over, and this PR closes that hole for `consequence_of` while leaving it open next door. One test asserting every roster entry that can be `ScopedGrantable` is answered by `standing_scope_of` would finish the job.

Two smaller notes:

- Worth stating plainly since it's adjacent to the fail-closed question: for an *unknown* tool the answer is a name-prefix heuristic, not fail-closed. Any unregistered tool starting with `read`/`list`/`get`/`search`/`query`/`peek`/`inspect`/`view` gets `Reach::Nothing` (`:1819-1848`). Pre-existing, deliberately documented, untouched here — but it means "unknown ⇒ requires approval" isn't literally true, and the coverage test is the only thing keeping registered tools off that path.
- The new tests are set-algebra over the roster, so they'd catch an enumeration regression but not a grade flip. You say so explicitly, and the deny-direction coverage genuinely does exist in the pre-existing suite (`:3483`, `:3681`, `:3767`, `:3861+` drive `consequence_of(SHELL|GIT_OPERATIONS, ...)`). Honest accounting, no objection. `tool_names` going O(roster × DECLARED) per call is negligible given `declared_tools()` is tests plus `judgement.rs:452`.

**Your CI failures are all infra.** `Rust (openhuman, tinycortex)` hung 60m in "Install system build dependencies" (apt) and hit `timeout-minutes: 60`; `Desktop` hung 45m in "Install webkit and GTK"; `Console E2E (live brain)` hung 45m in "Install Chromium". Run 32268756479 concluded `cancelled`, not `failure`, and none reached toolchain, build, clippy or test. Same `Install Chromium` hang hit #1086, and ~6 other PR runs in the 15:13–16:14 window are cancelled. The green lanes — `Rust`, `Rust (mongodb)`, `Gated host binary`, `Console E2E` — all passed, and default-feature `Rust` does compile and run your four new roster tests. One note: your comment blaming `Desktop` on main/#1129 is stale — main's Desktop failed at 14:08 (#1126) but passed at 15:10 once #1158 fixed it, and this branch already merged that green main. Just needs a re-run.

Worth knowing: this PR has had **no substantive automated review**. tinysweeper auto-approved at push+36s on an embedding-only pass with zero inline comments, and CodeRabbit's `pass` masks an explicit "Review rate limited" — including on your own `@coderabbitai review` request. On a change inside the approvals gate, this is the only real review it's had.

Issue #1103, opened 2026-08-19
console: agent / teammate / member are used interchangeably — one dialog says both, and the host already declared the standard


oxoxDev:
## Summary

The product calls the same thing three names, sometimes in one dialog. Pick one word and use it everywhere.

The clearest case, `frontend/src/views/TeamView.tsx`:

- Dialog title (line 693): **"Define an agent"**
- Its own submit button (line 731): **"Add member"**
- The button that opens it (line 304): **"Add member"**
- The empty-state action (line 342): **"Define an agent"**

`frontend/src/views/chat/AddMemberDialog.tsx` repeats it exactly — title "Define an agent", button "Add member" — and `MembersPane.tsx` labels its control "Add member", while `OrgChartView.tsx` says **"Add teammate"** for the same operation.

Rough count of the three words across UI strings in `frontend/src/views`: **agent 76, member 28, teammate 24**.

## The standard already exists, and the console is the side that disagrees

`src/server/ops/language.rs`:

```rust
/// A teammate (never "agent") — the prosumer word for a roster member.
pub const TEAMMATE: &str = "teammate";
```

The host declared **teammate**, said "never agent" in as many words, and routes its own operator-facing strings through that constant. The console then uses `agent` more than twice as often as `teammate`. So this is not an open question about which word is nicer — it is the console drifting from a convention the codebase already states.

## Suggested direction

1. **Settle the word.** The host says `teammate`; the honest options are to follow it, or to change `language.rs` and its constants deliberately. What should not continue is two halves of one product disagreeing.
2. **Fix the one-dialog contradictions first** — title and submit button in the same box is the worst of it and the cheapest to fix.
3. **Keep `agent` where it is a genuine technical term**, not an operator-facing noun: `agentId`, `AgentDetailView`, `#/team/<agentId>`, API fields. Renaming those is a separate and much larger change, and this issue should not smuggle it in.
4. **Consider a lint or a test** over user-facing strings once the word is chosen, since 128 occurrences across three spellings will otherwise drift straight back.

## Notes

The i18n surface matters for sizing: any string that changes has to change in every locale, so the count above is the floor, not the total.

Issue #1102, opened 2026-08-19
console: teammate rows on the Company page are inert — no detail, no edit, and the "Not on a desk" chips look tappable but are not


oxoxDev:
## Summary

On the Company page, clicking a teammate does nothing. Not the name, not the row. No detail view, no edit modal, not even a tooltip. The **Not on a desk** chips have the same problem and are worse about it — they are bordered pills that read as tappable and are inert.

A teammate detail page **already exists** (`#/team/<agentId>`, `AgentDetailView`), so this is an unwired affordance rather than a missing feature.

## Where it comes from

`frontend/src/views/company/OrgChartView.tsx`.

**Staffed rows (`Seat`)** — a `<div role="treeitem">` carrying drag handlers and three action buttons (move up, move down, remove from desk). The name itself is plain text with no click target: everything interactive on the row acts *on the seat*, and nothing opens the teammate.

**Not on a desk / People chips** — plain list items:

```tsx
<li key={member.id} className="rounded-md border px-2 py-1 text-xs">
  {member.name}
</li>
```

Border, padding, rounded corners, chip styling — no `onClick`, no `href`, no `title`, no `role`. They carry every visual signal of a control and none of the behaviour, which is why they get clicked.

## Suggested direction

1. **Make a teammate row open its detail page** — `#/team/<agentId>`, the surface `TeamView` already routes to. That is the whole fix for the reported symptom, and it introduces no new destination.
2. **Do the same for the Not-on-a-desk chips**, since they name the same teammates.
3. **If a chip is deliberately not a link, stop styling it as one** — drop the pill treatment, or give it a `title` explaining what it is. The current styling makes a promise the markup does not keep.

Worth deciding in this issue: whether the whole row is the click target or just the name. The row already owns drag-to-restaff and three buttons, so a full-row click risks colliding with a drag gesture — the name (or an explicit chevron) may be the safer target. Whichever way it goes, the element that becomes clickable needs to be a real `<button>`/`<a>` with a focus ring, not a `div` with a handler, so it is reachable by keyboard.

## Reproduce

1. Company page.
2. Click a teammate's name inside any desk → nothing happens.
3. Scroll to **Not on a desk** and click a chip → nothing happens.

## Related

The People chips in the same component have the identical shape, so whatever is decided here should cover them.

PR #1126, opened 2026-08-19
feat(console): give every teammate a face

graycyrus:
Teammates were initials on a tone tile. They now get one of eleven mascot avatars.

## How a teammate gets its face

`avatarFor(seed)` hashes the same seed `toneFor` already uses, and indexes the avatar list.

**A hash rather than a random draw**, for the reason that matters to an operator: the teammate keeps the same face across reloads, browsers and machines, with nothing persisted. Drawing randomly at creation would need a stored field on the roster; drawing randomly at render would hand the same teammate a new face on every reload.

Seeded on the id where there is one — the same argument `toneFor` is given — so renaming a teammate does not change its face.

## Where the mascot is *not* drawn

`markOnly` is the existing caller-side signal for "this tile is too small to hold a glyph" — the component's own comment notes a 16px facepile fits two letters only below 10px, "and below 10px is not a size, it is a bug". A detailed drawing is subject to that limit at least as much as two letters are, so those tiles keep the tone colour alone. Shipping the mascot there would have made small facepiles worse while making large avatars nicer.

Where it does draw, the tone tile stays behind the image, so a 404 or a slow load shows the old tile rather than collapsing a message gutter to a blank square.

## Assets

WebP, 192px square — **32KB for all eleven**, against 236KB as PNG for the same set. 192 covers a 96px avatar on a 2x display.

Format notes for anyone revisiting this: SVG is not an option for this artwork (painted raster with soft gradients and grain — tracing it would be larger and worse), and JPG is wrong for icons with rounded corners over a coloured ground. If the Figma source is vector, exporting SVG from Figma directly would beat all of this.

## Verification

- `npm run typecheck`, `typecheck:unit`, `typecheck:e2e` — 0 errors
- `npm test` — **1107 passed across 114 files**

New `teammate-avatar.test.ts` covers stability across repeated calls, that a roster does not collapse onto one face, that the id seeds it, and — the one that matters most — that **every key the hash can return has a file behind it**. A key with no asset renders as a broken image only for the teammates unlucky enough to hash onto it, which is exactly the kind of bug that reaches production.

Four existing fixtures gained the new required field.


graycyrus:
## Verified in a running browser

Ran a host from this branch (loopback, own data dir) with `agentic_software_company` loaded, signed in, and drove the console with Chromium at 1440×900.

**What renders:** 13 avatar images on the chat surface, **10 distinct colourways**, **0 broken** (`naturalWidth === 0` count is zero, so every key the hash produced resolved to a real file). Box sizes in the wild are **20px** (DM rail) and **48px** (channel header).

**At 20px it holds up.** I expected a smudge and it isn't one — the silhouette reads and the colours separate cleanly down the DM list. Light and dark both fine.

**One cosmetic note:** the artwork has a baked-in lavender ground, so at 48px in dark mode the tile reads as a light square against the dark chrome. It looks deliberate rather than broken, but exporting the source with a transparent background would let the existing tone tile show through instead, which would sit better in both themes.

## A gap this PR does not close

**The Team and Company pages render zero avatars.** They do not use `views/chat/Avatar` at all — `TeamView.tsx:431-435` draws its own tile from `TEAM_TONES[member.tone]` plus `initials(member.name)`, and the org chart does likewise.

So a teammate has a face in chat and initials on the roster page, which is the page you would most expect to see it. Extending `avatarFor` to those surfaces is the obvious follow-up; I have left it out of this PR deliberately because #1102 and #1103 are editing exactly those two files right now, and a three-way edit would be worse than a follow-up.

Issue #1127, opened 2026-08-19
desktop: a key-order assertion contradicts the crate's inherited `preserve_order`, so the `Desktop` lane is red on main and on every PR since 11:26Z


M3gA-Mind:
The `Desktop` lane is failing on `main` and on every pull request opened or updated since **2026-08-19 11:26:22Z**. One test, `commands::test::an_instance_row_answers_in_the_keys_the_console_reads`, asserts a JSON object's keys in **alphabetical** order while the crate it runs in emits them in **struct** order.

Neither of the two commits that produced this is wrong on its own. They are wrong together, and they merged **19 seconds apart**, so neither pull request's own lane could ever have seen the other.

## Mechanism

`serde_json`'s `preserve_order` feature swaps the backing map for a JSON object from `BTreeMap` (keys emitted alphabetically) to `IndexMap` (keys emitted in insertion order, which for a derived `Serialize` is struct field order).

- Root `Cargo.toml:86` enables it: `serde_json = { version = "1", features = ["preserve_order"] }`
- `src-tauri/Cargo.toml:40` depends on the root crate by path: `opencompany = { path = "..", default-features = false, features = ["sqlite", "platform-jwt", "oauth"] }`
- `src-tauri/Cargo.toml:55` declares its own `serde_json = "1"` with no features
- `src-tauri/Cargo.toml:16` makes `src-tauri` an independent workspace, but that is irrelevant: `opencompany` is a normal path dependency inside the desktop build graph, so Cargo unions features on the shared `serde_json` and the desktop crate compiles against `preserve_order` too

`default-features = false` on the `opencompany` dependency does not help. It controls `opencompany`'s own features; `serde_json` is a plain unconditional entry in its `[dependencies]`, so the feature travels regardless.

Confirmed directly, offline, with no build:

```
$ cargo tree --manifest-path src-tauri/Cargo.toml -i serde_json -e features --offline
...
├── serde_json feature "preserve_order"
│   └── opencompany v0.1.0 (…/opencompany)
```

## The two commits

| | Commit | PR | Merged | Did |
|---|---|---|---|---|
| **A** | `40e75970` *chore(deps): bump ref-cast and pin serde_json preserve_order* | **#985** "Add agent-authored internal dashboard pages" | 2026-08-19 **11:26:03Z** (merge `7b101d71`) | Added the `preserve_order` feature at `Cargo.toml:86`, with its rationale at `:78-85` |
[...]

PR #1128, opened 2026-08-19
fix(console): put the New-desk name error at the name field

### LINKED ISSUE #1100: console: the New-desk validation error renders below the whole member list, far from the field it is about
## Summary

Open **New desk**, leave **Name** empty, press **Create desk**. The dialog answers `"Give the desk a name."` — at the very bottom of the form, underneath the entire roster picker. On a company with a real roster that is well below the fold, while the field it is talking about is at the top. Nothing scrolls to the message, nothing focuses the input, so the button appears to do nothing.

## Where it comes from

`frontend/src/views/company/DeskCreateDialog.tsx`.

- `submit()` sets the message into a single `error` state: `if (!name.trim()) { setError("Give the desk a name."); return; }`
- That state renders once, in JSX order **after** the Name field, the Description field, and the whole Members block — one button per roster teammate, unbounded — as `{error && <Alert variant="destructive">…}`.
- The container is `<DialogContent className="max-h-[85vh] overflow-y-auto …">`, so the form scrolls and the alert sits past the end of the list.

The distance is therefore proportional to roster size: the bigger the company, the further the error is from the empty field, and the more the Create button looks inert.

Two smaller things on the same path:

- The Name `<Input>` carries no `aria-invalid` and no `aria-describedby` pointing at the message, so a screen-reader user gets nothing at all — not even the far-away version.
- The same `error` slot is reused for host failures from `createDesk`. Those are genuinely about the whole form, so a bottom-of-form banner is defensible for them; a field-level validation message is not.

## Suggested direction

This exact bug class was already fixed next door in the workflow create dialog (PR #1019, closing part of #813) — worth copying rather than re-inventing:

```ts
requestAnimationFrame(() => {
  errorRef.current?.scrollIntoView({ block: "center", behavior: "smooth" });
  errorRef.current?.focus();
});
```

and `aria-invalid={Boolean(error)}` on the offending control.

For this dialog the fuller fix is to separate the two kinds of message: render a field-level error **next to Name** (with `aria-invalid` + `aria-describedby`, and focus moved to the input), and keep the bottom banner for submit failures the host returns. Either way, an invalid submit must move the operator to the problem instead of leaving them looking at an unchanged screen.

## Reproduce

1. A company with enough roster teammates that the member list scrolls (the staging companies qualify).
2. Company page → **New desk**.
3. Leave **Name** blank, press **Create desk**.
4. The dialog does not visibly react; the message is at the bottom of the scroll container.

graycyrus:
## Summary

The New-desk dialog's empty-name refusal now renders **at the Name field** instead of at the bottom of the form.

`submit()` set `"Give the desk a name."` into the same `error` state the host's refusal uses, and that state renders once — after Name, after Description, after the entire roster picker, inside a `max-h-[85vh] overflow-y-auto` dialog. On a company with a real roster the message landed below the fold while the empty field sat at the top; nothing scrolled to it and nothing focused the input, so pressing **Create desk** looked like pressing nothing. The distance grew with the roster.

## What changed

- **Two message slots, kept apart.** A new `nameError` renders inline under the Name input; `error` keeps the banner above the footer for whole-form failures the host returns (those are genuinely about the form, and the banner already sits directly above the button that triggered them).
- **The operator is moved to the problem.** On an invalid submit, the frame after the state commits focuses the input (`preventScroll`, so the focus move does not jump the container out from under the smooth scroll) and `scrollIntoView({ block: "center", behavior: "smooth" })`.
- **Announced, not merely rendered.** `role="alert"` on the message, `aria-describedby` from the input to it, `aria-invalid` on the input — which also lights the destructive ring `ui/input` already styles. Without these a screen-reader operator got exactly the silence a sighted one got.
- **The first keystroke retires it**, so the complaint never outlives its subject.
- **`ui/input` now forwards its ref.** React 18 drops `ref` on a plain function component, so the focus move would otherwise hold a permanently `null` ref; the Base UI primitive underneath is already `forwardRef`. Behaviourally a no-op for every existing caller.

## Other validation paths

There is only one client-side validation on this dialog — the empty name. The only other message on the path is `createDesk`'s rejection, which is a whole-form failure and stays in the banner. So nothing else needed relocating; the two are now separate states rather than one shared slot, which is what would have made a second field-level rule land in the wrong place.

## Tests

`frontend/test/unit/desk-create-name-error.test.ts` (jsdom + `react-dom/client` + `act` + `createElement`, no testing-library) covers:

1. An empty-name submit never posts, renders the message as the Name input's next sibling in the same field group (and asserts the roster picker *follows* it), wires `role="alert"` / `aria-invalid` / `aria-describedby`, focuses the input and calls `scrollIntoView`.
2. The first keystroke drops the message and the invalid state.
3. A host refusal lands in the bottom banner with the draft intact, and leaves the Name field un-flagged.

`Element.prototype.scrollIntoView` is stubbed in the test file — jsdom does not ship it, and a throw from the effect would surface as an unhandled error on the Console lane even with every test passing. The component is deliberately not guarded; the call is correct in a browser.

## Commands run

```
npm run typecheck        # clean
npm run typecheck:unit   # clean
npm test                 # 116 files, 1123 tests passed
```

Closes #1100

## Summary by CodeRabbit

* **Bug Fixes**
  * Added inline validation for missing desk names.
  * Automatically focuses and scrolls to the Name field when validation fails.
  * Clears the field error when users begin editing.
  * Keeps creation errors visible in a separate form-level banner while preserving entered details and allowing retries.

* **Accessibility**
  * Improved error messaging and field associations for assistive technologies.

* **Documentation**
  * Documented field-level and form-level validation behavior.

### DIFF
--- a/frontend/src/components/ui/input.tsx
+++ b/frontend/src/components/ui/input.tsx
@@ -3,18 +3,26 @@ import { Input as InputPrimitive } from "@base-ui/react/input"
 
 import { cn } from "@/lib/utils"
 
-function Input({ className, type, ...props }: React.ComponentProps<"input">) {
-  return (
-    <InputPrimitive
-      type={type}
-      data-slot="input"
-      className={cn(
-        "h-8 w-full min-w-0 rounded-lg border border-input bg-transparent px-2.5 py-1 text-base transition-colors outline-none file:inline-flex file:h-6 file:border-0 file:bg-transparent file:text-sm file:font-medium file:text-foreground placeholder:text-muted-foreground focus-visible:border-ring focus-visible:ring-3 focus-visible:ring-ring/50 disabled:pointer-events-none disabled:cursor-not-allowed disabled:bg-input/50 disabled:opacity-50 aria-invalid:border-destructive aria-invalid:ring-3 aria-invalid:ring-destructive/20 md:text-sm dark:bg-input/30 dark:disabled:bg-input/80 dark:aria-invalid:border-destructive/50 dark:aria-invalid:ring-destructive/40",
-        className
-      )}
-      {...props}
-    />
-  )
-}
+// `forwardRef` because React 18 does not pass `ref` through a plain function
+// component: a caller that needs to move focus to this field (the desk
+// creator's empty-name validation, issue #1100) would otherwise hold a ref
+// that is permanently `null`. The Base UI primitive underneath already
+// forwards to the native `<input>`.
+const Input = React.forwardRef<HTMLInputElement, React.ComponentPropsWithoutRef<"input">>(
+  function Input({ className, type, ...props }, ref) {
+    return (
+      <InputPrimitive
+        ref={ref as React.Ref<HTMLElement>}
+        type={type}
+        data-slot="input"
+        className={cn(
+          "h-8 w-full min-w-0 rounded-lg border border-input bg-transparent px-2.5 py-1 text-base transition-colors outline-none file:inline-flex file:h-6 file:border-0 file:bg-transparent file:text-sm file:font-medium file:text-foreground placeholder:text-muted-foreground focus-visible:border-ring focus-visible:ring-3 focus-visible:ring-ring/50 disabled:pointer-events-none disabled:cursor-not-allowed disabled:bg-input/50 disabled:opacity-50 aria-invalid:border-destructive aria-invalid:ring-3 aria-invalid:ring-destructive/20 md:text-sm dark:bg-input/30 dark:disabled:bg-input/80 dark:aria-invalid:border-destructive/50 dark:aria-invalid:ring-destructive/40",
+          className
+        )}
+        {...props}
+      />
+    )
+  }
+)
 
 export { Input }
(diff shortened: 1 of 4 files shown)


graycyrus:
CI note: the `Desktop` lane's failure here is not this PR's. `cargo test --manifest-path src-tauri/Cargo.toml` fails on `commands::test::an_instance_row_answers_in_the_keys_the_console_reads` — a serde_json map-ORDER assertion that the crate's inherited `preserve_order` now contradicts. It is red on `main` itself (025d97e1) and on every open PR, and is already tracked by #1127 / #1129. Nothing in this PR touches Rust.

`Console E2E` also went red once on `workflow-selection-persists.spec.ts` ("probe A" vs "probe B") — a workflows-tab spec this PR does not touch, which passed on the identical tree one run earlier. Re-run requested.

Frontend gates run locally on this tree: `npm run typecheck`, `npm run typecheck:unit`, `npm test` (116 files, 1123 tests) all clean.

graycyrus:
@coderabbitai review

graycyrus:
@coderabbitai full review

graycyrus:
@coderabbitai full review

PR #1130, opened 2026-08-19
fix(workflows): align the dry-run SSRF pre-flight with the real guard

### LINKED ISSUE #1075: workflows: opencompany copies upstream's SSRF predicates because url_guard is private

M3gA-Mind:
## Summary

`preflight_refusal` in `src/workflows/caps/http.rs` is a hand copy of openhuman's
private `url_guard` predicates — it exists only because `mod url_guard;` is private
upstream (`src/openhuman/tools/impl/network/mod.rs:14`) and a dry run performs
nothing, so it cannot call `HttpRequestTool` to get the guard's verdict.

The copy had already drifted, **including once in the direction the module header
explicitly rules out**. This is a live false refusal on `main`, not a future risk.

**The defect.** `host_of` did not trim a trailing dot from the **host**, while
`normalize` trims one from allowlist **entries** (`http.rs:224`, mirroring
upstream's `normalize_domain:205`). Upstream's `extract_host:220-255` trims both.
So with `web_allowed_domains = ["example.com"]` and node URL
`https://example.com./x`:

| | verdict |
|---|---|
| real guard | `extract_host` trims the dot → `example.com` matches → **allowed** |
| dry run | host stays `example.com.` → **`URL not in allowed domains: example.com.`** |

A fully-qualified trailing dot is legal and appears in real configuration, so Test
run blocked a graph that runs — and an operator cannot tell that from a genuine
refusal without arming it, which is the whole point of the control.

**Three further divergences**, all in the safer *missing refusal* direction: the
copy read past userinfo with `rsplit('@')` and unwrapped IPv6 brackets where
upstream bails outright, and it refused neither a non-http scheme nor whitespace
even though the module header claims the scheme is decided here.

**Why nothing caught it.** `dry_run_refusal_matches_the_real_client`
(`dry_run.rs:343-374`) asserts, for six URLs, that the real client refuses **and**
the dry run refuses. It is one-directional: it proves the dry run is not too
*permissive*, and contains no case where the real client allows, so it structurally
cannot see the dry run becoming too *strict*. That is the failure mode the test
exists to prevent.

### What changed

- `host_of` now mirrors `extract_host` exactly — trailing dot trimmed off the host,
  userinfo and IPv6 literals rejected outright rather than parsed past, each with
  upstream's own wording. It returns `Result<String, &'static str>` instead of
  `Option<String>`, because "the guard refuses this" and "this is unreadable here"
  are different answers and the old signature collapsed them.
- `preflight_refusal` mirrors `validate_url`'s preamble (empty / whitespace /
  scheme), which the header already claimed was decided here.
- `dry_run_refusal_matches_the_real_client` gains an open-allowlist block covering
  the URL-shape refusals, and a **new second direction**:
  `dry_run_does_not_refuse_what_the_real_client_allows`.

### Scope, and what this is not

This is the fallback the issue names, **not** its resolution. The copy exists only
because `url_guard` is private, and it should be deleted — `preflight_refusal`
collapsing to `validate_url(url, allowed_domains).err().map(|e| e.to_string())` —
once upstream exports the predicates. That needs openhuman to move first, and
nothing was filed there, so I opened **tinyhumansai/openhuman#5589** asking for a
narrow `pub use` façade over the DNS-free half (deliberately excluding
`validate_url_with_dns_check`, since resolving is a network effect a dry run must
not perform). The OpenCompany deletion then follows a submodule pin bump; it is not
in this PR.

Deliberately left as-is: the **fail-closed sentinel**. Upstream substitutes
`<misconfigured-allowlist>` when every entry is malformed and then refuses
everything (`normalize_allowed_domains:176-186`); the copy returns `None` and checks
nothing (`http.rs:180-186`). That is the safe direction, it is deliberate and
documented in place, and it disappears on its own once `validate_url` is the single
implementation.

## API Or Behavior Changes

Dry-run behavior only — no public API, no live request path. The real client is
untouched; `GuardedHttpClient` still routes everything through openhuman's guard.

- Test run **no longer refuses** a URL whose host carries a trailing dot when the
  bare host is allowlisted (`https://example.com./x` vs `["example.com"]`). This is
  the fix.
- Test run **now refuses** four shapes it used to wave through, each of which the
  real run refuses: `user@host` userinfo, an IPv6 literal, a private-IP literal
  written with a trailing dot (`http://127.0.0.1./`), and a non-http(s) scheme.
  A node aimed at any of these was already failing on the real run; it now fails on
  Test run too, in the same shape and with upstream's wording.

## Tests

Run on the feature lane — `src/workflows/` is gated behind `feature = "openhuman"`
(`src/lib.rs:66`), so a default-feature run compiles none of the changed code.

- [x] `cargo fmt --all -- --check` — pass.
- [x] `cargo clippy --all-targets -- -D warnings` — ran as CI does, `cargo clippy
      --locked --no-deps --features openhuman,tinycortex --all-targets -- -D
      warnings`: pass (the bare form fails on vendored-dependency lints CI never
      reaches).
- [x] `cargo build --all-targets` — N/A as written; `cargo check --locked --features
      openhuman,tinycortex --lib` was run instead (pass). A full `--all-targets`
      build of this tree is what the `cargo test --tests` run below already links.
- [x] `cargo test` — ran as `cargo test --locked --features openhuman,tinycortex
      --tests`: **4438 + 10 + 1 + 2 + 11 + 2 passed, 0 failed**, 3 ignored.

**Both new directions were proven against the reverted fix.** With
`src/workflows/caps/http.rs` stashed and only the new tests in place:

```
dry_run_does_not_refuse_what_the_real_client_allows ... FAILED
  the real guard allows https://parity.invalid./x but the dry run refused it
  — Err(Capability("http_request: URL not in allowed domains: parity.invalid."))

dry_run_refusal_matches_the_real_client ... FAILED
  PROOF-RUN missing refusals: ["https://user@example.com/x",
    "http://[2606:4700::1111]/x", "http://127.0.0.1./", "ftp://example.com/x"]
```

(the second line is from a one-off run with the per-case assert swapped for an
accumulator, so every open-allowlist case is shown failing rather than just the
first; the accumulator was not committed). Both pass with the fix restored.

**On the new test making no request.** The real client cannot be driven all the way
to "allowed" without issuing the request, which this suite may not do. So the
allow-direction host is an RFC 2606 `.invalid` name: it clears every guard rule and
the run then stops at DNS resolution, which cannot succeed — no connection is ever
made. The assertion is "the guard did not refuse it", which is exactly the claim
under test.

Frontend: N/A — no frontend files touched, no user-facing strings, no i18n keys.

## Documentation

N/A as a separate doc change — the rationale lives in the module's own doc comments,
which are the documentation for this code and were updated in place: the "never
stricter than the real guard" section now records that the invariant was broken by
inspection alone and why a one-directional parity test could not see it, and
`host_of`'s doc names each rule it mirrors.

Closes #1075

### DIFF
--- a/src/workflows/caps/dry_run.rs
+++ b/src/workflows/caps/dry_run.rs
@@ -330,15 +330,12 @@ mod tests {
     /// and asserts the two agree, so it keeps holding when upstream changes its
     /// internals and fails loudly the day upstream loosens a rule.
     ///
-    /// Only refusals are compared, and that is not a gap: the real guard rejects
-    /// these **before it dials anything**, so the comparison needs no network and
-    /// performs no effect. An *allowed* URL cannot be compared this way, because
-    /// confirming it would mean actually issuing the request — which is the one
-    /// thing neither a dry run nor this suite may do.
-    ///
-    /// The direction that matters is the one this covers. A dry run that refuses
-    /// what a real run would allow blocks a working graph, and an operator cannot
-    /// tell that from a real refusal without arming it anyway.
+    /// This half covers the *too permissive* direction only — the real guard
+    /// rejects every URL below **before it dials anything**, so the comparison
+    /// needs no network and performs no effect. The other direction, a dry run
+    /// stricter than the guard, is covered by
+    /// [`dry_run_does_not_refuse_what_the_real_client_allows`]; asserting only
+    /// this one is what let #1075's trailing-dot break sit on `main` unnoticed.
     #[tokio::test]
     async fn dry_run_refusal_matches_the_real_client() {
         use crate::workflows::caps::http::GuardedHttpClient;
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Rebased onto current `upstream/main` (was 11 behind). Re-verified on the rebased commit, not just the working tree:

- `cargo fmt --all -- --check` — pass
- `cargo check --locked --features openhuman,tinycortex --lib` — pass
- `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` — pass
- `cargo test --locked --features openhuman,tinycortex --tests` — 4438 + 10 + 1 + 2 + 11 + 2 passed, 0 failed

Note on the `Desktop` lane: it is red on `main` independently of this PR — `an_instance_row_answers_in_the_keys_the_console_reads` (`src-tauri/src/commands.rs:439`) asserts alphabetical JSON key order while `Cargo.toml`'s `serde_json` `preserve_order` yields struct order. Nothing in this PR touches `src-tauri/`. It is filed and owned separately, so it is deliberately not patched here.

@coderabbitai review

Issue #1135, opened 2026-08-19
workflows: the detail toolbar still carries index-level controls, on three ragged rows


graycyrus:
Now that the Workflows tab opens on the list (#1110), the detail view's toolbar carries controls that no longer make sense there, on three ragged rows.

```
CURRENT
← All workflows │ Daily Agent Harness Twee…  ● Manual run ·2m ago  [Schedule paused]  Every day at 9am, searches…
     [Daily Agent Harness Tweets ▾] [What should this run work on?] [▶Run] [Test run] [Copilot] [History 3] [Resume] [Edit] [Delete]
                                                                                                    [+ New workflow]
```

## Five problems

1. **The workflow selector duplicates the title.** You are already inside this workflow — its name is in the heading and "All workflows" goes back. Switching workflows is what the index is for now.
2. **"New workflow" is stranded** on a row of its own, right-aligned, and it is an *index-level* action living on a detail page.
3. **Nine controls in one undifferentiated strip.** Run — the point of the screen — reads as just another pill between a text field and six neighbours.
4. **Delete sits flush against Edit** with nothing separating the destructive action.
5. **The description competes with the status chips** on row 1 and truncates early.

## Proposed

```
row 1 · identity & state
← All workflows │ Daily Agent Harness Tweets    ● Manual run · 2m ago   [Schedule paused]
                  Every day at 9am, searches for trending tweets matching 'Agent harness'…

row 2 · do something
[ What should this run work on?          ] [ ▶ Run ] │ [Test run] [Copilot] [History 3] [Resume] │ [Edit] [Delete]
                                             PRIMARY   ── secondary ──                            ── utility · destructive ──
```

- Drop the selector.
- Move **New workflow** to the index and make it the **primary** button there. This screen's primary is **Run**; two filled buttons on one screen means neither reads as the main action. If it must stay reachable here, a ghost button in row 1 — not a second primary.
- Group the toolbar: run intent · secondary actions · utility, with Delete separated.
- Give the description its own line so it stops fighting the chips.

PR #1139, opened 2026-08-19
fix(workflows): give the List view three columns with fixed edges

### LINKED ISSUE #1136: workflows: the List view has three columns and no shared vertical edge
In List mode the workflow index has three columns and not one shared vertical edge between them.

```
CURRENT
│ Feature pipeline        [in source]  A feature request goes from spec…    ● Manual run blocked ·21h │
│ AJ                       Should triage across all closed github issues…   ● Manual run ok ·7d       │
│ AI trends                   Every day at 9 am look for trending ai…            No recent runs       │
                          ↑ ragged left edge                                ↑ ragged
```

Descriptions are right-aligned, so their left edge zigzags by up to a couple of hundred pixels from row to row. The status cell floats too, so the state dots never line up and "No recent runs" lands somewhere different again. The eye has nothing to track down.

## Proposed

Three columns with fixed edges:

```
│ Feature pipeline  in source │ A feature request goes from spec to a…  │ ● Manual run blocked   21h │
│ AJ                          │ Should triage across all closed githu…  │ ● Manual run ok         7d │
│ AI trends                   │ Every day at 9 am look for trending a…  │   No recent runs           │
  fixed ~280, truncate          flex-1, LEFT aligned, truncate            fixed ~210 + time column
```

- **Name** — fixed width, left aligned, truncate with a title attribute. The `in source` badge sits with the name it qualifies.
- **Description** — takes the remaining width, **left aligned**, single-line truncate.
- **Status** — fixed width, left aligned *within its column* so the dots form a vertical line; the relative time in its own narrow right-aligned column so "21h ago" and "7d ago" agree.

A CSS grid with named column widths is the honest tool here — the current `justify-between` cannot produce aligned columns because each row sizes itself independently.

Check the narrow-viewport behaviour while implementing: below some width three columns stop fitting, and the description should be the one that yields.

graycyrus:
## Summary

In List mode the workflow index had three columns and not one shared vertical edge between them. Every row laid itself out with `justify-between`, so each row sized its own cells: the descriptions were pushed right and their left edge zigzagged by a couple of hundred pixels from row to row, the status cell floated so the state dots never formed a line, and "No recent runs" landed somewhere different again.

The row is a CSS grid now, with the column widths issue #1136 approved:

| Column | Track | Alignment |
| --- | --- | --- |
| Name (+ the `in source` badge that qualifies it) | `17.5rem` fixed | left, truncate, `title` |
| Description | `minmax(0, 1fr)` | **left**, single-line truncate, `title` |
| Status (dot + reading) | `13rem` fixed | left *within its column*, so the dots form a line |
| Relative time | `4.5rem` fixed | right, so "21h ago" and "7d ago" end together |

Fixed lengths rather than `auto` is the whole fix: a row is its own element, so a grid drawn there is a grid of **one** row and `auto`/`fr` tracks are measured against that row's own content — exactly the per-row sizing that produced the ragged edge. The single `1fr` is the description, and it is the same width on every row because everything around it is.

**Narrow viewports — the description yields.** Below `@3xl` (48rem of *list* width) the three fixed tracks plus a readable description stop fitting, and the description would truncate to a word or two; it is dropped there and the name takes the space. A **container** query rather than a viewport breakpoint, because the tab sits beside a sidebar — a viewport breakpoint drops the description while there is still room and keeps it past the point where there isn't. `hidden` rather than a conditional render, so a `display: none` cell simply stops being a grid item and the remaining three fall into the three-track template with no second code path.

**The status column is one sentence** instead of a label plus badges. A badge cannot truncate — a "2 not delivered" pill beside the label in a 13rem column leaves the label four characters, two pills leave it none — so `runSummaryLine` folds the counts into prose that truncates from the right, dropping the count before the state, with the full string on a `title`. Nothing is lost against the card: `runTone` already reports an undelivered report and a waiting approval *as* the state, so the badge only added the number. When the state already names the condition the count goes in brackets beside it (`Manual run awaiting approval (1)`) rather than repeating the words (`awaiting approval · 1 awaiting approval`).

Cards mode is untouched, and so is the toolbar above the list (#1135 owns it).

## Verified in a browser

Against a real host (`companies/agentic_software_company`) seeded with 14 workflows over its own API — names from "AJ" to "Incident postmortem drafting service", descriptions from "Bump." to a full sentence, and one with none. The company-wide run page was intercepted to supply the state mix a host with no inference source can never journal (blocked, failed, running, stopped, not-delivered, awaiting-approval, and three workflows with no runs at all).

Measured across all 14 rows, light and dark:

| Viewport | Name | Description | Status | Time |
| --- | --- | --- | --- | --- |
| 1440 | 245→525 | 537→1107 | 1119→1327 | 1339→1411 |
| 1100 | 245→525 | 537→767 | 779→987 | 999→1071 |
| 820 | 245→487 | *(dropped)* | 499→707 | 719→791 |

One left edge and one right edge per column, on every row, at every width.

## Tests

- `frontend/test/e2e/workflow-list-columns.spec.ts` — twelve stubbed workflows of varied description length and nine runs in a spread of states; asserts one distinct left and right edge per column, one right edge for the time column, and one left edge for the state dots. A second test asserts the description is the column that disappears when the list narrows, and that the rest stay aligned. Geometry is the assertion because geometry is the defect — a spec reading the rendered text passes against the broken build. **Confirmed failing on `main`'s layout** ("column 0 right edges: 701, 751, 779, 743, 739, 669, 736, 738, 658").
- `frontend/test/unit/workflow-run-summary-line.test.ts` — seven cases pinning what the status sentence says and what it must not repeat.

## Commands run

```
cd frontend
npm run typecheck            # clean
npm run typecheck:unit       # clean
npm run typecheck:e2e        # clean
npm test                     # 116 files, 1127 tests passing
npx playwright test workflow-list-columns workflow-selection-persists \
    workflow-live-list workflow-edit-delete workflow-create-affordance   # 19 passed
scripts/ci/assert-design-tokens.sh   # clean
```

Closes #1136

### DIFF
--- a/frontend/src/views/workflows/WorkflowIndex.tsx
+++ b/frontend/src/views/workflows/WorkflowIndex.tsx
@@ -24,13 +24,27 @@ import { Badge } from "@/components/ui/badge";
 import { Skeleton } from "@/components/ui/skeleton";
 
 import { failedNodeOf } from "./graph";
-import { pendingCount, relativeTime, runTone, undeliveredCount } from "./run-health";
+import {
+  pendingCount,
+  relativeTime,
+  runSummaryLine,
+  runTone,
+  undeliveredCount,
+} from "./run-health";
 
 /** Which rendering the index is showing. Persisted by the caller. */
 export type IndexMode = "cards" | "list";
 
 /** How many recent runs the health strip shows per workflow. */
 const STRIP_RUNS = 5;
+
+/** The two "no run to report" readings, named once because a card and a list
+ * row both say them (issue #1136) and two copies of a sentence this careful
+ * drift apart. What each one means is argued in {@link HealthLine}. */
+const NO_RUNS_LABEL = "No recent runs";
+const NO_RUNS_TITLE =
+  "No runs in the recent company-wide page. Open the workflow to read its own run history.";
+const LOADING_RUNS_LABEL = "Loading runs…";
 
 export function WorkflowIndex({
   workflows,
@@ -94,7 +108,11 @@ export function WorkflowIndex({
           ))}
         </div>
       ) : (
-        <div className="divide-y rounded-xl border">
+        // `@container`: the row's breakpoint is the LIST's width, not the
+        // window's — the tab sits beside a sidebar, so a viewport breakpoint
+        // would drop the description while there was still room for it, and
+        // keep it past the point where there wasn't (issue #1136).
+        <div className="@container divide-y rounded-xl border">
           {workflows.map((w) => (
             <WorkflowRow
               key={w.id}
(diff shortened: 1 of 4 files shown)


graycyrus:
@coderabbitai review

graycyrus:
CI on `2bb428e`: everything green except **Desktop**, which is failing on `main` too and is unrelated to this change (frontend TypeScript only).

The Desktop lane's `Test` step fails in `src-tauri/src/commands.rs:454` — `commands::test::an_instance_row_answers_in_the_keys_the_console_reads` asserts an instance row's JSON keys in insertion order against a sorted list:

```
  left: ["id", "label", "dataDir", "running", "baseUrl", "instanceId", "operatorEmail", "companies"]
 right: ["baseUrl", "companies", "dataDir", "id", "instanceId", "label", "operatorEmail", "running"]
```

Same failure on the base commit's own `main` run (32250830983, head `025d97e1`).

Green here and worth naming: **Console E2E** and **Console E2E (live brain)**, which are what run the new `workflow-list-columns` spec.

graycyrus:
@coderabbitai review

Issue #1101, opened 2026-08-19
console: a board whose work is all in later columns reads as empty — nothing scrolls to it, and nothing says there is more to the right


oxoxDev:
## Summary

Open Tasks on a company whose work has moved on, and the board reads as dead. **To-do, Planning and In progress all show 0** — those are the three columns that fit on screen — while the real work sits in **Paused** and **In review**, off to the right. Seen with **101 cards** in those later columns and a board that looked completely empty.

You only find them if you happen to scroll right. Nothing scrolls there for you, and nothing on screen says there is anything there to scroll to.

## Where it comes from

`frontend/src/views/TasksView.tsx`.

- The board is one horizontal scroller: `className="flex min-h-0 flex-1 gap-4 overflow-x-auto py-4 pl-4"`, rendering `TASK_COLUMNS.map(...)` in fixed lifecycle order — `todo → planning → in_progress → paused → in_review → done`.
- Nothing sets the initial scroll position. The only code that touches `scrollLeft` is the drag edge-scroll (`el.scrollLeft += edgeSpeed.current`), which needs a card already in hand.
- There is no overflow affordance — no fade, no arrow, no "N more" — so an overflowing board is visually identical to one that ends at In progress.

The per-column count (`{items.length}`) is rendered, so the three visible zeros are accurate. That is what makes it read as *dead* rather than *cut off*: the operator gets three confident zeros and no hint that the count they want is two columns right.

This is worse the healthier the company is. A board only reaches this state once work has actually moved through the early columns, so the failure lands exactly on the companies doing the most.

## Suggested direction

Two options, and they are not exclusive:

1. **Scroll to the first populated column on mount.** Matches what the operator means by "show me the board". Wants a couple of decisions: only when everything left of it is empty (so it never yanks a board someone is already reading), and only on mount rather than on every refetch — this view re-reads on a live stream, and a board that re-centres itself under the reader would be its own bug.
2. **A visible overflow affordance.** An edge fade or arrow on the scroller, or column counts summarised somewhere always-visible, so a full board is distinguishable from a finished one even before scrolling.

The minimum that fixes the reported symptom is (2); (1) is what makes the board useful on open. Deciding whether auto-scroll is desirable at all is worth doing in this issue before either is built — it moves the viewport without being asked, which is the kind of thing that is right the first time and irritating the fifth.

[...]

PR #1144, opened 2026-08-19
console: retire the Tasks page, leaving Ledgers the one place tasks live

### LINKED ISSUE #1140: console: retire the Tasks page and let Ledgers be the one place tasks live
The board and the `tasks` ledger show the same records, and the operator meets them twice. Remove the Tasks nav entry and let Ledgers be the one place tasks live.

## What makes this safe to do

The board **is** the `tasks` ledger — `LedgersView.tsx:19` says so outright, and it already renders the board as columns inside Ledgers, routing a drop through `patchTask` rather than `record_entry` because entering a column *fires work* (the dispatch edge, the planning pass, the settle).

## What must not be lost

Deleting the page naively would take two things with it:

1. **Task creation.** Ledgers deliberately has **no compose box** for `tasks` — a generic ledger form would produce a save the host refuses. The board's "+ Add task" (into `todo` only, `ADD_TASK_COLUMN`) is the sole creation affordance in the console. **Ledgers must gain an equivalent** before Tasks goes, or nothing can open a card.
2. **The card detail page.** `#/tasks/<id>` carries timeline, plan, discussion, attempts and the steering controls. Ledgers explicitly does not reproduce it and links out to it. **Keep the route**; only the nav entry and the standalone board page go.

## Scope

- Remove the `tasks` nav entry and the standalone board view.
- Keep `#/tasks/<id>` resolving, from wherever it is linked (Ledgers rows, chat, approvals, the paused-card Resume).
- Add a create-task affordance to the Ledgers `tasks` view, matching the board's rule that new cards land in `todo`.
- Redirect `#/tasks` (no id) to the `tasks` ledger rather than 404ing, so existing links and muscle memory survive.
- Check every in-app link to the board — the empty states, the tour, and the Overview graph all point at it.

Recently merged board work that must keep working inside Ledgers: #1101 (empty columns collapse to a rail, and stay drop targets that auto-expand on drag-over).

graycyrus:
The board and the `tasks` ledger showed the same records through the same component, and an operator met their work twice. This removes the Tasks nav entry and the standalone board page. The board is the `tasks` ledger's columns under `#/ledgers/tasks`.

Closes #1140

## The two things a naive deletion takes with it

**Task creation.** Already there — `LedgersView` renders `+ Add task` for the native board and opens `CreateTaskDialog`, which posts `POST …/tasks` and lands the card in To-do (the `ADD_TASK_COLUMN` rule, unchanged). What was wrong was the header comment claiming Ledgers has no compose box; it now says why creation keeps its own dialog rather than becoming a `record_entry` form.

**The card detail.** `#/tasks/<id>` is linked from chat, approval cards, workflow run rows and every card on the board, and Ledgers reproduces none of it. `tasks` moves into `HIDDEN_VIEWS` (routable, unlisted) and the new `TaskDetailRoute` mounts `TaskDetailView`. Drop it from that list and `useHashView` discards the head *and* its sub-page, so every one of those links silently lands on Overview — that is the trap, and it is called out in the code.

`#/tasks` with no card — or with a malformed one like `#/tasks/%` — is rewritten to `#/ledgers/tasks` through a new optional `rewrite` argument to `useHashView`. It resolves by **replace, never push**: a retired address that pushed its replacement would sit one Back away and bounce the operator forward again forever.

## The card moved rather than dying with the page

`TaskItem` and its helpers are now `views/TaskCard.tsx`, byte-for-byte. `LedgersView` joins the ledger's rows to `…/tasks` records by id and hands the native board that card, so it keeps the priority, assignee, cost, workflow chip, plan badge, output link, and — the one that mattered most — a paused card's blocked reason and its Resume button (#883). A row whose record has not landed yet renders as an ordinary ledger row rather than waiting, so a card never flickers out of its column. Every other ledger's card is untouched.

Two things the board had and this screen had lost: the `taskEventTick` re-read (#464), so work opened from chat appears without a reload, and the dispatch/resume toasts. There is deliberately **no** 4s interval poll — a board that re-renders on a timer is what `board-columns.spec.ts`'s shared-`DataTransfer` note warns about straddling a drag.

One fix the redirect forced: `LedgersView`'s `selected` seeded from `sub` and never followed a later one, so the rewrite would have changed the address while the screen stayed on whichever ledger was already open. It follows now, which also fixes Back and hand-edited `#/ledgers/<slug>` addresses.

## Verified in a browser

Host on `127.0.0.1:8364` with `companies/agentic_software_company`, console via the Vite dev proxy, driven with the repo-pinned Playwright. 19 checks, all passing:

- `#/tasks` lands on `#/ledgers/tasks`; the six #183 columns render in order; no Tasks entry in the sidebar
- exactly one `Add task`; the created card appears and lands in `todo`; it renders as a **task** card (priority badge)
- #1101 inside Ledgers: an empty column is a rail (`data-collapsed=true`), **opens under drag-over**, and takes the drop — the card moved to `planning`
- clicking a card opens `#/tasks/<id>` with its timeline/attempts/artifacts/discussion; `Board` returns to `#/ledgers/tasks`; a cold-load deep link resolves
- a paused card offers Resume, says "Resumed — the assignee is working on it.", and the card is re-dispatched (`dispatched` in its timeline; this brain-less host then settles it back to `todo`, which is host behaviour)
- a declared ledger (`goals`) still renders with no `Add task`; `#/tasks` reaches the board even with another ledger already open

## Commands run

`npm run typecheck`, `npm run typecheck:unit`, `npm run typecheck:e2e`, `npm test` — 116 files / 1128 tests, green (8 new).

## Specs moved, not deleted

Every `page.goto("/#/tasks")` became `/#/ledgers/tasks`, because the board is there now and the selectors (`ledger-board`, `column-label`, `board-column`, `[draggable=true]`, `Add task`, `#new-prompt`) are the shared component's and did not change:

| spec | why it visits the board |
| --- | --- |
| `board-columns.spec.ts` | the board's shape: columns in order, one prompt box landing in To-do, the #1101 rail taking a drop, the live-brain planning settle. **Plus a new test** asserting the redirect and that `#/tasks/<id>` still opens the card — the regression guard for this issue |
| `assignee-picker.spec.ts` | creates via the prompt box and reads the assignee **off the board card** — the spec that proves `TaskCard` survived the move |
| `alert-dialog-confirm.spec.ts` | creates via `Add task`, opens, Edit → Delete, re-checks the board |
| `connection-degrade.spec.ts`, `desktop-connections.spec.ts` | use the board as "the console is alive" and probe `Add task` |
| `desktop-embedded-identity.spec.ts`, `desktop-first-run-signin.spec.ts` | same, no `Add task` probe |
| `board-drag.spec.ts` | already on `#/ledgers/tasks`; its header claimed two boards at two addresses. Prose only |

`test/unit/task-blocked-card.test.ts` changed one import line and passes untouched otherwise, which is the proof the card move was faithful. `test/tools/contrast-audit*.mjs` point at `ledgers` instead of a view that no longer exists.

## Docs

`docs/spec/runtime/ledgers.md` (the two-entry-point section becomes one screen with two cards; the #334 history gains this deletion as its own evidence; the `#/tasks/<id>` contract now names how it survives), `frontend/ARCHITECTURE.md`, `frontend/README.md`, `docs/plans/09-verification.md`.

## CI

The Desktop lane is red on `main` (#1127) and will fail here for the same reason — not from this change.


graycyrus:
@coderabbitai review

Issue #1035, opened 2026-08-18
chat: the composer's intent signal never reaches the card-opening path, so an operator can only ever add a card


M3gA-Mind:
## Summary

The composer carries an explicit intent control — **"Do it once" / "Build me the workflow"** — and the operator's choice arrives on the event as `CompanyEvent::OperatorMessage { deliverable, .. }` (`src/ports/types.rs:277`).

It reaches exactly two readers, and **neither can suppress a card**:

- `src/server/operator.rs:1308` — `workflow_requested`, which **opens** a card whatever the triage said (#845).
- `src/runtime/cycle.rs:708` — `inject_workflow_builder_awareness`, which folds a briefing into the message text (#845 / #859).

`grep deliverable src/runtime/delegation.rs` returns nothing relevant. That file holds **both runner card paths** — `open_direct_work_card` and the hand-off card path, which funnel through `open_work_card` — and the signal never arrives. `MessageContext`, the struct built specifically to carry "what was already decided about this operator message" into those paths, has no field for it.

So the operator's stated intent is **write-only in one direction**: it can add a card it wasn't going to get, and can never withhold one it was.

## Why this is worth its own issue

#984 fixed the *classification* half — a message the triage model reads as conversation no longer opens a card (PR #1034). That leaves the classifier as the only thing deciding, with no operator override in either direction.

The issue text for #984 named the override as a distinct requirement, and its reasoning stands on its own: a classifier is fallible, so the person who typed the message should be able to outrank it. Today they can outrank it toward *more* cards and not toward *fewer*, which is the wrong asymmetry given a board that filled with 78 chat lines.

There is also a live gap the control itself has: per #984, the client only ever sends the workflow value, so the three-way control has one live position and "Do it once" is effectively pressed by default — asserting an intent the operator never gave.

## What it would take

`handle_operator_message` (`src/runtime/delegation.rs`) would carry the deliverable through to `MessageContext`, and `open_work_card` would honour it in both directions, outranking the classifier.

Sizing it honestly, because this is the reason it was not folded into #1034: **one production caller** (`src/harness/brain.rs:2650`) and **~50 test call sites**. It also touches the same region as #982 / PR #997.

## Suggested acceptance

- An explicit "just chat" intent suppresses the card even where `is_trackable_work` would track it.
- An explicit work intent opens the card even where the triage model returned `chatter` — the override outranks the classifier both ways.
- No intent sent (older client, unaddressed message) behaves exactly as today.
[...]

Issue #984, opened 2026-08-18
chat: an ordinary message opens a task card, so the board fills with conversation


oxoxDev:
## Summary

A deliberately trivial probe — `"QA send-button mouse-click check — reply ACK."` — opened a task card. The board on smoke1 then read 98 cards, **78 of them In Review**, dominated by cards that are chat lines rather than work items.

The intent classifier is not at fault: hand-tracing the probe through `triage_message_detailed` gives Chatter/Abstained, so the documented Layer A opened nothing. The card came from a **third card path the contract never describes**: because the probe was a DM, the responder is the addressed desk member rather than the orchestrator, so `open_direct_work_card` runs and defers to `is_trackable_work` — a detector with the opposite default, documented as such ("that one asks 'is this unambiguously an instruction?', this one asks 'is there any reason NOT to track it?'").

What it accepts today, stated plainly: **everything except an empty string, a syntactic question, and a greeting of six tokens or fewer.**

## The decisive finding

The model-judged answer already exists and is discarded. The escalation classifier is wired in production, runs on exactly the abstention set, and returned Chatter for this probe — and the caller drops it, by its own doc: *"`Work` and `Chatter` therefore both leave the gate where the abstention left it, and only `Answer` moves it."*

We very likely already paid a model to call this conversation, and opened a card anyway.

## Fix

Stop discarding that verdict — zero added latency, zero added spend, only a consumer. Carry it as a distinct fact rather than by reusing the "answering" flag, whose existing argument (Chatter must never withdraw the model's board tools) still stands. Both runner-opened card paths stand down on it.

Plus:
- Widen the compiled-in small-talk floor (acknowledgement and meta vocabulary — `ack`, `qa`, `test`, `ignore`, `fyi`…) and raise the token cap, for builds where no model answered. Strictly subtractive from a set that is currently "everything".
- An explicit operator override on the request, honoured in both directions, outranking the classifier.
- A three-way composer control — `Just chat | Do it once | Build me the workflow` — with **nothing pre-pressed**. Today "Do it once" is pressed for you, asserting an intent never given, and the client only ever sends the workflow value, so the control has one live position.
- Make the "Card opened" chip dismissable. #442's justification — "a spurious card can be dismissed in one click" — is currently false, because that click does not exist on the chat surface.

## Why not the alternatives

[...]

Issue #1016, opened 2026-08-18
workflows: renaming a node orphans its edges, and node config is accepted at save but fails at run


graycyrus:
## Renaming a node id orphans its edges

Edge endpoints are raw id strings in `Select`s fed by the current node ids (`frontend/src/views/WorkflowCreateDialog.tsx:1739-1763`), and `updateNode` (`:587`) never rewrites them. Rename a node and every edge points at a value that is no longer in its own option list; Save then fails with ``Edge starts at `old-id` `` (`:749`). `removeNode` (`:686`) prunes by the *current* id, so rename-then-delete leaves dangling edges behind. The backend rejects with an edge **index** rather than the node name (`src/company/workflow_file.rs:1183-1196`), so the message does not point at the problem.

**Fix:** cascade an id rename into `edges`; prune by row `key` rather than by id.

## Config is accepted at save and fails at run

`src/company/workflow_file.rs:1502` falls through with `_ => {}` — there is no schema for `transform`, `output_parser`, `merge` or `split_out`. `sub_workflow.workflow_id` is checked only for self-reference (`:1035-1040`), and `http_request.url` is checked only for non-emptiness (`:1470-1487`).

When a 400 does come back, `src/company/workflow_create.rs:1221` joins every problem into one flat string with no field paths, so the dialog cannot highlight the offending node.

**Fix:** add per-kind config validation at save; return structured problems (`{node_id, field, message}`) so the editor can point at the node.

PR #1146, opened 2026-08-19
chat: let a spurious board card be dismissed from the chip

M3gA-Mind:
## Summary

Issue #442 allowed a turn to open a board card from an ordinary chat message, and justified it in as many words: *"a spurious card can be dismissed in one click."* On both chat surfaces that click did not exist. The "Card opened" chip was a bare `<a href="#/tasks/…">`, so clearing a mis-fired card meant leaving the conversation, finding the card on the board, and deleting it there. That is the mechanism behind #984's headline — the board fills with conversation and nothing on the chat surface can push back.

This adds the dismissal to the chip, on both surfaces, behind a confirm, over the `deleteTask` route (`frontend/src/api/tasks.ts:740`, `DELETE .../tasks/{id}`) that already existed. **No host change.**

### Seams

| File | What changed |
|---|---|
| `frontend/src/lib/chat.ts` | New pure `clearTaskCard(messages, taskId)` |
| `frontend/src/views/Conversation.tsx` | `dismissCard` callback; `CardChip` gains the confirmed control |
| `frontend/src/views/chat/StepTimeline.tsx` | The exported `CardChip` gains an optional `onDismiss` |
| `frontend/src/views/chat/MessageRow.tsx` | Threads `onDismissCard` / `dismissingCardId` |
| `frontend/src/views/chat/MessageTimeline.tsx` | Threads the same |
| `frontend/src/views/ChatView.tsx` | `dismissCard` handler |

### Three decisions worth reviewing

**Keyed on the card, not on the row clicked.** One card can be named by several lines — a turn journals the id onto its reply, and "Add to board" writes it onto the operator's own message. Clearing only the clicked bubble leaves the other chips pointing at a card the host no longer has, i.e. a link to a 404, which reads as the delete having failed. That is why the state transition is a named, tested helper rather than two lines inside a `setState`, and it follows `reconcileIds` directly above it.

**Deliberately NOT optimistic**, unlike the `react` toggle immediately beside it in `ChatView`. A reaction that rolls back costs nothing. A chip that vanishes while the card survives tells the operator the board is clean when it is not, which is the exact confusion this issue is about. So: delete on the host first, clear the chip only on success, leave it exactly where it was on a refusal and say why.

**`onDismissCard` is required** through `MessageTimeline` and `MessageRow`, so no future caller can render a chip with no way to dismiss it — a silently-missing affordance is the failure this issue is about. `CardChip.onDismiss` itself stays optional for a genuinely read-only transcript. Making it required is what surfaced two existing unit tests constructing `MessageTimeline`; both are updated rather than the props being weakened to optional.

The confirm is the repo's `AlertDialog` (the same component `TaskEditDialog` uses to confirm a task delete), **not** a native `confirm()` — no browser modal is involved.

## API Or Behavior Changes

**Console only; no host, no API, no wire change.** `deleteTask` was already there and is already used by `TaskEditDialog`.

Behaviour added: the "Card opened" / "Added to the board" chip now carries a dismiss control which, after a confirm, deletes the card and removes the chip from every line naming that card. The chip's link half is unchanged.

## What this does NOT do — #984 stays open

`Refs #984`, not `Closes`. This is item 3 of four. The rest, in the issue's own dependency order:

- **Item 1 — the operator override: not here, it has its own issue, #1035** (`oxoxDev`). The composer's `deliverable` reaches two readers and neither can suppress a card, so the operator can outrank the classifier toward *more* cards and never toward fewer.
- **Item 2 — the three-way composer control: blocked on #1035.** `MessageComposer.tsx:70` is `useState<TaskDeliverable>("once")` with `once` pre-pressed, and `client.ts:438` only ever puts the workflow value on the wire, so the control has one live position. But a "just chat" button on a composer whose signal cannot suppress a card is a control that does nothing. It needs #1035 first.
- **Item 4 — the 78 existing cards: no migration.** The issue is explicit that it must be operator-driven with a reviewed selection, and the reasoning holds: the classifier whose fallibility motivates the override is weaker evidence after the fact than it was at send time. Worth re-checking whether the board is still bad at all now that #1034 has stopped the inflow.

The headline itself is already fixed by **#1034** (`868fca9a`), which wired `TriageVerdict::is_chatter()` into `open_work_card`. Not re-litigating its deliberately narrow lexical floor — `qa`, `test` and `ignore` were refused on purpose, each opening a legitimate short instruction.

## Tests

Five new unit tests in `frontend/test/unit/chat-card-dismiss.test.ts`, over the pure helper. The unit runner is for pure functions by design (see `vitest.config.ts`), which is part of why the state transition was extracted rather than inlined.

**Each assertion is proven against a mutation** — no test here passes either way:

| Mutation | Result |
|---|---|
| `clearTaskCard` becomes the naive inline two-liner (`{ ...m, taskId: undefined }`, always a new array) | **4 of 5 fail** — the key-removal test, the field-integrity test, the array-identity test, and the untouched-lines test |
| `clearTaskCard` clears only the *first* line naming the card | **the 5th fails** — "drops the card from every line that names it" |

One of those five needed strengthening to earn its place: the field-integrity assertion originally used `toEqual`, which treats a present-but-`undefined` key as absent and therefore passed against the naive mutation. It is now `toStrictEqual`, and fails.

The affordance itself (the confirm, the disabled state) is not unit-tested: this suite is `environment: "node"` and pure-function-scoped, and a rendering test belongs in `test/e2e/`. Said plainly rather than claimed as covered.

Commands run locally, mirroring the `Console` lane:

```
npm run typecheck        # tsc -b            EXIT 0
npm run typecheck:e2e                      # EXIT 0
npm run typecheck:unit                     # EXIT 0
npm test                 # vitest run       EXIT 0 — 116 files, 1125 tests
```

- [x] `cargo fmt --all -- --check` — N/A: no Rust file is touched. This is a console-only change; `git diff --stat` is nine files under `frontend/`.
- [x] `cargo clippy --all-targets -- -D warnings` — N/A: no Rust file is touched, and the `Changed areas` lane skips the Rust jobs for a console-only diff.
- [x] `cargo build --all-targets` — N/A: no Rust file is touched.
- [x] `cargo test` — N/A for Rust. The console equivalent is `npm test`, run above: 116 files, 1125 tests, 0 failures, including the five new ones and the two updated `MessageTimeline` constructions.

`npm run build` was deliberately not run locally (the standing instruction against full builds on the shared machine); CI's `Console` lane runs it.

## Documentation

`frontend/src/views/chat/README.md`'s component table now names the card chip and its dismissal on the `MessageRow.tsx` row. The reasoning lives with the code: `clearTaskCard` documents why it keys on the card, and both `dismissCard` handlers document why they are not optimistic.

## One correction to the research notes

The notes say `StepTimeline.tsx:189` "renders the same label in the step trace". It does not — `StepTimeline.tsx` only *defines* `CardChip`; the sole render is `MessageRow.tsx:88`, as a message-bubble chip beneath the body. So the two chips are the same position on two different chat surfaces (the parked `Conversation` and the live `ChatView`), not a bubble and a step trace. Nothing about the work changes, but the next reader should not go looking for a chip inside the step list.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to dismiss linked board cards directly from chat messages.
  * Added confirmation prompts, loading states, and disabled controls during card deletion.
  * Successfully deleted cards are removed from all matching chat references.

* **Bug Fixes**
  * Failed deletions preserve card references and display an error notification.

* **Tests**
  * Added coverage for removing card references while preserving unrelated message content.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3217,7 +3217,7 @@ dependencies = [
 
 [[package]]
 name = "openhuman"
-version = "0.63.11"
+version = "0.63.9"
 dependencies = [
  "aes-gcm",
  "aho-corasick",
--- a/frontend/src/lib/chat.ts
+++ b/frontend/src/lib/chat.ts
@@ -392,3 +392,38 @@ export function reconcileIds(
   });
   return changed ? next : messages;
 }
+
+/**
+ * Forget the board card `taskId` on every line that carries it (issue #984).
+ *
+ * The dismissal half of "Add to board": #442 justified opening cards from chat
+ * on the grounds that *"a spurious card can be dismissed in one click"*, and
+ * the chat surfaces offered no such click — the chip was a bare link to the
+ * card's detail screen. Deleting the card on the host is only half of it; this
+ * is what stops the console still drawing a chip for a card that is gone.
+ *
+ * Keyed on the **card**, not on the message the operator clicked, and that is
+ * the reason this is a named function rather than two lines inside a
+ * `setState`. One card can be named by more than one line — a turn journals the
+ * id onto its reply, and "Add to board" writes it onto the operator's own
+ * message — so clearing only the clicked bubble would leave the other chips
+ * pointing at a card the host no longer has, i.e. a link to a 404. A dismissal
+ * that leaves a stale chip on screen reads as the delete having failed.
+ *
+ * Pure and total: an unknown id passes through, and a list with nothing to
+ * change is returned as-is so React sees no new array.
+ */
+export function clearTaskCard(messages: ChatMessage[], taskId: string): ChatMessage[] {
+  let changed = false;
+  const next = messages.map((m) => {
+    if (m.taskId !== taskId) return m;
+    changed = true;
+    // Drop the key rather than setting it to `undefined`: `taskId` is an
+    // optional field, and every render site tests it for truthiness, but a
+    // present-and-undefined key survives a `JSON.stringify` round trip
+    // differently from an absent one — and these rows are persisted.
+    const { taskId: _dropped, ...rest } = m;
+    return rest;
+  });
+  return changed ? next : messages;
+}
--- a/frontend/src/lib/chat.ts
+++ b/frontend/src/lib/chat.ts
@@ -412,6 +412,16 @@ export function reconcileIds(
  *
  * Pure and total: an unknown id passes through, and a list with nothing to
  * change is returned as-is so React sees no new array.
+ *
+ * # This is the client half only
+ *
+ * It clears the chip from state that is never serialised, so on its own it
+ * survives a thread switch and **not** a reload: the console rehydrates from
+ * `GET …/chat/history` (see {@link fromHistory}) and merges by message id, so an
+ * empty transcript takes every row back. The host is what stops the chip
+ * returning — its history projection blanks `task_id` for a card the board no
+ * longer has, whoever deleted it (`server::chat_history::drop_dead_cards`,
+ * issue #984). Neither half is sufficient alone.
  */
 export function clearTaskCard(messages: ChatMessage[], taskId: string): ChatMessage[] {
   let changed = false;
@@ -419,9 +429,15 @@ export function clearTaskCard(messages: ChatMessage[], taskId: string): ChatMess
     if (m.taskId !== taskId) return m;
     changed = true;
     // Drop the key rather than setting it to `undefined`: `taskId` is an
-    // optional field, and every render site tests it for truthiness, but a
-    // present-and-undefined key survives a `JSON.stringify` round trip
-    // differently from an absent one — and these rows are persisted.
+    // optional field and every render site tests it for truthiness, but the two
+    // are not interchangeable to `Object.keys`, a spread, or a deep-equality
+    // assertion, and an absent key is what "this line has no card" means
+    // everywhere else in this module.
+    //
+    // NOT because these rows are persisted — they are not. `transcripts` is
+    // React state and is never serialised. The rows that persist are the
+    // host's, and they are why this helper is only half the dismissal: see the
+    // note on the function above.
     const { taskId: _dropped, ...rest } = m;
     return rest;
   });
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -13,7 +13,7 @@ import { toast } from "sonner";
 
 import { listPeople, me as fetchMe, type Person } from "@/api/auth";
 import type { OpenCompanyClient } from "@/api/client";
-import { deleteTask, type MessageIntent, type TaskDeliverable } from "@/api/tasks";
+import { deleteTask, type MessageIntent } from "@/api/tasks";
 import { setInboxEnabled } from "@/api/inbox";
 import {
   ApiError,
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -13,7 +13,7 @@ import { toast } from "sonner";
 
 import { listPeople, me as fetchMe, type Person } from "@/api/auth";
 import type { OpenCompanyClient } from "@/api/client";
-import type { MessageIntent } from "@/api/tasks";
+import { deleteTask, type MessageIntent, type TaskDeliverable } from "@/api/tasks";
 import { setInboxEnabled } from "@/api/inbox";
 import {
   ApiError,
@@ -26,6 +26,7 @@ import {
 import { Button } from "@/components/ui/button";
 import { Skeleton } from "@/components/ui/skeleton";
 import {
+  clearTaskCard,
   makeMessage,
   reconcileIds,
   toHostMessageId,
@@ -207,6 +208,7 @@ export function ChatView({
   const [desksError, setDesksError] = useState<string | null>(null);
   const [sending, setSending] = useState(false);
   const [openThreadId, setOpenThreadId] = useState<string | null>(null);
+  const [dismissingCardId, setDismissingCardId] = useState<string | null>(null);
   const [membersOpen, setMembersOpen] = useState(false);
   const [addOpen, setAddOpen] = useState(false);
   const [mobilePane, setMobilePane] = useState<"rail" | "chat">("chat");
@@ -712,6 +714,47 @@ export function ChatView({
   }
 
   /**
+   * Delete the board card a line opened, and stop drawing its chip
+   * (issue #984).
+   *
+   * #442 allowed a turn to open a card from an ordinary message on the
+   * grounds that *"a spurious card can be dismissed in one click"*. That click
+   * did not exist here: the chip was a bare link to the card's detail screen,
+   * so clearing a mis-fired card meant leaving the channel, finding the card
+   * on the board, and deleting it there — which is how the board filled up.
+   *
+   * NOT optimistic, unlike `react` directly above, and the asymmetry is the
+   * point. A reaction that rolls back costs nothing; a chip that vanishes
+   * while the card survives tells the operator the board is clean when it is
+   * not, which is the very confusion this issue is about. So: delete on the
+   * host first, clear the chip only on success, and leave it exactly where it
+   * was on a refusal.
+   *
+   * Clears by CARD id, not by the row clicked — see {@link clearTaskCard}.
+   * Once the card is gone every chip naming it is a link to a 404, and they
+   * can sit on different lines.
+   */
+  async function dismissCard(taskId: string) {
+    if (dismissingCardId) return;
+    setDismissingCardId(taskId);
+    try {
+      await deleteTask(client, company, taskId);
+      setTranscripts((t) => ({ ...t, [active.id]: clearTaskCard(t[active.id] ?? [], taskId) }));
+      toast.success("Card dismissed.");
+    } catch (error) {
+      toast.error(
+        error instanceof ApiError && error.status === 404
+          ? "That card is already gone."
+          : error instanceof Error
+            ? error.message
+            : "Couldn't dismiss that card.",
+      );
+    } finally {
+      setDismissingCardId(null);
+    }
+  }
+
+  /**
    * Give a teammate an inbox, or take it away, on the host — keyed by the
    * roster **agent id**, which is the `InboxStore` key the Inbox page reads and
    * the ingest webhook files mail under. Nothing is persisted client-side: if
@@ -882,6 +925,8 @@ export function ChatView({
               liveSteps={openThreadId ? undefined : liveSteps}
               onOpenThread={setOpenThreadId}
               onReact={react}
+              onDismissCard={(taskId) => void dismissCard(taskId)}
+              dismissingCardId={dismissingCardId}
               onAddPeople={() => setMembersOpen(true)}
               now={now}
               askerNames={askerNames}
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -26,7 +26,6 @@ import {
 import { Button } from "@/components/ui/button";
 import { Skeleton } from "@/components/ui/skeleton";
 import {
-  clearTaskCard,
   makeMessage,
   reconcileIds,
   toHostMessageId,
@@ -63,6 +62,7 @@ import {
   firstChannel,
   historyReady,
   HISTORY_UNTRACKED,
+  clearTaskCardEverywhere,
   offersDeliverableChoice,
   resolveDmChannelId,
   toggleReaction,
@@ -730,28 +730,40 @@ export function ChatView({
    * host first, clear the chip only on success, and leave it exactly where it
    * was on a refusal.
    *
-   * Clears by CARD id, not by the row clicked — see {@link clearTaskCard}.
-   * Once the card is gone every chip naming it is a link to a 404, and they
-   * can sit on different lines.
+   * Clears by CARD id, not by the row clicked, and across every channel rather
+   * than the active one — see {@link clearTaskCardEverywhere}. Once the card is
+   * gone every chip naming it is a link to a 404, and they can sit on different
+   * lines and in different channels.
    */
   async function dismissCard(taskId: string) {
     if (dismissingCardId) return;
     setDismissingCardId(taskId);
     try {
       await deleteTask(client, company, taskId);
-      setTranscripts((t) => ({ ...t, [active.id]: clearTaskCard(t[active.id] ?? [], taskId) }));
+      clearCardEverywhere(taskId);
       toast.success("Card dismissed.");
     } catch (error) {
-      toast.error(
-        error instanceof ApiError && error.status === 404
-          ? "That card is already gone."
-          : error instanceof Error
-            ? error.message
-            : "Couldn't dismiss that card.",
-      );
+      // A 404 is positive proof the card is gone — most likely deleted from the
+      // board itself, where nothing tells the open chat surface about it. The
+      // chip is then a permanent link to a 404 that no amount of clicking can
+      // remove, so this is a success for the operator's purpose: clear it and
+      // say so. Only a refusal we cannot interpret leaves the chip in place.
+      if (error instanceof ApiError && error.status === 404) {
+        clearCardEverywhere(taskId);
+        toast.success("That card was already gone — chip cleared.");
+      } else {
+        toast.error(
+          error instanceof Error && error.message ? error.message : "Couldn't dismiss that card.",
+        );
+      }
     } finally {
       setDismissingCardId(null);
     }
+  }
+
+  /** Drop the card from every channel — see {@link clearTaskCardEverywhere}. */
+  function clearCardEverywhere(taskId: string) {
+    setTranscripts((t) => clearTaskCardEverywhere(t, taskId));
   }
 
   /**
--- a/frontend/src/views/Conversation.tsx
+++ b/frontend/src/views/Conversation.tsx
@@ -22,17 +22,29 @@ import type { OpenCompanyClient } from "@/api/client";
 import { ApiError, type TurnStep, type TurnStepKind } from "@/api/types";
 import {
   createTask,
+  deleteTask,
   listInflight,
   steerTask,
   type InflightRun,
   type SteerAction,
 } from "@/api/tasks";
 import { Markdown } from "@/components/markdown";
+import {
+  AlertDialog,
+  AlertDialogAction,
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+  AlertDialogTrigger,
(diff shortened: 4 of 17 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (changes requested):
The affordance is right and the pure-helper design is clean — and not making the update optimistic is the correct call for a hard delete. I also checked the authz and it's fine: `ScopedCompany` blocks cross-company, and member-level rather than admin matches the pre-existing `TaskEditDialog` delete, so no finding there.

The problem is that "No host change" turns out to be load-bearing in the wrong direction.

**Blocking — the dismissal doesn't survive a reload, and the chip returns pointing at a 404.**

`clearTaskCard` mutates React state only; `transcripts` is `useState<Transcripts>({})` (`app-shell.tsx:382`) and never serialized. But the host persists `task_id` on the chat row and `fromHistory` rehydrates it (`lib/chat.ts:323-325`), merged by message id (`app-shell.tsx:697-702`). So a full reload re-adds the chip for a card that no longer exists — which is exactly the state `clearTaskCard`'s own docstring says must not happen ("a link to a 404 … reads as the delete having failed").

The repo already pins that rehydration: `frontend/test/e2e/chat-to-card.spec.ts:100-113` is literally "…and survives a reload". Fixing it needs the host change this PR set out to avoid — either clear/blank `task_id` on the journaled rows when the card is deleted, or have `fromHistory` drop ids the board no longer has.

**Blocking — dismiss-during-dispatch resurrects the card, and by then the chip is gone.**

`delete_task` never consults the in-flight registry, and the harness settle path does an unconditional `tasks.upsert(&company, &card)` from its in-memory clone (`harness/brain.rs:1245`, `:1355`, `:2155`). So a card deleted mid-turn is re-created when the turn lands, in `in_review`/`done`, while `clearTaskCard` has already removed every chip naming it. Net result is a zombie card on the board that can no longer be dismissed from chat — the "board fills with conversation" failure #984 is about.

Not a corner case, either: `CardChip` renders on "the card a turn opened, **or the one it dispatched to**" (`StepTimeline.tsx:191-200`), i.e. the card most likely to be running. The host already has both halves needed — `runtime.steer().list()` (see `list_inflight`, `ops/tasks.rs:2303`) and a `cancel` steer requiring `confirm:true` (`:2315`). Either refuse the delete on an in-flight card, or cancel-then-delete.

**Major**

- The `404` path never clears the chip. `ChatView.tsx`'s `dismissCard` catch toasts "That card is already gone." and leaves the chip in place, and `Conversation.tsx` doesn't special-case it at all. So a card deleted from the board first (via `TaskEditDialog`) has a chat chip that is permanently un-dismissable and permanently links to a 404. A 404 is positive proof the card is gone — that branch should call `clearTaskCard` and succeed.
- The affordance itself has zero coverage, and the test that would have caught the reload bug is a few lines from one that exists. Five tests, all on the pure helper; `chat-to-card.spec.ts` already drives both surfaces, already asserts the chip, and already reloads. I credit you saying the gap plainly rather than claiming coverage — but it's still the gap.

**Non-blocking**

- `ChatView.dismissCard` clears only `t[active.id]` (`ChatView.tsx:736`). The PR's own thesis is that one card can be named by several lines — that holds across channels too, since a dispatch marker lands in the origin thread's channel. Chips in a non-active channel keep pointing at the deleted card.
- The docstring and test comment "*and these rows are persisted*" (`lib/chat.ts` ~:420, `chat-card-dismiss.test.ts:38-42`) is wrong about the client store — it isn't persisted. The rows that *are* persisted are the host's, which is the first blocker. Right conclusion, wrong reason, and the wrong reason is what hides the bug.

**Nits:** error copy differs between surfaces (`ChatView` maps 404 and sentence-cases; `Conversation.tsx:314` falls back to lowercase "could not dismiss this card"). And `dismissingId` in `Conversation.tsx` holds a *task* id while its sibling `addingId` holds a *message* id — same name shape, different namespace.

Idempotency and double-click are fine — single-flight guard plus `disabled` on the other chips, dialog closes on click, server 404s on a repeat. No i18n owed; this repo has no i18n layer, unlike openhuman.

All 11 CI lanes are green. One caveat on the bot approvals though: **CodeRabbit never actually reviewed this** — it posted "Review limit reached" on `2c993d7b`, and its `@coderabbitai review` reply said "Already reviewed" pointing at the pre-merge `8a6eb29a`, while the check still reads `pass`. tinysweeper approved on an embeddings-only pass with zero inline comments. So the two green approvals aren't carrying any signal here.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Both blockers were real. I checked each claim at the lines you cited before touching anything, and every one of them held — including the two non-blocking ones. Nothing here is pushed back on. `57387554`, rebased onto current `main`.

## Blocking — the dismissal didn't survive a reload

Confirmed exactly as described. Worth adding one detail that makes it sharper: the merge at `app-shell.tsx:697-702` filters hydrated rows by `known` message ids, so a **thread switch** does keep the dismissal — the row is already in state and the rehydrated copy is dropped. A **full reload** starts from `useState({})`, so `known` is empty and every row comes back, `task_id` included. The dismissal looked durable in every way except the one that mattered.

**Fixed in the projection rather than by a write, and that is the substantive decision here.** Your first option — clear/blank `task_id` on the journaled rows — is not available: `task_id` is a field of the `CompanyEvent::AgentReply` that *happened*, and the journal is append-only. Rewriting it would be editing history to record that a turn never opened a card, when it did.

So it is your second option, `server::chat_history::drop_dead_cards`: the projection blanks a `task_id` whose card the board no longer has. That is not just the available option, it is the better one:

- It covers a card deleted by **any** path. `TaskEditDialog`'s delete leaves exactly the same stale chip and always has — which is also the root of the `404` finding below.
- It covers cards deleted **before** this change, which no write-time fix could reach.
- It cannot drift. One board, read at render time, rather than a denormalised copy that a missed call site leaves stale.

Cost is one board read per history and none at all when no row in the window carries a card — the same shape as the existing single roster read, not a read per message.

## Blocking — dismiss-during-dispatch resurrected the card

Confirmed: `delete_task` consulted nothing, and the settle path upserts from the in-memory clone. Now a `409` naming the steer route.

**I took "refuse" over "cancel-then-delete", and not because it was smaller.** A cancel is cooperative — it sets the run's `SteerControl` and the turn stops at its *next iteration boundary*, which may be after the settle write has already gone out. Cancel-then-delete would therefore reintroduce the very race it was meant to close, just less often. That is the worst shape a fix can have: it passes a test and fails in production. Refusing leaves the operator in a state they can act on — cancel, watch it stop, then delete.

The check is inside the existing board-write lock, so a run registering between the check and the delete cannot slip through.

## Major — the 404 never cleared the chip

Right, and it was the more common path than it looks: it is what a card deleted from the board first always produced. Both surfaces now treat a `404` as the success it is for the operator's purpose — `clearTaskCard` runs and the toast says the card was already gone. Only a refusal that cannot be interpreted leaves the chip in place.

## Major — the affordance had no coverage

Fixed, and on the path you pointed at. `chat-to-card.spec.ts` now has a dismissal test: dismiss → confirm → chip gone → card gone from the board → **reload** → still gone. It runs on the "Add to board" path rather than the scripted-backend one, so it needs no `LIVE_BRAIN` and runs on every CI.

It is deliberately the mirror image of your existing "…and survives a reload": that one proves a live card's chip comes back, mine proves a dismissed card's chip does not. Neither is safe alone — a host that blanked every `task_id` would pass mine and fail yours, which is why I also added a Rust test asserting the chip *is* projected while the card exists, before the one asserting it is dropped once deleted.

## Non-blocking — only the active channel was cleared

Right, and it holds for the reason you give. Now `clearTaskCardEverywhere` in `views/chat/model.ts`, a pure function over `Transcripts` that preserves object identity when nothing changed. Made a helper rather than inlined in the `setState` for the same reason `clearTaskCard` was: the unit runner can reach it, and it is now covered four ways.

## Non-blocking — "and these rows are persisted" was wrong

Corrected in both the docstring and the test comment, and you are right that the wrong reason is what hid the bug — the sentence asserted the exact property that was false. The real reason to drop the key rather than blank it is that a present-and-`undefined` key is not the same thing to `Object.keys`, a spread, or a deep-equality assertion. The function docstring now also states plainly that it is the client half only, and names the host half that stops the chip returning.

## Nits

Both taken. Error copy is identical on both surfaces now. `dismissingId` → `dismissingCardId`, since it holds a task id while its sibling `addingId` holds a message id.

## On the bot approvals

Noted, and thank you for checking rather than counting the green ticks — a `pass` from a review that never ran is worse than no check, because it looks like signal. I have re-requested CodeRabbit on this push.

## Verification

Every new test fails with its fix reverted; each mutation was applied alone and reverted after:

| mutation | test that fails |
|---|---|
| `drop_dead_cards` not called | deleted-card projection |
| it blanks every `task_id` | live-card projection |
| in-flight guard removed | running-card refusal |
| guard refuses any card while anything runs | not-running delete still works |
| `clearTaskCardEverywhere` clears only one channel | every-channel |
| it always returns a new object | identity preserved |

The last two mutations are the reason the "not running" and "identity" tests exist: without them, a guard that refused every delete and a helper that always re-allocated would both have passed.

Two existing fixtures (`graphql/test.rs`, `operator.rs`) journaled a reply naming `t-77` without ever putting `t-77` on the board — an impossible state that the projection now correctly blanks. They seed the card, which is what they were asserting all along. I checked the ordering in `harness/brain.rs` first to be sure a real turn never journals a reply before its card exists; it does not, and the dispatch reply is not desk history at all (`chat_id` is the card id, which matches no desk).

Local, on the rebased tree:

- `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` — clean
- `cargo test --locked --features openhuman,tinycortex --lib` — **4592 passed / 0 failed / 3 ignored**
- `npm run typecheck`, `typecheck:unit`, `typecheck:e2e` — all clean
- `npx vitest run` — **1282 passed / 132 files**

The e2e test is the one thing I could not run here: it needs the live host the e2e harness brings up. Saying that rather than implying it is proven.

@coderabbitai review

M3gA-Mind:
@oxoxDev — rebased onto current `main` (`9bab7d7e`, the branch was 88 behind). Head is now `01da9954`. The point-by-point reply to all six findings is [above](https://github.com/tinyhumansai/opencompany/pull/1146#issuecomment-5347263042); this is the short version of the blocking one plus what it means for your reload test.

## The host change

`server::chat_history::drop_dead_cards` — the history projection blanks `task_id` on any row whose card the board no longer has. One board read per history, none at all when no row in the window carries a card.

**Not** the first option you offered (clear/blank `task_id` on the journaled rows), and worth saying why: `task_id` is a field of the `CompanyEvent::AgentReply` that *happened*, and the journal is append-only. Rewriting it would be editing history to record that a turn never opened a card, when it did. So it is your second option, and it turns out to be the better one anyway — it covers a card deleted from the **board** (`TaskEditDialog`, which left the identical stale chip and always has), it covers cards deleted before this change, and there is one board read at render time rather than a denormalised copy a missed call site leaves stale.

Second blocker is a `409` in `delete_task` when the card is in flight, naming the steer route. I took refuse over cancel-then-delete deliberately: a cancel stops the turn at its *next iteration boundary*, which can be after the settle write has gone out, so cancel-then-delete would reintroduce the same race less often — passes a test, fails in production.

## Against your reload test

`chat-to-card.spec.ts:100-113` is untouched, and that is deliberate — it pins the property that must keep holding.

I added a new test beside it, `a dismissed card's chip goes away and does not come back on reload`, which is its **mirror image**: yours proves a live card's chip comes back after a reload, mine proves a dismissed card's chip does not. Neither is safe alone — a host that blanked every `task_id` would pass mine and fail yours, which is exactly the failure mode a one-sided fix would have.

It runs on the "Add to board" path rather than the scripted-backend one, so it needs no `LIVE_BRAIN` and runs on every CI — unlike yours, which is `test.skip`ped without it.

The same pairing exists on the Rust side: `a_reply_keeps_its_card_while_the_card_exists` is asserted **before** `a_reply_loses_its_card_once_the_card_is_deleted`, so the projection cannot pass by dropping every id it sees. Reverting `drop_dead_cards` fails the second; making it blank every `task_id` fails the first.

## Two fixtures I had to correct

`graphql/test.rs` and `operator.rs` each journaled a reply naming `t-77` without ever putting `t-77` on the board — a state a real turn cannot produce, which the projection now correctly blanks. They seed the card, which is what they were asserting all along. I checked `harness/brain.rs` first to confirm no real path journals a reply before its card exists; none does, and the dispatch reply is not desk history at all (`chat_id` is the card id, which matches no desk).

## On the rebased tree

- `cargo check --locked --features openhuman,tinycortex --lib` — clean
- `cargo check --locked --features openhuman,tinycortex --tests` — clean (no sign of #1197 here; `main` already carries #1200)
- `npm run typecheck` — clean

Full `cargo test --lib` (4592 passed / 0 failed) and `vitest` (1282 passed) were green pre-rebase; CI is re-running both on this head. The e2e is the one thing I cannot run locally — `Console E2E` is its first real execution, and I am watching it rather than claiming it.

M3gA-Mind:
### `Rust (openhuman, tinycortex)` went red on `01da9954` — it is a pre-existing flake on `main`, not this branch

Recording the evidence rather than just re-running it, because "re-run and it went green" is not proof of anything.

The failures are in `store::memory::test::namespace_driver_conformance` (`src/store/conformance.rs`), under the `acp,runner,tinymemory-embedded` feature set — a different subsystem from anything here, and a feature set my local `openhuman,tinycortex` runs never compile.

**It is non-deterministic.** Same commit, same test binary, five consecutive runs on this branch:

| run | result | failed |
|---|---|---|
| 1 | ok 67 | — |
| 2 | ok 67 | — |
| 3 | FAILED 65/2 | `isolation_by_company_holds`, `export_totality_holds` |
| 4 | ok 67 | — |
| 5 | FAILED 66/1 | `export_totality_holds` |

CI's own failure on this head was a *third* combination (`context_chunk_stamps_holds` + `export_totality_holds`). A regression from a diff fails the same tests every time; this does not.

**And it reproduces on `main` with none of this branch's code.** Detached at `9bab7d7e` — the exact commit this is rebased onto — same five runs:

| run | result | failed |
|---|---|---|
| 1 | FAILED 66/1 | `export_totality_holds` |
| 2 | ok 67 | — |
| 3 | FAILED 66/1 | `export_totality_holds` |
| 4 | ok 67 | — |
| 5 | FAILED 66/1 | `export_totality_holds` |

Three failures in five runs on clean `main`. So the lane is flaky independently of this PR, and `export_totality_holds_on_the_namespace_driver` is the most frequent offender.

I have re-run the lane. Flagging it as its own defect worth filing — a suite that fails ~50% of the time on `main` will keep costing every PR that touches this repo a red lane and an argument about whose fault it is. Happy to file it if nobody has.

Everything else on this head is green, including **`Console E2E`**, which is the first real run of the new dismissal-and-reload test.

M3gA-Mind:
### Addendum on the `store::memory` flake — attempt 2 failed the same way, and two obvious causes are ruled out

The re-run failed on the identical command (`--features acp,runner,tinymemory-embedded --lib store::memory`). Expected: at a ~50% per-run failure rate, two reds in a row is ordinary, so **I have stopped re-running it** — chasing a green here would prove nothing.

Two extra data points, so whoever picks this up does not re-tread them:

**It is not test parallelism.** Six runs with `--test-threads=1` on this branch: **3 failed, 3 passed.** Each test already builds its own `tempfile::tempdir`, so the tempdirs are not shared either. I had assumed parallel interference through `tinymemory-core`'s process-global statics (`embedding_host::HOST`, `learning_candidate::GLOBAL_BUFFER`, `events::SINK`) — that hypothesis is **wrong**, and I would rather say so than leave it in the record as a lead.

**It is not same-millisecond trace reordering.** `ProviderMemoryStore::ordered_traces` (`src/store/memory/facades.rs:506`) already imposes a total order — `at_millis` then `cycle_id` — with a comment saying exactly why. So `assert_eq!(recent, traces)` at `conformance.rs:907` is not flaking on ordering.

I have not identified the root cause and am not going to guess at one in a PR that does not touch this subsystem.

**What is established** is the part that matters here: clean `main` at `9bab7d7e`, none of this branch's code, fails **3 of 5 runs**. This lane is red for reasons that predate this PR and will keep going red for other PRs until someone fixes it. Worth its own issue; happy to take it if it is routed to me.

Every other lane on `01da9954` is green, including `Console E2E` and `Console E2E (live brain)`.

oxoxDev:
Verified this review's fix commit (`01da9954`, now rebased to `60568495` in the current branch) against each point before rebasing further — nothing left open on my side:

- **Blocker 1 (reload resurrects the chip):** fixed in the projection, not a write — `chat_history::drop_dead_cards` blanks `task_id` on any row whose card the board no longer has, at read time (`src/server/chat_history.rs`). Correctly framed as a projection concern since the journal is append-only.
- **Blocker 2 (dismiss-during-dispatch resurrects the card):** `delete_task` now checks `runtime.steer().list()` under the existing `task_writes` lock and returns `409` naming the steer route instead of racing the settle-path upsert (`src/server/ops/tasks.rs`). Two new tests assert both the refusal *and* that the card survives it, plus that a non-running delete still works — so a guard that refused everything wouldn't pass.
- **Major (404 never clears the chip):** both `ChatView.tsx` and `Conversation.tsx` now treat a 404 as proof-of-deletion and clear the chip instead of leaving it stuck.
- **Major (zero coverage):** `chat-to-card.spec.ts` gained an e2e case mirroring the existing "survives a reload" test, plus 9 new unit tests on the pure helpers.
- **Non-blocking (other channels keep a stale chip):** new `clearTaskCardEverywhere` sweeps every channel's transcript, not just the active one.
- **Nits:** error copy now matches across both surfaces; `dismissingId` → `dismissingCardId`.

Separately, rebasing this onto current `upstream/main` (to clear the `DIRTY` merge state) surfaced two mechanical follow-ups, now on top as their own commits:
- `Cargo.lock` needed a resync — `vendor/openhuman`'s pin moved from 0.63.11 to 0.63.9 as part of the rebase, and `cargo metadata --locked` was failing on the old lock entry. Regenerated; only the `openhuman` package's locked version line changed.
- `ChatView.tsx` had a dead `TaskDeliverable` type import left over from a merge conflict — upstream's #1016 rename cascade changed `send()`'s second parameter from `deliverable?: TaskDeliverable` to `intent?: MessageIntent` before this branch got here. `tsc -b --noEmit` caught it (TS6133); removed.

Full gate suite green post-rebase: `cargo fmt --check`, both clippy lanes (base + `--features openhuman,tinycortex`), both test lanes (3114 + 4728 passed, 0 failed), `cargo metadata --locked`, all three TS typecheck lanes, `npm run build`, `vitest run` (1413 passed). The one red CI lane (`Rust (openhuman, tinycortex)` on the pre-rebase run) was the pre-existing `namespace_driver_conformance` failure tracked in #1238 — reproduces on `upstream/main` itself, not something this PR introduced or needs to fix.

Leaving the review decision for a human re-review rather than flipping it myself.

PR #1154, opened 2026-08-19
Give host-authored notices an author that is honestly not an agent

### LINKED ISSUE #966: console: desk history renders pre-#885 replies with a confident but unrecoverable author

M3gA-Mind:
## Summary

**This is a recovery, not a new change.** PR #1060 built exactly this fix, was reviewed, approved by `tinysweeper` and CI-green — and never reached `main`. It was opened with `baseRefName: fix/966-confined-turn-author`, a branch whose own PR (#1040) had already landed, and was "merged" into that dead branch on 2026-08-19 05:16 UTC. It reads as merged on the PR list and on the issue timeline, and **none of its five files changed on `main`**.

Verified by symbol rather than by the merged badge, since a merged PR whose base was a feature branch looks exactly like one that landed:

- `git merge-base --is-ancestor a9b4e472 upstream/main` → **NO** (the merge `1830ab61` likewise)
- `grep -rn SYSTEM_AUTHOR src/` on `main` → nothing
- `src/server/chat_history.rs:547` `is_known_author` → still only `CONFINED_AGENT_ID` + roster
- `src/company/runtime.rs:1898` → still `agent_id: OPERATOR_CHANNEL`

Commit **`a9b4e472`** is re-raised here as `1a9ad8a4`, cherry-picked onto current `main` with an identical diffstat, so it can be diffed against the reviewed original. **Full credit to that commit — the content is unchanged and was already reviewed; only its destination was wrong.** This is not a duplicate of #1060; #1060 changed nothing on `main`.

### What the fix does

Three sites journal prose the runtime wrote itself — the approval-overflow notice, the `"Acknowledged."` cycle fallback, and the failed-continuation report — as an ordinary `AgentReply` carrying `"operator"` in the author field. That made a **correct** system row byte-identical on disk to a reply whose author the pre-#885 defect overwrote, which is why a reader-side marker was not buildable: no stored field separated them.

`SYSTEM_AUTHOR` is a reserved id in the ungated `ports::ids`, carried on the `OutboundMessage.agent` field #885 added, set at the three sites, and admitted by `is_known_author` so the attribution audit does not count three correct rows as damage. No new event variant, no schema change, no reader change.

### Two things I changed on top of the recovered commit

**`685041ed` — a doc comment on the wrong function.** The recovered commit inserted `system_notice` between `confined_bubble`'s doc comment and its `fn`, so all seventeen lines about the workflow-copilot turn attached to `system_notice` and `confined_bubble` was left undocumented. Kept as its own commit so `1a9ad8a4` stays byte-identical to `a9b4e472`.

**`bcaae55a` — the tests did not guard the fix.** `a9b4e472`'s message records that an earlier version of its tests was vacuous and that naming the constructors fixed it. It fixed half. The tests pin what the constructors *build*; nothing forced the production paths to *call* them. Reverting each site with the message text left byte-identical, so only the author moved:

| Reverted | Result before `bcaae55a` |
|---|---|
| `is_known_author` admission | fails — guarded |
| `system_notice` body | fails — guarded |
| `continuation_failure_notice` body | fails — guarded |
| `"Acknowledged."` **call site** → inline `agent: None` | **4441 passed, 0 failed** |
| continuation-failure **call site** → inline `OPERATOR_CHANNEL` | **4441 passed, 0 failed** |
| approval-overflow **call site** → inline `agent: None` | **4441 passed, 0 failed** |

The journal is the whole subject of this issue, so an unpinned journal path is the gap worth closing. Both new assertions go on tests that already drive the real path rather than on new fixtures: `no_events_still_acknowledges` runs a real cycle, and `a_failed_continuation_tells_the_operator` goes through the HTTP approve and reads the journaled row. Each now fails on its call-site revert — `left: None / right: Some("system")` and `left: ["operator"] / right: ["system"]`.

**One site is still unpinned, stated rather than papered over:** the approval-overflow call site. Reaching it through a cycle needs a gated tool call, a scripted model and a parking host — a far larger fixture than the two above. Its constructor is guarded; its call site is not.

### Deliberately out of scope

The issue has a second half — how to present pre-cutover rows — which the research notes leave as a product decision (their own lean is option 2, "mark as legacy", with option 1 now costing more than the issue body assumed because the affected set provably contains correct system rows). Nothing here decides that. What this does is make every row journaled *after* it cleanly separable, so whichever presentation wins can be applied without slandering a system row.

Forward-only, and permanently so: notices already journaled keep `"operator"` and stay indistinguishable. The distinguishing information was never written down.

## API Or Behavior Changes

**Two wire-visible changes, both on the same three notice rows.**

**1. The author, and with it the rendering.** Those three notices move from rendering as a company bubble to rendering as a centred system pill. `MessageView::project` already maps an `AgentReply`'s `agent_id` into `author`, and `SYSTEM_AUTHOR`'s value matches the literal the projection already uses for the `DeskTaskCompleted` marker — so the console reaches its existing system rendering with no reader change.

**2. `MessageView.channel`, which the first draft of this body did not mention** (caught in review). The `AgentReply` projection arm derives `channel` from `agent_id`, so these three rows also go `channel: "operator"` → `channel: "system"`. That is consistent with the pill rendering and no consumer keys on it, but it is a second field changing shape, not a side effect of the first. Holding `channel` at its old value would mean special-casing the projection for three rows — trading an accurate wire field for a cosmetic one — so it is left to follow the author.

No new event variant, no schema change, no route change. `GET {scope}/chat/attribution-audit` stops counting post-fix system notices as damage, which is a correction to that figure rather than a change of contract — and `AttributionAudit`'s doc now states that the figure is **not comparable across the cutover**: a company's `affected` can fall without a single row being repaired, purely because it stopped minting new false positives.

**`SYSTEM_AUTHOR` is a reserved agent id.** `RESERVED_AGENT_IDS` grows from three entries to four, so `mint_agent_id` will never hand `system` to a teammate. Without that, the id was mintable (`agent_slug("System")` produces it) and a teammate holding it would have rendered as the host — a worse confusion than the one this fixes, given the value it replaces was already reserved.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — PASS at default features, and as CI's gated form `--no-deps --features openhuman,tinycortex --all-targets`. Default features matter here: `ports::ids` is ungated on purpose, or `chat_history` would not compile without the harness features.
- [x] `cargo build --all-targets` — N/A as a separate step; covered by the clippy and test runs above, which compile every target in both feature sets.
- [x] `cargo test` — **4441 passed / 0 failed** gated (`--features openhuman,tinycortex --lib`) and **2919 passed / 0 failed** at default features.

Every new assertion was proven by reverting its fix and showing it fails, then restoring — the table above records which reverts failed nothing before `bcaae55a`, and the one site that still fails nothing.

## Documentation

The reasoning sits at the seams it governs: `SYSTEM_AUTHOR`'s doc says why it is neither an agent nor a destination and that it is forward-only; `is_known_author`'s doc says why counting these rows would inflate the one figure in #965 that has to be trustworthy; `system_notice` and `continuation_failure_notice` each say why they are named rather than inlined. The full discriminator table — every field that fails to separate a notice from a damaged reply — is recorded on the issue.

No `docs/` change: no route, launcher behaviour or `tiny*` integration changes.

Closes #966

## Summary by CodeRabbit

- **New Features**
  - Runtime-generated notices are now clearly identified as system messages.
  - Failed-continuation, approval-limit, and empty-cycle notices use consistent system attribution.
  - System notices remain directed to the appropriate conversation thread.

- **Bug Fixes**
  - System notices no longer appear as company-authored messages or task cards.
  - Legacy operator-authored notices continue to display correctly.
  - Host-authored notices are excluded from affected-message attribution counts.

### DIFF
--- a/frontend/test/unit/chat-system-notice.test.ts
+++ b/frontend/test/unit/chat-system-notice.test.ts
@@ -1,0 +1,85 @@
+import { describe, expect, it } from "vitest";
+
+import type { ChatHistoryMessageDto } from "@/api/types";
+import { fromHistory } from "@/lib/chat";
+
+/**
+ * Issue #966 — the console half of the host-authored notice.
+ *
+ * Three sites emit prose the runtime wrote itself: an approval-overflow notice,
+ * the `"Acknowledged."` cycle fallback, and a failed-continuation report. They
+ * used to journal under `"operator"`, which made a correct system row identical
+ * on disk to a reply whose author the pre-#885 defect had overwritten. The host
+ * now journals them under `SYSTEM_AUTHOR`, and this file pins what that buys on
+ * the surface a reader actually looks at.
+ *
+ * **What is and is not covered here.** The `author === "system"` branch itself
+ * predates this issue (#377 added it for the dispatch marker), so these are not
+ * a regression guard for the host-side change — the marker tests already cover
+ * that branch, and they only ever exercise it through a row carrying a card.
+ * What was untested is this shape: a system-authored row with **no** `taskId`,
+ * which is what all three notices are. Together with the host-side assertion
+ * that `SYSTEM_AUTHOR` is literally `"system"`, this is the coupling that keeps
+ * the two halves spelled the same.
+ */
+
+const AT = 1_700_000_000_000;
+
+/** A notice as the host now journals it: authored by the runtime, no card. */
+function notice(over: Partial<ChatHistoryMessageDto> = {}): ChatHistoryMessageDto {
+  return {
+    id: "512",
+    channel: "system",
+    author: "system",
+    text: "Acknowledged.",
+    atMillis: AT,
+    mine: false,
+    ...over,
+  };
+}
+
+describe("a host-authored notice", () => {
+  it("renders as a centred pill, not as the company speaking", () => {
+    const [line] = fromHistory([notice()]);
+
+    expect(line.from).toBe("system");
+    expect(line.text).toBe("Acknowledged.");
+  });
+
+  /**
+   * The notices carry no card. The marker tests only ever drive the system
+   * branch through a row that has one, so this is the shape they leave open.
+   */
+  it("carries no card chip", () => {
+    expect(fromHistory([notice()])[0].taskId).toBeUndefined();
+  });
+
+  /** All three notices, including the one that bypasses `OutboundMessage`. */
+  it("reads the same for the failed-continuation report", () => {
+    const [line] = fromHistory([
+      notice({
+        id: "513",
+        text: "Your approval was recorded, but the agent could not pick the work back up.",
+      }),
+    ]);
+
+    expect(line.from).toBe("system");
+  });
+});
+
+describe("a notice journaled before the fix", () => {
+  /**
+   * The forward-only boundary, stated as a test rather than only in prose.
+   *
+   * A row already on disk keeps `"operator"` and is indistinguishable from a
+   * reply whose author was overwritten, so it still reads as the company. That
+   * is not a defect this change can repair — the distinguishing information was
+   * never written down — and pinning it here keeps the limit visible to anyone
+   * who later wonders why old transcripts still look wrong.
+   */
+  it("still reads as a company bubble, which no fix can undo", () => {
+    const [line] = fromHistory([notice({ channel: "operator", author: "operator" })]);
+
+    expect(line.from).toBe("company");
+  });
+});
--- a/src/company/runtime.rs
+++ b/src/company/runtime.rs
@@ -384,6 +384,28 @@ pub struct CompanyRuntime {
     pub(crate) mcp: Option<Arc<crate::harness::mcp::McpRuntime>>,
 }
 
+/// The event the runtime appends when a continuation could not be picked back
+/// up (issue #469, defect 4).
+///
+/// Named so its **author** can be asserted (issue #966). This site writes the
+/// `AgentReply` directly rather than going through `OutboundMessage`, so it does
+/// not get the `agent` field's fallback and has to name the author itself. It
+/// used to store `OPERATOR_CHANNEL`, which made a correct system row
+/// indistinguishable on disk from a reply whose author the pre-#885 defect
+/// overwrote.
+fn continuation_failure_notice(thread: String, parent: Option<EventSeq>) -> CompanyEvent {
+    CompanyEvent::AgentReply {
+        parent,
+        chat_id: thread,
+        agent_id: crate::ports::SYSTEM_AUTHOR.to_string(),
+        text: "Your approval was recorded, but the agent could not pick the work back up. \
+               Nothing was half-done — approving again is safe and will retry it."
+            .to_string(),
+        steps: Vec::new(),
+        task_id: None,
+    }
+}
+
 impl CompanyRuntime {
     /// Assembles a runtime from its ports. Most callers use
     /// [`RuntimeBuilder`](crate::runtime::RuntimeBuilder) instead.
@@ -1996,20 +2018,7 @@ impl CompanyRuntime {
         let parent = self.resolvable_parent(conversation.parent, &thread).await;
         if let Err(err) = self
             .events
-            .append(
-                &self.id,
-                CompanyEvent::AgentReply {
-                    parent,
-                    chat_id: thread,
-                    agent_id: crate::runtime::channel::OPERATOR_CHANNEL.to_string(),
-                    text: "Your approval was recorded, but the agent could not pick the work \
-                           back up. Nothing was half-done — approving again is safe and will \
-                           retry it."
-                        .to_string(),
-                    steps: Vec::new(),
-                    task_id: None,
-                },
-            )
+            .append(&self.id, continuation_failure_notice(thread, parent))
             .await
         {
             tracing::warn!(
@@ -2806,7 +2815,10 @@ impl std::fmt::Debug for CompanyRuntime {
 
 #[cfg(test)]
 mod tests {
-    use super::{emergency_from_load, task_enters_in_progress, task_enters_planning};
+    use super::{
+        CompanyEvent, continuation_failure_notice, emergency_from_load, task_enters_in_progress,
+        task_enters_planning,
+    };
 
     /// Issue #880: which parked approvals name a workflow run, and which must
     /// not.
@@ -3916,6 +3928,34 @@ mod tests {
             rt.resolvable_parent(Some(roots[0]), "desk-ops").await,
             None,
             "and the General desk is not a named one",
+        );
+    }
+
+    /// Issue #966: the failed-continuation report is authored by the runtime.
+    ///
+    /// This site appends the `AgentReply` itself, so it never sees
+    /// `OutboundMessage::agent` or its `channel` fallback — it has to name the
+    /// author, and it used to name `OPERATOR_CHANNEL`. That made a correct
+    /// system row byte-identical on disk to a reply the pre-#885 defect had
+    /// damaged, which is the finding recorded on #966.
+    #[test]
+    fn a_failed_continuation_report_is_authored_by_the_runtime_not_the_operator() {
+        let event = continuation_failure_notice("desk-general".to_string(), None);
+        let CompanyEvent::AgentReply {
+            agent_id, chat_id, ..
+        } = event
+        else {
+            panic!("the notice must stay an AgentReply — the console renders it from that arm");
+        };
+        assert_eq!(agent_id, crate::ports::SYSTEM_AUTHOR);
+        assert_ne!(
+            agent_id,
+            crate::runtime::channel::OPERATOR_CHANNEL,
+            "a notice must not store the author a destination-overwrite produces"
+        );
+        assert_eq!(
+            chat_id, "desk-general",
+            "it still lands in the thread it answers"
         );
     }
 }
--- a/src/harness/brain.rs
+++ b/src/harness/brain.rs
@@ -230,23 +230,6 @@ pub struct HarnessBrain {
     runs: Option<Arc<dyn crate::ports::RunStore>>,
 }
 
-/// The single bubble a workflow-copilot turn returns (issues #416, #966).
-///
-/// Named rather than inlined because its **author** is the load-bearing field
-/// and it was wrong. #885 taught the main operator bubble to carry its
-/// responder and left this branch on `None`, so a genuine copilot reply kept
-/// journaling as `agent_id: "operator"` — the #885 defect still happening, not
-/// history needing a label. A function is what lets that be asserted without
-/// standing up a scripted model endpoint and a whole harness pool.
-///
-/// `CONFINED_AGENT_ID` is deliberately not a roster id (see [`confine`]): it
-/// names no teammate and cannot be addressed. That makes it a **truthful**
-/// author rather than a resolvable one, which is why
-/// `chat_history::is_known_author` has to know it — otherwise this row trades
-/// one wrong answer for a permanent false positive in the attribution audit.
-///
-/// No card, by construction: a confined turn has no `spawn_task` to call, and
-/// the chat handler does not open one from a copilot message either.
 /// A bubble the **runtime** wrote, not an agent (issue #966).
 ///
 /// Two sites emit these on the operator channel: the approval-overflow notice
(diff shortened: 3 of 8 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (changes requested):
The recovery is clean and I verified the claim — `git grep SYSTEM_AUTHOR origin/main -- src/` returns nothing, so #1060 genuinely never landed, the cherry-pick is clean, and separating the follow-ups so `1a9ad8a4` stays byte-identical to the reviewed `a9b4e472` is the right structure. The added call-site pinning is a real improvement on the original.

One thing to fix, and it's a one-liner, but it inverts the change.

**Major — `SYSTEM_AUTHOR` is spoofable, and the value it replaces was not.**

`ports/ids.rs:71` sets `SYSTEM_AUTHOR = "system"`, but `RESERVED_AGENT_IDS = ["operator", "Agents", "Desks"]` (`ports/types.rs:2922`, pinned at `:4937`) — `"system"` isn't in it. `agent_slug("System")` → `system` (lowercase ASCII, starts with a letter, so a legal roster id), and `roster_id_taken` (`:3280`) checks only that reserved list, the roster, and desk/group-chat names. So `mint_agent_id("System")` hands out `system`.

The chain then renders that teammate as the host: `chat_history.rs:369-370` projects `AgentReply` as `channel: agent_id.clone(), author: agent_id`, and `frontend/src/lib/chat.ts:297` does `entry.author === "system" ? "system" : …` → the centred system pill. `is_known_author` (`chat_history.rs:556-557`) also short-circuits on the string before consulting the roster, so the attribution audit rubber-stamps it.

It's reachable by an agent unattended — the doc at `types.rs:3215` names the orchestrator's `add_agent` tool alongside `POST …/team` as minting sites. The sharp edge is that `"operator"`, the value being replaced, **is** reserved, and the sibling `CONFINED_AGENT_ID = "workflow-copilot"` is unmintable by construction (hyphen; slugs only emit underscores). `SYSTEM_AUTHOR` is the only reserved author without that protection, so as written this moves the host author from a reserved name to an unreserved one.

Fix: add `crate::ports::SYSTEM_AUTHOR` to `RESERVED_AGENT_IDS` — the list already documents itself as holding references to the real constants — and update the `:4937` pin.

**Non-blocking**

- `MessageView.channel` silently changes for these three notices. Because the `AgentReply` arm derives `channel` from `agent_id`, they go from `channel: "operator"` to `channel: "system"`. Consistent with the intended pill rendering and I found no consumer that breaks on it, but it's a second wire-visible change the body describes only as an author change.
- No test on the surface that actually displays them. The three notices now render as centred pills instead of company bubbles — a real visible change resting on inspection alone, given the console has no test for it and you're already honest that the approval-overflow call site is unpinned.
- Nit: attribution-audit figures become non-comparable across the cutover, since pre-cutover system notices still count as damage and post-cutover ones don't. Acknowledged as forward-only, but the #965 figure the body says "has to be trustworthy" now has a discontinuity nothing labels.

**On the historical notices:** not migrated, and correctly so. Forward-only is stated in the body, the `SYSTEM_AUTHOR` doc and the commit message, and the distinguishing information was never written to disk so no backfill is possible. Deferring the issue's second half as a product decision is the right call, honestly disclosed.

**Downstream consumers check out:** `is_known_author` is the only backend consumer keying on the author string and it's updated; the frontend keys on `author === "system"` and already handles it via the pre-existing `DeskTaskCompleted` path; nothing in permissions, filtering or memory ingestion keys off it. No new user-facing strings, so no i18n owed.

CI: both failures are infra, not yours. `Rust (openhuman, tinycortex)` was cancelled after 59 minutes stuck on `apt-get update` at `Get:1 file:/etc/apt/apt-mirrors.txt Mirrorlist`, with every downstream step skipped so no test ever ran; `Console E2E (live brain)` cancelled at `Install Chromium`. Both distinct from main's earlier red (#1126's typecheck failure, since fixed by #1158). Worth a cancel + `gh run rerun --failed` rather than any code change.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Thanks — the major finding is correct, and I verified it rather than taking it on trust. Fixed in `e5db6feb`, along with all three non-blocking items. Rebased onto current `main` (`22934221`) first.

## Major — `SYSTEM_AUTHOR` is spoofable

**Confirmed, and the sharpest way to say it is your framing: the change moved the host author from a reserved name to an unreserved one.** I wrote the assertion *before* applying the fix so the vulnerability is demonstrated rather than argued:

```
mint_agent_id_never_returns_the_host_author ... FAILED
assertion `left != right` failed: a teammate must never be minted onto the id the runtime speaks under
  left: "system"   right: "system"
```

Every link checked at the lines you cited: `RESERVED_AGENT_IDS` was `[OPERATOR_CHANNEL, AGENTS_ROOT, DESKS_ROOT]` with no `"system"`; `roster_id_taken` consults only that list, the roster, and desk/group-chat ids and names; `chat_history.rs:369-370` is `channel: agent_id.clone(), author: agent_id`; `chat.ts:297` keys on `author === "system"`; and `is_known_author` does short-circuit on the string ahead of the roster.

Fixed exactly as you suggested — `crate::ports::SYSTEM_AUTHOR` added as a constant reference, which is what the list documents itself as holding, and the `:5039` pin updated with `mint_agent_id("System") == "system_2"` alongside it. I also recorded in the list's doc why `CONFINED_AGENT_ID` needs no entry (slugs never emit a hyphen), since that asymmetry is the thing a future reader would otherwise trip on.

## Non-blocking

**`MessageView.channel` changes too.** You are right that the body described only an author change. Now stated in the PR body as a second wire-visible change: because the `AgentReply` arm derives `channel` from `agent_id`, the three notices go `channel: "operator"` → `"system"`. I did not try to hold `channel` at its old value — that would mean special-casing the projection arm for three rows, which trades an accurate wire field for a cosmetic one.

**No test on the surface that displays them.** Added `frontend/test/unit/chat-system-notice.test.ts` (4 tests), and paired it with a host-side assertion.

Worth being precise about what each one buys, because a frontend test here is easy to over-claim: the `author === "system"` branch predates this issue (#377), so a console test is a **coupling** guard, not a regression guard for the host change — its doc says exactly that. What was genuinely untested is the notice *shape*: a system-authored row with **no** `taskId`. The marker tests only ever drive that branch through a row carrying a card.

The host-side half is the one that pays for itself. `the_notice_author_is_the_literal_the_console_keys_on` pins that the constant's *value* is the contract with `chat.ts`. Redefining `SYSTEM_AUTHOR` to `"host"` leaves **all four** original #966 tests green while the console silently returns these rows to company bubbles — I checked, and only the new test fails. Two copies of one literal, for the same reason `dispatch_marker_text` already carries two.

The console test also pins the forward-only boundary as a test rather than only as prose: a pre-cutover row still reads as a company bubble, and nothing can undo that.

**Audit discontinuity.** Labelled on `AttributionAudit`. The wording is that a company's `affected` can fall without a single row being repaired, purely because it stopped minting new false positives — so a decline across the cutover reads as "the bleeding stopped", never "history got better".

## CI

Agreed on both, and neither was code: `Rust (openhuman, tinycortex)` cancelled after 59 minutes stuck on `apt-get update`, `Console E2E (live brain)` at `Install Chromium`. This push re-runs everything from scratch, so it supersedes both rather than needing a targeted rerun.

Verified locally on the rebased tree: `cargo fmt --all -- --check`, `cargo check --locked --features openhuman,tinycortex --lib` and `--tests`, `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`, `cargo test --locked --features openhuman,tinycortex --lib` (**4590 passed / 0 failed**), `npm run typecheck`, and the new vitest file.

Nothing pushed back on — the review was accurate on every point I could check.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Cross-linking: **#1187** (oxoxDev) re-raises the same lost commit `a9b4e472` — I diffed the two patches and they are byte-identical apart from hunk line numbers, so the recovered content is the same on both sides.

Full comparison posted there: https://github.com/tinyhumansai/opencompany/pull/1187#issuecomment-5347728563

The one difference that is a defect rather than a preference is the `SYSTEM_AUTHOR` spoofability from oxoxDev's review on this PR: #1187 does not reserve the id, and I verified that against its own tree (`mint_agent_id("System")` still returns `"system"` there) rather than inferring it from the file list. Whichever PR is preferred, that reservation is worth carrying across.

Holding here — no further changes to this branch — until oxoxDev decides which to take. This PR stays open and intact either way, so nothing is lost.

M3gA-Mind:
@oxoxDev — what changed since your review, so you don't have to re-derive it. Rebased onto current `main` (`9bab7d7e`); the branch was 103 behind. Four commits, and the first is unchanged from what you reviewed.

**`539d45b2` — `fix(chat): give host-authored notices an author…`** *(unchanged)*
The rescue of `a9b4e472`, byte-identical to the reviewed original apart from its rebase. Kept as its own commit so it stays diffable against that one.

**`2ccf61d2` — `docs(brain): put confined_bubble's doc back on confined_bubble`**
`a9b4e472` inserted `system_notice` between `confined_bubble`'s doc comment and its `fn`, so all seventeen lines about the workflow-copilot turn attached to `system_notice` and `confined_bubble` was left undocumented. Moved back. No code change.

**`c4f86c4a` — `test(chat): pin the notice authors at the call sites, not just the constructors`**
The original's tests pinned what the constructors *build*; nothing forced the production paths to *call* them. Reverting any of the three call sites with the message text left byte-identical passed all 4441 tests. Assertions added to two tests that already drive the real paths — `no_events_still_acknowledges` (`brain.rs`, real cycle) and `a_failed_continuation_tells_the_operator` (`operator.rs`, real HTTP approve, reading the journaled row). The approval-overflow call site remains unpinned and the commit says so; reaching it needs a gated tool call, a scripted model and a parking host.

**`68fa22bf` — `fix(ids): reserve the host author so no teammate can be minted onto it`** — **your Major**
`crate::ports::SYSTEM_AUTHOR` added to `RESERVED_AGENT_IDS` (`types.rs:2941`, now `[&str; 4]`) as a constant reference, and the `:5039` pin updated with `mint_agent_id("System") == "system_2"`. Confirmed before fixing, by writing the assertion first: `left: "system"  right: "system"`. The list's doc now also records why `CONFINED_AGENT_ID` needs no entry — slugs never emit a hyphen.

Also in that commit, from your non-blocking three:
- `the_notice_author_is_the_literal_the_console_keys_on` (`chat_history.rs`) pins that the constant's **value** is the contract with `chat.ts`. Renaming it to `"host"` leaves all four original #966 tests green while the console silently reverts to company bubbles; only this test catches it.
- `frontend/test/unit/chat-system-notice.test.ts` covers the console half — a system-authored row with no `taskId`, the shape the marker tests never drive. Its doc is explicit that the `author === "system"` branch predates #966, so it is a coupling guard rather than a regression guard.
- `AttributionAudit`'s doc now states the figure is not comparable across the cutover.

The `MessageView.channel` point is recorded in the PR body as a second wire-visible change rather than a side effect of the author change.

Verified on the rebased tree: `cargo fmt --all -- --check`, `cargo check --locked --features openhuman,tinycortex --lib` and `--tests` — all clean, and no trace of #1197 now that #1200 (`ef8fe006`) is in `main`.

Issue #1143, opened 2026-08-19
approvals: a workflow run parked before a restart is unreleasable — the queue is empty and the run waits forever


oxoxDev:
## Summary

Four workflow runs on the `smoke1` staging tenant are parked awaiting approval and **cannot ever be released**. The approvals queue is empty, so there is nothing to approve; the runs keep their `pendingApprovals` forever.

Found in the 2026-08-19 QA pass, on a tenant whose product code is identical to `main` (deployed `9d5f50c0`, behind only by CI-only commits).

## What the operator sees

The run drawer for `feature_pipeline` says, in the product's own words:

> Not finished — "backend" needs your approval, so it produced nothing and the steps after it did not run. This run parked 3 approvals. **Approve them in Approvals** and this run continues on its own.
>
> "backend" gated shell, curl — **decide in Approvals: shell, curl, curl**

`shell`, `curl`, `curl` are rendered as links (#1094, merged today). Clicking one navigates to Approvals, which says:

> **All clear** — Nothing is waiting on you.

The workflow header simultaneously advertises `Manual run · 1 awaiting approval`. So the product tells the operator three times to go somewhere, links them to it, and the destination is empty.

## The four stuck runs

| workflow | node | parked |
| --- | --- | --- |
| `daily-agent-harness-tweets-to-gtm` | `fetch` | 2026-08-19 08:36 |
| `github-issues-to-tasks` | `list_issues` | 2026-08-19 06:44 |
| `feature_pipeline` | `backend` | 2026-08-18 15:49 |
| `daily-open-prs-to-slack` | `fetch_prs` | 2026-08-13 13:55 |

`GET {scope}/approvals` returns `[]`. So does every variant — `?status=pending`, `?all=true`. `/approvals/pending` is 405. The console is querying the only surface there is; the data is not in it.

## Root cause

`ApprovalParked` is written at `Durability::Process` (`src/runtime/journal.rs:340`), and the comment states the reasoning:

```rust
// Losing a park loses the *question*: the approval vanishes and the
// agent parks it again on its next attempt. Nothing external fired.
Self::ApprovalParked { .. } => Durability::Process,
```

**That tolerance is correct for a chat turn and wrong for a workflow run.** A chat turn retries and re-parks. A workflow run has already halted at the blocked node — there is no next attempt to re-park it. So when the park record is lost (container replacement on a hosted tenant; all four runs above predate the most recent pod restart), the two halves of the state diverge permanently:

[...]

Issue #978, opened 2026-08-18
workflows: approving a parallel gated node re-dispatches the whole run, so each approval round creates more approvals than it clears


oxoxDev:
## Summary
A workflow whose trigger fans out to N gated `tool_call` nodes cannot be cleared by approving. **Each approval independently re-dispatches the entire run.** The new run executes only the one node its single-use grant covers and re-parks the other N-1. Approving N gives you N new runs and N(N-1) new approvals.

For the three-node case: **3 → 6 → 12 → 24 …**

The operator experience is an approvals queue that grows every time they work through it.

## Reproduced on staging just now
Workflow `daily-sports-news-blog` (marketing tenant): `start` fans out to three `web_fetch` nodes (BBC, ESPN, Guardian), converging on `rank` → `draft` → `send`.

A cold manual run parked all three and executed nothing:
```
runId 01a013e576f7   nodes: []   pendingApprovals: [fetch_bbc, fetch_espn, fetch_guardian]
```

I approved all three, each with the console's default verdict (scope `once`). Within 3 seconds, three new runs appeared — timestamps line up one-to-one with the three POSTs:

| run | started (ms) | nodes run | re-parked |
|---|---|---|---|
| `01a013eb31ee` | 1787040576379 | `fetch_bbc` | 2 |
| `01a013eb36d1` | 1787040577601 | `fetch_espn` | 2 |
| `01a013eb3b81` | 1787040578804 | `fetch_guardian` | 2 |

Net: cleared 3, created 6. No node beyond the fetches ever ran; nothing was delivered.

## Root cause
Every one of the three approvals answered:

```json
{"responses":[{"channel":"operator","text":"Acknowledged."}],"stillAwaiting":0}
```

`stillAwaiting: 0` on **all three**, including the first — while two siblings were still pending on the same run.

#469 makes a turn continue exactly once, when the last decision it parked lands. That accounting is evidently per-node (or per-branch) rather than per-run, so for a parallel fan-out every branch believes it is the last decision and every branch re-dispatches. Combined with #243 — approving does not resume, it mints a single-use grant and re-dispatches — each re-dispatch replays the run from the trigger with exactly one usable grant.

So the two mechanisms compose into amplification: the run is replayed N times, and each replay re-gates the N-1 nodes it has no grant for.

## Scale of the damage on that tenant
- **77 runs** recorded for this one workflow; **73 predate** my test — i.e. accumulated from this loop, not from a schedule (the workflow is `enabled: false`).
[...]

PR #1159, opened 2026-08-19
journal: make a workflow gate's park survive what its resume assumes

### LINKED ISSUE #1145: approvals: decide whether a workflow-gating park is host-durable, or make the re-park assumption true
## Summary

Split out of #1143. That issue stops the **lie** — a run that claims to await an approval the system no longer holds will be reported as stranded rather than actionable. This issue is the **root cause**: deciding whether a workflow run's approval question should survive what an effect commit already survives.

Stated separately and deliberately. #1143's half is a read-side honesty fix and changes no durability guarantee; this one is a durability decision on a data path, and those should be argued on their merits rather than settled inside a bug fix.

## The assumption that does not hold

`ApprovalParked` is written at `Durability::Process` (`src/runtime/journal.rs:340`), with the reasoning stated in the code:

```rust
// Losing a park loses the *question*: the approval vanishes and the
// agent parks it again on its next attempt. Nothing external fired.
Self::ApprovalParked { .. } => Durability::Process,
```

"Nothing external fired" is true and is the property that matters for safety — no effect escapes. The second clause is the problem:

- **A chat turn** does re-park. The agent runs again, hits the same gate, mints a new approval. The operator is asked a second time. Cost: one extra question. The comment's tolerance is exactly right here.
- **A workflow run** does not. It halted at the blocked node when the park happened. There is no next attempt, nothing re-enters the gate, and nothing re-parks. The run's own `pending_approvals` is durable and outlives the question it refers to, so the two halves of the state diverge permanently.

The gap is not that the durability level is wrong in general. It is that one caller's retry behaviour was generalised to a caller that has none.

## Why it matters beyond one stuck run

A blocked run holds its whole downstream. On the tenant where this was found, `feature_pipeline` stopped at `backend` and never ran the frontend, QA, docs or ship steps — and no operator action recovers it, because the decision it is waiting for does not exist anywhere.

The type system already has the concept: `WorkflowApprovalOutcome::unparkable()` (`src/ports/workflow_runner.rs:257`) marks "a call the operator will **never** be asked about" for `ParkFailed` and `Discarded`. A park lost to a restart is a third way into that same state, arrived at after the fact.

## The options, none of them free

1. **Write a workflow-gating park at `Durability::Host`.** Matches the durability the effect commits already carry. Costs a flush on a record written at operator-decision scale — the same trade `GrantConsumed` already accepted and documented. Narrowest change; does not touch chat.
2. **Make the re-park assumption true.** Something re-enters the gate when a run resumes, so the comment's reasoning becomes correct for both callers. Larger, and it means a run has to be resumable at a blocked node, which is closer to #899's continuation work than to a durability change.
3. **Accept the loss and make it loud.** Keep `Process`, and have a run whose park is gone fail explicitly at boot with a reason, rather than sitting in the list. This is #1143's half taken further — honest, but it means a scheduled pipeline can be silently killed by a routine deploy and the answer is "re-run it".

Option 1 is the smallest thing that makes the failure impossible rather than merely visible. Option 3 is the smallest thing full stop, and is what #1143 already half-buys. The choice between them is a real trade and is why this is filed rather than assumed.

## Not a suppression

Recording this explicitly, because the pair could read that way: #1143 alone would leave a live defect wearing a better label. This issue is the fix, #1143 is the honesty, and neither is a substitute for the other. If this is closed as won't-fix, that should be a stated decision about acceptable data loss on a deploy — not a thing that happens by #1143 merging first.

Part of #1143.

M3gA-Mind:
## Summary

`ApprovalParked` is written at `Durability::Process` on the reasoning stated in the code:

```rust
// Losing a park loses the *question*: the approval vanishes and the
// agent parks it again on its next attempt. Nothing external fired.
```

"Nothing external fired" is true and is the safety property. The re-park clause is a property of the **caller**, not of the record, and it was generalised to a caller that has no next attempt.

A **chat turn** does re-park — it re-enters the gate and mints a new approval, so the loss costs one extra question. The tolerance is exactly right there, and that is where the volume is. A **paused workflow run** does not.

This makes a workflow gate's park `Durability::Host`. Every other park is untouched.

## Why the workflow path cannot re-park — verified, not assumed

`workflow_resume`'s module docs state it as the fact its whole design turns on:

> **Resume is a re-run, because a paused run is settled.** A paused tinyflows run is **not suspended**. Nothing holds a task, a connection, or a continuation — the engine returns, the future completes, and the run is over.

So the parked effect is not a record *of* the continuation, it **is** the continuation — it carries the whole trigger input — and `WorkflowGateQueue::rearm` rebuilds the live gate set at recovery from exactly these still-parked lines. Lose the line and the run keeps a durable `pending_approvals` naming a question that exists nowhere: no card to decide, no re-park coming, and the whole downstream of that pipeline held behind it.

## This is less a new guarantee than one the crate already claims twice

`workflow_resume.rs`, listing what its design gets for free:

> **restart durability.** The parked effect carries the whole input, so a host that dies between the park and the approval **loses nothing**: journal replay rehydrates the card and approving it still resumes.

`CompanyCycle::park`, on the single write path into the queue:

> a parked effect is journaled exactly one way and **survives a restart** with its original `ApprovalId` regardless of who decided it.

Both hold for a *process* restart and fail for the host death the first one names, because `Process` is page-cache-resident by definition. So this is an internal inconsistency being closed, not a fresh durability policy being proposed.

## What is deliberately NOT included, and why it would have been a false fix

**An agent node's gated tool call.** Those park inside a workflow run too, and strand it just as badly — they are the `shell`, `curl`, `curl` in #1143's `feature_pipeline`. Flushing them would buy **nothing**. Their continuation needs the workflow id and the trigger input, which the tool-call effect does not carry, so `BlockedNodeQueue` stashes them **in memory** and — unlike the gate queue — does not rehydrate from the journal at all. `blocked_nodes.rs` says so itself:

> A restart in the middle of a blocked node therefore comes back with the `ContinuationQueue` counter re-armed but the stash empty […] That is a strictly worse restart story than the gate path's, and it is the Stage-1 boundary — the durable-card version is Stage 2 territory.

A durable record whose continuation died with the process is a flush that changes nothing, and it would read as a fix. That path is #899 Stage 2.

**This also rules out the discriminator that first looks obvious.** `Effect.run_id` cannot do the job: it is stamped with a *task attempt* id at the dispatch boundary and with a *workflow run* id by `gate_effect`, so `run_id.is_some()` would select card runs and blocked-node parks — the two cases this must not widen to. `WORKFLOW_APPROVE_KIND` is minted in exactly one place, for exactly the parks whose loss strands a run nothing will re-enter.

## The trade, priced

One flush on a record written at operator-decision scale, on the one effect kind a human is about to be asked about — the identical trade `GrantConsumed` accepted four arms up the same `match`, with the same reasoning written out. The host-durable set goes from three unconditional kinds to three plus one conditional arm; the other nine, and every non-workflow park, stay `Process`.

## One honesty note on the blast radius

`Durability::Process` survives a *process* restart, and on the mongodb backend it is an acknowledged insert, which survives a tenant container replacement too. The loss window is a host death on the fs backend, or a primary loss without `j:true` on mongodb. The durability gap is real and worth closing on its own terms — but I would not assume it is the proximate cause of all four stuck runs in #1143 without a separate look at that tenant's backend. This makes the failure impossible; it does not by itself prove what happened on `smoke1`.

Recommendation and full reasoning were [posted on #1145](https://github.com/tinyhumansai/opencompany/issues/1145#issuecomment-5343538738) before any code was written, as the issue asked.

## API Or Behavior Changes

No API change and no wire/schema change — `ApprovalParked`'s shape is untouched, so every line already on disk replays identically. The behaviour change is one `sync_data` / `synchronous=FULL` / `j:true` write on workflow-gate parks only.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — both ways: default features (the bare `Rust` lane's command) exit 0, and `--locked --no-deps --features openhuman,tinycortex --all-targets` exit 0.
- [x] `cargo build --all-targets` — `--locked --features openhuman,tinycortex --all-targets`, exit 0.
- [x] `cargo test` — both configurations: `--locked --features openhuman,tinycortex --tests` → **4518 passed, 0 failed, 3 ignored** (+26 across integration targets); `cargo test --locked` (default features) → **2980 passed, 0 failed**.

`only_a_workflow_gate_park_is_host_durable` pins **both arms in one test**, because the assertion *is* the distinction: the `Host` half alone would pass if every park were flushed — taxing the approval path to fix one caller — and the `Process` half alone passes on the code this replaces.

**Prove-red:** reverting the new arm fails it with *"a workflow gate's park is the run's only continuation — losing it strands the run behind a question that exists nowhere, and nothing re-parks it"* (exit 101).

The existing #392 guard `host_durable_kinds_are_exactly_the_three_that_could_repeat_an_action` is kept **unloosened** — its fixture's `ApprovalParked` carries an ordinary effect and stays on the `Process` side, so the policy pin that exists to notice a kind moving across the line still notices. The new conditional arm is pinned separately rather than by widening that list, which is the assertion that would otherwise have stopped noticing.

## Documentation

`append`'s durability contract updated — it enumerated "the three `Host` kinds" and "the other ten", which this change makes stale. Now states the three unconditional kinds, the one conditional arm and why it is decided by contents rather than tag, and the other nine.

Closes #1145

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow gate approvals now remain available across host restarts.
  * Other parked approvals retain their existing process-lifetime behavior.

* **Documentation**
  * Clarified durability expectations for workflow gates, chat approvals, tool approvals, and other approval types.

### DIFF
--- a/src/runtime/journal.rs
+++ b/src/runtime/journal.rs
@@ -366,6 +366,57 @@ impl JournalRecord {
             Self::GrantConsumed { .. } => Durability::Host,
             // Losing a park loses the *question*: the approval vanishes and the
             // agent parks it again on its next attempt. Nothing external fired.
+            //
+            // **Except for a workflow gate, which has no next attempt (issue
+            // #1145).** The re-park reasoning above is a property of the
+            // *caller*, not of the record, and it was generalised to a caller
+            // that has none. A chat turn re-enters its gate and mints a new
+            // approval, so the cost of the loss is one extra question — the
+            // tolerance is exactly right there, and that is where the volume is.
+            // A workflow run does not: `workflow_resume` turns on the fact that
+            // "resume is a re-run, because a paused run is settled" — the engine
+            // returned, the future completed, and nothing is holding a
+            // continuation. So the parked effect is not a record *of* the
+            // continuation, it **is** the continuation, carrying the whole
+            // trigger input, and `WorkflowGateQueue::rearm` rebuilds the live
+            // gate set at recovery from exactly these still-parked lines. Lose
+            // the line and the run keeps a durable `pending_approvals` naming a
+            // question that exists nowhere: no card to decide, no re-park
+            // coming, and the whole downstream of that pipeline held behind it.
+            //
+            // Not a new guarantee so much as the one this crate already claims
+            // in two other places and did not deliver — `workflow_resume`'s
+            // "restart durability … a host that dies between the park and the
+            // approval loses nothing", and `CompanyCycle::park`'s "survives a
+            // restart with its original `ApprovalId`". Both hold for a *process*
+            // restart and fail for the host death the first one names, because
+            // `Process` is page-cache-resident by definition. The flush is the
+            // same trade `GrantConsumed` accepted four arms up: one flush on a
+            // record written at operator-decision scale.
+            //
+            // **Deliberately NOT extended to an agent node's gated tool call**,
+            // though those park inside a workflow run too and strand it just as
+            // badly. Flushing them would buy nothing: their continuation needs
+            // the workflow id and trigger input, which the tool-call effect does
+            // not carry, so `BlockedNodeQueue` stashes them **in memory** and
+            // — unlike the gate queue — does not rehydrate from the journal at
+            // all (`blocked_nodes`, stated there as the #899 Stage-1 boundary).
+            // A durable record whose continuation died with the process is a
+            // flush that changes nothing, and it would read as a fix. Making
+            // that path survive is #899 Stage 2.
+            //
+            // Keyed on the effect kind rather than on `run_id`, which cannot do
+            // this job: `run_id` is stamped with a *task attempt* id at the
+            // dispatch boundary and with a *workflow run* id by `gate_effect`,
+            // so `is_some()` also selects card runs and blocked-node parks —
+            // the two cases this must not widen to. `WORKFLOW_APPROVE_KIND` is
+            // minted in exactly one place, for exactly the parks whose loss
+            // strands a run nothing will re-enter.
+            Self::ApprovalParked { effect, .. }
+                if effect.kind == crate::runtime::WORKFLOW_APPROVE_KIND =>
+            {
+                Durability::Host
+            }
             Self::ApprovalParked { .. } => Durability::Process,
             // Bookkeeping after the decision. A ghost approval that is approved
             // a second time cannot duplicate the effect, because the effect's
@@ -1769,14 +1820,22 @@ impl RuntimeJournal {
     /// [`JournalRecord::durability`], and this is the single choke point that
     /// passes that decision to the sink:
     ///
-    /// * The three [`Durability::Host`] kinds — `EffectExecuted`,
+    /// * The three unconditional [`Durability::Host`] kinds — `EffectExecuted`,
     ///   `GrantConsumed` and `StandingGrantRevoked` — are on stable storage
     ///   before this returns. So the at-most-once contract holds against
     ///   **losing the machine** for precisely the records whose loss would
     ///   repeat an external action, and a failed flush fails the append — which
     ///   aborts `execute_effect_once` before `perform_effect` and so cannot
     ///   produce the duplicate it is guarding against.
-    /// * The other ten are [`Durability::Process`]: killing the process cannot
+    /// * `ApprovalParked` is [`Durability::Host`] **for a workflow gate only**
+    ///   (issue #1145), and `Process` for every other park. It is the one kind
+    ///   whose level is decided by its contents rather than by its tag, because
+    ///   the reasoning behind `Process` is a property of the caller: a chat turn
+    ///   re-enters its gate and re-parks, a paused workflow run has already
+    ///   settled and never will. For that run the parked effect *is* the
+    ///   continuation, so its loss strands the run permanently rather than
+    ///   costing a second question.
+    /// * The other nine are [`Durability::Process`]: killing the process cannot
     ///   lose them, a host crash can. That is the decision, not a gap left open.
     ///   Losing any of them makes the runtime **re-ask** — an approval is parked
     ///   again, an operator is prompted again, a cycle bracket reads as
@@ -3610,6 +3669,12 @@ mod test {
     /// across the line: flipping `EffectExecuted` to `Process` compiles, passes
     /// every other test in this file, and silently gives up the one guarantee
     /// the journal exists for. This is the test that notices.
+    ///
+    /// `every_record_kind`'s `ApprovalParked` sample carries an ordinary effect,
+    /// so it belongs on the `Process` side here. Its workflow-gate arm is the
+    /// one kind whose level depends on contents rather than tag, and it is
+    /// pinned separately below (issue #1145) — deliberately not by loosening
+    /// this list, which is the assertion that would have stopped noticing.
     #[test]
     fn host_durable_kinds_are_exactly_the_three_that_could_repeat_an_action() {
         let all = every_record_kind();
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

oxoxDev (approved):
Approving. Correctly targeted one-arm durability change, and the reasoning holds up against the code rather than just being asserted.

**The recovery claim is verified.** `builder.rs:1760-1768` (boot) and `cycle.rs:1305-1315` (recover) both rebuild `WorkflowGateQueue` from `journal.pending()`, so the journal line genuinely *is* the sole rehydration source — `Process` durability was a real gap, not a theoretical one. `WorkflowGateQueue::rearm` (`workflow_gates.rs:234`) replaces per turn rather than accumulating, so a double replay is safe.

**The discriminator is sound.** `WORKFLOW_APPROVE_KIND` is constructed in exactly one place (`workflow_resume.rs:645`, inside `gate_effect`); every other reference is a comparison. And the rejected alternative is correctly rejected — `run_id` is stamped both at the task-attempt dispatch boundary and by `gate_effect`, so keying on `is_some()` would have widened to card runs and blocked-node parks.

**Excluding blocked-node tool-call parks is right, not a gap.** `builder.rs:1777-1783` confirms `BlockedNodeQueue` is a bare `default()` on boot and never rehydrated from the journal, so flushing those records would produce a durable line whose continuation died with the process. Good call deferring to #899 Stage 2 and saying so in-code rather than leaving a TODO.

**Ordering is correct.** `CompanyCycle::park` (`cycle.rs:2282`) writes the journal line, then arms the continuation counter and emits the advisory event, both explicitly after the write — so a crash between them replays as "still parked". One pre-existing wrinkle you don't touch: `rt.approvals.park()` runs *before* the journal write, so host death in that window leaves an approvals-store entry with no journal line. Not a regression, and the journal is documented as the binding record.

**On double-application:** can't happen, and this PR doesn't change that. Idempotency is enforced upstream by the `approval:<id>` effect key and the `EffectExecuted` record — one of the three unconditional `Host` kinds, so the at-most-once guard was already flushed. `on_gate_approved` only banks the decision, with the spawn moved to the batch release in `continue_turn` under #978. A stale result can't resurrect a drained park either: `append` holds the write lock *around* the store call (`journal.rs:1856-1862`) precisely so a park can't be sequenced after the resolution that drains it.

**On the durability window:** your honesty note is accurate. Before, a process restart survived but host death didn't — `Durability::Process` is page-cache-resident by definition (an `O_APPEND` write with no `sync_data`, sqlite at `synchronous=NORMAL`, mongodb without `j:true`). After, workflow gates get `sync_data` / `synchronous=FULL` / `j:true`. On the mongodb backend an acknowledged insert already survived container replacement, so what's actually closed is host death on fs, or primary loss without `j:true`. Declining to claim this explains all four stuck runs in #1143 is the right level of confidence.

This fixes the cause rather than the symptom — a run stuck behind a card that doesn't exist gets fixed by making the record that *is* the continuation durable, not by patching recovery. And keeping the #392 policy pin unloosened while pinning the new arm separately is the right structure, since widening that list is exactly the assertion that would stop catching future drift.

Two notes, neither blocking:

- `effect.kind` is a `String` an agent can influence, so a tool named `workflow.approve` would take the flush path and already hits `cycle.rs:1490`'s `on_gate_approved`. Pre-existing and out of scope here, but if that namespace ever becomes agent-reachable it wants a reserved-prefix guard.
- `journal.rs:369-414` carries 45 lines of comment for a 5-line arm. Consistent with the file's house style — just noting the "#899 Stage 2" and `blocked_nodes` cross-references will rot when that lands.

CI: the `UNSTABLE` was pending, not broken — the gated Rust suite, gated host binary and Console E2E were still in progress, everything else green, and the Console jobs correctly `skipping` via the `Changed areas` gate on a Rust-only diff. Main's earlier red (`4bca2ab7` / #1126, a `src/lib/team.ts` TS6133) was already fixed by #1158 at 15:10, so nothing here is attributable to this PR.

PR #1160, opened 2026-08-19
fix(harness): tell a teammate the name the console calls it (#1105)

### LINKED ISSUE #1105: harness: a teammate is never told its own name, so it denies being the person the console says it is
## Summary

A teammate the operator added as **Alex** (role: Content Writer) does not know it is Alex, and says so:

> **You** — hey alex
> **Alex** — Hey! I'm actually the Content Writer here, not Alex — but happy to help regardless.

The console is naming this teammate Alex all over the screen: the DM header says **Alex**, the subtitle says *"This is the start of your direct message with Alex — content Writer."*, the composer placeholder says **Message Alex**. The model is told none of it, so it contradicts the interface it is speaking through — which reads to an operator as an agent confused about who it is.

Its answer is *correct given its prompt*. The name never reaches the model.

## Where the name is lost

Two sites, both in the host.

**1. The persona prompt never mentions a name** — `src/company/prompt.rs`:

```rust
let mut prompt = format!(
    "You are the {role} at {company}. Speak in the first person as this role.",
    role = agent.role,
    company = company_name,
);
```

Only `role`, then `description`, then any custom `prompt`. There is no name in the string, and `Agent` (`src/company/types.rs`) has no `name` field to put there — a manifest `[[agent]]` is identified by `id` + `role` only.

**2. The overlay teammate's name is dropped on the way in** — `overlay_agent_to_manifest` in `src/harness/mod.rs`:

```rust
ManifestAgent {
    id: overlay.id.clone(),
    role: overlay.role.clone(),
    description: overlay.description.clone(),
    …
}
```

An operator-added teammate is an `OverlayAgent`, and **`OverlayAgent.name` is the only place "Alex" exists**. The conversion copies id, role, description and tools; the name has nowhere to go, so it stops here. Everything downstream — the persona, the turn, the reply — is built from a struct that has never heard of Alex.

Meanwhile the console reads that same `overlay.name` for every label the operator sees (`roster_display_names` in `src/metering/mod.rs` maps overlay ids to `member.name`, manifest ids to `role`). So the two halves of the product disagree by construction: **the UI addresses by name, the model reasons by role.**

## Suggested direction

The fix is to carry the name into the persona, but it forces a decision the issue should settle first:

1. **What is a manifest teammate's name?** Manifest agents have only a role, so either they gain an optional `name`, or the persona says the name only when one exists and keeps today's wording otherwise. The second is smaller and avoids a manifest schema change.
2. **Where the name enters.** `persona_prompt(company_name, agent)` currently takes only the agent, so either `Agent`/`ManifestAgent` gains an optional `name` (which `overlay_agent_to_manifest` then fills), or the signature takes the display name alongside. The first keeps one source of truth; the second avoids touching a serialised type.
3. **The wording.** Something like `"You are Alex, the Content Writer at Acme."` — the model should answer to the name *and* keep the role, not swap one for the other.

Worth stating in whatever lands: the name is an addressing handle, not an identity claim. A teammate should not be encouraged to invent a persona around it.

## Reproduce

1. Team → add a teammate with a name (e.g. **Alex**) and a role (**Content Writer**).
2. Open its DM.
3. Address it by name: `hey alex`.
4. It denies being Alex while the header, subtitle and composer all call it Alex.

sanil-23:
## Summary

An operator adds a teammate called **Alex**, opens its DM, says `hey alex` — and it replies *"I'm actually the Content Writer here, not Alex."* It is right, given its prompt: the name never reached the model.

The name lives in exactly one place, `OverlayAgent.name`, and `overlay_agent_to_manifest` dropped it on the way into the roster — its own doc comment declared that intentional ("The overlay's `name` is a display label only … so the persona is framed from `role`/`description` alone"). Meanwhile the console labels that teammate from the same field everywhere: DM header, subtitle, composer placeholder. So the two halves disagreed by construction — the UI addressed by name, the model reasoned by role.

This carries the name through to the persona:

1. `company::Agent` gains `name: Option<String>`.
2. `overlay_agent_to_manifest` fills it, and its doc comment now says what the code does.
3. `persona_prompt` frames a named teammate as *"You are Alex, the Content Writer at Acme. … Teammates and the operator address you as Alex; it is how you are called here, not a separate character to play."*

The name is stated as an **addressing handle, not an identity claim**, which is what the issue asked for — a teammate should answer to it without inventing a persona around it. The role is never replaced.

Closes #1105.

## API Or Behavior Changes

**Behavior:** an operator-added teammate's persona now names it. Manifest teammates are unchanged — they are addressed by role, and their persona string is byte-for-byte what it was. Two guards keep it that way: a blank/whitespace name falls back to the role-only wording, and so does a name that only restates the role (otherwise: *"You are Content Writer, the Content Writer at Acme."*).

**API:** one new public field, `company::Agent::name`. It is `#[serde(skip)]` — an in-memory carrier filled by the roster build, deliberately not a manifest key:

- **No migration, and nothing to migrate.** The serialized shape of `Agent` is unchanged, so existing `company.toml` and `agents/*.toml` parse identically and stored records round-trip exactly as before.
- No `[[agent]]` block or per-agent file in `companies/` or `globals/` uses a `name` key today, and `AgentFile` has no such field — so nothing authored changes meaning.
- Making it authorable would mean adding it to `AgentFile` too, or a `name` in an `agents/*.toml` would be silently ignored while the same key worked in `company.toml`. That is a schema decision for a follow-up, not a side effect of this fix.

The overlay→manifest conversion is transient (never written back into `manifest.agents`), so the name keeps living in the persisted `OverlayAgent` and is re-derived on every roster build.

## Tests

- [x] `cargo clippy --features openhuman --all-targets -- -D warnings`
- [x] `cargo check --features openhuman --tests`
- [x] `cargo test --features openhuman` — **4429 passed, 0 failed**

Five new tests: the persona names the teammate *and* the role; an unnamed teammate keeps today's exact string (pinned literally, so a future edit to that arm fails loudly); a blank name falls back; a name restating the role falls back; and the conversion carries the name through to the persona end-to-end.

## Documentation

`docs/spec/runtime/agents.md` gains "A named teammate is told its name", beside the existing rule that `prompt` is appended rather than substituted.

## Noted while here — not fixed

`query_company` pushes `overlay.name` into the **id** slot of its Team list (`src/harness/orchestrator.rs`), while `delegate_to_teammate` grounds targets by id only. It usually survives because an id is a slug of the name, but a **renamed** teammate is listed under a name that grounds to nothing, since `mint_agent_id` deliberately does not re-mint on rename. Separate bug, worth its own issue.


sanil-23:
The `query_company` id/name-slot bug noted at the end of the description is now filed separately as #1162 — it turned out to be broader than described here: it breaks for *any* operator-added teammate whose display name is not already its slug (`mint_agent_id("Dana Designer") == "dana_designer"`), not only renamed ones. Nothing in this PR depends on it.

sanil-23:
@coderabbitai review

Issue #1104, opened 2026-08-19
console: graph labels are all-or-nothing — anonymous circles at rest, and an unreadable pile-up inside a focused pillar


oxoxDev:
## Summary

Two complaints about the Overview knowledge graph that turn out to be **one rule failing at both ends**:

1. **At default zoom, the agent nodes around "This company" are anonymous circles.** You have to click in to find out who is who.
2. **Zoom into a pillar cluster and the stage / tool / agent labels pile on top of each other.** Unreadable — there is no collision avoidance on a graph that dense.

Asked for: hovering an icon should show that node's name.

## The rule behind both

`frontend/src/views/overview/kg/KnowledgeGraph.tsx`:

```ts
const showLabel =
  n.kind === 'self' || n.kind === 'team' || inFocus ||
  (hoverId ? (lit?.has(n.id) ?? false) : false);
```

- **At rest** only `self` and `team` are labelled, so every agent, tool, task and stage renders as a bare circle. That is complaint 1, exactly.
- **In focus** `inFocus` is true for *every node in the tree at once*, so the entire cluster's labels turn on simultaneously. That is complaint 2.

The only mitigation today is `labelDy` — within a depth band, labels alternate between two heights (`i % 2 === 0 ? 0 : 11`) sorted by x. That is a two-row stagger, not collision avoidance: it assumes labels are short and siblings evenly spaced, and it does nothing about a third overlapping neighbour or a long name.

Worth knowing while fixing: labels are deliberately **constant on-screen size** at every camera depth (`fixedLabel` counter-scales through `--kg-cam-k`), so text does not shrink as density rises — zooming changes node spacing, never label footprint.

## On the hover request

Hover labelling already exists, but it is **chain-wide, not node-scoped**: `litFor(id)` returns the hovered node *plus its whole pillar chain* (`lit` then drives `showLabel`). So pointing at one agent lights a crowd of labels — which at rest is more than asked for, and inside a focused cluster adds to the pile rather than cutting through it.

What was asked for is narrower and, I think, better: **hovering an icon names that icon.**

## Suggested direction

1. **Node-scoped hover label** — hovering any node shows its own name, at rest and in focus, independent of the chain lighting. This alone fixes complaint 1 without adding any resting clutter.
[...]

PR #1164, opened 2026-08-19
docs(kg): describe the label rule the code actually implements

graycyrus:
Follow-up to #1148, which implemented the label rationing and is merged. This carries only the documentation corrections that landed after it.

**Two claims in `overview/README.md` were stronger than the code** — CodeRabbit caught both on review:

- *"hovering names the node under the pointer, **and only that node**"* — `KnowledgeGraph.tsx` raises only the hovered node to `LABEL_PRIORITY.hovered`; the resting labels stay put. Hover **adds** a name, it does not reduce the graph to one.
- *"Selection **always** keeps its name"* — `LABEL_PRIORITY.selected` is 900 against `hovered`'s 1000, so a selected label loses a collision to a hovered one. False in exactly the case a reader would care about.

Plus one stale doc reference in `label-plan.ts`: the comment said `{@link LABEL_GAP}`; the constant is `LABEL_GAP_X`.

Documentation that asserts stronger invariants than the implementation is worth correcting on its own — it gets read later as a spec, and the code gets "fixed" toward it.

## Correction to this PR's own history

I opened this believing #1104's work was orphaned by a session limit. It was not: PR #1148 had already been raised from the same branch and merged at 14:05, and what remained unpushed was these doc commits on top. The original body described the whole label rule and was wrong; this is the accurate scope.

## Verification

`npm run typecheck` clean · `npm test` **1228 passed**.

Browser check of the merged behaviour (on this branch, which contains #1148): at rest 16 labels — the company, all three departments, and the full roster — and a bounding-box test across all pairs returned **0 overlaps**, so the collision pass holds. Hover and focused-cluster remain unverified; my probe hovered a non-label-bearing element and proved nothing either way.


graycyrus:
Fixed in the commit above — both claims were genuinely stronger than the code, thank you.

To be explicit about what was wrong rather than just applying the diff:

- **Hover.** `KnowledgeGraph.tsx` raises only the hovered node to `LABEL_PRIORITY.hovered`; the resting labels stay where they are. So hover *adds* a name, it does not reduce the graph to one name. The old wording described a rule the code never implemented.
- **Selection.** `LABEL_PRIORITY.selected` is 900 and `hovered` is 1000, so a selected label loses a collision to a hovered one. "Always keeps its name" was false in exactly the case a reader would care about — pointing at something else.

Also merged `upstream/main` to clear the conflict. Local re-verify on the merged tree: `npm run typecheck` clean, `npm test` **1228 passed**.

Still outstanding and worth a reviewer knowing: **this PR has not been checked in a browser.** The implementing agent was interrupted before it could, so the visual result at rest, on hover and inside a focused cluster is unverified. I would rather say that than let green tests imply otherwise.

graycyrus:
## Browser check done — the gap I flagged is now closed

Ran a host from this branch (loopback, own data dir, `agentic_software_company`), signed in, drove the Overview graph with Chromium at 1440×900.

**At rest, 16 labels:** `This company`, all three departments (Engineering, Product & Design, Go-to-Market), and the roster — Backend Engineer, Frontend Engineer, Security Engineer, Writer, Researcher, Page Builder, Operations, Documentation Writer, Developer Relations, Customer Support, Designer — plus the human `mithil`. That is exactly the intended rule: ring-1 named, the numerous deeper tiers bare.

**Collision pass measured, not eyeballed.** I computed the bounding box of every rendered label and tested all pairs for intersection:

```
{ total: 16, overlapping: 0, pairs: [] }
```

Zero label-on-label overlaps. I had initially read the screenshot as a pile-up and was wrong — the measurement corrected me.

**One honest observation, not a blocker:** labels do cross *node glyphs and edges* in the denser quadrant (e.g. the "Engineering" label sits over a nearby circle). That is a different and milder problem than the label-on-label collision this issue was filed about, and the declutter pass deliberately scores label boxes rather than node art. Worth its own issue if it bothers an operator; I have not filed one.

**Not covered:** the hover case. My probe hovered an `svg circle` that turned out not to be a label-bearing node, so the label count did not change — that proves nothing either way about node-scoped hover, and I would rather say so than count it as verified. The focused-cluster case is likewise unverified.

Local on this head: `npm run typecheck` clean, `npm test` **1228 passed**.

oxoxDev (approved):
Approving. Both corrections are right, and I checked them against the code rather than the prose.

- `LABEL_PRIORITY` at `label-plan.ts:58-62` is `hovered: 1000`, `selected: 900`. So "Selection **always** keeps its name" was indeed false in exactly the case a reader cares about, and "ranks just below hover" is accurate.
- `LABEL_GAP_X = 3` at `:89`, used at `:101`. There is no `LABEL_GAP`, so the `{@link}` was dangling and now resolves.

The framing in the body is the part I'd underline: documentation that asserts stronger invariants than the implementation gets read later as a spec, and then the code gets "fixed" toward it. Worth correcting on its own, agreed.

Also appreciated the correction to the PR's own history — noting that #1148 had already merged from the same branch, rather than quietly re-scoping. And the honesty about the browser check is the right level of precision: 16 labels at rest with 0 overlaps across all pairs is a real result for the collision pass, and saying that the hover probe hit a non-label-bearing element and proved nothing is more useful than claiming hover was verified.

One nit: the reflow lost a wrap. The third line of that paragraph now runs to ~120 chars —

```
tooltip is the floor wherever a drawn label is suppressed. Tasks, workflow stages and tools are the numerous tiers, which is
```

— where the surrounding prose wraps at ~76. Worth re-wrapping to match the file.

**On CodeRabbit's CHANGES_REQUESTED — I'd disregard it.** Its single actionable comment asks for this exact change: its proposed wording for both the hover row and the selection sentence is what the PR already contains, essentially verbatim. It's re-flagging a fix that's already applied, evidently reviewing against the base rather than the head. Its four nitpicks all target `KnowledgeGraph.tsx`, which isn't in this diff at all (this PR is `README.md` + `label-plan.ts`) — that's #1148's merged content. The `selectedOrgId` hoist it suggests may well be worth doing, but not here.

Worth knowing that's the third distinct way that bot has produced a misleading signal in this repo today — reporting `pass` while rate-limited without reviewing, approving against a pre-force-push SHA, and now requesting changes for a change already made.

`Console E2E (live brain)` is still pending; a doc-only diff can't affect it.

PR #1179, opened 2026-08-19
fix(chat): draw the teammate's avatar in a DM header

### LINKED ISSUE #1170: chat: a DM header shows a generic circle where the sidebar shows the teammate's avatar
Open a DM and the teammate has two different faces on the same screen: the sidebar row shows their mascot avatar, and the header above the conversation shows a generic outline circle.

```
sidebar                     header
 (◕) Backend Engineer        ⊙  Backend Engineer   │ Backend Engineer
  ↑ the teammate's avatar    ↑ a generic CircleDot glyph
```

## Cause

`frontend/src/views/chat/ChatHeader.tsx:101-105`:

```ts
function KindIcon({ channel }: { channel: Channel }) {
  if (channel.kind === "dm") return <CircleDot className={cls} aria-hidden />;
  …
  return <Hash className={cls} aria-hidden />;
}
```

A DM gets `CircleDot`. The sidebar row draws the real `Avatar` for the same teammate, so the two disagree about who you are talking to.

`#` for a channel is right — a channel has no face. Only the DM case is wrong.

## Fix

Render the same `Avatar` the sidebar uses when `channel.kind === "dm"`, seeded identically so the mascot matches. Avatar assignment is `avatarFor(seed)` in `lib/team.ts`, seeded on the agent id where there is one — the header must use the same seed as the row, or the two will show different mascots, which is worse than the current inconsistency.

Keep `Hash` for channels and `Lock` for private ones.

## Worth checking while there

- The header repeats the name twice — title, then the same string again in muted text after a divider (`ChatHeader.tsx:78`). For a DM that second slot looks like it wants the teammate's role or description, not a repeat. Worth a look, though it is not what this issue is about.
- The avatar in the header should carry the same `markOnly` reasoning as elsewhere: below roughly 24px the mascot is a smudge and the tone tile is the honest fallback.

graycyrus:
## Summary

A DM showed the teammate twice with two different faces: the rail row drew their
mascot `Avatar`, and the header a few pixels away drew a generic `CircleDot`.
The header now draws the same mascot.

`#` for a channel and `Lock` for a private one are untouched — a channel has no
face, so a glyph is the honest mark for it.

## The seed, and why it is a shared helper

`Avatar` hashes its mascot out of the `name` it is handed. Seeding the header on
anything else — the teammate's id, its role — would have put *two different
mascots* on one person on one screen, which is worse than the glyph it replaced.
So the seed is not written out twice: both call sites go through one pure
`dmFace(channel)` in `views/chat/model.ts`, which returns `{ name, tone }` for a
DM with a roster entry and `null` for everything else (a channel, and a DM whose
teammate the roster cannot name — both keep their glyph). Drift is now
structurally impossible rather than merely watched for.

This deliberately seeds exactly the way the rail row already did — on the name.
`TeamMember.avatar` is id-seeded and currently rendered by nothing; using it here
would have made the header disagree with the rail. Unifying that properly is
#1185, and this PR stays out of its way.

Size: `size-6` (24px) rather than the rail's 20px. The header has the room, and
24px is the floor below which `Avatar`'s `markOnly` says a mascot is a smudge and
the bare tone tile is the honest fallback — so at 24px the drawing is worth
rendering.

## Verified in a browser

Host on `:8364` (`companies/agentic_software_company`), console on Vite, on the
merged tree at `ca5ed29`.

A screenshot showing *a* mascot would prove nothing — the seeds have to match. So
the check walks **every teammate in the roster**, opens their DM, and compares the
`src` of the header's `` against the `src` of that same teammate's rail row:

```
MATCH  Backend Engineer      header=/avatars/blob-indigo.webp   row=/avatars/blob-indigo.webp    px=24/20
MATCH  Customer Support      header=/avatars/blob-blue.webp     row=/avatars/blob-blue.webp      px=24/20
MATCH  Designer              header=/avatars/blob-graphite.webp row=/avatars/blob-graphite.webp  px=24/20
MATCH  Developer Relations   header=/avatars/blob-green.webp    row=/avatars/blob-green.webp     px=24/20
MATCH  Documentation Writer  header=/avatars/blob-cloud.webp    row=/avatars/blob-cloud.webp     px=24/20
MATCH  Frontend Engineer     header=/avatars/blob-green.webp    row=/avatars/blob-green.webp     px=24/20
MATCH  Product Manager       header=/avatars/blob-blue.webp     row=/avatars/blob-blue.webp      px=24/20
MATCH  QA Engineer           header=/avatars/blob-ember.webp    row=/avatars/blob-ember.webp     px=24/20
MATCH  Security Engineer     header=/avatars/blob-rose.webp     row=/avatars/blob-rose.webp      px=24/20
MATCH  Operations            header=/avatars/blob-clay.webp     row=/avatars/blob-clay.webp      px=24/20
MATCH  Page Builder          header=/avatars/blob-ember.webp    row=/avatars/blob-ember.webp     px=24/20
MATCH  Researcher            header=/avatars/blob-indigo.webp   row=/avatars/blob-indigo.webp    px=24/20
MATCH  Writer                header=/avatars/blob-cloud.webp    row=/avatars/blob-cloud.webp     px=24/20

light: 13/13 match; 8 distinct mascots across the roster
dark:  13/13 match; 8 distinct mascots across the roster
```

**The two mascots are the same file, for every teammate, in both themes.** The
8-distinct count is there so this cannot pass by everyone collapsing onto one
face. A channel header is unchanged: it still reports `lucide-hash size-4` and no
``.

## Commands run

- `npm run typecheck`
- `npm run typecheck:unit`
- `npm test` — 133 files, 1293 tests, including the 6 new `dmFace` cases

## Follow-up filed separately

The header prints the DM's name and then the teammate's role after a divider, and
a teammate the host sends no display name for takes its role *as* its name
(`fromDto`: `dto.name?.trim() || dto.role`), so the bar reads "Backend Engineer │
Backend Engineer". Out of scope here — filed as #1180.

Closes #1170

### DIFF
--- a/frontend/src/views/chat/ChannelRail.tsx
+++ b/frontend/src/views/chat/ChannelRail.tsx
@@ -4,7 +4,7 @@ import { ChevronRight, CircleDot, Hash, Lock, SquarePen } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import { cn } from "@/lib/utils";
 import { Avatar } from "./Avatar";
-import type { Channel, ChannelSection } from "./model";
+import { dmFace, type Channel, type ChannelSection } from "./model";
 
 /**
  * What an unread badge actually claims (issue #364).
@@ -178,8 +178,9 @@ function ChannelRow({
 
 function ChannelIcon({ channel }: { channel: Channel }) {
   if (channel.kind === "dm") {
-    return channel.member ? (
-      <Avatar name={channel.name} tone={channel.tone} className="size-5 text-3xs" />
+    const face = dmFace(channel);
+    return face ? (
+      <Avatar {...face} className="size-5 text-3xs" />
     ) : (
       <CircleDot className="size-4 shrink-0" aria-hidden />
     );
--- a/frontend/src/views/chat/ChatHeader.tsx
+++ b/frontend/src/views/chat/ChatHeader.tsx
@@ -3,7 +3,8 @@ import { Check, CircleDot, Copy, Hash, Lock, PanelLeft, Users } from "lucide-rea
 
 import { Button } from "@/components/ui/button";
 import { cn } from "@/lib/utils";
-import { channelTitle, type Channel } from "./model";
+import { Avatar } from "./Avatar";
+import { channelTitle, dmFace, type Channel } from "./model";
 
 interface Props {
   channel: Channel;
@@ -18,10 +19,10 @@ interface Props {
 /**
  * The bar above a channel's timeline.
  *
- * It is deliberately thin — a kind icon, the name, and the purpose in the
- * tooltip — with the two things you actually reach for on the right: the
- * member pane and a copy of the channel name. The copy button only appears on
- * hover so the title reads clean at rest.
+ * It is deliberately thin — a kind mark (the teammate's avatar on a DM), the
+ * name, and the purpose in the tooltip — with the two things you actually
+ * reach for on the right: the member pane and a copy of the channel name. The
+ * copy button only appears on hover so the title reads clean at rest.
  */
 export function ChatHeader({
   channel,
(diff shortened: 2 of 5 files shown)


graycyrus:
Follow-up for the repeated name in the header filed as #1180.

graycyrus:
@coderabbitai review

graycyrus:
Merged `upstream/main` (through #1154) into the branch — clean, no conflicts. Re-ran the gates on the merged tree: typecheck, typecheck:unit, and 1293 unit tests all pass, and the browser check was redone against `ca5ed29`, walking all 13 teammates in both themes.

CI on head `ca5ed295c600e38799cc5b20df29071315b8dcff`: every lane green, zero failures. (The `Desktop` cancellation on the previous head was an apt-install flake in `Install webkit and GTK`, before the job reached any build step; it passes on this head.)

Note on CodeRabbit: it returned "Review limit reached" rather than a review, both automatically and on an explicit `@coderabbitai review`. That is an account quota, not a verdict on this diff — **this PR has not been reviewed by CodeRabbit.** tinysweeper approved with 0 findings across all lanes.

PR #1188, opened 2026-08-19
console: an inset, centred app frame with the colour on the outer surface

### LINKED ISSUE #1178: console: an inset, centred app frame with the colour on the outer surface
Give the console an inset, centred app frame — the content sits in a rounded, elevated container with breathing room on top, left and right, and the surrounding surface carries the colour.

Reference the operator supplied (a desktop chat app): the window's outer surface is a soft tinted ground; the working area floats inside it as a rounded card with a visible edge and a slight lift; the sidebar shares that inset frame rather than running to the window edge; and the main column is a white/near-white panel set inside the tint.

## Why this is more than a margin

Today every surface runs edge-to-edge, so the sidebar, the header and the content all meet the viewport with nothing separating the app from the window. An inset frame does three things at once: it makes the working area read as a single object, it gives status colour somewhere to live that is not the content, and it stops long lines running the full width of a wide display.

## What the design has to answer

1. **Where the colour goes.** The outer ground is the only new surface; it must come from existing semantic tokens, and it has to work in both themes — `.dark` re-declares the full token set, and `scripts/ci/assert-design-tokens.sh` fails on raw hex or raw Tailwind palette classes. A tint that reads as "warm paper" in light mode must have a dark-mode counterpart that is not simply lighter black.
2. **What happens to the scroll containers.** The console already has several full-height panels (chat timeline, the workflows canvas and its rails, the ledgers board, the workspace tree). An inset frame changes what "100% height" means for each; the board and the canvas are the risky ones because they scroll horizontally.
3. **Narrow viewports.** Below some width the inset costs more than it gives — the frame should collapse to edge-to-edge rather than squeezing content into a thin column inside a thick border.
4. **The desktop shell.** Tauri renders this in a native window; an inset frame there interacts with the title bar and must not produce a double border.

## Deliverable

A **draft PR** with screenshots, so the direction can be judged before it is finished. Verify in a real browser at 1440 and a narrow width, in **both themes**, across at least: Overview, Chat, Workflows (canvas + rail), Ledgers board, and a Settings sub-page.

graycyrus:
Closes #1178.

The console stops being a wallpaper and becomes an object. A tinted ground fills
the window; the app sits inset inside it as one rounded card holding the sidebar
and the working column together, instead of each of them meeting the window on
its own.

Draft, because the direction is the thing to judge. The rough edges are named at
the bottom rather than left for you to find.

## The shape: inset on three sides, flush to the bottom

The issue asked for "breathing room on top, left and right", and I took that
literally rather than as shorthand for a margin. Three reasons:

- **Vertical space is the scarcest thing here.** Chat's composer, the ledgers
  board and the workflows canvas all want every pixel of height, and a bottom
  gutter costs each of them twice — once for the gap and once for the corner
  radius above it. Only one inset comes out of `100svh`, not two.
- **A card that meets the bottom edge still reads as an object.** The three
  edges that *are* inset are the three the eye tracks. It is what Slack and the
  macOS sidebar apps the reference is drawn from do.
- **On the desktop it is free.** A macOS window rounds its own bottom corners,
  so the OS supplies the radius our card deliberately does not draw.

| Token | ≥ `lg` | ≥ `2xl` | Below `lg` |
| --- | --- | --- | --- |
| `--app-frame-inset` | 12px | 18px | 0 |
| `--app-frame-radius` | 14px | 14px | 0 |
| `--app-frame-border` | 1px | 1px | 0 |

All three collapse together, and that mattered: my first pass kept the radius at
zero inset and the corners rounded *against the window edge*, letting the ground
show through as two stray tinted notches. Below `lg` there is no frame — not a
flattened one.

## Where the colour lives

One new rung on the surface ladder, `ground`, and it is the only rung that is not
inside the app.

| | Light | Dark |
| --- | --- | --- |
| `--app-ground` | `#EBE7F7` (L 0.936, C 0.022) | `#15131B` (L 0.193, C 0.017) |
| `--app-frame-edge` | rung `4` `#DADAE5` | rung `bg` `#08090B` |

**It is the most chromatic neutral in the system, on purpose.** It is the one
surface that carries no text, no stroke and no control, so nothing on it can be
harmed by a tint — and a grey there reads as an empty margin rather than as a
ground the app is standing on. Hue 296.5° in both themes, ~10° warmer than the
neutral ladder's 285–286°.

**On "warm paper": I went as warm as this palette honestly goes, and no
further.** A genuinely warm ground (amber, 70–90°) would put a second hue family
into a console that owns exactly one, and it would go muddy precisely where it
meets the cool greys — at the frame's edge, which is the only place this colour
is ever seen against another. So the ground takes the brand's own hue, pushed
warm within the family and dropped to near-neutral chroma. It reads as tinted
paper rather than as violet.

**The two themes take opposite sides of their canvas, and that is the real
decision.** Light sits a step *below* the canvas (0.936 vs 0.978), so the app
stays the brightest thing on screen. Dark sits a step *above* it (0.193 vs
0.140), because there is no rung left below near-black: darker is `#000`, and at
that lightness the chroma is invisible, so "the ground" would read as nothing at
all — the "lighter black" the issue warned about, in its other form. Going up
also gives the shadow something to fall on, which is the first time in dark mode
this system has had that. The rule that surfaces climb toward the viewer governs
panels stacked *inside* the app; the window ground is not in that stack.

`--app-frame-edge` is `--border` in light but the *canvas* value in dark. In dark
the ground is lighter than the app, so rung 4 (`#161719`) sits between the two
and reads as a third surface wedged into the seam; the canvas value disappears
against the content it borders and stays a dark line against the ground and the
sidebar. A raised edge on a dark screen is an absence of light, not a stroke.

## The judgement call you may want to veto: `--sidebar` moves to rung 2

Light mode's sidebar was rung 1, pure white — one step *above* the canvas beside
it — while the stylesheet's own comment on that line claimed it was "slightly
recessed from the canvas so the working area is the brightest thing on screen".

Nothing depended on the contradiction until the app gained a ground. Now there is
a ladder running from the window inward and it has to be monotonic, or the frame
reads as three unrelated greys rather than as an app sitting on a desk:

```
light   ground #EBE7F7 → sidebar #F1F1F8 → canvas #F7F7FC → card #FFFFFF
dark    ground #15131B → sidebar #0C0D0F → canvas #08090B → card #0C0D0F
```

Dark already ran in the right order. Light is the theme that was inverted, and
one value fixes it. It is also the half of the reference that the frame alone
does not deliver — "the main column is a near-white panel set inside the tint".

**No contrast regression is possible**, because rung 2 sits *between* two grounds
the ink ramp was already tuned against:

| Ground | primary | secondary | tertiary | hint | muted |
| --- | --- | --- | --- | --- | --- |
| canvas `#F7F7FC` | 18.44 | 9.04 | 7.38 | 6.09 | 5.05 |
| **sidebar `#F1F1F8`** | **17.51** | **8.58** | **7.01** | **5.79** | **4.79** |
| active `#ECE9FC` (documented worst case) | 16.53 | 8.10 | 6.62 | 5.46 | 4.52 |

This is one line and fully reversible. Say the word and it goes back.

## The breakpoint: `lg` (1024px)

`md` (768px) is the obvious candidate and it is the wrong one: that is where the
sidebar becomes an overlay sheet, so a frame surviving down to it would be a
border drawn around a single column — useless, and useless for the whole
768–1024 band on the way there.

1024 is where the inset starts to *cost* instead of give. With the sidebar
expanded (13.5rem) the working column is 779px at 1024 and 523px at 768; the
frame's 24px is 3% of the first and 4.6% of the second, taken from a column that
by then is the only one left. Turning it off at `lg` means the frame is never on
while the console is already short of width — which is the whole argument for
having a breakpoint at all.

## Scroll containers

The shell was already a pure flex chain rooted at one `h-svh`, so most surfaces
absorbed a shorter viewport for free. Two things needed doing, and one thing
needed checking:

- **The sidebar was `position: fixed` against the *viewport*.** That is what an
  inset frame cannot have — 0 and "the left edge of the app" stopped being the
  same number, and the column stood outside the frame. It is now `absolute`
  against the frame. I deliberately did *not* make the frame a containing block
  for `fixed` (a `transform` or `contain: paint` would do it): that recaptures
  every other fixed descendant with it — toasts, the tour overlay, any dialog
  that is not portalled — and clips them to the frame.
- **`Overview` sized itself `h-svh`** and was the only view that did. It
  overflowed its slot by exactly the inset and the shell's `overflow-hidden`
  clipped the bottom of the graph off with no scrollbar to say so. Now
  `flex-1 min-h-0`, and pinned by a test.
- **Checked and unchanged:** the ledgers board (scrolled to its far right,
  including #1121's collapsing columns), the workflows canvas with the run
  rail open, chat's timeline, settings' nav, and the workspace tree. At 1440 in
  both themes I walked the DOM for any element whose box escapes the frame —
  the only hits are the frame's own ancestors. `documentScrollWidth/Height`
  overflow is 0 at every width from 768 to 1600.

## The desktop shell — partially verified, and I will say where it stops

I built and launched the real Tauri window rather than reasoning from the config
alone. What I could confirm:

- The window is 1280×900 with `minWidth: 960`, and the accessibility tree
  reports the webview at the full window size with the traffic lights at the top.
- `decorations` and `titleBarStyle` are both unset, so the default is a standard
  visible title bar and the webview does **not** extend under it. Structurally
  there is one system separator, then 12px of ground, then our 1px hairline —
  a gap between two lines, which is not a double border.
- At the default 1280 the frame is on; at the 960 minimum it collapses to
  edge-to-edge. Screenshots of both are below.

**What I could not test: the actual pixels.** Screen recording is denied in this
environment (`screencapture` fails with "could not create image from display"),
so I could not photograph the boundary between the native title bar and the
frame. The specific thing I am unsure about is stated in the rough edges below.

## Screenshots

In the scratchpad, not the repo:
`/private/tmp/claude-504/.../scratchpad/wt5-1178/shots/`

Taken on the merged head. `upstream/main` moved twice under this branch and both
moves touched the chrome I was designing against — **#1176** removed the
auto-collapse (so the sidebar now stays wherever the operator left it), **#1174**
put the host switcher in the sidebar header, and **#1172** took Pages out of the
nav. Every shot below is after all three.

**40 shots: 5 views × 2 themes × 2 widths × 2 sidebar states.**
`f-<view>-<theme>-<width>[-collapsed].png`

Views: `overview`, `chat`, `workflows`, `ledgers`, `settings`.
Widths: `1440` (frame on) and `820` (frame off).
Sidebar: expanded (no suffix) and `-collapsed` (the 3rem icon rail).

What each one is for:

- **Overview** — the knowledge graph, which is the only full-bleed view and the
  one that had `h-svh`. Its legend sits flush to the frame's bottom edge; if the
  clipping regression came back this is where you would see it.
- **Chat** — three columns inside one frame (shell nav, channel rail, timeline),
  composer flush to the bottom. This is where the step from the ground into the
  work reads most clearly: tinted ground → rung-2 sidebar → near-white rail →
  canvas.
- **Workflows** — a workflow open with the run-history rail showing. Rail and
  canvas both run the full height of the frame; the canvas keeps ~880px beside
  the rail at 1440, well over the 640px floor `workflow-run-history-rail.spec.ts`
  pins. At 820 the rail correctly falls back to a strip under the canvas.
- **Ledgers** — the board. Also `f-board-end-<theme>.png`, scrolled to its far
  right, where Done is a full column ending inside the frame's inner edge rather
  than against the window's.
- **Settings → General** — the densest column arithmetic in the console: shell
  nav, settings nav, and content, all inside the frame.

At **820** every one of these is edge to edge — no ground, no radius, no border.
The frame is gone, not thinned.

Detail and shell shots:

- `d-tl-light.png` / `d-tl-dark.png` — the top-left corner at 3×, where the
  radius, the hairline and the ground → sidebar → canvas step are all legible.
- `f-workspace-light.png` / `-dark.png` — the workspace tree.
- `f-tauri-default-1280.png`, `f-tauri-min-960.png` — the frame at Tauri's
  default window width and at its minimum, i.e. either side of the breakpoint.
- `ov-tour.png`, `ov-dialog.png` — the first-run tour and a dialog, both
  portalled, both still centred on the window and above the frame.

## What I am unsure about

1. **The macOS title-bar boundary, in light mode.** The light ground `#EBE7F7`
   is close in lightness to the macOS light title bar. The 12px band may read as
   part of the window chrome rather than as the app's ground, which would make
   the console look like it simply starts lower. I could not photograph this.
   If it lands badly, the fix is one value, not a redesign.
2. **12px may be too quiet.** At 1440 the ground is present but modest. I chose
   restraint over a band you cannot miss, on the grounds that this is chrome and
   not content — but this is exactly the sort of thing to have an opinion about
   from the screenshots.
3. **The active nav row is softer in light mode.** `--sidebar-accent` (`#ECE9FC`)
   against rung 2 rather than white is a smaller step in the fill. The brand-hued
   label (6.90:1) still carries it, but if the row reads weak, bumping
   `--sidebar-accent` toward `--brand-100` is the follow-up.
4. **The ReactFlow minimap** sits at the bottom-right of the workflows canvas and
   its lower edge is close to the frame's. It looks the same as it did before
   this change, so I have left it, but it is the one control that lives where the
   frame's flush bottom edge is.
5. **The collapsed sidebar is the weakest case in light mode, and #1176 made it
   a state people stay in.** The icon rail is 3rem wide and painted rung 2
   (`#F1F1F8`) against a ground of `#EBE7F7` — 0.024 apart in lightness, which
   reads fine across a 13.5rem column and has very little area to register in
   across a 48px strip. Compare `f-overview-light-1440.png` with
   `f-overview-light-1440-collapsed.png`: in the second the console arguably
   reads as starting at the canvas rather than at the rail.

   **Dark does not have this problem** — there the ground is *lighter* than the
   app, so the collapsed rail is the darker of the two and separates cleanly
   (`f-workflows-dark-1440-collapsed.png`). If light needs fixing, the honest
   fix is to open the gap between the ground and rung 2 rather than to
   special-case the rail, and that is a one-value change to
   `--surface-light-ground`.

Taken on the merged head, after `upstream/main` moved under me — #1176 landed
while I was working, so the sidebar no longer auto-collapses on Chat and these
shots show it expanded.

## Commands run

```
frontend: npm run typecheck · typecheck:unit · typecheck:e2e · npm test (130 files, 1268 tests)
          npm run build
          npx playwright test app-frame sidebar-host-switcher board-drag board-columns \
              workflow-list-columns workflow-run-history-rail tour-popover-in-viewport \
              sidebar-toggle-reachable
root:     scripts/ci/assert-design-tokens.sh · scripts/ci/assert-md-line-cap.sh
tauri:    cargo build (src-tauri) and launched the window
```

### Tests updated, and why

- **`sidebar-host-switcher.spec.ts`** — asserted `sidebar.x === 0`. That was the
  right assertion when the window edge and the app edge were the same line; now
  it would pin the sidebar *outside* the frame, which is the exact bug the spec
  exists to catch, one frame further out. Restated against the frame: the app is
  inset from the window, and the sidebar starts at the app's left edge inside its
  1px border. **Tighter, not looser** — it now checks two facts where it checked
  one, and it reads the frame's real box rather than a magic 13.
- **`app-frame.spec.ts`** — new, four cases. Pins the three insets and the flush
  bottom, the collapse at `lg` (1023 vs 1024), that the frame never grows the
  document, that no view overflows it, and — because #1176 landed mid-branch —
  that collapsing the sidebar to its icon rail does not move the frame or let
  the rail slip onto the ground. Geometry is the assertion because geometry is
  what a stylesheet edit can silently take away while the console still renders
  fine.

Everything else passed unchanged: `board-drag`, `board-columns`,
`workflow-list-columns`, `workflow-run-history-rail`,
`tour-popover-in-viewport`, `sidebar-toggle-reachable`.

### DIFF
--- a/docs/design-system/color.md
+++ b/docs/design-system/color.md
@@ -78,9 +78,32 @@ is the primitive and the theme picks a set.
 | `3` | `#E8E8F2` | `#131317` | Icon circles |
 | `4` | `#DADAE5` | `#161719` | Card stroke · dividers |
 | `active` | `#ECE9FC` | `#1E1E28` | The nav row you are standing on |
+| `ground` | `#EBE7F7` | `#15131B` | The surface the app frame sits on |
 
-Declared as `--surface-light-*` and `--surface-dark-*`. Every value is the
-brand guide's, unchanged.
+Declared as `--surface-light-*` and `--surface-dark-*`. Every value except
+`ground` is the brand guide's, unchanged.
+
+`ground` is the one rung that is not inside the app (issue #1178). The console
+is inset from the window, and this is what shows around it. Two things make it
+different from the six above:
+
+- **It is the most chromatic neutral in the system** (chroma 0.017–0.022
+  against 0.004–0.026 for the rest, at hue 296.5° rather than the ladder's
+  285–286°). It carries no text, no stroke and no control, so nothing on it can
+  be harmed by a tint — and a grey there reads as an empty margin rather than
+  as a ground the app is standing on. The ~10° warmer hue is what separates it
+  from "another panel". A genuinely warm ground would put a second hue family
+  into a palette that owns one, and would go muddy exactly where it meets the
+  cool greys — at the frame's edge, the one place this colour is ever seen
+  against another.
+- **The two themes take opposite sides of their canvas.** Light sits a step
+  *below* it (L 0.936 against 0.978), so the app stays the brightest thing on
+  screen. Dark sits a step *above* it (L 0.193 against 0.140), because there is
+  no rung left below near-black — darker is `#000`, and at that lightness the
+  chroma is invisible. It is also what makes the frame's shadow work: a shadow
+  needs something to fall on. The rule that surfaces climb toward the viewer
+  governs panels stacked *inside* the app; the window ground is not in that
+  stack.
 
 ### Ink levels
 
@@ -130,15 +153,40 @@ Layer 2. These are what components use.
 | `--secondary` | rung `3` | rung `active` | Secondary button fill |
 | `--surface-icon` | rung `3` | rung `3` | Ground behind an icon circle |
 | `--accent` | rung `active` | rung `active` | Hover/rest tint under rows |
-| `--sidebar` | rung `1` | rung `1` | Nav column |
+| `--sidebar` | rung `2` | rung `1` | Nav column |
 | `--sidebar-accent` | rung `active` | rung `active` | Active nav row background |
 | `--border` | rung `4` | rung `4` | The hairline |
 | `--input` | rung `4` | rung `active` | Field borders, stronger rules |
 | `--ring` | `brand-500` | `brand-400` | Focus |
+| `--app-ground` | rung `ground` | rung `ground` | Outside the app frame |
+| `--app-frame-edge` | rung `4` | rung `bg` | The frame's own hairline |
 
 Light surfaces climb *toward* the viewer with lightness (canvas `#F7F7FC` →
 card white), and dark does the same (`#08090B` → `#0C0D0F` → `#161719`). A card
 lifts off the page before any shadow is applied.
+
+**The ladder runs from the window inward, and it has to be monotonic.** With a
+ground outside the frame there are four surfaces between the window edge and a
+card, and if they do not step in one direction the frame reads as three
+unrelated greys rather than as an app sitting on a desk:
+
+```
+light   ground #EBE7F7 → sidebar #F1F1F8 → canvas #F7F7FC → card #FFFFFF
+dark    ground #15131B → sidebar #0C0D0F → canvas #08090B → card #0C0D0F
+```
+
+Dark already ran in that order. Light did not: `--sidebar` was rung 1, pure
+white, one step *above* the canvas it sits beside — while the stylesheet's own
+comment on the line claimed it was "slightly recessed from the canvas so the
+working area is the brightest thing on screen". Issue #1178 moved it to rung 2,
+which is what that sentence meant.
+
+`--app-frame-edge` is `--border` in light and the *canvas* value in dark, which
+looks inconsistent and is not. In dark the ground is lighter than the app, so
+rung 4 (`#161719`) sits between the two and reads as a third surface wedged into
+the seam. The canvas value disappears against the content it borders and stays a
+dark line against the ground and the sidebar — a raised edge on a dark screen is
+an absence of light, not a stroke.
 
 Dark borders were translucent white, so one value could read against the
 canvas, the card and the popover at once. The guide names the stroke outright —
--- a/docs/design-system/components.md
+++ b/docs/design-system/components.md
@@ -209,6 +209,11 @@ zero below `lg` (1024px):
 | `--app-frame-radius` | 14px | 14px | 0 |
 | `--app-frame-border` | 1px | 1px | 0 |
 
+The breakpoint is `lg` and not `md`: `md` is where the sidebar becomes an
+overlay sheet, so a frame surviving down to it would be a border around a single
+column. `lg` is where the inset starts to cost — 24px out of a 779px working
+column at 1024, 4.6% of a 523px one at 768.
+
 **Inset on three sides, flush to the bottom.** Vertical space is the scarcest
 thing in this app — chat's composer, the ledgers board and the workflows canvas
 all want every pixel of height — so only one inset comes out of `100svh`, and
--- a/docs/design-system/components.md
+++ b/docs/design-system/components.md
@@ -189,6 +189,48 @@ Never nest a modal inside a modal.
 
 ---
 
+## App frame
+
+Not a primitive but a shape the whole console takes, so it is stated once here
+rather than rediscovered per view (issue #1178). `SidebarProvider` renders two
+boxes:
+
+| Box | `data-slot` | What it is |
+| --- | --- | --- |
+| Ground | `sidebar-ground` | `100svh`, painted `bg-app-ground`. Holds nothing. |
+| Frame | `sidebar-wrapper` | The console: one edge, one radius, one shadow. |
+
+Three geometry tokens drive it, all declared at `:root` in `index.css` and all
+zero below `lg` (1024px):
+
+| Token | ≥ `lg` | ≥ `2xl` | Below `lg` |
+| --- | --- | --- | --- |
+| `--app-frame-inset` | 12px | 18px | 0 |
+| `--app-frame-radius` | 14px | 14px | 0 |
+| `--app-frame-border` | 1px | 1px | 0 |
+
+**Inset on three sides, flush to the bottom.** Vertical space is the scarcest
+thing in this app — chat's composer, the ledgers board and the workflows canvas
+all want every pixel of height — so only one inset comes out of `100svh`, and
+only the two top corners are rounded.
+
+**All three collapse together.** A radius kept at zero inset rounds the corners
+against the window edge and lets the ground show through as two stray tinted
+notches; a border kept there draws a hairline down the outside of the screen.
+Below `lg` there is no frame, not a flattened one.
+
+**Nothing inside may use a viewport unit for height.** The frame is one inset
+shorter than the window, and the shell clips rather than scrolls, so an `h-svh`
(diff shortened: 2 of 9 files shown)


graycyrus:
Closing — the result doesn't work. Not pursuing this direction further.

PR #1195, opened 2026-08-19
feat(chat): a "Just chatting" intent that withholds the card (#1152)

### LINKED ISSUE #1152: chat: an operator has no way to say a message is NOT work, so the classifier is the only thing that can withhold a card
## What is needed

An operator can make the company open a card it was not going to open. They cannot stop it opening one it was.

Split out of #1035, whose fix threads the composer's choice into the runtime and closes the defect that signal's absence caused — a workflow request the triage did not recognise as work became **two** cards, because the runtime could not tell the handler had already carded it. That is landed. This issue is the half that was named in #1035 and deliberately not attempted, because it cannot be built without a product decision first.

## Why it is not a code change yet

The override is asymmetric by construction, not by oversight:

- `TaskDeliverable` has exactly two variants, `Once` and `Workflow` (`src/ports/tasks.rs`). **Both mean "do work."**
- The console sends the field *only* when the choice is `workflow` — `if (deliverable === "workflow") body.deliverable = deliverable;` (`frontend/src/api/client.ts`). Picking "Do it once" puts nothing on the wire.

So there is no value an operator can send that means *"this is not work — do not card it."* Threading the existing signal harder cannot produce one. The classifier remains the only thing that can withhold a card, and #984's own reasoning says the person who typed the message should be able to outrank it.

## The decision to make first

Someone has to choose what the third intent **is**, before any of it is buildable:

1. **A third composer option** — "just talking" / "no card" beside "Do it once" and "Build me the workflow". Explicit, discoverable, and costs a third control in a two-control affordance.
2. **A different mechanism** — a modifier on send, a per-thread default, or making an unmarked message mean "let the classifier decide" while an explicit *Once* means "card it" and some new gesture means "don't".

Option 1 is the obvious shape and probably right, but it is a UX call about the most-used control in the product, and it should not be smuggled into a bug fix. That is why #1035 shipped without it.

- A value on the wire that means "not work", carried end to end: composer → `OperatorMessage` → `MessageContext` → both card paths.
- The runtime path already exists after #1035 — `MessageContext` reaches `open_work_card` and `open_direct_work_card`, and `carded_by_handler` is now derived correctly. The seam is open; only the meaning is missing.
- The REST handler must honour it too, or the two surfaces disagree about the same message — the exact failure #1035 fixed, pointed the other way.

## Related

- #1035 — the threading + double-card fix (landed)
- #984 — the classification half: a message the model reads as conversation no longer opens a card
- #463 — one message, one card

oxoxDev:
## Summary

An operator could make the company open a card the classifier had declined, but never stop it opening one. `TaskDeliverable` has two values and both mean "do work", so there was no value on the wire meaning "this is not work". The classifier was the only thing that could withhold a card.

This adds a third composer position, **"Just chatting"**, and honours it in the chat route and the runtime.

The runtime seam already existed. `MessageContext.chatter` stands both deterministic card paths down — but it is the *model's* verdict, set only where the lexical layer abstained. What was missing was a way for the person who typed the message to assert the same thing.

## Problem

Concretely, before this change:

- `TaskDeliverable` is `Once | Workflow`. Both open a card.
- The console sent the field only for `workflow`, so choosing "Do it once" put nothing on the wire.
- So the operator's only lever was additive. `#1035` fixed the case where that lever fired twice; this is the half `#1035` named and deliberately left, because it needed a product decision first.

## Solution

**A separate `MessageIntent { Chat, Once, Workflow }`, not a third `TaskDeliverable` variant.** `TaskDeliverable` is a stored card field, editable later in the task dialog. A card can never *be* "not work", so a third variant would be representable-but-invalid in every persisted position. `MessageIntent::deliverable()` returns `Option<TaskDeliverable>`, and `Chat => None` is the type-level statement "no card, therefore no deliverable".

**It rides the existing `deliverable` key** — one operator choice, one field. A second field would make `{deliverable: "workflow", intent: "chat"}` representable and force identical reconciliation in two surfaces, which is precisely the split-brain `#1035` fixed.

**A peer field in the runtime, not a reuse of `chatter`.** `not_work` is a person's statement, settled before any model runs; `chatter` is the model's verdict. Folding them together would falsify `chatter`'s documented meaning and attribute an operator's choice to the model in the debug log. The guard sits above the `chatter` check in `open_work_card`, where all three card paths funnel.

**Wire compatibility:** `"chat"` is sent explicitly; `"once"` and no-choice send nothing, so an unmarked message is byte-identical to today's. "Do it once" stays pre-pressed, so nothing changes for anyone who does not touch the new control.

### Known limit — please read before approving

Only the **deterministic** card paths stand down. The orchestrator's own `spawn_task` tool is untouched, deliberately: narrowing board tools would change which delegation-queue claim the turn runs under, which this issue does not justify.

So "Just chatting" means the company will not *automatically* card the message — not that a card can never appear. If that reads as too weak a promise for the label, say so and the label is the cheap thing to change.

## Submission Checklist

- [x] `cargo check --features openhuman,mcp,telegram` and `cargo check --all-features`
- [x] `cargo fmt --check`
- [x] Touched-scope tests, re-run after rebasing onto current main: `ports::types` 97, `runtime::delegation` 80, `runtime::cycle` 76, `server::operator` 105, `harness::brain` 97 — all passing
- [x] Console: `npx vitest run` 131 files / 1283 tests, `typecheck` + `typecheck:unit` + `typecheck:e2e`, `npm run build`
- [ ] `cargo clippy --all-targets --no-deps -- -D warnings` — passes at the CI lane's feature set. With `mcp` added it fails on `src/app/types.rs` (`empty_line_after_doc_comments`), a file this branch does not touch; the site is `#[cfg(feature = "mcp")]` so the CI lane never compiles it. Pre-existing on main, reported separately rather than swept in here.
- [x] Manual UI verification — done, evidence in the thread below

## Impact

Behaviour is unchanged for every message that does not use the new control. An unmarked message is byte-identical on the wire, and no journaled record migrates: `OperatorMessage` blobs with no `deliverable` and with `"workflow"` both still parse and re-serialize identically, which is asserted by test rather than by comment.

## Tests

The behaviour tests were checked by mutation, not by reasoning. Disabling the two guards failed all four new behaviour tests; restoring them turned them green. Widening the builder-injection pattern turned the no-briefing assertion red. Reverting the console's wire rule failed exactly the one case that asserts `"chat"` reaches the body.

One assertion from the plan was **moved because it was vacuous where it was specified**: "outranks a `Track` triage" is already true in the runtime before this change, since `carded_by_handler` covers it — the test would have passed for the wrong reason. It now lives in the REST handler test, on a fixture whose triage verdict is asserted `Track`, where it is real. The runtime tests instead assert the two things that *are* real there: `not_work` beats `is_trackable_work`'s "everything is work" default, and it beats a model `Work` verdict from the triage escalation.

## Related

Closes #1152. Follows #1035, which threaded the signal and fixed the double-card. Part of #984 — this takes two bullets from its Fix list (the operator-side override), and deliberately leaves `chatter` and the "nothing pre-pressed" default alone, since both belong to that issue.

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -7,7 +7,7 @@
 //     and calls use `/api/v1/companies/{id}/*`.
 
 import type { ConsoleConfig } from "../config";
-import type { TaskDeliverable } from "./tasks";
+import type { MessageIntent } from "./tasks";
 import { defaultTransport, needsCarriedSession } from "./transport";
 import type { StreamHandlers, Transport, TransportResponse } from "./transport";
 import {
@@ -423,20 +423,29 @@ export class OpenCompanyClient {
      */
     parent?: string | null,
     /**
-     * The once-vs-workflow choice for the card this line opens (issue #580).
-     * Only `"workflow"` reaches the wire: `"once"` (and the default) is sent as
-     * *nothing at all*, so an ordinary message keeps the exact body shape it had
-     * before #580 — the same omitted-field compatibility rule the deliverable
-     * field follows everywhere (see `CreateTask.deliverable`).
+     * What this message is for (issues #580, #845, #1152): `"once"` and
+     * `"workflow"` say what the card it opens produces, `"chat"` says it is not
+     * a request for work and no card should be opened for it.
+     *
+     * Everything except `"once"` reaches the wire; `"once"` and the default are
+     * sent as *nothing at all*. That is what keeps "Do it once" the default: an
+     * unmarked message posts the exact body shape it posted before any of these
+     * controls existed, so the host cannot tell one from a pre-#580 client —
+     * the same omitted-field compatibility rule the deliverable field follows
+     * everywhere (see `CreateTask.deliverable`).
+     *
+     * One key, not two. `"chat"` rides `deliverable` rather than arriving as a
+     * second `intent` field, so a body cannot claim "build me the workflow" and
+     * "just chatting" about the same message.
      */
-    deliverable?: TaskDeliverable,
+    intent?: MessageIntent,
   ): Promise<ChatResponse> {
-    const body: { text: string; chat?: string; parent?: string; deliverable?: TaskDeliverable } = {
+    const body: { text: string; chat?: string; parent?: string; deliverable?: MessageIntent } = {
       text,
     };
     if (chat) body.chat = chat;
     if (parent) body.parent = parent;
-    if (deliverable === "workflow") body.deliverable = deliverable;
+    if (intent && intent !== "once") body.deliverable = intent;
     return this.request<ChatResponse>("POST", `${this.scope(company)}/chat`, body);
   }
 
--- a/frontend/src/api/tasks.ts
+++ b/frontend/src/api/tasks.ts
@@ -203,6 +203,24 @@ export interface AssigneeCandidate {
  */
 export type TaskDeliverable = "once" | "workflow";
 
+/**
+ * What the operator says one chat **message** is for (issue #1152).
+ *
+ * The two work words mean exactly what they mean on a card. `"chat"` is the
+ * operator saying this message is not a request for work at all — the composer's
+ * "Just chatting" — so no deterministic path opens a card for it.
+ *
+ * Deliberately **not** a widening of `TaskDeliverable`. That type is the card
+ * field, shared by `CreateTask`, `TaskPatch` and `TaskDto`, and a card can never
+ * *be* "not work": widening it would make `"chat"` assignable in every one of
+ * those positions and put the compiler on the wrong side of the invariant. This
+ * is message-scoped, and the host draws the same line (`MessageIntent` beside
+ * `TaskDeliverable` in `src/ports/types.rs`).
+ *
+ * Sent only when it is not the default — see `OpenCompanyClient.chat`.
+ */
+export type MessageIntent = "chat" | TaskDeliverable;
+
 /** A positive source-currency USD amount or an explicit role-redacted state. */
 export interface TaskCost {
   amountUsd?: number;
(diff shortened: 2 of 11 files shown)


oxoxDev:
Note for #984 (@M3gA-Mind): this takes only the operator-side override — a third composer position that withholds the card. It deliberately does **not** touch `MessageContext.chatter` (the model-verdict half) and does **not** adopt #984's "nothing pre-pressed" default, since changing what an *unmarked* message means affects every message and belongs with that issue rather than inside this one.

oxoxDev:
Manual verification done end to end — real host, built console, staging inference, `companies/e2e_harness`. The same sentence sent twice into `#engineering-desk`, changing only the composer position:

| Send | On the wire | Turn | Card |
|---|---|---|---|
| **Just chatting** | `deliverable: "chat"` | `pending → running → succeeded`, agent replied | **none** |
| unmarked | key absent entirely | `pending → running → succeeded`, agent replied | opened in `planning`, moved to `todo` |

Both turns reached `succeeded` with an `AgentReply`, which is the part that matters: the withheld case is a turn that ran and chose not to card, not a turn that failed. A failed turn would produce the same empty board and prove nothing.

Also confirmed in the running console:

- Three positions render in a channel composer, `aria-pressed` = `false / true / false` — "Do it once" is pre-pressed, so nothing changes for anyone who ignores the new control.
- The group label reads "What this message is for" rather than "What this should produce" — "Just chatting" produces nothing, so the old wording no longer described the control.
- The intent resets to "Do it once" after each send.
- The unmarked send confirms the compatibility rule: choosing "Do it once" puts **no** `deliverable` key on the wire, so an unmarked message is byte-identical to a pre-#1152 one.

One thing reviewers should see rather than take on trust: in the withheld turn the agent's own reply proposed "Open a task card for the build with acceptance criteria" as a next step. It said that and no card appeared — which is exactly the documented limit in the PR body. The deterministic paths stand down; the orchestrator's own `spawn_task` tool is untouched. "Just chatting" means the company will not automatically card the message, not that a card can never appear.

Issue #1197, opened 2026-08-19
main does not compile its gated test targets: #1134 and #1182 collided semantically


M3gA-Mind:
## Summary

`upstream/main` does not compile its **gated test targets**. Two PRs that were each green on their own lane merged 1h20m apart and left main red — a semantic merge conflict, invisible to both.

Reproduced on a clean checkout of `main` (`8d58d442`), with no local changes:

```
$ cargo check --locked --features openhuman,tinycortex --tests
error[E0560]: struct `ports::types::CycleRequest` has no field named `compressed_history`
   --> src/harness/spend_halt_turn_test.rs:379:9
error[E0560]: struct `ports::types::CycleRequest` has no field named `roster`
   --> src/harness/spend_halt_turn_test.rs:380:9
error[E0560]: struct `ports::types::CycleRequest` has no field named `context_index`
   --> src/harness/spend_halt_turn_test.rs:381:9
error: could not compile `opencompany` (lib test) due to 3 previous errors
```

## The two halves

| | |
| --- | --- |
| **#1134** (`harness: say when a turn was stopped for spend`) | merged 19:32:37Z. Added `src/harness/spend_halt_turn_test.rs`, whose `CycleRequest` literal sets the three fields. |
| **#1182** (`chore/dead-memory-path`, commit `c778c33e` "drop the CycleRequest fields no brain reads") | merged after. Removed those three fields. |

Neither is wrong. #1182's own lane was green against a main that did not yet contain the new test file, and #1134's was green against a `CycleRequest` that still had the fields. The merge is textually clean and semantically broken — nothing in either PR's lane could have caught it.

## Why it needs fixing now rather than at leisure

`cargo check --tests` is what the gated `Rust (openhuman, tinycortex)` lane compiles, so **every open PR's gated lane now fails for a reason that has nothing to do with its diff.** With several PRs in flight that is a lot of wasted investigation, and the real risk is the one this shape always carries: a genuine breakage arriving while the lane is already red gets read as "the known main breakage" and merged past.

`--lib` alone stays green, which is exactly why this survived: a lib-only check is blind to a struct-field change that only test fixtures depend on.

## The fix

Three lines. `src/harness/spend_halt_turn_test.rs:379-381` is the **only** remaining consumer in the tree —

```
$ grep -rn "compressed_history" src/
src/harness/spend_halt_turn_test.rs:379:        compressed_history: Vec::new(),
[...]

Issue #1059, opened 2026-08-18
runtime: a runtime built without with_harness serves a board that silently cannot dispatch


M3gA-Mind:
## Summary

A `CompanyRuntime` built without `RuntimeBuilder::with_harness(...)` hands back a working board that **cannot dispatch anything**, and says nothing about it. A card dragged into In Progress simply sits there: no dispatch, no run row, no error, no log line.

## Where it comes from

`RuntimeBuilder::harness` is `Option<Arc<HarnessPool>>` and defaults to `None` (`src/runtime/builder.rs:448`). `CompanyRuntime::dispatch_task` is explicit that this is intentional:

> Fires the detached `TaskDispatched` cycle for a task when a harness is attached. […] **In the default build (no harness) this is a no-op, keeping the board inert.**

Only `src/bin/opencompany.rs:589` calls `with_harness`. Anything else that builds a runtime — a test, a tool, an embedding — gets the inert board unless its author knows to add that one line.

Note this is **not** the `openhuman` feature being off. With `--features openhuman` and no `with_harness`, the harness brain is still never constructed, because the branch at `src/runtime/builder.rs:2106` matches on `self.harness` before it looks at inference at all.

## Why it is worth fixing rather than knowing

The failure has no signal. Everything downstream looks healthy:

- `POST /tasks` → 200, card created
- `PATCH /tasks/{id} {"column":"in_progress"}` → 200, card moves
- `GET /tasks/{id}` → the card sits in `in_progress` with an **empty timeline** for as long as you care to poll

There is no warning to grep for, and the nearby diagnostic is misleading by omission: when inference fails to resolve the builder logs a clear `no inference source resolved … the openhuman harness is disabled` (`builder.rs:2625`). When the *pool* is missing it logs nothing, so the one case that does warn is the case that is easier to spot.

Found while writing the offline end-to-end lane (#579): the first version of that test had a card inert in In Progress with an empty timeline and zero inference calls, and the cause was one missing builder line rather than anything about being offline.

## Options

1. **Warn on build** when a runtime is constructed with no harness pool — the same shape as the existing inference warning, so the two cases are diagnosable the same way. Cheapest, and would have turned the above into a one-line answer.
2. **Make the no-op loud at the edge**: `dispatch_task` logs (once per company, not per card) that it dropped a dispatch because no harness is attached.
[...]

PR #1199, opened 2026-08-19
board: name the remedy the operator's own build actually has

### LINKED ISSUE #1198: inert-board warning names `with_harness` in builds where it isn't compiled
## Summary

The inert-board warning added by #1059 / PR #1084 names a remedy that does not exist in the build most likely to see it. Raised by @oxoxDev as a `CHANGES_REQUESTED` finding on #1084 (2026-08-19 16:38) and **merged without being addressed** (`c83de73c`), so it is now on `main`.

His finding, quoted:

> **Major — in a default-feature build the warning names a remedy that doesn't exist.**
> The warn block (`company/runtime.rs:906-921`) isn't cfg-gated, but `RuntimeBuilder::with_harness` is — `#[cfg(feature = "openhuman")]` at `runtime/builder.rs:941`. So a host built without `openhuman` tells the operator *"Wire one with `RuntimeBuilder::with_harness(...)`"* about a method that isn't compiled into their binary. That build isn't hypothetical: `Dockerfile:5,44-46` ships `FEATURES=""` as a first-class configuration and Cargo.toml describes the default build as "offline and echo-brained". The right remedy there is *rebuild with `--features openhuman`*.

## Verified still present on `main` (`8d58d442`)

All three legs, read off the current tree rather than inferred from the commit list:

* `src/runtime/builder.rs:941-942` — `#[cfg(feature = "openhuman")] pub fn with_harness(...)`
* `src/company/runtime.rs:~912-923` — the `tracing::warn!` is **not** cfg-gated and interpolates ``Wire one with `RuntimeBuilder::with_harness(...)` ``
* `Dockerfile:7` — `ARG FEATURES=""`, defaulting the feature set to empty

`dispatch_task` itself is not gated either (only the early-return arms inside it are), so the warning genuinely executes in a default build.

## Why it is not the same cause as "nobody wired a pool"

There are two causes with two different fixes, and the message currently assumes the first:

| Cause | What helps |
| --- | --- |
| Feature is on, nobody called `with_harness` | wire one — the current text |
| Binary built without `openhuman` | rebuild with `--features openhuman` |

@oxoxDev already answered the rebuttal on the thread — *"showing that a wired harness can still be absent proves the feature isn't sufficient, but the feature being off is a separate cause with a separate fix, and that's the one the message gets wrong"* — so this is not re-litigating the point, only carrying it out.

## Suggested shape

Cfg-split **only the remedy sentence**, keeping the symptom half in one literal shared by both builds so the description of what happened cannot drift while the two remedies diverge.

The existing test `an_inert_board_says_it_cannot_dispatch_once` (`runtime/builder.rs`) asserts `text.contains("with_harness")`, and its module is `#[cfg(test)]` with no feature gate — so it runs in the default build and will need splitting with the message. Worth having each arm assert the *other* remedy is **absent** as well as its own present: without that, a message hedging by carrying both strings would satisfy every `contains` and still hand a default-build operator a dead end.

## Note

I have this implemented and verified locally (three lines plus the test split; `fmt --check`, gated `--lib`, and default-features `--tests` all clean) — it was built on a fresh branch off `main`, never on #1084's branch, which was already merged. Ready to raise on request.

⚠️ Its gated test-compile cannot be verified until #1197 lands: `main` currently fails `cargo check --features openhuman,tinycortex --tests` for an unrelated reason.

M3gA-Mind:
## Summary

The inert-board warning added by #1059 names a remedy the operator's build may not have. Raised by @oxoxDev in review on #1084, and merged before it was addressed, so it has been live on `main` since `c83de73c`:

> The warn block isn't cfg-gated, but `RuntimeBuilder::with_harness` is — `#[cfg(feature = "openhuman")]` at `runtime/builder.rs:941`. So a host built without `openhuman` tells the operator *"Wire one with `RuntimeBuilder::with_harness(...)`"* about a method that isn't compiled into their binary.

Re-verified against `upstream/main` before writing anything, rather than trusting the issue:

* `src/runtime/builder.rs:941-942` — `#[cfg(feature = "openhuman")] pub fn with_harness(...)`
* `src/company/runtime.rs:~950` — the `tracing::warn!` is **not** gated and names it
* `Dockerfile:7` — `ARG FEATURES=""`

`dispatch_task` itself is not gated either (only the early-return arms inside it are), so the warning genuinely executes in a default build.

## Two causes, two remedies

| Cause | What actually helps |
| --- | --- |
| Feature on, nobody called `with_harness` | wire one — the existing text |
| Binary built without `openhuman` | rebuild with `--features openhuman` |

Only the first was ever printed, so the build most likely to meet the warning was the one it could not help. Not re-argued here: the rebuttal was made on #1084 and @oxoxDev answered it — *"the feature being off is a separate cause with a separate fix, and that's the one the message gets wrong."*

## What changed

**Only the remedy sentence is split.** The symptom stays one literal shared by both builds, so the half that describes what happened cannot drift between them while the half that says what to do about it is the only thing the `cfg` decides.

```rust
#[cfg(feature = "openhuman")]
const REMEDY: &str = "Wire one with `RuntimeBuilder::with_harness(...)` …";
#[cfg(not(feature = "openhuman"))]
const REMEDY: &str = "This binary was built without the `openhuman` feature, so it has no \
     harness to wire — rebuild with `--features openhuman` (the `FEATURES` build arg in \
     `Dockerfile`), or move the card by hand.";
```

The latch, the level, the fields and the symptom text are untouched.

## The test had to move with it, and that is not incidental

`an_inert_board_says_it_cannot_dispatch_once` asserted `text.contains("with_harness")`, and its module is `#[cfg(test)]` with **no feature gate** — so it runs in the default build and would have failed on the corrected message. Split with it, and each arm now pins the **other** remedy *absent* as well as its own present.

That second half is the load-bearing one: without it, a message that hedged by carrying both strings would satisfy every `contains()` and still hand a default-build operator a dead end.

## API Or Behavior Changes

None. One log line's wording, in one build configuration. No API, no signature, no wire change.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — N/A as a separate run: the change is a `const` and a `format!` capture inside an existing block, and both builds were compiled through `cargo check --tests` below. Flagging honestly rather than ticking a command I did not run.
- [x] `cargo build --all-targets` — N/A as a separate run; covered by the `check --tests` compiles in both feature sets, which build every test target.
- [x] `cargo test` — the affected test, run in **both** builds, because the whole point is that the message differs by build:
  - default features → `an_inert_board_says_it_cannot_dispatch_once ... ok`
  - `--features openhuman,tinycortex` → `an_inert_board_says_it_cannot_dispatch_once ... ok`

Compiles, both feature sets, `--lib` **and** `--tests`:

| Command | Result |
| --- | --- |
| `cargo check --locked --tests` (default) | exit 0 |
| `cargo check --locked --features openhuman,tinycortex --lib` | exit 0 |
| `cargo check --locked --features openhuman,tinycortex --tests` | exit 0 † |

† **`main` currently fails that command for an unrelated reason** — #1197, `CycleRequest has no field named compressed_history` in `src/harness/spend_halt_turn_test.rs`, from #1134 and #1182 colliding semantically. #1197 had not landed when this was verified, so I applied its three-line deletion **locally, for verification only, and reverted it before committing**. This branch does **not** carry that fix; its diff is two files. Once #1197 merges, the gated lane here should be green on its own.

### Prove-red

Reverting `runtime.rs` to main's unconditional message while keeping the new test fails it in the default build (exit 101), and the panic prints the defect verbatim:

```
a default-feature build has no harness to wire, so rebuilding with the feature is the
only thing that helps: "… WARN … [board] a card was dispatched but this runtime has no
agent pool … Wire one with `RuntimeBuilder::with_harness(...)` …"
```

Stated plainly: the **gated** arm is a regression guard, not a mutation proof of this bug. Main's message already satisfies it, which is correct — the defect only ever existed in the default build. It exists so a future "fix" that swaps the message wholesale cannot silently take the right remedy away from the build that does have `with_harness`.

## Documentation

No spec doc changes. The reasoning — two causes, why only the remedy is split, and why each test arm pins the other's absence — is recorded at both sites.

Closes #1198

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated no-harness dispatch warnings to provide build-specific remediation guidance.
  * OpenHuman builds now recommend connecting a harness.
  * Default builds now explain that the feature is unavailable and recommend rebuilding with the `openhuman` option.
  * Preserved the existing warning behavior and one-time reporting.

### DIFF
--- a/src/company/runtime.rs
+++ b/src/company/runtime.rs
@@ -944,13 +944,34 @@ impl CompanyRuntime {
             .inert_board_reported
             .swap(true, std::sync::atomic::Ordering::Relaxed)
         {
+            // The remedy is per build, because there are two different causes
+            // and only one of them is "nobody called `with_harness`" (issue
+            // #1059 review). `RuntimeBuilder::with_harness` is itself
+            // `#[cfg(feature = "openhuman")]`, so in a default-feature build
+            // naming it sends the operator looking for a method that is not
+            // compiled into their binary — and that build is not hypothetical:
+            // `Dockerfile`'s `ARG FEATURES=""` ships it as a first-class
+            // configuration, and Cargo.toml describes the default as offline
+            // and echo-brained. There the thing that would actually help is to
+            // rebuild with the feature.
+            //
+            // Only the remedy is split. The symptom stays one literal shared by
+            // both builds, so the half that describes what happened cannot
+            // drift between them while the half that says what to do about it
+            // is the only thing the cfg decides.
+            #[cfg(feature = "openhuman")]
+            const REMEDY: &str = "Wire one with `RuntimeBuilder::with_harness(...)` (see \
+                 `src/bin/opencompany.rs`), or move the card by hand.";
+            #[cfg(not(feature = "openhuman"))]
+            const REMEDY: &str = "This binary was built without the `openhuman` feature, so it \
+                 has no harness to wire — rebuild with `--features openhuman` (the `FEATURES` \
+                 build arg in `Dockerfile`), or move the card by hand.";
             tracing::warn!(
                 company = %self.id,
                 task = %task.id,
                 "[board] a card was dispatched but this runtime has no agent pool, so nothing \
-                 will work it: the card stays in `in_progress` with no attempt row. Wire one with \
-                 `RuntimeBuilder::with_harness(...)` (see `src/bin/opencompany.rs`), or move the \
-                 card by hand. Reported once per runtime."
+                 will work it: the card stays in `in_progress` with no attempt row. {REMEDY} \
+                 Reported once per runtime."
             );
         }
         let _ = task;
--- a/src/runtime/builder.rs
+++ b/src/runtime/builder.rs
@@ -4060,6 +4060,13 @@ mod test {
     /// The second dispatch pins the latch. An inert board with fifty cards has
     /// one problem, not fifty, and a per-card warning is the kind of noise that
     /// gets a useful line filtered out.
+    ///
+    /// The remedy is asserted per build (issue #1059 review). "No agent pool"
+    /// has two causes with two different fixes — nobody called `with_harness`,
+    /// or the binary was built without the feature that compiles it — and a
+    /// message naming the wrong one is a dead end dressed as help. Each arm
+    /// pins the other's remedy *absent* as well as its own present, so a
+    /// message that hedged by carrying both would fail here.
     #[tokio::test]
     async fn an_inert_board_says_it_cannot_dispatch_once() {
         use std::sync::{Arc as StdArc, Mutex as StdMutex};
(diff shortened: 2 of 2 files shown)


M3gA-Mind:
@coderabbitai review

PR #1200, opened 2026-08-19
fix(harness): drop the CycleRequest fields #1182 removed from the spend-halt fixture

### LINKED ISSUE #1197: main does not compile its gated test targets: #1134 and #1182 collided semantically

M3gA-Mind:
Closes #1197

## Summary

`main` did not compile its **gated test targets**, so the `Rust (openhuman, tinycortex)` lane was failing on every open PR for a reason unrelated to its diff.

Three lines deleted from one test fixture.

## Problem

```
$ cargo check --locked --features openhuman,tinycortex --tests
error[E0560]: struct `ports::types::CycleRequest` has no field named `compressed_history`
   --> src/harness/spend_halt_turn_test.rs:379:9
error[E0560]: struct `ports::types::CycleRequest` has no field named `roster`
   --> src/harness/spend_halt_turn_test.rs:380:9
error[E0560]: struct `ports::types::CycleRequest` has no field named `context_index`
   --> src/harness/spend_halt_turn_test.rs:381:9
error: could not compile `opencompany` (lib test) due to 3 previous errors
```

A semantic merge conflict between two PRs that were each green:

| | |
|---|---|
| **#1134** | added `src/harness/spend_halt_turn_test.rs`, whose `CycleRequest` literal sets the three fields |
| **#1182** (`c778c33e`) | removed those three fields, merged 1h20m later |

Neither lane could have caught it: #1182's ran against a main without this file, #1134's against a struct that still had the fields.

## Solution

**I checked the direction before assuming the fixture was the wrong half.** `c778c33e` is explicit, and it is a fix rather than an oversight:

> `compressed_history`, `context_index` and `roster` were populated on every cycle and read by no Brain implementation. Populating the first two cost a recent_traces read plus an unbounded `ContextStore::list` scan per cycle for a Vec that was dropped; the index grows with every turn a company has ever run.

So the removal stays and the fixture moves. Three supporting checks:

1. **No other consumer.** Across `src/`, the only `CycleRequest`-field uses of these names are the three lines here. The remaining `compressed_history` hit is #1182's own doc note at `ports/types.rs:2001` recording the removal — correct, and untouched. Other `context_index` / `roster` hits are unrelated (`context_index_jsonl` store paths, manifest rosters).
2. **Nothing asserts on them.** All three were `Vec::new()`; they were required fields, not fixture inputs. Deleting them changes nothing the test pins.
3. **It matches its sibling.** `cap_turn_test.rs` — the test this one was modelled on — already carries exactly this 4-field form, since #1182 updated it. The two literals are now the same shape again.

## Verification

Run on this branch, with the submodules initialised:

| command | before | after |
|---|---|---|
| `cargo check --locked --features openhuman,tinycortex --tests` | **exit 101** — the three `E0560`s above | **exit 0** |
| `cargo check --locked --features openhuman,tinycortex --lib` | exit 0 | exit 0 |
| `cargo fmt --all -- --check` | — | exit 0 |
| `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings` | — | exit 0 |
| `cargo test --locked --features openhuman,tinycortex --lib spend_halt` | — | **10 passed, 0 failed** |

I reproduced the red on a pristine checkout of `upstream/main` before touching anything, so the before column is measured rather than quoted from the issue.

**Note the `--lib` row.** It is green in both columns — a struct-field change breaks struct *literals*, which live in test fixtures, so a lib-only check is blind to this entire class. That is why this reached main, and it is why the `--tests` row is the one that matters.

The 10 spend-halt tests still pass, so this restores compilation without weakening what #1134 was written to assert.

## API Or Behavior Changes

**None.** Test-fixture only; no production code, no public type, no lockfile change.

## Tests

- [x] `cargo clippy --all-targets -- -D warnings` — ran the gated form CI uses, `cargo clippy --locked --no-deps --features openhuman,tinycortex --all-targets -- -D warnings`: clean.
- [x] `cargo build --all-targets` — ran `cargo check --locked --features openhuman,tinycortex --tests` instead, which is both the failing command and the one the gated lane compiles: exit 0. `Cargo.lock` is unchanged.
- [x] `cargo test` — ran the affected target, `cargo test --locked --features openhuman,tinycortex --lib spend_halt`: **10 passed, 0 failed**.

### No new test, deliberately

The assertion here is the compiler, and the gated lane already runs it — a test asserting "this struct literal type-checks" is what `cargo check --tests` is. What would have caught this class is a gated `--tests` compile **on push to main** rather than only on PRs; #1197 raises that as a follow-up thought, and I have deliberately not bundled it into this fix.

## Documentation

None needed. #1182's note at `ports/types.rs:2001` already documents why the fields are gone, and it stays as written.

### DIFF
--- a/src/harness/spend_halt_turn_test.rs
+++ b/src/harness/spend_halt_turn_test.rs
@@ -376,9 +376,6 @@ fn chat(text: &str) -> CycleRequest {
             deliverable: None,
         }],
         event_seqs: Vec::new(),
-        compressed_history: Vec::new(),
-        roster: Vec::new(),
-        context_index: Vec::new(),
     }
 }
 


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
## The fix works. The gated lane is now red for a different, pre-existing reason.

**The compile is fixed — that is measurable from the two logs side by side:**

| | `main` (`ef0436f4`, [run 32299709882](https://github.com/tinyhumansai/opencompany/actions/runs/32299709882)) | this PR |
|---|---|---|
| outcome | `error[E0560]` ×3, **could not compile** | compiles |
| tests executed | **0** — it never got that far | **4,656+**, one failed |

So the lane moved from "cannot build the test targets" to "builds and runs the suite". That is exactly what #1197 asked for.

## What is failing now: a flaky conformance test, not this diff

```
store::memory::test::namespace_driver_conformance::isolation_by_company_holds_on_the_namespace_driver
  panicked at src/store/conformance.rs:341:5: assertion `left == right` failed
    left: 0   right: 1
```

Different feature set (`--features acp,runner,tinymemory-embedded --lib store::memory`), different subsystem (the in-memory namespace store driver). This PR deletes three `Vec::new()` initialisers from a harness test fixture.

**I did not stop at "it can't be mine" — I ran it.** Seven consecutive runs of the command on this branch, byte-identical tree, nothing changed between them:

```
ok, ok, ok, FAILED, ok, FAILED, ok        → 5 passed, 2 failed
```

Identical code, alternating results. And the member that fails varies: CI hit `isolation_by_company_holds_on_the_namespace_driver` (`conformance.rs:341`), locally it was `export_totality_holds_on_the_namespace_driver` (`conformance.rs:911`) both times. Same suite, different victim, non-deterministic — the signature of shared state or ordering between the conformance cases rather than a defect in any one of them.

*(Method note, since it matters for reading the numbers: I first intended a with/without comparison, but the fix was already committed, so the `git stash push -- <file>` I ran stashed nothing. All seven runs were therefore the same tree — which is a stronger flake demonstration than the comparison would have been, but it is seven runs of one configuration, not a controlled A/B.)*

It is ~30% failing locally, so a re-run should clear the lane. Flagging it rather than folding a fix into an urgent three-line unblock — it deserves its own issue, and it is worth one: an intermittently-red gated lane is precisely what lets a real breakage get waved through as "the known flake", which is the same failure mode #1197 was filed about.

Every other lane is green: `Rust`, `Rust (mongodb)`, `Desktop`, `Gated host binary`, `Console E2E`, `Changed areas`, `Actions pinned`.

M3gA-Mind:
Followed this up rather than leaving it in a comment: **#1201**.

The re-run cleared the lane (`Rust (openhuman, tinycortex)` pass, 27m58s) and #1200 is merged, so `main` compiles its gated test targets again — verified directly against `upstream/main` after the merge: `cargo check --locked --features openhuman,tinycortex --tests` exits **0**.

Two corrections to what I wrote above, now that I have measured further:

1. **It is not test parallelism.** Single-threaded runs still fail 2 of 4, which kills that hypothesis.
2. **It may not be a test defect at all.** The failing assertion shows three traces written and two read back, with the **middle** one missing (`c1` gone, `c2` written after it present). A flush lag truncates the tail; losing a middle record while the newer one lands is a different thing. #1201 has the full data and flags the timestamp-collision theory as unverified.

Across 11 local runs on an unchanged tree: 4 failures, ~36%.

Issue #1201, opened 2026-08-19
Embedded namespace memory driver intermittently loses writes: namespace_driver_conformance fails ~36% of runs, dropping a middle record


M3gA-Mind:
## Summary

`store::memory::test::namespace_driver_conformance` fails intermittently on a byte-identical tree — **~36% of runs** across 11 local runs. It reddens the `Rust (openhuman, tinycortex)` lane at random on any PR.

More importantly, what it is catching does not look like a test defect. **The embedded `namespace` memory driver appears to lose writes**: a read-back returns two of three records, and the *middle* one is the one missing.

Found while fixing #1197 (PR #1200); it is unrelated to that diff, which deleted three `Vec::new()` initialisers from a harness fixture.

## The evidence that it is a dropped write, not a lag

`assert_export_totality` writes three compressed traces and reads them back (`conformance.rs:907`, *"All traces round-trip, newest last"*):

```
left  (read back): [c0 @ 1787178633770,                          c2 @ 1787178633774]
right (expected):  [c0 @ 1787178633770, c1 @ 1787178633773, c2 @ 1787178633774]
```

**`c1` is gone while `c2`, written after it, survived.** A flush lag or eventual-consistency window would truncate the *tail*; losing a middle record while the newer one lands is a different failure.

The same suite also fails one assertion further on (`conformance.rs:911`, context chunks): `left: 2, right: 3` — one of three chunks not listable. So it is not specific to traces.

**A timestamp collision is the obvious hypothesis and is NOT verified.** `c1 @ ...773` and `c2 @ ...774` are 1 ms apart, while the surviving `c0` is 3 ms earlier — consistent with records being keyed or deduplicated by `at_millis` and colliding when two writes land in adjacent milliseconds. That would also explain the intermittency directly: whether two writes share a millisecond depends on machine speed and load. **Someone who knows the driver should confirm or kill this before acting on it.**

## Reproduction

```
cargo test --locked --features acp,runner,tinymemory-embedded --lib store::memory
```

| configuration | runs | failures |
|---|---|---|
| default threads | 7 | 2 |
| `-- --test-threads=1` | 4 | 2 |
| **total** | **11** | **4 (~36%)** |

Nothing changed between runs — same commit, same binary, no edits.

**It is not test parallelism.** That was my first hypothesis and single-threaded runs killed it: it still fails 2 of 4 serially. Each test also builds its own `tempfile::tempdir()`, so it is not a shared fixture directory either.

[...]

PR #1203, opened 2026-08-20
fix(app): attach a stranded note to the test it describes

oxoxDev:
## Summary

`cargo clippy --all-targets --no-deps --features openhuman,mcp,telegram -- -D warnings` fails on `src/app/types.rs` with `empty_line_after_doc_comments`. This fixes it, but not the way the lint suggests.

The stranded comment reads:

> The registered-company half of this lives in `server::setup::test`, beside the helper that can build a real runtime: `spec_reports_setup_complete_once_a_company_is_registered`.

That describes `spec_reports_setup_incomplete_for_an_empty_unstamped_host`, the test **above** it — the "incomplete" half whose counterpart lives elsewhere. The item **below** it is `parked_oauth_flow_is_single_use`, an MCP OAuth test with no relation to setup completeness.

So the one-line fix the lint nudges you toward — close the blank line — would have silenced it by making the note that OAuth test's documentation, planting a false cross-reference for the next reader. Moved onto the test it actually describes instead.

## API Or Behavior Changes

None. Comment placement only, inside a `#[cfg(test)]` module.

## Tests

- `cargo fmt --check` — clean
- `cargo clippy --all-targets --no-deps --features openhuman,mcp,telegram -- -D warnings` — now clean; fails on `main` without this

## Documentation

None needed.

## Why CI never caught it

The lint site is `#[cfg(feature = "mcp")]`, and the gated lane runs `--features openhuman,tinycortex`. The test is never compiled there, so the lint never fires in CI while being red for anyone building locally with `mcp` — an instance of the known full-feature gap, where no lane builds the `openhuman,mcp,telegram` combination.

Worth deciding separately whether a lane should cover that combination. This PR only clears the existing breakage.

## Related

Found while running local gates for #1186. Independent of it — cut from a clean `main`.

## Summary by CodeRabbit

* **Tests**
  * Clarified test comments documenting setup-completion coverage.
  * Removed a duplicate comment related to the MCP OAuth test.

### DIFF
--- a/src/app/types.rs
+++ b/src/app/types.rs
@@ -1204,6 +1204,10 @@ mod tests {
     }
 
     /// A host with nothing to open is the only one the wizard is for.
+    ///
+    /// The registered-company half of this lives in `server::setup::test`,
+    /// beside the helper that can build a real runtime:
+    /// `spec_reports_setup_complete_once_a_company_is_registered`.
     #[test]
     fn spec_reports_setup_incomplete_for_an_empty_unstamped_host() {
         let spec = AppState::new(AppConfig::default()).spec();
@@ -1213,10 +1217,6 @@ mod tests {
             "no stamp and no companies is exactly the first-run case"
         );
     }
-
-    /// The registered-company half of this lives in `server::setup::test`,
-    /// beside the helper that can build a real runtime:
-    /// `spec_reports_setup_complete_once_a_company_is_registered`.
 
     #[cfg(feature = "mcp")]
     #[test]


oxoxDev:
CI note — the red lane here is not this change.

`Rust (openhuman, tinycortex)` fails on `store::memory::test::namespace_driver_conformance::export_totality_holds_on_the_namespace_driver`. This branch changes one doc comment in `src/app/types.rs`; it cannot reach `store::memory`.

Evidence:

- The same test fails the same lane on an unrelated branch of mine, #1234, with a different symptom — there it drops the oldest trace (`c0`) at `conformance.rs:907`; here it drops a context chunk at `:911`.
- It passes locally on this exact commit: `cargo test --features acp,runner,tinymemory-embedded --lib store::memory` → `67 passed; 0 failed`. Failed twice in CI on the same commit.
- The lane is `SKIPPED` on most PRs and on main merge commits, which is why this is not more visible.

Already tracked: #1201 (M3gA-Mind / YellowSnnowmann) and #1238. Those two look like the same defect from two angles — worth deduping.

Nothing to do on this PR beyond a rerun once the driver bug lands.

PR #1213, opened 2026-08-20
workflows: take the run input off the detail toolbar, behind Run

### LINKED ISSUE #1204: workflows: take the run-input box off the detail toolbar without losing the capability
The workflow detail toolbar carries a wide free-text field — **"What should this run work on?"** — sitting between the workflow's description and the Run button. It should come off the toolbar.

It is the widest element in a row that otherwise holds nine controls, it is empty on almost every visit, and it puts a text input in the operator's way for the common case: pressing Run.

## What it does, so removing it is done with eyes open

It is not decoration. `WorkflowsView.tsx:346` (`ranWith`, issue #154) is the trigger input: the host seeds it as the trigger node's item (`src/server/ops/workflows.rs:1190-1195` — *"an optional trigger `input` payload seeded as the trigger node's item. An empty object (`{}`) runs with a null input."*). It then flows downstream through `=items` bindings, and it is echoed back on the run's detail as "Requested:".

So a graph whose first step reads `=items` genuinely depends on this. **Deleting the field outright would silently break those workflows** — the run would proceed with a null input rather than refuse, which is the worst version of this change.

## What to do

Take it off the toolbar without taking the capability away. Options, in rough order of preference:

1. **Behind Run.** A split control — Run runs; a small adjacent affordance opens "Run with input…". The common case is one click, the rarer case is one more.
2. **In the run dialog**, if one is introduced.
3. **Only when the graph needs it** — if the trigger or any node references `=items`, show the field; otherwise do not. Sharpest for the operator, but it means reading the graph's bindings, so only do this if that is cheap and reliable.

**Do not simply delete it.** If option 3 is chosen, be careful: a graph that reads `=items` indirectly must not be misread as not needing input.

## Also worth checking

`ranWith` is pinned only on success (`WorkflowsView.tsx:1094`), so the "Requested:" echo on a failed run may be missing or stale — worth confirming while you are in this code, and filing separately if it is wrong.

graycyrus:
Closes #1204

## What changed

The workflow detail toolbar's free-text field — **"What should this run work on?"** — is off the bar. `Run` is now a split control: the left half runs, the attached right half opens a **Run with input** dialog carrying the field.

| before | after |
| --- | --- |
| `[ What should this run work on?…… ] [Run] │ [Test run] [Copilot] [History] [Edit] [Delete]` | `[Run│⌄] │ [Test run] [Copilot] [History] [Edit] [Delete]` |

## Why option 1 (behind Run), and not 2 or 3

**Option 2 (in a run dialog)** presupposes a run dialog this console does not have. Introducing one would put a modal in front of the *common* case — pressing Run — which is the same complaint the issue is making, moved one step later.

**Option 3 (show it only when the graph references `=items`)** is the one the issue itself flags as conditional on the graph read being "cheap and reliable", and it is neither. A node can reach the trigger's item through a chain of upstream bindings, through a `sub_workflow` whose own graph is a separate document, or through an expression that merely contains `items` as a substring. Every one of those has to be resolved correctly or the field disappears from a workflow that needs it — and the failure mode is the exact one the issue names as the worst outcome, arriving quietly and only for the graphs that depend on the payload most. Option 1 costs the rare case one extra click and cannot be wrong about which graphs those are.

## The capability is intact, and is now harder to break

`ranWith`, the `{ request: asked }` payload, the trim, the `runFailureFrom` capture, and the "Requested:" echo are all unchanged. Three things were added around them:

1. **The payload is a parameter of `run()`, not state it closes over.** This is the load-bearing change. `run()` used to read `request` out of its closure, so simply relocating the box behind a dialog would have left a draft typed once riding along on every later press of Run, with nothing on screen to say so — a visible field turned into an invisible one, which is worse than the bar it was removed from. Now the toolbar's Run and Test run pass nothing and therefore genuinely run with no input.
2. **The dialog carries both dispatches.** The old box fed Run *and* Test run, so a dialog offering only Run would have quietly removed "prove this graph against a real input without sending anything" — the rehearsal that matters most for exactly the workflows this input exists for.
3. **The draft is cleared on workflow switch**, alongside the run drawers that are already cleared there for the same reason.

Enter-to-run, the disabled predicates, and the `sr-only` accessible name are all preserved.

## `ranWith` on a failed run — checked, and it is fine

The issue asks whether the "Requested:" echo is missing or stale on a failed run, since `ranWith` is pinned only on the success arm. It is correct today, and #1007 is why:

- `setResult` has six call sites and only one sets a value — the success arm, one line after `setRanWith(asked)`. `RunResultPanel` cannot mount against any other run, so its `request={ranWith}` is always that run's.
- The failed path renders `RunFailurePanel`, which reads `failure.request` — captured in the same `catch` from the same `asked` local, independently of `ranWith`.

Verified in the browser: a run that failed showed **`Requested: the Q3 board deck`** in the failure drawer. No separate issue filed.

## Verification

Browser, against a local host with the console on Vite, light and dark:

- toolbar before/after — the field is gone, the row is eight controls instead of nine plus a 256px input;
- the dialog's Run put **`{"input":{"request":"the Q3 board deck"}}`** on the wire (captured off the outgoing POST, not inferred from the UI);
- the run detail echoed `Requested: the Q3 board deck` — on a host **with** execution wired (`--features openhuman` + the suite's `mock-brain.mjs`), so this is the success drawer and a real run, not the failure path;
- that run's engine output confirms the payload went where issue #154 says it does — `run.trigger = {"request":"the Q3 board deck"}`, seeded as the trigger node's `items`, and carried downstream to the next node's `items`. The `=items` path is intact;
- reopening the dialog restored the draft.

Commands: `npm run typecheck`, `typecheck:unit`, `typecheck:e2e`, `npm test` (137 files, 1330 tests), and `npx playwright test workflow-` — 49 passed, 4 skipped (the `LIVE_BRAIN`-gated ones).

## Tests

Updated, not weakened:

- **`test/unit/workflow-toolbar-layout.test.ts`** — `"puts the actions on row 2…"` gains `workflow-run-split` / `workflow-run-with-input` in the expected order, and its old "the field and Run share a parent" assertion is replaced by the **stronger** "the toolbar mounts no request field at all", which is false for the old layout and for any half-measure that merely narrowed the box. `"makes Run the detail screen's primary, and nothing else"` now expects the two halves of the one split control, *and* asserts every filled button is inside `workflow-run-split` — so a genuinely second primary elsewhere still fails.
- **`test/e2e/workflow-toolbar-reachable.spec.ts`** — the `Run` entry moves from `getByRole("button", { name: "Run", exact: false }).first()` (ambiguous the moment a second Run-ish control exists) to `getByTestId("workflow-run")`, and `Run with input` is added so the newest control on the row is held to the same in-viewport/clickable property.

Added:

- **`test/unit/workflow-run-input.test.ts`** (10 tests) — delivery, trimming, empty-field `{}`, test-run parity, Enter-to-run, draft retention, the "Requested:" echo, and the invariant that the toolbar's Run and Test run post `input: {}` even with a draft sitting in the dialog.
- **`test/e2e/workflow-run-with-input.spec.ts`** — the same end to end, asserting on the **outgoing POST body** rather than on the rendering, so it does not depend on whether the host under test has a brain.

### DIFF
--- a/frontend/src/views/WorkflowsView.tsx
+++ b/frontend/src/views/WorkflowsView.tsx
@@ -12,6 +12,7 @@ import { useTheme } from "next-themes";
 import {
   ArrowLeft,
   Bot,
+  ChevronDown,
   FlaskConical,
   History,
   LayoutGrid,
@@ -66,6 +67,14 @@ import {
 import { Alert, AlertDescription } from "@/components/ui/alert";
 import { Badge } from "@/components/ui/badge";
 import { Button, buttonVariants } from "@/components/ui/button";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
 import { Input } from "@/components/ui/input";
 import { Skeleton } from "@/components/ui/skeleton";
 import { WorkflowNode } from "@/components/workflow-node";
@@ -342,8 +351,21 @@ export function WorkflowsView({
   // Issue #154: what the operator is asking this run to work on. `ranWith` is
   // pinned when the run is dispatched so the result panel echoes the request the
   // shown output came from, not whatever has been typed since.
+  //
+  // Issue #1204: `request` is the DRAFT held by the "Run with input" dialog, not
+  // an implicit argument to running. It used to be a text box wired straight
+  // into `run()`'s closure, sitting open on the toolbar; taking the box off the
+  // bar without also cutting that wire would have been worse than leaving it
+  // there, because a draft typed once and then forgotten would keep riding
+  // along on every later press of Run with nothing on screen to say so. So the
+  // payload is a PARAMETER of `run()` now (see below) and this state is only
+  // ever read at the moment a dialog button is pressed. It survives the dialog
+  // closing on purpose — reopening restores what was typed, which is what makes
+  // "run it again with a tweak" one edit rather than a retype.
   const [request, setRequest] = useState("");
   const [ranWith, setRanWith] = useState("");
+  // Issue #1204: whether that dialog is up.
+  const [runInputOpen, setRunInputOpen] = useState(false);
   // Issue #1002: the roster read behind the cards' "Asked by" line, keyed to the
   // approvals of the run on screen rather than to the whole queue — so a console
   // with a run drawer closed does not fetch the roster for cards it is not
@@ -1079,7 +1101,13 @@ export function WorkflowsView({
     return byId;
   }, [indexRuns]);
 
-  const run = useCallback(async (dryRun = false) => {
+  // `input` is the trigger payload for THIS dispatch (issue #1204), handed in by
+  // whichever control fired rather than read out of `request` state. The toolbar's
+  // Run and Test run pass nothing and therefore genuinely run with no input; only
+  // the "Run with input" dialog passes its draft. That is the whole guarantee the
+  // dialog rests on: there is no way to run with a payload the operator cannot
+  // see at the moment they press the button.
+  const run = useCallback(async (dryRun = false, input = "") => {
     if (!selectedId) return;
     setStarting(true);
     // Clear the last refusal and the own-run-start watch before this attempt, so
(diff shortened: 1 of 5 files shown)


graycyrus:
@coderabbitai review

Issue #1180, opened 2026-08-19
chat: a DM header prints the teammate's name twice when the host sends no display name


graycyrus:
Open a DM with a teammate the host sends no display name for and the chat header
says the same words twice:

```
🫥  Backend Engineer  │  Backend Engineer
    ↑ the title           ↑ the muted slot after the divider
```

Confirmed on `companies/agentic_software_company` — every teammate in that
company reads this way.

## Cause

The two slots hold different fields that happen to resolve to the same string.

`frontend/src/views/chat/ChatHeader.tsx:78` renders the muted slot as
`channel.purpose`, and `buildChannels` sets a DM's `purpose` to the teammate's
**role** (`frontend/src/views/chat/model.ts`, the `dms` map: `purpose: m.role`).
The title is `channel.name`, which is the teammate's **name**.

Those are only distinct when the host actually sends a name. `fromDto` in
`frontend/src/lib/team.ts` falls back:

```ts
const name = dto.name?.trim() || dto.role;
```

So a manifest that declares a role and no `name` — which is every agent in the
sample company — makes name and role the same string, and the header prints it
on both sides of the divider.

## Why it is worth fixing

The muted slot is the one place in the header with room to say something about
who you are talking to, and a DM has a `description` sitting unused on its
roster entry (`TeamMember.description`, already carried through by `fromDto`).
The conversation-intro line under the timeline already reaches for it. A header
that repeats itself is spending that slot on nothing.

Some shape of: for a DM, prefer the teammate's `description`, fall back to the
role, and draw nothing rather than a repeat when the fallback equals the title.
A channel's `purpose` is a genuinely different string from its slug and should
keep working exactly as it does.

## Notes

Split out of #1170 (PR #1179), which fixed the header's avatar and deliberately
left this alone.

PR #1215, opened 2026-08-20
fix(company): say the teammate's name once

### LINKED ISSUE #1208: company: teammate cards and org-chart seats both print the teammate's name twice
Every seat on the Company page prints the same words twice — the teammate's
name, then their role, when the two are the same string.

On `companies/agentic_software_company`, all 12 seats read this way. Measured
from the DOM (`[data-seat-id]`, innerText, newlines shown as `|`):

```
backend_engineer    Backend Engineer    | Backend Engineer    | Blueprint
frontend_engineer   Frontend Engineer   | Frontend Engineer   | Blueprint
qa_engineer         QA Engineer         | QA Engineer         | Blueprint
security_engineer   Security Engineer   | Security Engineer   | Blueprint
product_manager     Product Manager     | Product Manager     | Blueprint
designer            Designer            | Designer            | Blueprint
devrel              Developer Relations | Developer Relations | Blueprint
docs_writer         Documentation Writer| Documentation Writer| Blueprint
customer_support    Customer Support    | Customer Support    | Blueprint
growth_analyst      Growth Analyst      | Growth Analyst
researcher          Researcher          | Researcher
writer              Writer              | Writer
```

12 of 12. Not an edge case in this company — it is the whole chart.

## Cause

Same root as #1180, on a different surface. `Seat`'s label renders both fields
back to back:

- `frontend/src/views/company/OrgChartView.tsx:795` — `{seat.name}`
- `frontend/src/views/company/OrgChartView.tsx:797-801` — `{seat.role &&
  <span …>{seat.role}</span>}`

and `buildOrgTree` fills both from the same roster row
(`frontend/src/lib/org.ts:154-155`):

```ts
name: member?.name ?? id,
role: member?.role ?? "",
```

where `member` came through `fromDto` (`frontend/src/lib/team.ts`):

```ts
const name = dto.name?.trim() || dto.role;
```

The host genuinely sends no `name` for a manifest-declared agent — `GET
{scope}/team` answers `{"id":"backend_engineer","role":"Backend Engineer",…}`
with no `name` key at all — so `fromDto` falls back to the role and the two
slots collapse onto one string.

## Why it matters

The muted second slot is the only place on a seat row with room to say
something about the teammate, and every roster row already carries a
`description` (`"Build and operate the backend and services."`) that the chart
never shows. A row that repeats itself spends that slot on nothing, and at 12
rows the repetition is what the page mostly consists of.

It also makes the chart read as a rendering fault rather than as data — the
first question an operator asks is "why is it saying it twice", not "who is on
this desk".

## What a fix would be

In `Seat`, draw the role slot only when it differs from the name — the same
rule #1180 proposes for the DM header — and prefer the teammate's
`description` for that slot when there is one, falling back to the role, and to
nothing when the fallback equals the title.

`OrgSeat` would need `description` carried through `buildOrgTree` (it is
already on `TeamMember`, so it is a field copy, not a new read).

## Repro

```
cargo build --bin opencompany --features openhuman,tinycortex,mcp
OPENCOMPANY_DATA_DIR=<dir> OPENCOMPANY_ADMIN_EMAIL=<you> \
  OPENCOMPANY_CONSOLE_DIR=frontend/dist \
  ./opencompany serve --bind 127.0.0.1:8391 --company companies/agentic_software_company
# sign in, open #/company
```

graycyrus:
## Summary

The Company page said every teammate's name twice.

A roster row carries a `name` and a `role`. `fromDto` resolves the name as
`dto.name?.trim() || dto.role`, and a manifest-declared agent sends no `name` at
all — `GET {scope}/team` answers `{"id":"backend_engineer","role":"Backend
Engineer",…}` — so for every agent in every shipped company the two fields are
one string. Both surfaces that drew the title and then the role underneath drew
the same words twice.

Measured on `companies/agentic_software_company` before this change: 14 of 14
cards and 12 of 12 org-chart seats. Not an edge case — the whole page.

```
before                          after
Backend Engineer                Backend Engineer
Backend Engineer                Build and operate the backend and services.
Build and operate the …         Idle · 2 open tasks
Idle · 2 open tasks
```

On the card the repeat cost the most: it sat directly above the description
that actually says what the teammate does.

## What changed

- `frontend/src/lib/team.ts` — `roleSubtitle(name, role)`, the single rule.
  Trimmed and case-insensitive, because the two strings come from different
  places (one may have been typed into the add-teammate dialog, the other came
  out of a manifest) and "Backend engineer" under "Backend Engineer" is the same
  repeat in another costume. Answers `null` rather than `""` so a caller skips
  the element instead of rendering an empty one and keeping the gap.
- `frontend/src/views/TeamView.tsx` — the Company card's second line.
- `frontend/src/views/company/OrgChartView.tsx` — the org-chart seat's role span.

Nothing else moves: a teammate the host *does* name keeps both lines, which is
the whole point of the slot.

## Verified in the browser

Against a real host (`--features openhuman,tinycortex,mcp`, sample company,
signed in), with a teammate added specifically to hold the positive case:

```
#/company        BE  Backend Engineer / Build and operate the backend and services.
                 AL  Ada Lovelace / Data Scientist / Model the funnel and forecast.
#/company/desks  backend_engineer -> Backend Engineer | Blueprint
                 ada_lovelace     -> Ada Lovelace | Data Scientist
```

## Commands run

```
npx vitest run test/unit/role-subtitle.test.ts   # 7 passed
npm run typecheck
npm run typecheck:unit
npm run build
scripts/ci/assert-design-tokens.sh
```

## Not in scope

#1180 — the DM header's version of the same repeat. It has a different remedy
(that slot wants the teammate's `description`, not nothing), and it is filed
separately.

Closes #1208

### DIFF
--- a/frontend/src/lib/team.ts
+++ b/frontend/src/lib/team.ts
@@ -145,6 +145,33 @@ export function initials(name: string): string {
   );
 }
 
+/**
+ * The muted line that sits under a teammate's name, or `null` when there is
+ * nothing left to say (issue #1208).
+ *
+ * A roster row carries a `name` and a `role`, and the console's own fallback
+ * makes them the same string whenever the host sends no display name:
+ * `fromDto` below resolves `name` as `dto.name?.trim() || dto.role`, and a
+ * manifest-declared agent has no `name` at all — `GET {scope}/team` answers
+ * `{"id":"backend_engineer","role":"Backend Engineer",…}`. Every surface that
+ * drew the title and then the role drew the same words twice.
+ *
+ * Answering `null` rather than an empty string is deliberate: a caller has to
+ * skip the element, not render an empty one, or the layout keeps the gap the
+ * repeat used to fill.
+ *
+ * Compared case-insensitively and whitespace-trimmed, because the two strings
+ * come from different places — one may have been typed by an operator into the
+ * add-teammate dialog while the other came out of a manifest — and "Backend
+ * engineer" under "Backend Engineer" is the same repeat wearing a different
+ * capital.
+ */
+export function roleSubtitle(name: string, role: string): string | null {
+  const trimmed = role.trim();
+  if (!trimmed) return null;
+  return trimmed.toLowerCase() === name.trim().toLowerCase() ? null : trimmed;
+}
+
 /** Map a host roster entry into the console's team model. */
 export function fromDto(dto: TeamMemberDto): TeamMember {
   const name = dto.name?.trim() || dto.role;
--- a/frontend/src/views/TeamView.tsx
+++ b/frontend/src/views/TeamView.tsx
@@ -36,7 +36,7 @@ import {
   reportAddMember,
   type MissedStep,
 } from "@/lib/member-feedback";
-import { fromDto, newMember, type TeamMember } from "@/lib/team";
+import { fromDto, newMember, roleSubtitle, type TeamMember } from "@/lib/team";
 import { workloadByAssignee, type Workload } from "@/lib/team-workload";
 import { cn } from "@/lib/utils";
 import { Avatar } from "@/views/chat/Avatar";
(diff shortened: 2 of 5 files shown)


graycyrus:
Extended to a third surface found while re-driving the console: the **chat
members pane** draws the same pair, and 14 of its 15 rows read the name twice.

```
before                          after
BE  Backend Engineer  LEAD      BE  Backend Engineer  LEAD
    Backend Engineer
AL  Ada Lovelace                AL  Ada Lovelace
    Data Scientist                  Data Scientist          ← still both
```

Same one-line rule, same helper — `frontend/src/views/chat/MembersPane.tsx`.

That makes three surfaces on one rule: Company cards, org-chart seats, chat
members pane. The DM header (#1180) is still deliberately out of scope: its
slot wants the teammate's `description` rather than nothing, which is a
different remedy.

PR #1242, opened 2026-08-20
fix(console): show a boot state instead of a white flash before React mounts

### LINKED ISSUE #1241: the console shows a blank white page while a scaled-to-zero tenant cold-starts
## What you see

Open a hosted tenant that has been idle and the browser sits on a **blank white
page** for several seconds, with no spinner, no message, and no indication
anything is happening. It then renders correctly.

Reported against `marketing.staging.opencompany.work`, which looked broken
beside `smoke1.staging.opencompany.work` rendering instantly. Nothing was broken:
`smoke1` is polled continuously so it is always awake, and `marketing` had to
cold-start.

## Why it happens

Hosted tenants scale to zero when idle (opencompany-microservice#19). A request
to a parked tenant is **held** by the manager's proxy while the pod boots and
`/healthz` passes, then forwarded. Measured on staging today: **7.8s–12.8s**.

The infrastructure is already built for this — the edge nginx carries

```
# Must exceed the manager's OCM_STARTUP_TIMEOUT_SECONDS (30s) so a cold
# tenant wake (0 -> 1) does not 504 at the edge before the manager
# finishes booting the pod.
proxy_read_timeout 120s;
```

so every layer *except the console* knows a cold start can take tens of seconds.
The white page is simply the browser holding an empty document while the initial
request is in flight — React has not mounted, so not even the dark theme applies.

Caught mid-load, the sequence is visible:

```
t+0    blank white — index.html still in flight
t+~8s  shell renders, "Pick a pillar / 0 PILLARS", "Loading…"
t+~10s "3 PILLARS", full graph
```

## Why it is worth fixing rather than accepting

The trade behind scale-to-zero was roughly 3x density for an ~8s wake, and that
is a good trade. But it moved the cost onto first paint, and the console never
learned about it — so the platform got cheaper to run and *looks* broken to
whoever opens a quiet company. An operator cannot distinguish "waking" from
"down", which is exactly the distinction opencompany-microservice#19 phase 5
spent effort making legible at the API layer (typed 503s with `Retry-After`
instead of a masked 500). The browser is where that legibility matters most and
currently has none.

## Possible shapes, not a prescription

- Something in `index.html` before the bundle loads — a static skeleton or
  spinner that React replaces on mount. Cheapest, and covers the white-page
  window specifically, which is the part with no JS running at all.
- A "waking this company…" state, if the console can distinguish a cold wake from
  an ordinary slow load. Phase 5 gives the manager typed 503 + `Retry-After` for
  the retryable cases, so there is a signal to key on if it is surfaced.

The second is nicer and needs a decision about what the manager exposes; the
first is small and independently useful.

## Not a regression

Nothing changed to cause this. Scale-to-zero has been in place since phase 1, and
the console has never had a first-paint state. It became noticeable now because
more tenants sit parked more of the time — measured at **75.7% parked** across
the staging fleet over three days of sampling.

CodeGhost21:
Closes #1241.

## The problem

Open a hosted company that has been idle and the browser sits on a **blank white
page** for several seconds — no spinner, no message, nothing. It then renders
fine.

Reported as `marketing.staging.opencompany.work` looking broken next to
`smoke1.staging.opencompany.work`. Nothing was broken: `smoke1` is polled
continuously so it is always awake; `marketing` had to cold-start.

Hosted companies scale to zero when idle
(opencompany-microservice#19), so the first request to a quiet one is *held*
while the pod boots. Measured on staging: **8–13s**. For that entire window the
browser has an empty document — no bundle, no React, not even the theme, which
is why it is white rather than dark.

Caught mid-load, the sequence is visible:

```
t+0     blank white     index.html still in flight
t+~8s   shell renders   "0 PILLARS", "Loading…"
t+~10s  full graph      "3 PILLARS"
```

## Why it is worth fixing rather than accepting

Every other layer already knows cold starts happen — the edge nginx carries

```
# Must exceed the manager's OCM_STARTUP_TIMEOUT_SECONDS (30s) so a cold
# tenant wake (0 -> 1) does not 504 at the edge before the manager
# finishes booting the pod.
proxy_read_timeout 120s;
```

The console was the only layer with no notion of it, so it presented "waking" and
"down" identically. That is the same distinction #19 phase 5 spent real effort
making legible at the API layer (typed 503 + `Retry-After` instead of a masked
500) — missing in the one place a person actually looks.

## The change

One file. A boot placeholder inside `#root`, which `createRoot` replaces on first
render.

**Deliberately inert** — inline styles and markup, no script, no stylesheet link.
It paints on the first byte of HTML, which matters because the white window is
precisely the period when nothing else has loaded. It uses the app's own
`--surface-*-bg` values (`#F7F7FC` / `#08090B`) and follows
`prefers-color-scheme`, so handover to the real shell is not a flash of a
different colour.

**Copy says "waking", not "loading"** — that is the true claim. Nothing is broken
and nothing is slow; a machine is being started.

## The bit worth reviewing

**After 35s it stops claiming this is normal.** A placeholder saying "waking"
forever would be a *worse* lie than the blank page it replaces — someone whose
company genuinely failed to start would sit reading a reassuring message
indefinitely.

Done with a delayed CSS animation rather than script, so the placeholder stays
inert. 35s is past both the measured wake and the manager's 30s startup timeout.
`prefers-reduced-motion` keeps the indicator and drops the spin.

## Verification

- `npm run build` passes (`tsc -b && vite build`)
- the placeholder is present in `dist/index.html`, ahead of the deferred
  `type="module"` bundle
- serving the built output confirms React clears it on mount with nothing left
  behind

## Not a regression

Scale-to-zero has been in place since phase 1 and the console has never had a
first-paint state. It became noticeable because tenants now sit parked far more
of the time — **75.7%** across the staging fleet over three days of sampling.

## Summary by CodeRabbit

* **New Features**
  * Added a loading screen with a spinner and “Waking this company…” message during startup.
  * Displays a slow-start warning if loading takes longer than 35 seconds.
  * Added support for dark mode and reduced-motion preferences.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -21,11 +21,16 @@
     <!--
       Boot placeholder, replaced by React on mount.
 
-      A hosted company scales to zero when idle, so the first request to a quiet
-      one is *held* by the platform while its pod boots — measured at 8-13s. For
-      that whole window the browser has an empty document: no bundle, no React,
-      not even the theme. The page rendered plain white and was indistinguishable
-      from a broken deployment (#1241).
+      Covers the window between the HTML arriving and React mounting, when the
+      browser has a document but no bundle, no React and not even the theme —
+      so it would otherwise paint plain white.
+
+      It does NOT cover a cold start, and that limit is worth stating here so
+      nobody assumes otherwise. A hosted company scales to zero when idle and the
+      platform *holds* the request while its pod boots: measured TTFB on a parked
+      tenant is 8.2s, and this markup lives inside the HTML that has not arrived
+      yet. Fixing that needs a holding response from something that is always up
+      — the edge, or the manager's proxy — not the console (#1241).
 
       This is deliberately inert: inline styles and markup only, no script and no
       stylesheet link, so it paints on the first byte of HTML rather than waiting
@@ -50,7 +55,7 @@
           animation:boot-spin .8s linear infinite; opacity:.55;
         "></div>
         <div>Waking this company&hellip;</div>
-        <div id="boot-hint" style="font-size:12px; opacity:.6;">Idle companies start on demand, which takes a few seconds.</div>
+        <div id="boot-hint" style="font-size:12px;">Idle companies start on demand, which takes a few seconds.</div>
         <div id="boot-slow" style="font-size:12px; opacity:0; max-width:34ch; text-align:center;">
           This is taking longer than a normal wake. If it does not finish, the
           company may have failed to start rather than be starting.
@@ -64,13 +69,18 @@
          message that says "waking" forever would be a worse lie than the blank
          page it replaces. 35s is comfortably past both the measured 8-13s wake
          and the manager's 30s startup timeout. */
-      @keyframes boot-reveal { to { opacity:.75 } }
+      @keyframes boot-reveal { to { opacity:1 } }
       @keyframes boot-fade   { to { opacity:0 } }
       #boot-slow { animation: boot-reveal .4s ease 35s forwards }
       #boot-hint { animation: boot-fade .4s ease 35s forwards }
       /* Reduced motion: keep the indicator, drop the spin. */
       @media (prefers-reduced-motion: reduce) {
+        /* The spinner is decorative, so it stops entirely. The status swap is
+           not — it carries the "may have failed" message — so it keeps its 35s
+           delay and loses only the fade. `animation: none` here would suppress
+           the reveal and leave the misleading copy up forever. */
         #boot [style*="boot-spin"] { animation: none !important }
+        #boot-slow, #boot-hint { animation-duration: 0s !important }
       }
       @media (prefers-color-scheme: dark) {
         #boot { background:#08090B !important; color:#8A8A9A !important }
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -18,7 +18,64 @@
     -->
   </head>
   <body>
-    <div id="root"></div>
+    <!--
+      Boot placeholder, replaced by React on mount.
+
+      A hosted company scales to zero when idle, so the first request to a quiet
+      one is *held* by the platform while its pod boots — measured at 8-13s. For
+      that whole window the browser has an empty document: no bundle, no React,
+      not even the theme. The page rendered plain white and was indistinguishable
+      from a broken deployment (#1241).
+
+      This is deliberately inert: inline styles and markup only, no script and no
+      stylesheet link, so it paints on the first byte of HTML rather than waiting
+      for anything else to load. It matches the app's own `--surface-*-bg`
+      primitives and follows `prefers-color-scheme`, so the handover to the real
+      shell is not a flash of a different colour.
+
+      The copy says "waking" rather than "loading" because that is the true
+      claim: nothing is broken and nothing is slow, a machine is being started.
+    -->
+    <div id="root">
+      <div id="boot" style="
+        position:fixed; inset:0; display:flex; flex-direction:column;
+        align-items:center; justify-content:center; gap:14px;
+        background:#F7F7FC; color:#6B6B7B;
+        font-family:system-ui,-apple-system,'Segoe UI',Roboto,sans-serif;
+        font-size:14px; letter-spacing:.01em;
+      ">
+        <div style="
+          width:22px; height:22px; border-radius:50%;
+          border:2px solid currentColor; border-top-color:transparent;
+          animation:boot-spin .8s linear infinite; opacity:.55;
+        "></div>
+        <div>Waking this company&hellip;</div>
+        <div id="boot-hint" style="font-size:12px; opacity:.6;">Idle companies start on demand, which takes a few seconds.</div>
+        <div id="boot-slow" style="font-size:12px; opacity:0; max-width:34ch; text-align:center;">
+          This is taking longer than a normal wake. If it does not finish, the
+          company may have failed to start rather than be starting.
+        </div>
+      </div>
+    </div>
+    <style>
+      @keyframes boot-spin { to { transform: rotate(360deg) } }
+      /* After a wake budget's worth of waiting, stop claiming this is normal.
+         Done in CSS rather than script so the placeholder stays inert: a
+         message that says "waking" forever would be a worse lie than the blank
+         page it replaces. 35s is comfortably past both the measured 8-13s wake
+         and the manager's 30s startup timeout. */
+      @keyframes boot-reveal { to { opacity:.75 } }
+      @keyframes boot-fade   { to { opacity:0 } }
+      #boot-slow { animation: boot-reveal .4s ease 35s forwards }
+      #boot-hint { animation: boot-fade .4s ease 35s forwards }
+      /* Reduced motion: keep the indicator, drop the spin. */
+      @media (prefers-reduced-motion: reduce) {
+        #boot [style*="boot-spin"] { animation: none !important }
+      }
+      @media (prefers-color-scheme: dark) {
+        #boot { background:#08090B !important; color:#8A8A9A !important }
+      }
+    </style>
     <script type="module" src="/src/main.tsx"></script>
   </body>
 </html>
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -16,6 +16,33 @@
         };
       </script>
     -->
+    <style>
+      /* Styles for the boot placeholder in <body>. In <head> so they are
+         guaranteed parsed before the markup they target. Only keyframes and
+         state live here — the first paint is carried by inline styles, so the
+         placeholder still renders if this block is ever dropped. */
+      @keyframes boot-spin { to { transform: rotate(360deg) } }
+      /* After a wake budget's worth of waiting, stop claiming this is normal.
+         Done in CSS rather than script so the placeholder stays inert: a
+         message that says "waking" forever would be a worse lie than the blank
+         page it replaces. 35s is comfortably past both the measured 8-13s wake
+         and the manager's 30s startup timeout. */
+      @keyframes boot-reveal { to { opacity:1 } }
+      @keyframes boot-fade   { to { opacity:0 } }
+      #boot-slow { animation: boot-reveal .4s ease 35s forwards }
+      #boot-hint { animation: boot-fade .4s ease 35s forwards }
+      @media (prefers-reduced-motion: reduce) {
+        /* The spinner is decorative, so it stops entirely. The status swap is
+           not — it carries the "may have failed" message — so it keeps its 35s
+           delay and loses only the fade. `animation: none` here would suppress
+           the reveal and leave the misleading copy up forever. */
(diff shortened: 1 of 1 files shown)


oxoxDev (changes requested):
The craft here is good — inert markup with no script or stylesheet link so it paints on the first byte, the 35s de-escalation done in pure CSS, reduced-motion handling, and the reasoning written down rather than assumed. I also checked the colour claim and it's exact: `#F7F7FC` is `--surface-light-bg` and `#08090B` is `--surface-dark-bg` (`index.css:84,91`). That's the kind of detail that usually drifts, and it didn't.

But I think this is aimed one layer below the problem, and the PR's own timeline is the evidence.

**Blocking — the placeholder ships inside the document that is being held, so it can't paint during the window it targets.**

Your timeline says it outright:

```
t+0     blank white     index.html still in flight
```

For a hosted tenant, `index.html` is served by the tenant pod itself, not by an always-up edge. The tenant image bakes the console bundle in and points the binary at it — `Dockerfile:103` `COPY --from=console /console/dist /app/console`, `Dockerfile:114` `ENV OPENCOMPANY_CONSOLE_DIR=/app/console` — and `src/server/routes.rs` serves it through `ServeDir` with the SPA fallback answering `index.html`. On the manager side the tenant is a single container on one port (`kube_manifests.rs:618`), with no separate console container and no `CONSOLE_DIR` override.

So when a pod is scaled to zero, the request for `/` is held for the full 8–13s. The placeholder is *in* that response. It arrives when the pod is already up, and the window it actually covers is the bundle fetch plus React mount — real, but a fraction of a second, not the 8–13s in the issue. A user opening a cold `marketing.staging.opencompany.work` still sees blank white for 8–13s, then this for a moment.

I'd rather flag this than have it merge and #1241 stay open with a fix attached to it.

Two places the fix could actually live:

- **At the holding layer.** The manager already knows it is holding a request for a waking pod — that is the only component that can send bytes during the window. A minimal `text/html` holding page from the proxy (or a `503` + `Retry-After` with a body, which the API layer already does per #19 phase 5) would cover the real gap. This is the same "make waking legible" work, just at the layer that has the information.
- **Serve the console from an always-up layer.** `frontend/nginx/default.conf.template` already describes exactly this shape — static console from `/usr/share/nginx/html`, only `/api/`, `/healthz`, `/spec`, `/tiny` proxied to `${OC_UPSTREAM}`. If tenant consoles were served that way, `index.html` would be immediate and this placeholder would work as designed, doing precisely the job it is written for.

Worth confirming which path staging actually uses before acting — I verified the tenant-image path because that is what the manifests and Dockerfile point to, but if tenant traffic goes through that nginx image instead, then the premise flips and this lands as intended. That's a question you can answer faster than I can.

**Non-blocking — the theme can still flash for anyone who overrode it.**

`main.tsx:26` is `<ThemeProvider attribute="class" defaultTheme="system" enableSystem>`, so the default really does follow `prefers-color-scheme` and the placeholder matches. But `next-themes` persists an explicit choice in `localStorage`, so a user on a light system who picked dark gets a white placeholder handing over to a dark shell — the exact flash the comment says it avoids. Narrow, and only for users who overrode, but the comment currently claims more than it can deliver. Reading the stored key inline would need a script, which would break the "deliberately inert" property — so I'd just soften the comment to say it follows the *system* preference rather than the app's theme.

**Nit** — `#boot [style*="boot-spin"]` matches on a substring of the `style` attribute. It works, but it couples the reduced-motion rule to the exact spelling of an inline style, so reformatting that attribute silently disables it. A class or an id on the spinner div would be sturdier for two more characters.

**Nit** — the `<style>` block sits after the markup in `<body>`. Harmless in practice since the inline styles carry the first paint and the `<style>` only supplies keyframes, but in `<head>` it would be guaranteed parsed before the element it targets.

One thing I'd keep exactly as-is regardless of where this lands: the 35s de-escalation. "A message that says waking forever would be a worse lie than the blank page it replaces" is the right instinct, and doing it in CSS to preserve inertness is the right execution.

CI was still running when I looked (`Rust`, `Desktop`, `Gated host binary` pending, CodeRabbit in progress). For a one-file HTML change none of the Rust lanes are meaningful, and this repo has had a lot of infra-cancelled jobs in the last two days — worth reading any red on those with suspicion rather than as signal.

CodeGhost21:
All three findings verified against the code and fixed. The first one is right in
a way that changes what this PR is, so leading with it.

**This does not fix the reported cold start, and I claimed it did.**

Measured on a deliberately parked tenant:

```
TTFB = 8.174s   total = 8.174s   http=200
```

The first byte of HTML does not arrive for 8 seconds, because the platform holds
the request while the pod boots. This placeholder lives *inside* that HTML — so
during the exact window the reporter saw a white page, it does not exist yet.

What it actually covers is the smaller window after the HTML arrives and before
React mounts. That is a genuine improvement, and it is not the reported problem.
I have retitled the PR, removed `Closes #1241`, and put the limit in the markup's
own comment so nobody assumes otherwise later.

Covering the cold start needs a holding response from something that is always up
— the edge nginx or the manager's proxy — which is not the console. #1241 stays
open for that.

**Contrast failed AA.** `opacity:.6` blended `#6B6B7B` to `#A3A3AF`, giving
**2.34:1** against a required 4.5. Both base colours already pass at full opacity
(4.90:1 light, 5.86:1 dark), so the opacity is removed rather than the colours
retuned.

**Reduced motion suppressed the wrong element.** The rule only matched the
spinner. Worth noting `animation: none` on the status swap would have been
*worse* than leaving it alone — the 35s reveal would never fire and the
misleading "waking" copy would stay up indefinitely. It now keeps the delay and
drops only the fade; the spinner, being decorative, still stops entirely.

`npm run build` passes.

CodeGhost21:
Thanks — and you're right on the blocking point. I'd reached the same conclusion independently just before your review landed (measured `TTFB=8.174s http=200` on a deliberately parked tenant), and had already retitled this PR and dropped `Closes #1241`. Your reasoning gets there from the manifests rather than the wire, which is the better derivation.

**Answering the question you left open — staging uses the tenant-image path, so your premise holds.**

Three independent confirmations:

- `Dockerfile:52` says it in as many words: *"the runtime image serves at `/` (same-origin, no separate nginx container)."*
- The console is baked into the tenant image and served by the binary: `Dockerfile:103` `COPY --from=console /console/dist /app/console`, `Dockerfile:114` `ENV OPENCOMPANY_CONSOLE_DIR=/app/console`.
- The pod really is one container — `kube_manifests.rs` builds `containers: vec![container]`, no console sidecar, no `CONSOLE_DIR` override.

`frontend/Dockerfile`'s nginx runtime stage is a **separate image** for Compose/ECS (`ENV OC_UPSTREAM=opencompany:8080`), not what hosted tenants run. So the premise doesn't flip: `index.html` is held for the full wake, and this placeholder cannot paint during it.

That leaves your second suggested shape — a holding response from the layer that *knows* it is holding — as the real fix. #1241 stays open for it.

**Fixed in `72c8d2ce`:**

- **Theme claim.** Verified and you're right: `main.tsx:26` is `defaultTheme="system" enableSystem` with no custom `storageKey`, so an explicit override does persist to `localStorage` and only a script could read it. Rather than soften the wording alone, the comment now names the default it matches *and* states the override case as a known gap with the reason (keeping the placeholder inert). The overclaim was "the handover is not a flash of a different colour" — that sentence is gone.
- **Spinner hook.** `#boot [style*="boot-spin"]` → an id on the spinner. Your reasoning is the part that convinced me: it worked, but it made reformatting an attribute silently disable the rule, and nothing would have caught that.
- **`<style>` into `<head>`.** Moved, with a note that the first paint is still carried by the inline styles so the placeholder survives losing the block entirely.

`npm run build` passes and I checked the built `dist/index.html` rather than trusting the source: placeholder intact, `#boot-spinner { animation: none !important }` present, no `[style*=` selector left anywhere.

On CI — agreed, and thanks for flagging the cancelled-job noise. The Rust and Desktop lanes have since gone green on this one anyway, so there's nothing to read with suspicion.

CodeGhost21:
Correction to my previous comment: when I wrote that the Rust and Desktop lanes "have since gone green," I was reading the run for the *previous* commit — the run for `72c8d2ce` was still in progress and I hadn't observed it. I shouldn't have stated it.

The accurate result, now that it has finished:

```
Changed areas ........................ SUCCESS
Actions pinned ....................... SUCCESS
Rust ................................. SUCCESS
Rust (mongodb) ....................... SKIPPED
Rust (openhuman, tinycortex) ......... SKIPPED
Gated host binary .................... SUCCESS
Console .............................. SUCCESS
Console (current Node, advisory) ..... SUCCESS
Desktop .............................. SUCCESS
Console E2E .......................... SUCCESS
Console E2E (live brain) ............. SUCCESS
```

Green, and no cancelled jobs in this run — so the conclusion holds, but it holds because it finished, not because I checked.

oxoxDev (approved):
Approving. You took the finding the right way — the blocking issue couldn't be fixed in `index.html`, and rather than trying, you corrected the claim. The comment now states the limit plainly:

> It does NOT cover a cold start, and that limit is worth stating here so nobody assumes otherwise. […] measured TTFB on a parked tenant is 8.2s, and this markup lives inside the HTML that has not arrived yet. Fixing that needs a holding response from something that is always up — the edge, or the manager's proxy — not the console.

That 8.2s TTFB measurement is independent confirmation of the finding, which is better than taking my word for it. And documenting the boundary at the site of the code is more durable than a review thread — the next person to look at this file learns why it doesn't cover the wake, instead of assuming it does and closing #1241 twice.

**One thing before merge: the body still opens `Closes #1241.`** As written, merging auto-closes an issue titled *"the console shows a blank white page while a scaled-to-zero tenant wakes"* — the exact behaviour the code comment now says this does not fix. `Part of #1241` or `Refs #1241` keeps it open for the holding-response work. Worth reconciling the body's framing with the comment too; it still leads with the 8–13s cold start as the problem being solved.

Body-only, so no re-review needed from me — just don't let it merge with the `Closes`.

**You caught something in my own suggestion that I missed.** I asked for a sturdier hook than `[style*="boot-spin"]`, and the obvious `animation: none` under reduced motion would have suppressed the 35s reveal along with the spin — leaving the misleading "waking" copy up forever for exactly the users who opted out of motion. Splitting it is right:

```css
#boot-spinner { animation: none !important }
#boot-slow, #boot-hint { animation-duration: 0s !important }
```

Decorative animation stops; the status swap keeps its 35s delay and loses only the fade. That's a real bug avoided, and the comment explaining why is the part worth keeping.

The rest checks out: `id="boot-spinner"` replaces the attribute-substring match, the `<style>` moved to `<head>` with the note that inline styles still carry first paint if the block is ever dropped, and the theme comment now names the `localStorage` override case and calls it a deliberate trade rather than claiming no flash. The contrast fix on `#boot-hint` (dropping the `opacity:.6` and taking `boot-reveal` to full opacity) is yours, not something I flagged — `#6B6B7B` at 60% over `#F7F7FC` would not have passed.

All checks green.

Issue #1106, opened 2026-08-19
tasks: no way to choose who does a task at create time, and nothing disambiguates when several teammates plausibly fit


oxoxDev:
## Summary

Creating a task offers no choice of who does it, and when the operator does not choose, the host picks silently — even when two teammates plausibly fit the same work.

The example that prompted this: a task to *fetch trending tweets about "agent harness" and similar keywords*. The company had both a **DevRel agent** and a **Social Media Manager**. Either could reasonably own it. One got it. The operator was never asked, and never told there had been a choice to make.

Two things are wanted:

1. **Pick the assignee when defining the task.**
2. **When no assignee is given and several teammates fit, disambiguate** rather than quietly picking one.

## Background check — what exists today

**Creating a card offers no assignee.** The board's Add-task dialog (`frontend/src/views/TasksView.tsx`) posts only `{ title, note, deliverable? }`. The card is created with `assignee: ""` — nobody.

**The wire already supports one.** `POST …/tasks` takes an optional `assignee` (`src/server/ops/tasks.rs`), and `runtime::assignee::resolve` canonicalises it against the roster — a desk resolves to its lead, a teammate by id or display name.

**The picker already exists, just not at creation.** `frontend/src/views/AssigneeSelect.tsx` is wired into `TaskEditDialog` and `TaskDetailView`. So an operator *can* set an owner — only after the card exists, by opening it and editing.

**When nobody sets one, the model decides.** An unassigned card is routed by an orchestrator/planning turn through the `assign_task` delegation tool; the planning prompt is handed `Currently assigned to: nobody` and picks. That is the same class of behaviour #982 described from the chat side ("a card opened from a DM is assigned by content match, not to the teammate addressed", closed by #997).

**Ambiguity is already refused in one narrow case, and not in this one.** `assignee::resolve` returns a refusal when two teammates share a display name (`a_display_name_cannot_shadow_a_manifest_id`, `two_teammates_sharing_a_display_name_are_refused_not_guessed`) — the host will not guess between identical *names*. But two teammates with distinct names and **overlapping capability** are not ambiguous to that resolver at all, so nothing stops the model choosing one.

So the gap is real and the pieces are mostly there: a picker component, a wire field that accepts it, and an existing "refuse rather than guess" posture to extend.

## What to build

**1. Assignee at create time.** Add `AssigneeSelect` to the Add-task dialog, defaulting to unassigned so today's behaviour stays the default rather than forcing a choice on every quick card.

[...]

PR #1246, opened 2026-08-20
fix(harness): baseline teammate steps aside for a company-authored tie

### LINKED ISSUE #1196: planning: a baseline teammate is indistinguishable from one the company wrote, so #1106 asks about ties it need not raise
## Summary

The planner cannot tell a teammate the company wrote from one the global baseline gave it, so #1106's ownership question fires on ties it did not need to raise.

Every company carries four teammates it never authored — `operations`, `page_builder`, `researcher`, `writer` (`globals/agents/`). They are real and assignable, and that is fine. What is not fine is that `gather_evidence` renders them identically to the company's own roster:

```
- `devrel` — Developer Relations — may use: … — Talks to the developer audience: docs, demos, …
- `writer` — Writer — may use: … — …
```

Nothing in that list says one of those two was chosen for this company and the other arrives in every company. So when both plausibly fit, the model ties them, and since #1106 a tie parks the card and asks a person.

## Observed against a real model

Driven on `main` @ `d4646e51` with a live host and a real model (OpenRouter → `deepseek/deepseek-chat`), on a scratch company whose own roster is `devrel`, `social_manager`, `accountant`:

| card | candidates returned | outcome |
|---|---|---|
| "Post a short thread on X about our new agent harness release" | `devrel` + `social_manager` | parked — **correct**, both company-authored |
| "Reconcile last month's supplier invoices and file the VAT return" | `accountant` | dispatched — **correct**, no padding |
| "Rewrite the API reference page for the new authentication endpoints" | `devrel` + **`writer`** | parked — the case this issue is about |

The third is not the model behaving badly. `devrel`'s description says it owns developer docs; the global `writer` writes things. On the evidence it was shown, tying them is defensible. The problem is the evidence.

Two of the three ties involved a global, on a company with only three roles of its own. A company with a fuller roster gets four baseline teammates competing against every one of them.

## Why the fix is not a precedence rule in the host

The obvious move — "a company teammate outranks a global" enforced in `resolve_assignee_candidates` — is capability ranking in Rust, which is the thing #1106 explicitly rules out:

> it should not become a fuzzy matcher — the workflow copilot's grounding deliberately has *no* fuzzy name matching, and capability-fit there is the model's job against honest descriptions while the host enforces reality.

Excluding globals outright would also be wrong. A card that is purely "write this prose" may genuinely belong to `writer`, and a company that never staffed a writer of its own is exactly the company the baseline exists for.

## Proposed shape

Keep the split (model proposes, host enforces reality) and fix the evidence instead.

1. **Mark provenance in the prompt.** [`Agent::global`](src/company/types.rs) already records it — "Provenance, not configuration: no author writes it, and the merge sets it on every teammate it appends." `TeammateBrief` drops it, so it never reaches the model. Carry it, and render it (`— from the shared baseline`, or similar).
2. **State a preference, not a rule.** Tell the planner that when a company-authored teammate and a baseline one both fit, prefer the company's own — and only raise the ownership question when the tie is between roles the company actually chose. A preference the model applies against honest descriptions, not a filter the host applies to a list.

That leaves the good ties intact (row 1 above is two company roles and should still park) and removes the ones where the company has already expressed a choice.

## Acceptance

- The roster in the planning prompt distinguishes baseline teammates from the company's own.
- A card that fits one company teammate and one baseline teammate dispatches to the company's, rather than parking.
- A card that fits two company teammates still parks with both — #1106 unchanged.
- No ranking logic added to `resolve_assignee_candidates`; it keeps validating names and nothing else.

## Scope note

This is a planner-evidence gap, not a #1106 regression: before #1106 the same blindness produced a single silent `proposedAssignee` from the same undifferentiated list. #1106 made it visible by turning ties into questions.

sanil-23:
## Summary

Closes #1196.

`gather_evidence` rendered a global/baseline teammate (`globals/agents/` —
`operations`, `page_builder`, `researcher`, `writer`) identically to a
teammate the company authored itself, so the planner could tie a company role
against the baseline padding every company carries. Since #1106, a tie parks
the card and asks a person — but a company that has already staffed a role
itself has already answered that question.

Per the issue thread's resolved shape (host-side precedence, not a prompt-only
nudge — see the discussion on #1196), this adds a `prefer_company_over_baseline`
pass, run right after `resolve_assignee_candidates` in `run_planning_pass`
(`src/harness/built_in/planning.rs`). It only acts when the candidate set mixes
a company-authored teammate with a baseline one: the baseline candidate(s) are
dropped, so the pass either collapses to a single proposal (dispatches, no
park) or, in the two-company or two-baseline cases, is untouched (still
parks — #1106 unchanged). `resolve_assignee_candidates` itself is untouched —
it still only validates names.

Also carries `Agent::global` into `TeammateBrief` and renders it in the
planner's roster prompt (`— from the shared baseline`), so the model has that
provenance as evidence too, even though the tie itself is resolved
host-side.

## API Or Behavior Changes

- A card whose plan ties a company-authored teammate against a global
  baseline teammate now dispatches to the company's own teammate instead of
  parking with an ownership question.
- A tie between two company teammates, or between two baseline teammates,
  behaves exactly as before (#1106: parks, both candidates recorded).
- The planning prompt's roster now marks baseline teammates with
  `— from the shared baseline`.

## Tests

- [x] `cargo clippy --lib --features openhuman -- -D warnings`
- [x] `cargo test --lib --features openhuman` (4657 passed, 0 failed)
- [ ] `cargo build --all-targets` (not run — `--lib` build/test/clippy covers the changed module; happy to run the full target set if CI flags a gap)

New/updated tests in `src/harness/built_in/planning/test.rs`:
- `a_company_teammate_beats_a_baseline_tie_and_dispatches_without_parking`
- `two_baseline_teammates_still_park_with_both`
- `prefer_company_over_baseline_drops_only_a_true_mixed_tie` (unit-level)

## Documentation

No docs reference this prompt behavior directly; none updated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved assignee selection when company-authored and baseline candidates are both eligible.
  * Company-authored candidates now take precedence in mixed ties, while ties among baseline candidates remain unresolved.
  * Cards with unresolved teammate or desk ties continue to be parked for review.
  * Added baseline provenance to planning context to improve assignment decisions.

### DIFF
--- a/src/harness/built_in/planning.rs
+++ b/src/harness/built_in/planning.rs
@@ -1884,29 +1884,48 @@ fn resolve_assignee_candidates(
 /// baseline teammates, or between two company teammates, carries no such
 /// signal and is left untouched — issue #1106's park-and-ask stands for both.
 ///
-/// A candidate id that does not resolve to a manifest agent (a desk, or an
-/// overlay teammate — [`OverlayAgent`](crate::ports::types::OverlayAgent) has
-/// no `global` field, so it can never be one) counts as company-side: only a
-/// manifest agent explicitly marked `global` is baseline.
+/// A candidate id resolves to exactly one of three provenances: a manifest
+/// agent marked `global` is [`Baseline`](Provenance::Baseline); a manifest
+/// agent that is not, or an overlay teammate — which
+/// [`OverlayAgent`](crate::ports::types::OverlayAgent) can never be, having no
+/// `global` field at all — is [`Company`](Provenance::Company); anything else
+/// `resolve_assignee_candidates` could still have handed back (a desk) is
+/// neither. A desk is not the company's own choice of *teammate*, so its mere
+/// presence must not stand in for a real one: it neither triggers the drop nor
+/// is dropped by it, on either side of the tie.
+enum Provenance {
+    Baseline,
+    Company,
+}
+
+fn provenance_of(evidence: &Evidence, id: &str) -> Option<Provenance> {
+    if let Some(agent) = evidence.record.manifest.agents.iter().find(|a| a.id == id) {
+        return Some(if agent.global {
+            Provenance::Baseline
+        } else {
+            Provenance::Company
+        });
+    }
+    if evidence.record.overlay_agents.iter().any(|a| a.id == id) {
+        return Some(Provenance::Company);
+    }
+    None
+}
+
 fn prefer_company_over_baseline(
     evidence: &Evidence,
     candidates: Vec<AssigneeCandidate>,
 ) -> Vec<AssigneeCandidate> {
-    let is_baseline = |id: &str| {
-        evidence
-            .record
-            .manifest
-            .agents
-            .iter()
-            .find(|a| a.id == id)
-            .is_some_and(|a| a.global)
-    };
-    let has_company_side = candidates.iter().any(|c| !is_baseline(&c.id));
-    let has_baseline = candidates.iter().any(|c| is_baseline(&c.id));
-    if has_company_side && has_baseline {
+    let has_company = candidates
+        .iter()
+        .any(|c| matches!(provenance_of(evidence, &c.id), Some(Provenance::Company)));
+    let has_baseline = candidates
+        .iter()
+        .any(|c| matches!(provenance_of(evidence, &c.id), Some(Provenance::Baseline)));
+    if has_company && has_baseline {
         candidates
             .into_iter()
-            .filter(|c| !is_baseline(&c.id))
+            .filter(|c| !matches!(provenance_of(evidence, &c.id), Some(Provenance::Baseline)))
             .collect()
     } else {
         candidates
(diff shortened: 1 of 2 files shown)


sanil-23:
Pushed a commit addressing the actionable review comment (direct test coverage that the roster prompt marks a baseline teammate with "— from the shared baseline" and does not mis-mark a company-authored one).

Re: the failing `Rust` / `Rust (mongodb)` / `Rust (openhuman, tinycortex)` / `Gated host binary` checks — this is not caused by this PR. `Cargo.lock` on `main` (unchanged by this branch) pins `openhuman = "0.63.11"`, but the `vendor/openhuman` submodule commit currently recorded in the tree (`ded703dd`) is itself version `0.63.9`. Any `--locked` build fails on that mismatch. Confirmed this is pre-existing: the identical base commit (`7eb2e083`) has both a passing and a failing `main`-branch CI run with the exact same "cannot update the lock file" error, and two reruns on this PR reproduce it identically. Not something I should paper over by editing `Cargo.lock` in an unrelated issue's PR — flagging for a maintainer to re-sync the submodule pin and lockfile on `main`.

sanil-23:
Pushed a fix for the CodeRabbit finding: \`prefer_company_over_baseline\` was classifying *any* non-baseline id (including a desk, which is neither a company teammate nor a baseline one) as "company-side" — so a tie like \`[desk, baseline-teammate]\` could wrongly drop the baseline candidate even though no genuine company-authored teammate was actually in the mix. Replaced the boolean check with a three-way classification (Baseline / Company / neither) so a desk is neutral on both sides. Added the requested \`[sam, studio]\`-style test plus verified the existing teammate-only cases are unchanged. 52/52 tests pass, clippy clean.

Issue #66, opened 2026-07-23
Add a company→operator attention signal (bidirectional chat via GET /events SSE)


oxoxDev:
## Problem
The company chat is one-directional. When something needs the operator's attention (a task finished, an approval is pending, an MCP call failed, an agent has an update) the console does not proactively surface it.

## Evidence
- Backend has the machinery, unexposed: `EventLog::subscribe()` (`src/ports/events.rs`) backed by all stores, and a rich `CompanyEvent` enum (`src/ports/types.rs`) already modelling `AgentReply`, `TaskDispatched`, `McpCallFailed`, `ApprovalResolved`, etc.
- No HTTP surface serves it: `GET /events` was never implemented (noted as a follow-up in `src/server/operator.rs`), GraphQL uses `EmptySubscription`.
- Frontend only polls status + approvals every 5s (`frontend/src/hooks/use-company.ts`) → a passive `pending_approvals` badge. Nothing pops, toasts, or injects agent-initiated messages into the chat.

## Proposed fix
- Implement `GET /events` (SSE) over `EventLog::subscribe()`.
- Console: surface attention items (task/approval/MCP events) and inject `AgentReply` events into the chat transcript for true company→operator messaging.

Related: #26

PR #63, opened 2026-07-23
feat(chat): surface scrubbed tool-call/thinking steps in the operator chat

oxoxDev:
## Summary

The operator chat returned only final text, so a failed MCP call looked like a vague ack and a memory-served answer was indistinguishable from a tool-backed one. This adds a scrubbed `steps[]` activity trace per turn: `Agent::set_on_progress` captures ToolCallStarted/Completed + ThinkingDelta during the turn; a new `harness/steps.rs` folds them into `TurnStep`s and a `StepTimeline` renders above each chat bubble (collapsed "N steps · M failed", auto-expand on error). **Zero steps is the tell** that an answer was memory-served. Also re-skins MCP-hardening's failure surface into the same `steps[]` renderer. Stacked on MCP-hardening + #55 — diff includes them until they merge.

> Note: Sibling of the channels-telegram PR — both append a field to `OutboundMessage` (no `Default`) across ~12 construction sites; whichever merges second rebases.

## API Or Behavior Changes

Additive `steps: Vec<TurnStep>` on `OutboundMessage` (serde default, skip-empty); `HarnessPool::run` now returns `TurnOutcome{reply, steps}`.

## Tests

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [ ] `cargo build --all-targets`
- [x] `cargo test` (517 passing; `cargo test --features openhuman` = 644)

**Security**: the wire carries NO raw arguments/output/call_id — whitelist-only enrichment, errors are `cause_plain`/sanitize-class; `planted_secret_never_reaches_serialized_steps` asserts a planted secret (in output, nested args, error text) never survives serialization. Frontend typecheck + build clean.

## Documentation

`docs/spec/runtime/ports.md` documents `steps`.

PR #1271, opened 2026-08-20
chore(vendor): bump openhuman to the tinymemory-38a34d2 pin

YellowSnnowmann:
## Summary

Bump `vendor/openhuman` ded703dd → 3911b7b8 (openhuman#5608's head: upstream main 92bab8df + the tinymemory pin bump).

What rides in, for the memory surfacing work this unblocks:

- **tinymemory dc3a725 → 38a34d2** — the vendored tinymemory jumps from a mid-arc Aug-17 commit (97 behind) to the completed tinymemory#18 arc. Concretely: the `tinymemory-conformance` crate exists on disk (the next PR wires it into a CI lane), and the hosted adapters gain the #63–#66 production fixes — dead Cognee default endpoint removed, Mem0 cloud search no longer 400s on a null threshold, API keys no longer echoed into error text, named transport-failure classes, incremental response-body cap. Today's vendored adapters have all of those bugs, which is what made this bump a prerequisite for enabling `tinymemory` in shipped builds.
- **tinycortex 8401346b → bfd2ea21** — openhuman#5605's pin ("cc-corroboration").
- Everything else openhuman merged between ded703dd and current main.

Two OpenCompany-side changes ride along, both mechanical:

- **New `[patch."https://github.com/tinyhumansai/tinymemory"]` table** in the root `Cargo.toml`. The new tinycortex pin's `tinycortex-api` takes `tinymemory-api` **by git** (tinycortex#149), and openhuman's own patch table unifying that onto its vendored path copy is ignored when openhuman is a path dependency — the WS4 replication rule already documented above `[patch.crates-io]`. Without the entry cargo resolves two `tinymemory-api` crates and `tinymemory-core` fails with E0308s where the engine trait meets the contract (hit exactly as the handoff's "two-copies hazard" predicts; found by the compiler, fixed with the one-line patch mirroring openhuman's own).
- **`Cargo.lock` regenerated** for the vendored crate-version moves (`openhuman` 0.63.11 → 0.63.13, new `tinymemory-sources`/`tinymemory-sync` path crates, toml/winnow transitive bumps).

No OpenCompany source changes.

## Stacking

Depends on [openhuman#5608](https://github.com/tinyhumansai/openhuman/pull/5608) (draft). This PR pins its head commit so the chain can be validated now; **re-pin to the merge commit once #5608 lands, before this merges.**

Follow-ups that stack on this: the memory CI lane (union build + vendored driver-conformance), then enabling `tinymemory` in the staging tenant feature set.

## Verification

- `scripts/ci/init-vendored-submodules.sh` clean at the new pin
- `cargo metadata --locked` resolves after the lock regen
- `cargo check --features openhuman,tinycortex` (the gated tenant shape)
- `cargo check --features acp,runner,tinymemory-embedded` (the memory-seam shape)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -431,7 +431,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
 dependencies = [
  "axum-core",
- "axum-macros",
  "base64 0.22.1",
  "bytes",
  "form_urlencoded",
@@ -479,17 +478,6 @@ dependencies = [
  "tower-layer",
  "tower-service",
  "tracing",
-]
-
-[[package]]
-name = "axum-macros"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7aa268c23bfbbd2c4363b9cd302a4f504fb2a9dfe7e3451d66f35dd392e20aca"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.119",
 ]
 
 [[package]]
@@ -3204,7 +3192,7 @@ dependencies = [
  "tokio",
  "tokio-rustls",
  "tokio-util",
- "toml",
+ "toml 1.1.4+spec-1.1.0",
  "toml_edit",
  "tower",
  "tower-http 0.7.0",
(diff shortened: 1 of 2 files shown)


YellowSnnowmann:
Superseded by #1248, which carries the same vendor/openhuman bump to upstream main (92bab8df) and the identical [patch."…/tinymemory"] table — plus the #1201 fix itself. The one piece this PR had beyond that (advancing the vendored tinymemory f8bd9af → 38a34d2 via openhuman#5608, the #63–#66 hosted-adapter fixes) now rides #1273 as a single commit stacked on #1248. Merge order: openhuman#5608 → #1248 → #1273.

PR #1278, opened 2026-08-20
ci(staging): publish the tenant image via ConfigMap instead of restarting the manager

CodeGhost21:
Completes `opencompany-microservice#38`. **This is the step that actually stops the restarts** — the manager and cluster halves are already merged and deployed.

## The problem this workflow was causing

This workflow set `OCM_TENANT_IMAGE` with `kubectl set env` on the **manager's** Deployment. An env var is part of the pod template, so every tenant build created a new ReplicaSet — and the manager's strategy is `Recreate` (SQLite: never two writers), so the old pod stopped **completely** before the new one started.

The manager proxies every hosted tenant. Each of those gaps took the entire fleet offline — not degraded, failing.

**On 2026-08-20 there were eleven manager ReplicaSets. Three were manager deploys; the other nine came from this workflow.** Each gap is a handful of seconds, nothing alerts on it, and nobody had noticed. It surfaced only because a load test halted mid-run and the cause turned out to be unrelated to load.

## The change

Publish the image through a ConfigMap the manager reads via a mounted file. Updating a ConfigMap does not touch the Deployment, so nothing rolls.

```sh
kubectl -n "${K8S_NAMESPACE}" create configmap "${TENANT_IMAGE_CM}" \
  --from-literal=image="${IMAGE}" --dry-run=client -o yaml | kubectl apply -f -
```

## Two things that are deliberate, not incidental

**The old `set env` path is kept, behind a capability check.** A manager predating `OCM_TENANT_IMAGE_FILE` ignores the ConfigMap entirely, so writing only the ConfigMap would leave new tenants silently provisioning on the *previous* image — a worse failure than the restart, because nothing would report it. The check reads the env var straight off the live Deployment, so this picks the right path per-cluster with **no ordering requirement between repos** and no coordination needed.

**The write is verified, not assumed.** The kubelet refreshes a mount lazily, typically within a minute, so the workflow polls the manager until it reads back the exact image (up to 200s) and fails the deploy if it never does.

That second one matters: this change exists to remove a failure nothing was watching for. Replacing it with a *different* silent failure — a ConfigMap write that never reaches the manager — would be a poor trade.

## Verification

- workflow YAML parses
- the embedded remote script passes `bash -n`
- **the capability-detection jsonpath was run against live staging**, where it correctly selects the ConfigMap branch and reads back the currently-deployed image:

```
detection jsonpath -> /etc/opencompany/tenant-image/image
BRANCH: ConfigMap path (correct)
manager reads -> ...opencompany-tenant:staging-1e080ab4...
```

## Rollout state

```
✅ opencompany-microservice#39   manager reads the file        (merged, deployed)
✅ k8s#37                        ConfigMap + mount live         (merged, applied)
⬜ this PR                       workflow stops patching        ← stops the restarts
```

The ConfigMap on staging is already seeded with the current live image, so merging this changes *what* gets written, not what tenants run.

## Honest limits

- The fleet default only affects tenants provisioned **after** it changes — unchanged from before, and the `refresh_all_tenants` path still exists to move existing ones.
- The kubelet's ~60s refresh means the value is eventually-consistent rather than immediate. Fine for image rollouts, and the verify loop covers it, but worth stating rather than discovering.
- This does not stop the manager restarting on **its own** deploys, which is correct and unavoidable.

## Related

`opencompany-microservice#38` also notes that nothing watches for manager unavailability at all. Even after this, a `Recreate` manager has gaps on its own deploys, and today nothing distinguishes "brief expected gap" from "down for ten minutes".

## Summary by CodeRabbit

* **New Features**
  * Improved staging deployments by supporting tenant image updates through a shared configuration setting.
  * Added verification that supported managers use the updated image without requiring a restart.
  * Preserved fallback deployment behavior for older manager versions.
  * Enhanced deployment reporting to show the active tenant image and identify fallback configuration usage.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -208,10 +208,24 @@ jobs:
           # alone (opencompany-microservice#38).
           #
           # Updating a ConfigMap does not touch the Deployment, so nothing rolls.
-          if kubectl -n "${K8S_NAMESPACE}" get deploy "${MANAGER_DEPLOY}" \
-               -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="OCM_TENANT_IMAGE_FILE")].value}' \
-             | grep -q .; then
+          # Resolved once and reused by the closing report, so the two can never
+          # disagree about which source is actually governing provisioning.
+          MOUNT_PATH=$(kubectl -n "${K8S_NAMESPACE}" get deploy "${MANAGER_DEPLOY}" \
+            -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="OCM_TENANT_IMAGE_FILE")].value}')
+          if [ -n "${MOUNT_PATH}" ]; then
             echo "Publishing OCM_TENANT_IMAGE=${IMAGE} via ConfigMap (no manager restart)"
+
+            # Captured BEFORE the write so a failed verification can put the
+            # cluster back as it was. `HAD_CM` distinguishes "no ConfigMap yet"
+            # from "ConfigMap with an empty value", which need different undos.
+            HAD_CM=no
+            PRIOR_IMAGE=
+            if kubectl -n "${K8S_NAMESPACE}" get configmap "${TENANT_IMAGE_CM}" >/dev/null 2>&1; then
+              HAD_CM=yes
+              PRIOR_IMAGE=$(kubectl -n "${K8S_NAMESPACE}" get configmap "${TENANT_IMAGE_CM}" \
+                -o jsonpath='{.data.image}' 2>/dev/null || true)
+            fi
+
             kubectl -n "${K8S_NAMESPACE}" create configmap "${TENANT_IMAGE_CM}" \
               --from-literal=image="${IMAGE}" --dry-run=client -o yaml \
             | kubectl apply -f -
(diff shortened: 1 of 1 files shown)


CodeGhost21:
Both findings verified against the code and fixed in `6ae68f53`. Both were right.

## 1. A failed verification left the new value in place

Confirmed. `exit 1` stopped the deploy but not the write. The kubelet refreshes the mount on its own schedule, so a deploy that **explicitly could not confirm the manager saw an image** would still, minutes later, govern what new tenants provision on.

That is exactly the silent miss the verification exists to prevent — arriving late instead of never, which is arguably worse, because the deploy is red and the cluster quietly disagrees.

The prior value is now captured *before* the write and restored on failure.

One detail worth calling out: `HAD_CM` distinguishes **"no ConfigMap yet"** from **"ConfigMap holding an empty value"**. They need different undos — delete versus restore — and conflating them would either leave an orphan on the first run or delete a resource this deploy did not create.

## 2. The report keyed on existence, not capability

Confirmed. A manager without `OCM_TENANT_IMAGE_FILE` would print a stale ConfigMap left by an earlier deploy *beside* the env var, with nothing to indicate which one a tenant would actually get.

It now reuses the same capability check that decided where the value was written, so the two cannot disagree, and emits exactly one authoritative source. That required hoisting the check above the branch as `MOUNT_PATH`, resolved once instead of twice — which is what makes a single source possible at all.

## Verification

- workflow YAML parses; embedded remote script passes `bash -n`
- `MOUNT_PATH` is defined before all four of its uses
- **both restore branches exercised against the live cluster** with a throwaway ConfigMap:

```
case A: absent -> created -> failure   => deleted        (correct)
case B: OLD -> overwritten NEW -> failure => back to OLD (correct)
real opencompany-tenant-image ConfigMap: untouched
```

I tested the undo path specifically because an undo that doesn't undo is worse than not having one — it would turn a visible failure into a silent partial apply.

Issue #1261, opened 2026-08-20
workflows: the canvas cannot zoom out past 0.5x, so fit-view crops the first/last node on almost every shipped workflow


graycyrus:
## What's happening

`<ReactFlow>` in `frontend/src/views/WorkflowsView.tsx:2741` is mounted with
`fitView` / `fitViewOptions={{ padding: 0.2 }}` and no `minZoom` override, so
it inherits React Flow's library default `minZoom={0.5}`. Any workflow whose
node span needs a scale below 0.5 to fit the pane cannot ever be fully shown:
`fitView` computes the ideal fit-to-content scale, then the library clamps it
to `minZoom`, so the graph is zoomed to 0.5 and simply overflows the pane —
its first and/or last node land off-screen, permanently, with no way to see
the whole thing at once.

This is not a one-time layout glitch — the **Zoom Out** control button in the
canvas's own `<Controls>` is `disabled` from the moment the page loads,
because the viewport is already pinned at the zoom floor. Clicking the
"fit view" control (`.react-flow__controls-fitview`) does nothing new: the
computed transform before and after the click is identical
(`translate(-110.5px, 375px) scale(0.5)` → `translate(-122.5px, 369.5px)
scale(0.5)`), still at the floor, still cropping the graph.

## Reproduction (twice, two workflows, two companies)

Built this worktree at `main`'s current tip (`8eac7720`) with `cargo build
--bin opencompany --features openhuman,tinycortex,mcp`, ran it locally with a
seeded mock inference provider, and opened a workflow detail canvas at
1440×900 in Chromium (Playwright):

1. `agentic_software_company` → **Feature pipeline**: 10 nodes, `COL_GAP`
   apart along a single row (`request` at x=0 through `ship` at x=2700, all
   `y=0`), each 190px wide. Fitting all 10 into the ~1224px-wide pane at
   `padding: 0.2` needs roughly `1224 / (2890 * 1.4) ≈ 0.30` zoom. The
   viewport instead sits at exactly `scale(0.5)` before and after clicking
   "fit view", and the **Zoom Out** button is disabled throughout. On screen,
   only 8 of the 10 nodes are visible — the `request` trigger node and the
   `ship` output node are both cut off past the pane's edges. Panning is the
   only way to reach either one; there is no way to see the whole pipeline at
   once.
2. `agentic_research_lab` → **Research loop**: 9 nodes, same shape (single
   row, `y=0` for every node bar the layout's per-layer counter). Same
[...]

Issue #1230, opened 2026-08-20
workflows: the canvas minimap is an empty box in every graph — node dimensions never reach it


graycyrus:
## Symptom

Every workflow canvas carries a 200×150 minimap in its bottom-right corner. It is **always empty** — no node shapes, in any workflow, in either theme. It is `pannable zoomable`, so it looks and behaves like a navigation aid, and it shows nothing to navigate by.

`#/workflows/customer-escalation`, a four-node graph, 1440px. The whole minimap SVG:

```html
<div class="react-flow__panel react-flow__minimap ..." data-testid="rf__minimap">
  <svg width="200" height="150" viewBox="-141.83 -488.18 1374.66 1047.36" class="react-flow__minimap-svg">
    <title>Mini Map</title>
    <path class="react-flow__minimap-mask" d="M-174.56,-520.91h1440.12v1112.82h-1440.12
        M-109.10,-393.41h1309.2v857.83h-1309.2" fill-rule="evenodd"/>
  </svg>
</div>
```

One `<path>`: the viewport mask. Not a single `react-flow__minimap-node` rect. Note the `viewBox` **is** the node bounding box — so the minimap knows where the nodes are and still paints none of them.

The four nodes are measured and on screen:

```
start   180x56   translate(0px, 0px)
triage  180x71   translate(300px, 0px)
reply   204x71   translate(600px, 0px)
out     191x56   translate(900px, 0px)
```

Theming is fine — background is `rgb(255,255,255)` in light and `rgb(20,20,20)` in dark, so this is not a contrast bug. It is an empty box in both.

## Cause

`frontend/src/views/WorkflowsView.tsx:2725` renders `<MiniMap pannable zoomable />`, and `frontend/src/views/workflows/graph.ts:422` builds the node array with **no dimensions**:

```ts
const nodes: Node<WorkflowNodeData>[] = graph.nodes.map((n) => {
  ...
  return { id: n.id, type: "oc", position: { x: layer * COL_GAP, y: row * ROW_GAP }, data: { ... } };
});
```

React Flow's `MiniMapNodes` reads the **user's** node object, not the measured internals — `@xyflow/react@12.11.2`:

```js
const userNode = node.internals.userNode;
const { width, height } = getNodeDimensions(userNode);
...
if (!node || node.hidden || !nodeHasDimensions(node)) {
  return null;                        // <- every node, every render
}
```

and `@xyflow/system`:

```js
function nodeHasDimensions(node) {
[...]

PR #1292, opened 2026-08-20
fix(workflows): stop the minimap squashing linear pipelines into a sliver

### LINKED ISSUE #1259: workflows: the minimap now draws node rects (#1230), but every linear graph squashes them into an invisible sliver
## What's happening

Issue #1230 fixed the canvas minimap drawing *nothing at all* (nodes had no
`measured`/`width`/`height`, so `MiniMapNodes` skipped every one). That fix
landed in `frontend/src/views/workflows/graph.ts` (commit 9e8fb0a0, PR #1235)
by stamping `initialWidth: 190` / `initialHeight: 64` on every node.

The minimap's `<rect>` elements now genuinely exist with correct width/height —
verified in the DOM. But the minimap is still visually indistinguishable from
empty, for a different reason: `layout()` places every node's *depth* on the
x-axis and *sibling index* on the y-axis (`position: { x: layer * COL_GAP, y:
row * ROW_GAP }`), and every shipped company workflow I opened lays out with
exactly one node per depth layer — i.e. every node sits at `y: 0`. The real
content bounding box is therefore ~2800px wide by 64px tall (a 45:1 aspect
ratio).

React Flow's default `<MiniMap>` fits that bounding box into its 200×150 SVG
by padding the *height* out to match the container's aspect ratio, rather than
letterboxing. The resulting `viewBox` ends up e.g. `-66 -1024 2772 2112` — a
2112-unit-tall viewBox holding 64 units of actual node rects. On screen, each
node rect renders as a ~3px sliver, in React Flow's default light-gray fill
(`rgb(226, 226, 226)`) on a near-white background. The net result: the
minimap looks exactly as empty as it did before #1230, just for an aspect-
ratio/contrast reason instead of a missing-dimensions reason.

## Reproduction (twice, two different workflows)

Built this worktree at `8eac7720` (current `main`, already includes 9e8fb0a0)
with `cargo build --bin opencompany --features openhuman,tinycortex,mcp`, ran
it locally, opened:

1. `agentic_software_company` → **Feature pipeline** (10 nodes, purely linear)
2. `agentic_research_lab` → **Research loop** (9 nodes, includes a `gate`
   condition branch — still zero vertical spread, because `rethink` and
   `critique` land on different depths rather than becoming true siblings)

For both, `document.querySelectorAll('.react-flow__minimap-node')` returns
one `<rect>` per node with real `width`/`height` (e.g. `width="190"
height="64"`), all at `y="0"`, while the minimap `<svg>`'s `viewBox` reports a
height 30-40x the content's actual height. A cropped screenshot of the
minimap alone shows only the pannable-viewport dashed outline — no node
markers are perceptible.

## Why this matters

Every workflow shipped in `companies/*/workflows/*.toml` I checked is a
sequential agent hand-off with no two nodes at the same depth, so this isn't
an edge case — it's the shape the minimap will see on essentially every
company's every workflow. The feature reads as still-broken to an operator,
just via a subtler mechanism than #1230's.

## Where

`frontend/src/views/WorkflowsView.tsx:2757` — `<MiniMap pannable zoomable
className="!hidden sm:!block" />` with no `nodeColor`/aspect override.
`frontend/src/views/workflows/graph.ts`'s `layout()` is what produces the
zero-height bounding box that trips this up.

Not proposing a fix — just the finding, per the sweep's scope.

graycyrus:
## Summary

Closes #1259 — a follow-on to #1230, which made the workflow canvas minimap draw real node `<rect>`s (they existed in the DOM but were invisible). This PR fixes the actual visibility: every linear/low-vertical-spread graph (most shipped company workflow templates — 8-10 nodes in a straight horizontal line) still rendered as an imperceptible sliver.

- **Root cause 1**: React Flow's built-in `<MiniMap>` "contain"-fits its content bounding box into its 200x150 default container via `viewScale = Math.max(scaledWidth, scaledHeight)`, not configurable via any prop. For a graph whose real content is ~2800px wide x 64px tall (44:1), that pads the viewBox's height ~30-40x past the actual content.
- **Root cause 2 (found while verifying the first fix in a real browser)**: shrinking the built-in `<MiniMap>`'s own `style` container measured as *no improvement*. Its bounding box is the union of node bounds **and the current pan/zoom viewport** — and for a wide graph, the canvas itself ends up zoomed way out to fit that width, leaving a viewport rect far taller than any node. That viewport height dominates the bounding box regardless of container size, so shrinking the container just flips which axis inflates.

**Fix**: `WorkflowMiniMap` (`frontend/src/views/workflows/WorkflowMiniMap.tsx`) replaces the built-in `<MiniMap>`. It draws the same node rects (`@xyflow/react`'s public `MiniMapNode`) at their real layout positions, but scales strictly off a new `minimapViewBox`/`contentBounds`/`minimapDimensions` trio in `graph.ts` — node positions only, never the live viewport — so a node's real height reliably fills a healthy fraction of the minimap. Drag-to-pan and wheel-to-zoom are reimplemented on the public `useReactFlow` API (`setCenter`/`zoomIn`/`zoomOut`) rather than the built-in component's internal `@xyflow/system` wiring, since that package is only a transitive dependency here, not a direct one.

Verified live against both of the issue's own repro cases:
- `agentic_software_company` → Feature pipeline (10 nodes): rects went from ~3x1 CSS px (imperceptible) to ~10x3.4 CSS px (clearly distinguishable), in both light and dark mode.
- `agentic_research_lab` → Research loop (9 nodes, includes a branch): same result.

Merged `upstream/main` before opening — it now includes #1261 (`minZoom={0.1}` on the canvas, so `fitView` isn't clamped to the library's 0.5 default and can show the full pipeline). That's compatible by design: this minimap's scale never reads the live viewport/zoom, so it's unaffected either way — confirmed by re-verifying live post-merge (identical minimap geometry, canvas now correctly zoomed to 0.35 instead of clamped at 0.5).

## Test plan

- [x] `npm run typecheck`
- [x] `npm run typecheck:e2e`
- [x] `npx vitest run` — 157 files / 1482 tests pass
- [x] New tests in `frontend/test/unit/workflow-minimap-aspect.test.ts` covering `minimapDimensions` and `minimapViewBox`: linear chain shrinks/keeps a healthy node-height fraction, a graph with real vertical spread (siblings) keeps the untouched 200x150 default, empty/single-node graphs stay finite and sane.
- [x] Manual verification in a real browser (`cargo build --bin opencompany --features openhuman,tinycortex,mcp` + `npm run build`, served via `opencompany serve`): confirmed visible, distinct minimap node rects on the Feature pipeline and Research loop templates, in both color modes; confirmed drag-to-pan and wheel-to-zoom on the new minimap actually move the canvas viewport.


graycyrus:
@coderabbitai review

PR #1301, opened 2026-08-20
lint: Rust 1.98's result_large_err turned main red — crate-wide allow with reasoning

YellowSnnowmann:
CI runs unpinned stable; clippy 1.98 (landed in CI within the last hours — 1290's run at 17:40Z shows `clippy 1.98.0` where earlier runs used 1.97.1) promoted `result_large_err` into the default set, and it fires on **~80 pre-existing handler sites** (`operator.rs` ×19, `users/` ×24, …): every `Result<_, ApiError>` return, because `ApiError` wraps `OpenCompanyError` by value. Every branch — including main on its next run — is red with zero code change.

This is the minimal unblock: a crate-level allow with the reasoning at the attribute (cold error path, one construction per failed request, immediately serialized; boxing ~100 signatures costs an allocation per error and buys latency nothing). The alternative — pinning the CI toolchain — trades this class of surprise for staleness and is a bigger policy call than one lint.

Verified locally on rustc/clippy 1.98.0: `cargo clippy --locked --all-targets -- -D warnings` goes from 80 errors to 0 with only this attribute.

Flagging directly since every open PR is blocked on it: @oxoxDev

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,17 @@
 //! TinyHumans Rust modules. The default build stays small; enable the `tiny`
 //! feature to compile against the sibling `tiny*` crates.
 
+// Rust 1.98's clippy promoted `result_large_err` into the default set and it
+// fires on ~80 existing Axum handlers: every `Result<_, ApiError>` return,
+// because `ApiError` wraps `OpenCompanyError` by value (>128 bytes). CI runs
+// unpinned stable, so this turned every branch red overnight with no code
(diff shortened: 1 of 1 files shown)


YellowSnnowmann:
Superseded by #1302, which landed the better version of both halves: the toolchain pin WITH a drift guard (the policy call my body punted on), and the lint deferral scoped to src/server/mod.rs where all 80 findings live, rather than crate-wide. Closing; my open PRs drop their interim cherry-picks in favor of main's fix.

PR #1360, opened 2026-08-20
fix: make the sticky dialog footer opaque (#1359)

### LINKED ISSUE #1359: The sticky dialog footer is translucent, so scrolled content reads through Cancel and the submit button
Found during a design audit of the **Company** surface (`#/company`, `#/company/desks`, `#/team/<id>`).

## What happens

`DialogContent` is `max-h-[85vh] overflow-y-auto` and `DialogFooter` is `sticky bottom-[-1rem] … bg-muted/50`. On any dialog whose body outgrows the popup, the body keeps scrolling *underneath* the footer — which is the intended behaviour — but at 50% opacity the content is still legible through it.

The clearest instance is **New desk** (`DeskCreateDialog`) on `#/company/desks`. It renders one bordered row per roster teammate; the seeded `agentic-software-company` has thirteen, so the list scrolls, and the last two rows print across `Cancel` and `Create desk`:

- light: "Page Builder" and "Researcher" are fully readable behind both buttons
- dark: same, and the row borders read as a boxed artifact around the footer

It reads as a rendering fault, and it makes the two controls the dialog exists for harder to read than the list behind them.

## Expected

A sticky bar is opaque. `bg-muted` (not `bg-muted/50`) still differs from the popup's own `--popover` in both themes — lighter in light, darker in dark — so the tonal separation the translucency was reaching for survives.

## Scope

`frontend/src/components/ui/dialog.tsx` — the shared primitive, so every dialog with a scrolling body benefits.

Severity: **defect**, cosmetic but highly visible and on a primary create flow.

graycyrus:
Closes #1359.

## The failure

`DialogContent` is `max-h-[85vh] overflow-y-auto`; `DialogFooter` is `sticky bottom-[-1rem] … bg-muted/50`. On a dialog whose body outgrows the popup the body scrolls *behind* the footer — intended — but at 50% opacity it stays legible through it.

**New desk** on `#/company/desks` is the visible case: one bordered row per roster teammate, thirteen on the seeded company, so "Page Builder" and "Researcher" print across `Cancel` and `Create desk` in both themes.

## The change

One utility: `bg-muted/50` → `bg-muted`, with the reasoning recorded at the site. `--muted` still differs from the popup's `--popover` in both themes (lighter in light, darker in dark), so the tonal separation the translucency reached for is unchanged — only the bleed-through goes.

## Test

`frontend/test/unit/dialog-footer-opaque.test.ts` — a jsdom render asserting the mounted footer is `sticky`, carries `bg-muted`, and carries **no** `bg-*/nn` alpha modifier at all. Asserted on the class list the browser sees rather than on the source, so a later `className` override that reintroduces translucency is caught too.

## Commands run locally

- `npx vitest run test/unit/dialog-footer-opaque.test.ts` — 3 passed
- Browser-verified against a live seeded host, **New desk** dialog, light and dark at 1440px: rows now clip cleanly at the footer edge.

## Behaviour change

Purely visual, and it applies to every dialog with a scrolling body — that is the point.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated dialog footers to use a fully opaque background, improving visual consistency and readability.

* **Tests**
  * Added coverage to verify the footer remains sticky and does not apply unintended transparency.

### DIFF
--- a/frontend/src/components/ui/dialog.tsx
+++ b/frontend/src/components/ui/dialog.tsx
@@ -102,7 +102,18 @@ function DialogFooter({
     <div
       data-slot="dialog-footer"
       className={cn(
-        "sticky bottom-[-1rem] -mx-4 -mb-4 flex flex-col-reverse gap-2 rounded-b-xl border-t bg-muted/50 p-4 sm:flex-row sm:justify-end",
+        // `bg-muted`, not `bg-muted/50`. This footer is `sticky`, so on any
+        // dialog whose body scrolls — the desk creator's roster picker is the
+        // one with thirteen rows in it — the content underneath keeps running
+        // past the bottom of the popup. At 50% opacity that content reads
+        // *through* the footer: teammate names sit legibly behind Cancel and
+        // the submit button, which looks like a rendering fault and makes both
+        // harder to read. A sticky bar has to be opaque to do its job.
+        //
+        // `--muted` still differs from the popup's own `--popover` in both
+        // themes (lighter in light, darker in dark), so the tonal separation
+        // the translucency was reaching for survives the change.
+        "sticky bottom-[-1rem] -mx-4 -mb-4 flex flex-col-reverse gap-2 rounded-b-xl border-t bg-muted p-4 sm:flex-row sm:justify-end",
         className
       )}
       {...props}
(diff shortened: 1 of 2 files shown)


oxoxDev:
@coderabbitai full review

Issue #1316, opened 2026-08-20
Overview: the legend draws Workflow and Stage in the same colour


graycyrus:
## What

The Overview legend gives **Workflow** and **Stage** the same colour:

```
{ label: 'Workflow', color: CAT.workflow.color, ... }   // var(--brain-2)
{ label: 'Stage',    color: CAT.step.color,     ... }   // var(--brain-2)
```

`KnowledgeGraph.tsx:1503-1538`, `CAT` at `:48-57`. The two are separated only by a 12px lucide glyph (`Workflow` vs `Milestone`) rendered at `h-3 w-3`.

On the graph itself these are two different rings that sit adjacent — a workflow on ring 2 and its stages on ring 3 — so they are exactly the pair a reader most needs to tell apart, and the legend is where they go to learn how.

## Also on the legend

- Seven entries in one un-wrapping flex row (~615px). It clears the 300px detail rail down to about a 700px container, below which it is clipped by `overflow-hidden` rather than wrapping.
- **`Notes`** is the first entry and is the one label that does not name a node kind the way the others do — it is the company core's memory constellation, but the core itself is drawn black/white as the *company*, so the legend's pink `Notes` swatch matches nothing visible until you click the core.

Found in a design audit of `#/overview`; this is taste more than defect, but the colour collision is a real legibility cost.

Issue #1310, opened 2026-08-20
Overview: graph labels are drawn at 9px and 8.5px, under the design system's own 10px floor


graycyrus:
## What

`KnowledgeGraph.tsx:92-95`:

```ts
const labelFontPx = (kind: KGNodeKind): number =>
  kind === 'self' || kind === 'team' ? 10 : kind === 'task' ? 8.5 : 9;
```

Everything that is not the company core or a department is drawn at **9px**, and SOP tasks at **8.5px**. Agents, humans, tools, workflows and stages are all in the 9px band — that is most of the graph.

`index.css:607`:

> **Below 10px is not a size, it is a bug.** Nothing smaller is defined, and the audit in `docs/design-system/typography.md` lists the six sites currently under it that need raising.

`docs/design-system/typography.md:110`:

> ## Sizes below the scale
> **None remain.** Below 10px is not a size, it is a bug, and the console no longer has one.

Both of those are false while the landing page draws its labels at 8.5px and 9px. The graph escapes `scripts/ci/assert-design-tokens.sh` because it sets its size as an SVG `fontSize` in TypeScript, not as a `text-[8.5px]` class — the guard greps for the class form only.

## It is worse than nominal

`fixedLabel` (`KnowledgeGraph.tsx:194-200`) emits `calc(<px>px * var(--kg-cam-k))`, which cancels the *camera* zoom but not the **container** scale: the effective size is `fontPx x min(containerW/880, containerH/600)`. On a short viewport the labels render *smaller* than the number says. At a 1440x760 content area that is `760/600 = 1.27`, fine; at 1100x620 it is `~1.03`; a 1400x600 window puts task labels at 8.5px exactly.

`planLabels` explicitly refuses to shrink (`label-plan.ts:217-220`) — which is the right call — so the sizes are the only lever.

## Note before changing it

Raising the floor to 10px makes every label box wider, which makes the declutter pass drop more of them. The fix is probably not just three numbers; it likely needs the density work to go with it. That is why this is an issue rather than a one-line PR.

Found in a design audit of `#/overview`.

Issue #1309, opened 2026-08-20
Overview: the pillar selector offers three anonymous dots for three named desks


graycyrus:
## What

At rest the pillar control reads:

```
Pick a pillar
3 PILLARS          ● ● ●
```

Three 10px dots at 50% opacity, and no names. The company's three desks — *Engineering*, *Product & Design*, *Go-to-Market* — are **named right there on the graph**, in their own colours, three inches away. The control that exists to select one of them refuses to say which is which.

The name is only in each dot's `title` / `aria-label`, so it is a hover away on a pointer and unreachable on touch. Once you pick one, the control finally says the name (`Engineering — 1 / 3`). You have to click a blind dot to learn what it was.

`KnowledgeGraphFullscreen.tsx:83-124`.

## Also

- The text block is inert — not a button, not a menu, no cursor change. The only hit targets are the 24x20 dots.
- `"Pick a pillar"` is an imperative with no visible object.
- `3 PILLARS` is redundant with three visible dots, and becomes actively wrong at the empty state (`0 PILLARS`, no dots, and the instruction still says pick one).
- The paddles that do the same job say `aria-label="Previous department"` but `title="Previous pillar (←)"` — and the rest of the product calls these things **desks**. Three words for one concept across four surfaces.
- Nothing on screen discloses that `←` / `→` / `Escape` work.

## Suggested fix

Name the pillars. A row of colour-swatched name chips costs the same horizontal space at three desks as the dot row plus the caption does, and it answers "which" and "what for" at rest instead of on hover.

Found in a design audit of `#/overview`; the purpose of this control was not self-evident to me on arrival either.

Issue #1308, opened 2026-08-20
Overview: the knowledge graph has no way out — no node links to the surface it names


graycyrus:
## What

`#/overview` is the console's landing page. It draws every teammate, every open board card, every workflow, every stage, every tool and every desk the company declares — roughly 100 nodes on a seeded company — and **not one of them links anywhere.**

```
$ grep -rn "href\|onNavigate\|window.location\|navigate" frontend/src/views/overview/kg/*.tsx
(nothing — the only hits in the folder are a `hashStr` helper for animation seeding)
```

Click a teammate and you get a card naming their tools, their harness tier and their last run. There is no way from that card to `#/team/<agentId>`, which is the page that would let you do something about any of it. Click a task and you get its steps; `#/tasks/<id>` — the timeline, the plan brief, the steer controls — is a sidebar trip and a search away. Same for workflows, desks and notes.

## Why it matters

Every other view in this console is a place you *do* something. Overview is a place you *look at* something and then use the sidebar anyway. That is what makes it read as decoration rather than as a landing page: not the force layout, not the density — the fact that no path leads out of it.

It is also the cheapest thing about the page to fix. Every destination already exists and is already addressable:

| node kind | destination that already exists |
|---|---|
| AI teammate | `#/team/<agentId>` |
| SOP task | `#/tasks/<taskId>` |
| workflow | `#/workflows/<workflowId>` |
| department (desk) | `#/company/<deskId>` |
| human | `#/settings/people` |
| note | `#/memory` |

## Related

The card also tells operators to *"trigger it from /agents"* (`KnowledgeDetail.tsx:261`) — a route that does not exist in `VIEWS` at all.

Found in a design audit of `#/overview`.

PR #1368, opened 2026-08-20
fix(overview): keep the right-edge chrome clear of the detail rail

### LINKED ISSUE #1307: Overview: the detail card buries the next-pillar paddle and the Snapshot/Refresh control
## What

Opening any detail card on `#/overview` silently hides two controls the operator needs: the **Next pillar** paddle and the **Snapshot / Refresh** chip.

`KnowledgeGraphFullscreen.tsx:129-131` carries this comment:

> The right paddle steps aside when the detail panel is open

**No code implements that.** The paddle renders whenever `!coreOpen`; the detail aside is `z-30` at `right-0` with `w-[300px]`, and the paddle is `z-20` at `right-2` with `w-12` — entirely inside the aside's band.

The snapshot chip is worse, because it belongs to a different component and nobody was looking: `Overview.tsx:260` puts it at `absolute right-3 top-3 z-10`, and the aside resolves against the same containing block (its parent at `KnowledgeGraphFullscreen.tsx:73` has no `position`, so it lands on `.oc-kg.relative` in `Overview.tsx:247`). `z-30` wins. **The staleness signal, the Refresh button, and the "Could not reach the company" error alert are all buried behind any open detail card.**

## Measured, at 1440x900 against a seeded host

```
nextPaddle            [1384, 386, 48, 128]
aside                 [1140,   0, 300, 900]
refreshChip           [1356,  17,  63,  20]
elementFromPoint(paddle centre)  -> DIV.flex-1 overflow-y-auto px-3 py-2.5   (the aside's body)
elementFromPoint(refresh centre) -> BUTTON  (the aside's Back button)
```

Both are not merely obscured — they are **not clickable**, because the aside is opaque (`bg-os-bg/95`) and on top.

## Below 820px it is both paddles

The aside becomes a bottom sheet (`max-[820px]:inset-x-0 bottom-0 max-h-[62vh]`). Its top edge sits at 38vh; both paddles are centred at 50vh. The legend at `bottom-5 left-5` goes under it too.

## Why it matters

The graph's whole freshness contract is "this is a snapshot, and here is how you take another" (`Overview.tsx:85-93`). Clicking a node — the primary interaction on the page — revokes that contract with no notice. An operator inspecting a stale card has no way to see it is stale and no way to refresh.

## Suggested fix

Either shift the graph pane's right edge by the aside's width when it is open (the honest fix — the camera then frames into the space that is actually visible), or at minimum translate the right paddle and the snapshot column left by `300px` while the aside is open, and keep both above it.

Found in a design audit of `#/overview`; this one is a defect, not taste.

graycyrus:
Closes #1307.

## The bug

The graph's detail rail is a 300px `z-30` overlay. Two controls live on the right edge underneath it:

| control | was | rail |
|---|---|---|
| Next pillar paddle | `right-2 z-20` | `right-0 w-[300px] z-30` |
| Snapshot / Refresh / outage alert | `right-3 z-10` (owned by `Overview`) | same |

Measured at 1440x900 against a seeded host, with a teammate card open:

```
nextPaddle   [1384, 386, 48, 128]      aside  [1140, 0, 300, 900]
refreshChip  [1356,  17, 63,  20]
elementFromPoint(paddle centre)  -> the aside's scroll body
elementFromPoint(refresh centre) -> the aside's Back button
```

The rail is opaque (`bg-os-bg/95`), so both were **unclickable**, not merely obscured. Clicking a node — the primary interaction on this page — silently revoked the page's own freshness contract: an operator inspecting a card could not see that the snapshot was stale and could not refresh it. The total-outage alert from #1219 went with them.

`KnowledgeGraphFullscreen.tsx:129-131` already claimed the behaviour —

> The right paddle steps aside when the detail panel is open

— above a paddle nailed to `right-2` with nothing implementing it.

## The fix

The rail stays an overlay. Resizing the canvas under it reflowed the graph on every open and close, which is the glitch the `<aside>`'s own comment is about, so the controls move instead of the canvas.

- `clearOfRail` / `paddleTop` in the shell hold the offsets, so the rail-awareness is in one place rather than spread across two components.
- The snapshot line becomes a **slot** (`statusSlot`) the shell positions. `Overview` cannot see the rail; only the shell knows how much of the right edge is left. It also moves from a lone `right-3 top-3` inset onto the `top-5 / right-5` inset the pillar selector and legend already share.
- Both controls go to `z-40`, so a future change that moves the rail rather than the controls cannot re-bury them by stacking order.
- Below 820px the rail is a bottom sheet (`max-h-[62vh]`, anchored bottom), which buried *both* paddles at `top-1/2`. The offset reverts there and the paddles rise to `top-[19vh]` — the middle of the band the sheet leaves.
- The Suspense fallback keeps drawing the line at the shell's inset, so a cold load still says "Loading…" and it does not appear to pop into place when the physics chunk lands.

## Tests

`frontend/test/unit/overview-detail-rail-clearance.test.ts` — 4 cases. jsdom does no layout, so it asserts the class contract: at rest neither control carries the clearance; with a card open both do, and both are `z-40`; closing the card returns them. The point of the test is that this behaviour was previously a *sentence*, and a sentence is not a guard.

`overview-unreachable-host.test.ts`'s `KnowledgeGraph` stand-in now renders the slot it is handed — a mock that dropped it would take the surface under test with it and pass by drawing nothing.

## Verified

`npm run typecheck`, `typecheck:unit`, `typecheck:e2e`, `npx vitest run` (168 files / 1600 tests), `scripts/ci/assert-design-tokens.sh`.

In a browser against the seeded `agentic-software-company` host, at 1440x900 light **and** dark and at 1100x800, opening a teammate card:

```
light 1440  nextPaddle [1076,386,48,128] reachable=true   refresh [1052,25,63,20] reachable=true
dark  1440  nextPaddle [1076,386,48,128] reachable=true   refresh [1052,25,63,20] reachable=true
light 1100  nextPaddle [ 736,336,48,128] reachable=true   refresh [ 712,25,63,20] reachable=true
```

No console errors.

## Not in this PR

Found in the same audit, filed separately: the graph has no outbound navigation at all (#1308), the pillar selector shows anonymous dots (#1309), labels are drawn under the design system's 10px floor (#1310), and the legend is clipped by the rail at narrow widths (#1316).

---
Raised as a draft.

Design audit of `#/overview`; findings are professional judgement — no Figma reference exists for this surface.

### DIFF
--- a/frontend/src/views/Overview.tsx
+++ b/frontend/src/views/Overview.tsx
@@ -240,6 +240,59 @@ export function Overview({ client, company, companyName }: Props) {
 
   const memoryGraph = useMemo(() => buildMemoryGraph(sources.memories), [sources.memories]);
 
+  /**
+   * The snapshot line: what the page is, when it was taken, and the way to
+   * take another. The graph is not live, and says so rather than leaving an
+   * operator to assume a stale wheel is the current company.
+   *
+   * Handed to the graph as a slot rather than positioned here (issue #1307).
+   * This used to be an `absolute right-3 top-3 z-10` corner of its own, which
+   * put it squarely underneath the graph's `z-30` detail rail: opening any
+   * node's card hid the staleness signal, the Refresh control *and* the
+   * outage alert, and left them unclickable behind an opaque panel. Only the
+   * graph's shell knows how much of the right edge the rail is using, so the
+   * shell is what places this.
+   */
+  const statusSlot = (
+    <>
+      {/* Every source failed at once (issue #1219): a host that could not be
+          reached, said out loud, in the same corner that already owns the
+          staleness signal — rather than a graph redrawn empty with no
+          explanation. */}
+      {loadError && (
+        <div
+          role="alert"
+          className="max-w-64 rounded-md border border-destructive/40 bg-destructive/10 px-2 py-1 text-2xs text-destructive shadow-sm backdrop-blur"
+        >
+          {loadError}
+        </div>
+      )}
+      <div className="flex items-center gap-1.5 rounded-md border bg-background/90 px-2 py-1 text-2xs text-muted-foreground shadow-sm backdrop-blur">
+        <span className="truncate">
+          {sources.fetchedAt === null
+            ? loading
+              ? "Loading…"
+              : "No snapshot yet"
+            : `Snapshot ${new Date(sources.fetchedAt).toLocaleTimeString([], {
+                hour: "2-digit",
+                minute: "2-digit",
+              })}`}
+        </span>
+        <button
+          type="button"
+          onClick={refresh}
+          disabled={loading}
+          title="This page is a snapshot, not a live view. Re-read the company."
+          aria-label="Refresh the graph"
+          className="inline-flex items-center gap-1 rounded px-1 py-0.5 hover:bg-muted disabled:opacity-50"
+        >
+          <RefreshCw className={`size-3 ${loading ? "animate-spin" : ""}`} aria-hidden />
+          Refresh
+        </button>
+      </div>
+    </>
+  );
+
   return (
     // The whole viewport: the shell hides its top bar for this view, so there
     // is nothing above to subtract.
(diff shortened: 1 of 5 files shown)


oxoxDev:
@coderabbitai full review

oxoxDev:
@coderabbitai review

Issue #1141, opened 2026-08-19
console: lead the Company page with teammate cards, opening a full detail page


graycyrus:
The Company page is an org chart of desks. It should lead with the teammates themselves: a card each with avatar, name, description, status and workload, opening a full detail page.

## What already exists, and is not reachable

`#/team` already renders exactly this card grid — `grid gap-4 sm:grid-cols-2 lg:grid-cols-3` of `MemberCard` — and `#/team/<agentId>` already opens a **detail sub-page** (`TeamView.tsx:279-287`), deliberately a page rather than a modal so it can be linked. **But `team` is routable and absent from the nav** (`app-shell.tsx:172`), so no operator can get to it.

So most of this ask is surfacing and finishing what is there, not building it new.

## Proposed

**One Company page with a Cards ⇄ Org chart toggle**, matching the pattern Workflows already uses for Cards/List. Cards lead; the desk hierarchy stays reachable rather than being deleted.

```
Company                                        [ ▣ Cards │ ⌸ Org chart ]

┌────────────────────┐ ┌────────────────────┐ ┌────────────────────┐
│ ◕  Maya            │ │ ◕  Ravi            │ │ ◕  Priya           │
│    Research Lead   │ │    Analyst         │ │    Writer          │
│ Tracks competitor  │ │ Digs through the   │ │ Turns findings in… │
│ moves and drafts…  │ │ numbers and repo…  │ │                    │
│ ● idle · 3 tasks   │ │ ● working · 7 task │ │ ● idle · 0 tasks   │
└────────────────────┘ └────────────────────┘ └────────────────────┘
```

Click → `#/team/<agentId>`:

```
Company  ›  Maya                                          [ Edit ]
─────────────────────────────────────────────────────────────────
◕  Maya · Research Lead · on Research Desk
   Tracks competitor moves and drafts the weekly brief.

   Status ● idle      Tasks 3 open       Today $0.42 / $5.00 cap
   Tools  web_search, fetch_page, publish_artifact
   Desks  Research
```

## What is genuinely missing today

Cards show name, role and budget controls only — **description, status and task count are not there**. The detail page needs a **breadcrumb** and an **edit affordance**. Those are the real work; the grid and the routing are not.

## Sequencing

PR #1131 is editing `OrgChartView.tsx` right now to wire its rows to `#/team/<id>`. **Land that first**, then build this on top, so the two do not fight over the same file.

PR #1374, opened 2026-08-20
fix: pin a teammate card's status line to the bottom of the card (#1373)

### LINKED ISSUE #1373: Company: a teammate card's status line lands on a different baseline in every card of a row
Design audit of the **Company** surface. Judgement call, no Figma reference — but the mechanism is objective.

## What happens

`#/company` renders the roster in `grid gap-4 sm:grid-cols-2 lg:grid-cols-3`, so every card in a row is the same height. Inside, `CardContent` is `flex h-full flex-col gap-3` and everything is **top-aligned**. The description is `line-clamp-3`, so it is 1–3 lines depending on the teammate.

Result: `Working · 2 open tasks` — the one line an operator scans a roster *for* — sits wherever the description happened to end. On the seeded `agentic-software-company` (13 teammates) the status line lands on two or three different baselines in every row, and each card carries 30–60px of dead space beneath it.

Measured before the fix, row 1 at 1440px: two cards put the status at y=321, the one with a two-line description at y=349.

## Why it matters

"Who is working, and how much is on them" is the reason the workload line was added (#1141). A ragged baseline is exactly what stops a grid being scannable: the eye has to re-find the line per card instead of reading down a column.

## Expected

The running facts (workload + daily budget) are anchored to the bottom of the card. Every card in a row then puts them in the same place, and the empty tail goes away.

Severity: **defect** (layout), high visibility, whole-page.

graycyrus:
Closes #1373.

## The failure

`#/company` lays the roster out in a stretched grid, so every card in a row is the same height — but `CardContent` top-aligns everything and the description is `line-clamp-3`. The workload line therefore followed the description: on the seeded company, row 1 at 1440px put `Idle · 2 open tasks` at y=321 on two cards and y=349 on the one whose description wrapped to two lines, with 30–60px of empty card underneath each.

That line is the reason the workload was put on the card at all (#1141). A grid is scannable when a fact sits in the same place on every tile; this one moved.

## The change

The workload line and the daily-budget line move into one `mt-auto` wrapper, so the slack collects above them instead of below. Nothing else moves — the avatar, name, menu and description keep their position.

`empty:hidden` on the wrapper because both children are conditional: a host that cannot answer the board draws no workload (deliberately — `IDLE` vs `undefined`), and an uncapped teammate draws no budget line. Without it those cards would carry a `gap-3` for a wrapper with nothing in it.

## Test

`frontend/test/e2e/company-cards.spec.ts` — "the status line sits on one baseline across a row, whatever the description". It first asserts the fixture's two descriptions really do wrap to different heights (so it cannot pass vacuously), then asserts the three cards' status lines share a `y` **and** share the gap between the status line and the card's bottom edge — bottom-anchored, not accidentally-equal content.

Verified it fails on the parent commit and passes here.

## Commands run locally

- `npx vitest run` — 1596 passed / 167 files
- `PW_BASE_URL=… npx playwright test test/e2e/company-cards.spec.ts` against a served console — 12 passed
- Browser-verified against a live seeded host at 1440px and 1100px, light and dark: 13 cards, one distinct bottom gap (32px) across all of them, one baseline per grid row.

## Behaviour change

Visual only. No props, no data, no routes.

### DIFF
--- a/frontend/src/views/TeamView.tsx
+++ b/frontend/src/views/TeamView.tsx
@@ -607,8 +607,30 @@ function MemberCard({
             {member.description}
           </p>
         )}
-        {workload && <WorkloadLine workload={workload} />}
-        <DailyBudgetLine member={member} setByLabel={setByLabel} />
+        {/*
+          Pinned to the bottom of the card, not left floating under whatever
+          length the description happened to be.
+
+          `CardContent` is a `h-full` column inside a stretched grid row, so
+          every card in a row is the same height — but the content was all
+          top-aligned, and the description is `line-clamp-3`. A one-line
+          description therefore put this block ~36px higher than the two-line
+          card beside it, and the status line is the one thing a roster is
+          scanned for: "who is working, and how much is on them" was on two or
+          three different baselines in every row, with dead space underneath
+          each card.
+
+          `mt-auto` takes the slack instead, so the running facts line up
+          across a row and the card has no empty tail. Wrapped rather than
+          applied to `WorkloadLine` directly because a host that cannot answer
+          the board renders no workload at all (see `IDLE` and the `workload`
+          prop) — the budget line has to inherit the same anchor, or the two
+          shapes of card disagree again.
+        */}
+        <div className="mt-auto space-y-1.5 empty:hidden">
+          {workload && <WorkloadLine workload={workload} />}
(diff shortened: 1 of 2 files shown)


oxoxDev:
@coderabbitai full review

PR #1376, opened 2026-08-20
design: give board columns a real surface, and collapse only when they do not fit

### LINKED ISSUE #1343: Work board: dark-mode columns and cards have no surface separation from the page
## What happens

`LedgerBoard` paints every column `bg-card/40`. In dark that is `#0C0D0F` at 40% over a `#08090B` page — a difference of roughly one value step at 40% alpha, i.e. invisible. At 1440px dark the board reads as a flat black field with hairline borders; the cards (`bg-card`, `#0C0D0F`) are equally indistinguishable from the column they sit in.

The Figma dark reference (`927:2294`, file `iHAo3wYujP9bRlREmlTVBK`) shows the opposite: the column is a clearly raised surface against the page, and the drop zone inside it is a third rung.

The same alpha trick costs light mode less (white at 40% over `#F7F7FC`) but still leaves the column flatter than the reference.

## Fix

Paint the column at a full surface rung rather than an alpha of one, so the three-rung hierarchy (page / column / card) survives in both themes. Colour through tokens only.

`frontend/src/views/LedgerBoard.tsx`

### LINKED ISSUE #1345: Work board: the collapsed column rail pins its count ~650px below its label
## What happens

A collapsed column (issue #1101) renders the rotated label at the top of a full-height 40px rail and the count with `mt-auto`, which drops it to the very bottom. At a 900px viewport the label sits at y≈205 and its own count at y≈851.

The two halves of one fact are 650px apart, separated by nothing. The count reads as an orphan.

## Fix

Keep the label and count together at the top of the rail.

`frontend/src/views/LedgerBoard.tsx`

### LINKED ISSUE #1346: Work board: empty columns collapse to rails even when the board has room to spare
## What happens

`collapsible` in `LedgerBoard` is true as soon as **any** column holds a card:

```ts
const collapsible = !loading && columns.some((c) => (held.get(c.id)?.length ?? 0) > 0);
```

It never asks whether the board is actually short of room. Consequences:

1. **A young declared list looks broken.** `#/ledgers/incidents` (3 stages, 1 resolved row) renders as two 40px rails of rotated text, one real column, and ~800px of empty page to the right. Every declared list is in that state on day one.
2. **The board reflows mid-drag.** A rail opens under the pointer (correctly — that part of #1101 works), which pushes every column to its right 248px while the operator is holding a card and aiming at one of them.

The #1101 premise — "the board that has moved the most work looks the most finished" — only bites when the columns genuinely do not fit. That is the condition to test.

## Fix

Gate collapsing on the board actually overflowing its container, not merely on some column having cards.

`frontend/src/views/LedgerBoard.tsx`

### LINKED ISSUE #1350: Work board: column geometry and empty-state copy are off the Figma reference
## Measured against `927:2643` (light) / `927:2294` (dark), file `iHAo3wYujP9bRlREmlTVBK`

| | app | Figma |
|---|---|---|
| column width | 288px (`w-72`) | 260px |
| gutter | 16px (`gap-4`) | 19px |
| empty-column copy | "Nothing here" | "Drop tasks here" |
| count badge at 0 | shown | omitted |
| light-mode column | flat | vertical gradient wash |
| scroll-edge treatment | hard clip | bleed, with the wash carrying the eye |

Note for the record: the columns measure **260/19**, not the 208/28 that has been quoted second-hand.

Two of these are more than pixel-matching:

- **"Nothing here" is a status; "Drop tasks here" is an affordance.** The board`s whole gesture is the drop, and the empty column is the one place to say so.
- **The right-edge bleed reads as clipping, not as a scroll affordance**, because nothing fades. At 1100px the app shows 3.5 of 6 columns and cuts a Resume button mid-word with no visual cue that more exists.

## Fix

Adopt the reference geometry, the drop-affordance copy, the omitted zero badge, and add the edge treatment that makes the bleed legible as "there is more this way".

`frontend/src/views/LedgerBoard.tsx`

graycyrus:
Closes #1343, closes #1345, closes #1346, closes #1350.

From a design audit of the **Work** surface. Four findings, one component (`LedgerBoard.tsx`).

## 1. Dark-mode columns had no surface (#1343)

Columns were painted `bg-card/40` — one surface rung at 40% alpha. Over the dark page that is `#0C0D0F` on `#08090B` at 40%, i.e. no step at all: the board read as a flat black field of hairlines, with cards the same colour as the column holding them. The Figma dark frame (`927:2294`) shows the opposite — a clearly raised column, a third rung for the drop zone.

Now a full rung plus the reference's vertical wash, which also stops the light board reading as one flat card.

## 2. Empty columns collapsed with room to spare (#1346)

`collapsible` asked *"does some other column have cards?"*. #1101's premise is a claim about **width** — "the 101 cards you came for are two columns off the right edge" — so width is what it should be tested against.

- `#/ledgers/incidents` (three stages, one resolved row) rendered as two 40px rails of rotated text, one real column, and ~800px of empty page. Every declared list is in that state on day one.
- A rail opening under a drag (which is correct, and still works) shoved every column to its right by 248px while the operator was aiming at one of them.

`fits` now does the arithmetic — `columns × 260 + gutters + trailing ≤ viewport` — measured through a `ResizeObserver` in a layout effect. Deliberately arithmetic rather than reading the rendered board's `scrollWidth`: the rendered board is already collapsed, so that reading would report "fits", uncollapse, report "does not", and oscillate.

## 3. The rail's count sat 650px from its label (#1345)

`mt-auto` dropped it to the foot of a full-height rail. Two halves of one fact with nothing between them.

## 4. Geometry and empty-state copy (#1350)

Measured against `927:2643` / `927:2294` (file `iHAo3wYujP9bRlREmlTVBK`):

| | before | now |
|---|---|---|
| column | 288px | **260px** |
| gutter | 16px | **20px** |
| empty copy | "Nothing here" | **"Drop tasks here"** |
| right edge | hard clip | **fade** |

"Nothing here" reports a fact the empty column already makes obvious; the board's whole gesture is the drop, and the empty column is the one place with room to say so. The bleed is deliberate in the reference, but a hard clip is indistinguishable from a layout bug — at 1100px the app was cutting a button through the middle of its own label. The fade is what makes the same bleed read as *more this way*.

For the record: the columns measure **260/19** in the frame, not the 208/28 that has been quoted second-hand.

## Tests

`test/unit/board-collapsed-columns.test.ts` gains three cases: a board with room collapses nothing however the work is distributed; one column too many brings the rails back; and the pixel geometry `fits` reasons about is asserted equal to the Tailwind classes that paint it, because nothing else would notice those drifting apart.

Every pre-existing case in that suite is untouched and green — jsdom reports `clientWidth: 0`, which the board reads as *nothing is known to fit*, the conservative answer.

## Commands run

- `npm run typecheck` / `typecheck:unit` / `typecheck:e2e`
- `npx vitest run` — 1599 passed, 167 files
- `scripts/ci/assert-design-tokens.sh`
- Browser-verified at 1440px and 1100px, light and dark, against a live seeded host


oxoxDev:
@coderabbitai full review

PR #1379, opened 2026-08-20
fix(chat): make the top of an empty channel readable

### LINKED ISSUE #1323: Chat: an empty channel or DM pins its intro and action cards to the bottom of an empty screen
Audit of `#/chat` at 1440px and 1100px, light and dark, against a seeded host.

## What happens

Open any channel or DM with no messages — e.g. a DM with any teammate on a fresh company. The identity block (avatar, name, "This is the start of your direct message with …") and the two **Create teammate / Add people** action cards are rendered flush against the composer at the very bottom of the pane, under roughly 850px of empty canvas.

The two action cards are the primary call to action on an empty channel, and they are the *last* thing the eye reaches. The pane reads as a rendering failure rather than an invitation.

## Cause

`frontend/src/views/chat/MessageTimeline.tsx` wraps the transcript in:

```tsx
<div className="flex min-h-full flex-col justify-end pb-4">
```

`justify-end` on a `min-h-full` flex column bottom-pins content that is shorter than the viewport. That is the right rule for a short *transcript* — a three-message channel should sit above the composer the way Slack does — but an empty channel has no transcript, only the intro, and bottom-pinning it is what produces the void.

`ChannelIntro` already carries `empty ? "pt-16" : "pt-6"`, which was clearly written expecting the block to be top-anchored.

## Expected

When the channel has no items, the intro block anchors to the **top** of the pane, as the Figma reference does (file `iHAo3wYujP9bRlREmlTVBK`, nodes `927:2097` / `927:2386`): heading, description, then the action cards, reading downward from the top.

Channels that do have messages keep todays bottom-anchored behaviour unchanged.

### LINKED ISSUE #1327: Chat: the channel intro draws a person's mascot for a channel that is not a person
At the top of `#engineering` (and every channel except `main`) the intro block renders a large 48px teammate mascot — a purple cartoon figure — above the heading.

A channel is not a person. The mascot is seeded on the channel *name*, so `#engineering` gets an arbitrary face that belongs to nobody, sits at the largest avatar size on the surface, and is the first thing in the pane.

`frontend/src/views/chat/MessageTimeline.tsx`, `ChannelIntro`:

```tsx
<TeammateAvatar
  name={channel.voice ?? channel.name}
  tone={channel.tone}
  avatar={channel.member?.avatar}
  company={channel.kind === "channel" && channel.id === "main"}
  className="mb-3 size-12 rounded-lg text-base"
/>
```

`company` is only true for `main`, so every other channel falls through to the mascot branch.

The header a few pixels above already gets this right — `ChatHeader`s `KindIcon` draws `#` for a channel, a lock when private, and the teammates face only for a DM, precisely so that "a channel has no face" (issue #1170). The intro block contradicts it on the same screen.

The Figma reference (file `iHAo3wYujP9bRlREmlTVBK`, nodes `927:2097` / `927:2386`) shows no avatar in the channel intro at all — just `#general` as a heading with the description beneath.

## Expected

A DM keeps the teammates face. A channel gets its kind mark (`#` / lock) on a tone tile, or no mark at all — but not a stranger.

### LINKED ISSUE #1339: Chat: the empty-channel action cards are flat where the reference gives them a soft brand glow
The **Create teammate** / **Add people** cards on an empty channel are `rounded-xl border bg-card` and nothing else. Against the near-black canvas in dark mode they are almost invisible — the border is the only thing separating them from the ground, and they read as two empty boxes rather than the surfaces two primary actions.

The Figma reference (file `iHAo3wYujP9bRlREmlTVBK`, nodes `927:2097` dark / `927:2386` light) gives each card a soft radial brand glow bleeding from its lower-left, which is what lifts them off the canvas and marks them as the invitation on an otherwise empty screen.

This has to go through the token layer — `scripts/ci/assert-design-tokens.sh` rejects raw hex and raw Tailwind palette steps, and correctly so, since a glow that cannot theme is a glow that is wrong in one of the two modes. `--shadow-brand` in `frontend/src/index.css` is the existing precedent for a brand-tinted effect built with `color-mix(in oklab, var(--brand-500) …)`; a radial-gradient token belongs beside it.

Site: `frontend/src/views/chat/MessageTimeline.tsx`, `ActionCard`.

graycyrus:
Design audit of the Chat surface, driven against a running seeded host at 1440px and 1100px in both themes. Three faults in one block — the intro above a channel's transcript.

## 1. It was pinned to the bottom of an empty screen (#1323)

`justify-end` on a `min-h-full` flex column bottom-pins content shorter than the viewport. Right for a short *transcript* — three messages belong above the composer, as every chat client puts them — wrong for a channel with no transcript, where the only thing bottom-pinned is the intro. The two action cards that are the entire point of an empty channel rendered crushed against the composer under roughly 850px of dead canvas, so the primary invitation was the last thing the eye reached.

An empty channel now reads downward from the top. Two behaviours deliberately unchanged:
- a channel **with rows** keeps the bottom anchor;
- a channel **still loading** keeps it too, because the skeleton fills the space the real rows will — flipping to the top there would lift the intro and drop it back the moment history landed, which is the jump the skeleton exists to prevent.

## 2. It drew a face for a thing with no face (#1327)

Every channel but `main` fell through to `TeammateAvatar` seeded on the channel *name*, so `#engineering` grew an arbitrary mascot belonging to nobody — at the largest avatar size on the surface, as the first thing in the pane — while `ChatHeader` a few pixels above drew `#` for the same channel. Two marks for one thing, disagreeing on screen.

`IntroMark` now applies the rule the header settled in #1170: a DM wears its teammate's face (seeded through the shared `dmFace`, so the intro, the rail row and the header cannot drift about who a DM is with); `main` keeps the company brand mark; every other channel wears its kind (`#`, or the lock when private).

## 3. The action cards did not lift off the canvas (#1339)

`rounded-xl border bg-card` and nothing else read as two empty boxes against near-black. `--glow-brand-card` adds the soft radial brand glow the design reference gives them.

Tokenised, beside `--shadow-brand` and built the same way (`color-mix(in oklab, var(--brand-N) …)`): a glow is a colour, and a colour that cannot theme is wrong in one of the two modes — the light value is barely there, and dark needs `--brand-400` at roughly twice the mix to separate a card from the canvas. Exposed through `--background-image-*`, which is the namespace that generates a `bg-*` utility for a gradient; **verified against the emitted CSS**, not assumed:

```
.bg-glow-brand-card{background-image:var(--glow-brand-card)}
```

## Tests

- `frontend/test/unit/chat-scroll-anchor.test.ts` — a new block for which end short content settles against. Deliberately separate from the scroll rules already in that file: a pane shorter than its viewport has no scroll range, so every anchoring call is a no-op there and only the flex alignment decides. Covers empty, has-rows, and still-loading.
- `frontend/test/unit/chat-channel-intro.test.ts` (new) — which mark the intro draws, told apart by what the element is made of rather than by a class name (only the mascot branch renders an ``), so a query cannot pass by construction. Covers DM, plain channel, private channel, `main`, and a DM with no roster entry.
- No unit test for the glow — it is pure CSS, covered by `assert-design-tokens.sh` and the build check above.

## Gates run locally

| gate | result |
|---|---|
| `npx vitest run` | 168 files, 1604 tests, all pass |
| `npm run build` | pass (and used to verify the utility above) |

Browser-verified in **both themes** at 1440px against a real host: the empty DM, the empty channel with its cards, and a channel with a transcript to confirm the bottom anchor is undisturbed.

Closes #1323, closes #1327, closes #1339.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -329,6 +329,22 @@
        surface. Not a substitute for the focus ring — an addition to it. */
     --shadow-brand: 0 2px 8px -2px color-mix(in oklab, var(--brand-500) 32%, transparent);
 
+    /* A soft brand glow bleeding from a card's lower-left corner, for the
+       standing invitations on an otherwise empty surface — the "create a
+       teammate" / "add people" pair on a channel with no transcript.
+
+       A token rather than a gradient written at the call site because a glow
+       is a *colour*, and a colour that cannot theme is wrong in one of the two
+       modes: the value below is barely there against a light canvas, and the
+       `.dark` counterpart has to work several times harder to lift a card off
+       near-black. One utility, two tunings, no call site that has to know
+       which mode it is in. */
+    --glow-brand-card: radial-gradient(
+        120% 108% at 14% 112%,
+        color-mix(in oklab, var(--brand-500) 13%, transparent) 0%,
+        transparent 62%
+    );
+
     /*
      * Third-party brand colours.
      *
@@ -463,6 +479,13 @@
     --shadow-xl: 0 8px 16px -8px hsl(var(--shadow-color) / 0.55),
         0 24px 56px -12px hsl(var(--shadow-color) / 0.6), inset 0 1px 0 0 oklch(1 0 0 / 7%);
     --shadow-brand: 0 2px 12px -2px color-mix(in oklab, var(--brand-400) 40%, transparent);
+    /* The light value is invisible here — `--brand-400` at roughly twice the
+       mix is what actually separates the card from the canvas. See `:root`. */
+    --glow-brand-card: radial-gradient(
+        120% 108% at 14% 112%,
+        color-mix(in oklab, var(--brand-400) 22%, transparent) 0%,
+        transparent 62%
+    );
 }
 
 /* ============================================================================
@@ -579,6 +602,11 @@
     --shadow-lg: var(--shadow-lg);
     --shadow-xl: var(--shadow-xl);
     --shadow-brand: var(--shadow-brand);
+
+    /* `--background-image-*` is the namespace that generates a `bg-*` utility
+       for a gradient, the way `--color-*` does for a flat fill. Gives
+       `bg-glow-brand-card`. */
+    --background-image-glow-brand-card: var(--glow-brand-card);
 
     --radius-sm: calc(var(--radius) * 0.6);
     --radius-md: calc(var(--radius) * 0.8);
--- a/frontend/src/views/chat/MessageTimeline.tsx
+++ b/frontend/src/views/chat/MessageTimeline.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useLayoutEffect, useRef } from "react";
-import { Bot, UserPlus } from "lucide-react";
+import { Bot, CircleDot, Hash, Lock, UserPlus } from "lucide-react";
 
 import type { ApprovalSummary, GrantScope, TurnStep, Verdict } from "@/api/types";
 import { TeammateAvatar } from "@/components/teammate-avatar";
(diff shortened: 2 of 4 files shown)


oxoxDev:
@coderabbitai full review

Issue #1185, opened 2026-08-19
console: move Avatar out of views/chat, and settle the mascot seed on the teammate id


graycyrus:
Two loose ends left by #1181 (fixed in PR #1168), both deliberately not bundled there.

## 1. `Avatar` still lives under `views/chat/`

#1181 noted the import path misleads: `views/chat/Avatar.tsx` now serves the
Company cards and the teammate detail header too, and #1170 will add a sixth
surface.

It was **not** moved in #1168 because `frontend/src/components/ui/avatar.tsx`
already exists — the Base UI primitive. Moving the chat one to
`components/avatar.tsx` would put two different components named `Avatar`
side by side under `components/`, which is a worse import path than the one
being fixed. Naming it properly (`TeammateAvatar`? `VoiceAvatar`?) is its own
decision, and doing it inside a Company-page PR would have buried a rename of
six import sites in an unrelated diff.

## 2. The mascot seed is name-based everywhere, and the id-seeded field is dead

This is the more interesting one, and #1181's premise had it backwards.

`lib/team.ts` documents `avatarFor` as seeded on the id — *"renaming a teammate
does not change its face"* — and `fromDto` computes `TeamMember.avatar` from
`dto.id || name`.

**Nothing renders that field.** Every `Avatar` call site in the console passes
`name` and `tone` only — `ChannelRail`, `MembersPane`, `MessageRow`,
`ThreadPanel`, `MessageTimeline`, and now the two Company surfaces — so all of
them fall through to `avatarFor(name)`.

The two seeds disagree for essentially every real roster row:

| teammate | `avatarFor(id)` | `avatarFor(name)` — what renders |
|---|---|---|
| `backend_engineer` | violet | indigo |
| `security_engineer` | ember | rose |
| `designer` | green | graphite |
| `product_manager` | rose | blue |
| `researcher` | clay | indigo |

So the console is **consistent but on the undocumented seed**, and a teammate
renamed on the host silently changes face everywhere — the exact instability the
id seed was introduced to prevent.

### What to do

Pass `TeamMember.avatar` at every call site **in one change**, never a subset —
a partial migration gives one teammate two faces, which is harder to spot than
the current state because each screen stays internally consistent.

Tractable, because the roster entry is already in hand at most sites:

- `MembersPane` — the row renders a full `TeamMember`; one prop.
- `ChannelRail` — the DM branch already guards on `channel.member`.
- `MessageTimeline` — `channel.member?.avatar` for DM channels.
[...]

PR #1393, opened 2026-08-20
fix(chat): give a thread's summary row faces and an age

### LINKED ISSUE #1324: Chat: the thread facepile renders as three blank grey blocks
On any message with replies, the thread affordance reads:

    [▪][▪][▪]  6 replies   Last reply 1:08 AM

The three tiles are identical featureless dark rectangles. They look like a loading skeleton or three broken images, not like faces, and they carry no information at all.

## Cause

`frontend/src/views/chat/MessageRow.tsx`, `ReplyFacepile`:

```tsx
<TeammateAvatar
  key={r.id}
  name={r.from === "you" ? "You" : (r.channel ?? "Company")}
  tone={r.channel}
  markOnly
  className="size-4 rounded-[3px] ring-1 ring-background"
/>
```

Two things combine:

1. `markOnly` renders an **empty** span — a tone tile with no initials and no mascot, by design, because 16px is below the size at which either is legible.
2. `tone={r.channel}` seeds the tile on the *channel*, and every reply in a thread shares one channel. So all three tiles hash to the same tone and the facepile has nothing to differentiate.

The result is three copies of one grey square. Whatever `markOnly` claims to say — "the colour still distinguishes one voice from the next" — it cannot say it when the seed is constant.

## Expected

Either seed the tile on the **author** so the pile shows distinct voices, or drop the pile and let the reply count and the last-reply time carry the affordance on their own. A facepile that cannot distinguish anybody is worse than no facepile.

Related: the tiles are 16px, below the floor `TeammateAvatar` itself documents for legibility. If they stay, they want to be ~20px.

### LINKED ISSUE #1328: Chat: a thread's "Last reply" should be relative, not a wall-clock time
The thread affordance under a message reads:

    6 replies   Last reply 1:08 AM

`formatTime` gives an absolute clock time. For "when did this thread last move" that is the least useful form: on a transcript older than a day it is ambiguous without the date, and even inside one day it makes the reader do the subtraction.

The Figma reference (file `iHAo3wYujP9bRlREmlTVBK`, node `927:2097`) shows `3 replies · last reply 6 hours ago`, which is what every other chat client does, and what the phrase "last reply" is actually asking.

Two smaller things in the same line:

- There is no separator between the count and the timestamp. The reference puts a `·` between them; today they run together as `6 repliesLast reply 1:08 AM` with only a space.
- The day divider directly above already carries "Today", so the absolute time is doubly redundant on the common case.

Site: `frontend/src/views/chat/MessageRow.tsx`, the `replies.length > 0` block.

graycyrus:
Design audit of the Chat surface. The summary row under a threaded message was carrying two pieces of information and getting both wrong.

Before / after, same thread, same host:

| | |
|---|---|
| before | `▪▪▪  6 replies  Last reply 1:08 AM` |
| after | `🟢🟣  6 replies · Last reply 26m ago` |

## 1. The facepile said nothing (#1324)

Each tile was seeded on the *message's* channel and drawn `markOnly`. Both halves defeated it:

- every reply in a thread carries the same channel, so all three tiles hashed to **one tone**;
- `markOnly` renders a deliberately **empty** tile, because 16px is below the size at which initials or a mascot can be read.

So the row drew three identical featureless grey squares. They read as a loading skeleton, and they carried no information — a facepile that cannot distinguish anybody is worse than no facepile.

Both halves are fixed by resolving the senders properly. `TimelineEntry` gains `replySenders`: the distinct voices behind a row's replies, deduped by `Sender.key`, in first-spoken order, resolved through **the same `senderOf` every rendered row uses** — so a face in the summary is the same face the thread shows when opened. That is the agreement #1170 and #1185 established for DMs, extended to threads.

Resolving it in the model rather than the row is what makes it possible at all: a sender needs the channel and the roster, and neither reaches the renderer. It also puts the logic somewhere a pure unit test can reach.

With genuinely different people to draw, the tile goes to **20px** — the size the rail already draws a DM's face at, legibly — so the real mascot fits and there is nothing left for `markOnly` to apologise for.

A system line among the replies draws no face: it has no voice, and counting it would claim one more participant than the thread has.

## 2. "Last reply 1:08 AM" answered the wrong question (#1328)

"Last reply" asks how long a thread has been quiet, and a wall-clock time is the one form that does not say it: on a transcript older than a day it is ambiguous without the date, and inside one day it makes the reader do the subtraction — while the day divider above already carries "Today", so it was doubly redundant on the common case.

Now `timeAgo`, the helper `ApprovalsView`, `TaskCard` and `FeedbackView` already share (reused rather than added — the point of that module is that two surfaces cannot drift into two vocabularies for one idea), plus the `·` that was missing between the count and the age.

## Tests

`frontend/test/unit/chat-timeline.test.ts` — the model change is pure, so it is tested as one. Five cases: two company voices the message channel alone could not tell apart; one person who replied three times counting once (a pile is people, not messages); first-spoken ordering held when someone replies again; a system line drawing nothing; and a row with no replies having no voices.

## Gates run locally

| gate | result |
|---|---|
| `npx vitest run` | 167 files, 1601 tests, all pass |
| `npm run build` | pass |

Browser-verified in **both themes** at 1440px against a real host, on a six-reply thread.

Closes #1324, closes #1328.

### DIFF
--- a/frontend/src/views/chat/MessageRow.tsx
+++ b/frontend/src/views/chat/MessageRow.tsx
@@ -4,6 +4,7 @@ import { Markdown } from "@/components/markdown";
 import { TeammateAvatar } from "@/components/teammate-avatar";
 import { Button } from "@/components/ui/button";
 import { isHostMessageId, type ChatMessage } from "@/lib/chat";
+import { timeAgo } from "@/lib/language";
 import { cn } from "@/lib/utils";
 import {
   formatTime,
@@ -125,10 +126,16 @@ export function MessageRow({
             onClick={() => onOpenThread(message.id)}
             className="mt-1 flex w-fit items-center gap-2 rounded-md px-1.5 py-1 text-xs font-medium text-primary transition-colors hover:bg-accent"
           >
-            <ReplyFacepile replies={replies} />
+            <ReplyFacepile senders={entry.replySenders} />
             {replies.length} {replies.length === 1 ? "reply" : "replies"}
+            {/* "Last reply" asks how long a thread has been quiet, and a
+                wall-clock time is the one form that does not answer it: on a
+                transcript older than a day it is ambiguous without the date,
+                and inside one day it makes the reader do the subtraction. The
+                day divider above already carries "Today", so the absolute time
+                was doubly redundant on the common case (issue #1328). */}
             <span className="font-normal text-muted-foreground">
-              Last reply {formatTime(replies[replies.length - 1].at)}
+              · Last reply {timeAgo(replies[replies.length - 1].at, Date.now())}
             </span>
           </button>
         )}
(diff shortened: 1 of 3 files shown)


oxoxDev:
@coderabbitai review

Issue #1224, opened 2026-08-20
chat: a channel with real history opens at the top and then stops following new messages for the rest of the session


graycyrus:
Open the console on a channel with any real history and you land at the **very
beginning** of it — "This is the very beginning of #engineering", messages from
40 minutes ago — with 3,438px of conversation below the fold. The view never
corrects, and from then on the channel **stops following new messages
altogether** for the rest of the session.

Measured, cold load of `#/chat/engineering` on the sample company:

```
                              scrollTop   distance from bottom
cold load, settled                   41                  3438
+8s (nothing touched)                41                  3438
after a new message arrives          41                  3492   ← view did not move
```

The new message is not the only casualty: everything already sent is off-screen
too. Screenshot in the thread — the channel reads like a brand-new one.

## The animation captures a pixel offset, then the transcript grows under it

A `requestAnimationFrame` trace of the scroll container from page load
(`t` in ms from navigation, `sh` = scrollHeight, `ch` = clientHeight):

```
t=66    top=0    sh=709   ch=709    empty — history still on the wire
t=79    top=0    sh=750   ch=709    the channel intro block; max scrollTop is now 41
t=104   top=4    sh=750   ch=709    a smooth scrollTo is under way, target 41
t=112   top=14   sh=750   ch=709
t=141   top=25   sh=4188  ch=709    ← history hydrates. The target is still 41.
t=152   top=34   sh=4188  ch=709
t=178   top=41   sh=4188  ch=709    animation ends. 3438px from the bottom, forever.
```

Two separate things go wrong in that window, and the second is what makes it
permanent.

**1. The smooth scroll is aimed at a number, not at "the bottom."**
`MessageTimeline.tsx:147-155` (rule 2) fires while `items` is still empty and
the container is 750px tall, so `el.scrollTo({ top: el.scrollHeight })` captures
**41**. The browser then eases to 41 regardless of the transcript becoming
4188px mid-flight.

**2. The animation's own scroll events convince the timeline the operator
scrolled away.** `trackFollowing` (`MessageTimeline.tsx:118-123`) runs on every
`scroll` event and cannot tell the browser's smooth-scroll animation from a
human wheel:

```ts
const fromBottom = el.scrollHeight - el.scrollTop - el.clientHeight;
following.current = fromBottom <= BOTTOM_SLACK_PX;   // 32
```

At `t=141` the content jumps to 4188 while the animation sits at `top=25`, so
[...]

Issue #757, opened 2026-08-12
chat: opening a channel animates a scroll to the bottom instead of starting there


oxoxDev:
## What the operator sees

Opening a channel does not *start* at the newest message — it starts higher up and then visibly animates downward. Every channel switch replays the scroll, so moving between channels is a series of little downward slides before the view settles. It should already be at the bottom on the first painted frame.

## Where it comes from

`frontend/src/views/chat/MessageTimeline.tsx` has one scroll effect, and it always animates:

```ts
useEffect(() => {
  const el = scroller.current;
  if (!el) return;
  el.scrollTo({ top: el.scrollHeight, behavior: "smooth" });
}, [items.length, typing, liveStepCount]);
```

`behavior: "smooth"` is right for the case it was written for — a tool row or reply arriving *while you watch*, where the movement shows you something changed. It is wrong on mount and on channel switch, where there is no prior position worth animating from and the animation is the entire artefact being complained about. The longer the transcript, the longer the slide.

The container is bottom-anchored by layout (`min-h-full flex-col justify-end`), so a short channel looks correct and a long one does not — which is why this reads as intermittent.

## A second defect in the same effect

The dependency list is `[items.length, typing, liveStepCount]` — **the channel identity is not in it.** Two consequences:

1. Switching between two channels whose `items.length` happens to match fires no effect at all, so the new channel can open at the previous channel's scroll offset.
2. `items.length` is a proxy for "something arrived". A transcript that replaces content without changing the count does not re-anchor.

## And a third: it steals the scroll

There is no "only stick if the operator is already at the bottom" check. Scroll up to read history, and the next inbound message or tool row yanks the view back to the bottom mid-read. Standard chat behaviour is to hold position when the reader has deliberately scrolled away, usually with a "jump to latest" affordance.

## Direction

- Anchor **instantly** (`behavior: "auto"`) on mount and on channel change; reserve `"smooth"` for growth while the channel is already open and already at the bottom.
- Put the channel id in the dependency list so every switch re-anchors.
- Track whether the scroller is near the bottom, and only auto-follow when it is. When it is not, leave the position alone.

## Acceptance

- Opening any channel paints at the newest message with no visible animation, regardless of transcript length.
- Switching between two channels with equal message counts still anchors correctly.
- A message arriving while the operator has scrolled up does not move the viewport.
- A message arriving while the operator is at the bottom still follows, smoothly.

## Related

[...]

PR #1412, opened 2026-08-20
fix(chat): follow the bottom when the composer grows

### LINKED ISSUE #1325: Chat: the transcript does not re-anchor when the composer grows, hiding the newest messages while you type
Measured against a running host at 1440x900, `#/chat/engineering`, on a transcript long enough to scroll.

Probing the scroller (`.flex-1.overflow-y-auto`) and the composer textarea:

| state | composer height | scrollTop | distance from bottom |
|---|---|---|---|
| at rest | 40px | 2293 | **0** |
| after typing a long draft | 136px | 2293 | **96** |
| after clearing the draft | 40px | 2293 | 0 |

`scrollTop` never moves. The composer grows (`field-sizing-content`, up to `max-h-48`), the scroller `clientHeight` shrinks by the same amount, and the transcript silently slides up behind it — so the last ~96px of the conversation, which is very often the message being replied to, is hidden for exactly as long as the draft is long.

At the `max-h-48` cap the composer can take ~150px+, so up to about two message rows can disappear while composing.

## Why it happens

`frontend/src/views/chat/MessageTimeline.tsx` re-anchors in an effect keyed on
`[channel.id, historyPending, items.length, typing, liveStepCount]`. Composer
height is not in that list and cannot be — it lives in a sibling component. Nothing
observes the scrollers own size, so a container resize is invisible to the
anchoring rules.

Note the near-miss: `BOTTOM_SLACK_PX` is 32, and 96 is past it. A container resize
fires no scroll event, so `trackFollowing` does not run and `following.current`
stays true — follow-mode survives. But any rule that *reads* the distance while a
draft is open now reads the view as scrolled away.

## Expected

Rule 2 in that files doc comment — "growth follows, but only if they are already
at the bottom" — should hold for growth of the *viewport* as well as growth of the
content. A `ResizeObserver` on the scroller that re-anchors when
`following.current` is the natural fix, and it costs nothing when the operator has
deliberately scrolled up.

graycyrus:
Found auditing the Chat surface: composing a long message hides the newest part of the conversation you are composing about.

## The measurement

Probing the scroller and the composer textarea against a running host at 1440x900, on a transcript long enough to scroll.

**Before**

| state | composer | scrollTop | clientHeight | from bottom |
|---|---|---|---|---|
| at rest | 40px | 2293 | 709 | **0** |
| long draft open | 136px | 2293 | 613 | **96** |
| draft cleared | 40px | 2293 | 709 | 0 |

**After**

| state | composer | scrollTop | clientHeight | from bottom |
|---|---|---|---|---|
| at rest | 40px | 2293 | 709 | **0** |
| long draft open | 136px | **2389** | 613 | **0** |
| draft cleared | 40px | 2293 | 709 | 0 |

96px of transcript hidden on a two-line draft, and up to ~150px at the `max-h-48` cap — roughly two message rows — for exactly as long as the draft is long.

## Why it happened

The file documents two anchoring rules (#757, #1224), and both watch the **content**. Neither watches the **box**, and the box moves: the composer grows with the draft (`field-sizing-content`) and takes its height out of this scroller's `clientHeight` while `scrollTop` stands still.

It could not be fixed by adding a dependency to rule 2 — the composer is a sibling component and its height is not a value this one is given. The element's own size is, through `ResizeObserver`. Observing the box also covers the window resizing and the thread panel opening, which want the same answer.

Two deliberate choices, both documented at the call site:

- **Gated on `following.current`**, the same gate rule 2 uses, so a reader who has scrolled up to read history is left alone. Opening a draft must not yank them to the newest message.
- **Instant, not smooth.** Rule 2 animates because a message arriving should travel into view. This fires as the composer grows a line at a time, so an animation per keystroke would be a permanent wobble. Setting `scrollTop` does not resize anything, so there is no feedback loop.

There is a near-miss worth recording: `BOTTOM_SLACK_PX` is 32 and the gap was 96, so a view with a draft open reads as "scrolled away" to anything measuring the distance. A container resize fires no scroll event, so `trackFollowing` never ran and follow-mode survived — but the margin was luck, not design.

## Tests

`frontend/test/unit/chat-scroll-anchor.test.ts`. The harness already stubs geometry because jsdom performs no layout; this adds a **mutable viewport height** (mirroring how `contentHeight` was already mutable for #1224) and a controllable `ResizeObserver`, since jsdom ships none. The global is restored in `afterEach` under the same rule `restoreGeometry` follows.

Four cases: following the bottom down when the pane shrinks; leaving a scrolled-up reader alone; continuing to follow as the composer grows a line at a time; and not animating.

**Verified non-vacuous** — with the observer body short-circuited, three of the four fail. The fourth is the no-regression guard, which correctly passes either way. The "does not animate" case asserts the anchor *moved* alongside the absence of a smooth call, since "no smooth call" is otherwise satisfied by doing nothing.

## Gates run locally

| gate | result |
|---|---|
| `npx vitest run` | 167 files, 1600 tests, all pass |

Browser-verified in **both themes** at 1440px against a real host, plus the numeric probe above re-run against the fixed build.

Closes #1325.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat scrolling when the viewport or message composer changes size.
  * Users at the bottom of the conversation remain anchored automatically.
  * Users who have scrolled up can continue reading without being repositioned.
  * Repeated viewport changes now preserve the expected scroll position without animation.


oxoxDev:
@coderabbitai review

Issue #839, opened 2026-08-13
Company page: no way to create a teammate, and desk hierarchy reorders one click at a time instead of by dragging


oxoxDev:
## Summary

Two gaps on the Company page, reported from operating the product. They are filed together because they are the same surface and the same complaint: the page that owns company structure cannot do the two structural things an operator most wants to do there.

## 1. There is no way to create a teammate from the Company page

The Company page is where desks and their members live, and since the flat Desks page was removed it is the only route into desk creation and membership. It can create a **desk**, and it can attach an **existing** teammate to one. It cannot create a teammate.

The capability exists. `ApiClient.addTeamMember` posts to `{scope}/team` and the host persists it, but only two surfaces call it:

- `ChatView`'s members pane (`AddMemberDialog`), and
- `TeamView` — which is **parked**: it is in `HIDDEN_VIEWS` in `app-shell.tsx`, absent from the sidebar, reachable only by typing `#/team`.

So creating an agent means either going to Chat and opening a channel's members pane, or knowing a URL that is deliberately unlisted. Neither is discoverable from the page whose entire job is company structure, and an operator setting up a company hits this immediately — you cannot staff a desk you just made without leaving the page that made it.

**What good looks like:** a create action on the Company page that mints a teammate and, ideally, drops it straight onto the desk it was created from. The write path already exists; this is a UI gap, not a backend one.

## 2. Desk hierarchy reordering should be drag and drop

Members inside a desk are reordered with up/down chevrons — `OrgChartView` renders `ChevronUp`/`ChevronDown` per row and calls `onMove(index, "up" | "down")`, one position per click. Moving someone from the bottom of a six-person desk to the top is five clicks with a re-render between each.

Dragging is the natural gesture for a hierarchy, and two things make it cheap here:

- **The API is already the right shape.** `setDeskOrder` takes the *full* ordered member list, not a swap or a delta — exactly what a drop handler produces. No backend change.
- **The repo already drags.** `TasksView` implements board drag and has e2e coverage in `frontend/test/e2e/board-drag.spec.ts`. The pattern, and a spec to model a new one on, both exist.

Worth keeping the chevrons as a keyboard-accessible fallback rather than replacing them outright — drag alone is not reachable without a pointer.

## Why together

[...]

PR #1428, opened 2026-08-20
fix: one add-teammate control per desk, with the create path inside it (#1424)

### LINKED ISSUE #1424: Desks: "Add teammate" names two different actions, and the create path is an unlabelled icon
Design audit of the **Company** surface. Judgement call, no Figma reference.

## What happens

Every desk on `#/company/desks` ends with two controls, side by side with a 4px gap:

1. a full-width `outline` button reading **"Add teammate"** — a dropdown of roster teammates *not yet on this desk*. It **seats an existing** teammate.
2. flush against it, a `ghost` icon button with **no visible label**, `aria-label="Create teammate on <desk>"`, wearing `UserPlus` — it **creates a new** teammate on this desk.

Three problems, all the same problem:

- **"Add teammate" names two different actions on the same surface.** On `#/company` the header's "Add teammate" *creates* a teammate. Inside a desk, "Add teammate" *seats an existing one*. The Desks page header meanwhile calls creating "New teammate" — and wears the same `UserPlus` glyph as the unlabelled button that also creates.
- **An unlabelled icon touching a button that already says the words is not discoverable.** Nobody hunting for "define a new teammate on this desk" finds a bare glyph 4px from a labelled control.
- **When the roster is exhausted the labelled control goes disabled** and reads "Everyone is on this desk", which leaves the invisible affordance as the only way in. Creating a teammate here is still perfectly valid.

Also: `DropdownMenuContent` inherited the trigger's width. The trigger is full-bleed across a desk card, so at 1440px the menu is over a thousand pixels wide for ten short names, all left-aligned down the edge of an empty panel.

## Expected

One control per desk, labelled, never disabled by a full roster, whose menu carries both paths — whoever is left on the roster, then `New teammate…`. The menu is a fixed, readable width, and the create item stays visible rather than falling below a scroll.

Severity: **defect** (affordance + terminology).

Related: the page-level vocabulary is still split — the roster header says "Add teammate", the Desks header says "New teammate", and the create dialog is titled "Define a teammate". Worth a follow-up of its own; this issue is scoped to the desk footer.

graycyrus:
Closes #1424.

## The failure

Every desk on `#/company/desks` ended with two controls, 4px apart:

- a full-width **"Add teammate"** dropdown that *seated an existing* roster teammate, and
- flush beside it, an **unlabelled** `UserPlus` icon button that *created a new* one.

So "Add teammate" meant "create" on `#/company` and "seat an existing one" here; the icon that actually creates wore the same glyph as the page header's labelled "New teammate"; and when every roster teammate was already seated the labelled control went **disabled** ("Everyone is on this desk"), leaving the invisible affordance as the only way in.

The menu was also the trigger's width — full-bleed across a desk card, so over a thousand pixels at 1440 for ten short names.

## The change

One control per desk. It always says "Add teammate", is disabled only by an in-flight write, and its menu carries both paths:

- `On the roster` — whoever is left to seat, in a group that scrolls
- a separator, then `New teammate…` (`aria-label="Create teammate on <desk>"`, kept so the intent is still nameable)

with `"Everyone on the roster is already here."` as a line **inside** the menu rather than a dead trigger.

Two details worth naming:

- `DropdownMenuLabel` is Base UI's `Menu.GroupLabel` and **throws outside a `Menu.Group`** — a blank page, not a warning. Hence the `DropdownMenuGroup`; the empty-roster line is a plain `<p>` because there is no group there to label.
- The `max-h`/`overflow-y-auto` sits on the **group**, not the popup. On the popup, `New teammate…` falls below the fold on any company with a roster — which is the one item that had to become findable for merging the controls to be worth anything.

Menu width fixed at `w-64`.

## Tests

- `org-tree.spec.ts` — new: "a desk offers one add control, and it stays usable when the roster is exhausted". Asserts exactly one add control per desk, seats every remaining roster teammate through the menu, then asserts the trigger is still **enabled** and the menu still offers `Create teammate on Engineering` alongside "Everyone on the roster is already here.".
- The two existing `#839` create-on-a-desk tests updated to reach the item through the menu.

## Commands run locally

- `npx vitest run` — 1596 passed / 167 files
- `PW_BASE_URL=… npx playwright test test/e2e/org-tree.spec.ts` against a served console — **25 passed**
- Browser-verified against a live seeded host: 1440px dark, 1440px light, 1100px light — menu at 256px, `New teammate…` pinned below the scrolling roster group.

## Behaviour change

The per-desk create action moves from a bare icon button to a menu item. `aria-label` is unchanged, so anything locating it by accessible name still finds it — as a `menuitem` rather than a `button`.

### DIFF
--- a/frontend/src/views/company/OrgChartView.tsx
+++ b/frontend/src/views/company/OrgChartView.tsx
@@ -42,7 +42,10 @@ import { Button } from "@/components/ui/button";
 import {
   DropdownMenu,
   DropdownMenuContent,
+  DropdownMenuGroup,
   DropdownMenuItem,
+  DropdownMenuLabel,
+  DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
 import { Skeleton } from "@/components/ui/skeleton";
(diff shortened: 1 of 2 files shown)


graycyrus:
@coderabbitai review

Issue #1320, opened 2026-08-20
Overview: one concept, four names — pillar / department / desk


graycyrus:
## What

One concept, four names, all visible at once on `#/overview`:

| where | word |
|---|---|
| the graph's pillar control, and its paddles' `title` | **pillar** |
| the same paddles' `aria-label` | **department** |
| `views/overview/README.md`, ring 1 | **department (pillars)** |
| the Company page, the org chart, `company.toml`, the API | **desk** |

So the paddle a sighted operator reads as *"Previous pillar (←)"* is announced to a screen-reader user as *"Previous department"*, and both of them are pointing at the thing the rest of the product calls a **desk**.

`KnowledgeGraphFullscreen.tsx:129-151`, `:83-124`; `adapter.ts` ring 1; `docs/spec/runtime/*` and `Company` for `desk`.

## Why it matters

The graph is the landing page, so "pillar" is the first word a new operator meets for the company's central structural concept — and it is a word they will never see again anywhere else in the console, in the manifest, or in the docs. The mismatch between `title` and `aria-label` on the same button is a straightforward accessibility bug on top.

## Suggested fix

Pick one. `desk` is the declared, documented, API-level name and the one an operator will meet in `company.toml`; the graph is the outlier. If `pillar` is worth keeping as the graph's own visual metaphor, it should at least never disagree with its own `aria-label`.

Found in a design audit of `#/overview`.

Issue #1313, opened 2026-08-20
Overview: a company with no desks lands on a blank canvas that says "Pick a pillar / 0 PILLARS"


graycyrus:
## What

A company with no desks — which every newly provisioned company is until somebody seats one — lands on this:

- one lone core node in the middle of a 1440px grid
- `Pick a pillar` / `0 PILLARS`, with **no dots to pick**
- two large paddle buttons that render, take focus, and do nothing (`step()` returns early)
- a seven-entry legend for seven kinds of node, **none of which are on screen**
- a snapshot chip

No sentence explaining what this is. No next step. No link to the setup flow, to Company, to "add a teammate", to anything.

This is the **first screen of the product** for every new operator, and it is a blank canvas with an instruction it makes impossible to follow.

## Screenshot

Reproduce by stubbing `/desks`, `/team`, `/tasks`, `/workflows`, `/users`, `/memory` to empty arrays, or by provisioning a company and signing in before seating a desk.

## Suggested fix

An empty graph should say what the graph is for and offer the one action that would fill it. `ApprovalsView` already does this well (`EmptyApprovals` — *"All clear / Nothing is waiting on you"*); Overview has no equivalent. The pillar control should also stop instructing at `0 PILLARS`.

Found in a design audit of `#/overview`.

PR #1429, opened 2026-08-20
feat(overview): name the pillars instead of numbering them

### LINKED ISSUE #1309: Overview: the pillar selector offers three anonymous dots for three named desks

graycyrus:
Closes #1309.

## The problem

At rest the pillar control read:

```
Pick a pillar
3 PILLARS          ● ● ●
```

Three 10px dots at 50% opacity. The company's three desks — *Engineering*, *Product & Design*, *Go-to-Market* — are named right there on the graph, in their own colours, a few inches away. The control that exists to choose one of them refused to say which was which.

Each name lived only in its dot's `title` / `aria-label`: a second of hover on a 24x20 target with a pointer, and unreachable on touch. Once you picked one it finally said the name — so you had to click a blind dot to learn what it was.

Three smaller failures rode along:

- `Pick a pillar` is an imperative with no visible object.
- `3 PILLARS` is redundant next to three visible dots, and becomes actively wrong at `0 PILLARS` — an instruction the page makes impossible to follow, on the first screen a newly provisioned company sees.
- The text block was inert: not a button, not a menu, no cursor change. The only hit targets were the dots.

## The fix

One named chip per desk, in the host's own desk order — never re-sorted, the same rule ring 1 follows, so the selector cannot disagree with the wheel.

Each chip carries the colour its pillar's node and label already carry, so the chip and the pillar are visibly the same object. The active one takes that colour and goes bold; the rest sit in `--os-muted`. `aria-current` marks the focused desk.

The caption names the group (`PILLARS`) instead of instructing, and answers `No desks yet` when the company has seated none — which also gives the empty state one honest sentence where it had an impossible instruction.

**They wrap, rather than scroll or truncate the row.** Ten desks become three short lines in the top-left corner, which is a legible answer; a clipped row is not. Each name still truncates at `12rem` on its own so one very long desk name cannot push the others off.

The `←` / `→` paddles and `Escape` are untouched.

## What this gives up

The `1 / 3` position readout. Names replace it: knowing you are on *Go-to-Market* and that the others are *Engineering* and *Product & Design* is strictly more than knowing you are on the second of three.

## Tests

`frontend/test/unit/overview-pillar-selector.test.ts` — 6 cases: every desk named at rest with nothing focused; the host's order preserved (a selector that alphabetised would disagree with the wheel); `aria-current` on the focused desk and only that one; nothing marked at rest; a click bringing that desk forward; and the empty company saying `No desks yet` rather than `Pick a pillar`.

## Verified

`npm run typecheck`, `typecheck:unit`, `typecheck:e2e`, `npx vitest run` (168 files / 1602 tests), `scripts/ci/assert-design-tokens.sh`.

In a browser against the seeded `agentic-software-company` host:

```
light 1440  [Engineering, Product & Design, Go-to-Market]  none current
dark  1100  [Engineering, Product & Design, Go-to-Market]  Go-to-Market current
```

No console errors. Colours confirmed matching the pillar nodes in both themes.

## Not in this PR

Same audit, filed separately: the chrome buried under the detail rail (#1307, PR #1368), the graph having no way out (#1308, PR #1398), labels under the design system's 10px floor (#1310), the empty state generally (#1313), and `pillar` / `department` / `desk` being three words for one concept (#1320) — this PR keeps saying "pillar" rather than pre-empting that decision.

---
Raised as a draft.

Design audit of `#/overview`; findings are professional judgement — no Figma reference exists for this surface.

### DIFF
--- a/frontend/src/views/overview/README.md
+++ b/frontend/src/views/overview/README.md
@@ -141,6 +141,11 @@ the two board predicates the adapter needs. Theme tokens live under `.oc-kg`
 in `src/index.css`.
 
 The graph is the whole page, so its chrome stays minimal: a pillar selector, a
-kind legend, the side paddles, and the detail card. The docked directory index
+kind legend, the side paddles, and the detail card. The selector names every
+desk rather than numbering them (issue #1309) — it used to draw one anonymous
+dot per desk under the words "Pick a pillar", so the control that exists to
+choose a desk said which was which only on hover, and said nothing at all to a
+touch device. The chips wrap rather than scroll or clip; each carries the
+colour its pillar's node and label already carry. The docked directory index
 and the entity/function/action lenses were removed — with nothing else on the
 page competing for attention, they covered more of the graph than they earned.
--- a/frontend/src/views/overview/kg/KnowledgeGraphFullscreen.tsx
+++ b/frontend/src/views/overview/kg/KnowledgeGraphFullscreen.tsx
@@ -80,46 +80,59 @@ export function KnowledgeGraphFullscreen({
         {/* vault search — top-left while the Notes core is open */}
         {searchSlot && <div className="absolute left-5 top-5 z-10">{searchSlot}</div>}
 
-        {/* pillar selector — compact, TOP-LEFT: convenient, not in
-            the graph's way): current pillar + one dot per pillar to jump */}
+        {/* pillar selector — compact, TOP-LEFT: convenient, not in the
+            graph's way. One named chip per desk (issue #1309).
+
+            It used to be three 10px dots at 50% opacity under the words "Pick
+            a pillar", and the names existed only in each dot's `title` — so
+            the control that exists to choose a desk refused to say which desk
+            was which, while the graph named all three in their own colours a
+            few inches away. You had to click a blind dot to learn what it was.
+
+            The chips wrap rather than scroll or truncate the row: a company
+            with ten desks gets three short lines in the corner, which is a
+            legible answer, where a clipped row is not. The colour is the same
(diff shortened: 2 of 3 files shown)


graycyrus:
@coderabbitai review

Issue #1338, opened 2026-08-20
Retired top-level addresses #/connections, #/mcp and #/people silently collapse onto Overview — including a legacy OAuth failure bounce-back


graycyrus:
## What happens

`#/connections`, `#/mcp` and `#/people` were live top-level nav routes until commit `e9dd8e51` ("group configuration under a Settings section with its own nav") moved them under `#/settings/<page>`. Every one of them now resolves to **Overview**, silently, with no toast and no explanation.

```
#/connections  ->  #/overview   [Company overview]
#/mcp          ->  #/overview   [Company overview]
#/people       ->  #/overview   [Company overview]
```

The codebase already has the mechanism for exactly this — `REWRITE_RETIRED` in `frontend/src/components/app-shell.tsx`, whose own doc comment explains why an address whose view no longer exists must be "sent somewhere real instead of silently collapsing onto `fallback`". It carries rules for `#/tasks` (#1140) and `#/team` (#1141). The three Settings pages never got one.

## Why this is worse than a stale bookmark

The `connect_error` bounce-back handler sits ~20 lines below `REWRITE_RETIRED` in the same file and its comment says:

> Connections is a page of the Settings section now (`#/settings/connections`), so the bounce-back lands there rather than on a top-level view.

So the file knows the address moved. The redirect handler was updated; the router was not.

## Repro

| # | Address entered | Lands on | Expected |
|---|---|---|---|
| 1 | `http://127.0.0.1:8450/login#/connections` | `#/overview` — "Company overview" | `#/settings/connections` |
| 2 | `http://127.0.0.1:8450/login#/connections?connect_error=denied&provider=slack` | `#/overview`, **no toast at all** | `#/settings/connections` + "Slack connection was cancelled." |
| 3 | `http://127.0.0.1:8450/login#/mcp` | `#/overview` | `#/settings/mcp` |
| 4 | `http://127.0.0.1:8450/login#/people` | `#/overview` | `#/settings/people` |

Case 2 is the one that stings: the equivalent **path** form still works —
`http://127.0.0.1:8450/connections?connect_error=denied&provider=slack` correctly lands on `#/settings/connections` and shows "Slack connection was cancelled. You can try again whenever you're ready." — so the same failed OAuth handshake is recoverable or a dead end depending on whether the recorded address used a `?` or a `#?`. An operator cannot tell those apart from a browser history entry.

## Suggested fix

Add to `REWRITE_RETIRED`:

```ts
[...]

Issue #1337, opened 2026-08-20
Parked surfaces (Inbox, Finances, Pages) present as fully-live pages with no entry point and no sign they are parked


graycyrus:
**Type: design / IA coherence** — a judgement call rather than a broken control, filed because the pattern is now applied to four surfaces and has no stated rule. From a read-only design audit against a seeded host.

## The situation

`app-shell.tsx` keeps a set of views out of `NAV` but alive as routes:

```ts
const HIDDEN_VIEWS: View[] = ["feedback", "inbox", "finances", "conversation", "team", "tasks"];
```

Three of these have coherent reasons to be reachable-but-unlisted:

- `feedback` **is** linked, from the sidebar footer. Good — unlisted in `NAV` but discoverable.
- `tasks` and `team` exist only for `#/tasks/<id>` and `#/team/<agentId>`, deep links written by other surfaces. Their parent routes are redirected or unused. Also coherent, and both carry comments explaining it.

`inbox` and `finances` are different, and so is `pages` (which is currently not routable at all — separate issue). Each renders a **complete, first-class-looking page** — full-width title, subtitle, KPI row, cards, empty states written for real operators — that nothing in the product links to and nothing in the product mentions. `#/finances` draws a finance dashboard. `#/inbox` draws a mail client. An operator can only reach either by typing a URL they have no way to learn.

## Why this is worth deciding rather than leaving

**1. The pages actively assert they are live.** Nothing on `#/finances` says "this surface is parked". It says "What your company is earning and spending this month" and shows a wallet balance. An operator who lands there — from an old bookmark, a doc, a support answer, a teammate's link — has no way to know they are looking at something the product has withdrawn. That is a worse state than a 404, because a 404 is honest.

**2. The empty states give instructions into a retired IA.** `#/inbox`'s empty state tells you to go to "the Team page", which no longer exists (filed separately). This is the predictable failure of a parked surface: it stops being maintained but keeps rendering, so its copy drifts against an IA that moved on without it.

**3. The reason for parking is invisible at the point of use.** The rationale lives entirely in source comments (#302: "unmounted from the console — hidden, not retired … Do not delete it as dead code"). That is the right note for a maintainer and no help at all to the person looking at the page.

[...]

Issue #1221, opened 2026-08-20
console: every page has two nested main landmarks, and most have no h1


graycyrus:
Every page of the console ships two `main` landmarks, one nested inside the
other, and most pages ship no `h1` at all. Both are structural, both are on
every view, and both are cheap to fix.

## Measured

Walking the console with the DOM, not by eye:

```
view          <main> count   nested   chain                            <h1>
#/overview    2              1        main[sidebar-inset] > main       []
#/company     2              1        main[sidebar-inset] > main       []
#/chat        2              1        main[sidebar-inset] > main       ["engineering"]
#/approvals   2              1        main[sidebar-inset] > main       []
#/ledgers     2              1        main[sidebar-inset] > main       ["Ledgers"]
#/settings    2              1        main[sidebar-inset] > main       []
```

Playwright's own strict-mode locator refuses `page.locator("main")` on any of
these pages for the same reason — it resolves to two elements — which is how
this was found.

## Why it matters

**Two main landmarks.** `<main>` must not be a descendant of another `<main>`;
the HTML spec allows one visible main per document, and assistive technology
exposes exactly one "main" region for a reason: it is the "skip to content"
target. Landmark navigation now offers two identical destinations on every
screen, and the outer one contains the whole app frame rather than the content.
A screen-reader user picking "main" gets the wrong box first, every time.

**No `h1`.** Overview, Company, Approvals and Settings have no level-1 heading.
Their page titles are `h2` — "Company", "3 things need your approval". Pressing
`1` in a screen reader finds nothing on those pages, and the first heading a
reader lands on is a level 2 with no level 1 above it, which reads as a
document that starts mid-outline.

Chat and Ledgers do have an `h1`, and they disagree about what it is for:
Ledgers uses it for the page ("Ledgers"), Chat uses it for the channel
("engineering"). So the outline also changes shape as you move between views.

## Cause

**The nesting** is `SidebarInset` meeting the shell's own wrapper.
`frontend/src/components/ui/sidebar.tsx:316-329` renders a `<main>`:

```ts
function SidebarInset({ className, ...props }: React.ComponentProps<"main">) {
  return (
    <main data-slot="sidebar-inset" className={cn(…)} {…props} />
```

and `frontend/src/components/app-shell.tsx:1368-1369` puts a second one inside
it:

```tsx
<SidebarInset className="min-h-0 min-w-0">
  <main className="flex min-h-0 flex-1 flex-col overflow-hidden">
```

[...]

Issue #1171, opened 2026-08-19
console: take Pages out of the sidebar nav


graycyrus:
Take **Pages** out of the sidebar navigation.

`frontend/src/components/app-shell.tsx:159` — `{ view: "pages", label: "Pages", icon: AppWindow }`.

## How, and why not simply delete it

Comment the entry out rather than removing the feature. The view itself (`PagesView`, lazily imported at `:111`, rendered at `:1592-1600`) and the `"pages"` route stay intact, so:

- `#/pages` still resolves for anyone with the address or a link to it;
- nothing else that references the view breaks;
- restoring it later is one line, not a revival.

This matches how the console already treats views that exist but are not offered — `feedback`, `inbox`, `memory`, `finances`, `conversation` and `team` are all routable and absent from the nav list. Hiding the entry is the established pattern here, not an exception.

Leave a comment saying it is deliberately hidden, so the next reader does not "fix" the omission by adding it back.

PR #1441, opened 2026-08-20
console: make a view routable by existing, not by having a nav row (#1311)

### LINKED ISSUE #1311: #/pages does not resolve: PagesView is unreachable, contradicting the comment that kept its route alive
**Type: defect.** Found in a read-only design audit against a seeded host (`agentic-software-company`).

## Repro

Signed in, either of:

- Navigate to `http://<host>/#/pages` directly, or
- From inside the console set `window.location.hash = '/pages'`.

**Observed (both):** the URL is rewritten to `#/overview` and the Overview page renders. `PagesView` never mounts.

**Expected:** the Pages surface renders, as the code says it should.

## Where it comes from

`frontend/src/components/app-shell.tsx`. #1172 commented the nav row out and left this note:

```
// Pages is deliberately not offered in the nav. The view and its `#/pages`
// route stay live, so an address or an existing link still resolves — this
// is the same treatment `feedback`, `inbox`, `finances`, `conversation`
// and `team` already get. Do not "fix" the omission by
// adding it back.
// { view: "pages", label: "Pages", icon: AppWindow },
```

But that treatment has two halves, and only one was applied. The other five named views are listed in `HIDDEN_VIEWS`; `pages` is not:

```ts
const HIDDEN_VIEWS: View[] = ["feedback", "inbox", "finances", "conversation", "team", "tasks"];
const VIEWS: View[] = [...NAV.map((i) => i.view), ...HIDDEN_VIEWS];
```

`VIEWS` is the allow-list `useHashView` validates against, and an unrecognised head collapses to the fallback and is canonicalised into the URL:

```ts
if (!(valid as readonly string[]).includes(head)) return [fallback, null];
```

With `pages` in neither `NAV` nor `HIDDEN_VIEWS` it is absent from `VIEWS`, so `#/pages` is an unknown hash. The `{view === "pages" && ...}` block further down `app-shell.tsx` — and the `lazy()` import above it — are dead code that cannot execute.

## Why it matters

An entire feature is unreachable by any means. `PagesView` is not a stub: it carries the sandboxed-iframe host, the per-document capability rotation, and the `oc:graphql` postMessage bridge (`docs/spec/runtime/pages.md`) — a security-sensitive surface with its own spec and its own recent hardening work (#1122, #1221). None of it can be exercised in the product.

It also silently defeats the deliberate design of `nav_visible = false` in `page.toml`, which is documented in `PagesView.tsx` as keeping a page "off the nav, reachable only by direct URL". There is no direct URL any more — the parent route is gone, so no page is reachable, nav-visible or not.

And the empty state tells operators to "Ask the **Page Builder** to design one … and it shows up here" — pointing at a *here* that cannot be opened.

## What good looks like

- Add `"pages"` to `HIDDEN_VIEWS`, which is all the quoted comment intended. That restores `#/pages` without re-adding the nav row it deliberately removed.
- Give the comment a line saying `HIDDEN_VIEWS` membership is what keeps the route live, so the next removal doesn't repeat the split.
- A route test asserting `#/pages` resolves to the pages view and is not rewritten — the same guard `#/tasks/<id>` and `#/team/<agentId>` have comments demanding but which nothing enforces for this view.
- Separately worth deciding: with no nav row and no in-product link anywhere, even a working `#/pages` is reachable only by someone who already knows the address. If Pages is meant to be used, something has to link to it (see the sibling IA issue).

graycyrus:
Closes #1311.

`#/pages` did not resolve. Verified against the current tree before touching anything: `pages` is in the `View` union and has a live `{view === "pages" && …}` render block in `app-shell.tsx`, but it is in neither `NAV` (#1172 commented the row out) nor `HIDDEN_VIEWS` — and `VIEWS`, the allow-list `useHashView` validates against, was `[...NAV.map((i) => i.view), ...HIDDEN_VIEWS]`. So the head was unknown, `useHashView` collapsed it to `overview` and canonicalised the URL, and the entire sandboxed-iframe / `oc:graphql` Pages surface was unreachable by any address.

## The decision: restore, not remove

Two coherent outcomes were on the table — restore reachability, or delete the dead branch and the misleading comment. **Restored**, deliberately:

- #1171 and #1172 are explicit that this was a *nav* removal: "Commented, not deleted — `PagesView` and the `#/pages` route stay intact, so an address or an existing link still resolves." Nobody decided to retire the feature; half a change went missing.
- Pages is not parked code. It has a spec (`docs/spec/runtime/pages.md`), four harness tools (`src/harness/pages_tools.rs`), three always-compiled HTTP routes (`src/server/ops/pages.rs`), a `page_builder` agent in `globals/`, a `build:pages-sdk` target in `frontend/package.json` — and two hardening PRs, #1122 (per-document bridge capability) and #1221 (landmark/heading fixes), **landed after** #1172 removed the nav row. Deleting it would retire a feature that is still being worked on, on the strength of a comment nobody meant that way.

## The fix is structural, not a one-line list edit

Adding `"pages"` to `HIDDEN_VIEWS` fixes this instance and leaves the trap armed: routing was the nav list plus a second, hand-maintained list, so "hide a surface" was a two-part treatment that a reviewer could see half of and believe.

The route table now lives in `frontend/src/lib/console-routes.ts`, where `ROUTABLE` is a `Record<View, true>` — **complete by construction**. A view in the union with no entry is a compile error, and `npm run typecheck` is the first step of every console CI lane:

```
src/lib/console-routes.ts(70,7): error TS2741: Property 'pages' is missing in type
'{ overview: true; … }' but required in type 'Record<View, true>'.
```

(that is a real run, with the `pages` entry commented out). There is no second list to forget any more: a surface the shell can render is a surface an address can reach. `NAV` keeps its job — presentation — and its Pages comment now says what actually keeps the route alive instead of the opposite.

Every word of the old `HIDDEN_VIEWS` prose (why `feedback`, `inbox`, `finances`, `conversation`, `team`, `tasks` have no nav row; #302, #360, #1140, #1141) is carried over onto the corresponding entries. `REWRITE_RETIRED` and `NAV_ALWAYS_PARENT` are untouched.

## Regression coverage

`frontend/test/unit/console-routes.test.ts` (20 tests) imports the **real** table rather than transcribing it — the transcription in `task-route.test.ts` is exactly why this rotted unseen:

- `#/pages` is routable, named as its own test so a failure reads as "Pages is unreachable again";
- every member of `VIEWS`, table-driven: `#/<view>` resolves to itself and the address is not rewritten;
- `#/team/agent-1` still carries its sub-page (an unknown head discards the sub-page too — the failure mode #1140/#1141 warn about);
- `#/nope` still collapses onto Overview and is canonicalised, so widening the allow-list did not widen it into accepting anything.

## Relationship to #1338 and #1337

- **#1338** (retired `#/connections`, `#/mcp`, `#/people` collapsing onto Overview) is a different root cause — addresses whose view no longer exists, needing `REWRITE_RETIRED` rules — and a disjoint route set. This PR does **not** subsume it and deliberately does not touch `REWRITE_RETIRED`, to stay out of its way. What it does contribute is the general guard #1338 asks for at the end: the type-level completeness check makes "a surface the shell renders that no address reaches" impossible to reintroduce.
- **#1337** (parked surfaces have no honest "this is parked" state, and six changes produced four treatments) is the design question of what a parked page should *say*. This PR does not answer it and does not preempt it: it decides only that Pages still *answers*, which is what #1172 intended. `console-routes.ts` says so in as many words and points at #1337 for the rest. If #1337 lands on "parked-and-gone" for some surface, the honest removal is now a single, visible edit in one file.

## Verification

VERIFICATION_PLACEHOLDER

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -53,6 +53,7 @@ import { useLedgerNav } from "@/hooks/use-ledger-nav";
 import type { WorkspaceEvent } from "@/views/WorkspaceView";
 import { useHashView } from "@/hooks/use-hash-view";
 import { BOARD_LEDGER } from "@/lib/board-columns";
+import { VIEWS, type View } from "@/lib/console-routes";
 import { taskIdFromSegment } from "@/lib/task-route";
 import { toast } from "sonner";
 
@@ -112,23 +113,11 @@ const FinancesView = lazy(() =>
 // as the other heavier, less-visited surfaces.
 const PagesView = lazy(() => import("@/views/PagesView").then((m) => ({ default: m.PagesView })));
 
-export type View =
-  | "overview"
-  | "company"
-  | "chat"
-  | "conversation"
-  | "inbox"
-  | "tasks"
-  | "ledgers"
-  | "team"
-  | "workspace"
-  | "memory"
-  | "approvals"
-  | "workflows"
-  | "pages"
-  | "finances"
-  | "settings"
-  | "feedback";
+// The route table lives in `@/lib/console-routes` — a plain module the unit
+// lane can import, and the single place a surface is declared routable (issue
+// #1311). Re-exported because the console has always imported `View` from the
+// shell that renders those views.
+export type { View };
 
 interface NavItem {
   view: View;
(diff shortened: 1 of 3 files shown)


graycyrus:
@coderabbitai review

oxoxDev:
@coderabbitai review

PR #1442, opened 2026-08-20
fix(workspace): say where the open note lives, and reveal it in the tree

### LINKED ISSUE #1371: Workspace: an open note never says where it lives, and the tree does not reveal it
## What I saw

Open a note that lives at `Standards/Engineering/Backend/Rust/API design/Pagination.md` — by clicking a search hit, by following a `[[wikilink]]`, or by landing on `#/workspace/<nodeId>` from the Artifacts tab's "Open in workspace".

The editor header reads:

> **Pagination**  ·  Aug 21, 1:13 AM

That is the entire location information the operator gets. And the explorer tree beside it stays exactly where it was — collapsed, showing the root folders, with `Standards` not expanded and the open note nowhere on screen. The tree does not scroll to the open node and does not expand the chain of folders leading to it.

## Why it matters

A workspace of any real size has repeated names. The seeded company already has three notes called `README`; a real one will have several `Overview.md`, several `Notes.md`, one per area. With no path in the header and no reveal in the tree, the operator reading a note cannot answer "which one is this?" or "what else is next to it?" without hunting for it by hand.

Search already knows the answer — `GET …/workspace/search` returns a `path` on every hit, and `SearchResults.tsx` renders it. That path is thrown away the moment the hit is opened, which is the moment it becomes most useful.

## What I'd expect

1. **A breadcrumb in the note header**: the parent path, ellipsised at the *start* if it does not fit (`…/Rust/API design`), each segment clickable to expand-and-scroll the tree to that folder.
2. **Reveal on open**: opening a node expands every ancestor and scrolls the tree row into view. This should fire for every route into a note — tree click, search hit, wikilink, `initialNodeId`, and the SSE "opened" event.

`pathOf` already exists in `src/lib/workspace.ts`, and `Tree` already takes an `expanded: Set<string>`, so both halves have the pieces.

## Where

- `frontend/src/views/WorkspaceView.tsx` — the note-pane header (~line 1420) and the `open()` callback
- `frontend/src/lib/workspace.ts` — `pathOf`

Found during a design audit of the Workspace surface at 1440px and 1100px, light and dark.

graycyrus:
Closes #1371.

## What was wrong

The note-pane header read `Pagination · Aug 21, 1:13 AM` and that was the entire location information the operator got. The explorer beside it stayed exactly where it was — `Standards` collapsed, the open note nowhere on screen.

The ancestor expansion in `open()` looked like it should have handled this, but it was dead on the route that needs it most. It read `nodes` out of its own closure, and on the deep-link route (`#/workspace/<id>`, which the Artifacts tab's "Open in workspace" uses) that closure runs **before the tree has loaded** — so `pathOf` walked an empty array and expanded nothing. Measured against a live host:

```
deep note reveal: {"rowCount":32,"activeLabel":null,"activeInView":null,"scrollTop":0}
```

32 rows rendered, and the row for the note the operator had just been sent to was not one of them.

## What changed

**Reveal** is now an effect keyed on `[openId, nodes]`, so the pass that runs once the tree arrives is the one that does the work. Expanding is idempotent, so re-running costs nothing; the effect bails early when every ancestor is already open, so it does not churn state.

**Scrolling** belongs to the row, not the pane — only the row knows when it exists, since the ancestors expand on one render and the row mounts on the next. `block: "nearest"`, so a row already on screen (the ordinary tree click) does not move at all. `revealId` is a one-shot event cleared by the row that consumed it, so an operator who scrolls the tree away from the open note is not yanked back on the next render.

**Breadcrumb**: `breadcrumbOf` elides the **middle**, keeping the root and the last two folders — `Standards / … / Rust / API design`. Truncating a path *string* (which is what `truncate` on a single span does) ellipsises the tail, so every note under `Standards/Engineering/…` renders an identical prefix and the discriminating end is exactly what gets thrown away. Each crumb is clickable and reveals that folder.

The note's own name is deliberately not in the trail — it is the heading immediately beside it.

## Verified

Against a real host, deep link and search-hit routes, 1440px and 1100px, light and dark:

```
deep-link reveal: {"rowCount":41,"activeLabel":"Pagination","activeInView":true,"scrollTop":314}
breadcrumb:       Standards/…/Rust/API design/

search-hit reveal:{"rowCount":48,"activeLabel":"Refund after chargeback","activeInView":true,"scrollTop":26}
breadcrumb:       Product/Billing/Edge cases/
```

## Commands run

- `npx vitest run` — 168 files, 1605 tests, all passing (9 new)
- Browser-verified light and dark, 1440px and 1100px, against a live host

Found during a design audit of the Workspace surface.

### DIFF
--- a/frontend/src/lib/workspace.ts
+++ b/frontend/src/lib/workspace.ts
@@ -94,6 +94,50 @@ export function pathOf(nodes: FsNode[], id: string | null): FsNode[] {
   return path;
 }
 
+/**
+ * One step of a note's location, for the header breadcrumb (issue #1371).
+ *
+ * `null` is the elided middle — a crumb the trail could not fit, rendered as an
+ * ellipsis rather than dropped, so the operator can see that the path is longer
+ * than what is shown instead of reading a shortened path as the whole truth.
+ */
+export type Crumb = FsNode | null;
+
+/**
+ * The folders a note sits inside, shortened to at most `max` of them.
+ *
+ * The note itself is **not** in the trail: its name is already the heading beside
+ * it, and repeating it would spend the widest crumb saying what the operator is
+ * looking straight at.
+ *
+ * When the trail is too long, the **root and the last two folders** survive and
+ * the middle collapses. That split is the useful one: the root says which part
+ * of the company this belongs to, and the last two say what it sits next to.
+ * Truncating the string instead — which is what `truncate` on a single span did
+ * — ellipsises the *tail*, so every note under `Standards/Engineering/…` renders
+ * the identical prefix and the discriminating end is exactly what is thrown away.
+ */
+export function breadcrumbOf(nodes: FsNode[], id: string | null, max = 3): Crumb[] {
+  const folders = pathOf(nodes, id).slice(0, -1);
+  if (folders.length <= max) return folders;
+  // `max - 2` leading crumbs, the ellipsis, then the final two.
+  return [...folders.slice(0, Math.max(max - 2, 1)), null, ...folders.slice(-2)];
+}
+
+/**
+ * The ids of every folder a node is nested in — what has to be expanded for the
+ * node's own row to exist in the tree (issue #1371).
+ *
+ * Excludes the node itself even when it is a folder: revealing a folder means
+ * showing its row, not opening it.
+ */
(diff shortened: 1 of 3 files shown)


graycyrus:
@coderabbitai review

PR #1444, opened 2026-08-20
perf(mongodb): create indexes concurrently instead of one round trip at a time

CodeGhost21:
Second of the cold-start reductions. A hosted tenant's wake was measured at ~8s, and **~3.5s of it was here**.

## The measurement

`ensure_indexes` issued **44 `create_index` calls sequentially**, each awaited before the next. Against a local MongoDB that is free. Against a remote one it is 44 round trips, and the round trips are the entire cost.

Staging runs MongoDB Atlas over the internet, and the tenant's own boot log shows exactly where the time goes:

```
20:07:01.018  agent journal, keyring
20:07:04.492  storage backend: Mongodb     ← 3.47s
20:07:04.759  listening on 0.0.0.0:8080
```

That is paid on **every boot**, re-creating indexes that already exist — and every scale-to-zero wake pays it again.

## The change

Issued concurrently, bounded at 10.

Safe because index creation is **independent per index and idempotent**: no ordering between any two, nothing reads what another writes, and re-creating an existing index is a server-side no-op. Bounded rather than unbounded because the driver's connection pool is 10 by default — more in flight would queue inside the pool anyway while making the burst look less deliberate than it is.

The nine indexes that lived after the loop are folded into the same array so they parallelise too, with every issue-referencing comment kept attached to its own entry.

## Verified as a parity refactor, not assumed

**This matters more than usual here.** The MongoDB conformance tests are skipped unless `OPENCOMPANY_TEST_MONGODB_URI` points at a live server — so the 3261 passing tests **did not exercise this code at all**. A green run means very little on its own.

Against a real MongoDB in Docker:

- 43 conformance tests pass
- a temporary probe dumped the resulting index set before and after, sorted by collection and name

```
IDENTICAL — 74 index rows, no difference
```

Same indexes, same names. Different only in how they are issued.

## What is arithmetic rather than measured

The expected saving is 44 sequential round trips collapsing to about five batches. **That is reasoning, not a number I can show**: locally there is no latency to remove, so the improvement is invisible here. The real figure will come from staging, where the 3.5s was measured in the first place, and I will post it once this deploys.

## Verification

```
cargo test --features mongodb,sqlite    3261 + 14 + 1 passed, 0 failed
cargo clippy --all-targets --features mongodb,sqlite -- -D warnings    exit 0
cargo fmt --check (crate-wide)          clean
```

## Context

One of three contributors to the ~8s:

| Phase | Time | Status |
|---|---|---|
| k8s scheduling + container start | ~3.3s | untouched |
| **MongoDB index creation** | **~3.5s** | **this PR** |
| readiness probe lag | ~1.1s | opencompany-microservice#47 |
| rest of app boot | ~0.3s | — |

Sub-1s from a parked tenant stays unreachable regardless — Kubernetes spends ~3.3s before the tenant's code runs. Sub-second wakes remain a question about `IdleResident` and the park timeout rather than about boot speed.

## Summary by CodeRabbit

* **Performance**
  * MongoDB index setup now runs concurrently, improving startup efficiency while limiting resource usage.
* **Reliability**
  * Index creation continues to report errors after all configured indexes have been processed.

### DIFF
--- a/src/store/mongodb.rs
+++ b/src/store/mongodb.rs
@@ -356,76 +356,85 @@ impl MongoStore {
                 nonunique(doc! {"company_id": 1, "workflow_id": 1, "created_ms": -1}),
             ),
         ];
-        for (name, index) in plans {
-            self.collection(name)
-                .create_index(index)
-                .await
-                .map_err(mongo_err)?;
-        }
-        self.collection("owners")
-            .create_index(unique(doc! {"company_id": 1}))
-            .await
-            .map_err(mongo_err)?;
-        // Issue #241: the cross-replica arbiter. This unique compound index is
-        // what turns two replicas racing one schedule minute into one winning
-        // `insert_one` and one `E11000` — the case that actually matters, since
-        // hosted replicas share the tenant database. Created outside the array
-        // above so adding it does not disturb that array's fixed length.
-        self.collection("schedule_fires")
-            .create_index(unique(
-                doc! {"company_id": 1, "schedule_id": 1, "scheduled_for": 1},
-            ))
-            .await
-            .map_err(mongo_err)?;
-        // Issue #596: one durable per-node output snapshot per run. The unique
-        // `(company_id, run_id)` index makes the upsert last-write-wins per run;
-        // the recency index backs the newest-N prune. Created outside the fixed
-        // array above so adding it does not disturb that array's length.
-        self.collection("run_outputs")
-            .create_index(unique(doc! {"company_id": 1, "run_id": 1}))
-            .await
-            .map_err(mongo_err)?;
-        self.collection("run_outputs")
-            .create_index(nonunique(doc! {"company_id": 1, "at_ms": -1}))
-            .await
-            .map_err(mongo_err)?;
-        // Issue #726: the runtime journal. `(company_id, seq)` is unique because
-        // two records sharing a sequence would order arbitrarily on replay, and
-        // an at-most-once key replayed before the resolution that consumed it is
-        // the failure the journal exists to prevent. The import receipt is one
-        // document per company. Created outside the fixed array above so adding
-        // them does not disturb its length.
-        self.collection(JOURNAL)
-            .create_index(unique(doc! {"company_id": 1, "seq": 1}))
-            .await
-            .map_err(mongo_err)?;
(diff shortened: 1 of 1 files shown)


CodeGhost21:
Verified and fixed in `3dd3476f`. The finding is valid, though for a more specific reason than stated — worth spelling out because it changes what the fix is protecting against.

## It is not a regression

`try_collect` short-circuits, which is precisely what the sequential loop it replaced did:

```rust
for (name, index) in plans {
    self.collection(name).create_index(index).await.map_err(mongo_err)?;   // also abandons the rest
}
```

So on "does a failure stop later indexes being created", the behaviour was unchanged.

## But the concurrency introduced a hazard the old code could not have

Sequentially there was **never an operation in flight to abandon**. Concurrently, a short-circuit drops up to `INDEX_CONCURRENCY - 1` futures mid-await — and a driver operation cancelled after its request is sent but before its reply is read leaves a connection the pool cannot safely reuse.

That hazard belongs to this change rather than being inherited, which is what makes the fix worth doing. Letting them all land is also simply better: a transient failure on one index no longer decides whether the other forty-three exist.

The error now also names the failing collection, and reports the count when more than one fails — one failing index is a different problem from every index failing, and the first error alone cannot distinguish them.

## The test needed two attempts

Worth recording, because the first version was worthless in a way that would have passed review.

It tried to seed the conflicting index *before* `ensure_indexes` had run — impossible, since `store()` connects and connecting creates the indexes. It failed loudly, which was lucky. The real problem was quieter: **its assertion checked for indexes the first connect had already created**, so it would have passed no matter what the pipeline did.

The version that landed drops a known index first, induces the failure by replacing the unique `owners` index with a conflicting non-unique one, then asserts the dropped index came back — which only holds if the run continued past the failure.

Confirmed by reintroducing the short-circuit:

```
```

## Verification

Run against a real MongoDB rather than the skipped lane, since these tests are no-ops without `OPENCOMPANY_TEST_MONGODB_URI`:

```
cargo test --features mongodb,sqlite    3262 + 14 + 1 passed, 0 failed
cargo clippy --all-targets --features mongodb,sqlite -- -D warnings    exit 0
cargo fmt --check (crate-wide)          clean
```

Issue #1300, opened 2026-08-20
ContextStore: multi-label-per-address semantics diverge across backends, and shared addresses are a forget-griefing vector


YellowSnnowmann:
Spun out of the #1290 review (its shared-address minor). Two related defects in how content-addressed context chunks handle byte-identical bodies under different labels:

**1. Backend divergence on the second label.** fs appends an index entry per `put` — one address can carry N labels. sqlite (`INSERT OR IGNORE`) and mongodb (`$setOnInsert`) and the provider facade are first-write-wins on `(company, addr)` — the second `put` of identical content under a different label returns a success receipt (the addr) for a row that never landed under that label. A caller listing by their label finds nothing on those backends, everything on fs. The conformance suite uses distinct bodies everywhere, so nothing pins either behavior.

**2. Shared addresses make memories permanently un-forgettable (fs).** `memory_forget` and the fact-mirror reap both refuse address-level deletion when any other label shares the address — correct fail-safe, but on fs any agent can make another agent's memory (or an operator-fact mirror) permanently un-deletable from those surfaces by storing byte-identical text. No escape hatch exists below the operator console.

Options to evaluate: label-scoped delete on the port (removes one index entry; blob refcounted — fs is close already, sqlite/mongo need a labels column/array); salting the address with the label (kills dedup, breaks existing addrs — probably no); or documenting first-write-wins as the contract and making fs conform (loses fs's multi-label listing, changes existing data). Whichever way, the conformance suite needs identical-body-two-label cases so the chosen semantics are pinned on every backend.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

**Addendum (from the #1290 CodeRabbit round):** both shared-address guards (`memory_forget`, `reap_fact_mirror`) check a snapshot and then delete — a byte-identical write landing in between loses its row (TOCTOU). The label-scoped delete option closes this by construction (deleting one label entry cannot orphan another's); any other chosen semantics needs a conditional delete on the port. The interleaving regression tests belong with whichever lands.

PR #1496, opened 2026-08-21
fix(memory): unfreeze Brain list, guard multibyte slices, serve bodies in one pass (#1488)

### LINKED ISSUE #1488: Memory audit follow-ups: Brain view freezes on the oldest 500 chunks, agent-reachable snippet panics, shared-address guard is fs-only, and the tinymemory pin predates the keyed-CRUD line
## Summary

Product-side findings from the eight-lane memory-stack audit (run after tinymemory #69/#71 and oc #1290 merged). Every major was verified in source at the cited lines on the merged head. Companion adapter/engine issue: tinyhumansai/tinymemory#75.

## Majors

- [ ] **The vendored tinymemory pin predates everything the audit just improved** — `vendor/openhuman/vendor/tinymemory` sits at `38a34d2`, before the #68 error taxonomy and the whole #69/#71 keyed-CRUD line. Provably in the compiled product, every keyed remote get is still `dialect.entries().find(...)` — a full account enumeration per get — with no read retries and the shallow boot health probe. Now that the PRs are merged this is unblocked: bump openhuman's tinymemory pin, then oc's openhuman pin (the oh#5608→oc#1248 shape), and refresh `vendor/tinycortex`'s stale `tinymemory-api` rev (4549cda, pre-§A4) in the same sweep. Note tinymemory's post-#73 `crates/` layout when bumping.
- [ ] **Brain tab freezes on the OLDEST 500 chunks and never shows a new memory again** — `src/server/ops/memory.rs:334-373`. `list_facts` takes the first `MAX_CONTEXT_ENTRIES = 500` metas with no sort, and every backend lists oldest-first (fs = index append order; the provider facade explicitly sorts `stored_at_millis` ascending). An active company crosses 500 chunks in normal use; after that GET /memory silently drops every new task outcome and agent memory while GET /memory/stats keeps reporting a growing count. Fix: sort by `stored_at_millis` **descending** before the cap.
- [ ] **The same route re-fetches every chunk body one-by-one after a list that already carried the bodies** — up to 500 sequential `peek` calls per console load; on the provider overlay each peek is a fresh provider `get` (at the current pin: a full account enumeration *each*). Fix: serve bodies from the same enumeration (a with-bodies list variant or a bulk read for the console route).
- [ ] **Byte-offset slicing panics on multibyte content in fs/sqlite/mongo search snippets and ranged peek — agent-reachable** — `src/store/fs.rs:1387-1392`, `sqlite.rs:1069`, `mongodb.rs:1007`. Snippets slice `body[pos-24 .. pos+q+24]` on raw byte offsets; a multibyte character within 24 bytes of the match lands mid-codepoint and panics ("byte index is not a char boundary"). Ranged peek clamps to length but not to char boundaries in all three backends. The provider facade and the tinycortex store both defend this exact case (`slice_on_char_boundaries` — its doc says a mid-character bound is "reachable with any non-ASCII body"); the three base backends don't. `memory_recall` routes agent queries straight into `context.search` (`src/harness/built_in/memory_tools.rs:293`), so an agent recalling near any non-ASCII memory can panic the call. Fix: reuse the boundary-widening helpers at the four sites.
- [ ] **The forget guard's shared-address rule only exists on the fs backend** — `src/harness/built_in/memory_tools.rs:380` reasons over "every label at this address", but only `FsContextStore` accumulates multiple labels per address. sqlite (`PRIMARY KEY (company_id, addr)` + `INSERT OR IGNORE`, `sqlite.rs:98/:958`), mongo (`$setOnInsert`, `mongodb.rs:906-916`), and the provider facade (explicit first-write-wins, `facades.rs:372-407`) all silently drop the second label. Concrete cross-agent failure: agent B stores byte-identical text and is told "Remembered" while nothing landed; agent A later forgets the address, passes the guard (only its own label visible), and deletes what B believes exists. Found independently by two audit lanes. Fix: label side-table (sqlite) / `$addToSet` labels array (mongo) / label set (provider) so identical content accumulates labels like fs — or declare single-label semantics everywhere and make fs + the guard match. Belongs with #1300's label-scoped-delete design.

## Minors

- `memory_forget` / `reap_fact_mirror` enumerate the entire company context per call — at the current pin, that is an account enumeration per forget on hosted engines.
- fs `put` rewrites an already-indexed blob in place with a plain O_TRUNC write while `peek`/`search` readers may be mid-read.
- fs `delete` reports failure after the index rows are already gone when blob removal errors, stranding the caller with a half-applied delete reported as none.
- The scratch firewall and per-agent/per-desk partitions are minted but never wired — a dead seam presented as live in the overlay's own docs.
- Stale comment: the overlay doc still claims the inbound taint port has "no production writer yet" while `cycle.rs` now routes external cycles through it (shipped in #1290).

## Related

- tinyhumansai/tinymemory#75 — the adapter/embedded-engine half of this audit (blocker: Supermemory re-store 400s).
- #1300 — label-scoped delete; the shared-address finding above should fold into its design.

YellowSnnowmann:
## Summary

Three in-repo majors from the memory-stack audit (#1488). Every fix ships with a biting test.

- **Brain tab unfrozen.** `list_facts` (`src/server/ops/memory.rs`) took the first `MAX_CONTEXT_ENTRIES = 500` metas with no sort, and every backend lists oldest-first — so past 500 chunks GET `/memory` silently dropped every new memory while `/memory/stats` kept counting up. Now sorts by `stored_at_millis` **descending** before the cap (extracted as `newest_context_chunks` so it is unit-testable).
- **Multibyte slice panic closed (agent-reachable).** Search snippets and ranged `peek` in `src/store/{fs,sqlite,mongodb}.rs` sliced on raw byte offsets, panicking mid-codepoint ("byte index is not a char boundary"); `memory_recall` routes agent queries straight into `context.search`, so an agent recalling near any non-ASCII memory could panic the call. All sites now go through the shared `slice_on_char_boundaries` / `search_snippet` helpers in `src/store/mod.rs`.
- **Bodies served in one pass.** The console route re-fetched every chunk body one-by-one after a list that already carried them (up to 500 sequential `peek`s; on the provider overlay each a fresh `get`). A new `ContextStore::list_with_bodies` (default = list+peek; fs/sqlite/mongodb/provider-overlay override to read bodies in one enumeration) is wired into the console route.

A cross-backend `assert_context_multibyte_and_bodies` conformance assertion covers the last two across fs/sqlite/mongodb.

## API Or Behavior Changes

Internal store/route behavior only; the `/memory` response shape is unchanged. New `ContextStore::list_with_bodies` port method (defaulted, so existing implementors compile). No manifest/wire changes.

**Deferred (explicitly not in this PR):** the vendored tinymemory pin bump (cross-repo submodule dance) and the forget-guard shared-address / label side-table change (belongs with #1300's design).

## Tests

Biting revert-checks confirmed: reverting the sort fails the cap test with "the stalest chunk must be dropped, not frozen in"; reverting the char-boundary helper panics the conformance test `inside 'é'`.

- [x] `cargo clippy --all-targets -- -D warnings` (default and `openhuman,tinycortex`)
- [x] `cargo test` (default `3206 passed`; sqlite lane `44`; openhuman store `175` / memory-ops `15` / write_test `133` — 0 failures)

## Documentation

Inline docs on the new port method and helpers; no spec/module doc restructure needed.

Closes #1488

### DIFF
--- a/src/ports/context.rs
+++ b/src/ports/context.rs
@@ -5,7 +5,7 @@ use std::ops::Range;
 use async_trait::async_trait;
 
 use crate::Result;
-use crate::ports::types::{ChunkAddr, ChunkHit, ChunkMeta, CompanyId, ContextChunk};
+use crate::ports::types::{ChunkAddr, ChunkHit, ChunkMeta, ChunkWithBody, CompanyId, ContextChunk};
 
 /// The RLM environment: addressable context chunks. Mirrors Medulla's
 /// `ContextStore` port.
@@ -15,6 +15,24 @@ pub trait ContextStore: Send + Sync {
     async fn put(&self, id: &CompanyId, chunk: ContextChunk) -> Result<ChunkAddr>;
     /// Lists chunk metadata under `prefix`.
     async fn list(&self, id: &CompanyId, prefix: &str) -> Result<Vec<ChunkMeta>>;
+    /// Lists chunk metadata under `prefix`, each paired with its body.
+    ///
+    /// For a caller that needs every body — the console's Brain list — this is
+    /// one enumeration rather than a `list` plus one `peek` per row. The default
+    /// preserves exactly that shape (`list` then a `peek` each) so every existing
+    /// backend keeps working unchanged; the durable backends override it to read
+    /// bodies in the same pass. A read fault surfaces rather than being laundered
+    /// into a blank body — the durable overrides read each body in the row that
+    /// already carries it, so there is no separate per-row read to lose.
+    async fn list_with_bodies(&self, id: &CompanyId, prefix: &str) -> Result<Vec<ChunkWithBody>> {
+        let metas = self.list(id, prefix).await?;
+        let mut out = Vec::with_capacity(metas.len());
+        for meta in metas {
+            let body = self.peek(id, &meta.addr, None).await?;
+            out.push(ChunkWithBody { meta, body });
+        }
+        Ok(out)
+    }
     /// Reads a chunk (optionally a byte range) as text.
     async fn peek(
         &self,
--- a/src/ports/types.rs
+++ b/src/ports/types.rs
@@ -2020,6 +2020,21 @@ pub struct ChunkMeta {
     pub stored_at_millis: u64,
 }
 
+/// A context chunk's metadata paired with its body, returned by
+/// [`ContextStore::list_with_bodies`](crate::ports::ContextStore::list_with_bodies).
+///
+/// Lets a caller that needs every body under a prefix — the console's Brain
+/// list is the one such caller — get them in a single enumeration instead of a
+/// `list` followed by one `peek` per row, which on the provider overlay was a
+/// full account read each.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ChunkWithBody {
+    /// The chunk's metadata (address, label, length, store time).
+    pub meta: ChunkMeta,
+    /// The chunk's full body.
+    pub body: String,
+}
+
 /// A single ledger movement.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct LedgerEntry {
--- a/src/server/ops/memory.rs
+++ b/src/server/ops/memory.rs
@@ -29,7 +29,7 @@ use serde::{Deserialize, Serialize};
 use crate::AppState;
 use crate::error::OpenCompanyError;
 use crate::ports::facts::{FactKind, FactRecord};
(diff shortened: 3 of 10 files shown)


YellowSnnowmann:
Closing as a duplicate — superseded by #1489, which covers the same three #1488 majors (newest-first sort+cap, one bulk body read, multibyte snippet panics) and more: it also fixes a fourth tinycortex slice site, an origin-bucketing render-order bug, and fs delete/put honesty. #1489 should carry #1488.

PR #1531, opened 2026-08-21
fix(console): centre the sign-in card below the header (#1332)

### LINKED ISSUE #1332: Login: the sign-in card is not vertically centred — justify-center is a no-op on a non-flex parent
**Type: defect** (intent is stated in the markup and not achieved). Found in a read-only design audit of `frontend/src/views/Login.tsx` at 1440x900 and 1100x900, light and dark.

## Repro

Open the host signed out. At 1440x900 the card's bottom edge sits around y=360 and the remaining ~540px of viewport is empty. The composition is top-heavy: a 64px header bar, a small card immediately under it, then more than half the screen of blank background.

## Where it comes from

```tsx
<div className="min-h-svh bg-background">
  <header className="flex items-center justify-between border-b px-6 py-4"> … </header>
  <main className="mx-auto flex w-full max-w-md flex-col justify-center px-6 py-16">
```

`justify-center` on `main` distributes free space along its own main axis — but `main`'s parent is a plain block `div`, so `main` is only as tall as its content and there is no free space to distribute. The property is inert. The `min-h-svh` that would have created the space is on the parent, and the parent is not a flex container.

The `justify-center` is unambiguous evidence the intended design was a vertically centred card; the layout just doesn't produce one.

## Why it matters

This is the only screen every user of a hosted company sees before anything else, and the only one seen by people who never get in. A sign-in card pinned to the top of a mostly-empty page reads as an unfinished page rather than a composed one — the standard treatment, and the one the markup is reaching for, is optically centred in the space below the header.

At 1100x900 the same gap is present. It is most visible on short laptop viewports, where the empty region is still the majority of the page.

## What good looks like

Make the parent the flex column that owns the viewport height, so the existing `justify-center` starts working:

```tsx
<div className="flex min-h-svh flex-col bg-background">
  <header className="shrink-0 …">
  <main className="mx-auto flex w-full max-w-md flex-1 flex-col justify-center px-6 py-16">
```

Two things to keep while doing it:

- Optical centring, not geometric — a sign-in card usually wants to sit slightly above true centre. `justify-center` with the existing `py-16` gets close; check it against the tallest variant of this screen (hub provider buttons + terms + divider + form + password toggle), which must still scroll rather than clip on a short viewport.
- The screen has four modes (`email`, `wallet`, `none`, plus the "Check your email" state) of quite different heights. Whatever lands should be checked in all four, since the centring only reads as deliberate if it holds as the card grows.

CodeGhost21:
Fixes #1332.

## Root cause

The intent and the mechanism were on two different elements, and neither reached the other:

```tsx
<div className="min-h-svh bg-background">          // owns the viewport height, but is a block box
  <header …>
  <main className="… flex flex-col justify-center …">   // is the flex container, but is content-sized
```

`justify-center` distributes free space along its **own** main axis. `main` is a block-level flex item of a plain block parent, so its height is exactly its content height — free space is zero and the property is inert. The `min-h-svh` that would have created that space sits on a parent that is not a flex container, so nothing hands it down.

Measured on a live host at 1440x900 before the fix: `main` is **402px** tall against the **835px** the header leaves over, so the card ends around y=465 with the rest of the page empty below it — the composition in the report.

Not a regression: `git blame` puts both lines in the commit that introduced the view. It has never worked; it simply never failed anything, because the markup *reads* as centred, so neither review nor any assertion on class names would catch it.

## Change

`frontend/src/views/Login.tsx` — three classes, so the existing `justify-center` starts working, plus a comment recording why it was inert:

```tsx
<div className="flex min-h-svh flex-col bg-background">
  <header className="flex shrink-0 items-center …">
  <main className="mx-auto flex w-full max-w-md flex-1 flex-col justify-center px-6 py-16">
```

`flex-1` cannot clip the taller variants: a flex item's `min-height: auto` floors it at its content height, so on a short viewport `main` grows past its share, the page grows past `min-h-svh`, and the document scrolls.

One note on the issue's framing: centring in the region below the header puts the composition ~32px *below* the viewport's true centre, not above it — the header pushes it down. That is the treatment the issue prescribes and it reads well on screen; the offset is documented in the test rather than chased with padding.

## Tests

`frontend/test/e2e/login-centred.spec.ts` (new, 4 tests) — geometric, because the markup read as centred throughout and only a measured layout tells a stated intent from an achieved one. Each test signs itself out; the suite's shared storage state is an authenticated admin, who never sees this screen.

- centred below the header at 1440x900 and 1100x900 — `main` fills the leftover height exactly, and the content's midpoint lands on the region's midpoint (±2px)
- centring holds across the heights this screen changes through: magic link, password, "Check your email"
- at 1440x360 the page scrolls rather than clipping, and nothing lands above the header where scrolling could not reach it

Confirmed non-vacuous: reverting the three classes fails the suite with `main` at 402 vs 835 and the midpoint off by 216px.

`wallet` and `none` are the same `main` with a different card inside it, so they are covered by the geometry rather than by a host mode the harness cannot stand up.

## Commands run

- `npm run typecheck` — clean
- `npx tsc -p tsconfig.e2e.json` — clean
- `npx playwright test login-centred.spec.ts` — 4 passed, against a live host serving this branch's console bundle

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sign-in page layout to keep content vertically centered across viewport sizes.
  * Ensured taller sign-in content remains accessible and scrollable on shorter screens.

* **Tests**
  * Added automated coverage for sign-in centering, varying viewport widths, card states, and short-screen scrolling.

### DIFF
--- a/frontend/src/views/Login.tsx
+++ b/frontend/src/views/Login.tsx
@@ -263,8 +263,24 @@ export function Login({
   }
 
   return (
-    <div className="min-h-svh bg-background">
-      <header className="flex items-center justify-between border-b px-6 py-4">
+    /*
+      The box that owns the viewport height has to be the flex container too.
+
+      `justify-center` distributes free space along its *own* main axis, so on a
+      `main` whose parent is a plain block it is inert: that `main` is exactly as
+      tall as its content and has no free space to distribute. The height lived
+      on this div and the centring lived on `main`, and neither reached the
+      other — the card sat pinned under the header with half the viewport empty
+      below it (#1332). `flex flex-col` here plus `flex-1` there hands `main`
+      the height the header left over, which is what the `justify-center` below
+      was always reaching for.
+
+      `flex-1` cannot clip the taller variants: a flex item's `min-height: auto`
+      floors it at its content height, so on a short viewport `main` grows past
+      its share, the page grows past `min-h-svh`, and the document scrolls.
+    */
+    <div className="flex min-h-svh flex-col bg-background">
+      <header className="flex shrink-0 items-center justify-between border-b px-6 py-4">
         <div className="flex items-center gap-2">
           <div className="flex size-7 items-center justify-center rounded-md bg-primary text-primary-foreground">
             <Building2 className="size-4" />
@@ -274,7 +290,7 @@ export function Login({
         <ThemeToggle />
       </header>
 
-      <main className="mx-auto flex w-full max-w-md flex-col justify-center px-6 py-16">
+      <main className="mx-auto flex w-full max-w-md flex-1 flex-col justify-center px-6 py-16">
         {/*
           The refusal, above the heading, where the eye lands first.
 
--- a/frontend/test/e2e/login-centred.spec.ts
+++ b/frontend/test/e2e/login-centred.spec.ts
@@ -1,0 +1,159 @@
+import { expect, test, type Browser, type Page } from "@playwright/test";
+
+/**
+ * Regression proof for #1332 — the sign-in composition must sit in the middle
+ * of the space below the header, not pinned under it.
+ *
+ * `main` already said `justify-center`, and had since the view was written. It
+ * was inert: `justify-center` distributes free space along its own main axis,
+ * and `main`'s parent was a plain block box, so `main` was exactly as tall as
+ * its content and had no free space to distribute. Measured at 1440x900 before
+ * the fix, `main` was 402px tall against the 835px the header left over — the
+ * card ended around y=465 with the remaining half of the page empty below it.
+ *
+ * The assertions are geometric on purpose. The markup *read* as centred
+ * throughout, which is why this survived review and would survive any check on
+ * class names: only a real layout distinguishes a stated intent from an
+ * achieved one. Anything that re-breaks the chain — the height moved back onto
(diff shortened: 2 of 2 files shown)


CodeGhost21:
Rebased onto current `main` and addressed the review.

**The red `Console` check was not from this branch.** It was the markdown 500-line cap failing on `docs/spec/runtime/desktop.md` (524 lines at my base commit, `650a2e42`) — a file this PR does not touch. `d97b14fc` has since split it to 446, so rebasing onto `7912b7cf` clears it; `./scripts/ci/assert-md-line-cap.sh` and `./scripts/ci/assert-design-tokens.sh` both pass locally on the rebased branch.

**@coderabbitai's nitpick — `wallet` and `none` geometry — is now covered** (`3ce61e1c`). No host renders more than one mode: `auth/config` is what picks between them and a host answers one way for its whole life. So the spec stubs that answer, which is the only input that selects a mode, and lets the console build the real card for it — the layout itself is never faked. `wallet` lands on its no-wallet-installed variant, headless Chromium having no wallet to find, which is the shortest card this screen has and so the one with the most free space to get wrong.

Both new tests fail without the layout fix, with `main` at 320px and 308px against the 835px the header leaves over. Suite is 6 tests, all passing against a live host.

Issue #1524, opened 2026-08-21
Port the memory seam onto the loadable TinyMemory module


YellowSnnowmann:
> ## Status: superseded — the tree this plan is written against no longer exists
>
> #1568 (merged 23 Aug) removed the embedded TinyCortex memory backend from
> opencompany. `MemoryMode` now carries a single variant (`Remote`);
> `MemoryBackend` is `Store | Remote | Null`. `tinymemory-embedded`, the
> `namespace` driver and the `tinycortex` feature are gone, along with the tests
> that rode them.
>
> That removes both of this plan's load-bearing premises:
>
> - **Where the driver was to be routed.** Wave 1 PR 2 bound the module inside the
>   `MemoryMode::Embedded` arm of `open_driver` on `driver_id == "module"`, and
>   the design turned on adding no new mode and no new backend variant. That arm
>   is gone, so the module can only be a new variant — which is a different
>   design, not a rebase.
> - **What made it non-breaking.** The no-breaking-changes table rests on
>   *"`tinymemory-embedded` is never deleted — removing it would kill the
>   regression net."* It was deleted.
>
> **And none of it moved the compile time**, which was the point of the
> programme. On `main` today:
>
> ```
> $ cargo tree --offline -e normal --no-default-features \
>     --features mongodb,openhuman,tinymemory -i tinymemory-core
> tinymemory-core
> ├── openhuman
> │   └── opencompany
> └── tinymemory-tinycortex
>     └── openhuman
> ```
>
> `TENANT_FEATURES` ships `openhuman` (`deploy-staging.yml:75`), and
> `vendor/openhuman/Cargo.toml` declares the engine crates without
> `optional = true` — so every tenant image still compiles the engine, exactly as
> the **Goal, split honestly** table below predicted. #1568's validation ran
> `cargo tree --features tinymemory -i tinycortex` *without* `openhuman`, which is
> why it read clean.
>
> The whole remaining win therefore lives in **openhuman#5560**. opencompany's
> only surviving share of it is recorded in the closing comment.
>
> Everything below the line is the original plan, kept unedited. Its survey of
> tinybus's loader constraints, the module's process-singleton behaviour, the 5s
> `MODULE_INIT_DEADLINE`, the `workspace_dir` trap and the answered gates
> (Q1–Q4) are still accurate about **tinymemory and tinybus** — only
> opencompany's side of it moved.
>
> ---

## Goal, split honestly

[...]

PR #1551, opened 2026-08-21
chore(deploy): compile the TinyMemory module driver into tenant images (#1524 PR 3, build half)

YellowSnnowmann:
## Summary

The build half of #1524's PR 3: `tinymemory-module` joins `TENANT_FEATURES`, so staging tenant images carry the loadable-module driver — **inert** until a tenant's env sets both `OPENCOMPANY_MEMORY=embedded` and `OPENCOMPANY_MEMORY_DRIVER=module` (the driver knob alone does nothing under the `store` default the manager currently implies — verified in the manager's own manifest assembly, which injects no `OPENCOMPANY_MEMORY` at all).

The comment above the list excludes `tinymemory-embedded` for its bundled-rusqlite C build; the module driver does not share that problem and the new comment records the proof: under this exact feature set, `cargo tree -i libsqlite3-sys` and `-i cc` are both empty — the provider is a bus proxy and the loader stack is pure Rust.

The env half (the actual canary flip) lives in opencompany-microservice: the manager's tenant env assembly has **no per-tenant override channel today** (its `extra_env` carries only minted mongo/mail credentials), so that mechanism is being added there — #1524 carries the finding.

Depends on #1550 (the feature must exist to compile). No behavior change to any running tenant.

## API Or Behavior Changes

None — a compiled-in, default-off feature.

## Tests

- [x] `cargo tree -e normal --features mongodb,openhuman,tinycortex,tinymemory,tinymemory-module -i libsqlite3-sys` — empty
- [x] Same tree `-i cc` — empty
- [x] Workflow YAML change only; the tenant-image build proves the rest in CI on merge

## Documentation

The env pair, prior-value recording, and rollback rule are in `docs/spec/runtime/memory-module-port.md` (#1550).

## Summary by CodeRabbit

* **New Features**
  * Enabled the TinyMemory module for the staging environment.
  * Added support for its activation conditions and dependency behavior in staging deployments.


YellowSnnowmann:
Closing with #1550 — this only adds `tinymemory-module` to `TENANT_FEATURES`, and that feature exists only on the branch just closed. On its own it would name a feature the crate does not define, so it fails rather than does nothing.

Worth noting what its conflict with #1568 nearly caused, since the same shape will recur on any long-lived branch touching that line. #1568 removed `tinycortex` from `TENANT_FEATURES`; this branch had added `tinymemory-module` to the pre-#1568 version of the same line. Resolving in favour of "ours" would have restored `tinycortex` to the shipping image as a side effect of keeping an unrelated addition — invisible in the diff, since the line is long and the eye goes to the part that changed on purpose.

The note above that list also quoted crate counts measured before #1568, which it invalidated.

PR #1552, opened 2026-08-21
setup: give each curated profile its own instructions on top of its shape

sanil-23:
## Summary

Follow-up to #1547, which gave setup-built teammates standing instructions keyed on `AgentFocus`. A shape is shared, so that layer collapses where a template leans on one: `analysis` covers **seven of the thirty** curated profiles, meaning an SEO Specialist and an Accountant were told the same thing however carefully that text was written. Widening the vocabulary in #1547 got `software` to four distinct sets out of five and left `ecommerce` at three.

Each `TemplateAgent` now carries its own line, appended after the shape's — what *this role* is judged on, rather than how its kind of work is done:

> **Operations Manager** (`coordination`) — A stock-out costs more than reordering slightly early. Say so the second time a supplier slips, not the fifth.
> **QA Engineer** (`build`) — Reproduce before you report, and give the exact steps. A bug nobody can trigger twice is a rumour.
> **Account Manager** (`coordination`) — The client hears bad news from you, early, with an option already attached. Never let scope grow quietly between invoices.

Neither of those is something its shape could say, and the two `coordination` teammates above no longer read alike. **Every template now reads five distinct instruction sets for five teammates.**

| template | #1547 | this PR |
|---|---|---|
| ecommerce | 3/5 | **5/5** |
| content | 3/5 | **5/5** |
| agency | 5/5 | 5/5 |
| consulting | 5/5 | 5/5 |
| software | 4/5 | **5/5** |
| generic | 5/5 | 5/5 |

## The text is looked up, never carried

The obvious implementation hangs it on `ProposedAgent` beside `focus` and lets it ride the review-screen round trip. That would be a hole, and the difference is worth being precise about: `focus` survives that trip safely **because** it is a value from a closed enum the host re-parses — the worst a crafted request achieves is the wrong belt from a list the host wrote. Free-form instruction text posted back would reach a teammate's system prompt verbatim, authored by whoever made the call, and the company-scoped setup route is deliberately open to any member rather than only the operator (`src/server/ops/setup.rs`).

So `manifest_from_setup` re-matches the template from the same answers and reads the text out of its own compiled-in tables. No new wire field, and `instruction_text_cannot_be_posted_in` posts one to prove it is ignored.

An operator who **renames** a role on the review screen drops its profile line and keeps its shape — the answer rather than a gap, since the host no longer knows the teammate is that profile, and inheriting a mandate from a role somebody deliberately changed is worse than falling back.

## API Or Behavior Changes

- New public `profile_instructions()` and `standing_instructions()`; new `TemplateAgent.instructions` field (not `Option` — every curated profile must have one, for the same reason `focus` is not optional).
- **Behaviour:** a teammate whose role matches a curated profile is now built with shape instructions *and* a profile line. A model-designed teammate whose role matches nothing is unchanged from #1547 — shape only.
- No route, DTO or wire shape changes. No tool-belt change.

## Tests

- [x] `cargo test` — 3372 + 14 + 1 passed, 0 failed

Six new tests. The ones worth reading:

- `instruction_text_cannot_be_posted_in` — the boundary above, exercised with a request carrying an `instructions` field containing an injection string.
- `no_profile_repeats_its_shape_or_a_globals_prompt` — extends #1547's six-word-run check so a profile line cannot restate the shape it sits on, which would make the second layer decorative.
- `a_curated_teammate_is_told_its_shape_then_its_profile` — pins the order.
- `a_renamed_role_falls_back_to_its_shape`, `a_designed_teammate_gets_the_shape_alone`.

One existing test needed **narrowing rather than fixing**, and it is the most interesting thing in the diff. `an_unreadable_focus_gets_no_invented_instructions` asserted that an unreadable focus leaves a teammate with no instructions at all — using the role `"Analyst"`, which is itself a curated `generic` profile. Adding this layer made it fail, correctly. The true claim is smaller: an unknown shape costs the *shape* text only, and a role the host's own table names keeps its profile line, because matching a role against a compiled-in table is not guessing a work shape. The test now covers both halves — a stranger role gets nothing, a known role keeps its profile.

## Documentation

`docs/spec/runtime/company-setup-guarantees.md` — new subsection under D9 covering the two layers, the lookup-not-carry boundary, and the rename fallback.

## Not in this PR

The in-console dialog still cannot carry any of this. `AddMember` and `OverlayAgent` have no `prompt` field, so a company staffed from inside the console gets uninstructed teammates while a host-wizard company gets instructed ones — the same two-paths-disagree shape as the id minting, where `manifest_from_setup` derives ids from the role and `add_member` from the display name. Both want fixing together, in a change that touches the durable overlay record.

## Summary by CodeRabbit

* **New Features**
  * Curated roster profiles now provide role-specific guidance alongside shared focus instructions.
  * Profile guidance is automatically selected from the matched role template.
  * Renamed or unmatched roles retain shared instructions without incorrect profile-specific guidance.

* **Documentation**
  * Added documentation describing instruction layering, template matching, and fallback behavior.

### DIFF
--- a/docs/spec/runtime/company-setup-guarantees.md
+++ b/docs/spec/runtime/company-setup-guarantees.md
@@ -130,6 +130,42 @@ be a staler one. They are written in the globals' register without reusing its
 sentences: a global teammate is on the same roster, and two agents given the
 same instructions are one agent twice.
 
+### The shape is a floor, and the profile sits on it
+
+A shape cannot be the whole answer, because a shape is shared. `analysis` covers
+seven of the thirty curated profiles, so an SEO Specialist and an Accountant were
+told the same thing however carefully that text was written — the collision
+above, one level down.
+
+So each curated profile carries its own line too, appended after the shape's:
+what *this role* is judged on, rather than how its kind of work is done. "A
+stock-out costs more than reordering slightly early" is not something the
+`coordination` shape can say. Every template now reads five distinct instruction
+sets for five teammates, where the widened vocabulary alone got `software` to
+four and `ecommerce` to three.
+
+Shape first, profile second, because the profile qualifies the general case —
+the order the persona already reads in, where role and mandate arrive before
+anything about how to work.
+
+**The profile text is looked up, never carried.** The obvious implementation
+hangs it on `ProposedAgent` beside `focus` and lets it ride the review-screen
+round trip. That would be a hole. `focus` survives the trip safely *because* it
+is a value from a closed enum the host re-parses — the worst a crafted request
+achieves is the wrong belt from a list the host wrote. Free-form instruction text
+posted back would reach a teammate's system prompt verbatim, authored by whoever
+made the call, and the company-scoped setup route is deliberately open to any
+member rather than only the operator. So `manifest_from_setup` re-matches the
+template from the same answers and reads the text out of its own compiled-in
+tables. A request that invents an `instructions` field is ignored, and a test
+posts one to prove it.
+
+An operator who **renames** a role on the review screen drops its profile line
+and keeps its shape. That is the answer rather than a gap: once "Report Writer"
+is "Reports", the host no longer knows the teammate is that profile, and
+inheriting a mandate from a role somebody deliberately changed is worse than
+falling back.
+
 An unrecognised focus is instructed with **nothing**, which is the opposite of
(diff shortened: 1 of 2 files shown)


sanil-23:
@coderabbitai review

sanil-23:
@coderabbitai review

PR #1573, opened 2026-08-23
fix(memory): move health status behind operator auth

### LINKED ISSUE #1277: /spec memory.healthy: staleness marker or re-probe, and operator-auth placement
Follow-ups from @oxoxDev's #1273 review, deferred there deliberately:

1. **Staleness.** `memory.healthy` is written once by the boot probe and never refreshed. A long-lived process serves a stale `true` after the hosted engine goes down (worse than absent) and a stale `false` after it recovers. Wanted: a probe timestamp beside the bit, or a short-TTL re-probe on `/spec` — belongs with the P2 Console panel work, which renders the tri-state today.

2. **Auth placement.** `/spec` is unauthenticated for the manager's sake, but the manager's need really covers `healthz`/`busy`; the `memory` block (driver id + healthy) hands an unauthenticated caller mild operational intel about a tenant's third-party memory provider. Consider moving the memory block behind operator auth, with `/spec` keeping only what the handshake needs.

senamakel:
## Summary
- remove memory engine identity, capabilities, and health from the unauthenticated /spec handshake
- refresh the bounded health probe on authenticated memory-engine reads without mutating live engine state
- move the Settings display to the authenticated Brain engine surface and document the new verification path

Closes #1277

## Validation
- cargo fmt --all -- --check
- cargo test spec_does_not_disclose_memory_engine_details
- cargo test the_default_host_reports_the_built_in_store_as_editable
- cargo test --features tinymemory read_reprobes_the_live_memory_engine (running during PR creation)
- pnpm --dir frontend typecheck
- pnpm --dir frontend test -- memory-engine.test.ts

## Interpretation
I used the issue's re-probe alternative on the existing authenticated GET /memory/engine surface rather than adding a timestamp. Each read returns its own fresh, three-second-bounded probe result. The live overlay is not overwritten by that read, so it cannot race a concurrent engine switch.


senamakel:
@codex review

PR #1576, opened 2026-08-23
fix(ledgers): hide empty notices on boards

### LINKED ISSUE #1486: Work board: an empty list says it is empty twice, in two different voices
On a company with no cards, `#/ledgers/tasks` renders:

- `Nothing recorded here yet.` — a line above the board, and
- `Nothing here` (now `Drop tasks here`) — inside every one of the six columns.

Seven statements of one fact, in two vocabularies, one of which is a status and the other an affordance.

The above-board line is `data-testid="ledger-empty"` and exists for the **list** view, where there are no columns to say it in. On a board it is redundant, and it is the first thing an operator meets on a brand-new company — which is exactly when the board should be showing them where work goes, not confirming that none has.

Same for its sibling: `filteredEmptyNotice` renders `No rows match…` above a board whose columns each independently show their own empty state.

## Fix

Suppress the above-board empty line while the board view is showing; keep it for the list view, which needs it.

`frontend/src/views/LedgersView.tsx`

senamakel:
## Summary
- Hide the above-view empty and filtered-empty notices in board mode, where each column already provides its empty-state affordance.
- Retain both notices in list mode, where there are no columns to communicate the state.
- Add a live-host browser regression covering empty and filtered-empty board and list views using an isolated declared ledger.

Closes #1486

## Validation
- `npm run typecheck:e2e`
- `npm run typecheck`
- `npm test`
- `npm run build`

I attempted the focused Playwright regression, but the environment's `/tmp` quota prevented the required host-binary probe from starting (the sandbox could not mount `/tmp/.git`). The regression is included for CI's live-host suite.

### DIFF
--- a/frontend/src/views/LedgersView.tsx
+++ b/frontend/src/views/LedgersView.tsx
@@ -911,7 +911,8 @@ export function LedgersView({
                   indistinguishably — the read is filtered server-side. The two
                   are not the same claim, and saying the stronger one over a
                   ledger the nav is counting 14 rows for is simply false. */}
-              {read &&
+              {mode === "list" &&
+                read &&
                 read.entries.length === 0 &&
                 !reading &&
                 (emptyNotice ? (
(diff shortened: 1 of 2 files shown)


senamakel:
@codex review

Issue #1430, opened 2026-08-20
Desks renders every person as bare text, while the roster went to some trouble to give them faces


graycyrus:
Design audit of the **Company** surface. Judgement call, no Figma reference.

#1185 made teammate avatars id-seeded so a rename never changes a face, and #1181 put the shared chat mascot on the roster card and the detail header so a teammate looks the same in a DM and on the page that is about them.

`#/company/desks` renders the *same people* as plain text everywhere:

- every seat in every desk — name, optional role, a crown if they lead
- the "Not on a desk" chips at the bottom
- the per-desk "Add teammate" menu
- the **New desk** dialog's roster picker (13 bare rows)

So an operator who has learned the roster by face — which is what the seeded companies actually give you, since every teammate's name *is* its role and the text is nearly interchangeable ("Backend Engineer", "Frontend Engineer", "Security Engineer", "QA Engineer") — loses that the moment they move to the page about structure. Scanning a desk for "the purple one" is impossible; you have to read.

Faces are the cheapest disambiguator this surface has and it is the surface that needs it most.

## Expected

A small `TeammateAvatar` (~20px, the component already handles the "too small for a mascot, fall back to the tone tile" case) on each seat, on each unplaced chip, and in both pickers.

Severity: **taste**, but cross-surface — it undoes the identity work #1181/#1185 paid for.

PR #1584, opened 2026-08-23
feat(desk-create): add teammate filter and selection checkmarks

### LINKED ISSUE #1438: New desk: the roster picker doesn't look pickable, and doesn't scale
Design audit of the **Company** surface. Judgement call, no Figma reference.

The **New desk** dialog (`DeskCreateDialog`, reached from `#/company/desks`) asks the operator to "Pick who staffs it — the first teammate you choose leads it", and offers one bordered row per roster teammate.

**Nothing on a row says it is selectable.** Each row is a `<button aria-pressed>` styled as a plain bordered box; selection shows only as `border-primary/50 bg-primary/10` plus a tiny right-aligned `lead` / `2` / `3`. At rest the picker reads as a list, not a set of choices — in dark mode particularly, where the row border is close to invisible against the popup and the thirteen rows read as a plain roster dump. There is no checkbox, no check glyph, no count.

**It doesn't scale.** Thirteen rows on the seeded company, no search, no avatars, and the ordering inherits the roster's own unexplained split (see #1436). On a company with forty teammates this is a scroll-and-hunt.

**The ordering rule is invisible.** "the first teammate you choose leads it" is stated once in the dialog description; afterwards the only trace is a `lead` label on one row and a `Lead: X` line under the list. Choosing in a different order to change who leads is not something the picker's shape suggests.

## Expected

- a real selected/unselected affordance per row (checkbox or check glyph), so state is readable at rest
- a name filter above the list once the roster is long
- avatars, per #1430
- the selection order rendered as a numbered sequence rather than a bare `2` / `3`, so "first = lead" is legible from the list itself

Severity: **taste**, on the one dialog that can create structure.

senamakel:
## Summary
- add a name filter for desk rosters longer than eight teammates
- add checkbox-style selection indicators and render selection order as 1 · Lead, 2, 3
- cover filtering, visible selection state, and lead ordering in the desk-create dialog unit test

Closes #1438

## Validation
- pnpm --dir frontend test -- desk-create-name-error (the configured Vitest command ran the full unit suite)
- pnpm --dir frontend typecheck
- pnpm --dir frontend build

## Interpretation
The issue does not set a roster-length threshold, so the filter appears at nine or more teammates. Existing avatars from #1430 are preserved; no avatar work was duplicated.

### DIFF
--- a/frontend/src/views/company/DeskCreateDialog.tsx
+++ b/frontend/src/views/company/DeskCreateDialog.tsx
@@ -9,7 +9,7 @@
 // dialog at all, so creating a desk was impossible outside the manifest.
 
 import { useEffect, useId, useRef, useState } from "react";
-import { Crown } from "lucide-react";
+import { Check, Crown } from "lucide-react";
 
 import type { OpenCompanyClient } from "@/api/client";
 import type { DeskDto, TeamMemberDto } from "@/api/types";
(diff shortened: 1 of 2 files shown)


senamakel:
@codex review

PR #1594, opened 2026-08-23
fix(usage): distinguish unavailable usage from zero spend

### LINKED ISSUE #1425: settings/usage: a host with no usage surface renders as a company that has spent nothing — 404 is swallowed into a dashboard of zeros
Found in a UI/UX audit of Settings, judging whether pages degrade honestly when a feature isn't in the build. Seven of the eight Settings pages do. Usage is the one that doesn't.

## What happens

`UsageView` swallows both 404s and substitutes an empty dataset:

```ts
.catch(() => { if (alive) setData(EMPTY_USAGE); })          // usage
.catch(() => { if (alive) setCaps({ configured: false }); }) // capabilities
```

A host that serves no usage surface at all therefore renders a complete, confident dashboard: `Total tokens 0`, `Cost $0.00`, `OAuth calls 0`, `Web searches 0`, flat charts, and `"No token plan configured."`

That is pixel-identical to a company that is wired correctly and simply hasn't spent anything.

## Why it matters here specifically

This is the page an operator opens to answer "is this costing me anything?" — and the two answers it cannot distinguish are "no" and "I can't see". Those have opposite follow-up actions. A zero that might mean "not measured" is worse than no number, because a real zero is a perfectly plausible reading and nothing prompts a second look.

The code comment acknowledges the adjacent decision and gets it right — *"Real-or-zero: an empty series renders flat rather than inventing numbers"* — but that reasoning covers an empty series from a working surface. It doesn't cover an absent surface.

The rest of Settings sets the standard this should meet. Billing: *"This host was built without Chargebee support, so these settings will be stored and have no effect."* Hosting, MCP Servers, Connections all name the specific reason they are inert. Usage says nothing.

Note the page already knows how to do this correctly one level down: its per-capability badges (`mediaStatus`, `composioStatus`, `searchStatus`, `mcpDirectoryStatus`) each check `inBuild` first, and `composioStatus` has an explicit `undefined` → `"Couldn't check"` rung added by #886 precisely so unknown is never shown in the alarm colour. The page-level read is the only part that doesn't distinguish unknown from zero.

## Proposed fix

Keep a 404 non-fatal — a missing usage surface should not replace the page with an error, and the current instinct to stay calm is right. But distinguish the states:

1. Track "the host has no usage surface" separately from "the usage surface reports zero".
2. In the first case, say so where the totals are, in the Billing/Hosting register, instead of rendering `0` and `$0.00`.
3. Leave the real-zero rendering exactly as it is.

`frontend/src/views/UsageView.tsx`.

senamakel:
## Summary
- show an explicit unavailable state instead of zero usage KPIs when a host does not serve the usage endpoint
- keep a successful zero-usage response unchanged, and distinguish an unavailable capability-status endpoint from an unconfigured plan
- add DOM regression coverage for unavailable and genuine-zero responses

Closes #1425

## Validation
- npm run typecheck
- npm run typecheck:e2e
- npm run typecheck:unit
- npm test
- npm run build

## Scope
I treated rejected usage and capability-status reads as unavailable, regardless of whether the host reports a 404 or another request failure, because the client does not expose an HTTP-status-specific branch. No backend behavior was changed.

### DIFF
--- a/frontend/src/views/UsageView.tsx
+++ b/frontend/src/views/UsageView.tsx
@@ -313,7 +313,7 @@ export function UsageView({ client, company }: Props) {
                   <CapabilityRow key={tier.namespace} tier={tier} />
                 ))}
               </div>
-            ) : capsLoaded && caps?.configured && caps.total ? null : (
+            ) : capsLoaded && caps?.configured && caps.total ? null : capsUnavailable ? null : (
               <p className="py-2 text-sm text-muted-foreground">
                 {capsLoaded ? "No token plan configured." : "Loading budgets…"}
               </p>
--- a/frontend/src/views/UsageView.tsx
+++ b/frontend/src/views/UsageView.tsx
@@ -9,11 +9,12 @@ import {
   XAxis,
   YAxis,
 } from "recharts";
-import { Coins, CreditCard, Gauge, Plug, Search, Zap } from "lucide-react";
+import { Coins, CreditCard, Gauge, Plug, Search, TriangleAlert, Zap } from "lucide-react";
 
 import type { OpenCompanyClient } from "@/api/client";
 import type { CapabilityStatusDto, UsageDto } from "@/api/types";
 import { Badge } from "@/components/ui/badge";
+import { Alert, AlertDescription } from "@/components/ui/alert";
 import {
   Card,
   CardContent,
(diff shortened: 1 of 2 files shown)


senamakel:
@codex review

PR #1599, opened 2026-08-23
fix(settings): canonicalize invalid Settings sub-hashes

### LINKED ISSUE #1422: settings: the sub-nav is a flat list of eight with hints wrapping beside an empty gutter, and three different page-title treatments across it
Found in a UI/UX audit of Settings. Taste and craft rather than a defect — the page works, it just makes the operator do the sorting.

## Sub-navigation is a flat list of eight

`SETTINGS_PAGES` (`frontend/src/views/SettingsSection.tsx:29-47`) renders one undifferentiated column:

```
General       Connection, lifecycle, domain, mail
People        Who can sign in, and as what
Connections   Third-party accounts
MCP Servers   Tool servers and their tools
Billing       Invoicing through Chargebee
Hosting       Where this company's sites go live
Skills        Playbooks your teammates read
Usage         What this company is spending
```

There are at least three obvious groups in there — the company itself (General, People), what it acts through (Connections, MCP Servers, Billing, Hosting), and what it knows and spends (Skills, Usage). The comments in the file show the ordering was argued about pairwise ("Sits beside Connections rather than inside it…", "Beside Billing for the same reason…"), which is what happens when a list has structure but no way to express it. Eight is past the point where a flat list scans.

## The hints wrap next to empty space

The rail is a fixed `w-60`. At 1440px, three of the eight hints wrap to two lines — "Connection, lifecycle, domain, mail", "Where this company's sites go live", "Playbooks your teammates read" — while roughly 140px of empty gutter sits between the rail and the content card. The rail is too narrow for its own copy, and the space to fix it is immediately to its right and unused. Same at ~1100px.

## Page titles are inconsistent across the eight

Three treatments, plus an absence:

- `text-2xl font-semibold tracking-tight` — Connections, MCP Servers, Skills, Usage, People
- `flex items-center gap-2 text-lg font-medium` with an icon — Billing, Hosting
- `sr-only` — General, deliberately, per #1221's reasoning that the rail beside it already says "Settings"

General's argument is sound but it is the only page that takes it, so the section has a visible heading on seven pages and not on the eighth, in two different sizes. (#1392 covers missing `<h1>`s on other surfaces including Brain — this is the narrower "the eight Settings pages should agree with each other" version, and should be settled consistently with it.)

## Also: an unknown sub-page renders General while the address bar keeps the bad hash

`resolvePage()` falls back to `general` for any unrecognised segment, but `canonicalize()` in `use-hash-view.ts` never rewrites it — the resolution is internal to `SettingsSection`. So `#/settings/nonsense` shows General while the URL keeps claiming otherwise, and a bookmark or a shared link silently means something different from what it says.

## Proposed fix

1. Group the eight under two or three quiet section labels in the rail, using the groupings the file's own comments already imply.
2. Widen the rail, or reflow so hints stop wrapping at the two widths the console targets.
3. Pick one page-title treatment for all eight and settle it against #1392.
4. Canonicalise an unknown Settings sub-hash to `#/settings/general`.

`frontend/src/views/SettingsSection.tsx`, `frontend/src/hooks/use-hash-view.ts`.

senamakel:
## Summary
- Canonicalize an unknown Settings sub-hash to #/settings/general before the section renders its default page.
- Use the shared visible text-2xl heading treatment on all nine current Settings pages.
- Add route and navigation regression coverage.

Closes #1422

## Validation
- npm test -- --run test/unit/settings-navigation.test.ts test/unit/settings-route.test.ts
- npm run typecheck:unit
- npm run typecheck
- npm run build
- npm test (fails outside this change: workspace-repair-reveal-no-write.test.ts observes an unexpected write; WorkflowCreateDialog delayed timers also raise client.post is not a function.)

## Scope
Issue #1583 had already grouped the rail and addressed its responsive layout, so this PR deliberately leaves that landed work intact. The current Settings table has nine pages after Billing moved to Finance; this change applies the shared heading treatment to all nine.

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -100,6 +100,7 @@ import { InboxView } from "@/views/InboxView";
 import { MemoryView } from "@/views/MemoryView";
 import { FeedbackView } from "@/views/FeedbackView";
 import { SettingsSection } from "@/views/SettingsSection";
+import { isSettingsPage } from "@/views/settings-pages";
 import { useLocalScope } from "@/connections/ConnectionContext";
 
 // React Flow is heavy and only used here — load it on demand.
@@ -231,6 +232,10 @@ const REWRITE_RETIRED = (
   sub: string | null,
 ): [View, string | null] | null => {
   if (head === "tasks" && taskIdFromSegment(sub) === null) return ["ledgers", BOARD_LEDGER];
+  // Settings owns a fixed table of sub-pages, unlike the entity ids beneath
+  // Team and Workspace. Do not render General under an address that names no
+  // page: a bookmark or shared link must say where it actually lands.
+  if (head === "settings" && sub !== null && !isSettingsPage(sub)) return ["settings", "general"];
   // Bare `#/team` is the Company page now (issue #1141). It rendered the
   // teammate card grid from a route with no nav entry, so nobody arrived at it;
   // the grid is Company's Cards half, and leaving `#/team` answering as well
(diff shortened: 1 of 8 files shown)


senamakel:
@codex review

PR #1607, opened 2026-08-23
fix(accessibility): name filter and composer controls

### LINKED ISSUE #1395: Three select triggers have no accessible name (axe critical), and five text fields are named only by their placeholder
Two input-labelling problems found by walking the accessibility tree on every route.

### 1. Three select triggers have no accessible name at all

axe `button-name`, **critical**, at every viewport and in both themes:

| route | control |
|---|---|
| `#/ledgers` | the status filter (`Every status`) |
| `#/memory` | the type filter (`All types`) |
| `#/settings/usage` | the range filter |

All three render as `<button role="combobox" aria-haspopup="listbox">` with no `aria-label`, no `aria-labelledby` and no visible `<label>`. A screen reader announces "combobox" and nothing else — the operator has no way to know what they are about to filter.

### 2. Several text fields are named only by their placeholder

`element.labels` is empty and there is no `aria-label`; the only text is the placeholder, which disappears the moment the user types (WCAG 3.3.2, and 1.3.5 for autocomplete):

| route | field | placeholder |
|---|---|---|
| `#/ledgers` | board search | `Search every field` |
| `#/memory` | memory search | `Search memory…` |
| `#/settings/general` | SMTP host | `mail.acme.com` |
| `#/chat` | the composer textarea | `Message #engineering` |
| `#/workspace` | the upload `<input type=file>` | — (no name at all) |

The chat composer is the one that matters most: it is the primary input of the surface and it announces as an unlabelled textarea.

A visually-hidden `<label>` or an `aria-label` on each is the whole fix. The search fields already have a magnifier icon that could carry the `sr-only` text.

### Repro

1. `#/ledgers`, run axe → `button-name` critical on the status select.
2. Focus the search field with a screen reader running → "edit, blank" once anything is typed.

senamakel:
## Summary
- add explicit accessible names to the ledger and memory search fields and their filter triggers
- name the usage date-range trigger and preserve the chat composer’s contextual name after text is entered
- add a focused source-contract regression test for every corrected control

Closes #1395

## Validation
- `pnpm exec vitest run test/unit/accessibility-control-names.test.ts`
- `pnpm typecheck`

I also started `pnpm test` and `pnpm build`; the host command boundary ended each before its final summary, although the unit run reached and passed this regression test with no active runner left afterward.

## Scope
The issue’s SMTP host field already has a visible associated label, and the workspace file input is hidden behind the exposed upload control, so neither was changed.

### DIFF
--- a/frontend/src/views/LedgersView.tsx
+++ b/frontend/src/views/LedgersView.tsx
@@ -836,6 +836,7 @@ export function LedgersView({
                   <Search className="pointer-events-none absolute left-2 top-1/2 size-4 -translate-y-1/2 text-muted-foreground" />
                   <Input
                     className="pl-8"
+                    aria-label="Search ledger entries"
                     placeholder="Search every field"
                     value={query}
                     onChange={(e) => setQuery(e.target.value)}
@@ -853,7 +854,7 @@ export function LedgersView({
                       this read `all` while the list under it read "Every
                       status", and a chosen board column read `in_progress`
                       beside columns headed "In progress". */}
-                  <SelectTrigger className="w-[12rem]">
+                  <SelectTrigger className="w-[12rem]" aria-label="Filter by status">
                     <SelectValue>
                       {statusFilterLabel(ledger, statusFilter)}
                     </SelectValue>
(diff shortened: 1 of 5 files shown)


senamakel:
@codex review

PR #1619, opened 2026-08-23
fix(console): preserve raw run failure details

### LINKED ISSUE #1366: Run failures show raw engine prefix chains ("harness error: capability error: http_request:") to operators
**Designer's audit finding — professional judgement, no Figma reference for this surface.**

## What happens

A real failed run, rendered verbatim in the history row and in the failure panel:

> This run failed at “Fetch invoices”: **harness error: capability error: http_request:** Blocked local/private host: 127.0.0.1

Three layers of internal prefix sit between the operator and the one clause they can act on. "harness error" and "capability error" name subsystems this console's audience has never heard of and cannot do anything about; `http_request:` restates the node kind the sentence has already located by name.

The console's own half of that sentence is excellent — #1007 went to real trouble to say `“Fetch invoices”` rather than `fetch`. The host's half then undoes it.

## Proposed fix

The operator-facing sentence and the raw chain are two different artefacts and should be rendered as two:

- Lead with the clause that carries the fact ("Blocked local/private host: 127.0.0.1"), keeping #1007's `at “<node name>”`.
- Put the full untouched string behind a small "Details" disclosure, in mono — nothing is discarded, because the prefix chain is exactly what a bug report needs.

Strip in the console rather than at the host: the host's string is a contract other readers have (#981's verdict fold, the QA harness), and a console that trims for display cannot break them. A small, well-tested helper beside `run-error.ts` with the prefix set named and pinned, falling back to the whole string whenever it does not recognise the shape — an unrecognised error must never render as less than it is.

## Acceptance

- The failed row leads with the actionable clause and names the node.
- The complete original string is reachable in one interaction and is never lost.
- An error that does not match the known prefix shape renders exactly as it does today.

senamakel:
## Summary

- lead failed-run history and failure-panel messages with the actionable error clause
- preserve the untouched host diagnostic behind a closed, monospaced Details disclosure in both surfaces
- add renderer coverage for the visible leaf and raw diagnostic

Closes #1366

## Validation

- `npm test -- workflow-run-failure.test.ts workflow-run-error-message.test.ts`
- `npm run typecheck:unit`
- `npm run typecheck`
- `npm run build`

`npm test` was also run: 250 files / 2,265 tests passed, but the unrelated `test/unit/mock-brain.test.ts` timed out starting its local server.

## Scope

No host error contract was changed; the console retains raw errors only for the Details disclosure and uses the existing known-prefix helper for display.

### DIFF
--- a/frontend/src/views/workflows/RunFailurePanel.tsx
+++ b/frontend/src/views/workflows/RunFailurePanel.tsx
@@ -107,6 +107,14 @@ export function RunFailurePanel({
             {message}
           </AlertDescription>
         </Alert>
+        <details className="mt-2">
+          <summary className="cursor-pointer text-2xs text-muted-foreground">
+            Details
+          </summary>
+          <pre className="mt-1 overflow-auto rounded border bg-muted/40 p-2 font-mono text-2xs leading-snug text-foreground">
+            {failure.message}
+          </pre>
+        </details>
         <p className="mt-2 text-2xs text-muted-foreground">
           Failed after {formatDuration(ranFor)} ·{" "}
           {new Date(failure.atMillis).toLocaleString()}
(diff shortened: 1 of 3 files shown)


senamakel:
@codex review

PR #1623, opened 2026-08-23
feat(create-task): add priority picker to the new-task dialog

### LINKED ISSUE #1357: New task dialog cannot set priority, though priority is the loudest thing on the card
## What happens

`+ Add task` collects title, deliverable and owner. It does not collect **priority** — which is the badge rendered top-right on every card and coloured with the status hues. A newly-created card is always `low`, and changing it means creating the card, opening it, and editing it.

Smaller notes on the same dialog:

- `Deliverable` offers the raw enum value `once` in lowercase in a narrow select.
- The disabled `Create` button is the filled brand button at ~45% opacity, which reads as "loading" rather than "not yet".

`frontend/src/views/CreateTaskDialog.tsx`

senamakel:
## Summary
- add low, medium, and high priority choices to the new-task dialog and send explicit non-default selections to the existing tasks API
- preserve the host’s medium default by omitting untouched priority, and make an empty-prompt Create button neutral instead of a dimmed primary action
- cover payload omission/selection and the browser creation flow

Closes #1357

## Validation
- pnpm exec vitest run test/unit/create-task-body.test.ts
- pnpm run typecheck:unit
- pnpm run typecheck:e2e
- pnpm run typecheck
- pnpm run build
- pnpm test (2,265 tests passed; unrelated mock-brain.test.ts local-server startup hook timed out)

## Scope
Priority is limited to the existing low/medium/high vocabulary and existing create-task API field. The dialog continues to omit column so the server’s intake default remains authoritative. The browser e2e spec could not be run locally because the required Rust host build did not finish within the guarded command window during concurrent cold compilation.


senamakel:
@codex review

PR #1632, opened 2026-08-23
fix(chat): remove unavailable chat controls

### LINKED ISSUE #1336: Chat: three permanently-disabled controls sit in the chat chrome
Auditing `#/chat`, three controls are rendered but can never be used:

1. **New message** (`SquarePen`) in the channel-rail header — `disabled`, titled `New message — coming soon`. It is the only action in that header, at the top of the rail.
2. **Attach a file** (paperclip) in the composer toolbar — `disabled`, no explanation.
3. **Add an emoji** (smiley) in the composer toolbar — `disabled`, no explanation.

Plus two more inside the composers formatting row when it is opened: **Bulleted list** and **Link**, both `disabled`.

The composer toolbar is four icons wide and **two of the four do nothing**, with no tooltip saying why — so the operator learns by clicking. The Figma reference (file `iHAo3wYujP9bRlREmlTVBK`, node `927:2097`) shows the same four marks, which is presumably where they came from, but a reference frame is a picture of an intent rather than a promise to ship a dead button.

A disabled control with no reason reads as a bug; a disabled control with "coming soon" reads as an apology occupying permanent space. Either give them a reason on hover, or hold them back until they work — an empty toolbar slot costs nothing and a dead button costs trust.

Sites: `frontend/src/views/chat/ChannelRail.tsx`, `frontend/src/views/chat/MessageComposer.tsx`.

senamakel:
## Summary
- remove the permanently disabled New message action from the channel rail
- remove unavailable file, emoji, list, and link actions from the composer
- add a rendered UI regression test that keeps the working controls visible and the unavailable ones absent

## Validation
- `pnpm typecheck`
- `pnpm test`
- `pnpm build`

## Scope
Takes issue #1336's hold-back-until-functional interpretation: unavailable controls are absent rather than shown disabled with a tooltip. Working mention, formatting, and markdown-wrap actions are unchanged.

Closes #1336

### DIFF
--- a/frontend/src/views/chat/ChannelRail.tsx
+++ b/frontend/src/views/chat/ChannelRail.tsx
@@ -2,7 +2,6 @@ import { useState } from "react";
 import { ChevronRight, CircleDot, Hash, Lock } from "lucide-react";
 
 import { TeammateAvatar } from "@/components/teammate-avatar";
-import { Button } from "@/components/ui/button";
 import { cn } from "@/lib/utils";
 import { channelSubtitle, dmFace, type Channel, type ChannelSection } from "./model";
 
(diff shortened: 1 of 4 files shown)


senamakel:
@codex review

PR #1636, opened 2026-08-23
fix(console): link inbox empty state to company

### LINKED ISSUE #1331: Inbox empty state sends operators to a "Team page" that no longer exists
**Type: defect.** Found in a read-only design audit of `frontend/src/views/InboxView.tsx` against a seeded host.

## Repro

Sign in, go to `#/inbox` on a company with no teammate inboxes enabled (`GET /api/v1/companies/<co>/inboxes` → `[]`).

**Observed** — the empty state reads:

> **No inboxes yet**
> Give a teammate its own inbox from the **Team** page — flip on the inbox toggle for anyone who needs to receive email. Mail sent to that address shows up here.

There is no Team page. It is not in the sidebar, and `#/team` does not resolve to one — `REWRITE_RETIRED` in `app-shell.tsx` sends bare `#/team` to `#/company` (#1141):

```ts
// Bare `#/team` is the Company page now (issue #1141). It rendered the
// teammate card grid from a route with no nav entry, so nobody arrived at it;
if (head === "team" && !sub) return ["company", null];
```

Only `#/team/<agentId>` survives, as the teammate detail sub-page.

## Why it matters

This is the whole content of the empty state — the one moment the surface has to tell someone how to make it non-empty — and its single instruction names a destination that was retired. An operator following it finds no "Team" anywhere in the nav; if they guess the URL they are bounced to Company, which does not present itself as the place the instruction meant.

The word "Team" is styled `font-medium`, which in this console's vocabulary reads as an emphasised proper noun and invites a click, but it is plain text inside a `<p>` — not a link, not focusable. So the one actionable-looking element is inert as well as wrong.

`#/inbox` is itself a parked, nav-less surface (#302), so this is not on a hot path — but it is a two-word fix and the kind of rot that makes the rest of a parked surface untrustworthy when someone comes back to un-park it.

## What good looks like

- Point at wherever the inbox toggle actually lives today (Company, or the teammate's own `#/team/<agentId>` detail page) and verify the toggle is reachable there.
- Make it a real link, so the empty state ends in an action rather than a direction.
- Worth a sweep for the same rot: grep the console's empty states and helper copy for other references to retired destinations ("Team page", "the Desks page", "the board") left behind by #302, #1140 and #1141. This one was found by reading a single empty state; there are likely siblings.

senamakel:
## Summary
- replace the retired, inert Team page instruction in Inbox with a real Company page hash link
- explain that operators open a teammate from Company to enable its inbox, and add a DOM regression test for that anchor
- sweep and update the remaining active setup copy that named the retired Team page

## Validation
- npm run typecheck:unit
- npm test -- --run test/unit/inbox-empty-state.test.ts
- npm run typecheck
- npm run build
- npm test (2265 tests passed; the suite still fails on an unrelated mock-brain.test.ts helper-process startup timeout in this environment)

## Scope
The empty state has no teammate id, so it links to #/company rather than attempting an invalid teammate detail deep link. The active-copy sweep was limited to retired-destination wording; historical implementation comments were left intact.

Closes #1331

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -1070,7 +1070,7 @@ export interface InboxDto {
   name: string;
   /** The full address (`{key}@{domain}` when a domain is configured). */
   address: string;
-  /** Whether the inbox is enabled on the Team page. */
+  /** Whether the inbox is enabled on this teammate's detail page. */
   enabled: boolean;
   /** The number of unread received (inbound) messages. */
   unread: number;
--- a/frontend/src/setup/SetupDialog.tsx
+++ b/frontend/src/setup/SetupDialog.tsx
@@ -327,8 +327,8 @@ function BuildOut({
         <DialogDescription>
           {finished
             ? fallback
-              ? "A general starting team for your industry — we couldn't reach a model to tailor it to your answers. Rename, retire, or add anyone from the Team page."
-              : "Built from your answers. A starting point — rename, retire, or add anyone from the Team page."
+              ? "A general starting team for your industry — we couldn't reach a model to tailor it to your answers. Rename, retire, or add anyone from the Company page."
+              : "Built from your answers. A starting point — rename, retire, or add anyone from the Company page."
             : buildOutLabel(created, agents.length)}
         </DialogDescription>
       </DialogHeader>
--- a/frontend/src/views/InboxView.tsx
+++ b/frontend/src/views/InboxView.tsx
@@ -177,9 +177,12 @@ export function InboxView({ client, company }: Props) {
         <div className="space-y-1">
           <p className="font-medium text-foreground">No inboxes yet</p>
           <p className="max-w-sm text-sm">
-            Give a teammate its own inbox from the <span className="font-medium">Team</span> page —
-            flip on the inbox toggle for anyone who needs to receive email. Mail sent to that
-            address shows up here.
+            Give a teammate its own inbox from the{" "}
+            <a className="font-medium text-foreground underline-offset-4 hover:underline" href="#/company">
+              Company page
+            </a>{" "}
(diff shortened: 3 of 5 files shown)


senamakel:
@codex review

senamakel:
### PR babysitter — ready for approval

**Head inspected:** `9ea80323a822271befd3867e06d48e1053acc96c` (stable across the loop)

**Fixes pushed this run:** none — the head arrived green and clean.

**CI:** 12 SUCCESS, 0 pending, 0 failing. `Rust (openhuman, tinymemory)` and `Rust (mongodb)` are SKIPPED by the path filter (the diff is frontend-only); the `tinysweeper/*` sub-checks are NEUTRAL/informational.

**Validation run locally:**
- `npx vitest run test/unit/inbox-empty-state.test.ts` → 1 passed. Asserts the empty state renders an `<a href="#/company">Company page</a>`, that the new "open a teammate to flip on the inbox toggle" copy is present, and that no "Team page" text survives. It fails on base, where the empty state rendered a plain `<span>` with no href.

**Review feedback:**
- `tinysweeper` — APPROVED, nothing blocking.
- `chatgpt-codex-connector` — "Didn't find any major issues" on `9ea80323a8`. (Two earlier comments were usage-limit notices, not findings.)
- `coderabbitai` — review-limit notice; its check reports SUCCESS. No findings raised.
- Unresolved review threads: 0. Reviews requesting changes: 0.

**Route check:** `#/company` is a member of the `View` union and has a `ROUTABLE` entry in `src/lib/console-routes.ts`, so the address resolves. The plain hash `href` matches the existing convention (`MessageTimeline.tsx:478` links the same way); `useHashView`'s canonicalize re-asserts `?host=` after navigation, so no `withHostParam` wrapper is needed at this call site.

**Next action:** hand to `pr-approval-reviewer`.

PR #1650, opened 2026-08-23
fix(chat): leave message intent unselected

### LINKED ISSUE #984: chat: an ordinary message opens a task card, so the board fills with conversation

senamakel:
## Summary

- leave all three composer intent choices unselected initially and after each send
- send no operator override for an unmarked message, allowing normal host triage to decide whether it is work or conversation
- add a DOM regression test covering the unselected state, ordinary send, explicit workflow choice, and reset

Closes #984

## Validation

- `pnpm vitest run test/unit/chat-deliverable-choice.test.ts`
- `pnpm typecheck:unit`
- `pnpm typecheck`
- `pnpm build`

## Scope

This implements the remaining composer-default requirement. The classifier, explicit Just chatting override, and dismissible card chip were already present on `main`; bulk cleanup of existing cards is deliberately left out because the issue specifies an operator-reviewed flow and no such selection/review contract is implemented here.

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -454,12 +454,12 @@ export class OpenCompanyClient {
      * `"workflow"` say what the card it opens produces, `"chat"` says it is not
      * a request for work and no card should be opened for it.
      *
-     * Everything except `"once"` reaches the wire; `"once"` and the default are
-     * sent as *nothing at all*. That is what keeps "Do it once" the default: an
-     * unmarked message posts the exact body shape it posted before any of these
-     * controls existed, so the host cannot tell one from a pre-#580 client —
-     * the same omitted-field compatibility rule the deliverable field follows
-     * everywhere (see `CreateTask.deliverable`).
+     * Everything except `"once"` reaches the wire; `"once"` and no selection
+     * are sent as *nothing at all*. That preserves the historical wire shape:
+     * an unmarked message posts exactly what it did before any of these controls
+     * existed, so the host can apply its normal triage without a browser-asserted
+     * override — the same omitted-field compatibility rule the deliverable field
+     * follows everywhere (see `CreateTask.deliverable`).
      *
      * One key, not two. `"chat"` rides `deliverable` rather than arriving as a
      * second `intent` field, so a body cannot claim "build me the workflow" and
--- a/frontend/src/views/chat/MessageComposer.tsx
+++ b/frontend/src/views/chat/MessageComposer.tsx
@@ -65,15 +65,12 @@ export function MessageComposer({
   deliverableChoice,
 }: Props) {
   const [draft, setDraft] = useState("");
-  // What the NEXT line is for, and only the next one. It resets to "once"
-  // after every send so neither a workflow request nor a "just chatting" mark
-  // silently carries into the message after it — each is an explicit, per-line
-  // decision.
-  //
-  // "once" is the initial value, and stays it (issue #1152): "Just chatting" is
-  // a third position, not the new default, so an unmarked message is
-  // byte-identical on the wire to one sent before this control existed.
-  const [intent, setIntent] = useState<MessageIntent>("once");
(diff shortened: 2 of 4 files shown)


senamakel:
@codex review

PR #1654, opened 2026-08-23
fix(server): box pre-rendered handler rejections

### LINKED ISSUE #1299: Reshape server handler rejections so clippy::result_large_err can be re-enabled
Follow-up to #1298, which unblocked CI by scoping
`#![allow(clippy::result_large_err)]` to `src/server`. That allow is a
deliberate deferral; this issue is the deferred work.

## What the lint is actually complaining about

Not `OpenCompanyError`, and not `ApiError` — both are comfortably under the
128-byte threshold. Every one of the 80 findings names the same foreign type:

```
= help: try reducing the size of `axum::http::Response<axum::body::Body>`
  the `Err`-variant is at least 128 bytes
```

`axum::http::Response<Body>` is exactly 128 bytes, which is exactly clippy's
default `large-error-threshold`. Handlers across the server surface return
`Result<T, Response>` — the "pre-rendered rejection" shape — and so every one
of them returns a 128-byte `Err`. The lint's point is real: those bytes are
moved on the success path too, since `Result<T, E>` is as big as its larger arm.

## Where it fires

80 handler signatures across 16 files:

| count | file |
| --- | --- |
| 19 | `src/server/operator.rs` |
| 10 | `src/server/users/admin.rs` |
| 10 | `src/server/users/routes.rs` |
| 5 | `src/server/ops/policy.rs` |
| 5 | `src/server/ops/team.rs` |
| 5 | `src/server/setup.rs` |
| 4 | `src/server/feedback.rs` |
| 4 | `src/server/ops/workflows.rs` |
| 4 | `src/server/users/devices.rs` |
| 3 | `src/server/ops/repos.rs` |
| 3 | `src/server/ops/setup.rs` |
| 2 | `src/server/ops/domain.rs` |
| 2 | `src/server/ops/read_state.rs` |
| 2 | `src/server/ops/smtp.rs` |
| 1 | `src/server/hooks_chargebee.rs` |
| 1 | `src/server/ops/team_agent.rs` |

## Why it was not fixed in #1298

`Response` is axum's type, so "box the large variant" is not available — there
is no variant of ours to box. The available shapes are all wider than a
CI unblock should be:

- `Result<T, Box<Response>>` — axum requires the `Err` type to implement
  `IntoResponse`, and `Box<Response>` does not, so this needs a newtype
  (`struct Rejection(Box<Response>)` + `IntoResponse`) threaded through all 80
  signatures and every `?` and `map_err` that feeds them.
- Return `ApiError` instead of a pre-rendered `Response` wherever the rejection
  is expressible as an `OpenCompanyError`. This is the better end state — it
  moves these handlers onto the same envelope mapping `src/server/error.rs`
  already owns — but some of these rejections are genuinely not
  `OpenCompanyError`-shaped (auth redirects, cookie-setting responses,
  setup-flow HTML), so it is a per-handler judgement call, not a mechanical
  rewrite.

Either way it is a reviewable refactor of the whole server surface, which is
not something to bury in a CI-unblock PR.

## Suggested approach

Do it per-module rather than in one sweep, narrowing the `allow` from
`src/server/mod.rs` down to the modules not yet converted as you go, so the
lint starts covering each module the moment it is clean. `operator.rs` is a
quarter of the total on its own and is a reasonable first slice.

## Done when

- `src/server/mod.rs` no longer carries `#![allow(clippy::result_large_err)]`.
- `cargo clippy --all-targets --features openhuman,tinycortex,mcp -- -D warnings`
  is clean without it.

senamakel:
Summary:
- replace the server-wide result_large_err suppression with a compact Rejection wrapper for pre-rendered Axum responses
- migrate handler and helper error arms to the wrapper while preserving status, headers, cookies, and bodies
- add coverage proving a pre-rendered redirect response is preserved

Closes #1299

Validation:
- cargo fmt --all -- --check
- cargo check --all-targets --features openhuman,mcp

Interpretation: tinycortex is no longer a Cargo feature on current main, so validation uses the current openhuman,mcp feature set.


senamakel:
@codex review

PR #1655, opened 2026-08-23
ci: run full checks for merge queue entries

### LINKED ISSUE #793: ci: evaluate a batched merge queue, so main is verified before a merge rather than after
## Summary

Split out of #770. That issue fixes the immediate defect — `ci.yml` cancelled `main`'s
post-merge verification on every merge — by scoping `cancel-in-progress` to non-default
refs. This issue is about the answer that makes the question moot, which is outside a
PR's power to apply.

## Why the #770 fix is not the end of it

Turning cancellation off on `main` buys a conclusive run **on the tip**, not per-merge
attribution. The concurrency group still admits at most one running plus one pending run,
and a newer arrival supersedes the older *pending* one. Verified empirically on a fork
while fixing #770: with three pushes to a default branch minutes apart, the running run
survived, the older pending run was cancelled as superseded, and the newest queued.

So in a fast burst of merges, a red run on the tip covers the **union** of merges since
the last conclusive one. It is bisectable, and that is strictly better than today's
"nothing concluded at all" — but it is not the same as knowing which merge broke `main`.

Post-merge verification is also structurally late. It tells you `main` is broken after
everyone's branches were already cut from it.

## What a merge queue changes

A merge queue verifies the **exact merged state** before it becomes `main`. A change that
breaks the combination never lands, so the cancellation question disappears rather than
being traded off, and per-merge attribution comes free — each queue entry is verified on
its own.

This is the same defect family as #592 and #475: a check that is quietly not doing what
everyone assumes. A merge queue removes the assumption instead of tuning it.

## What it costs, honestly

- **A `merge_group` trigger** in `ci.yml`, plus branch-protection settings changed in the
  repository — not something a PR can do.
- **Throughput.** The critical path is ~26-30 minutes (`Rust (openhuman, tinycortex)`,
  then `Console E2E (live brain)` behind it). Serialised, that is ~2 merges/hour, which
  the current merge rate would exceed immediately.
- So it needs **batching** (`merge_group` with a group size > 1), which restores
  throughput but reintroduces union-attribution *within a batch* — with the important
  difference that a failing batch is rejected instead of landing.

Trimming the critical path is the lever that makes the throughput arithmetic comfortable,
and is worth pricing alongside this.

## Ask

Maintainers to evaluate a **batched** merge queue for this repo, and decide between it and
the #770 status quo with the costs above stated.

## Related

- #770 — the immediate fix (scope cancellation to non-default refs)
- #592, #475 — same defect family: a check quietly not doing what it appears to

senamakel:
## Summary

- trigger CI for GitHub merge-queue `merge_group.checks_requested` events
- run every conditional Rust, console, and desktop lane for a queue's synthetic merge commit instead of relying on pull-request path metadata
- add a CI guard that fails if the merge-queue trigger or full-lane override drifts

Closes #793

## Validation

- `python3 -c 'import yaml; yaml.compose(open(".github/workflows/ci.yml")); print("YAML syntax parses")'`
- `bash -n scripts/ci/assert-merge-group-workflow.sh`
- `scripts/ci/assert-merge-group-workflow.sh`
- `scripts/ci/assert-toolchain-pin.sh`
- `cargo clippy --all-targets -- -D warnings`

I also attempted `cargo test --locked --lib`, but this shared-host worktree was unable to acquire Cargo's shared build-directory lock before the harness ended the invocation. CI will run the relevant suite normally.

## Scope

This PR implements the workflow side only. Enabling GitHub's merge queue, setting its required checks, and choosing a batch size remain maintainer-controlled repository ruleset settings. Queue events deliberately select all conditional lanes because they contain a synthetic combined merge rather than pull-request changed-file metadata; that avoids a green queue entry produced by skipped checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,19 @@ name: CI
 # concurrency block below exists to protect) and every pull request, including
 # ones from forks. What it drops: runs on branches with no PR open yet, which
 # are re-run in full the moment the PR is opened.
+#
+# Issue #793. A merge queue dispatches `merge_group.checks_requested` for the
+# synthetic commit that would land on the default branch. That is the state a
+# required check must validate: if it fails, GitHub removes the entry before it
+# becomes `main`. Enabling the queue and choosing its batch size are repository
+# ruleset settings, deliberately left to maintainers; this trigger is the
+# workflow half GitHub requires once they enable it.
 on:
   push:
     branches: [main]
   pull_request:
+  merge_group:
+    types: [checks_requested]
 
 # Issue #770. Cancellation is scoped to NON-DEFAULT refs, and the asymmetry is
 # the whole point: on a PR branch a new push makes the previous run's answer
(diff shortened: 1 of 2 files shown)


senamakel:
@codex review

PR #1709, opened 2026-08-24
fix(server): convert Response to Rejection in edit_agent

oxoxDev:
## Summary

`main` is currently red. Every PR that rebuilds against it inherits three
`E0308: mismatched types` compile errors in `src/server/ops/team_agent.rs`
(lines 667, 691, 711), breaking the `Rust`, `Desktop`, and `Gated host binary`
CI lanes. This PR fixes the type mismatch — no behavior change.

`edit_agent`'s three error-refusal sites build an `ApiError`, call
`.into_response()` (which produces `axum::http::Response<Body>`), and return
it directly from a function whose error type is `server::error::Rejection`.
`Response` and `Rejection` are different types — `Rejection` wraps a
`Box<Response>` and only gets there via `Into`/`From`, so the bare
`.into_response()` value doesn't satisfy the return type.

Introduced by `0075205f8` (feat(team_agent): allow editing native fields on
manifest teammates) and `bdcf82565` (chore(team_agent): remove redundant
instruction processing in edit_agent).

## API Or Behavior Changes

None. This is a type-level fix only — the HTTP responses these three refusals
produce are unchanged (same status codes, same messages). Confirmed by the
existing tests covering all three paths (see Tests).

## Tests

The fix appends `.into()` after `.into_response()` at each of the three
sites, converting `Response` → `Rejection` via `Rejection`'s
`From<Response>` impl. This mirrors the idiom already established twice in
the same function: the `is_roster_agent` check a few lines above (line ~605)
and the `detail(...).map_err(|e| e.into_response().into())` call at the
function's tail — both use the identical `.into_response().into()` shape.
Rustc's own suggestion happened to match, but the choice was made by reading
the surrounding code, not by taking the compiler's first suggestion.

No new test was added. All three fixed error paths already have reachable-path
coverage in `src/server/ops/team_agent.rs`'s test module, and all pass against
the fixed code:
- `an_admin_can_pin_and_clear_a_teammates_harness` — asserts `400` for an
  unknown harness id (the site at line 667).
- `a_model_override_is_refused_off_an_acp_harness` — asserts `400` for a
  model set on a non-ACP harness (line 691).
- `a_model_is_refused_on_a_runner_bound_harness` — asserts `400` for a model
  set on a `runner`-transport harness (line 711).

Commands run locally:

- [x] `cargo check --all-targets`
- [x] `cargo clippy --locked --all-targets --no-deps -- -D warnings`
- [x] `cargo test --locked --lib server::ops::team_agent` — 38 passed, 0 failed (includes all three sites above)
- [ ] N/A: `cargo test --locked` (full suite) — could not complete locally; this
  machine is out of disk space (unrelated to this change, ~400GB+ tied up in
  other stale worktrees). The scoped run above exercises every path this diff
  touches.

## Documentation

None needed — internal error-type fix, no public API or behavior change.

## Related

Unblocks CI on:
- #1694
- #1705
- #1706 (already `APPROVED`, blocked only by this)
- #1708

### DIFF
--- a/src/server/ops/team_agent.rs
+++ b/src/server/ops/team_agent.rs
@@ -667,7 +667,8 @@ async fn edit_agent(
         return Err(ApiError(OpenCompanyError::InvalidRequest(format!(
             "no harness named `{id}` is available for this company."
         )))
-        .into_response());
+        .into_response()
+        .into());
     }
 
     // A model override only means anything on an `acp` harness — the same
@@ -693,7 +694,8 @@ async fn edit_agent(
                  transport to forward it to. Bind it to an ACP harness first, or clear \
                  the model."
             )))
-            .into_response());
+            .into_response()
+            .into());
         }
         // `kind = "acp"` is not sufficient: a `runner` transport is ACP and
         // still cannot carry a model, because the runner wire protocol has no
@@ -713,7 +715,8 @@ async fn edit_agent(
                  `transport = \"runner\"`. Model overrides aren't supported for a runner \
                  yet — the runner wire protocol doesn't carry them."
             )))
-            .into_response());
+            .into_response()
+            .into());
         }
     }
 


oxoxDev:
Closing this — `main` fixed the same three sites independently while this was open, so there is nothing left here to land.

Both fixes are correct, they just take different routes to `Rejection`, which is why the branch now conflicts on exactly the lines it was meant to repair:

```rust
// main
return Err(Rejection::from(ApiError(OpenCompanyError::InvalidRequest(...))));

// this PR
return Err(ApiError(OpenCompanyError::InvalidRequest(...)).into_response().into());
```

Current `main` has all three call sites on `Rejection::from(ApiError(...))`, no `Err(ApiError(..).into_response())` left in the file, and CI green at `37603caf` / `50575a6a` (last red was `8975859e`). The break that was reddening `Rust`, `Desktop` and `Gated host binary` on every open PR is gone at the source — anything still showing it just needs to pick up current main.

Worth noting for consistency, not as a blocker: the file now mixes both idioms — `Rejection::from(ApiError(..))` at the three sites above and `.into_response().into()` at the tail of the same function. Fine either way, but if one is preferred as the house style it would be a good small follow-up.

PR #1713, opened 2026-08-24
feat(approvals): add Approve all / Decline all to the inbox

theamazinghenk:
A visible operator drowning in a full approvals inbox has to click every row. There is no way to clear the queue at once.

## Change

Add **Approve all** / **Decline all** buttons to the approvals inbox header, shown only when more than one request is pending (a single row already has its own card buttons).

Deliberately client-side and sequential over the existing, validated resolve path, rather than a new batch endpoint:

- each decision goes through the same per-approval resolve call, so every existing guard (grant-scope validation, per-row drop-safety) still applies;
- the host already serialises decisions behind its per-company lock, so sending them one at a time respects that lock instead of fighting it;
- if one fails partway, the rows that landed drop out and the rest stay in the queue — a second click picks up exactly the remainder.

Scope is always `once`: a bulk action never mints a standing grant — that stays a deliberate per-tool choice.

Approve is not terminal (each approval resumes the teammate, so approving many at once starts several follow-up turns), so **Approve all** confirms first with the count. Both buttons disable while any decision is in flight.

### DIFF
--- a/frontend/src/views/ApprovalsView.tsx
+++ b/frontend/src/views/ApprovalsView.tsx
@@ -318,6 +318,41 @@ export function ApprovalsView({
     }
   }
 
+  // Bulk decisions for the whole visible queue.
+  //
+  // Client-side and sequential on purpose, rather than a new batch route: each
+  // call goes through the existing, validated resolve path, and the host already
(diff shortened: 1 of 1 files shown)


senamakel:
love it bro

PR #1747, opened 2026-08-25
fix(chat): resolve thread-panel avatar per message sender (#1729)

### LINKED ISSUE #1729: Bug: Thread panel shows wrong avatar for current user — renders agent avatar instead
## Description

In the thread panel, the current user's avatar is incorrectly rendered as the agent's avatar (green mascot), instead of the user's own avatar (yellow mascot).

## Steps to reproduce

1. Open any channel with an agent conversation (e.g. #engineering)
2. Send a message — avatar shows correctly in main chat (yellow)
3. Open the thread panel for that conversation
4. In the thread panel, the user's reply avatar renders as the agent's avatar (green) instead of the user's (yellow)

## Expected behaviour

Thread panel shows each participant's correct avatar — user gets their own avatar, agent gets the agent avatar.

## Actual behaviour

The user's avatar in the thread panel is replaced with the agent's avatar. Both user and agent appear with the same green mascot, making it impossible to distinguish who said what.

## Root cause hypothesis

Thread panel likely resolves participant identity/avatar from a shared context or thread metadata that incorrectly maps the user to the agent's identity. Possibly the thread render component is using a single avatar source for all participants rather than resolving per-message sender.

## Impact

High — breaks message attribution in threads. User cannot visually tell their own messages from the agent's.

## Screenshots

Thread panel (right side): user "You" at 13:06 shows green mascot instead of yellow.
Main chat (left): same user correctly shows yellow mascot.

oxoxDev:
## Summary

In the thread panel the current user's reply avatar rendered as the agent avatar (green mascot) instead of the user's own (yellow), breaking message attribution.

Root cause: `ThreadPanel`'s internal `Line` called `senderOf(message, channel, members)` **without** the 4th `youAvatar` argument. For a current-user line `senderOf` returns `{ kind: "you", avatar: youAvatar }`; with it `undefined`, `AvatarTile` fell back to `avatarFor("You")` — a mascot hashed off the literal string `"You"`, which renders green. The main timeline was correct because `buildTimeline(...)` already threads `youAvatar` into the same `senderOf`.

Fix: add a `youAvatar` prop to `ThreadPanel`, thread it into both `Line` renders (parent message + each reply) and pass it as the 4th arg to `senderOf`, then wire `youAvatar={youAvatar}` at the `ChatView` call site (state already existed and already feeds `buildTimeline`). `MessageRow` unchanged. Reuses the exact resolution path the main timeline uses.

Closes #1729

## API Or Behavior Changes

None (rendering fix only). Thread-panel avatars now resolve per message sender.

## Tests

Frontend-only change; Rust gates not applicable.

- [x] `npm run typecheck` (`tsc -b --noEmit`) — PASS
- [ ] `cargo fmt --all -- --check` — N/A: no Rust changed
- [ ] `cargo clippy --all-targets -- -D warnings` — N/A: no Rust changed
- [ ] `cargo build --all-targets` — N/A: no Rust changed
- [ ] `cargo test` — N/A: no Rust changed

## Documentation

None needed — UI rendering fix, no public API or docs surface affected.

### DIFF
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -1669,6 +1669,7 @@ export function ChatView({
             <ThreadPanel
               channel={channel}
               members={members}
+              youAvatar={youAvatar}
               parent={parent}
               replies={threadReplies}
               sending={sending}
--- a/frontend/src/views/chat/ThreadPanel.tsx
+++ b/frontend/src/views/chat/ThreadPanel.tsx
@@ -21,6 +21,13 @@ interface Props {
    * agent up by id (issue #1185).
    */
   members: TeamMember[];
+  /**
+   * Your own avatar reference, so a reply *you* wrote wears your face here the
+   * same way it does in the main timeline (issue #1729). Without it a "you"
+   * line resolves through `senderOf` with no avatar and `TeammateAvatar` seeds
+   * a mascot off the literal string "You" — a wrong (agent-looking) face.
+   */
+  youAvatar?: string;
   /** The message the thread hangs off. */
   parent: ChatMessage;
   replies: ChatMessage[];
@@ -77,6 +84,7 @@ interface Props {
 export function ThreadPanel({
   channel,
   members,
+  youAvatar,
   parent,
   replies,
   sending,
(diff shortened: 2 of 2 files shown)


oxoxDev:
Superseded by #1766 (commit 2e9470c37, "fix(chat): draw your own face on your own line in the thread panel"), which landed the full #1729 fix — ThreadPanel youAvatar threading, ChatView wiring, and the chat-thread-avatar tests — more completely than this branch. Every hunk here is already on main; a rebase would be empty. Closing as redundant.

Issue #1732, opened 2026-08-25
ci: macOS desktop release pipeline — DMG (unsigned CI artifact + gated Developer-ID sign/notarize for public releases)


oxoxDev:
## Summary
Net-new CI pipeline to package the OpenCompany Tauri desktop app as a **macOS DMG**. No desktop packaging exists today (`ci.yml` compiles `src-tauri` on ubuntu but never runs `tauri build`; `release.yml` builds the server; `deploy-staging` is the tenant image). Templated on openhuman's `build-desktop.yml` + `scripts/release/sign-and-notarize-macos.sh` / `repackage-dmg.sh`, scoped down (OC uses system WebKit/wry, no CEF, host embedded in-process).

## Behavior
- New `.github/workflows/release-desktop-macos.yml`, `workflow_dispatch` (inputs: `tag`, `sign` bool default false, `create_release` bool). macos-latest, **arm64 first** (`aarch64-apple-darwin`); x86_64/universal deferred.
- Steps: checkout (submodules) → `scripts/ci/init-vendored-submodules.sh` → rust toolchain + target → node (`frontend/.nvmrc`) → `npm ci && npm run build` in `frontend/` → `cd src-tauri && tauri build --target aarch64-apple-darwin` → **(sign=true only)** Developer-ID codesign (hardened) → notarize (`notarytool --wait`) → staple → repackage DMG → notarize+staple DMG → upload artifact / attach to Release.
- **Unsigned by default** (Actions artifact, for CI/internal). **Signed+notarized only when `sign=true` AND the `APPLE_*` secrets are present** (validated, fails fast if requested-but-missing). Never ship an unsigned DMG to a public release.

## New repo assets
- the workflow; a `bundle.macOS` block in `src-tauri/tauri.conf.json` (entitlements, `minimumSystemVersion`); `src-tauri/entitlements.plist` (hardened runtime, non-sandboxed: `network.client` + `network.server` only — server binds the loopback embedded host, client for platform/tenant/SSH); ported+simplified sign/notarize + repackage scripts (CEF/sidecar branches dropped).

## Secrets (Phase 2, admin action — NOT in repo yet)
[...]

PR #1812, opened 2026-08-26
ci(release-desktop-macos): matrix-build arm64 + x86_64 DMGs

oxoxDev:
## Summary

Adds `x86_64-apple-darwin` (Intel) to the macOS DMG release pipeline. The single-target `build` job in `release-desktop-macos.yml` is now a `strategy.matrix.target` over `[aarch64-apple-darwin, x86_64-apple-darwin]`, so one dispatch produces both an Apple-Silicon and an Intel DMG.

- `fail-fast: false` — one leg's failure never cancels the other; a partial success still publishes.
- `runs-on: macos-latest` (arm64) cross-compiles the Intel target once `rustup target add` installs it.
- Every hardcoded `aarch64-apple-darwin` replaced with `${{ matrix.target }}` (toolchain `targets:`, `rustup target add`, `tauri build --target`, and the `TARGET` env that drives `bundle_dir`).
- New **Name DMG per target** step renames the bundle to `<base>_<target>.dmg` — Tauri names both DMGs identically (productName + version), so without this the two artifacts/Release assets would collide. The notarization staple lives inside the `.dmg`, so renaming the file on disk preserves it.
- Per-target-distinct publishing: Actions artifact `opencompany-macos-${{ matrix.target }}-dmg`; Release asset name derives from the per-target filename, so both attach to one Release.

Signing/notarization gating (`Guard release-requires-signing`, `if: inputs.sign`) and all SHA-pinned `uses:` are unchanged — they now run per matrix leg.

## API Or Behavior Changes

None (CI workflow only). One dispatch now emits two DMGs instead of one.

## Tests

CI-only change — no local Rust gates apply. Validated:

- [x] `actionlint` clean on the workflow
- [ ] Live matrix run — not possible pre-merge: `workflow_dispatch` resolves on the default branch, so the matrix run can only be dispatched after this lands on `main`.

## Documentation

The header comment block in the workflow is updated to describe the two-target matrix.

## Related

Part of #1732 (macOS DMG packaging).

## Summary by CodeRabbit

* **New Features**
  * Added macOS builds for both Apple Silicon and Intel devices.
  * Release downloads now include clearly labeled, architecture-specific DMG files.
  * Both macOS builds are uploaded as separate release assets and artifacts.

### DIFF
--- a/.github/workflows/release-desktop-macos.yml
+++ b/.github/workflows/release-desktop-macos.yml
@@ -1,7 +1,9 @@
 # Issue #1732. Package the OpenCompany Tauri desktop shell as a macOS DMG.
 #
-# By default this produces an UNSIGNED arm64 (aarch64-apple-darwin) .dmg and
-# uploads it as an Actions artifact — the fast path for internal / CI builds.
+# A matrix runs both macOS targets — arm64 (aarch64-apple-darwin) and Intel
+# (x86_64-apple-darwin) — so one dispatch produces both DMGs. By default each
+# leg produces an UNSIGNED .dmg and uploads it as a per-target Actions
+# artifact — the fast path for internal / CI builds.
 # Developer-ID signing + Apple notarization is GATED behind the `sign` input
 # and the presence of the six APPLE_* secrets; only a signed+notarized DMG is
 # ever attached to a GitHub Release (`create_release`). An unsigned DMG never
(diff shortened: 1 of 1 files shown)


oxoxDev:
Superseded by #1813 (sanil-23), which does the x86_64 matrix leg plus removes the `acp`/`composio` feature gates and mirrors the Desktop CI lane. Closing this in favor of that broader, approved PR. Branch `ci/desktop-macos-dmg-x86_64` left on the fork; no further action here.

PR #1815, opened 2026-08-26
fix(acp): stop offering a coding CLI the build cannot run (#1814)

### LINKED ISSUE #1814: acp: the console offers claude/codex as bindable on a desktop build compiled without `acp`, then every turn fails telling the user to run it from the desktop app
## Summary

`can_run_local` uses *"was an `AcpAgentFactory` handed to `AppState`?"* as a proxy for *"can this build actually run a local ACP harness?"*. Those coincide everywhere except one configuration — a **desktop build compiled without the `acp` feature**, which is what `src-tauri/Cargo.toml` produces today.

| Build | Factory wired? | Runtime can use it? | `can_run_local` | Correct? |
|---|---|---|---|---|
| `opencompany serve` | no | no | `false` | ✅ |
| desktop **with** `acp` | yes | yes | `true` | ✅ |
| desktop **without** `acp` | **yes** | **no** | **`true`** | ❌ |

## Why the factory is wired even without the feature

`AppState::with_acp_agents` (`src/app/types.rs:587`) is ungated — deliberately, since `ports::acp` is ungated so the desktop can name the trait without pulling in the whole OpenHuman harness. So `src-tauri/src/embedded.rs:214` compiles and runs on *any* desktop build.

## Why the runtime still cannot use it

`src/runtime/builder.rs:3124` forces `let acp_agents = None` under `#[cfg(not(feature = "acp"))]`, and the non-`acp` `resolve_acp_engine` (`src/harness/lanes.rs:126`) returns `Err` unconditionally — its parameter is typed `&Infallible`, so `Some` is uninhabited.

## The two affected sites

- **`src/server/ops/harnesses.rs:126`** — the picker. `can_run_local` gates both `runs_here` and the `detected` list, so `claude` and `codex` (`ACP_AGENTS`, `src/company/types.rs:75`) surface as `Harness::implicit_local` with `detected: true, runs_here: true`.
- **`src/server/ops/team_agent.rs:795`** — the `PATCH` validator: `declared || (ACP_AGENTS.contains(&id.as_str()) && state.acp_agents().is_some())`, so the binding is accepted.

## Repro

On a desktop build at default features (no `acp`):

1. Open a company's harness picker → `claude` and `codex` are listed as detected and runnable.
2. Bind a teammate to one → `PATCH` returns 200.
3. Run a turn → it fails with `src/harness/lanes.rs:75`:

   > it is an ACP harness and this build has no ACP transport wired — **run it from the desktop app**, or bind these agents to a `built_in` harness

The remedy names the thing the user is already doing. That message was written on the assumption that only *server* builds would lack `acp`, which held until a desktop build shipped without it.

## Why this is the guard's own stated purpose

The comments above **both** sites describe preventing exactly this — "accepted by `PATCH`, then dead on the next rebuild, because the server has nothing to launch." They work for the server and miss the desktop, because they test whether a factory was *handed over* rather than whether the build can *use* it.

There is an accurate signal already on the wire and nothing renders it: `src/server/setup.rs:630` reports `acp_in_build: cfg!(feature = "acp")`, and `frontend/src/api/setup.ts:65` only declares the field.

## Suggested fix

```rust
let can_run_local = cfg!(feature = "acp") && state.acp_agents().is_some();
```

and the matching conjunct at `team_agent.rs:795` — preferably as one shared helper, since the two are already copy-paste siblings with copy-paste comments and this is precisely the drift that produced the gap.

## Relationship to #1813

PR #1813 makes the *released* DMG carry `acp`, so shipped artifacts stop landing in the broken row of the table above. It does **not** fix this: anyone running the desktop from source at default features (`cargo run`, any local dev build) still hits it. Worth fixing on its own merits — a picker that offers a lane the build cannot run is wrong regardless of which builds happen to ship the feature.

sanil-23:
## Summary

Closes #1814.

`can_run_local` asked whether an `AcpAgentFactory` was **handed over**, and used that to answer whether a local ACP harness can **run**. Those coincide on every server build and on a desktop compiled with `acp` — and diverge on exactly one configuration, a desktop compiled *without* it, which is what `src-tauri/Cargo.toml` produces today.

| Build | Factory wired? | Runtime can use it? | old `can_run_local` | Correct? |
|---|---|---|---|---|
| `opencompany serve` | no | no | `false` | ✅ |
| desktop **with** `acp` | yes | yes | `true` | ✅ |
| desktop **without** `acp` | **yes** | **no** | **`true`** | ❌ |

- `AppState::with_acp_agents` is deliberately ungated, so `src-tauri` can hand over a factory without pulling the embedded harness in behind `crate::harness` — `crate::ports::acp` exists for that reason. The desktop shell calls it unconditionally.
- The runtime cannot use what it was given: `RuntimeBuilder` forces `acp_agents = None` under `cfg(not(feature = "acp"))`, and `lanes::resolve_acp_engine` is an unconditional `Err` there — its factory parameter is typed `Infallible`, so `Some` is uninhabited.

The result: the picker offered `claude` and `codex` as detected and runnable, the teammate `PATCH` accepted the binding, and every turn then failed with `lanes.rs`'s *"run it from the desktop app"* — advice for somebody already in the desktop app.

The comments above **both** call sites describe stopping precisely this ("accepted by `PATCH`, then dead on the next rebuild"). They worked for the server and missed the desktop, because they tested the wrong thing.

`AppState::can_run_local_acp()` is now the predicate, living next to the accessor it qualifies. One method rather than the same conjunct twice: the two sites were already copy-paste siblings, comments included, and a predicate they can state differently is what opened the gap in the first place.

## API Or Behavior Changes

Behavior change, on a desktop build compiled **without** `acp` only. Nothing changes for `opencompany serve` or for a desktop build with the feature.

- `GET {scope}/harnesses` no longer lists `claude`/`codex` as detected, and no longer reports `runsHere: true` for a declared local ACP harness.
- `PATCH` on a teammate now refuses an undeclared `ACP_AGENTS` id with the existing "no harness named `{id}` is available for this company."

Both are the *intended* behavior of the guards as their own comments describe them — the write path and the picker agree again. A user who previously bound a teammate this way had a binding that could never run a turn.

No public API surface changes; `can_run_local_acp` is a new `AppState` method.

## Tests

Both lanes, since the `cfg!` branches must compile *and* assert correctly in each:

- [x] `cargo clippy --all-targets -- -D warnings` — clean on the default lane and on `--features acp,runner,tinymemory`
- [x] `cargo build --all-targets` — covered by the clippy runs above (`--all-targets`)
- [x] `cargo test` — `server::ops::harnesses` 6 passed, `server::ops::team_agent` 53 passed, on **both** the default lane and `--features acp,runner,tinymemory` (the lane `feature-lanes.txt` names for `acp`)

**The tests were encoding the bug.** `a_coding_cli_is_offered_only_where_this_host_can_run_one` and `an_undeclared_coding_cli_is_bindable_only_where_this_host_can_run_one` both asserted that a wired factory *alone* makes the CLIs available. `default` does not include `acp`, so they ran in the default `Rust` lane and passed *because they agreed with the bug*. They now assert the corrected invariant in both configurations.

**Verified by mutation.** Reverting `can_run_local_acp` to the old `acp_agents().is_some()` fails the picker test, with exactly the bug in the assertion output:

```
a build that cannot run `claude` must not offer it, even with a factory wired:
[… {"id":"claude","kind":"acp","detected":true,"runsHere":true},
    {"id":"codex","kind":"acp","detected":true,"runsHere":true}]
```

So the regression test guards the thing it claims to.

## Documentation

No doc files change. The rationale lives on `can_run_local_acp`'s own doc comment — why the two questions differ, which configuration separates them, and why it is one method instead of two conjuncts — with short pointers at both call sites, matching how the surrounding code documents its own guards.

## Notes

Deliberately **not** fixed here: `src/server/setup.rs` already reports `acp_in_build: cfg!(feature = "acp")` accurately, and `frontend/src/api/setup.ts` only declares the field — so the honest signal is already on the wire and nothing renders it. That is a console change and belongs on its own.

Related: #1813 makes the released DMG carry `acp`, so shipped artifacts stop landing in the broken row of the table above. It does not fix this — anyone running the desktop from source at default features still hits it, which is why this is separate.

### DIFF
--- a/src/app/types.rs
+++ b/src/app/types.rs
@@ -594,6 +594,35 @@ impl AppState {
         self.acp_agents.clone()
     }
 
+    /// Whether a `transport = "local"` ACP harness can actually run here.
+    ///
+    /// Issue #1814. Two callers — the harness picker and the teammate `PATCH`
+    /// validator — used to ask [`acp_agents`](Self::acp_agents) directly, which
+    /// answers a different question: whether a factory was HANDED OVER, not
+    /// whether this build can use one. Those coincide on every server build and
+    /// on a desktop compiled with `acp`, and diverge on exactly one
+    /// configuration — a desktop compiled WITHOUT it:
+    ///
+    /// * [`with_acp_agents`](Self::with_acp_agents) is deliberately ungated, so
+    ///   that `src-tauri` can hand over a factory without pulling the whole
+    ///   embedded harness in behind `crate::harness` (`crate::ports::acp` exists
+    ///   for that reason). The desktop shell calls it unconditionally.
+    /// * The runtime cannot use what it was given: `RuntimeBuilder` forces
+    ///   `acp_agents = None` under `cfg(not(feature = "acp"))`, and
+    ///   `lanes::resolve_acp_engine` is an unconditional `Err` there — its
+    ///   factory parameter is typed `Infallible`, so `Some` is uninhabited.
+    ///
+    /// The result was a picker that offered `claude` and `codex`, a `PATCH`
+    /// that accepted the binding, and then every turn failing with `lanes.rs`'s
+    /// "run it from the desktop app" — advice for somebody already in it.
+    ///
+    /// One method rather than the same conjunct at both call sites: they were
+    /// already copy-paste siblings, comments included, and a predicate the two
+    /// can state differently is what opened the gap in the first place.
+    pub fn can_run_local_acp(&self) -> bool {
+        cfg!(feature = "acp") && self.acp_agents.is_some()
+    }
+
     /// Sessions opened through the host's ACP HTTP transport.
     #[cfg(feature = "acp")]
     pub fn acp_sessions(&self) -> Arc<crate::server::acp::SessionRegistry> {
--- a/src/server/ops/harnesses.rs
+++ b/src/server/ops/harnesses.rs
@@ -123,7 +123,12 @@ async fn list_harnesses(
     // Declared entries are untouched: a manifest that names a `[[harness]]`
     // is stating what that deployment has, and this route does not get to
     // second-guess it.
-    let can_run_local = state.acp_agents().is_some();
+    //
+    // Issue #1814: `can_run_local_acp()`, not `acp_agents().is_some()`. The
+    // desktop wires a factory on every build, so the bare accessor said yes on
+    // a desktop compiled without `acp` — the one host where this route's whole
+    // purpose applied and it silently did nothing.
+    let can_run_local = state.can_run_local_acp();
     let runs_here = |harness: &Harness| match harness.kind.as_str() {
         "built_in" => true,
         "acp" => {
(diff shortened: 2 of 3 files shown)


sanil-23:
The red `Console E2E` here is **pre-existing on `main`**, not this change.

**It fails identically on the base commit.** Run [32935641746](https://github.com/tinyhumansai/opencompany/actions/runs/32935641746) on `main` at `def92a2a6` — which is exactly what this branch is based on — fails the same test with the same signature:

```
✘  53 test/e2e/chat-detached-post-failure.spec.ts:79:1 › a chat POST killed in flight still shows the reply the host went on to write
1 failed
333 passed
```

Same test, same index, same counts as both runs here (I reran the failed job once; it reproduced).

**This diff cannot reach that path.** The E2E host is the `opencompany-debug-binary` artifact, built as `cargo build --locked --bin opencompany` — a server build. `opencompany serve` never calls `with_acp_agents` (its only callers are `src-tauri/src/embedded.rs` and two unit tests), so `state.acp_agents()` is `None` there. Both predicates therefore agree on that binary:

- before: `acp_agents().is_some()` → `false`
- after: `cfg!(feature = "acp") && acp_agents().is_some()` → `false`

The only configuration whose behavior changes is a desktop build that wires a factory without the `acp` feature, and the E2E suite does not run the desktop shell. The failing spec is the detached chat-POST recovery test from #1000/#983 — it kills a request mid-flight and waits on a streamed reply — and touches nothing to do with harness selection.

`Console E2E (first run)` and `Console E2E (live brain)` both pass here, as do the Rust lanes on both the default and `acp` feature sets.

Happy to file the flaky spec separately if it is not already tracked — it appears intermittent with a high failure rate rather than deterministic (#1813, on the same base, passed it once).

Issue #1789, opened 2026-08-25
chat-detached-post-failure.spec.ts is flaky on main (3 of 8 runs), and every open PR inherits it


graycyrus:
## The flake

`frontend/test/e2e/chat-detached-post-failure.spec.ts:79` — *"a chat POST killed in flight still shows the reply the host went on to write"* — fails intermittently in the `Console E2E` lane, **on `main`**.

Of `main`'s ten most recent CI runs on 2026-08-25, eight ran the lane and **three failed**. All three are the same test, the same line, the same error:

```
Error: expect(locator).toBeVisible() failed
Locator: locator('article[data-message-id]').filter({ hasText: 'You said: cut-…' })
Timeout: 60000ms
  at .../chat-detached-post-failure.spec.ts:165:37
```

Failing job ids: `97815420188`, `97785229633`, `97702453846`. No other test failed that lane on `main` in the window.

## Why this is urgent rather than annoying

A ~3-in-8 failure rate on `main` means **every open PR inherits a coin-flip**. Eight of sixteen currently-open PRs failed this lane, and investigation found none of them was actually broken — the failures were this test, on unrelated diffs.

The cost is not the rerun. It is that a red lane on eight PRs is indistinguishable from eight real regressions, so either everything gets rerun on faith or a genuine failure gets waved through as "that flake again". One of those eight (#1760) *was* a real break, in a different lane, and it was nearly lost in the noise.

## Probable cause — read, not proved

The spec fulfils every `**/chat/history*` request with `[]` from the send onward.

`reReadSettledThread` in `frontend/src/components/app-shell.tsx` early-returns on `entries.length === 0`, so an empty history cannot *wipe* an already-rendered reply — but it does disable the settled-turn re-read that normally covers a missed live frame.

That leaves the single SSE `agent_reply` frame as the **only** delivery path for the assertion. Any hiccup in that one frame is then a hard 60-second timeout rather than a recoverable miss. The test has barred its own backstop.

If that reading is right, the fix is to let history answer truthfully for the settled re-read while still exercising the killed-in-flight path the test exists for — the test's *subject* is the detached POST, not the absence of history.

## What to establish first

1. Confirm the mechanism. Instrument or log which delivery path satisfies the assertion on a passing run, and what is missing on a failing one. The reading above is inference from source, not observation.
[...]

PR #1832, opened 2026-08-26
integrations: let a connect page grant its own `[tools].allow` namespace

### LINKED ISSUE #1796: integrations: 'Connected' but no teammate can use it — the manifest grant gap has no fix from the connect page
## Problem (systemic — highest severity)
Connecting an integration stores the credential but the company **manifest `[tools].allow` grant** is a *separate* step, and the connect/Finance/integration page **cannot add the grant** ("it cannot be fixed from this page"). So an integration reads **Connected but reaches nobody** — no agent/teammate can use it. This is the same dead-end across **chargebee, paypal, hosting, search, composio**.

Live symptoms (marketing/agentic tenant screenshots):
- Finance→Invoicing: "Chargebee  Connected to null — but no teammate can use it … Add `chargebee` to [tools].allow … it cannot be fixed from this page."
- Finance→Wallet: "PayPal  Connected — but no teammate can use it … Add `paypal` to [tools].allow …"

## Evidence
- Grant gap + "cannot be fixed from this page": `frontend/src/views/HostingView.tsx:193-197`, `frontend/src/views/SearchView.tsx:217-221`, `frontend/src/views/connections/ComposioSection.tsx:218`, `frontend/src/views/connections/ProvidersSection.tsx:155`, `frontend/src/views/finance/health.ts:10`.
- The `*` catch-all deliberately does **not** confer these namespaces (media/composio/search/hosting): `src/harness/build.rs:372-373,398-399,448`, `src/workflows/caps/tools.rs:348-349`.
- "Connected to null": the "Connected to {name}" render resolves name to null on chargebee (display bug on top of the grant gap).

## Fix direction
Make the connect surface able to add the missing `[tools].allow` grant (one click), or auto-grant on connect with an explicit toggle — so "Connected" implies "usable". At minimum, replace "cannot be fixed from this page" with an actionable control. Fix the "Connected to null" name resolution.

CodeGhost21:
Closes #1796.

## The problem

Connecting an integration and granting its tool namespace are two separate steps, and only the first one had a write path anywhere in the product. Every connect surface stored the credential and then told the operator the grant *"cannot be fixed from this page."*

That sentence was **accurate**, which is what made it a product failure rather than a copy failure:

- Nothing under `src/server/` writes `[tools].allow` — confirmed by grep; every consumer only reads it.
- `runtime/builder.rs` states the invariant: a rebuild re-persists `record.manifest` from the seed, merging only `[workflows].enabled`; *"every other manifest field is seed-authoritative, and for `[tools]`/`[policy]` that is a security property"*. So a manifest write would be wiped by the next rebuild anyway.
- The catch-all `*` deliberately refuses these namespaces (`grants_chargebee_explicit` and siblings in `src/company/types.rs`).
- On a hosted tenant the manifest is a read-only boot snapshot baked into the image, so "edit `company.toml` and redeploy" was not available either.

Net effect: five integrations — chargebee, paypal, hosting, search, composio — shipped reading **Connected** and reaching nobody, with no page and no file the operator could go to.

A second, independent bug: `Connected to null`. `chargebeeHealth` tested `!granted` **before** asking whether any credential existed, so a company that had never touched Chargebee fell into the `not_granted` arm — which asserts "Connected" and interpolates `status.site` (null). `paypalHealth` had the same ordering and printed a bare "Connected —" over a company with no client id. The file's own doc already defined `not_granted` as *"credentials stored and STILL nothing reaches an agent"*; the guards just did not say it.

## The fix

`[policy]` had exactly this problem (#562) and was solved with an **overlay**: a durable, attributed operator decision resolved ahead of the manifest, with `carry_policy_override` dropping it when the seed's `[policy]` changes. Tool grants were simply never given that layer. This adds it.

- **`ToolGrantsOverride`** (`src/ports/types.rs`) — additive only, attributed, persisted through every store backend and the export bundle.
- **`carry_tool_grants_override`** (`src/runtime/builder.rs`) — version control still wins: a `[tools]` edit in `company.toml` clears every console grant. Sharper here than for `[policy]`, because this is the only overlay that *widens*. `seed_allow` recovers the seed's own list from the materialised manifest so the layer does not clear itself on its first rebuild.
- **Folded into the persisted manifest**, not merged per read. `[tools].allow` is consulted at ~three dozen sites (roster build, workflow capability bundle, harness tool wiring, all five status routes); a parallel resolution would have to reach every one, and the single missed site is a console saying "Connected" over a harness that wired nothing — the bug, reintroduced by its own fix.
- **`GET`/`PUT`/`DELETE {scope}/tools/grants`** (`src/server/ops/tool_grants.rs`) — admin-only, attributed, bounded by a closed `CONSOLE_GRANTABLE_NAMESPACES` list enforced at the route *and again* at resolution. `shell`, `code` and `web` are not grantable from any page: every entry on the list is a namespace `*` refuses **and** one the console holds a credential form for, so granting is the second half of an action the operator already took.
- **A grant fingerprint axis** on `HarnessPool` — the grant lands on the next turn, not the next restart.
- **Console** — a shared `GrantNamespace` control on all five surfaces, plus the `health.ts` precedence fix.

## Commands run locally

Run the way CI runs them — this console is `npm`, not pnpm.

| | |
|---|---|
| `cargo fmt -- --check` | clean |
| `cargo build --locked --features openhuman --all-targets` | clean |
| `cargo clippy --locked --no-deps --features openhuman --all-targets -- -D warnings` | clean |
| `cargo test --locked` | 3902 passed, 0 failed |
| `cargo test --locked --features openhuman --tests` | 5616 passed, 0 failed |
| `cargo test --locked --features sqlite --lib store::sqlite` | 53 passed, 0 failed |
| `npm ci` + `npm run typecheck` / `:unit` / `:e2e` | clean |
| `npm test` | 3272 passed, 0 failed |
| `playwright test` (default lane, default-feature binary) | 334 passed, 26 skipped, 0 failed |
| `PW_LIVE_BRAIN=1 playwright test tool-grant-from-connect-page` | 2 passed, run twice against one persistent host |

## Tests added

**Rust** — 18 route tests (`src/server/ops/tool_grants.rs`) covering the grant, the closed list at both enforcement points, seed-grant no-op, accumulation, idempotence, single and bulk withdrawal, and admin-only writes; 6 carry-rule tests and `a_console_tool_grant_survives_a_rebuild` (`src/runtime/builder.rs`), which is deliberately the twin of the existing `rebuild_keeps_every_other_manifest_field_seed_authoritative` — a raw manifest write is still discarded, an attributed overlay is carried; `a_tool_grant_written_through_the_store_rebuilds_the_roster_in_place` (`src/harness/built_in/mod.rs`), the no-restart proof, driven through one pool that is never reconstructed. `src/store/conformance.rs` carries a **non-empty** override so a backend that drops the field fails rather than passing silently.

**Console** — `tool-grant-control.test.ts` drives `HostingView` and asserts the click reaches `PUT …/tools/grants` and that a refusal leaves the warning standing; `finance-health.test.ts` gains the "Connected to null" regression for both providers.

**E2E** — `tool-grant-from-connect-page.spec.ts` drives the real Rust host: click Grant, **reload the page**, require the warning to stay gone, then read the grant back through the host's own API and assert it is in `allow` and not in `manifestAllow`. Gated on the live-brain lane because both surfaces gate their warning on `inBuild`, which for search and hosting is `cfg!(feature = "openhuman")`; the route itself is covered without a feature.

## Behaviour changes

- New routes `GET`/`PUT`/`DELETE {scope}/tools/grants`.
- `CompanyRecord` gains `overlay_tool_grants` (`#[serde(default)]`, so no migration — existing rows load as "the manifest still decides").
- `Tools` / `ComposioTools` gain `PartialEq`, for the carry rule's one question.
- Docs: `docs/spec/runtime/api-write-plane.md`, `tools.md`, `finance-console.md`.

## Review round 2 (@oxoxDev)

The Major finding was real and is fixed in `f231f44d`: the fold landed in a
local manifest clone at the save site, but `effective_grants` — whose result
becomes the tool provider's grant list, and which `call_tool` enforces against —
and the hosted brain's `tool_catalog` both read `self.manifest.tools.allow`
earlier in the same `build()`. The record load moves to the top of `build` and
the grants fold into `self.manifest` itself, matching the `[workflows].enabled`
shape; the comment claiming every other `self.manifest` reader is untouched by
that merge was true for `[workflows]` and is now corrected for `[tools]`.
`a_console_tool_grant_reaches_the_grant_list_the_provider_enforces` asserts
`effective_grants` on the rebuilt manifest and fails against the old shape.

All three Minors are fixed too — the finance panel's `canManage` gate, the e2e
fixture leak (`beforeAll`/`afterAll` revoke, verified by running the file twice
against one persistent host and reading `added: []` back), and the spliced doc
blocks in `store/export.rs` / `store/fs.rs` plus the docstring my export test cut
in half. The fingerprint asymmetry is recorded in a comment rather than changed.

The two body inaccuracies are corrected above: the route-test count and the
package manager.

## Deliberately out of scope

Console-side **narrowing** of `[tools].allow`. This layer can only widen; taking capability away already has a home in the per-desk ceiling (`overlay_desk_tools`), which is bounded by the allow-list rather than competing with it. A console that could quietly withdraw a capability version control confers would be a second, invisible authority over the same field.


oxoxDev (changes requested):
Read the grant route, the ports types, the builder fold, the harness `ensure` path, the store/export/conformance changes and all four new test files closely; skimmed the ~75 files that only add `overlay_tool_grants: None` to a struct literal.

The design is right and the harness path is correct — `HarnessPool::ensure` installs the effective list onto the fresh company's manifest before the roster is built. One gap in the runtime path, which I think blocks.

## Major — the fold does not reach two readers of `[tools].allow`

The console grant lands in a **local** manifest clone (`src/runtime/builder.rs:3412-3414`) that goes into the saved record. But `self.manifest.tools.allow` is read twice earlier in the same `build()`:

- `:1608` — `effective_grants(&self.manifest)` becomes the grant list handed to `StubToolProvider`/`OpenHumanToolProvider` at `:1814`/`:1819`, and that provider is what `call_tool` enforces against.
- `:3338` — the hosted brain's `tool_catalog` maps `self.manifest.tools.allow`.

Failure scenario: hosted company, `[tools].allow = ["*"]`, `BrainMode::Hosted` with a credential. Operator opens Settings → Search, clicks Grant. The route stores the override and folds `search` into the record's manifest; the grants endpoint and the search status route both report granted and the warning clears. The brain's `tool_catalog` still lacks `search`, so it is never advertised, and the provider's grant list still lacks it, so a `search.web` frame is refused as ungranted. A restart does not fix it — the rebuild re-parses `company.toml` into `self.manifest` and folds only into the record it saves, so both readers stay seed-only permanently.

This is the shape the PR body names as the thing to avoid: a console saying "Connected" over a harness that wired nothing.

Not a drop-in fix — `:1608` runs ~900 lines before the overlays load at `:2544`, so the `[workflows].enabled` precedent doesn't transfer. Either provider construction moves below the overlay read, or `grants`/`tool_catalog` are recomputed from the folded list. Either way the comment at `:2610-2612` ("every other `self.manifest` reader … reads fields this merge never touches") was true for `[workflows].enabled` and is now false.

## Minor

- `frontend/test/e2e/tool-grant-from-connect-page.spec.ts:74,:106` — both specs grant against the shared `e2e_harness` fixture and never revoke. With `reuseExistingServer: !CI`, a second local `e2e:live-brain` finds `search` already granted, so `search-not-granted` never renders and the first spec fails on its 30s wait. And with `fullyParallel: false, workers: 1`, every later spec in the lane runs against a fixture that now grants `search` and `hosting` — which the spec's own docstring says is withheld on purpose so a wildcard can't buy a metered request. An `afterAll` issuing `DELETE …/tools/grants` closes both.
- `src/store/export.rs:161-171` and `src/store/fs.rs:754-762` — the new field was spliced below the neighbouring doc block, so in both structs that paragraph now documents `overlay_tool_grants` and `overlay_desk_tools` is left undocumented. Same splice at `export.rs:2087` leaves an orphan doc line as another test's entire comment. (`OverlayBlob` in `ports/types.rs` got this right.)
- `src/runtime/builder.rs:2554` — rustfmt joined a continuation, leaving ~22 spaces mid-sentence in the dropped-grant log line. The siblings at `:2533` and `:2577` are split correctly.

## Notes

- The two earlier Codex comments (preserve seed grants on DELETE; strip folded grants from the export) were correct against `d9b487b5` and are already fixed by `105312d3` with regression tests. Stale, no action. The two from 11:27 look right and I haven't re-litigated them.
- `src/harness/built_in/mod.rs:1924`,`:2101` — `grants_fp` and the roster fold both base on `company.manifest.tools.allow`, which is now runtime-mutable. A `DELETE` landing after `refresh_record` snapshots leaves the fingerprint unmoved, so revocation waits a cycle. Bounded, and arguably the same window every axis has — not calling it a defect, but the policy overlay doesn't have this asymmetry and it's worth a sentence in the code.
- Test quality is good: the route tests, carry-rule/rebuild tests, export round-trip, `tool-grant-control` and the two `finance-health` "Connected to null" regressions would all fail against main. None is green-for-the-wrong-reason.
- Only red lane is Console E2E on `chat-detached-post-failure.spec.ts:79` — the known flake, not this PR. Live-brain is green, so the new spec did run.
- Body says 14 route tests (there are 16) and reports `pnpm typecheck`/`pnpm vitest`; this console is npm.

CodeGhost21:
Thanks — the Major was real, and the trace was exact. Fixed in `f231f44d`.

**The two readers.** Confirmed both: `effective_grants(&self.manifest)` feeding `StubToolProvider`/`OpenHumanToolProvider`, and the hosted brain's `tool_catalog`. The record load moves to the top of `build` — nothing between the old and new positions touches that store, and no `store.save` intervenes, so `existing` is identical and the first-boot gate at `existing.is_none()` still holds — and the grants fold into `self.manifest` itself rather than a clone. That makes it the same shape as the `[workflows].enabled` merge, and the comment there ("every other `self.manifest` reader … reads fields this merge never touches") now says explicitly that this is true for `[workflows]` and not for `[tools]`, which is why the tool-grant fold runs at the top instead.

`a_console_tool_grant_reaches_the_grant_list_the_provider_enforces` asserts `effective_grants` on the rebuilt manifest — the provider's actual input — and I checked it goes red against the pre-review shape rather than trusting that it would.

**Minors, all three.**

- `canManage` now lives on `ConnectionPanel`, defaulted closed, resolved per page the way `OAuthView` does. Tests for admin / member / not-in-build. (Same round also made the prop **required** on `GrantNamespace` after Codex caught the identical thing on Hosting and Search — a permission that defaults open is the bug, so the compiler now asks.)
- `beforeAll` **and** `afterAll` revoke. `afterAll` alone still leaves a run that inherits a dirty host — killed mid-suite, or `reuseExistingServer` — asserting a precondition it cannot rely on. Verified against the scenario rather than by inspection: started one host, ran the file twice against it, both green, then read the endpoint back — `added: []` and an allow-list byte-identical to `companies/e2e_harness/company.toml`.
- Doc blocks in `store/export.rs` and `store/fs.rs` reunited with their fields, and the `a_bundle_with_duplicate_budget_overrides_is_rejected` docstring my export test had cut in half is whole again.

**Fingerprint asymmetry** — recorded in a comment at `harness/built_in/mod.rs` rather than changed, in your framing: the policy axis reads both halves live, this one takes its base from a manifest field the fold has made runtime-mutable, so a withdrawal can trail a grant by a cycle. Worth a reader knowing that "works exactly like the policy axis" is not quite true.

**Body corrections** — route-test count (18, not 14) and the package manager: you're right, this console is `npm`. That was a habit from a different repo in the workspace; I've re-run the frontend checks under `npm ci` and the table now reports those numbers.

On the stale Codex pair: agreed, both were fixed by `105312d3` with regression tests. The two from 11:27 were also real — the `canManage` default and a retried `PUT` re-stamping `set_by`/`at_millis` over the original operator's grant — and are fixed in `43c62153`.

Only red lane remains Console E2E on `chat-detached-post-failure.spec.ts` (#1789). Untouched here deliberately: that issue's acceptance asks for repeated-run evidence and rules out timeout/retry fixes, so it wants its own PR.

CodeGhost21:
## Manual verification

Driven end to end in the **desktop app** (`build_commit f231f44d0dba`, `--features opencompany/acp,composio,paypal,chargebee`) against a real company — `agentic_marketing_agency`, whose `[tools].allow` is the globals default plus `chargebee`, so it grants neither `hosting` nor `paypal` without any fixture rigging.

### 1. Settings → Hosting — the one-click grant

| | |
|---|---|
| Before | `inBuild: true, granted: false` — warning ends *"Granting it here widens what this company's teammates can do"* + **Grant hosting** |
| Action | clicked **Grant hosting** |
| After | toast naming when it bites; warning gone; `granted: true` |
| Reload | grant still in force — it is in the store, not tab state |

On `main` that warning ends *"it cannot be fixed from this page"* with no control, and a packaged app has no `company.toml` to edit.

### 2. Finance → Wallet — both halves on one panel

The full arc, in order:

1. **No credential, ungranted** → panel reads **"Not connected"**, credential form open.
   This is the `Connected to null` fix. On `main` this exact state renders *"Connected — but no teammate can use it"* (and for Chargebee, *"Connected to null — …"*), because `!granted` was tested before asking whether any credential existed.
2. **Sandbox client ID + secret saved** → wallet reads a live sandbox balance and a transaction, so the connection is genuinely working.
3. **Connected, ungranted** → **"Connected — but no teammate can use it"**, `Attention` badge, *"This company does not grant `paypal`, so wallet tools reach no teammate."* + **Grant paypal**.
   Same sentence as (1) on `main`, but here it is *true* — which is the point of the precedence fix.
4. **Clicked Grant paypal** → **"Connected · sandbox"**, green badge, warning and `Attention` gone.

### 3. The layering held throughout

```
added        : ['hosting', 'paypal']                              ← console grants
manifestAllow: ['*', 'workspace.*', …, 'chargebee']               ← seed, never polluted
allow        : ['*', 'workspace.*', …, 'chargebee', 'hosting', 'paypal']
setBy        : <the signed-in admin's user id>
```

`manifestAllow` is unchanged after two grants, so the console never claims credit for a seed grant and `DELETE …/tools/grants` stays able to withdraw exactly what it added. Both grants are attributed.

### 4. Also exercised, in a standalone host

A throwaway company granting **none** of the five (`allow = ["mcp:*", "workspace", "workspace.*", "web"]`, host built `--features openhuman,chargebee,paypal,composio`): all five reported `inBuild: true, granted: false`; **Settings → Search** granted and survived a reload; **Finance → Invoicing** rendered **"Not connected"** rather than `Connected to null`.

### Not manually exercised

The **Composio** grant control (`ComposioSection` / the provider grid) — covered by unit tests and by the closed-list route tests, but not clicked, because every company to hand already grants `composio`. The non-admin path is unit-tested only (`does not offer a non-admin a control whose write is admin-only`, `treats a host with no user plane as a non-admin`).

### Automated

| | |
|---|---|
| `cargo test --locked` | 3902 passed |
| `cargo test --locked --features openhuman --tests` | 5616 passed |
| `cargo test --locked --features sqlite --lib store::sqlite` | 53 passed |
| `npm test` | 3272 passed |
| `playwright test` (default lane) | 334 passed, 26 skipped |
| `PW_LIVE_BRAIN=1 playwright test tool-grant-from-connect-page` | 2 passed — run **twice against one persistent host**, then the grants endpoint read back `added: []`, confirming the `beforeAll`/`afterAll` revoke leaves the shared fixture as `company.toml` declares it |

Each regression test was checked against the pre-fix shape rather than assumed to discriminate: the provider-path test, the `DELETE`-seed-grant test, the export round-trip and the attribution test all go red when the fix is reverted.

CodeGhost21:
Fixed in `a2bb92fe` — the third Codex finding (`tool_grants.rs:84`, *"Make grants live on non-harness cognition paths"*) was correct, and I'd missed it.

**The bug.** `HarnessPool::ensure` is the only live grant refresh, and it is where the new fingerprint axis hooks in — but it lives behind `openhuman` and serves companies on the **harness** cognition path. On the hosted, sidecar or echo path there is no `ensure`: `RuntimeBuilder::build` snapshotted the provider's grants and the hosted brain's `tool_catalog` at construction. So the route stored the grant, every status route began reporting `granted: true`, the belts stayed as they booted, and the response said *"on the next turn"* regardless.

That is the connect page asserting reach the runtime does not deliver — this issue reproduced one layer inside its own fix. Worth being blunt about: the PR body claimed folding into `self.manifest` meant "one place to be right", and that was true for the readers, not for the *delivery*.

**The fix.** `apply_to_runtime` resolves what actually happened, and `takesEffect` carries one of three answers rather than a constant:

| Path | Answer |
|---|---|
| harness | **next turn** — the fingerprint carries it |
| hosted / sidecar / echo, rebuilder wired | **now** — runtime rebuilt in place through the same issue-#290 seam `PUT …/inference` uses |
| no rebuilder, or the rebuild failed | **restart** |

A failed rebuild is logged rather than raised: the grant is stored and durable either way, and the company is running exactly what the restart wording describes — reporting that honestly beats a 500 over a write that landed. Withdrawals take the same path, and arguably need it more, since a revocation the runtime has not picked up is a capability the operator believes they removed and the agents still hold.

**On the test.** `the_response_says_when_the_grant_bites` asserted the old promise, and its fixture — echo brain, no rebuilder — is precisely the configuration that cannot honour it. It was green because it asserted the bug. It is replaced by four tests pinning each answer to the state that produces it (`a_grant_on_a_non_harness_runtime_reports_restart_not_next_turn`, the withdrawal twin, and two that keep the read/no-op wording), so flattening these back into one constant fails.

No console change was needed: it already rendered `takesEffect` verbatim, which is the payoff for showing the host's words instead of paraphrasing them.

`cargo test --locked` 3905 · `--features openhuman --tests` 5619 · `npm test` 3272 · fmt + both clippy lanes clean.

The two other comments on `f231f44d` are re-posts of findings already fixed in `105312d3` (export strip) and `43c62153` (idempotent-grant attribution) — stale, no action.

@oxoxDev this also touches the timing sentence you'd have read in the module doc, so it's worth a second look alongside the Major fix from the last round.

Issue #1738, opened 2026-08-25
Desktop excludes Composio on a rationale its own credential resolver contradicts, leaving every connector dark


graycyrus:
## What an operator sees

On a desktop build, `Connections` shows eight provider tiles — Gmail, Google Calendar, Google Drive, GitHub, Slack, Notion, Linear, Discord — every one of them reading **"not available here"**, above a banner:

> Composio's provider catalog could not be fetched (Composio is not compiled into this build), so this is a built-in starter list and may be incomplete.

And directly above that, the card invites the operator to do the thing the build cannot honour:

> "…or paste a Composio token below to use a Composio account of your own."

So the surface offers a BYO token, and the build has no code to use it.

## Why it is excluded

`src-tauri/Cargo.toml:56-58`:

```
# * `media`, `composio`, `search` and the rest of the managed backends — they
#   need a platform credential the desktop has no way to hold, and each fails
#   closed anyway.
```

## Why that rationale looks stale

`composio::resolve_credential` resolves over **three tiers**, and the platform credential is the *last* of them. From `src/harness/built_in/build.rs:494-495`:

```
//     `HarnessPool::ensure` through `composio::resolve_credential` — the BYO
//     `composio/token` override, else the company's own TinyHumans key, else
```

and again at `:525` — *"a resolver outcome over three tiers (BYO `composio/token`, …)"*.

Tier one is a **BYO token**, which is precisely what a desktop operator can hold and what the connections UI already asks them to paste. The exclusion's premise — that the credential must come from the platform — is not what the resolver does.

## It builds

Adding the features resolves and compiles clean:

```
opencompany = { path = "..", default-features = false,
                features = ["sqlite", "platform-jwt", "oauth", "mcp", "composio", "media", "acp"] }

cargo check --manifest-path src-tauri/Cargo.toml --bin opencompany-desktop
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 35s
```

Zero errors. So the exclusion is a product decision, not a technical constraint.

## What this costs

The desktop app's entire integration surface is dark, and the reason shown to the operator ("not compiled into this build") is one they cannot act on. Eight connectors and a token field that leads nowhere is a worse first impression than not offering them at all — which is the same principle `frontend/src/api/setup.ts:58` already states: the flow should "say 'not in this build' instead of offering a switch that does nothing."

## What to decide

This issue is a request for a decision, not a foregone fix. Either:

[...]

PR #1837, opened 2026-08-26
Adopt the winner of a user-materialization race instead of refusing

### LINKED ISSUE #1833: Desktop first boot can refuse its own operator: local_owner_record races itself and the console shows "Unreachable"
On a first launch of the desktop app, the console can come up with the host row reading **"This computer — Unreachable"** and the page saying **"There is no sign-in here"**, while the embedded host is healthy and listening. A restart fixes it permanently, which is what makes it easy to dismiss as a fluke.

It is a check-then-act race in `local_owner_record`, and the first boot is exactly when it is most likely to fire.

## What the host logs

```
10:48:46.366765 INFO  opencompany::desktop: seeded the first-run company company=agentic-marketing-agency
10:48:46.367149 INFO  embedded: embedded host listening address=127.0.0.1:50102 companies=1
10:48:47.834852 WARN  server::graphql::auth: could not materialize the none-mode local owner
                      company=agentic-marketing-agency error=conflict: another user already has the email local:owner
10:48:47.835919 WARN  server::graphql::auth: could not materialize the none-mode local owner
                      company=agentic-marketing-agency error=conflict: another user already has the email local:owner
```

`/healthz` answers 200 throughout. The host is fine; it is refusing the only identity it has.

## The mechanism

`src/server/users/routes.rs:443` — `local_owner_record` is find-then-create, and **generates a fresh random id on every call**:

```rust
if let Some(user) = runtime.users().find_user_by_email(id, &key).await? {
    return Ok(user);
}
let user = UserRecord { id: generate_id(), email: key, /* … */ };
runtime.users().upsert_user(id, &user).await?;
```

`upsert_user` is not the problem — it is correct. `src/store/fs_ops.rs:250-270` takes a path lock and rejects a record whose email is already held **by a different id**, which is the invariant that keeps `find_user_by_email` unambiguous.

So when N requests arrive concurrently on a cold store:

1. all N call `find_user_by_email` → `None`;
2. each builds a record with its **own** `generate_id()`;
3. the first `upsert_user` wins;
4. every other one is a *different* id claiming a taken email, so the store correctly returns `Conflict`.

The loser is not "a duplicate user" from the store's point of view — it is a second identity claiming the same address, which is precisely what that guard exists to stop. The caller is asking the wrong question.

## Timestamps confirm it is a race, not a stale read

From the `users.json` written during the failing boot:

```json
[{"id":"01a03dafd1ca-000000000011","email":"local:owner","role":"admin",
  "status":"active","createdAtMillis":1787741327818, …}]
```

`1787741327818` is **10:48:47.818** — 1.4s *after* the seed, and **17ms before** the two failures at `.834`/`.835`. So the record was not created by seeding and then missed by a stale read; three callers interleaved, one won at `.818`, and the other two lost 16ms later. The record is left correct on disk, which is why the next launch is clean.

## Why it matters more than a warning

`auth.rs:328` turns the error into `LocalOwnerOutcome::GatesRefused`, deliberately:

> a store outage while materializing the one identity `none` mode has must not read as "not `none` mode" and must not leave a bearer path to fall through to either.

That is the right call for a store outage. But a lost race is not a store outage — it means the identity **exists and was just created**. Refusing there turns a recoverable race into the desktop app's first impression: an operator who cannot get in, told the machine they are sitting at is unreachable.

## Reproduction

```sh
cd <checkout>
rm -rf target/desktop-dev && mkdir -p target/desktop-dev
OPENCOMPANY_DATA_DIR=$PWD/target/desktop-dev ./scripts/desktop-dev.sh
```

Timing-dependent — it reproduced on a first boot and not on the next two. A cold/slow store widens the window. Any `none`-mode host is affected; the desktop is simply the one that is always `none`-mode and always cold on first run.

## A direction, not a diagnosis

The mechanism above is verified; this part is a suggestion. `Conflict` from `upsert_user` on this path means "someone else materialized the owner first", so re-reading by email and returning that record would make the operation idempotent — the standard compare-and-retry. That keeps the genuine store-outage refusal intact, since only `Conflict` would be retried.

Worth checking whether the same shape exists wherever else a singleton identity is materialized on demand.

Found while verifying #1738 (PR #1830); unrelated to that change — nothing feature-gated touches user materialization, and it reproduces independent of `acp`/`composio`.

CodeGhost21:
Fixes #1833.

## The bug

Two functions in `src/server/users/routes.rs` materialize a user on demand, and both are find-then-create with a **fresh `generate_id()` on every call**:

```rust
if let Some(user) = runtime.users().find_user_by_email(id, email).await? { return Ok(user); }
let user = UserRecord { id: generate_id(), email: /* … */ };
runtime.users().upsert_user(id, &user).await?;
```

That pair is not atomic. When N callers arrive together on a cold store they all miss the lookup, each mints its own id, and each presents a different id for one address.

**The store is not at fault.** `fs_ops.rs` takes a path lock and rejects an email already held by *another id* — the invariant that keeps `find_user_by_email` unambiguous. The caller asked the wrong question: "may I create this user", when what it needed was "who owns this address now".

## Why it mattered

On a desktop first boot, three requests raced `local_owner_record`. One won at `10:48:47.818`; two were refused 16ms later:

```
WARN server::graphql::auth: could not materialize the none-mode local owner
     error=conflict: another user already has the email local:owner
```

`graphql::auth:328` turns that into `GatesRefused`, so the console reported the **healthy host it was talking to** as "This computer — Unreachable", with "There is no sign-in here" and no way forward. `/healthz` answered 200 throughout.

A restart cured it permanently — the record exists by then — which is exactly what made it read as a fluke rather than a bug.

## The fix

`insert_or_adopt` reads a `Conflict` as "somebody else got there first" and returns the winner. That is the idempotence both call sites already **claim** in their own docs — `local_owner_record` says outright:

> Idempotent: the record is keyed by `LoginIdentity::Local`, so every request after the first is a read.

It is safe by construction here: `LoginIdentity::Local` is documented as *"the single implicit owner of a company that has no login at all"* — a reserved synthetic key, not a mailbox — so a `Conflict` on `local:owner` can only mean the local owner already exists.

**`upsert_from_eligibility` is fixed in the same change.** It has the identical shape on the email sign-in path — the one real users authenticate through — and leaving it would be knowingly shipping a latent copy of the bug just fixed.

### What is deliberately not changed

- **Only `Conflict` is adopted.** Every other error still propagates, so the store-outage refusal `graphql::auth` deliberately makes fatal stays fatal: a store that cannot be read must not read as "this user is fine".
- **A `Conflict` with nothing behind it returns the original error** rather than retrying. That means the winner was deleted between the write and the re-read, and looping would spin against a store doing something this function has no business papering over.

## Tests

A race cannot be proven by a racing test alone, so there are two:

| Test | Role |
|---|---|
| `a_lost_materialization_race_adopts_the_winner` | **Deterministic.** Stages the exact losing state — address held, by an id this caller did not mint — and asserts the winner comes back, and that only one record survives. |
| `concurrent_local_owner_materialization_yields_one_record` | 16 concurrent callers on a multi-thread runtime. Timing-dependent by nature, so it is the companion rather than the proof; it catches a regression that reintroduces the shape. |

Both **fail against the pre-fix code**, with the production error string:

```
no racer is refused its own company owner: Conflict("another user already has the email local:owner")
```

I verified that by reverting only the adopt arm and re-running — the failure is the one from the desktop log, not an artificial one.

## Verified locally

- `cargo test --lib server::users` — **154 passed, 0 failed**
- `cargo clippy -p opencompany --all-targets --no-deps -- -D warnings` — clean

Found while verifying #1738 (PR #1830); independent of that change and deliberately kept out of it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple user creation requests occur simultaneously.
  * Conflicting requests now safely reuse the existing user record instead of failing.
  * Concurrent local-owner creation now converges on a single user record without errors or crashes.

* **Tests**
  * Added coverage for concurrent user creation and record adoption scenarios.

### DIFF
--- a/src/server/users/routes.rs
+++ b/src/server/users/routes.rs
@@ -389,6 +389,60 @@ async fn eligibility(
     Ok(invite.filter(|i| i.is_redeemable(now)).map(|i| i.role))
 }
 
+/// Write `user`, or adopt the record that reached its address first.
+///
+/// Every caller below is find-then-create: look the address up, and mint a
+/// record when it is absent. That pair is not atomic, and each caller mints a
+/// **fresh `generate_id()`**, so two requests arriving together both miss the
+/// lookup and then present two different ids for one address. The store is
+/// right to refuse the second — `upsert_user` holds a lock and rejects an email
+/// already held by another id, which is the invariant that keeps
+/// `find_user_by_email` unambiguous — but the caller's question was wrong. It
+/// asked "may I create this user", when what it needed to know was "who owns
+/// this address now".
+///
+/// So a `Conflict` here is read as "somebody else materialized it", and the
+/// winner is returned. That is what makes the operation idempotent under
+/// concurrency, which is what both callers' doc comments already claim.
+///
+/// Issue #1833. On a desktop first boot three requests raced
+/// [`local_owner_record`]; one won and two were refused 16ms later, and
+/// `graphql::auth` turns that refusal into `GatesRefused` — so the console
+/// reported the healthy host it was talking to as "Unreachable" and offered no
+/// way in. A restart fixed it permanently, because by then the record existed.
+///
+/// **Only `Conflict` is adopted.** Every other error still propagates, so the
+/// store-outage refusal that `graphql::auth` deliberately makes fatal stays
+/// fatal: a store that cannot be read must not read as "this user is fine".
+///
+/// A `Conflict` with nothing behind it returns the original error rather than
+/// retrying. That means the winner was deleted between the write and the
+/// re-read, and looping on it would spin against a store doing something this
+/// function has no business papering over.
+pub(super) async fn insert_or_adopt(
+    runtime: &CompanyRuntime,
+    user: UserRecord,
+) -> Result<UserRecord, OpenCompanyError> {
+    let id = runtime.id();
+    match runtime.users().upsert_user(id, &user).await {
+        Ok(()) => Ok(user),
+        Err(OpenCompanyError::Conflict(conflict)) => {
+            match runtime.users().find_user_by_email(id, &user.email).await? {
+                Some(winner) => {
+                    tracing::debug!(
+                        company = %id,
+                        email = %user.email,
+                        "adopted the user record that won the materialization race"
+                    );
+                    Ok(winner)
+                }
+                None => Err(OpenCompanyError::Conflict(conflict)),
+            }
+        }
+        Err(other) => Err(other),
+    }
+}
+
 /// Returns the existing user for `email`, or materializes one from their
 /// eligibility.
 ///
@@ -418,7 +472,7 @@ async fn upsert_from_eligibility(
         last_seen_at_millis: Some(now),
         updated_at_millis: now,
     };
-    runtime.users().upsert_user(id, &user).await?;
+    let user = insert_or_adopt(runtime, user).await?;
     // Mark any real invite as redeemed. A manifest-bootstrapped admin has no
     // invite record, so this is a no-op for them.
     if let Some(mut invite) = runtime.users().find_invite_by_email(id, email).await? {
@@ -464,8 +518,7 @@ pub(crate) async fn local_owner_record(
         last_seen_at_millis: Some(now),
         updated_at_millis: now,
     };
-    runtime.users().upsert_user(id, &user).await?;
-    Ok(user)
+    insert_or_adopt(runtime, user).await
 }
 
 // ---------------------------------------------------------------------------
--- a/src/server/users/routes_test.rs
+++ b/src/server/users/routes_test.rs
@@ -2278,3 +2278,130 @@ async fn a_profile_edit_needs_a_session() {
         .unwrap();
     assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
 }
+
+// ---------------------------------------------------------------------------
+// Materialization races (issue #1833)
+// ---------------------------------------------------------------------------
+
+/// A runtime over a company with no `[users] admins`, which is all these need:
+/// `local_owner_record` answers a mode question nobody asks it, so the manifest
+/// only has to produce a store.
+async fn users_runtime(home: &std::path::Path) -> Arc<crate::CompanyRuntime> {
+    let (connections, _sender) = mail_connections();
+    let state = state_from(
+        home,
+        manifest_without_admins(),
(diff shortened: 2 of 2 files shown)


oxoxDev:
Reviewed this after it merged, so this is for the record rather than a gate — nothing here needed to hold it.

The fix is structural, not a narrowed window, which is the thing worth confirming on a race fix. The old shape in both `upsert_from_eligibility` and `local_owner_record` was check-then-act with nothing between the read and the write. `insert_or_adopt` moves the correctness-bearing step to the write: it always attempts the upsert, treats the initial `find_user_by_email` as a warm-cache shortcut, and on `Conflict` re-reads and adopts the winner. Walked two actors through all three backends — fs serialises check-and-write under one held `path_lock` guard; sqlite can only fail via the `users_email` index, since `ON CONFLICT(company_id,id)` absorbs id collisions and the id is fresh per caller; mongo's `update_one` on `user_id` with `upsert:true` is the same logic. Both racers converge on one record and no caller gets a false "created".

Also checked the loser's data isn't silently dropped: the discarded fields are synthetic materialization defaults, and `role` is a pure function of `eligibility()`, so two racers can't compute divergent roles.

`concurrent_local_owner_materialization_yields_one_record` is a genuine interleaving test — 16 real `tokio::spawn` racers on a 4-worker multi-thread runtime against the actual `local_owner_record`, not the sequential call-twice pattern that would pass either way. Against main the losers get a raw `Conflict` and trip the `expect`.

Two things worth a follow-up, neither blocking:

- The new concurrency coverage only exercises `FsCompanyStore` (`state_from` hard-codes it). The sqlite/mongo `Conflict` mapping is correct on inspection but isn't exercised concurrently, so a regression in `is_unique_violation`/`is_duplicate_key` wouldn't be caught here.
- `src/store/sqlite.rs:1223` `is_unique_violation` matches any `ConstraintViolation`, not the email index specifically. Provably safe at this call site today, but `insert_or_adopt` now converts that whole class into "adopt the winner" — add a NOT NULL or CHECK constraint to `users` later and a real write failure would silently return an unrelated existing record. Latent, worth a comment at minimum.

Out of scope and confirmed not gaps: cross-process fs races stay open (`path_lock` is an in-process registry, #386), and `admin.rs:431`'s invite creation has its own separate check-then-act.

PR #1857, opened 2026-08-26
feat(harness): brand the web_search step label with its engine

sanil-23:
## Summary

The live step timeline showed every search as a humanized, provider-less "Web search". This brands the `web_search` row with the engine behind it, so a cycle's action transcript reads `Exa web search — query: …` instead:

- **Managed tool** (`src/harness/built_in/search.rs`): `WebSearchTool` overrides `Tool::display_label` to return `"Exa web search"` — the engine serving the managed platform's search surface.
- **BYO alias** (`src/harness/built_in/search_byo.rs`): `AliasedTool` presents each provider's tool under the canonical `web_search` name, but its label fell through to the trait default, which humanizes the alias name. It now carries the provider's own label: "Brave web search", "Exa web search", "Querit web search", "SearXNG web search".

No frontend or step-folding changes: `label_for` already prefers `display_label` over the humanized tool name, so every step row, live timeline, and persisted run trace picks the label up as-is.

This is the minimal attribution surface after the #1714 footer approach (issue #1695) was not taken: the engine is named where the operator watches the search happen, and nothing is stamped onto workspace documents.

## API Or Behavior Changes

Operator-facing only: the step-timeline label for `web_search` calls changes from "Web search" to the engine-branded label above. Tool names, schemas, and outputs are unchanged.

## Tests

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --features openhuman --lib -- -D warnings`
- [x] `cargo test --features openhuman --lib "built_in::search::"` (22 passed, incl. the extended contract test pinning the managed label)
- [x] `cargo test --features openhuman --lib search_byo` (10 passed, incl. new `the_aliased_tool_step_label_names_the_provider`)

## Documentation

None needed — no route, launcher, or feature-surface change; the label contract is documented on the constants and the new tests.

## Summary by CodeRabbit

* **New Features**
  * Added provider-specific labels for web search tools in tool-step displays.
  * Managed web search is now labeled “Exa web search.”
  * BYO search tools identify their configured providers, including Brave, Exa, Querit, and SearXNG.

PR #1879, opened 2026-08-27
fix(harness): let a tool's own step label reach the timeline

sanil-23:
## Summary

[#1857](https://github.com/tinyhumansai/opencompany/pull/1857) taught the managed web search to call itself "Exa web search" and each BYO belt to name its own provider. Nothing changed on screen — **`Tool::display_label` is never called.**

The turn loop's `ToolStarted` event carries a call id and a tool *name* and nothing else ([`tinyagents/.../events/types.rs`](https://github.com/tinyhumansai/tinyagents)), so the bridge that projects it into an `AgentProgress` fills `display_label` with `humanize_tool_name(tool_name)` — the same "Web Search" for every provider, on every tenant — and `fold_steps` then faithfully renders a label the tool never chose. Grepping the call sites of `display_label` across this crate *and* openhuman finds only wrapper delegations (`audit.rs`, `checkpoint.rs`) and tests. The trait method has been vestigial since the tinyagents migration.

OpenCompany assembles the tool set, so it is the layer that can answer the question the loop does not ask. `StepLabels` captures each tool's label once, at build time — keeping only labels that *differ* from the humanized default — and puts it back onto the event as that event enters the collector.

That collector is the single site where `set_on_progress` is attached in the whole crate, so chat, dispatched tasks, workflow nodes, the confined copilot and ACP are all covered by the one seam. Nothing downstream changed signature: the folded `TurnStep`s, the live stream frames and the durable run trace read the same rewritten event, so they cannot disagree about a step's name.

**Branding on the tool name alone could not have done this.** Every BYO provider is aliased to the one canonical `web_search`, so the name the loop labels from carries no provider at all — only the tool's own label tells a Brave tenant from an Exa one.

## API Or Behavior Changes

A `web_search` row on the step timeline now reads "Exa web search" (managed) or the wired provider's name (BYO: Brave / Exa / Querit / SearXNG) instead of "Web Search". No wire-format, route or schema change; `TurnStep.label` was already a free string.

`StepLabels` deliberately does **not** override a label the loop actually chose for a call — the unavailable-tool row keeps "`<name>` (unavailable)", and a loop that one day computes a real per-call label outranks a build-time snapshot.

That is also this shim's exit condition. `the_vendored_loop_still_labels_a_tool_row_from_its_name_alone` pins the needle in the vendored source, so if openhuman starts asking tools for their own labels the test fires and says to delete this rather than let it shadow the real answer.

## ⚠️ Depends on #1877

`Rust (openhuman, tinymemory)` does not compile on `main`: `chat_seed.rs`'s `CompanyRecord` test literal is missing three fields. [#1877](https://github.com/tinyhumansai/opencompany/pull/1877) fixes it and is still open. As that PR notes, the break only surfaces on a PR touching `src/harness/**` — which this one does, so it will read as this PR's fault. **Please land #1877 first**; the local runs below were done with its three lines applied and then reverted, so this branch carries none of them.

## Tests

Tests now assert through `fold_steps` rather than on the trait method — the gap that let #1857 ship green while changing nothing an operator sees.

- 8 new tests in `steps.rs` (capture rule, the rewrite, per-provider labels under one tool name, pass-through, loop-chosen labels surviving, the vendored-source pin)
- `search.rs` / `search_byo.rs` label assertions extended to fold a real `ToolCallStarted` end to end

Run locally with `--features openhuman` (the harness is feature-gated, so the default lanes below do not compile this code):

- [x] `cargo clippy --no-deps --features openhuman --all-targets -- -D warnings`
- [x] `cargo build --all-targets` (via `cargo check --features openhuman --lib --tests`)
- [x] `cargo test --features openhuman --lib` — **5734 passed, 0 failed, 3 ignored**

## Documentation

Module docs in `steps.rs` updated: the "Label" bullet in the security section now records that the loop does not ask a tool for its label and that `StepLabels` restores it, noting both halves stay registry-derived and neither widens what a label may contain. `StepLabels`' own doc carries the why, what it holds, and when to remove it.

### DIFF
--- a/src/harness/built_in/iteration_cap_turn_test.rs
+++ b/src/harness/built_in/iteration_cap_turn_test.rs
@@ -317,6 +317,7 @@ async fn company_agent(
         agent_id: "ceo".to_string(),
         role: "Chief Executive".to_string(),
         budget_usd_daily,
+        step_labels: crate::harness::steps::StepLabels::from_tools(agent.tools()),
         agent: tokio::sync::Mutex::new(agent),
         bound_chat: tokio::sync::Mutex::new(None),
     }
--- a/src/harness/built_in/mod.rs
+++ b/src/harness/built_in/mod.rs
@@ -602,6 +602,14 @@ pub struct CompanyAgent {
     /// The embedded openhuman session. A [`Mutex`] because a `turn` takes
     /// `&mut self` and one agent must serialise its own turns.
     agent: Mutex<Agent>,
+    /// The curated step labels of this agent's tools, captured from the built
+    /// tool set (see [`StepLabels`](steps::StepLabels) for why the turn loop
+    /// cannot supply them).
+    ///
+    /// Resolved once per agent build rather than per turn: the tool set is fixed
+    /// for the life of a pooled agent, and a rebuild — the only thing that can
+    /// change which search belt is wired — mints a new `CompanyAgent` anyway.
+    step_labels: steps::StepLabels,
     /// The chat/desk thread this pooled agent's in-memory history is currently
     /// bound to (issue #1725).
     ///
@@ -836,6 +844,14 @@ impl CompanyAgent {
         // same turns.
         let turn_company: Option<CompanyId> = stream.as_ref().map(|ctx| ctx.company.clone());
         let (tx, mut rx) = tokio::sync::mpsc::channel::<oh::agent::progress::AgentProgress>(1024);
+        // This agent's curated step labels, restored onto each tool-call start as
+        // it arrives. The turn loop labels a tool row from its *name* alone and
+        // never asks the tool what it calls itself, so a branded belt would
+        // otherwise render as the generic humanized name on every surface below
+        // (see `steps::StepLabels`). Applied here — once, at the single point the
+        // turn's events enter OpenCompany — so the live stream, the durable run
+        // trace, and the folded timeline cannot disagree about a step's name.
+        let step_labels = self.step_labels.clone();
         let collector = tokio::spawn(async move {
             let mut events = Vec::new();
             let mut seq: u64 = 0;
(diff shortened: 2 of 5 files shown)


sanil-23:
## CI: both red checks are pre-existing on `main`, neither is from this PR

**`Rust (openhuman, tinymemory)` — `chat_seed.rs:857`, the break #1877 fixes.**

```
error[E0063]: missing fields `activation_completed_at`, `name_confirmed` and `overlay_tool_grants`
  --> src/harness/built_in/chat_seed.rs:857:9
```

Byte-identical on main's own run [33045807125](https://github.com/tinyhumansai/opencompany/actions/runs/33045807125). This PR touches no file within a mile of it. Land #1877 and this goes green.

**`Console E2E` — one test, failing on `main` since at least 2026-08-26.**

`chat-detached-post-failure.spec.ts:79 › a chat POST killed in flight still shows the reply the host went on to write` (334 passed, 1 failed, 26 skipped). The same single test, in the same lane, on three consecutive main runs that predate this branch: [33045807125](https://github.com/tinyhumansai/opencompany/actions/runs/33045807125), [33039396552](https://github.com/tinyhumansai/opencompany/actions/runs/33039396552), [33009410010](https://github.com/tinyhumansai/opencompany/actions/runs/33009410010).

It also cannot be this PR by construction: that lane runs the **`opencompany-debug-binary`**, built by `cargo build --locked --bin opencompany` with no `--features openhuman`. Every line this PR changes is under `src/harness/`, which that binary does not compile. The lane that *does* carry this code — **`Console E2E (live brain)`**, on the gated binary — passed (10m26s), as did `Rust`, `Rust (mail)`, `Rust (mongodb)`, `Desktop` and `Gated host binary`.

Happy to file the E2E one separately rather than let it keep reading as each harness PR's fault.

Locally, with #1877's three lines applied and then reverted: `cargo test --features openhuman --lib` → **5734 passed, 0 failed**; clippy `-D warnings` and `cargo fmt --check` clean.

PR #1884, opened 2026-08-27
policy: declare the three hosting tools the vendor pin brought in

CodeGhost21:
## What this is

`src/harness/built_in/hosting.rs:161` returns `Account::tools()` wholesale, so the hosting tools wired onto a live agent are whatever `vendor/openhuman` ships. The pin at `01a436d` ships nine. This repository declares six.

The three that arrived with the pin — `hosting_rollback`, `hosting_list_deployments`, `hosting_domain_status` — are live on `main` right now with no row in `crate::policy::consequence::DECLARED`.

## Why that is not harmless

`undeclared()` decides "is this a read?" with `READ_ONLY_PREFIXES` matched by `name.starts_with`. Every name in this family begins with `hosting_`, so `list`/`get`/`read` can never match — the file says so itself at `src/policy/consequence.rs:758`, and pins it in `the_fallback_cannot_see_a_read_verb_behind_a_namespace`. So today:

| tool | what it is | what the gate says |
|---|---|---|
| `hosting_list_deployments` | "Read-only." | `Consequence` — parks, costs an approval |
| `hosting_domain_status` | "Read-only." | `Consequence` — parks, costs an approval |
| `hosting_rollback` | "External effect." | `Other` — no `deploy`/`publish`/`post` substring |

That is issue #1079's defect — reads parking, a live deployment classified `Other` — arriving through the vendor pin instead of through an edit.

## The rows

Classified from openhuman's own `hosting/README.md`, which labels each tool, the same source the six existing rows were argued from:

- **`hosting_list_deployments`** → `Other` / `Nothing`. "Read-only." It is also what hands `hosting_rollback` its deployment id, so parking it charges an approval on the way to every recovery.
- **`hosting_domain_status`** → `Other` / `Nothing`. "Read-only." The read half of `hosting_add_domain`: attaching is the effect, asking is not.
- **`hosting_rollback`** → `Publish` / `Consequence`. The shape the `NOTE` it replaces predicted. The tool marks itself `external_effect()` with the comment *"Changes what the public sees on a live site, so it gates"* — `hosting_launch_site`'s sentence with the build removed.

`hosting_rollback` parks, deliberately, even though rollback is the *recovery* path and parking delays a fix to a broken site. `supervised` exists to put a human in front of a change to what the public sees, and "the site is already broken" is an argument for approving quickly, not for not being asked. An operator who wants it unattended has `always_approve`.

It therefore also needs a `TOOL_LABELS` entry, or `every_consequence_tool_has_a_console_label` fails and the card reads "Use one of its tools" over a live site.

## The guard that should have caught this

`every_hosting_tool_is_declared` was a hardcoded list of six names. **A hardcoded list fails when a row is deleted. It cannot fail when the pin adds a tool** — which is why all three went live and CI stayed green.

This adds `every_wired_hosting_tool_is_declared`, which enumerates the belt `hosting_tools` actually builds and checks it against `declared_tools()`. Reverting just the `DECLARED` rows on this branch makes it name all three:

```
these hosting tools came in with the vendor pin and are wired onto live agents,
but nobody has said what they can reach [...]:
["hosting_list_deployments", "hosting_rollback", "hosting_domain_status"]
```

It is separate from `every_registered_tool_is_declared` because that test **cannot** see these tools: it enumerates a belt from `build_agent`, and hosting is wired only when a company has a *stored credential*, which no unit-test company has. That gap is the structural reason this drifted silently. `Account::connect` builds a client and does no I/O, so the dummy credential never leaves the process; the test has a vacuity guard for the empty-belt path.

The hardcoded list stays as the floor for lanes that do not have the `openhuman` feature.

## Verification

- `cargo test --features openhuman --tests` — 5735 + 32 pass, 0 fail (the CI lane's exact selection).
- `cargo test --lib policy::consequence` at default features — 55 pass.
- `cargo clippy --features openhuman --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- Negative check above: the new test fails on the unpatched table and names exactly the three tools.

Not verified: `frontend` typecheck (no `node_modules` in this worktree). The frontend change is one entry in a `Readonly<Record<string, string>>` literal plus comment text.

## Scope

Three declaration rows, one console label, one test. **No vendor pin change** — the pin already carries these tools, so the cross-repo sequencing #913 warns about is already resolved.

This does **not** deliver #913, which is labeled `phase-2`. `hosting_deployment_logs` is still unbuilt and blocked on tinyhumansai/tinyhosts#4. Refs #913, #1079.

## Summary by CodeRabbit

- **New Features**
  - Added hosting rollback support, allowing an earlier site version to be served.
  - Added visibility into deployment information and domain status.
- **Safety Improvements**
  - Hosting rollbacks now require approval before they take effect.
  - Hosting actions are consistently tracked and declared, improving safeguards and reliability.

### DIFF
--- a/frontend/src/lib/language.ts
+++ b/frontend/src/lib/language.ts
@@ -286,12 +286,12 @@ const TOOL_LABELS: Readonly<Record<string, string>> = {
   // count is the whole difference between it and `edit` above.
   apply_patch: "Edit several files in its workspace at once",
   csv_export: "Save data as a spreadsheet file in its workspace",
-  // Hosting (issue #1079). Only the three that park need words here; the three
-  // read tools are `Reach::Nothing` and never reach an approval card.
+  // Hosting (issues #1079, #913). Only the tools that park need words here;
+  // the read tools are `Reach::Nothing` and never reach an approval card.
   //
   // Each label names the thing an operator is actually consenting to, because
-  // these are the cards where a vague one costs the most: two of them change
-  // what the public sees at an address, and the third can write secrets at a
+  // these are the cards where a vague one costs the most: three of them change
+  // what the public sees at an address, and the fourth can write secrets at a
   // third-party provider. "Use one of its tools" over a live deployment is the
   // failure this issue is about.
   hosting_launch_site: "Deploy a site to the public internet",
@@ -302,6 +302,17 @@ const TOOL_LABELS: Readonly<Record<string, string>> = {
   // afterwards for a build-time variable to take effect, so a label promising
   // one would be describing an act that has not happened.
   hosting_set_env: "Set environment variables on one of its sites",
+  // Worded as what the visitor gets, not as what the provider calls it. The
+  // provider's own verb is "promote", and an operator reading "Promote a
+  // deployment" on a card has to already know that promoting an old build is
+  // how a rollback is spelled. "Serve an earlier version" is the consequence:
+  // the site the public loads changes, and it changes to something older.
+  //
+  // No "(rollback)" gloss, and no mention of recovery. The card is read when
+  // the agent asks, which is not always after a bad deploy — an agent can
+  // propose this from a stale premise, and a label that pre-agrees it is a fix
+  // argues the operator into the approval the card exists to ask for.
+  hosting_rollback: "Serve an earlier version of one of its sites",
   // `mcp_registry_tool_call` is deliberately absent: EFFECT_LABELS already
   // names it and is consulted first, so an entry here would be unreachable.
 };
--- a/src/harness/built_in/hosting.rs
+++ b/src/harness/built_in/hosting.rs
@@ -1,5 +1,11 @@
-//! The agent-facing bridge for hosting (TinyHosts): six tools that put a
+//! The agent-facing bridge for hosting (TinyHosts): the tools that put a
 //! company's workspace on a real hosting provider.
+//!
+//! The count is deliberately not written down here. [`hosting_tools`] returns
+//! `Account::tools()` wholesale, so the set is whatever the vendored OpenHuman
+//! ships and it moves with the pin — a number in this sentence is a number that
+//! goes stale silently. `every_wired_hosting_tool_is_declared` below is what
+//! holds the set to something this crate has actually classified.
 //!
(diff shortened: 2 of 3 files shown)


CodeGhost21:
## The red `Console E2E` lane is #1885, not this diff

Two runs of the **same commit** failed on **two different specs**, and each one is separately documented as failing on `main`:

| run | failing spec | documented in |
|---|---|---|
| [98512145240](https://github.com/tinyhumansai/opencompany/actions/runs/33069555900/job/98512145240) | `chat-detached-post-failure.spec.ts:79` | #1789, #1885 |
| [98520315621](https://github.com/tinyhumansai/opencompany/actions/runs/33069555900/job/98520315621) (rerun) | `chat-thread-avatar.spec.ts:68` | #1885 |

Both runs show the signature #1885 records: **1 failed / 26 skipped / 334 passed**. On the rerun, `chat-detached-post-failure` passed and a different load-sensitive spec took its place — which is the behaviour of a saturated runner, not of a diff.

#1885 lists `chat-thread-avatar.spec.ts:68` failing on `main` in run [33055333344](https://github.com/tinyhumansai/opencompany/actions/runs/33055333344), and `chat-detached-post-failure.spec.ts:79` on `main` in two more.

## Why this diff cannot reach either spec

The change is three things: two rows and a test in Rust policy/harness code, and **one string entry** in `TOOL_LABELS` (`hosting_rollback`). Neither failing spec references `TOOL_LABELS`, hosting, approvals, or the policy table — they are chat SSE delivery and thread-panel avatar rendering. `grep -E "TOOL_LABELS|hosting|approval"` over `chat-detached-post-failure.spec.ts` returns nothing.

The lane that actually exercises this change, **`Rust (openhuman, tinymemory)`**, passes — as does `Console`, which typechecks and builds the frontend edit.

Every other lane is green: `Rust`, `Rust (mail)`, `Rust (mongodb)`, `Console`, `Console (current Node, advisory)`, `Console E2E (live brain)`, `Desktop`, `Gated host binary`, `Actions pinned`, `Changed areas`.

I have not rerun a third time — #1885's point is that rerunning until green is the cost, not the fix.

Issue #1885, opened 2026-08-27
Console E2E: chat-detached-post-failure and chat-thread-avatar flake on main under runner load


CodeGhost21:
## Summary

Two Console E2E chat specs are load-sensitive and fail intermittently **on `main`**, not just on
PRs. Because `Console E2E` is a required-looking lane, every red run costs somebody a triage
pass on a branch that is not actually broken.

## Evidence

Same signature every time — **1 failed / 26 skipped / 334 passed** out of 361:

| Run | Branch | Failing spec |
|---|---|---|
| [33052757324](https://github.com/tinyhumansai/opencompany/actions/runs/33052757324) | `main` | `chat-detached-post-failure.spec.ts:79` |
| [33039396552](https://github.com/tinyhumansai/opencompany/actions/runs/33039396552) | `main` | `chat-detached-post-failure.spec.ts:79` |
| [33055333344](https://github.com/tinyhumansai/opencompany/actions/runs/33055333344) | `main` | `chat-thread-avatar.spec.ts:68` |
| [33060027775](https://github.com/tinyhumansai/opencompany/actions/runs/33060027775) | PR #1847 | `chat-detached-post-failure.spec.ts:79` |

A full local run of the same suite, against a default-feature `cargo build --locked --bin
opencompany` host (what the lane downloads from the `rust` job), is clean:

```
26 skipped
335 passed (5.6m)
```

`retries: 0` for the non-visual lane (`playwright.config.ts:410`), so a local pass is not a
retry masking a failure.

## Why it flakes

`chat-detached-post-failure.spec.ts` cuts a chat POST mid-flight, then asserts the reply held
during the cut is released:

```ts
await expect(reply(page, marker)).toBeVisible({ timeout: 60_000 });
```

The spec's own comment concedes the hazard:

> a saturated CI runner can delay the SSE frame well past the echo brain's millisecond answer

That is a wall-clock race against runner load. The 60s slack narrows the window; it does not
close it. `chat-thread-avatar.spec.ts:68` fails the same way under load.

## Suggested direction

Not a proposed fix — the right shape needs a look by whoever owns the SSE release path:

1. Assert on a **deterministic signal** that the held frame was released (a host-side event, a
   test hook) rather than on wall-clock visibility.
2. Or make the echo-brain harness deterministic for this spec so the frame cannot be delayed by
   runner load at all.
3. Retries are a last resort: they would hide the real detached-POST regression this spec was
[...]

Issue #1846, opened 2026-08-26
Never break mid-task: graceful budget pause on the top-level orchestrator path


oxoxDev:
## Summary
Never let a job break silently when credits run out mid-run. Close the confirmed asymmetry where a delegated sub-agent's budget exhaustion pauses gracefully but the top-level orchestrator turn dies with a generic error. Fixes the 13% churn cause ("job broke mid-task when credits ran out").

Source: OpenCompany pre-launch Retention Strategy, fix 04. Verified: the hypothesized `tinyagents` reclassifier (`format_failure_aggregate`) does NOT exist — the top-level path genuinely misses the graceful halt. Bucket = REAL DEFECT (fix, not suppress).

## Problem
- Delegated sub-agent budget failure is caught by `RepeatedToolFailureMiddleware::after_tool` (`vendor/openhuman/.../tinyagents/middleware.rs:2360`), gated on an envelope marker written only for sub-agent failures → graceful `Ok` reply + `SteeringCommand::Pause`.
- The top-level orchestrator's own inference call is not a tool result, carries no marker, so it returns `Err` → OC `classify_turn` (`src/harness/mod.rs:746`) matches only "empty response" → `AttemptOutcome::Hard(OpenCompanyError::Harness)` → generic HTTP `Err`. Simple non-delegating tasks (the common case) break silently mid-task.
- No pre-task cost/proximity primitive exists — all budget logic is binary at/past-cap.

## Scope (OpenCompany Rust + console; backend read-only)
1. **Close the asymmetry at the OC boundary** (not in vendored code): add `AttemptOutcome::BudgetPaused { summary }`; new `classify_turn` arm — an `Err` satisfying the single existing `is_budget_exhausted_message` (`vendor/.../inference/provider/billing_error.rs:8-35`, READ-ONLY reuse, do not fork) → `BudgetPaused` carrying the same actionable copy the sub-agent halt emits.
2. **Handle `BudgetPaused`** in the `run_inner` attempt-loop (`src/harness/mod.rs:705-722`): emit "paused — add credits to continue", set a budget-paused terminal state (distinct from failed), park a durable re-issue marker (reuse grant-reissue precedent `src/runtime/grants.rs` + `src/harness/policy.rs`). Per #561 (mid-turn checkpointing declined): this is end-turn-cleanly + re-issue on fresh dispatch after top-up, NOT true resume. Align cron budget-halt (`scheduler.rs:401-411`) to the same state.
[...]

PR #1886, opened 2026-08-27
fix(harness,delegation): stop offering budget-pause redemptions that cannot work (#1846)

CodeGhost21:
## Summary

Follow-up to #1847 (merged). Three budget-pause paths surfaced an "Add credits & resend"
CTA the server can never honour, or spent a second inference call on an already-exhausted
provider. All three are still present on `main` at `5037937`.

These were raised as review findings on #1847 and were still open when it merged.

## Problem

The console decides whether to draw the Add-Credits button by `isBudgetPauseNotice`, which
is `text.startsWith(BUDGET_PAUSE_NOTICE_PREFIX)` — there is no structured wire field for a
message's kind, so that prefix IS the contract. Two paths pause with no marker the generic
chat redeem can honour, but emitted the prefix anyway:

1. **Confined workflow copilot** (`confined_turn_bubble`). `run_confined` bypasses
   `run_inner`, the only place that parks a marker, and `CONFINED_AGENT_ID` names no
   addressable teammate. The function's own doc already claimed "with no CTA" — nothing
   enforced it. Every click did a GET returning `null`, then a POST that **404**'d.

2. **Approval continuation.** Runs through `run_steered_background`, so `run_inner` parks
   `background: true` — exactly the shape `redeem_budget_pause` refuses
   (`src/server/ops/budget_pause.rs`). Every click reserved the marker, restored it, and
   returned **400**.

3. **CEO relay after a delegated pause** (`delegation.rs`). "Answered" is the load-bearing
   word: a desk that paused has NOT answered — `desk.reply` is the placeholder — but the
   non-empty check launched the relay regardless. That fired a second call at the same
   exhausted provider, which paused too and parked a **second marker for the responder**
   that no notice points at. Being newer, it superseded the live CTA on the delegate's own
   notice — disabling the one button that would have worked.

## API Or Behavior Changes

- New `BUDGET_PAUSE_NOTICE_NO_RESEND_PREFIX` + `budget_pause_notice_no_resend()`: a sibling
  carrying the same information that deliberately does NOT match `isBudgetPauseNotice`, so
  those notices render as ordinary system bubbles with no unusable action. Paths (1) and (2)
  now use it.
- The CEO relay is gated on `budget_paused.is_none()`. The delegate's text is folded onto the
  operator bubble in the shape `build_relay_prompt` would have used, so skipping the relay
  drops nothing the operator would have read.

Resuming an approval continuation properly needs the grant's own identity, which the generic
chat-message redeem path does not carry. Until it does, a notice with no button is the honest
surface — stated rather than silently deferred.

## Tests

- **Rewrote** `a_confined_turns_budget_pause_is_a_system_notice_not_a_copilot_reply`, which
  asserted the *redeemable* prefix and therefore encoded the defect. Now asserts the
  no-resend prefix **and** the negative.
- `an_approval_continuation_pause_offers_no_redeem_cta`.
- `the_redeemable_and_no_resend_prefixes_are_not_prefixes_of_each_other` — a one-character
  edit to either constant silently restores the broken CTA.
- `a_delegates_budget_pause_does_not_launch_the_ceo_relay` — asserts on recorded calls, not
  parked markers: the marker registry is process-global and keyed by company id, and every
  `Fixture::nested()` shares one id, so a marker assertion passes or fails on test-execution
  order. Scripted with only two turns, so a regression makes `ScriptedTurns` run dry and fail
  loudly rather than quietly parking an extra marker.

Gates, all clean on this branch:

- `cargo check --features openhuman --all-targets`
- `cargo clippy --features openhuman --all-targets --no-deps -- -D warnings`
- `npm run typecheck`; `npx vitest run` — **366 files / 3446 tests passed**

## Documentation

Also documents the deliberate near-miss on `isBudgetPauseNotice`, so the two prefixes are not
"helpfully" merged by a later change — the near-miss is the mechanism.

Includes one comment-only commit correcting a lint-suppression rationale in `ChatView.tsx`
that contradicted the comment thirty lines above it (raised by CodeRabbit on #1847, unaddressed
at merge).

## Note on CI

**Console E2E will likely fail on this PR, and it is not this branch.** `main` currently fails
that lane in roughly four runs out of six, always on the same two chat specs — tracked in #1885
with the run-by-run evidence. #1847 itself merged with that lane red. Please don't re-triage it
from scratch.

Related: #1846, #1847, #1885

### DIFF
--- a/frontend/src/hooks/use-events.ts
+++ b/frontend/src/hooks/use-events.ts
@@ -425,7 +425,21 @@ export type CompanyStreamEvent =
  */
 export const BUDGET_PAUSE_NOTICE_PREFIX = "⏸ Paused — out of credits:";
 
-/** Whether a chat message's text is the unauthored budget-pause system notice. */
+/**
+ * Whether a chat message's text is the **redeemable** budget-pause system
+ * notice — the one that gets an "Add credits & resend" CTA.
+ *
+ * Issue #1846 review (Codex #3870562586 / #3870562590): the host has a
+ * deliberate SIBLING prefix, `BUDGET_PAUSE_NOTICE_NO_RESEND_PREFIX`
+ * ("⏸ Paused — out of credits (add credits, then start this again):"), for the
+ * two paths that pause with no marker this CTA can redeem — the confined
+ * workflow copilot (no marker is ever parked) and an approval continuation
+ * (its marker is parked `background: true`, which the redeem route refuses).
+ * That prefix must NOT match here: those notices render as ordinary system
+ * bubbles precisely so no unusable button is drawn. Do not "fix" the near-miss
+ * by loosening this to a substring check or by matching both prefixes — the
+ * near-miss is the mechanism.
+ */
 export function isBudgetPauseNotice(text: string): boolean {
   return text.startsWith(BUDGET_PAUSE_NOTICE_PREFIX);
 }
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -1118,8 +1118,10 @@ export function ChatView({
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps -- `client`/`company`
     // intentionally excluded: this effect's job is "a new notice appeared",
-    // not "the client changed"; a client/company change already clears
-    // `transcripts` (and therefore this map) via the boot/company-switch path.
+    // not "the client changed". A client/company change is handled by the
+    // reset effect above, which empties this map; `transcripts` resetting in
+    // `AppShell` then re-derives `budgetPauseMessageIdByAgent`, so this effect
+    // re-runs with the new scope's notices.
   }, [budgetPauseMessageIdByAgent]);
 
   // An open thread only makes sense while its parent is on screen; switching
--- a/src/harness/built_in/brain.rs
+++ b/src/harness/built_in/brain.rs
@@ -156,6 +156,40 @@ pub(crate) fn budget_pause_notice(pause: &crate::harness::BudgetPause) -> String
     format!("{BUDGET_PAUSE_NOTICE_PREFIX} {}", pause.summary)
 }
 
+/// The non-redeemable sibling of [`BUDGET_PAUSE_NOTICE_PREFIX`] (issue #1846
+/// review, Codex #3870562586 / #3870562590).
+///
+/// The console renders its "Add credits & resend" CTA off
+/// [`BUDGET_PAUSE_NOTICE_PREFIX`] alone — there is no structured wire field to
+/// key on yet, as that constant's own doc says. Two paths pause WITHOUT a
+/// marker the generic chat redeem can honour:
+///
+/// * the **confined workflow copilot** — `run_confined` bypasses `run_inner`,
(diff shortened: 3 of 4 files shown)


oxoxDev:
Reviewed this after it merged — my fault for being late, so this is for the record and a possible follow-up rather than a gate. Two things I think are still live in `main`.

**The fold the relay-skip is justified on is dead on the production path.**

`handle_operator_message` has exactly one non-test caller (`brain.rs:3290`; every `delegation.rs` call site is in the test module). Sixteen lines later, `brain.rs:3306-3308` does `if turn.budget_paused.is_some() { operator_reply = BUDGET_PAUSED_PLACEHOLDER_REPLY }`. On the skip path `desk_paused == true` implies `budget_paused` is `Some`, so that override always fires and the appended `"\n\n{member} replied:\n{reply}"` is discarded before any bubble is built.

Concretely: operator asks chief → chief hands to engineering → engineer's turn pauses for credits. The console shows `"(no reply — see the notice below)"` plus the pause notice; the engineer's text never appears. So the PR body's "the fold below keeps the delegate's own text on the operator bubble, so nothing the operator would have read is lost" (and the comment at `delegation.rs:1601-1604`) isn't accurate.

This is **not a regression** — on `main` the relay ran and `operator_reply = relay.reply`, which the same override also threw away. The stronger and true justification is: *the relay's reply was already unconditionally discarded whenever a pause folded through, so skipping it loses nothing and saves a call at an exhausted provider.* Worth either deleting the fold and correcting the rationale, or making it actually reach the operator (skip the placeholder override when the reply carries folded desk text). Two test assertions currently pin the wrong rationale — `a_delegates_budget_pause_does_not_launch_the_ceo_relay`'s final `contains("engineer")` and the two new assertions in `a_nested_delegates_budget_pause_reaches_the_operator_turn` all assert an operator-visible property production doesn't surface.

**Codex's P2 on `brain.rs:792` holds, and it reintroduces the class this PR closes.**

`model.ts:69` filters the supersession map through `isBudgetPauseNotice`, so a no-resend notice is invisible to it. Sequence: maya pauses on an interactive turn → redeemable notice N1, marker M1 (`background:false`); a pending approval for maya is approved; the continuation runs `run_steered_background`, pauses, and parks M2 (`background:true`) over M1. Post-merge its notice carries the no-resend prefix, so `latestBudgetPauseMessageIdByAgent` still reports N1 as maya's latest and N1's CTA stays enabled. Clicking sends `?id=M1.id` → `RedeemMatch::Stale` → 409 "refresh and try again", which refreshing never clears. `main` greyed N1 out. No wrong redeem and no double-spend — the `?id=` guard and the atomic reserve both hold — so it's UX-only, but it's a permanently-broken enabled button.

Smaller things: `an_approval_continuation_pause_offers_no_redeem_cta` is a tautology over `format!` (revert `brain.rs:792` and it still passes — the name promises coverage only the other test provides); `delegation.rs:1599` skips `drain_refusals`, losing the `tracing::warn!` for a hand-off aimed at a nonexistent desk; and `brain.rs:10535`'s doc claims a frontend fixture mirror in `budget-pause-notice.test.ts` that doesn't exist.

Budget correctness itself is clean — no double-count, no mis-attribution, markers keyed per-agent, redeem reserves atomically.

PR #1888, opened 2026-08-27
test(e2e): cut the detached-POST connection on an observation, not a guess (#1885)

CodeGhost21:
## Summary

Test-only. Replaces the wall-clock guess in `chat-detached-post-failure.spec.ts` with an
observation, so the spec can no longer confuse "the held frame was released" with "the frame
did not exist yet".

Fixes the diagnosis problem in #1885. **Whether it fixes the flake itself is unproven — see
"Honest status" below.**

## Problem

The spec asserts that a frame `PendingSyncPosts` held is released when the POST throws. That is
only what it tests if the reply was emitted *while the POST was unresolved*, so the frame was
captured. The spec enforced that by waiting a fixed 8s:

```ts
const response = await route.fetch();
await new Promise((resolve) => setTimeout(resolve, CUT_AFTER_MS)); // 8_000
```

Nothing made that true — it was assumed from "the offline echo brain answers in milliseconds".
When it did not hold, the failure surfaced as a 60-second wait for a DOM element, several steps
removed from the cause. In this issue that misdirection cost three wrong hypotheses before the
premise itself was questioned.

## What changed

`awaitJournaledReply()` polls the host until this turn's reply is genuinely journaled, then cuts:

```ts
const response = await route.fetch();
await awaitJournaledReply();
await new Promise((resolve) => setTimeout(resolve, FRAME_SETTLE_MS)); // 1s, one known push
```

- **`page.request` is deliberate.** An `APIRequestContext` is not subject to `page.route`, so it
  reads the real durable transcript while the page stays barred from it by the `chat/history`
  stub. The bar on the page is what keeps the assertions honest; it must not blind the test's own
  premise check.
- **Reuses the hydration request's own URL and `authorization`**, filtered to this thread's
  `?desk=` param — no second source of truth for scope or auth to drift from. (Capturing
  unfiltered points the probe at whichever desk hydrated last, which never sees this turn; caught
  locally.)
- **Matches `You said: <marker>`, not the bare marker.** The operator's own message carries the
  marker and is journaled the moment the host accepts the turn, so matching that cuts before any
  reply exists and defeats the premise. (Also caught locally.)

Two properties now hold regardless of what is behind the CI flake:

- a **slow host delays** this spec instead of failing it;
- a host that **never replies fails it saying so**, instead of as a downstream timeout.

## Honest status

**This does not demonstrate the flake is cured.** The CI failure never reproduced locally — the
spec passed **9/9** before this change, including at load average 58, and the load hypothesis in
#1885's description is retracted ([comment](https://github.com/tinyhumansai/opencompany/issues/1885#issuecomment-5439537467)).
Local passes show no regression, not a fix.

There is even mild evidence against this being the cure: removing the fixed 8s cut per-run time
from ~26s to ~22s, so `awaitJournaledReply` returns in well under a second — the host journals
the reply almost immediately. That makes "the frame did not exist yet" a *weaker* explanation for
CI and shifts suspicion toward delivery.

What it does buy, unconditionally, is the diagnostic split. If this spec still fails in CI it will
now say either:

- **"the host never journaled a reply …"** → a real delivery failure, a product gap; or
- a failure at the DOM assertion *with the reply provably journaled* → the frame existed and
  rendering lost it.

Today those are indistinguishable. That one bit is what tells the next person which half to look
at.

Validation is therefore repeated `Console E2E` runs on this PR compared against main's base rate
(~2 green in 6), not a green local run.

## Tests

- This spec: **8 runs, 8 passes** — 3 idle, 5 at load average 139.
- Full suite: **338 passed / 26 skipped / 0 failed**, `retries: 0`.
- `npm run typecheck:e2e` clean.

No product code is touched.

Related: #1885, #1847, #1886

## Summary by CodeRabbit

- **Tests**
  - Improved end-to-end coverage for chat reply failures by waiting for the reply to be durably recorded before continuing.
  - Added a timeout with a descriptive error when the expected reply is not recorded.
  - Reduced test flakiness by replacing a fixed delay with observable completion checks and a brief stream-settling pause.

### DIFF
--- a/frontend/test/e2e/chat-detached-post-failure.spec.ts
+++ b/frontend/test/e2e/chat-detached-post-failure.spec.ts
@@ -158,8 +158,8 @@ test("a chat POST killed in flight still shows the reply the host went on to wri
    * itself stays barred from it. The bar on the page is what keeps the
    * assertions honest; it must not blind the test's own premise check too.
    */
-  const awaitJournaledReply = async (): Promise<void> => {
-    const deadline = Date.now() + 60_000;
+  const awaitJournaledReply = async (): Promise<string | null> => {
+    const deadline = Date.now() + 45_000;
     let lastStatus = "no history request was captured during hydration";
     while (Date.now() < deadline) {
       if (historyProbe) {
@@ -176,17 +176,13 @@ test("a chat POST killed in flight still shows the reply the host went on to wri
           // accepts the turn, so matching that would cut before any reply
           // existed and defeat the premise this wait exists to establish.
           // Same text `reply()` locates in the DOM.
-          if (body.includes(`You said: ${marker}`)) return;
+          if (body.includes(`You said: ${marker}`)) return null;
           lastStatus = "the host has not journaled this turn's reply yet";
         }
       }
       await new Promise((resolve) => setTimeout(resolve, 250));
     }
-    throw new Error(
-      `the host never journaled a reply for ${marker} within 60s — ${lastStatus}. ` +
-        "That is a real delivery failure, not a timing artefact of this spec: the turn was " +
-        "accepted and the connection was still open, so the reply should exist.",
-    );
+    return lastStatus;
   };
 
   // What was on screen at the moment of the cut, read inside the handler and
@@ -195,6 +191,9 @@ test("a chat POST killed in flight still shows the reply the host went on to wri
   // run dies waiting for an error line that was never going to appear —
   // reporting a premise violation as a timeout somewhere else entirely.
   let repliesAtCut: number | null = null;
+  // Why the premise wait gave up, or `null` if it did not — asserted after the
+  // cut, never thrown inside the handler. See `awaitJournaledReply`.
+  let premiseFailure: string | null = null;
   let cutReady!: () => void;
   const cutReadyPromise = new Promise<void>((resolve) => {
     cutReady = resolve;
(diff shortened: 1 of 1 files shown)


oxoxDev:
Reviewed after merge — late on my side, so this is for the record and a follow-up rather than a gate.

`awaitJournaledReply()` is the right fix for the problem it diagnoses: replacing `CUT_AFTER_MS = 8_000` with an actual observation of the reply being journaled (polled via `page.request`, deliberately bypassing the route bar) removes a real unenforced assumption. Turning the throw-inside-route-handler into an asserted `premiseFailure` is a genuine robustness win too.

But the flake isn't closed, because the cut still fires on a second wall-clock guess.

**`FRAME_SETTLE_MS = 1_000` (`spec.ts:83`, slept at `:225`)** stands in for "the SSE frame arrived and `PendingSyncPosts` captured it" — a different race from the one `awaitJournaledReply` fixed. By the PR's own timing data the post-reply slack shrank from ~4–5s to 1s, so exposure to this race went *up*.

The failure mode is the bad kind — green while proving nothing:

> `route.fetch()` resolves → `awaitJournaledReply()` confirms the journal in <250ms → 1s elapses → cut fires, `repliesAtCut === 0` (correct) → `route.abort()` clears the client's pending-suppression state → the still-in-flight SSE frame arrives *after* suppression lifted and renders through the ordinary live path, not the release-on-error path. `expect(reply(...)).toBeVisible()` still passes.

The spec goes green having never exercised the `PendingSyncPosts` release logic it's named for — even if that logic is completely broken.

**Second, `historyProbe` capture races hydration.** It's only captured while `holdHistory` is `false` (`:121-129`), and `holdHistory` flips at `:242` right after `openChannel()` — which only awaits the composer placeholder, not the `chat/history` GET being issued. If hydration lags, the probe is never captured, `awaitJournaledReply()` burns its 45s deadline and returns "no history request was captured during hydration", and the test fails via the new `premiseFailure` assertion while the host answered perfectly normally.

Both are Codex's already-posted P2s; I confirmed the code at `:224-225` is byte-identical to what those comments quote, so neither was addressed by the second commit.

Suggested follow-up, keeping `awaitJournaledReply()` as-is: capture `historyProbe` unconditionally and gate only the *fulfill* on `holdHistory`; and replace the `FRAME_SETTLE_MS` sleep with a second observation — poll for "frame captured for this marker" via a test hook rather than sleeping. Until that second sleep is gone I don't think this spec can be called flake-closed, and it's the one that's been costing reruns across a lot of unrelated PRs.

The deterministic-cut property does hold — `route.abort("connectionaborted")` is unchanged and still a real interception, only *when* it fires moved.

PR #1913, opened 2026-08-28
test(e2e): cut the detached POST once the frame is HELD, not a second later (#1907)

### LINKED ISSUE #1907: chat-detached-post-failure: the cut still rides a 1s wall-clock guess, and historyProbe races hydration
Follow-up from the post-merge review of #1888 (comment: https://github.com/tinyhumansai/opencompany/pull/1888#issuecomment-5448907731).

`awaitJournaledReply()` was the right fix for the problem it diagnosed — replacing `CUT_AFTER_MS = 8_000` with a real observation of the reply being journaled (polled via `page.request`, deliberately bypassing the route bar) removed a genuine unenforced assumption. Converting the throw-inside-route-handler into an asserted `premiseFailure` was a real robustness win too. This issue is about what's left.

## 1. The cut still fires on a wall-clock guess, and the margin got smaller

`FRAME_SETTLE_MS = 1_000` (`frontend/test/e2e/chat-detached-post-failure.spec.ts:83`, slept at `:225`) stands in for "the SSE frame arrived and `PendingSyncPosts` captured it" — a *different* race from the one `awaitJournaledReply` closed. Per the PR's own timing numbers (~26s → ~22s per run) the post-reply slack shrank from roughly 4–5s to 1s, so exposure to this race increased.

The failure mode is the one that matters: **green while proving nothing.**

> `route.fetch()` resolves → `awaitJournaledReply()` confirms the journal in <250ms → 1s elapses → cut fires, `repliesAtCut === 0` (correct so far) → `route.abort()` clears the client's pending-suppression state → the still-in-flight SSE frame arrives *after* suppression lifted and renders through the ordinary live path, not the release-on-error path. `expect(reply(...)).toBeVisible()` still passes.

The spec goes green having never exercised the `PendingSyncPosts` release logic it is named for — even if that logic is entirely broken.

**Suggested fix:** replace the sleep with a second observation — poll for "frame captured for this marker" (a test hook on `PendingSyncPosts`'s in-memory state) rather than sleeping a fixed interval. Same technique that made `awaitJournaledReply` correct.

## 2. `historyProbe` capture races hydration

`historyProbe` is only captured while `holdHistory` is `false` (`:121-129`), and `holdHistory` flips at `:242` immediately after `openChannel()` returns. `openChannel` (`frontend/test/e2e/chat-helpers.ts:14-17`) only awaits the composer placeholder becoming visible — it does not wait for the desk-scoped `chat/history` GET to have been issued.

**Repro shape:** under CI load, the composer renders before the hydration effect fires. `holdHistory` becomes `true` first, `historyProbe` stays `null`, `awaitJournaledReply()` burns its full 45s deadline and returns `"no history request was captured during hydration"`, and the test fails via the new `premiseFailure` assertion — while the host answered perfectly normally.

**Suggested fix:** capture `historyProbe` unconditionally and gate only the *fulfill* on `holdHistory`.

## Why this is worth doing properly

This spec has flaked across a long run of unrelated PRs and has been costing a rerun cycle on most of them. The second-order cost is worse than the reruns: it trains everyone to treat a red E2E lane as noise, which is how a genuine failure gets waved through. A spec that both flakes *and* can silently stop testing its subject is worse than either problem alone — so it's worth closing the second guess rather than leaving it at "rarer".

Both items were also raised as Codex P2s on the PR; the code at `:224-225` is byte-identical to what those comments quote, so neither was addressed by the second commit.

The deterministic-cut property does hold — `route.abort("connectionaborted")` is unchanged and still a real `page.route` interception; only *when* it fires moved.

CodeGhost21:
Closes #1907.

## RCA

`awaitJournaledReply()` (#1888) closed one race — "has the turn finished?" — by observing the host's journal instead of guessing at 8s. It did not close the other one, and shortening the run made that one worse.

`FRAME_SETTLE_MS = 1_000` stood in for a **different** fact: *the SSE frame arrived and `PendingSyncPosts` captured it*. The reply being journaled says the host wrote it; it says nothing about whether this browser is holding it. The cut is only a test of the release path if it is.

The sequence when that 1s is not enough:

1. `route.fetch()` resolves; `awaitJournaledReply()` confirms the journal in <250ms.
2. 1s elapses. The frame has not arrived yet.
3. `route.abort()` fires → the POST throws → `onSendFailed` → `PendingSyncPosts.failed()` lifts suppression. There is nothing held, so there is nothing to release.
4. The frame lands *after* suppression is gone, `capture()` returns `false`, and it renders through the ordinary live path.
5. `repliesAtCut === 0` (true — it hadn't arrived), `reply(...)` is visible, count is 1. **Green.**

Every assertion passes and the release logic was never touched. It would pass identically if that logic were deleted.

### The failure demonstrated, not argued

Two variables, four runs, on `companies/e2e_harness`:

- **Mutation:** `PendingSyncPosts.failed()` changed to discard held frames instead of releasing them — i.e. the exact silent-loss regression this spec exists to catch, reintroduced.
- **Slow wire:** every `message` listener on the stream delayed 2.5s, standing in for the saturated CI runner (longer than `FRAME_SETTLE_MS`, and applied at the wire so relative ordering between listeners is untouched).

| spec | release logic | result |
|---|---|---|
| before | **broken** | **passed 2/2** ← the bug |
| after | **broken** | failed 2/2, on `expect(reply(...)).toBeVisible()` |
| after | correct | passed 2/2 — the slow wire *delays* the run (14.5s), it does not fail it |

The first row is the whole issue: the spec was green while the thing it is named for was entirely broken.

Neither experiment is in this diff.

## The fix

The sleep becomes an observation, the same technique that made `awaitJournaledReply` correct.

An `EventSource` recorder installed in `beforeEach` logs every `data:` payload the stream hands the page, and the cut waits until this turn's `You said: <marker>` is in that log before it fires.

It reads **the wire**, not the console's internals:

- No test hook on `PendingSyncPosts`. The issue suggested one; a test-only accessor on the single highest-risk rule in the detached design is a worse trade than reading the transport, which is a fact about the browser and cannot go stale when that rule is refactored. The console under test stays byte-identical to the shipped one — nothing under `src/` changes.
- The recorder's listener is added in the constructor, ahead of the transport's own `onmessage`, and `message` dispatch runs every listener synchronously — `handleEvent` → `injectAgentReply` → `capture` included. A poll from the test side is a later task by construction, so a frame the log has is a frame the console has already routed.
- Suppression is certainly up: `ChatView` calls `onSendStart` before `client.chat`, and the route handler is holding that very POST unresolved.

`repliesAtCut === 0` then becomes the other half of a complete reading rather than a weak "nothing rendered early": **the frame arrived, and nothing had drawn it, so the hold is the only place left for it.** The bubble that appears after the abort can only be the release.

`EventSource` is the right seam because it is the lane this console is on — `BrowserTransport.subscribe` only takes its `fetch` fallback for a credential `EventSource` cannot carry, and a same-origin console authenticates by `HttpOnly` cookie (the same fact `historyProbe` already relies on). If that ever changes, the recorder stays empty and the premise wait says so in as many words, rather than surfacing as a wait for something else.

The premise-failure message widens to cover both readings, and the journal wait is reported first (`??=`) — a host that never answered explains a frame that never arrived, so naming the second would bury the cause under the symptom.

## Item 2 (`historyProbe` races hydration) — already closed

The issue's second item was fixed by #1909, which merged ~2h after the issue was filed, by enforcement rather than by reordering: `historyProbe` capture is now synchronous, and `holdHistory` is not set until an `expect.poll` has confirmed the probe exists. The repro shape in the issue — `holdHistory` flipping true before the capture lands — can no longer occur, and a console that stops reading this desk's history now fails at that poll saying so. Nothing further needed; the "capture unconditionally" variant would be a no-op behind that gate.

## Verification

- `npm run typecheck:e2e` — clean.
- `npx playwright test chat-detached-post-failure` — **5/5 green at 10.9s per run**, down from ~22s. The run got faster because it no longer sleeps through a window it now observes.
- Mutation matrix above.

## Scope

One file, test-only: `frontend/test/e2e/chat-detached-post-failure.spec.ts`. No production change.

## Summary by CodeRabbit

* **Tests**
  * Improved detached chat end-to-end test reliability by replacing fixed delays with event-based waits.
  * Added validation that the expected live reply frame reaches the page before simulating a connection interruption.
  * Enhanced failure messages to clarify whether the reply was stored or delivered live.

### DIFF
--- a/frontend/test/e2e/chat-detached-post-failure.spec.ts
+++ b/frontend/test/e2e/chat-detached-post-failure.spec.ts
@@ -55,6 +55,22 @@ import { LIVE_BRAIN } from "./capabilities";
  * turn as undelivered, which is the same misdirection in a new place. The
  * capture is now synchronous, and the send waits on it.
  *
+ * Knowing the reply exists is still not knowing that THIS BROWSER is holding
+ * it, and the cut is only a test of the release path if it is. That gap was a
+ * one-second sleep — a second wall-clock guess, and a tighter one than the
+ * eight seconds it replaced, standing in for "the SSE frame arrived and
+ * `PendingSyncPosts` captured it" (issue #1907). Losing that race is the quiet
+ * failure rather than the loud one: `route.abort` lifts the suppression, the
+ * still-in-flight frame lands after it is gone and renders by the ordinary
+ * live path, and every assertion below still passes — green having never
+ * exercised the release logic this spec is named for, and green in exactly the
+ * same way if that logic were entirely broken.
+ *
+ * So that sleep is an observation too. The cut now waits until the frame has
+ * demonstrably reached the page (`awaitCapturedFrame`), which together with
+ * `repliesAtCut === 0` pins where the frame is rather than inferring it: it
+ * arrived, and nothing drew it, so the hold is the only place left for it.
+ *
  * Nothing else can draw this reply, and the spec makes that true rather than
  * assuming it. The `202` body never reached the page, so no turn id was learned
  * from the POST — but the shell also arms its turn poll from `listRuns` at
@@ -81,15 +97,20 @@ import { LIVE_BRAIN } from "./capabilities";
 const ENGINEERING = { id: "engineering", channel: "engineering-desk" };
 
 /**
- * How long the released `agent_reply` frame is given to traverse the stream and
- * be captured, once the host is KNOWN to have journaled the reply.
+ * The page-side key the stream recorder installed in `beforeEach` writes and
+ * `awaitCapturedFrame` reads: every `data:` payload this page has been handed
+ * on the company event stream, in arrival order.
  *
- * Issue #1885: this replaces a fixed 8s `CUT_AFTER_MS` that stood in for "the
- * turn has surely finished by now". The distinction is not the number — it is
- * that this waits on one already-sent push rather than on a turn of unknown
- * duration, so a slow host delays the spec instead of failing it.
+ * Test scaffolding, deliberately NOT a product surface. Nothing under `src/`
+ * knows this key exists, so the console being measured is byte-identical to the
+ * shipped one. The alternative was a debug accessor on `PendingSyncPosts`' held
+ * map, which would have put a test-only reader on the single highest-risk rule
+ * in the detached design (see that class's own doc) — a worse trade than
+ * reading the wire the frame arrives on, which is a fact about the browser
+ * rather than about the console's internals, and which therefore cannot go
+ * stale when that rule is refactored.
  */
-const FRAME_SETTLE_MS = 1_000;
+type FrameLogWindow = Window & { __ocLiveFrames?: string[] };
 
 test.beforeEach(async ({ page }) => {
   // Same tour-skip shim the rest of the suite uses — the first-run modal
@@ -98,6 +119,40 @@ test.beforeEach(async ({ page }) => {
     const real = Storage.prototype.getItem;
     Storage.prototype.getItem = function getItem(key: string) {
       return key.startsWith("oc-tour:") ? '{"skipped":true}' : real.call(this, key);
+    };
+  });
+
+  // Every frame the company event stream hands this page, recorded so the cut
+  // can wait on this turn's `agent_reply` ARRIVING rather than on a duration
+  // guessed to outlast its flight (issue #1907).
+  //
+  // `EventSource` is the right seam because it is the lane this console is on:
+  // `BrowserTransport.subscribe` only takes its `fetch` fallback for a
+  // credential an `EventSource` cannot carry, and a same-origin console like
+  // this one authenticates by `HttpOnly` cookie and sets no auth header at all
+  // (the same fact `historyProbe` below relies on). Should that ever change,
+  // the recorder stays empty and `awaitCapturedFrame` says so in as many
+  // words — which is the failure mode this spec keeps choosing: a premise that
+  // stopped holding reports itself, rather than surfacing as a wait for
+  // something else.
+  //
+  // The listener is added in the constructor, ahead of the transport's own
+  // `onmessage` assignment, and `message` dispatch runs every listener
+  // synchronously — `handleEvent` -> `injectAgentReply` -> `capture` included.
+  // A poll from the test side is a later task by construction, so a frame this
+  // log has is a frame the console has already routed. Nothing here inspects
+  // the console's state; the ordering is what makes that unnecessary.
+  await page.addInitScript(() => {
+    const frames: string[] = [];
+    (window as FrameLogWindow).__ocLiveFrames = frames;
+    const RealEventSource = window.EventSource;
+    window.EventSource = class extends RealEventSource {
+      constructor(url: string | URL, init?: EventSourceInit) {
+        super(url, init);
+        this.addEventListener("message", (event) => {
+          frames.push(String(event.data));
+        });
+      }
     };
   });
 });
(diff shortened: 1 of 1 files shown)


oxoxDev (approved):
This closes item 1 of #1907 properly, and it's a better fix than the one I suggested. Approving — with one line I'd like moved before merge (Codex's inline finding, which holds; I derived it independently before reading theirs).

## Item 1 — closed

`FRAME_SETTLE_MS` and its `setTimeout` are gone. The cut now blocks on `awaitCapturedFrame()`, which polls a `page.addInitScript` recorder that subclasses `EventSource` and pushes every `message` payload into `__ocLiveFrames`.

I checked the ordering claim rather than taking it: `BrowserTransport.subscribe` does `new EventSource(...)` then assigns `source.onmessage`, so the recorder's listener is registered strictly earlier — and more importantly the whole dispatch is one synchronous task through `onmessage → handleEvent → onAgentReply → injectAgentReply → capture()`, with no `await`/`queueMicrotask`/`setTimeout` on the path. A `page.evaluate` poll is a later task by construction, so anything in the log has already been routed.

Worth being precise about what that pins: the poll alone means *arrived and routed*, not *held*. The pin comes from pairing it with `repliesAtCut === 0` — if suppression weren't up, `capture` returns false, `renderAgentReply` runs, and the count is 1. Suppression is independently guaranteed because `onSendStart` fires before `client.chat(...)` and the handler is holding that POST. Arrived + nothing drawn ⇒ held. That reasoning is sound.

Reading the wire instead of adding a `PendingSyncPosts` debug hook was the right call, and for the right reason — `live-reply.ts` calls that rule the highest-risk part of the detached-chat change, and this keeps `src/` byte-identical.

Coverage is preserved: `chat/history` is fulfilled `[]` from the send onward, which also neuters the `listRuns` recovery path, and the `202` body was destroyed so no turn id was learned — so `:423` can only be satisfied by the release-on-error path. Off-path interleavings fail loudly rather than going green. The 5.9s CI pass (was ~22s) corroborates the recorder actually saw the frame; an empty recorder costs 45s.

**Item 2** is closed too, though by #1909 rather than here — `holdHistory` is now gated behind a poll on `historyProbe !== null`, so the interleaving I described is unreachable. The body says so and doesn't claim credit.

No new wall-clock sleep is introduced: the two new numbers are a 45s deadline and a 100ms poll cadence, the same shape `awaitJournaledReply` already had.

## One line to move first

`:396` waits up to 60s for `Couldn't send` **before** `:397` awaits the cut. That line is only reachable after `route.abort()`, and there's no client-side timeout on the chat POST, so the handler is the sole gate. Base worst case was `45s + 1s = 46s`, safely under 60. Head worst case is `45s + 45s = 90s`.

Failure scenario: saturated runner, journal confirms at T+20s, the frame lands at T+42s — inside its documented allowance, a run that should pass. The cut fires at T+62s, but `:396` already failed at T+60s with `waiting for getByText(/Couldn't send/)`, and `premiseFailure`'s carefully-worded message never gets evaluated. That's the misattribution this spec has spent three issues removing, with a wider window than before.

Moving `await cutReadyPromise` above `:396` fixes it. Note the reorder also surfaces that `test.setTimeout(150_000)` can't cover `90 + 60 + 60 + 5` — the declared budgets aren't all simultaneously spendable. That was already true in base (`46 + 60 + 60 + 5 = 171`), so it's a widening rather than a new defect, but worth a look while you're in there.

## Minor

`BrowserTransport.subscribe` uses `streamWithFetch` instead of `EventSource` whenever `authHeaders()` is non-empty — which is already true for any host where `carriesOwnSession` holds. If the harness ever moves to such a host, no `EventSource` is constructed, the recorder stays empty, and `awaitCapturedFrame` returns "the reply never reached this page over the event stream" while delivery worked fine over the fetch lane. The composite at `:411-412` would then call it "a real delivery failure, not a timing artefact". Cheap guard: set `__ocStreamLane = "eventsource"` in the init script and return a distinct message when no `EventSource` was ever constructed.

Also non-blocking: the needle matches the raw `data:` payload rather than `type === "agent_reply"`. I couldn't construct a false positive — only `agent_reply.text` carries reply prose, and the echo brain runs no tools — but a `JSON.parse(frame).type` guard in the poll would cost nothing.

Issue #1890, opened 2026-08-27
Epic: a thread is a rendering fold, not a context boundary


sanil-23:
## Problem

A chat thread is a rendering fold and nothing more. `OperatorMessage.parent` is
"a parent id, not a thread object" (`src/ports/types.rs:610`) and the console
folds a transcript by it, so a threaded reply renders in the thread panel
instead of the channel timeline. Nothing downstream of the renderer knows the
thread exists — **the model answering in a thread is handed the whole channel.**

Four leaks, all live today with user-initiated threads:

| # | Where | What happens |
|---|---|---|
| 1 | `src/harness/built_in/mod.rs:949` | `let switched = bound.as_deref() != Some(incoming)` — `incoming` is the **chat id**, i.e. the channel. Thread A → thread B in one channel is not a switch, so `clear_history()` and the re-seed never run and thread A's turns are still loaded in the agent's in-memory `history`. |
| 2 | `src/harness/built_in/chat_seed.rs:259` | The seed filters on `chat_history::owns(desk_id, desk_name, event)`, which matches on chat id only — no parent term. When a re-seed *does* run, it projects the channel's last `CHAT_SEED_WINDOW = 30` messages with every live thread interleaved by wall-clock. |
| 3 | `src/ports/tasks.rs:960` | `TaskRecord` records `origin_chat_id` — the channel — and no thread root. A card raised inside a thread settles its `DeskTaskCompleted` marker flat in the channel, so the thread that asked for the work never shows it finishing. |
| 4 | `src/server/chat_history.rs:125` | `owns` admits `DeskTaskCompleted`, but `build_chat_seed` maps it to `None` ("carries no conversational body — do not seed it as a turn"). Correct as written, but it means a channel agent cannot see that work raised in its own channel settled. The one durable fact answering "did that ship?" is on screen and absent from context. |

The seed's shape is what makes leak 2 undetectable to the model: it emits bare
`(role, content)` pairs — no author, no seq, no parent — so a sibling thread's
turn is indistinguishable from the current thread's own.

Concretely, in a channel with two live threads:

```
seq 41  OperatorMessage { chat:"growth", parent:None,  "draft the launch email" }   <- root A
seq 42  AgentReply      { chat_id:"growth", parent:41, "Here's a draft..." }
seq 43  OperatorMessage { chat:"growth", parent:None,  "what's our Q3 CAC?" }       <- root B
[...]

Issue #1865, opened 2026-08-26
workflows/tasks: the passive layer — honest run verdicts, failure notifications, and a board chip separating bounced from fresh


sanil-23:
## Problem

For board-created tasks and workflow runs, terminal states are dishonest or wordless — the passive surfaces (board, run history, notifications) misreport or under-report what happened. This is the fallback layer for everything #1860 does **not** park: non-answerable failures, unanswered/expired blockers, and quality degradations inside runs that settled green.

The silent-failure kinds, board/workflow only:

| Kind | Today | Class |
|---|---|---|
| Node error under `on_error: continue\|route` | run settles **green** — `WorkflowRunVerdict::of` (`src/ports/workflow_verdict.rs:234`) never reads `nodes[].status`; red only on the canvas overlay | false success |
| Agent node truncated at iteration cap | engine settles the node OK with partial output downstream (`tinyflows agent.rs:385`) while its attempt row settles Failed (`src/workflows/caps/mod.rs:1580`) | two surfaces disagree |
| Failed dispatch → card returns to `todo` | reason buried in the note; **`todo` is both the failure state and the fresh state** — a bounced card and an untouched card look identical | unannounced failure |
| Scheduled run fails overnight | journaled + host-stdout warn only; discovered by opening Run History | unannounced failure |
| Panic on the orchestrator `run_workflow` tool path (`orchestrator.rs:3647`, bypasses `WorkflowSpawn`'s `catch_unwind`) | run reads `running: true` until the next boot sweep — which does not run on rebuild | zombie |
| Stranded approvals (gates parked, queue unwired, `runner.rs:1611-1625`) | "awaiting approval" forever with nothing to decide; the sync run response hardcodes `stranded_approvals: 0` (`ops/workflows.rs:1608`) | zombie |
| Any failure at all | `NotificationStore` (`src/ports/notifications.rs`) defines `SubjectKind::{Task, Run, Workflow}` and has exactly **one** producer in the tree (`notify_mentions`) — nothing pushes | no delivery |

## Fix shape

[...]

Issue #1682, opened 2026-08-24
chat: attachment (paperclip) button is a permanently-disabled placeholder — wire chat attachments end-to-end


oxoxDev:
## Problem
The chat composer shows a paperclip "Attach a file" button, but it's greyed out and does nothing — there is no way to attach anything to a channel or DM message.

## Root cause
`frontend/src/views/chat/MessageComposer.tsx:191-200` renders the paperclip with a hardcoded `disabled` and no `onClick`. Chat attachments are **not wired at all**: `onSend` carries no file, the chat route is text-only, and the only multipart route (`/workspace/upload`, `api/workspace.ts:468`) is Workspace-scoped. History note: the button was **born disabled** in the #361 console rebuild (commit 8c1bc196a, 2026-08-04) — the pre-rebuild composer had no attach affordance either, so this never functioned; the rebuild added a disabled icon that raised the expectation.

## Fix
Wire attachments end-to-end: composer file picker + upload state → a chat multipart route (new backend) → message render with the attachment. (If the decision is instead to not support chat attachments yet, remove/hide the misleading control — but the ask is to make attaching work.)

## Scope
L (frontend picker + upload state + new backend multipart chat route + message rendering).

PR #1942, opened 2026-08-31
feat(chat): a turn is told what work this conversation raised has finished (#1890 C)

sanil-23:
## Summary

Sub-issue **C** of #1890 — *settle markers reach the model*. Stacked on #1939 (**B**),
which it depends on; review that one first.

`chat_history::owns` files a `finished → In review` marker back into the
conversation a card was raised from, so the **operator** can see the work settle.
The **model** cannot. The chat seed drops the marker for want of a conversational
body — correct, since a settle is not a turn — and the epic names the
consequence: *the one durable fact answering "did that ship?" is on screen and
absent from context*.

So a channel agent, asked whether the thing it was asked for shipped, has nothing
to answer from. `inject_handed_task_awareness` briefs on **open** work, filtered
`column != "done"` and matched by **assignee** — it answers "what are you working
on?", which is a different question.

This adds the second briefing, off the same card read.

## The two decisions

**It is a briefing, not a seed line.** `seed_resume_from_messages` recognises
`user`, `agent` and `assistant`, and **falls back to the user role for anything
else** — losing context being worse than mislabelling it, on its own terms. A
marker emitted into the seed under a `system` role would therefore reach the
model as though the operator had typed "finished → In review", which is worse
than dropping it. The seed is also rebuilt only on a *switch* (deliberately — the
projection is costly on the filesystem backend), so a settle landing
mid-conversation would never arrive at all.

**It reports the board, not the marker.** The console's marker is rendered from
the journal and frozen at settle time: a card dragged to Done afterwards still
reads `finished → In review` on screen. This reads the card's **current** column,
so it answers "did that ship?" with where the work actually is. The two can
disagree, deliberately — the screen is a record of what happened, this is a
statement of what is true now — and the annotation says so in the text the model
sees. It is also free: `inject_handed_task_awareness` already calls
`tasks().list()`, so this is a second filter over a list in hand rather than a
journal scan on every turn.

## Where it lands

| | |
|---|---|
| `SETTLED_WORK_ANNOTATION` | beside `OPEN_WORK_ANNOTATION` and `BUILDER_ANNOTATION`, on identical terms: in-memory only, never journaled |
| `inject_handed_task_awareness` | a second pass over the same cards, matched on the **conversation that raised the work** rather than on who it was handed to |
| `has_settled` | has this card *stopped* — not *succeeded* |
| `settled_briefing_line` | `- <title> — finished → <label>`, label from `ledger::board` |
| `operator_words` | **the fourth marker in the cut list** — see below |

Matched on both halves of the origin, which is what B was for: the channel through
`same_conversation` (folding General's four spellings) and the thread verbatim. No
desk-*name* term, unlike `chat_history::owns` — both sides here are the raw chat
id the operator addressed, stamped onto the card from that same field, never a
resolved desk.

## The two traps

**`operator_words` must cut on this marker.** That function exists because of a
live bug: a desk-addressed "thanks!" scored as substantial work — the appended
briefing is long, and `looks_like_work` scores length — and opened a card, which
lengthened the briefing, which made the next card likelier. It has now happened
three times (#176, #845, #1682). This is a fourth block, and its loop is the
nastiest of them: #176's briefing grew as cards were *opened*; this one grows as
cards **finish**, so every completed card would make the next message likelier to
open one, which would in time finish and lengthen the briefing again. Pinned by a
test built from the shared constant, so rewording the briefing fails the test
rather than silently restoring the bug.

**`todo` is both the failure landing and the fresh state.** A card that bounced
there off a failed run has settled; one nobody has touched has not, and they share
a column. `has_settled` asks `TaskRecord::bounced` — the field #1865 added to close
exactly this distinction on the board, for a human reader — rather than inventing a
second rule for a model reader. `planning` and `in_progress` are never settled;
calling them finished would be the "concluded the work had finished when it had in
fact parked" misreading #377 exists to remove.

**Truncation is declared.** Capped at 5, most recent first, with `(and N more, not
listed)` past the cap. A model handed 5 of 9 with no marker answers "that is
everything" confidently and wrongly — the rule the epic sets for its thread index,
applied here.

## API Or Behavior Changes

- **No wire change, no persisted change.** The briefing is folded into the
  in-memory `OperatorMessage` handed to the brain and never journaled — the same
  contract `OPEN_WORK_ANNOTATION` has had since #176.
- A desk-addressed turn in a conversation that has raised work now carries a few
  extra lines of context. A conversation that has raised none is byte-identical.
- Brain-agnostic: both brains read `req.events`.

## Tests

- `a_settled_card_briefs_the_thread_that_raised_it_and_no_other` — the whole of it, end to end through the injector, including that a **sibling thread's** settled work does not leak (a briefing that crossed threads would undo sub-issue A one message later). Confirmed to fail both with the briefing removed **and** with only the `origin_parent == thread` term dropped.
- `a_truncated_settled_briefing_declares_what_it_left_out` — the cap, the declaration, and most-recent-first ordering. Fails without the fix.
- `a_card_has_settled_only_once_its_run_stopped` — every column, with `todo` split by `bounced`.
- `a_settled_line_names_the_landing_and_a_bounce_names_its_reason`.
- `work_still_running_is_not_briefed_as_finished` and `a_settled_card_says_nothing_in_another_channel` — negative cases; these pass on `main` too (nothing is briefed there at all), so they guard against a too-lax rule rather than prove the fix.
- `the_cycles_settled_work_briefing_is_not_the_operators_request` and `operator_words_cuts_at_the_first_of_every_briefing` — the "thanks!" trap.

Run locally:

- [x] `cargo test --locked`

## Documentation

The constant is the documentation, on the pattern its two siblings set: why a
briefing rather than a seed line, why the board rather than the marker, and the
`operator_words` obligation. No wire surface changed, so `docs/spec/runtime/`
needs nothing.


sanil-23:
**Stacked on #1939 (sub-issue B) — review that one first.** This branch contains B's commits, so the diff against `main` shows both. The C-only change is `eeb2e6afc` (`src/runtime/cycle.rs`, `src/runtime/delegation.rs`); everything else is B and the `main` fix explained over there.

C depends on B and cannot be split from it: the briefing matches a card to a conversation on **both** halves of the origin, and `TaskRecord::origin_parent` is what B adds. Without it a settled card in one thread would brief every thread in its channel, which is sub-issue A undone one message later — there is a test for exactly that.

Once #1939 merges I will rebase this onto `main` so the diff narrows to the C commit alone.

Local gates on the merged tree, all green: `cargo test --locked`, `cargo test --locked --features openhuman --tests` (6323 passed), clippy on default / `openhuman` / `acp,runner,tinymemory`, `cargo fmt --all --check`.

One note for whoever watches CI here: `server::ops::workflows::tests::running::cancelling_an_unknown_or_settled_run_is_not_found` failed once in a full-suite run on this branch (200 instead of 404) and then passed 3/3 in isolation, on a branch touching only `cycle.rs` and `delegation.rs`. It looks like a pre-existing race under parallel load rather than anything here, and it is worth its own fix.

sanil-23:

PR #1945, opened 2026-08-31
feat(chat): an answer threads under the message that opened it (#1890 D)

sanil-23:
## Summary

Sub-issue **D** of #1890 — *replies land in a thread*. **Stacked on #1939 (B)**,
which it depends on; review that one first. The D-only commit is `e0bd81f56`.

A thread is something an operator opens by hand, and nothing else. Every answer
the company gives is journaled unparented, so a channel is a flat run of
questions and answers and the only structure in it is what a person built. This
threads an exchange under the message that opened it.

The epic is explicit that this cannot be split, and the reasons are concrete:
part 1 without 2 empties the channel on screen, 1 without 3 empties it for the
model, and 2 without 3 is the trap — an inline-rendered reply still carries a
`parentId`, so the answer is back on screen while the model still never sees it.
Two surfaces, one definition of "in this channel".

## The three parts

**1 — Journal, uniform.** `reply_thread(asked_in, message_seq)` returns
`Some(asked_in.unwrap_or(message_seq))`. Already in a thread → the same root, so
N messages in a thread stay one *topic* rather than becoming N threads. Not in
one → the message itself, so the exchange becomes a thread.

Never `None`, and that is the point. The tempting version decides here — "thread
it only if another question arrived while I was working" — which makes `parent` a
function of race timing. `parent` is permanent: two operators doing the identical
thing would get permanently different transcripts on microseconds, and the
console renders a reply as it streams, before the backend could know.

**2 — Timeline, adaptive.** `buildTimeline` dropped every parented line, so part
1 alone deletes every agent reply from the channel. A thread's **first reply now
renders inline** when nothing came between question and answer; the pair folds to
the summary row only when something did — the case the fold was always for, two
conversations racing in one channel.

"In between" is anything outside the root's own thread, which is wider than
"another root" deliberately: a sibling thread's reply interleaves the two
conversations on screen just as visibly as a new question does, and the rule is
about what a reader sees rather than the shape of the id graph.

**3 — Seed, matching.** `in_thread(None)` becomes roots **plus each root's first
reply**. The narrowing runs after the walk reverses to chronological order,
because it cannot be decided per event: a backward walk meets a root's *newest*
reply first and its oldest last. Deliberately not "one level flattened" — that is
the pre-#1890-A leak, siblings and all, arriving through the channel-level door
instead of the thread one.

## Three bugs my own change had, each caught by an existing test

Worth listing, because two of them are invisible when they break.

1. **A grandchild rendered.** Promoting any first reply inline promoted a
   reply-to-a-reply too, giving the console a second fold level. Not cosmetic:
   `cycle_conversation` (`src/runtime/cycle.rs`) parents an approval continuation
   to the thread *root* rather than to the message that raised it **precisely
   because** a grandchild is unrenderable, so #435's routing choice would have
   quietly stopped being necessary. Only a root's reply is promoted now.
2. **The chip leaked the second level.** With the reply now a rendered row, its
   own bucket hung off it — putting level two back on screen through the summary
   instead of through a row. Only a root carries a chip.
3. **The narrowing hit thread seeds too.** A turn answering a follow-up inside a
   thread was getting the question and one reply out of five — a worse seed than
   the leak #1890 A closed. The narrowing is the channel's rule alone.

The third is the one I would have shipped if `a_thread_sees_only_its_own_exchange`
had not existed.

## API Or Behavior Changes

- **No wire change and no schema change.** `AgentReply.parent` has existed since
  #364; this changes which value it carries on the chat path.
- **Existing transcripts are untouched.** Replies already journaled unparented
  stay unparented and keep rendering exactly as they do today — the new
  presentation applies to what is threaded, and nothing backfills.
- **A channel reads the same for a simple exchange** — question, answer inline —
  and folds only where two conversations overlap, which is where a flat render
  was already unreadable.
- **Not widened to the other reply paths.** `journal_dispatch_replies`
  (`parent: None`), `publish_continuation` (the approval's remembered parent) and
  `DeskChannel` are unchanged. The dispatch relay bubble is a real candidate now
  that B gives a card its thread, and is worth its own change rather than
  widening this one.

## Tests

- `an_answer_threads_under_the_message_that_opened_the_exchange` — part 1's rule, both arms. Confirmed to fail when the rule is reverted. **Part 1 had no coverage at all before this**, which is why changing it broke nothing and why I went looking rather than trusting the green run.
- `lays a first reply out inline when nothing came between question and answer` / `folds the pair instead once another conversation interleaved` / `counts a sibling thread's reply as interleaving too` — part 2's rule and both sides of it.
- `renders a grandchild nowhere: the fold is exactly one level deep` — pre-existing, and it caught bugs 1 and 2. Updated to the new shape; what it pins is unchanged.
- `the_channel_sees_roots_and_their_first_replies` — part 3, replacing `the_channel_sees_only_unparented_lines`. The follow-up typed inside the thread still stays out, which is the line between "the channel can see its own answers" and the pre-A leak.
- `a_thread_still_sees_its_whole_exchange` — the narrowing is the channel's alone. Caught bug 3.
- The facepile tests now interleave deliberately, so they keep testing the pile rather than accidentally testing the new inline rule.

Run locally, all green:

- [x] `cargo test --locked`
- [x] `cargo test --locked --features openhuman --tests` (6316 passed)
- [x] `pnpm typecheck`, `pnpm typecheck:unit`, `pnpm typecheck:e2e`
- [x] `pnpm vitest run` (3943 passed)
- [x] `./scripts/ci/assert-md-line-cap.sh`, `./scripts/ci/assert-design-tokens.sh`

## Documentation

The rule lives on `reply_thread` and `inlineFirstReplies`, each carrying why the
decision is where it is — the journal for permanence, the renderer for
presentation. No wire surface changed, so `docs/spec/runtime/` needs nothing.

## Summary by CodeRabbit

* **New Features**
  * Dispatch completion markers now appear in the thread where the task was raised, with correct handling for channel-level tasks.
  * The chat timeline displays simple first replies inline, while interleaved replies remain grouped in threads.
  * Replies whose original message is unavailable now remain visible in the channel.
* **Bug Fixes**
  * Completion notifications are no longer hidden unless the relevant thread is actually open.
  * Mentions in inline replies clear correctly when viewing the channel.
* **Compatibility**
  * Existing chat history and saved task data continue to load without migration.


sanil-23:
**Two gaps the E2E caught, both fixed in 553b56aad — and one of them reverses a documented rule, so it wants a reviewer's eye.**

I raised this without running Playwright locally, having flagged the E2E surface as the most at-risk part of the change. The unit tests all passed and told me nothing. I can run it locally now, and did: **350 passed**.

### 1. An answer whose parent this console never drew disappeared

`buildTimeline` dropped a reply pointing at an id the transcript does not hold rather than promoting it, reasoned as preventing the "the console lost my thread" symptom `reconcileIds` exists to fix. That was safe while only hand-opened threads carried a `parentId`. Part 1 gives **every** answer one, so the same rule silently deleted answers — in two ordinary cases, both caught by E2E:

- `chat-live-events`: a reply to a message **another client** sent. This console deliberately does not draw an operator line it did not send, so the parent is absent and the answer vanished.
- `chat-detached-post-failure`: a reply arriving before `reconcileIds` has swapped a locally-sent message's id for the host's — which a killed POST leaves pending permanently.

An orphan bucket now renders flat, all of it, since there is no summary row to put a remainder on. **The trade:** a reply whose root fell outside the history window reads without its question, versus an answer that is simply gone. I took the former — a lost answer is the failure this sub-issue exists to prevent — but the old rule had its own rationale and a reviewer may want to argue it. The depth rule is untouched: a grandchild whose own parent IS present still renders nowhere.

### 2. A visible summons stayed unread

`chat-mention-badge` states its premise in the test: *"replies are folded out of the main timeline — they render only inside the thread panel."* Part 2 breaks it. An inline reply is on screen the moment the channel is, so the badge could never clear and no thread panel existed to open.

`replyParents` now asks `buildTimeline`'s own rule through an exported `inlineReplyIds` rather than re-deriving it — two surfaces, one definition, the discipline `owns` enforces host-side. Both halves are pinned: a folded reply still defers (that fixture now interleaves deliberately, or it would no longer be testing deferral), and an inline one clears.

sanil-23:

PR #1952, opened 2026-08-31
perf(store): read a transcript page from the end of the journal (#1890 G)

sanil-23:
## Summary

Sub-issue **G** of #1890 — *bound the seed's first-page cost*. **Stacked on #1939
(B)**; the G-only commit is `71105dd20`.

`EventLog::read_before` is the primitive every transcript reader opens with: the
chat seed, the `chat/history` route, the GraphQL `Chat.history` projection. The
filesystem backend answered it by streaming from the **head** of the events JSONL
and `serde_json`-parsing every line to keep the last few in a sliding window — so
one page cost O(total company events) however few pages were asked for, and it
parsed everything it then discarded.

That cost predates this epic and was paid on each channel switch. Scoping the
chat seed to a thread (#1890 A) turned a rebind from per-channel into
per-**thread**, so on a company with any history it became the dominant cost of
answering a message. That is why it is in this epic rather than filed as an
ordinary performance nit.

## The change

`read_lines_backwards` seeks to the end and reads 64 KiB chunks, yielding whole
lines newest-first; `read_before` parses only the lines it actually returns.

A page **behind a cursor** still walks back from the tail to reach it. That is
the shape a transcript's own pagination already has — each page one step further
back — and no page re-reads the file to find its start, which is the property
that was missing.

## Measured

`cargo bench --bench journal_read_before`, macOS 15 / APFS / local SSD, one
512-record page:

| journal | `forward` (before) | `backward` (after) |
|---|---|---|
| 1,000 events | 759.9µs | 397.5µs |
| 10,000 events | 7.25ms | 394.0µs |
| 100,000 events | **72.80ms** | **399.8µs** |

`forward` is exactly linear — ten times the journal, ten times the time — and
`backward` is flat to within noise. **182× at 100k events**, and the ratio keeps
growing, because only one of the two curves has a slope.

The ~397µs floor is not overhead: it is the cost of reading and parsing the 512
records the caller asked for, which is the irreducible work. The whole of the old
implementation's excess was spent on records it discarded.

The bench is committed beside `journal_append` and **opt-in for the same reason**
— it measures the host's storage, so a shared runner's numbers would mean nothing
to gate on — and carries the same honesty section: the shape is the finding, a
local SSD does not predict a network volume, and a warm page cache flatters the
old implementation more than the new one because it is the one doing the extra
reading.

## API Or Behavior Changes

**None.** Same trait, same contract, same results — `read_before` returns the
same page, newest-first, for every input. This is the implementation behind it.

## Tests

The helper is tested directly, because a backwards reader fails by *silently
truncating*, and a transcript missing one line in the middle still renders:

- **a line longer than a whole chunk, and lines straddling chunk boundaries** — 801 lines across many reads, each whole and exactly once. This is the case worth the most here.
- a file with **no trailing newline** — the last line is a line, not a fragment to drop.
- blank lines skipped, exactly as the forward walk skipped them; empty file; absent file.
- **the walk stops as soon as the caller is done** — 3 lines read out of 5,000, which is the entire point of the change.

And the **shared conformance suite** gains the `usize::MAX` sentinel, bounded and
cursored. That is the input most likely to be wrong while looking right: an
implementation that reserves against the limit allocates 2^64 slots, and one
reading from the end must not treat it as a stopping count. Every caller of the
unbounded form is a full-history reader, so a page silently short there is a
reader silently missing history. All three backends are now held to it — fs and
sqlite verified locally, and mongodb already has an explicit `Unlimited` arm for
the sentinel.

Run locally, all green:

- [x] `cargo test --locked`
- [x] `cargo test --locked --features sqlite --lib store::sqlite`
- [x] `cargo test --locked --features openhuman --tests` (6321 passed)

## Documentation

`read_lines_backwards` carries why the direction changed and what it costs; the
bench carries the measurement and how to read it honestly. No wire or spec
surface changed.

Unblocks the `find_thread` half of the epic's **F**, which is unusable until the
read cost is bounded — `query_company` is the cautionary case, doing
`read_from(EventSeq::new(0), usize::MAX)` on every call.

## Summary by CodeRabbit

* **New Features**
  * Settled card markers now appear in the specific conversation thread where the card was raised.
  * Channel-level cards continue to display markers at the channel level.
  * Thread placement and visibility remain consistent after reloading chat history.

* **Performance**
  * Reading recent journal entries is faster, especially for large event histories.

* **Documentation**
  * Added documentation explaining settled-card markers, thread behavior, and origin tracking.


sanil-23:

PR #1955, opened 2026-08-31
fix(harness): scope an ACP session to its conversation, and bind an unstreamed turn (#1890 H, I)

sanil-23:
## Summary

Sub-issues **H** and **I** of #1890 — *the two places isolation does not reach
yet*, one PR as the epic's own split has them. **Stacked on #1939 (B).**

Both are pre-existing and both were made visible rather than caused by A. And
each turned out to have a **comment asserting the property the code did not
have**, which is the thread running through this PR.

## H — an ACP teammate had one conversation for the whole company

`AcpRunTurn::session_key` was `{company}::{agent}`, and its own doc read:

> Per (company, agent) so two desks do not share a conversation

The key holds no channel, so that is exactly what they did. Every desk, every DM
and every thread an ACP-backed teammate answered in was **one durable
conversation inside another program**.

### Scoped per channel, not per thread — and the port is why

The rest of this epic makes the *thread* the boundary, and for the built-in
harness it is: that binding is in-memory, so a new one costs nothing and a stale
one dies with the process that owns it.

An ACP session is neither. It is durable state in another program, reached
through a port with **no lifecycle at all** — `AcpAgent` has `prompt` and
`cancel`, and there is no create, close or reap anywhere on it. Keying per thread
would mint an unbounded number of external sessions on a busy channel with
nothing able to collect them, and giving them a lifetime means designing one
across a port whose implementation lives in the desktop shell rather than in this
crate.

A channel is bounded by the roster and changes rarely, so this is the isolation
available without inventing that lifecycle. Thread-level scoping stays open on H
with that recorded as the reason.

### The turn slot is deliberately **not** the session key

`session_key` was doing two jobs: naming the conversation *and* serialising the
agent's turns. Widening both would let one teammate's desks prompt an external
process concurrently — a change to how that process is driven, made as a side
effect of a change about conversation scope, and one this issue has no evidence
for. Split into `session_key` and `lock_key`, with a test pinning that two
conversations share a slot while holding different sessions.

That split also made `steered_with_grace`'s cancel reasoning stale — it argued a
queued turn's cancel "names the session, which belongs to whichever turn is
running". That is now only reachable between two turns in the **same** channel.
The guard stays and is still right; the comment now says what is true, because
leaving it would have repeated the defect fixed one function above.

**General's four spellings stay one session**, folded through the same
`is_general_chat` rule every other reader uses — otherwise the operator's main
line would forget itself depending on which id the caller addressed.

## I — chat identity was derived from live streaming

`run_with_steer` read `turn_chat_id` off the **live stream route**, so a turn that
has a conversation but publishes no frames could not be expressed. The approval
re-dispatch is exactly that turn: `run_steered_background` runs `LiveStream::Off`,
so the binding branch was skipped, the else branch invalidated `bound_chat`, and
the re-issued call ran against **whatever history the agent happened to be
holding** — then published its answer into the origin thread regardless.

The grant has known its conversation since #435 (`origin_thread` +
`origin_parent`, the same pair C uses for the publish path). It simply had no way
to pass it.

So: `run_steered_background` takes a `ChatTarget`, and identity is carried in its
own right. **The precedent is A's** — `run_with_steer` already took `thread_root`
separately, for the same reason ("the binding must not depend on an optional
dependency"); this finishes the job for the channel half.

A dispatched card passes `ChatTarget::default()` and keeps exactly its old
behaviour, including that it does **not** clear history — one background task can
span several turns that depend on what accumulated between them.

### The fallback I broke and put back

Reading identity purely from the caller dropped a normalisation the stream route
performs: an unaddressed message is folded onto `DEFAULT_DESK` there, so
`chat.chat_id` would hand back `None` and **unbind a turn that today binds to
General** — the behaviour `an_unaddressed_message_still_binds_to_its_thread`
exists to pin, and which #1896's review already established is correct rather
than a gap. The stream therefore still wins when present, making this a strict
extension: nothing that streams changes at all. Caught by that test, not by me.

**Residue, stated rather than left to be discovered:** an approval raised in an
*unaddressed* message still has `origin_thread: None`, so its re-issued call
binds to nothing as before. Closing that means teaching `ChatTarget` to tell
"unaddressed" from "no conversation at all" — a wider change than this one.

## API Or Behavior Changes

- **`RunTurn::run_steered_background` gains a `ChatTarget`.** Internal trait; the
  17 other implementations are test doubles that ignore it.
- **An ACP teammate starts fresh in a channel it has not spoken in.** That is the
  fix, and it is a behaviour change for anyone relying on cross-desk carry-over —
  which was the leak. An existing session is still found by the turns that were
  already sharing it: General and the unaddressed paths keep the old key exactly.
- **No wire, schema or console change.**

## Tests

- `a_session_key_separates_conversations` — two desks, and a desk vs a DM, no longer share a session. The property the old comment claimed.
- `every_spelling_of_the_general_desk_is_one_session` — four spellings, one session, or the main line forgets itself depending on which id was addressed.
- `the_turn_slot_stays_per_teammate` — the slot is not the session key, and two conversations share a slot while holding different sessions.
- `identity_and_streaming_are_separate_questions` — an unstreamed turn with a conversation is distinguishable from one without, which is the whole of I.
- `an_unaddressed_message_still_binds_to_its_thread` — pre-existing, and it caught the dropped fallback above.
- The `iteration_cap_turn_test` isolation cases now name their conversation explicitly rather than relying on the stream to imply it, which is what the change is about.

Run locally, all green:

- [x] `cargo clippy` on default, `openhuman`, and `acp,runner,tinymemory` — each `-D warnings`
- [x] `cargo test --locked`
- [x] `cargo test --locked --features acp,runner,tinymemory --tests`

## Documentation

The reasoning lives on `session_key`, `lock_key` and `run_steered_background` —
including why H stops at the channel and what closing it at the thread would
require. No spec surface changed.

With this, every item on #1890 has landed or is in flight except **E** and **F**.

## Summary by CodeRabbit

- **New Features**
  - Task completion markers now appear in the channel or thread where the task originated.
  - Thread context is preserved across task cards, approvals, scheduled runs, and conversation history.
  - Conversation-specific sessions prevent context from crossing between chats.
  - Workflow postconditions now validate structured results before downstream execution.

- **Bug Fixes**
  - Prevented duplicate completion markers across live updates and history.
  - Completion toasts appear only when the relevant marker is not visible.
  - Budget, approval, and answerable dispatch pauses now show clear system notices and paused task status.


sanil-23:
**Latent break fixed in f9fbde38e — found by building all seven #1890 branches together, not by CI.**

`main` gained three `run_steered_background` test doubles (in `workflows/caps/mod.rs` and `workflows/runner.rs`) after this branch widened that trait method. Git merges the two without complaint — this branch added the parameter to the doubles that existed, main added new ones elsewhere — and the result does not compile.

That is a semantic merge conflict, and the same class that took `main` down earlier today with `created_at_millis`: both sides green on their own merge ref, only the combination red — which is the one state no PR's CI ever builds. This branch would have gone red on its next rebuild, or broken `main` on merge.

Current `main` is merged in here now, so this is verified against the tree it will actually land on: `cargo test --locked --features openhuman --tests` (6446 passed), clippy on `openhuman`, and `cargo fmt --all --check`, all green.

sanil-23:

Issue #176, opened 2026-07-29
Brain-agnostic delegation seam: hosted/Medulla-path hand-back + recursive desk depth + handed-task awareness


oxoxDev:
## Summary

Delegation exists on only **one** of OpenCompany's two cognition brains. `HarnessBrain` (vendored OpenHuman engine, compiled in via `openhuman` feature) has the full orchestrator delegation stack — `delegate_to_desk`, `spawn_task`, the desk-lead relay, the single "CEO voice" hand-back (#175). The `HostedMedullaBrain` (cognition remote on the Medulla service; device is a thin `tool_call` client) has **none of it**: its tool catalog is built only from `manifest.tools.allow`, and `CycleHostImpl::call_tool` just does `rt.tools.invoke(...)` with no delegation arm. smoke1 and every hosted-cognition tenant run the Medulla brain, so on smoke1 the orchestrator **cannot delegate at all** and #175's hand-back is a no-op.

This epic makes delegation **brain-agnostic** and adds two capabilities on top:

- **(A) Hosted/Medulla-path delegation + hand-back** — lift the delegation runtime out of `HarnessBrain` into a shared, brain-agnostic seam so the Medulla path gets `delegate_to_desk` / `spawn_task` and the one-voice relay, mirroring OpenHuman's blocking sub-agent model (child reply returned as a **tool result**, continuation relays naturally).
- **(B) Recursive desk-member delegation depth** — today the tree is one hop deep (orchestrator → desk lead). Give desk members their own delegation tools so delegation can recurse into a bounded tree, with the same guards OpenHuman uses (allowlist + max-depth + circuit breakers).
- **(C) Handed-task awareness (NEW).** When a desk/agent is handed a task, that hand-off must land in the agent's working memory/context so that if the operator later asks that agent **directly**, it knows it was handed the task. Today delegation state is orchestrator-local; the target agent has no durable record of the assignment.

## Motivation

- smoke1 (hosted Medulla) has no orchestrator delegation — the entire desk hierarchy is inert on the production hosted path. Fixing it on only the harness path leaves every real tenant unserved.
- OpenCompany "runs a company," not a chat app — a shallow one-hop org can't model real desks-within-desks. Recursion is the structural unlock.
- An operator asking the engineering desk "what are you working on?" must get the truth. If a handed task is invisible to the desk when queried directly, the org is lying to the operator.

### Shared delegation seam (the crux for A + B)

[...]

PR #1961, opened 2026-08-31
feat(chat): a thread is told what else its channel is about (#1890 E)

sanil-23:
## Summary

Sub-issue **E** of #1890 — *cold-start briefings*. **Stacked on #1942 (C)**,
which it must be: both inject at the same seam, and the repo's rule is that
issues touching the same files run sequentially. The E-only commit is the tip.

A thread scoped to itself (#1890 A) is **cold by construction**. That is the
point — and it also means the turn answering in it does not know what else its
channel is about. "The other thread" resolves to nothing, and a channel-level
turn asked "where are we?" can speak only for the channel line.

This is the orientation, folded into the **prompt** rather than back into
history — because history is what A scoped, and widening it again would undo A.
Same seam, same terms as its three siblings: in-memory only, never journaled.

## Both directions, one rule

The index is *the channel's other threads*: it excludes the thread you are
answering in, and a channel-level turn is in none, so it sees them all. The
epic's "both directions" falls out of one exclusion rather than needing two
mechanisms.

```
- "draft the launch email" — 4 replies
- "what's our Q3 CAC?" — 1 reply
- "build the migration plan" — finished → In review
```

That last line is **B paying off**: a thread can say where its work landed
because a card records which thread raised it.

## The constraints are the design

Each is in the code with the reason attached, because each has a specific
failure on the other side of it:

- **Not LLM-generated.** The root's own opening words, `first_line(text, 120)`, verbatim. Summarising costs a model call per thread per turn and loses the exact words a later reference will echo — they are the discriminator, so "the launch email one" resolves.
- **Sized for deciding, never for knowing.** Opening words, state, recency, and that is the whole budget. If lines grow long enough to answer *from*, the flat channel window A removed has been rebuilt in the prompt and paid for twice.
- **Default is not to read.** The annotation carries the gate: *reference only — do not read or answer from them unless this message explicitly refers to one, and if a reference could mean more than one, ask which.* Over-reading is the failure to guard hardest: an agent that pulls three threads to be safe has silently undone A. The instruction is half the mechanism, which is why it ships in the same string as the index.
- **Liveness as a bound.** Drawn from one tail page — the same window the turn's own history comes from — so the index cannot name a conversation the turn could not otherwise have heard of. **Cheap because of G**: that page now reads from the end of the journal rather than streaming it from the head.
- **Truncation declared.** Capped at 5, newest first, `(and N older, not listed)`. A selection presented as an enumeration is answered from confidently and wrongly.

Recency is a thread's **last** activity, not when it opened, so a revived old
thread sorts above a quiet newer one — "the other one" should resolve to what is
actually live.

## Two bugs an existing test caught

`redeem_replays_the_markers_attachments` failed **reproducibly**, not flakily,
and both findings are worth a reviewer's eye.

**The budget-pause park was baking briefings into the operator's words.** It
captures what a redeem re-sends, documented as "the operator's own words" — but
it captured the *composed* turn text, so whatever the cycle appended went with
it. That was already reachable through #176's open-work briefing whenever the
agent had open cards; no test in that file has cards, so only the thread index
fired and exposed it. Fixed at the park **and** at the retire-match together:
they have to strip identically, or a briefing-laden candidate would never match
a clean parked marker and the pause would never retire.

**The index listed the message's own thread.** At channel level there is no
thread to exclude, but the operator's message is journaled before the cycle
runs — so it is an unparented root on the page, and the index showed a reader
their own message back as somebody else's conversation. Excluded by text match,
the same technique `chat_seed::strip_current_message` uses and for the same
reason (the in-memory event carries no sequence), with the same trap named: a
different thread opened with identical wording is excluded too, which costs a
line of orientation rather than showing something wrong.

## One refactor

Both briefings now share a single `tasks().list()` hoisted to the caller. C's
own comment already stated the principle ("two `list` calls to answer related
questions about one company is the cost the cheap exit exists to avoid"); E
would have made it a third.

## API Or Behavior Changes

- **No wire, schema, or console change.** The briefing is folded into the
  in-memory `OperatorMessage` handed to the brain and never journaled — the
  contract `OPEN_WORK_ANNOTATION` has had since #176.
- A turn in a channel with other live threads carries a few extra lines. A
  channel with nothing else open gets **no index at all** — an empty briefing is
  budget spent to say nothing.
- One journal read per addressed turn, bounded to a tail page.

## Tests

- `the_index_lists_the_other_threads_and_not_this_one` / `a_channel_level_turn_sees_every_thread` — both directions, from the one rule.
- `the_index_is_ordered_by_recency` — a revived old thread outranks a quiet newer one, since recency is last activity.
- `a_truncated_index_declares_what_it_left_out` — the cap, the count, and that it cuts the stale tail rather than the live head.
- `a_thread_whose_work_settled_says_where_it_landed` — B's field read back; state beats a reply count because it is what a reader is asking.
- `the_index_never_crosses_channels` — an index that crossed channels would be a wider leak than the one this epic closed.
- `an_agent_reply_is_never_a_root` — a reply is always parented to its question, so treating one as a root would invent a conversation nobody started.
- `a_threaded_turn_is_oriented_without_being_invited_to_read` — end to end, and it asserts **the gate travels with the index**.
- `a_channel_with_nothing_else_open_gets_no_index`.
- `a_message_is_not_listed_in_its_own_index` — the regression test for the second bug above.
- `the_cycles_thread_index_is_not_the_operators_request` — the "thanks!" trap, a fourth time. This block is the most work-shaped of the four, since every line is literally something an operator asked for; unstripped, a busy channel would open a card on every "thanks!", and each card would settle and add a `finished → …` line to the index that opened the next one.
- `operator_words_cuts_at_the_first_of_every_briefing` — now four markers, in several orders, since the cut is a `min` and nothing pins which lands first.

Run locally, all green:

- [x] `cargo clippy` on default, `openhuman`, `acp,runner,tinymemory` — each `-D warnings`
- [x] `cargo test --locked`

## Documentation

`THREAD_INDEX_ANNOTATION` carries why this is a prompt briefing rather than
history, why the lines are not summarised, and why the default is not to read.
No spec surface changed.

**With this, only F remains on #1890** — and its `find_thread` half is unblocked
by G.

## Summary by CodeRabbit

* **New Features**
  * Task completion markers now appear in the conversation thread where the task was created.
  * Thread context is preserved when viewing chat history or receiving live updates.
  * Activity summaries now distinguish open and settled work within relevant conversations.
  * Completion notifications are suppressed when the originating thread is already visible.

* **Bug Fixes**
  * Improved consistency between live completion markers and reloaded conversation history.
  * Added backward-compatible handling for older task records without thread information.

* **Documentation**
  * Added detailed documentation for conversation-linked completion markers and thread behavior.


sanil-23:

PR #1964, opened 2026-08-31
feat(harness): read another thread of this channel, on demand (#1890 F)

sanil-23:
## Summary

Sub-issue **F** of #1890 — *thread read tools*, and the last item on the epic.
**Stacked on #1961 (E)**, which it must be: the index E renders is where the
handles come from, and both change the same surface.

A scoped a turn to its own thread. E tells a turn what *else* its channel is
about and explicitly tells it **not to read any of it**. That leaves exactly one
gap, and it is the one the operator opens themselves:

> make it match the tone of the launch email thread

The agent can see that thread exists and cannot look at it. This is the looking.

**Escalate on demand rather than preload.** Cost then scales with how often
threads are actually cross-referenced instead of with how busy the channel is,
and *isolation breaks only where the operator asked it to*. A did not isolate
threads because cross-reference is bad — it isolated them because
**unconditional** cross-reference is.

## Scope: `read_thread` only

Exactly what the epic's own PR table scopes F to. `find_thread` was slated to
fold into G because it was unusable until the read cost was bounded; G shipped
that bounding without it, so `find_thread` now has no home and is worth its own
change rather than being folded in here silently. Noted as the remaining piece.

## The handle is an id, and that is a decision

E's index renders `- [41] "draft the launch email" — 4 replies`, and the tool
takes `read_thread(41)`.

The alternative was matching on the opening words, host-side, against the same
candidate set. It is tempting — the index stays clean, and ambiguity could reuse
E's "ask which" rule. It is also wrong, and writing the tool is what made that
clear: **a model paraphrases rather than quotes.** It sees `"draft the launch
email"` and calls `read_thread("the launch email one")`. Strict matching then
fails an obviously-correct reference; loose matching reads the *wrong* thread
while looking like success — which is the cross-thread leak A exists to prevent,
arriving through the tool instead of the seed.

An id has no matching rule to get wrong. It is either in the index or it is not,
and the model can only name what it was literally shown. The schema says
"exactly as shown in square brackets", which is a rule a model can follow.

E's "ask which" instruction stays where it belongs: governing the **operator's**
ambiguous reference, which is a human problem, rather than papering over a fuzzy
string match.

## The three constraints

- **Scoped to the current channel**, through the same `owns` predicate the seed uses. A tool able to read any thread anywhere reintroduces A's leak through the back door. The channel is ambient for the turn (a task-local, the pattern the existing `CHAT_ONLY_TURN` hint already uses) because a belt is built once per agent while a conversation changes every message.
- **Declares its truncation.** `query_company` is the cautionary case the epic names: a full log handed the orchestrator only its last ten rows and read as complete, so "we have no record of that" became a conclusion reachable from a partial list.
- **On every roster agent's belt**, not just the orchestrator's. The agent that needs it is the one answering in the channel; gating it on delegation grants would leave a desk lead able to see E's index and unable to follow any of it.

Two refusals are deliberate rather than incidental:

- **A root in another channel is refused by name**, not answered with silence. "Not yours to read" and "there is nothing there" are different facts, and the second invites a retry that will also fail. The refusal does not quote the thread it refuses.
- **A turn with no conversation may read nothing.** A dispatched card or a workflow node has no threads it is entitled to; `None` is a refusal, not a wildcard.

## API Or Behavior Changes

- **A new read-only tool on every roster agent's belt.** Read-only is what keeps it out of every approval and grant path a write would have to pass.
- **No wire, schema or console change.** It reads the company's own journal.
- One bounded tail page per call — cheap since #1890 G reads from the end.
- Registered only when an event log is wired; a host without one simply does not offer it.

## Tests

- `it_reads_a_thread_oldest_first` — the thread in the order it happened, root included.
- `it_returns_only_the_requested_thread` — a sibling thread's turns are not this thread's; reading them would be A's leak through the tool.
- `a_thread_in_another_channel_is_refused_and_says_so` — **the back door**, refused by name, and the refusal does not leak what it refuses.
- `a_turn_with_no_conversation_may_read_nothing`.
- `an_unknown_root_says_it_may_be_out_of_the_window` — honest about the bound rather than implying the thread does not exist.
- `a_long_thread_declares_what_it_left_out`.
- `a_missing_root_argument_is_a_refusal_not_a_panic`.
- `the_tool_is_read_only`.

Run locally, all green:

- [x] `cargo clippy` on default, `openhuman`, `acp,runner,tinymemory` — each `-D warnings`
- [x] `cargo test --locked`

## A note on stacking

F derives the turn's channel from `LiveStream`, because #1890 I's decoupling of
chat identity from live streaming lives on a sibling branch (#1955). When the two
meet, this reads the explicit target instead and nothing else changes — a
`read_thread` is only meaningful on a turn that has a conversation either way.
Noted at the site.

Part of #1890. **With this, every item on the epic has landed or is in flight**,
`find_thread` excepted.


sanil-23:

Issue #1936, opened 2026-08-31
Investigate shared hosting setup for Cortex DB as Open Company memory engine


Al629176:
## Summary

Investigate and design a **practical, production-ready way to run [`tinycortex`](https://github.com/tinyhumansai/tinycortex) ("Cortex DB") as a shared, hosted memory engine** that OpenCompany tenants bind to through the existing `remote` memory seam. Produce an architecture, an infra requirements list, and a concrete implementation plan.

## Background — where memory stands today

Memory is a separable overlay in OpenCompany. `OPENCOMPANY_STORAGE` picks the durable base for all fourteen ports; `OPENCOMPANY_MEMORY` optionally swaps the three knowledge ports (`MemoryStore`, `ContextStore`, `FactStore`) onto a dedicated engine bound through the `MemoryProvider` contract (`tinymemory-api`, vendored at `vendor/openhuman/vendor/tinymemory/api`). See [`docs/spec/runtime/memory-engine.md`](https://github.com/tinyhumansai/opencompany/blob/main/docs/spec/runtime/memory-engine.md) and [`docs/spec/runtime/storage.md`](https://github.com/tinyhumansai/opencompany/blob/main/docs/spec/runtime/storage.md).

Current `OPENCOMPANY_MEMORY` modes:

| Value | Engine | Notes |
|---|---|---|
| `store` (default) | Base backend's own memory | fs substring recall, or sqlite/mongodb |
| `remote` | Hosted memory service | Bound via `MemoryProvider`; needs driver + URL + credential. Drivers today: `supermemory`, `mem0`, `cognee` |
| `null` | Nothing | Writes discarded, reads empty |

**The in-pod `embedded`/`tinycortex`/`cortex` engine was removed in #1568** and now *refuses at boot* if still selected. So Cortex is not currently a first-class OpenCompany memory option. This issue asks the opposite question from #1568: rather than embedding Cortex in every pod, **can we host it once, centrally, and let every tenant bind to it over the `remote` seam** — putting Cortex on equal footing with `supermemory`/`mem0`/`cognee`, but as our own engine?

`tinycortex` = "The Fastest AI Memory Model — Your Second Brain"; `tinymemory` = the agent-friendly memory contract/driver layer the seam already speaks.

## Goal / outcome

A design doc (proposed home: `docs/spec/runtime/orchestration/` or a new `docs/plans/NN-cortex-hosted-memory.md`) that covers:

[...]

PR #1967, opened 2026-08-31
docs: record the hosted-Cortex investigation for #1936

CodeGhost21:
Design record for #1936. Adds `docs/spec/runtime/memory-engine-cortex.md` and links it from `memory-engine.md`. Docs only — nothing here is implemented.

I deployed a CortexDB instance and exercised it rather than answering from the product page, and three of the findings change the shape of the question.

**The isolation choice the issue poses collapses.** The self-hosted build cannot issue per-tenant credentials at all: the PASETO minter answers `NOT_CONFIGURED` under `cloud_shared_saas`, its keypair generator is not in the distribution, and the bundled `PRODUCTION_DEPLOYMENT.md` describes `CORTEX_API_KEY` as the bootstrap key for the *default* tenant, with real per-tenant keys coming from Cortex's own hosted key store. So "one shared Cortex with namespace separation" is definitionally the weak tier — not something better configuration reaches. Instance-per-tenant is the only self-hosted route to the tier `memory-engine.md` already assumes, which means "shared hosting" becomes shared *infrastructure*, not a shared engine process. That is a real departure from the issue's framing and the main thing worth arguing about here.

**Three of the five layers advertise and return nothing.** Facts, Beliefs and Understanding stay empty with the extraction and enrichment routers enabled and reporting healthy. Meanwhile the routes exist, boot logs say the routers are enabled, and `/v1/derivation/status` returns `built: {episodes: 3, beliefs: 3}` for a scope whose belief store is empty — those counters count scopes processed, not records produced. A `cortex` driver deriving `provides()` from routes, config, or engine self-report would pass `audit_capabilities` and then return empty inside a tenant, at the moment memory is needed. The doc makes `provides()`-from-observed-contents an acceptance gate on the driver plan.

**Cortex inverts the seam's no-silent-degradation rule.** With no embedding credential it does not refuse — it falls back to mock embeddings, pins the data dir to `mock::1536`, and reports `{"status":"healthy"}`. Recall then returns confident, meaningless results. A driver has to probe for this and refuse the bind itself; the engine will not.

**What does work** is worth stating plainly: ranked retrieval is genuinely good and needs no LLM lanes at all, and scope isolation held across every test. Cortex can back `MemoryStore` and `ContextStore` today. It cannot back `FactStore` with anything the layer model promises.

Two upstream defects are filed at cortexdbai/cortexdb-releases [#1](https://github.com/cortexdbai/cortexdb-releases/issues/1) and [#2](https://github.com/cortexdbai/cortexdb-releases/issues/2).

Phase 0 in the plan is licensing and topology — both decisions rather than engineering, and both gate the rest. My own recommendation is to park the driver work until the upstream defects are answered, since without Facts and Beliefs, Cortex is at parity with the `supermemory`/`mem0`/`cognee` drivers we already ship.

The weakest evidence in the doc is the ~18 GiB per-instance projection, and it is labelled as such — it comes from the binary's own config lint and did not move under any remediation the lint itself suggests. A real figure has to come from Cortex before any capacity plan is credible.

## Summary by CodeRabbit

* **Documentation**
  * Added a detailed design record evaluating CortexDB as a hosted memory engine.
  * Documented storage and memory responsibilities, namespace translation requirements, configuration constraints, licensing gates, and deployment considerations.
  * Clarified that Cortex can support all three knowledge ports, with embedded adapter conformance and HTTP exposure still under consideration.
  * Recorded capability-audit limitations, migration considerations, and an open question about fallback use.
  * Confirmed hosted Cortex support is investigational and not implemented.


CodeGhost21:
Pushed a second commit. While scoping what a `cortex` driver would actually involve, I hit a blocker that changes the phasing, so it is now recorded in the doc rather than left as a surprise for whoever picks up Phase 1.

**A conformant driver cannot be written against CortexDB v0.9.8 at all.** Not a capability gap — the data models disagree.

`Memory::store` is an idempotent upsert keyed by `(namespace, key)`, and the conformance suite asserts it directly (`assert_upsert_replaces_rather_than_duplicates`: write `first`, write `second`, expect one row holding `second`). CortexDB is an append-only event log:

| Attempt | Result |
|---|---|
| Same key, different content | `409 IDEMPOTENCY_CONFLICT` |
| Update route | none — events are GET-only, `/v1/experience` POST-only |
| Forget, then rewrite the key | forget succeeds, rewrite still `409`, **scope left holding nothing** |
| Carry our own key in the envelope | `422` — closed schema, and no metadata filter to query it by anyway |

The idempotency ledger is independent of the event store and survives `/v1/forget`, so delete-then-rewrite is strictly worse than not attempting it.

The only technically viable adapter keeps an external `(namespace, key) → event_id` index, writes with fresh idempotency keys, and forgets the prior event on overwrite — making the driver stateful across two non-atomic calls, on top of a `forget` that reported `deleted.events: 2` for a selector naming one event id. That is a lot of correctness risk to absorb for something an upstream `on_conflict: replace` removes entirely.

Filed as [cortexdb-releases#3](https://github.com/cortexdbai/cortexdb-releases/issues/3). Phase 1 is now marked **blocked on that, not on our effort** — which I think is the useful thing for this doc to say, since it means no amount of adapter work unblocks it.

CodeGhost21:
Correcting something in the doc: **I had licensing wrong, and it is not a blocker.**

I had been reading the release repo's README, which summarises the terms as "the only thing not permitted is reselling CortexDB as a hosted service to third parties". The licence text itself is more specific, and clause 2 permits exactly what this design does:

> The Software may be used to power internal or commercial applications, **including those sold to third parties**, provided that: (a) the third party does not access the Software directly as a general-purpose memory database (i.e., **you may build products on top of CortexDB and sell those products**; you may not resell CortexDB itself as a service); and (b) attribution to "CortexDB" appears in product documentation where reasonable.

Selling OpenCompany with Cortex behind the memory seam is building a product on top, not reselling the database. Condition (a) holds **by construction** here: tenants reach memory only through the `MemoryProvider` seam, and `memory-engine.md` guarantees the credential and endpoint never appear "in logs, `/healthz`, `/spec`, status output, or an export". A tenant cannot address the engine directly even deliberately.

Two obligations follow, neither of them a gate:

- **Never hand a tenant its own Cortex endpoint or credential.** Under instance-per-tenant that would be an easy thing to do casually — a BYO-engine option, or surfacing the URL in a console — and it is exactly what (a) forbids. The seam's existing redaction already prevents it; the point is to keep it that way.
- **Attribute CortexDB in product documentation** (clause 2b).

Phase 0 no longer lists licensing as blocking; the topology decision is what remains there.

**What has not changed:** the upsert gap. That is still the thing that makes a conformant driver impossible on v0.9.8, and it is still the only question whose answer decides whether Phase 1 can start at all.

CodeGhost21:
Thanks — pushed `122225c46`. Four of the findings were right, including one that caught a real inconsistency in the argument rather than a typo.

**The substantive one (empty tenant vs unsupported layer).** Correct, and it contradicted my own reasoning. I wrote that `provides()` should be "derived from observed layer contents", but on a freshly provisioned per-tenant instance *every* layer is legitimately empty, so that rule would refuse every family on day one. Emptiness means "new" and "inert" equally — I had noted exactly that ambiguity in #1968 and then wrote the requirement here as though it were solved.

Reframed: the signal has to be that a family *functions*, not that it currently holds data. For Cortex the available discriminator is a build endpoint reporting work done against a store that stays empty (`/v1/beliefs/build` → `built: 0` with `no_facts_in_scope`, while `derivation/status` claims success). That is engine-specific, so it belongs in each adapter and therefore in the conformance suite — which runs against a live engine and would cover every driver by construction — rather than in the host. A freshly provisioned empty instance is now an explicit acceptance case, since a suite that only tests a populated instance would pass a driver that fails precisely there.

**ScopePath regex.** Fixed. I had elided the repeating segment with a literal `…`, which is not valid regex in something that reads as normative. Now written out in full.

**Spelling.** Fixed — "license" throughout, matching the license's own name.

**Clause 2 conditions.** Agreed, and made concrete: not exposing a tenant its own engine URL or credential is now a named **provisioning gate** owned by Phase 2, and the attribution requirement a **release gate** that has to be satisfied before the first tenant is served.

**One deviation:** the fenced block was flagged for a `json` tag, but it carries a `POST /v1/beliefs/build` request line as well as the JSON response, so it is not valid JSON. Tagged `text` instead.

oxoxDev (changes requested):
The investigation is genuinely useful and most of it holds up — I verified twelve of its factual claims against `upstream/main` @ `25539f6b6` and the pinned tinymemory, including all four `memory-engine.md` quotes (verbatim), the conformance-suite behaviour, the port counts, the storage-mode list, and every issue reference. Currency is clean: nothing on main has touched `src/store/memory/`, `memory-engine.md`, `select.rs` or the `vendor/openhuman` gitlink since this opened.

The problem is the capability model. The document describes the seam in terms of OpenCompany host ports and Cortex layers, and neither is what a driver actually advertises — which carries into the headline conclusion and the Phase 1 gate. A reader who trusts it builds the wrong driver.

## The capability model

**There is no Facts / Beliefs / Understanding capability family.** The contract's `Capability` enum is a closed set (deliberately not `#[non_exhaustive]`, so adding one is a compile error rather than config) and contains no such variant. Cortex's five *layers* are not families, so `audit_provider` cannot fire on them. If there is an over-claim risk it lives in `Ingest` / `Entities` / `Tree` / `Retrieval`, and the doc should name which.

**"Cortex cannot back `FactStore`" is false**, and it is the document's headline. `MemoryStore`, `ContextStore` and `FactStore` are host ports, not driver families — a driver can neither advertise nor withhold them. All three are facades over the same `Bound` helper using only `store/get/list/forget/recall`, i.e. `Core` + `Recall`, both mandatory supertraits. Concretely:

```rust
// src/store/memory/facades.rs:389-391
async fn upsert(&self, company: &CompanyId, fact: &FactRecord) -> Result<()> {
    self.bound.put(company, &fact.id, fact, "fact").await
}
```

That writes operator-curated records into `oc/<company>/facts` via `MemoryCore::store`. It never reads a derived facts layer, so Cortex's empty `/v1/facts` is irrelevant to it. The true statement is narrower and still supports your conclusion: Cortex cannot deliver the *derived* fact/belief tier — which is a reason it offers nothing over `supermemory`/`mem0`/`cognee`, not a reason a port fails.

**The namespace claim is backwards.** `Namespace::company_root` emits `oc/<slug>-<32 hex>`, children append a segment (`…/context`, `…/facts`, `…/agent/<member>`), and `sanitize_segment` maps everything outside `[A-Za-z0-9-_]` to `_` — so a colon can never appear and no segment is ever `type:id`. Every store and recall would be rejected. A `cortex` driver needs an explicit **reversible** translation, and reversible is not optional: `Bound::recall` re-checks each returned entry with `Namespace::contains` and drops mismatches, so an unmapped return silently yields zero hits.

**`provides()` cannot be established against a live engine.** It is a defaulted trait method with a fixed body — `Core | Recall | Portability` hardcoded true, the rest `self.as_X().is_some()` — a structural Rust-type question, not a runtime one. Its own doc calls it "the implementation-side truth, as opposed to `capabilities()`, which is the advertised claim". Asking it to probe the engine would make `audit_provider` compare two runtime opinions instead of a claim against a structure, and the acceptance test as written is not something the audit can express — the audit never talks to the engine. The concern is real and it is exactly what **#1968** exists for; the lever is a live probe / conformance case, not `provides()`.

## Also

- `audit_capabilities` doesn't exist — it's `audit_provider`. The stale name comes from `memory-engine.md:82`, but this PR propagates it into a new file *and* is already editing that file, so it's one line in each.
- The Phase 1 gate names only upsert, but imprecise deletion is equally disqualifying — the doc itself records a `forget` reporting `deleted.events: 2` for a one-id selector. `Bound::forget`, `evict`'s archive-then-delete loop and `delete_label` all need key-exact deletion; an over-deleting `forget` destroys tenant memory even with upsert working.
- "The only technically viable adapter keeps an external index" overstates. The host already wraps records in a JSON envelope inside `content`, so an in-content `(key, content)` envelope resolved by a client-side scan is the same shape it already uses — slow and non-atomic, which is a fair objection, but not "only viable".
- Phase 1 is gated on cortexdb-releases#3, which is CLOSED; the doc notes that caveat for #1 and #2 but is silent on #3.
- Worth one sentence: `cortex` is registry-clean as a driver id, but `OPENCOMPANY_MEMORY=cortex` is still a hard boot refusal as a *mode* value. An operator reading Phase 1 will hit that.

## On the bot rounds

For what it's worth, most of the timeline noise isn't review content — seven Codex entries are the same rate-limit notice and three more are your own progress notes. The real content is five inline P2s, and three of them (namespace translation, layers-vs-families, precise deletion) are open, correct, and unanswered — they're the same three things above. CodeRabbit's CHANGES_REQUESTED is stale against head; both its rounds were cosmetic and are addressed in `122225c`, including the fence tag, where deviating was right since the block contains a request line and isn't valid JSON.

None of this changes the investigation's conclusions — it's a rewrite of the capability section, the `FactStore` sentence, the namespace paragraph, and the Phase 1 gate.

One aside, not for this PR: `assert-md-line-cap.sh` rides the Console job, which is path-skipped for docs-only changes — so the 500-line cap is unenforced on exactly the PRs that add documentation. Moot at 382 lines, but probably worth its own issue.

CodeGhost21:
This is a much better review than the document deserved — thank you for reading it against the code rather than for plausibility. All three substantive findings are correct; I verified each before changing anything, and each was a real error, not a wording problem. Pushed `4dc697f6d`.

**Layers are not capability families.** Correct, and this was the worst of the three because it pointed the driver plan at the wrong thing entirely. The `Capability` enum has no Facts/Beliefs/Understanding variant, so `audit_provider` cannot fire on Cortex's empty layers however empty they are. Rewritten as "Layers are not capability families": the empty tier is a **commercial** finding — it removes the reason to prefer Cortex over `supermemory`/`mem0`/`cognee` — not a bind-time one. I also took the point that any genuine over-claim risk would live in `Ingest`/`Entities`/`Tree`/`Retrieval`, and said a driver plan has to name which it advertises and on what evidence.

**"Cortex cannot back `FactStore`" was false.** Also correct, and it was the headline. `FactStore::upsert` is `self.bound.put(company, &fact.id, fact, "fact")` — operator-curated records through `MemoryCore::store` into `oc/<company>/facts`, never a read of a derived layer. The narrower true statement is now the one in the doc, and as you say it still supports the conclusion.

**The namespace claim was backwards.** `sanitize_segment` maps everything outside `[A-Za-z0-9-_]` to `_`, so a colon cannot appear and no segment is ever `type:id` — passed through unchanged, every call is rejected. Rewritten to say the formats are *incompatible* and a driver needs an explicit reversible translation, with your reason for why reversible is mandatory rather than tidy: `Bound::recall` re-checks with `Namespace::contains` and drops mismatches, so an unmapped return is silent zero hits rather than an error. That is a much worse failure than a rejection and it is now stated as such.

**`provides()`.** Agreed, and the previous wording was not implementable — it is a defaulted method with a fixed body, and asking it to consult the engine would have the audit compare two runtime opinions instead of a claim against a structure. The Phase 1 gate now names a live probe or conformance case. The related work is #1968, and #1973 proposes the boot-time probe of the mandatory families, which is where I think this actually belongs.

**The smaller ones**, all taken: key-exact deletion added as a Phase 1 gate — you are right that an over-deleting `forget` destroys tenant memory even with upsert working, and `deleted.events: 2` for a one-id selector is disqualifying on its own until explained; "only viable adapter" softened, since an in-content envelope is the shape `Bound::put` already uses via `encode(record)`; cortexdb-releases#3 now carries the same closed-tracker caveat as #1 and #2; the `OPENCOMPANY_MEMORY=cortex` mode refusal is called out for whoever starts Phase 1; and `audit_capabilities` → `audit_provider` in both files.

**On `assert-md-line-cap.sh`** — good catch, and I agree it is not this PR's business. Filing it separately.

oxoxDev (changes requested):
Re-reviewed at `df6d9fd2`. **Everything I filed last round is closed**, and I checked the replacement text against source rather than just confirming it changed:

- The new "Layers are not capability families" section is accurate — the enum is a closed set of 21 variants with no Facts/Beliefs/Understanding, and naming `Ingest`/`Entities`/`Tree`/`Retrieval` as where the real risk sits is right.
- The `FactStore` correction is exactly right, including quoting `bound.put(company, &fact.id, fact, "fact")` and the narrower true claim about the derived tier.
- The namespace section is now inverted correctly, and it keeps the part that matters most — reversibility isn't a nicety, because `Bound::recall` filters on `Namespace::contains` and an unmapped return yields zero hits *silently*.
- `provides()` is now described as the structural, defaulted method it is, with the lever pointed at #1968/#1973.
- `audit_capabilities` is gone from the diff entirely, and key-exact deletion is now a hard Phase 1 acceptance item.
- All three minors closed, including the honest "blocked on their answer, not on our effort" framing for the CortexDB tracker.

One new thing, and it's the same class as the `FactStore` error — driver obligations described in terms of what the host happens to exercise.

## The mandatory set is three, not two

Two sentences say otherwise:

> "So the three **mandatory** families can never fail the audit, and they are exactly the ones this host's knowledge ports are built on."

> "A `cortex` driver over `tinymemory-api` advertising only the families Cortex actually serves — the mandatory `Core` and `Recall`."

At the pinned `tinymemory@a764d6b`:

```rust
// crates/tinymemory-api/src/provider/driver.rs:84
pub trait MemoryProvider: MemoryCore + MemoryRecall + MemoryPortability + 'static {
```

`advertised_capabilities()` returns `{Core, Recall, Portability}` — "the mandatory three, and nothing else". So a reader takes two false things from this. First, that the knowledge ports depend on Portability; they don't — `Bound` never calls a Portability method, and the doc says so correctly elsewhere ("`Core` plus `Recall`, both mandatory supertraits"). Second, and more costly, that a Phase 1 `cortex` driver is a two-family job. It isn't: `MemoryPortability` is a supertrait, so the driver must implement `export_page`/`import_records` over an append-only event log **to compile at all** — and Phase 3 of this same plan runs migration over that family. Under-scoping the acceptance list is how that lands as a surprise mid-build.

Suggested: "the mandatory three can never fail the audit; two of them, `Core` and `Recall`, are what the knowledge ports are built on", and for Phase 1 "the mandatory three — `Core`, `Recall` and `Portability` — since `MemoryProvider` requires all three as supertraits; only `Core` and `Recall` are exercised by the knowledge ports".

## Minor

The §Licensing quote — the credential and endpoint "never appear in logs, `/healthz`, `/spec`, status output, or an export" — is attributed to `memory-engine.md`, but that exact string is `storage.md:104`. `memory-engine.md:60` says the same thing in different words ("Neither appears in…"). The claim is true either way; just re-quote or re-attribute. The other two attributed quotes check out verbatim.

Also still open but not yours: the `audit_capabilities` origin at `memory-engine.md:82` is stale, and this PR edits that file. #1978 doesn't fix it either — worth catching in one of the two.

CodeGhost21:
Pushed `8d42e6651`. Both findings correct, and the first is the same class as the `FactStore` one — describing a driver's obligations in terms of what the host happens to exercise.

**The mandatory set is three.** Verified at the pin: `pub trait MemoryProvider: MemoryCore + MemoryRecall + MemoryPortability + 'static` (`provider/driver.rs:84`), and `advertised_capabilities()` returns all three unconditionally (`mandatory/provider.rs:85`). So both of your consequences hold: the knowledge ports do not depend on Portability, and a Phase 1 driver is not a two-family job — it cannot compile without `export_page`/`import_records`.

I took your suggested split, and added the part that makes it matter rather than just correct: Portability is what **Phase 3 runs migration over**, and implementing it against an append-only event log is its own problem, not glue. Under-scoping it was the actual cost, exactly as you said.

**The quote attribution.** Right — the exact string is `storage.md:104`; `memory-engine.md:60` says the same thing in different words. Re-attributed to `storage.md` with a note that the sibling states the same rule, so the claim keeps both sources.

**The stale `audit_capabilities` at `memory-engine.md:82`** — caught here, since this PR already edits that file and #1978 did not touch it. Now `audit_provider`, called from `src/store/memory/driver.rs`.

For completeness on the CodeRabbit round: all three of its comments were already closed by `df6d9fd2` before I got here — the gate contradiction, the Phase 1 port/family wording, and the `oc/<slug>-<32 hex>/facts` namespace form. I checked rather than assuming, and left them alone.

Doc is 424 lines, still inside the cap — which this PR now actually gets checked against, since #1975 moved that job off the path-filtered Console lane.

PR #1978, opened 2026-09-01
docs: correct the tinymemory-api path, versions, and the argument built on them

### LINKED ISSUE #1976: memory-engine.md names a tinymemory-api path that does not exist at the pinned commit
## Summary

`docs/spec/runtime/memory-engine.md:146` names the contract crate's path as:

> `tinymemory-api`, at `vendor/openhuman/vendor/tinymemory/api` — the same path `vendor/openhuman` itself path-depends on, which is what keeps the `MemoryProvider` trait identity single across the process.

That directory does not exist at the pinned submodule commit. The crate lives at `vendor/openhuman/vendor/tinymemory/crates/tinymemory-api`; the repository is a cargo workspace with everything under `crates/`.

```
$ ls vendor/openhuman/vendor/tinymemory/api
ls: no such file or directory

$ ls -d vendor/openhuman/vendor/tinymemory/crates/tinymemory-api
vendor/openhuman/vendor/tinymemory/crates/tinymemory-api
```

## Why it is worth fixing rather than leaving

The sentence is not incidental — it is the paragraph that explains *why* the path matters: single trait identity across the process, which is the thing a wrong path would quietly break. Somebody checking that invariant follows the path, finds nothing, and has to reconstruct the layout to know whether the claim still holds.

The same stale path appears in the issue body of #1936 (quoting the spec), so it has already propagated once.

## Fix

One line in `memory-engine.md`. Worth a quick grep for the same string elsewhere in `docs/` while in there.

## Provenance

Noticed by @oxoxDev reviewing #1973. Deliberately not folded into that PR, which is about the boot probe.

CodeGhost21:
Closes #1976.

## The reported bug

`memory-engine.md:107` and `orchestration/memory.md:126` both name the contract crate at `vendor/openhuman/vendor/tinymemory/api`. That directory does not exist at the pinned commit — tinymemory is a cargo workspace and the crate is at `crates/tinymemory-api`.

The *claim* the sentence makes is still true, which is why this is a path fix rather than a rewrite. Both manifests already point at the real location:

```
vendor/openhuman/Cargo.toml:252  tinymemory-api = { path = "vendor/tinymemory/crates/tinymemory-api" }
Cargo.toml:300                   tinymemory-api = { path = "vendor/openhuman/vendor/tinymemory/crates/tinymemory-api" }
```

Same directory, so Cargo still unifies them and the `MemoryProvider` trait identity is still single across the process.

## Two more drifts at the same site

Checking the rest of that table rather than just the one cell:

**`CONTRACT_VERSION` is `(3, 0)`, not `(2, 0)`.** (`tinymemory-bus/src/version.rs:63`.)

**`tinycortex-api` no longer declares `(1, 0)`.** It re-exports the constant straight from `tinymemory_api` (`vendor/openhuman/vendor/tinycortex/api/src/lib.rs:87`), so both crates now report the same version.

That last one mattered beyond the number, because the prose was built on it:

> `is_compatible` is major-equality only, so the two contract crates are declared incompatible

They are not, and have not been since the re-export landed — `is_compatible` now answers *compatible* for the pair. The separation rests entirely on crate identity: separate crates, so distinct `MemoryProvider` types whatever the version says. That is the stronger guarantee but a silent one — binding the wrong crate is a type error, not a version refusal — so the paragraph now says that instead of pointing at a check that no longer fires.

## Scope

Docs only, two files. The tinycortex rows' *paths* are still accurate (`vendor/openhuman/vendor/tinycortex/api` exists and both manifests still path-dep it), so they are unchanged apart from the version cell.

`assert-md-line-cap.sh` passes — and thanks to #1975 it will now actually run on this PR, which is docs-only.

## Summary by CodeRabbit

- **Documentation**
  - Updated memory engine documentation to reflect the current `tinymemory-api` crate location.
  - Updated contract versions and compatibility guidance for memory-related APIs.
  - Clarified that compatibility is determined by matching versions, while distinct provider interfaces remain type-safe and separate.


oxoxDev (changes requested):
The three version claims are all correct — I verified each at the pinned refs rather than the on-disk vendor (which is at an older layout and would have given wrong answers):

- `CONTRACT_VERSION = (3, 0)` — `crates/tinymemory-bus/src/version.rs:63` at `tinymemory@a764d6b`.
- `tinycortex-api` genuinely re-exports rather than defining — `tinycortex/api/src/lib.rs:87` is `pub use tinymemory_api::{is_compatible, CONTRACT_VERSION};` and its `[dependencies]` is a single entry.
- `is_compatible` is major-equality — `version.rs:85-87`, "Compatible exactly when the major halves match."

So the rewritten argument is sound: both crates report the same constant, `is_compatible` answers *compatible*, and the separation rests on crate identity. The path correction is right too — `api/` doesn't exist at that pin.

One mechanism worth a footnote, since the claim leans on it: `tinycortex/api/Cargo.toml` takes `tinymemory-api` **by git rev**, not by path. It resolves to the same `(3, 0)` only because our root `Cargo.toml` carries a `[patch."https://github.com/tinyhumansai/tinymemory"]` entry pointing at the vendored crate. True in our build; a property of the patch table rather than of tinycortex-api's manifest.

## The next sentence in the same paragraph is now false

The rewrite fixes the two cells and then leaves this three lines below:

> "OpenHuman has additionally inlined the v2 contract into its own tree at `vendor/openhuman/src/openhuman/memory/api/`, and that inlined copy is what its live binding uses…"

At `openhuman@c6a06ab`, three things are wrong:

1. **The path.** There is no `src/openhuman/memory/api/` directory — it's `api.rs`, a single 6,372-byte file. I checked the tree at the pin.
2. **"inlined … and that inlined copy is what its live binding uses"** is the opposite of the truth, and the file says so itself. Its heading is *"Why this file is nine lines and not ten thousand"*, and it explains: *"This module used to be a directory holding 10,894 lines … Nothing behaved differently the day that landed, which is exactly what made it worth undoing."* The body is `pub use tinymemory_api::{…, CONTRACT_VERSION}`. The binding uses the crate.
3. **"the v2 contract"** — it's `(3, 0)`, which this PR corrects two cells above. The stale major survives in prose in the same paragraph it's fixing.

The quoted phrase is off too: openhuman's own text reads "`tinycortex-api` is now a deprecated re-export of **`tinymemory-bus`**" — of the bus crate, not of openhuman's module.

Roughly: "OpenHuman no longer keeps its own copy of the contract — `src/openhuman/memory/api.rs` is a re-export of `tinymemory_api`, and its live binding uses the crate directly. It treats the string `tinycortex` as a legacy driver-id alias." That last part is correct as written (`binding.rs:219`, `LEGACY_MODULE_ID`).

## Nit

The table marks `tinycortex-api` as "`(3, 0)`, re-exported" but gives `tinymemory-api` a bare `(3, 0)`. That one is also a re-export — the constant is defined in `tinymemory-bus`, as the PR body itself cites. A reader concludes `tinymemory-api` defines it. One word.

Worth knowing: both bot approvals are stamped at `2909f277` and head is now `770ead0`, so neither describes the current diff. Neither touched the version claims or the trailing paragraph.

PR #1980, opened 2026-09-01
Group the console title row into three clusters, and move Overview and Approvals into it

graycyrus:
Step 3 of the console IA restructure: the window title row is designed rather than accumulated, and the two jumps you make from anywhere move into it.

```
●●●  ⌂ Northwind Design Collective ▾  ······  ⌂ ⛨7 │ ∞ Full  Broadest execution autonomy. ▾ │ ◉
                                              go            state                            you
```

## What changed

**Three groups, hairline-separated.** Where you are going, what the agents may do, and you. The rules do what a gap cannot at any width that also fits: they say *these are different kinds of thing*, so the eye reads three objects rather than five loose controls. Each group carries its own divider as a `::before` rather than as an element of its own, so it dies with `empty:hidden` when that group's control renders nothing — the autonomy pill when the tier is unknown, the profile control on a host with no sign-in. A standalone divider would survive both as a stray mark.

**Overview and Approvals become glyphs in the first group.** A labelled button in a chrome band reads as content, which is the reverse of the trade the sidebar makes. The names move to `aria-label` and `title`; only the pixels are lost.

**The degradation ladder is decided in one object.** `TITLE_BAR_LADDER` in `window-title-bar.tsx` — the autonomy sentence below 1280, the company's name below 1024, the Overview glyph below 768, and never approvals, autonomy or you. `AutonomyPill` and `HostSwitcher` consume a rung rather than each choosing a breakpoint; a unit test fails if either restates one locally.

**The rail dot is deleted with the badge.** `SidebarMenuDot` existed for exactly one reason (issue #1018): `SidebarMenuBadge` carries `group-data-[collapsible=icon]:hidden`, so the approvals count — the sidebar's only attention signal — vanished the moment the rail collapsed, and a collapsed rail showing nothing is indistinguishable from all-clear. The title row does not collapse, so the disappearance cannot happen and both are removed rather than one kept. The primitive is deleted rather than left unused, with the reasoning left in its place.

What #1018 was *about* is kept: the signal must reach someone who is not reading a number. For an icon-only control the accessible name is the whole of what a screen reader gets, so the dot's `aria-label` sentence — "N approvals need you" — is now the button's name, word for word.

## Two decisions worth reviewing

**The autonomy sentence now goes at 1280 (`xl`), not 1024 (`lg`).** The plan's ladder table gives 1280; the code said `lg`. At `lg` the sentence and the company name went at the same width, so two rungs landed together and the ladder had one fewer step than it claimed. `autonomy-pill.test.ts` updated accordingly.

**The count sits beside the glyph, not on it.** Caught by validating at the real cap: as a corner mark, `99+` at 128 pending covered most of the shield and the control stopped reading as approvals. With something waiting the button grows to hold its count; height is unchanged, so the row's one `items-center` rule still holds.

## Not in this PR

The Approvals and Overview rows are still in `NAV` — removing them belongs to the sidebar restructure landing in parallel. `tour/steps.ts` still anchors on `nav-approvals`, which still exists, so the tour is unaffected by this change.

## Verified

- `npm run typecheck`, `npm run typecheck:unit`, `npm run typecheck:e2e` — all three clean.
- `npx vitest run` — 442 files, 4144 tests, 0 failures.
- `scripts/ci/assert-design-tokens.sh` — clean.
- Every new test mutation-proved: 11 mutations, each killing the intended test (pluralisation, the zero state, the count cap, the chip's `aria-hidden`, `aria-current`, the `pending` wiring, the hairline token, `empty:hidden`, the Overview rung, a locally restated breakpoint, and group order).
- Chromium at 1400 / 1200 / 900 / 700 / 390, light and dark, against a real host serving `companies/e2e_harness`: each rung drops exactly the item it should, nothing wraps and nothing scrolls at 390, and the collapsed rail carries no dot while the count stays visible in the row.


graycyrus:
@coderabbitai review

graycyrus:
Follow-up: **#1983** carries two corrections that were written against this PR's
tree and were still in flight when it merged at 10:15:49Z.

- **`docs/design-system/color.md` and `components.md`** illustrate "anything that
  cuts a hole in the chrome must ask for the chrome" with two examples, one of
  which is `SidebarMenuDot` — deleted here. `ring-chrome` now has exactly one
  call site left (`host-switcher.tsx:285`), so the example points at nothing.
- **`overview-button.tsx` and the `overview={...}` comment in `app-shell.tsx`**
  say "It was a labelled sidebar row", but `NAV` still carries both rows — as
  this PR's own description says under *Not in this PR*. What moved out of the
  sidebar here is the count, not the rows.

Neither came from a reviewer: this PR closed with **0 inline review comments and
0 review threads**, and CodeRabbit was rate limited for its entire open life and
never posted a review (its two comments here are the limit notice and a
"review rate limited" reply to an on-demand request). The only review on it is
tinysweeper's approval.

One thing worth a look on the merged commit rather than on #1983: this merged
while **`Console E2E` and `Console E2E (live brain)` were still `in_progress`**
on `9f5ea03a7` — the two lanes that actually drive a browser against a console
change. They are still running; their result lands on the merged SHA.

PR #1997, opened 2026-09-01
fix(console): keep the echo-brain notice on the read-only feed

graycyrus:
Follow-up to #1984, which merged while this was being verified. It fixes the
one review finding that arrived after the merge — a regression #1984 itself
introduced.

## The finding (codex, P2, on #1984)

> **Keep the echo warning visible in the read-only feed.** Suppressing the
> banner there removes the visible explanation and the Settings remedy from
> precisely that feed. The remaining chip just says "Placeholder", and its
> explanation exists only in a `title` on a non-focusable `<span>`, so touch and
> keyboard users cannot discover that the named teammate did not author the
> report.

Valid. #1984 gated the cognition banner on `echoing && !readOnly`; before it,
the banner was not channel-aware at all. **This removes the gate.**

## Why the suppression was wrong on its own terms

It was justified by "nothing can be sent here, so a caveat about what sending
produces has nothing left to qualify". But the sentence is *"the replies in this
conversation come from the offline echo brain rather than the teammate they
appear under"* — a claim about **attribution of what is already on screen**.
None of the four states mentions sending. It is fully true on a read-only feed,
and a feed the reader cannot reply to is where they are least able to test the
attribution by asking, so it is the last place to drop the only visible
statement of it. The copy therefore needed no read-only variant either.

The premise was checked rather than taken on trust: `MessageRow` calls
`echoMarkerFor(message, sender, cognition)` for every company-side row with no
`byPerson`, on **every** channel — nothing about it is channel-aware — so
`#Operator`'s workflow reports, rendered under a teammate's name and avatar, are
marked exactly like a channel reply. And `EchoPlaceholder` really is a
non-focusable `<span>` whose reason lives only in a `title`.

With no composer on that channel, the strip lands under the read-only notice as
the bottom strip of the pane.

## Docs

`docs/spec/runtime/ports-cognition.md` already promised a banner "on anything
but `configured`" and said nothing about where it sits. It now says the strip is
directly above the composer, below the typing line, and kept on a read-only
channel — and why.

## Verification

```
$ npm run typecheck        # tsc -b --noEmit             → clean
$ npm run typecheck:unit   # tsc -p tsconfig.unit.json   → clean
$ npm run typecheck:e2e    # tsc -p tsconfig.e2e.json    → clean
$ npx vitest run           # Test Files 445 passed · Tests 4197 passed
$ scripts/ci/assert-design-tokens.sh
✓ design tokens: no arbitrary sizes, no palette colours, no stray hex
```

Two new tests, both mutation-proved by restoring `!readOnly`:

| Test | Failure under the mutation |
|---|---|
| the read-only feed still states the false attribution | `expected null not to be null` |
| it sits under the read-only notice, as the last child of the column | `Cannot read properties of null (reading 'parentElement')` |

**Browser, light and dark, 1440×900**, `#Operator` against a real host with
`cognition: unconfigured`: no composer, no textarea, no `data-tour` anchor, and
DOM order `["TRANSCRIPT", "READONLY-NOTICE", "BANNER"]` with the banner as the
last child — reading *"Teammates can't think yet. This company has no model
configured, so the replies in this conversation come from the offline echo brain
rather than the teammate they appear under. Choose a provider in Settings →
Inference."*

#1984's own three fixes were re-run on this build and still hold: the draft
survives a trip to `#Operator`; the tour walks **8 of 8** stops from the Operator
feed; with someone typing, the order is `["TRANSCRIPT", "TYPING", "BANNER",
"COMPOSER"]` with nothing between the notice and the composer.

## Not in scope

`EchoPlaceholder`'s chip carries its reason only in a `title` — a pre-existing
gap, not one #1984 created, whose fix is an accessible name on the chip. With
the banner restored it is no longer the only explanation on any surface. Worth
its own issue.

### DIFF
--- a/docs/spec/runtime/ports-cognition.md
+++ b/docs/spec/runtime/ports-cognition.md
@@ -140,9 +140,14 @@ drift from reality. Reported on `GET …/capabilities` beside `mediaInBuild`,
 "what can this company actually do" — as `cognition`.
 
 The console consumes it in chat (issue #1734): on anything but `configured` the
-transcript carries a banner naming the cause and, where one exists, the remedy,
+pane carries a banner naming the cause and, where one exists, the remedy,
 and every company-side row is marked as a placeholder rather than presented as
-the teammate's own words. `ChatMessage` carries no provenance, so a company-level
+the teammate's own words. The banner is a strip **directly above the composer**
+rather than above the transcript — the caveat and the control it qualifies are
+read together — and below the typing line, so nothing sits between them. It is
+rendered on a read-only channel too, where there is no composer: the sentence is
+about attribution, not about sending, and `#Operator` is precisely a feed of
+company-authored reports rendered under a teammate's name. `ChatMessage` carries no provenance, so a company-level
 state is the only shape that answer has — see `MessageRow`'s `cognition` prop for
 why marking beats suppressing. The same state reaches `ThreadPanel`, because a
 reply read inside a thread is the same false attribution as one read in the
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -2236,9 +2236,21 @@ export function ChatView({
                 the order read correct with nobody typing and wrong with someone
                 typing.
 
-                Suppressed on a read-only channel: nothing can be sent there, so a
-                caveat about what sending produces has nothing left to qualify —
-                and the composer it would sit above is not rendered at all.
+                NOT suppressed on a read-only channel, though the composer it
+                sits above is gone there. It was, briefly, on the reasoning that
+                a caveat about sending has nothing left to qualify — but the
+                sentence is not about sending. It is about attribution: these
+                rows are not the teammate's words. `#Operator` carries
+                company-authored reports rendered under a teammate's name and
+                avatar, `MessageRow` marks each of them from the same cognition
+                state, and suppressing this left the operator with only
+                `EchoPlaceholder`'s chip — the word "Placeholder", with its
+                explanation in a `title` that touch and keyboard users never see
+                (codex review on PR #1984). A read-only feed is where a reader is
+                least able to test the attribution by asking, so it is the last
+                place to drop the only visible statement of it. With no composer
+                there, the strip simply lands under the read-only notice as the
+                bottom strip of the pane.
 
                 All four states below say "the replies in this conversation", not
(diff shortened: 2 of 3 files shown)


graycyrus:
Superseded by **#2159**, which lands the same fix fresh against current \`main\` (\`89b1f818a\`).

This branch is 1,410 commits behind, with a 185-line conflict region in \`ChatView.tsx\` — all of it re-indentation from #2130 / #2134 wrapping the Room route in \`{routeOpen && (\` and extracting the transcript into \`<MessageTimeline>\`. The payload is one token (\`!readOnly\` removed from a single JSX guard), so hand-merging that to land it risked silently dropping someone else's work for no benefit.

#2159 carries the same change, the same docs paragraph, and a rewritten test — with one thing from here deliberately **not** copied: the assertion that the notice is "the bottom strip of the pane" (\`expect(kids.indexOf(strip)).toBe(kids.length - 1)\`). That became false when \`InflightRunBar\` started rendering after the strip, deliberately outside the read-only branch; it still passed here only because the unit harness passes no \`inflightRuns\`. #2159 asserts order against the read-only notice only.

Everything is re-verified on #2159 — three typecheck gates, 4819 unit tests, design tokens, and a browser pass in light and dark against a default-feature host. Nothing from the 2026-09-01 verification here carries forward. Safe to close.

graycyrus:
Superseded by #2159, merged 2026-09-08 as `b2c8f0fe7`.

Rewritten fresh against `main` rather than rebased. The payload here was one token — `!readOnly` out of a single JSX guard — while this branch had drifted 1,410 commits behind with a 185-line conflict region in `ChatView.tsx` that was pure re-indentation from #2130/#2134 wrapping the Room route in `{routeOpen && (` and extracting the transcript into `<MessageTimeline>`. Hand-merging that to land one token risked silently dropping other work for no benefit.

One thing the rewrite deliberately dropped: this branch's second test asserted the notice is "the bottom strip of the pane" (`kids.indexOf(strip) === kids.length - 1`). That stopped being true when `InflightRunBar` started rendering after it — deliberately outside the read-only branch. The assertion still passed only because the unit harness omits `inflightRuns`/`onInflightSteered`, so it pinned a claim that was false in production. #2159 asserts order relative to the read-only notice instead.

Closing. Branch `fix/readonly-feed-echo-notice` is kept at `86257360` if anything here is wanted back.

PR #2000, opened 2026-09-01
fix(e2e): open the console at `/`, not at a view that used to be the default

### LINKED ISSUE #1999: Console E2E (first run): the setup dialog never opens, failing on every frontend PR
## Summary

`test/e2e/company-setup.spec.ts` fails on every PR whose frontend lanes run. Two specs, identical failure, identical timing:

```
✘  1 test/e2e/company-setup.spec.ts:161:1 › first-run setup builds a real team from three answers (20.3s)
✘  2 test/e2e/company-setup.spec.ts:242:1 › skipping setup leaves a way back in (20.2s)

Error: expect(locator).toBeVisible() failed
Error: element(s) not found
  > 176 |   await expect(dialog).toBeVisible({ timeout: 20_000 });
```

The first-run dialog never appears. Both specs time out waiting for `getByTestId("setup-dialog")`, at the 20s ceiling, so nothing is racing — it simply does not open.

## Not one PR's fault

Reproduced on three unrelated open PRs, same two specs, same line numbers, same ~20.3s:

| PR | What it touches |
|---|---|
| #1990 | workflows — `WorkflowCreateDialog`, `workflow-node` |
| #1981 | chat — `ChatView`, `ThreadPanel`, `MessageRow` |
| #1973 | a memory panel row in `SettingsView` |

None touches onboarding. PRs whose base predates the recent console work (#1995, #1991, #1982) still pass the lane, which points at something on `main` rather than at any of these branches.

## Not a testid rename

`data-testid="setup-dialog"` still exists, at `frontend/src/setup/SetupDialog.tsx:647`. The selector is right; the dialog is not opening. The spec's own comment says what is expected:

```ts
// 1. It opens by itself — nobody clicked anything.
```

So the auto-open condition is what regressed, not the markup.

## Where to look

`main` has had heavy console-shell churn recently — `507f9e310 feat(console): four fixed rows, then the section you are in`, `74a3eb917 fix(console): three codex findings — Overview on a phone, rail density, 3rem fit`, and the `#1987` merge, among others. The spec navigates to `/#/overview` and waits, so a change to route handling or the shell's first paint is the likely neighbourhood.

I have not bisected it — this is not my area and the diagnosis needs the console running locally.

## Why it matters beyond the noise

`Console E2E (first run)` is not a required check, so PRs are merging past it. That is the failure mode the lane exists to prevent: once it is red for everyone, a genuine first-run regression lands unnoticed, because nobody can tell their red from the standing red.

CodeGhost21:
Closes #1999.

## What was actually wrong

Nothing in the console. `Console E2E (first run)` has been red on every frontend PR — both specs timing out at the 20s ceiling waiting for `setup-dialog` — because the spec asserts against a default that moved.

First-run setup only offers itself when the operator "just opened the console", and the shell answers that with:

```ts
const deepLinked = useRef(view !== DEFAULT_VIEW || Boolean(sub)).current;
```

`DEFAULT_VIEW` changed from `overview` to `chat` in `a93b509d7` (*"move the channel list into the sidebar as Room"*). `company-setup.spec.ts` navigates to `page.goto("/#/overview")` — written when overview *was* the default — so it began arriving **deep-linked**, and the automatic offer correctly declined to open.

Production was never affected: an operator opening the console lands on an empty hash, which resolves to `DEFAULT_VIEW`, so `deepLinked` is false and setup still offers.

## The fix

Navigate to `/`. An empty hash resolves to `DEFAULT_VIEW` by definition, so it expresses "just opened the console" — the actual condition — and cannot drift when the default moves again. 24 other specs already open this way.

`/#/company` is left alone: that one is a deliberate deep link to the Team page's prompt.

## One of the three was worse than red

The reload check at the end of the first spec:

```ts
// 7. A reload does not re-offer setup: the roster is no longer empty, which is
// the whole reason emptiness is the signal rather than a stored flag.
await page.reload();
await page.goto("/#/overview");
await expect(dialog).toBeHidden();
```

Arriving deep-linked hid the dialog **regardless of the roster**, so this passed for the wrong reason and stopped testing its own comment. It only becomes meaningful with this change.

## Prevention

`DEFAULT_VIEW`'s doc comment already warned that a value changed in one place and not the other is "not a cosmetic drift, it is a company that can never be set up". It named two places. This adds the third — a test that hard-codes a view — since that is the one that actually broke.

## Validation

I could not run Playwright locally (no `frontend/node_modules` on this machine), so this is verified by reasoning and by CI: `Console E2E (first run)` is the lane that has been failing, and it runs on this PR because it touches `frontend/`. If the diagnosis is right, it goes green here.

Worth merging promptly — the lane is not a required check, so people have been merging past it, which is exactly how a genuine first-run regression would land unnoticed behind the standing red.

## Summary by CodeRabbit

* **Tests**
  * Updated company setup tests to open the console at its root, accurately validating automatic setup offers regardless of the configured default view.
  * Improved mock testing for messages containing incomplete or repeated directives, ensuring valid instructions are still recognized and processed.

* **Documentation**
  * Clarified guidance for tests that represent a newly opened console, ensuring they use the root route rather than a specific named view.

### DIFF
--- a/frontend/src/lib/console-routes.ts
+++ b/frontend/src/lib/console-routes.ts
@@ -198,6 +198,14 @@ export const VIEWS: View[] = Object.keys(ROUTABLE) as View[];
  * view against this one. First-run setup only offers itself when the answer is
  * "just opened" — so a default view changed in one place and not the other is
  * not a cosmetic drift, it is a company that can never be set up.
+ *
+ * There is a third place, and it is the one that actually broke (#1999): a test
+ * that opens the console at a *named* view is asserting against whatever this
+ * constant said the day it was written. When this moved from `overview` to
+ * `chat`, `company-setup.spec.ts` started arriving deep-linked and the
+ * first-run specs went red against a console that was working correctly. Tests
+ * that mean "just opened the console" must navigate to `/` — an empty hash
+ * resolves here by definition and cannot drift.
  */
 export const DEFAULT_VIEW: View = "chat";
 
--- a/frontend/test/e2e/company-setup.spec.ts
+++ b/frontend/test/e2e/company-setup.spec.ts
@@ -169,7 +169,13 @@ test("first-run setup builds a real team from three answers", async ({ page, req
     }
   });
 
-  await page.goto("/#/overview");
+  // `/` rather than a named view: `deepLinked` is `view !== DEFAULT_VIEW`, so
+  // naming any view hard-codes today's default and silently stops testing the
+  // automatic offer the day that default moves. It moved — `DEFAULT_VIEW` went
+  // from `overview` to `chat` — and this spec went red for a console that was
+  // working. An empty hash *is* "just opened the console", which is the
+  // condition the offer keys on.
+  await page.goto("/");
 
   // 1. It opens by itself — nobody clicked anything.
   const dialog = page.getByTestId("setup-dialog");
@@ -235,7 +241,11 @@ test("first-run setup builds a real team from three answers", async ({ page, req
   // 7. A reload does not re-offer setup: the roster is no longer empty, which is
   // the whole reason emptiness is the signal rather than a stored flag.
   await page.reload();
-  await page.goto("/#/overview");
+  // Also `/`, and this one is load-bearing: reopening at a named view would
+  // leave the dialog hidden because the arrival was a deep link, which is true
+  // whether or not the roster is staffed. The assertion would pass for the
+  // wrong reason and stop testing what the comment above claims.
+  await page.goto("/");
   await expect(dialog).toBeHidden();
 });
 
@@ -248,7 +258,7 @@ test("skipping setup leaves a way back in", async ({ page, request }) => {
     }
   });
 
-  await page.goto("/#/overview");
+  await page.goto("/");
   await expect(page.getByTestId("setup-dialog")).toBeVisible({ timeout: 20_000 });
 
   await page.getByTestId("setup-skip").click();
--- a/frontend/test/e2e/mock-brain.mjs
+++ b/frontend/test/e2e/mock-brain.mjs
@@ -215,6 +215,48 @@ const SPAWN_DIRECTIVE = "SPAWNONE";
 const PLAN_DIRECTIVE = "__MOCK_PLAN__";
 
 /**
+ * The host's own briefing blocks, appended to an operator message before it
+ * reaches a model.
+ *
+ * These matter here because two of them quote **other messages verbatim**.
+ * `THREAD_INDEX_ANNOTATION` lists the opening words of the channel's other
+ * threads, truncated — so once a spec has opened a thread with a directive in
+ * it, every later message in that channel carries a chopped-off copy of that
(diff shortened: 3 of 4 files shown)


CodeGhost21:
Carrying the sibling E2E fix so this PR's own CI is meaningful.

The two standing E2E regressions are fixed in separate PRs — #2000 (`Console E2E (first run)`: `DEFAULT_VIEW` moved from `overview` to `chat`, so the spec's hard-coded `/#/overview` began arriving deep-linked) and #2003 (`Console E2E (live brain)`: the host's `## Relevant prior work` digest quotes earlier messages truncated, so a chopped-off `__MOCK_PLAN__` sat ahead of the live one in the same message). Each PR was green on the lane it fixes and red on the lane the other fixes, which made every run here ambiguous.

Both branches are now merged into this one, so a red lane means this PR. They collapse to nothing once the two land.

M3gA-Mind:
Reviewed together with #2003. The `findPlan` change itself is right: occurrences are collected and scanned newest-first, and the loop falls through to older messages only when *nothing* in the current one parses — so it handles a decoy on either side of the live directive, and `at` is always bound by the time `text.slice(at)` runs. Carrying the sibling fix to get an unambiguous CI run was the right call too.

Two things before this lands.

## This PR and #2003 are byte-identical

I diffed both full patches: zero difference. The cross-merges (`origin/fix/2002-orchestration-review` into this branch, and the mirror into #2003) mean each now contains the whole union. #1973 carries the same four files as well.

So whichever of the two lands first fixes **both** bugs while auto-closing only **one** issue, and leaves the other as an empty PR still carrying a `Closes` keyword. Land one, then close the sibling and its issue by hand rather than merging a no-op — and rebase #1973 to drop the carried E2E commits once either is in.

## `specWords` is not exercised by anything

`HOST_ANNOTATIONS` and `specWords` are new, and no test reaches them. The fixture in `ignores a truncated directive the host echoed back` composes `## Relevant prior work` + `## Task`, and `## Relevant prior work` is not one of the five markers — so `specWords` returns the string unchanged and the test passes purely on the newest-occurrence scan. Revert `specWords(textOf(messages[i]))` to `textOf(messages[i])` and nothing goes red.

That matters more than a nit here, because `specWords` targets a *different* decoy source than the one the test reproduces: the thread-index annotation appends a truncated directive **after** the operator's words, where the occurrence scan already recovers. The case only `specWords` can fix is a host annotation whose quoted text parses as valid JSON — and that is the case with no coverage. One fixture ending in `\n\n[Other conversations in this channel` followed by a parseable `__MOCK_PLAN__` closes it.

## Minor: the two finders now disagree

`findPlan` strips host annotations; `findDirective` still reads raw `textOf`, and its `SPAWNONE` branch is a plain `indexOf`. An echoed truncated `SPAWNONE` inside a thread-index annotation still matches there, and since the id keys on the rest of the line it reads as a *new* directive — the duplicate-card class the comment right above it describes. Pre-existing rather than introduced here, so not blocking, but the asymmetry is now conspicuous enough to be worth a line either way.

PR #2003, opened 2026-09-01
fix(e2e): find the live plan, not the truncated copy quoted above it

### LINKED ISSUE #2002: Console E2E (live brain): an approved card never reaches done, since #1972
## Summary

`Console E2E (live brain)` fails on every PR based after #1972. `orchestration-simulation.spec.ts:97` times out waiting for an approved card to reach `done`:

```
Error: card "sim gather the sources …" was never accepted
Expected: "done"
Received: "working"
- Timeout 180000ms exceeded while waiting on the predicate
```

## Bisect

Clean bracket across merged PRs, by whether their base predates #1972:

| PR | live brain |
|---|---|
| #1984, #1983, #1980, #1977, #1975, #1969 | SUCCESS |
| **#1972** | **FAILURE** — merged anyway |
| #1987, #1995, and every open PR since (#1990, #1997, #1998, #2000) | FAILURE |

#1972 (`feat/1890-thread-as-context-boundary`) went in with this lane already red. `Console E2E (live brain)` is not a required check, so nothing stopped it.

## What is *not* broken

The backend transition is fine. Every Rust test covering it passes on `main` (`cargo test --features openhuman --lib review`), including:

- `an_approving_review_finishes_the_card_and_revise_sends_it_back`
- `every_finished_card_reaches_the_column_review_consumes`
- `a_delegated_card_now_stops_in_review_like_every_other_card`
- `review_task_tool_rejects_an_unknown_verdict_rather_than_approving`

The spec also gets *past* the earlier stages: step 4 (`stage == "in_review"`) and step 5 (both settled cards mark their thread) both pass. `in_review` is a stage inside the `working` column, so "received `working`" is consistent with a card that reached review and was never approved.

So the card arrives correctly and the approve never lands.

## Where it likely is

`review_task` is never *called*, rather than called and failing. The mock brain serves a `__MOCK_PLAN__` step only when the request's own `tools[]` carry every name the step uses — its own comment:

> Served only when the request's own `tools` carry every name the step uses, which is what stops a teammate reading the operator's message second-hand from honouring the orchestrator's plan.

#1972 changed how a turn is dispatched: `RunTurn::run_turn` gained a `chat: ChatTarget<'_>` parameter, a `TURN_CONVERSATION` task-local was added, and `src/harness/thread_tools.rs` (473 lines) joined the belt. If the closing operator message now runs as a differently-scoped turn, the orchestrator's `tools[]` for it may no longer carry `review_task`, and the plan would be silently declined — which looks exactly like this.

That is a hypothesis, not a diagnosis. `MOCK_BRAIN_DEBUG=1` dumps the `tools=[…]` the host actually sent, which should settle it in one run.

Not attempting a fix: this is inside #1890's in-flight turn-dispatch work, and guessing at it is more likely to be wrong than useful. cc @sanil-23.

## Worth fixing separately

This is the second standing E2E regression to reach `main` this week through a non-required lane — #1999 was the other, where the first-run setup dialog broke for every frontend PR. Once a lane is red for everyone, nobody can distinguish their own red from the standing red, which is how the next real regression lands unnoticed. Either these lanes should gate, or a red one should page someone.

CodeGhost21:
Closes #2002.

## What was wrong

`Console E2E (live brain)` has failed on every PR based after #1972. `orchestration-simulation.spec.ts` drives a goal to completion with `__MOCK_PLAN__` directives; cards reached `in_review` and the closing `review_task` never ran, so it timed out waiting for `done`.

The mock brain had been saying so in its own stderr the whole time:

```
[mock brain] __MOCK_PLAN__ found but its JSON payload did not parse
```

The host composes an operator turn as a `## Relevant prior work` digest of earlier tasks, then the operator's words under `## Task`. **The digest quotes those earlier messages truncated.** So once the simulation opens its thread with a directive in the first message, every later turn carries a chopped-off copy of it — *earlier in the same string* than the live one:

```
## Relevant prior work
- Task: Ship a short market digest… __MOCK_PLAN__ [[{"name":"spawn_task",…{"title":"sim gather the sou
## Task
Both pieces are back — take a look and close them out. __MOCK_PLAN__ [[{"name":"review_task",…
```

`findPlan` took `indexOf`, met the decoy, and gave up on the whole thread.

**The host is behaving correctly.** Quoting prior work truncated is the point of that digest. What broke is the harness's assumption that the first directive in a message is the one the spec wrote.

## The fix

Collect every occurrence of the directive in a message and try them newest-first, so the operator's own directive wins over anything quoted above it. Also: on a parse failure keep scanning rather than abandoning the thread, and ignore the host's appended briefing blocks (`operator_words` does the same thing host-side, for the same reason).

## Verified by running it, not by reasoning

Worth saying plainly, because my first attempt was wrong. I initially fixed only the "keep scanning older messages" half — which is useless when the live directive is further down the *same* message. Running the spec is what caught it:

```
[mock brain] plan step 0: spawn_task + spawn_task      ← opening plan, always worked
[mock brain] plan step 0: review_task + review_task    ← the closing plan, never served before
1 passed (18.8s)
```

The unit test reproduces the real composition — decoy before the live directive in one message — and fails if the search stops at the first occurrence. I checked it fails without the fix; an earlier version of that test passed either way, which is no test at all.

## Note for #1999's neighbour

This is the second standing E2E regression this week to reach `main` through a lane that does not gate — #1972 merged with this lane already red. Once a lane is red for everyone, nobody can tell their own red from the standing red.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automated task interpretation when messages contain repeated or incomplete directives.
  - The system now ignores truncated or echoed instructions and continues searching for the most recent valid plan.
  - Added clearer diagnostics when a directive cannot be parsed.
  - Prevented prior-work summaries and host-generated annotations from interfering with live task instructions.

- **Tests**
  - Added coverage for cases where an incomplete directive appears before a valid task plan.

### DIFF
--- a/frontend/src/lib/console-routes.ts
+++ b/frontend/src/lib/console-routes.ts
@@ -198,6 +198,14 @@ export const VIEWS: View[] = Object.keys(ROUTABLE) as View[];
  * view against this one. First-run setup only offers itself when the answer is
  * "just opened" — so a default view changed in one place and not the other is
  * not a cosmetic drift, it is a company that can never be set up.
+ *
+ * There is a third place, and it is the one that actually broke (#1999): a test
+ * that opens the console at a *named* view is asserting against whatever this
+ * constant said the day it was written. When this moved from `overview` to
+ * `chat`, `company-setup.spec.ts` started arriving deep-linked and the
+ * first-run specs went red against a console that was working correctly. Tests
+ * that mean "just opened the console" must navigate to `/` — an empty hash
+ * resolves here by definition and cannot drift.
  */
 export const DEFAULT_VIEW: View = "chat";
 
--- a/frontend/test/e2e/company-setup.spec.ts
+++ b/frontend/test/e2e/company-setup.spec.ts
@@ -169,7 +169,13 @@ test("first-run setup builds a real team from three answers", async ({ page, req
     }
   });
 
-  await page.goto("/#/overview");
+  // `/` rather than a named view: `deepLinked` is `view !== DEFAULT_VIEW`, so
+  // naming any view hard-codes today's default and silently stops testing the
+  // automatic offer the day that default moves. It moved — `DEFAULT_VIEW` went
+  // from `overview` to `chat` — and this spec went red for a console that was
+  // working. An empty hash *is* "just opened the console", which is the
+  // condition the offer keys on.
+  await page.goto("/");
 
   // 1. It opens by itself — nobody clicked anything.
   const dialog = page.getByTestId("setup-dialog");
@@ -235,7 +241,11 @@ test("first-run setup builds a real team from three answers", async ({ page, req
   // 7. A reload does not re-offer setup: the roster is no longer empty, which is
   // the whole reason emptiness is the signal rather than a stored flag.
   await page.reload();
-  await page.goto("/#/overview");
+  // Also `/`, and this one is load-bearing: reopening at a named view would
+  // leave the dialog hidden because the arrival was a deep link, which is true
+  // whether or not the roster is staffed. The assertion would pass for the
+  // wrong reason and stop testing what the comment above claims.
+  await page.goto("/");
   await expect(dialog).toBeHidden();
 });
 
@@ -248,7 +258,7 @@ test("skipping setup leaves a way back in", async ({ page, request }) => {
     }
   });
 
-  await page.goto("/#/overview");
+  await page.goto("/");
   await expect(page.getByTestId("setup-dialog")).toBeVisible({ timeout: 20_000 });
 
   await page.getByTestId("setup-skip").click();
--- a/frontend/test/e2e/mock-brain.mjs
+++ b/frontend/test/e2e/mock-brain.mjs
@@ -215,6 +215,48 @@ const SPAWN_DIRECTIVE = "SPAWNONE";
 const PLAN_DIRECTIVE = "__MOCK_PLAN__";
 
 /**
+ * The host's own briefing blocks, appended to an operator message before it
+ * reaches a model.
+ *
+ * These matter here because two of them quote **other messages verbatim**.
+ * `THREAD_INDEX_ANNOTATION` lists the opening words of the channel's other
+ * threads, truncated — so once a spec has opened a thread with a directive in
+ * it, every later message in that channel carries a chopped-off copy of that
+ * directive. A directive cut mid-payload is not a directive; it is a decoy that
+ * parses as nothing.
+ *
+ * Cutting at the earliest of these is what the host itself does to recover the
+ * operator's own words (`operator_words` in `src/runtime/delegation.rs`), and
+ * for the same reason: everything past the first marker was written by the
+ * host, not by the spec.
+ *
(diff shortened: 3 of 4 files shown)


CodeGhost21:
Carrying the sibling E2E fix so this PR's own CI is meaningful.

The two standing E2E regressions are fixed in separate PRs — #2000 (`Console E2E (first run)`: `DEFAULT_VIEW` moved from `overview` to `chat`, so the spec's hard-coded `/#/overview` began arriving deep-linked) and #2003 (`Console E2E (live brain)`: the host's `## Relevant prior work` digest quotes earlier messages truncated, so a chopped-off `__MOCK_PLAN__` sat ahead of the live one in the same message). Each PR was green on the lane it fixes and red on the lane the other fixes, which made every run here ambiguous.

Both branches are now merged into this one, so a red lane means this PR. They collapse to nothing once the two land.

M3gA-Mind:
Reviewed together with #2000. The `findPlan` change itself is right: occurrences are collected and scanned newest-first, and the loop falls through to older messages only when *nothing* in the current one parses — so it handles a decoy on either side of the live directive, and `at` is always bound by the time `text.slice(at)` runs. Carrying the sibling fix to get an unambiguous CI run was the right call too.

Two things before this lands.

## This PR and #2000 are byte-identical

I diffed both full patches: zero difference. The cross-merges (`origin/fix/1999-setup-dialog` into this branch, and the mirror into #2000) mean each now contains the whole union. #1973 carries the same four files as well.

So whichever of the two lands first fixes **both** bugs while auto-closing only **one** issue, and leaves the other as an empty PR still carrying a `Closes` keyword. Land one, then close the sibling and its issue by hand rather than merging a no-op — and rebase #1973 to drop the carried E2E commits once either is in.

## `specWords` is not exercised by anything

`HOST_ANNOTATIONS` and `specWords` are new, and no test reaches them. The fixture in `ignores a truncated directive the host echoed back` composes `## Relevant prior work` + `## Task`, and `## Relevant prior work` is not one of the five markers — so `specWords` returns the string unchanged and the test passes purely on the newest-occurrence scan. Revert `specWords(textOf(messages[i]))` to `textOf(messages[i])` and nothing goes red.

That matters more than a nit here, because `specWords` targets a *different* decoy source than the one the test reproduces: the thread-index annotation appends a truncated directive **after** the operator's words, where the occurrence scan already recovers. The case only `specWords` can fix is a host annotation whose quoted text parses as valid JSON — and that is the case with no coverage. One fixture ending in `\n\n[Other conversations in this channel` followed by a parseable `__MOCK_PLAN__` closes it.

## Minor: the two finders now disagree

`findPlan` strips host annotations; `findDirective` still reads raw `textOf`, and its `SPAWNONE` branch is a plain `indexOf`. An echoed truncated `SPAWNONE` inside a thread-index annotation still matches there, and since the id keys on the rest of the line it reads as a *new* directive — the duplicate-card class the comment right above it describes. Pre-existing rather than introduced here, so not blocking, but the asymmetry is now conspicuous enough to be worth a line either way.

PR #2043, opened 2026-09-03
docs: correct the Cortex record — the driver was written, and works

CodeGhost21:
## Summary

`docs/spec/runtime/memory-engine-cortex.md` says a conformant driver **cannot** be
written against CortexDB v0.9.8, and that Phase 1 is blocked upstream. Both were
true when it was written and are wrong now — wrong in the direction that costs
somebody a week, because a reader would conclude the work cannot start when it is
finished and merged as tinyhumansai/tinymemory#128.

The upsert gap is real; it is not a wall. Appending every write under a fresh
idempotency key carrying the logical key in the payload, then folding the log to
newest-per-key on read, reconstructs replacement on the read side and passes the
full conformance suite against a live engine. This PR replaces the impossibility
claim with what the workaround actually costs — a scan per keyed read, seconds
per write, and superseded values left in the ranking corpus — because the cost is
what should decide this, and it points at the same recommendation from firmer
ground.

Documentation only. No code, no behaviour, no new architecture — mostly
corrections and deletions.

## Related issue

#1936. Does not close it: the remaining deliverable is a topology decision, not a
document.

## API or behavior changes

None.

## What changed

- **Finding 5 and its section.** "A conformant driver cannot be written" → the
  gap and its price. The measured table of failed direct mappings stays; it is
  still why the workaround is necessary.
- **Phase 1: blocked → done, and unbound.** With the point that matters for
  review: `SUPPORTED_REMOTE_DRIVERS` and `remote_provider()` are untouched, so
  nothing here can select the driver. Registering it is a decision, not a task.
- **The `deleted.events: 2` alarm was ours.** This record called it
  "disqualifying on its own until explained". The explanation is that the request
  named `selector.event_ids`, which is not a schema field, alongside
  `confirm_all: true` — an unrecognised selector field deserialises as an *empty*
  selector, meaning the whole scope, which `confirm_all` then authorises. With
  `selector.memory_ids` deletion is exact, and two engine interlocks refuse that
  mistake on its own. I would rather this correction be in the record than have
  someone inherit a phantom defect.
- **Two open questions answered** instead of left hanging. Ranked recall is
  vector search over the event log — confirmed on a fresh scope queried before
  any derived layer had built — so on retrieval alone Cortex does not beat the
  drivers we already ship.
- **New: the five engine behaviours that only surfaced against a running
  instance.** Each produced a driver that passed every offline test and was wrong
  in production, because the conformance double and the driver were both written
  from the same documentation and agreed with each other rather than with the
  service. They are properties of the engine, so whoever revisits this meets them
  again.
- **Infrastructure stays incomplete and now says so.** #1936 asked for HA,
  backup/restore and an upgrade path. They describe operating a deployment this
  record recommends against, so they wait on the Phase 0 topology decision. An
  explicit gap beats a plan written to fill a checkbox.

## Validation

- [x] `bash scripts/ci/assert-md-line-cap.sh` — passes (the file lands at 500,
      which is why several passages are tighter than a first draft; the cap is
      the one I added in #1975, so it would have been poor form to trip it)
- [x] Every stale claim confirmed gone by grep: "Nothing here is implemented",
      "cannot be written against v0.9.8", "Currently blocked upstream", "blocked
      on their answer", "This blocks Phase 1 outright", "disqualifying on its
      own"
- [x] `memory-engine.md`'s link to this file still resolves

No Rust touched, so the Rust lanes are unaffected.

## Tests

None — documentation. The claims it now makes are the ones tinymemory#128's test
suite and live lane already assert; this PR stops the spec contradicting them.

## Documentation

This *is* the documentation change.

## Checklist

- [x] The change is focused on one logical change
- [x] No new `#[allow(...)]`, `#[ignore]`, or relaxed lints
- [x] No secrets, tokens, or `.env` contents in the diff or the description

## Summary by CodeRabbit

- **Documentation**
  - Updated runtime memory documentation with the Cortex driver’s live conformance status and unregistered state.
  - Clarified that the reachable token minter does not enforce scope isolation.
  - Added driver guidance covering append-and-fold reconstruction, scan-based reads, delayed write visibility, and stale recalled values.
  - Documented exact deletion behavior and resolved the prior deletion alarm.
  - Marked Phase 1 complete and refreshed infrastructure details, acceptance criteria, and open questions.

### DIFF
--- a/docs/spec/runtime/memory-engine-cortex-driver.md
+++ b/docs/spec/runtime/memory-engine-cortex-driver.md
@@ -60,9 +60,19 @@ wraps records in a JSON envelope inside `content` (`Bound::put` calls
 `encode(record)`), so that is the shape it uses anyway — but a scan-per-read is
 slow and still non-atomic.
 
-Both rest on a `forget` that reported `deleted.events: 2` for a selector naming
-one event id. That is a lot of correctness risk to absorb for something an
-upstream `on_conflict: replace` would remove entirely.
+Only the first rests on `forget`, and that matters, because `forget` is where our
+worst scare came from: a call reporting `deleted.events: 2` for a selector naming
+one event id. That turned out to be our own malformed request rather than an
+engine defect — the record's Phase 1 acceptance notes set out exactly which layer
+refuses what — but an emulation that deletes the previous event on every
+overwrite puts that call on the write path of every store, which is a lot of
+blast radius to accept for something an upstream `on_conflict: replace` would
+remove entirely.
+
+The envelope-and-fold path does not delete anything. It appends the new event and
+lets the fold prefer the newest per key, so the previous version stays on disk and
+is simply not returned. That is the path the shipped driver takes, and it is why
+its `forget` is confined to `delete`, where a caller asked for removal.
 
 The reproduction is recorded at
 [cortexdb-releases#3](https://github.com/cortexdbai/cortexdb-releases/issues/3)
--- a/docs/spec/runtime/memory-engine-cortex-driver.md
+++ b/docs/spec/runtime/memory-engine-cortex-driver.md
@@ -1,0 +1,116 @@
+# Writing a driver against CortexDB v0.9.8
+
+Companion to [`memory-engine-cortex.md`](memory-engine-cortex.md), which asks
+whether we should host Cortex at all. This one assumes that question is still
+open and answers a narrower one: what a `MemoryProvider` driver has to do to
+speak to CortexDB v0.9.8, and what that costs at every call.
+
+Split out of the design record when it outgrew the 500-line cap. The record kept
+the decision; this kept the mechanics behind it.
+
+The driver these notes describe is merged as
+[tinymemory#128](https://github.com/tinyhumansai/tinymemory/pull/128) and is
+deliberately unregistered — see the record's phased plan.
+
+## The upsert gap, and what it costs to work around
+
+This is the finding that decides the phasing, and it is not about capability
+breadth — it is that the two data models disagree.
+
+`Memory::store` is an **idempotent upsert keyed by `(namespace, key)`**: storing
+twice at one key replaces the previous content "rather than erroring or creating
+a duplicate". The conformance suite asserts it directly
+(`assert_upsert_replaces_rather_than_duplicates`) — write `first`, write
+`second`, expect one row holding `second`. `memory-engine.md` makes that suite
+the thing that retired the unproven-remote flag, so passing it is not optional.
+
+CortexDB is an append-only event log. Measured against the deployment:
+
+| Attempt | Result |
+|---|---|
+| Same key, different content | `409 IDEMPOTENCY_CONFLICT` — "idempotency key reused with a different body" |
+| Same key, identical content | `202`, `replayed_from_idempotency: true` |
+| Update route | none — `/v1/events` and `/v1/events/{id}` are GET-only, `/v1/experience` POST-only |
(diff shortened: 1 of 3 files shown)


CodeGhost21:
All four points were right, and two of them changed the argument rather than the wording. Fixed in e7e00ff89.

**"All three costs disappear with an upsert" was wrong** (@chatgpt-codex-connector, @coderabbitai both). Only the third does. Keyed reads still scan — that needs a metadata filter on `/v1/events`. Writes still wait — that needs a readiness signal the engine doesn't expose. Those are three separate upstream asks, and collapsing them into one overstated what the vendor conversation would buy us, in a document whose whole job is informing a go/no-go. Also flagged that `on_conflict: "replace"` is our proposed ask, not a documented CortexDB option.

**Superseded values reaching `recall` is not merely a cost.** I checked the claim rather than taking it: `assert_upsert_replaces_rather_than_duplicates` reads back through `list` and never calls `recall`. So the suite does not decide this axis in either direction, and citing a green suite as proof of conformance here was wrong. It is now recorded as an open contract question — whether that is a violation or permitted backend-defined behaviour — with the coverage gap named explicitly.

**The deletion interlocks needed their enforcement boundary named.** Fair, and the honest version is less comfortable than what I wrote. The scope wipe was *observed, not simulated*, and nothing blocked it. The engine refuses an empty selector alone (`422`) and refuses selector + `confirm_all` (`400`), but it cannot refuse the combination we actually sent, because "empty selector + confirm_all" is a valid whole-scope wipe request. The only thing between a typo and a destroyed tenant is driver-side. There is now a table naming which layer stops which request.

**The selection recipe is a boot refusal today.** Verified: `SUPPORTED_REMOTE_DRIVERS` holds `supermemory`, `mem0`, `cognee`, and `remote_provider()` rejects everything else. Marked as the target configuration rather than a usable one.

## One structural change

The record hit the 500-line cap three times in this session, and I had been buying room by shaving prose, which was making it worse. These fixes needed real content, so I split instead: `memory-engine-cortex-driver.md` now holds the upsert gap, what the workaround costs per call, and the engine behaviours that only appear against a running instance. The record keeps the decision and links to it; `memory-engine.md` links to both.

That is the split CLAUDE.md prescribes for this situation, and it should have happened on merit rather than because a limit forced it — the decision and the mechanics have different readers.

Validation: `scripts/ci/assert-md-line-cap.sh` passes (431 and 117 lines), no dangling cross-references, and the one "see the section below" that the split broke now points at the companion.

CodeGhost21:
Valid, and it caught a sentence that contradicted the paragraph three lines below it. Fixed in 224be24bb.

"Both rest on a `forget`" was wrong. The **external-index** emulation forgets the prior event on every overwrite, so it does. The **envelope-and-fold** path appends the new event and lets the fold prefer the newest per key — the previous version stays on disk and is simply not returned. Nothing is deleted.

That distinction is the entire argument for the path we actually took, so flattening the two together undersold it. Verified against the driver rather than reasoned: `cortex.rs` has exactly one `/v1/forget` call site, inside `delete()`, and `upsert()` has none.

I also fixed the framing of the scare while I was in there, since it had travelled from an older draft. The `deleted.events: 2` call was our own malformed request, not an engine defect — the record's Phase 1 notes set out which layer refuses what. It is still worth mentioning, but for a different reason than the text implied: an emulation that deletes on every overwrite would put that call on the write path of *every store*, which is the blast radius that makes the external-index option unattractive.

One thing this surfaced that the review did not ask about: the record said "the `deleted.events: 2` alarm **above**", and the passage it pointed at had moved into the driver notes when I split the file. That back-reference is now a link. Worth me re-reading the rest of the split for the same class of breakage rather than waiting to be told.

CI green, line cap passes on both files.

PR #2065, opened 2026-09-03
Make `cortex` a selectable memory engine (#1936)

CodeGhost21:
## Summary

Registers the CortexDB driver from tinyhumansai/tinymemory#128 so this host can
bind it. **Selectable, not selected** — `OPENCOMPANY_MEMORY` defaults exactly as
before and nothing changes for an instance that does not ask for it.

Choosing it is a separate decision, and the design record argues against it
(`docs/spec/runtime/memory-engine-cortex.md`): the derived fact and belief tier
produces nothing, so what Cortex offers over the drivers we already ship is
ordinary vector search, at a scan per keyed read and seconds per write. This PR
exists so that decision is a config change rather than a project.

## Related issue

#1936. Does not close it.

## API or behavior changes

Additive. `cortex` becomes a valid `OPENCOMPANY_MEMORY_DRIVER` value and appears
in the console's engine catalog. No existing engine changes.

## Validation

- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo build --all-targets` — clean
- [x] `cargo test --lib` — 4566 passed, 0 failed

Plus the live lane against a real CortexDB v0.9.8 on staging: both
`the_live_cortex_service_upholds_the_provider_contract` and
`..._upholds_the_host_ports` pass, in **295 seconds**. That number is a finding
rather than an aside — the engine indexes asynchronously, so the driver waits
for its own write to become readable and a caller feels every one of those
seconds. It is the same cost the design record quantifies.

## What a reviewer should actually look at

**The pin bump.** `vendor/openhuman` 61d25fe2 → b7f4e949, which repins tinymemory
v1.13.7 → v1.13.8. The `Cargo.lock` delta is exactly those two versions and
nothing else. OpenHuman had already repinned tinymemory itself
(YellowSnnowmann's #6013), so this carries that through rather than forcing it.

**No tinymemory registry change was needed, and that is deliberate.** `cortex` is
not in `DriverRegistry::builtin()`'s reserved table, but `admit` accepts an
unreserved id when the host supplies the class — and this host already declares
every remote driver `External` with `TRUSTED`. So the class stays host-decided
rather than self-reported, which is the property that table exists to protect. If
you would rather it were reserved upstream, say so and I will send that instead;
it costs a tinymemory release and two pin bumps.

**Three coupling tests caught omissions I had made**, which is worth noting
because it is the argument for keeping them: `catalog_matches_driver_registry`,
`every_offered_engine_resolves_to_a_runtime_selection` and
`engine_ids_round_trip_through_a_selection` all failed until the console catalog
entry and the `split_engine` mapping existed. Registering a driver in one place
is not enough here, and the suite says so.

## Tests

The standard per-driver pair against a double, plus the `live_hosted` pair.

The double is deliberately thinner than the one upstream in `tinymemory-remote`,
and its docs say why: duplicating that fidelity here would be two copies to keep
honest. What this one must get right is the two behaviours the *binding* depends
on — that a write becomes readable (the driver waits for its own write, so a
double that never settles would hang the wait rather than fail it), and that
`/v1/recall` renders content with the speaker prefixed while `/v1/events` returns
it as stored. A driver parsing only the stored form lists correctly and searches
to nothing, which is the failure this catches.

## Documentation

None needed — `memory-engine-cortex.md` and its driver companion landed in #2043
and already describe this. The catalog entry's description points at them.

## Checklist

- [x] The change is focused on one logical change
- [x] No new `#[allow(...)]`, `#[ignore]`, or relaxed lints
- [x] No secrets, tokens, or `.env` contents in the diff or the description

## Summary by CodeRabbit

* **New Features**
  * Added CortexDB as a selectable memory engine.
  * Added support for connecting to CortexDB through managed or self-hosted deployments.
  * CortexDB supports durable memory, ranked recall, event listing, forgetting, scope listing, and health checks.

* **Documentation**
  * Updated memory engine documentation with CortexDB configuration and selection guidance.

* **Tests**
  * Added coverage for CortexDB provider behavior, offline operation, and hosted deployments.


CodeGhost21:
Valid, and a good catch — fixed in e52221097.

This PR registers `cortex` and points the console catalog at the design record, while that record still told an operator the driver was unregistered and the selection recipe was a boot refusal. Anyone following the pointer would have landed on a document saying the supported configuration will fail.

Four places, all corrected:

- the record's header and its Phase 1 entry, both of which said "deliberately unregistered";
- the operational note, which described the recipe as not working. It now says how to select it, and why registering needed no change to tinymemory's reserved table — `admit` takes an unreserved id when the host declares the class, and this host declares every remote driver `External` with `TRUSTED`;
- `memory-engine.md`, which said nothing was implemented and left `cortex` out of the valid driver list;
- the driver companion, which repeated the same claim.

The recommendation is untouched throughout: selectable is not selected, the record still argues against choosing it, and the default is unchanged. `scripts/ci/assert-md-line-cap.sh` passes.

Worth noting the shape of this one, since it is the second time in two PRs: I changed behaviour in code and left prose asserting the old behaviour. The line-cap script is the only Markdown check that runs, so nothing catches a doc that contradicts the code it documents — it took a reviewer both times.

CodeGhost21:
Both critique findings are false positives. Verified rather than asserted, and no code change — flagging them here so the lane can clear.

**"Reference to nonexistent feature gate `tinymemory`" (high).** `tinymemory` is a local binding, not a feature name used bare. `catalog()` opens with it at line 235:

```rust
let tinymemory = cfg!(feature = "tinymemory");
```

and every remote entry uses it identically — lines 256, 267, 278, **291** (the new one) and 302. The suggested `available: cfg!(feature = "tinymemory")` would be correct in isolation but inconsistent with the other four, and the existing comment above that binding explains why it is resolved once: `cfg!` rather than `#[cfg]` blocks, so the catalog is one list in one order in every build.

The claim that it "will cause a compile error" is also directly falsifiable: `cargo check`, `cargo clippy --all-targets -- -D warnings` and `cargo test --lib` (4566 passing) all ran clean on this commit. The finding appears to come from reading the diff hunk without the enclosing function.

**"Reference to a documentation file that is not added in this PR" (medium).** `docs/spec/runtime/memory-engine-cortex.md` exists on `main` — it merged in #2043 — and this PR *modifies* it rather than adding it, along with `memory-engine-cortex-driver.md` and `memory-engine.md`. The link resolves. The finding treats "absent from the added-files list" as "does not exist".

Happy to be shown wrong on either, but I would rather not change working, consistent code to satisfy a reading of a partial diff.

CodeGhost21:
`tinysweeper/critique` is red with 4 findings. All four are false, and the bot
says why in its own header: _"The code index is behind this pull request
(indexed at `31f7f6e65dcd`)."_ That SHA is not reachable in `opencompany`,
`openhuman`, or `vendor/openhuman` — the index is pointing at something that
does not exist here, not merely at an older commit.

The decisive counter-evidence is CI itself. Findings 1 and 2 claim
`memory_engine.rs` **will not compile**. `Rust (openhuman, tinymemory)` — the
lane that compiles this code — passed in 37m50s, including the steps
`Test the memory engine selection surface`,
`Test the memory engine union the tenant ships`, and
`Test the memory engine union selection`.

Point by point:

**1 + 2 — "nonexistent feature gate `tinymemory`" (`memory_engine.rs:291`, `:292`).**
`tinymemory` is a local binding and `feature` a local closure, declared 55 lines
above the flagged code:

```rust
// src/server/ops/memory_engine.rs:235
let tinymemory = cfg!(feature = "tinymemory");
let feature = |on: bool, name: &str| {
    (!on).then(|| format!("this build was compiled without the `{name}` feature"))
};
```

All six catalog entries use both identically; only the new `cortex` entry was
flagged. The suggested fix (`available: cfg!(feature = "tinymemory")`) restates
what the binding already holds and would make the new entry inconsistent with
the other five.

**3 — "cortex not in the driver registry" (`:324`).**
It is. `src/store/memory/driver.rs:182` is now `[&str; 4]`:

```rust
pub const SUPPORTED_REMOTE_DRIVERS: [&str; 4] = [
    SUPERMEMORY_DRIVER_ID,
    MEM0_DRIVER_ID,
    COGNEE_DRIVER_ID,
    tinymemory_remote::CORTEX_DRIVER_ID,
];
```

with the dispatch arm at `driver.rs:165`. The finding cites the 3-element
version from `test.rs`, which is the stale snapshot. The
`catalog_matches_driver_registry` test (`ops/memory_engine/test.rs:161`) exists
precisely to keep the catalog and the registry in sync, and it passes.

**4 — "doc file not added in this PR" (`:287`).**
`docs/spec/runtime/memory-engine-cortex.md` is in the diff, alongside
`memory-engine-cortex-driver.md` and an update to `memory-engine.md`.

Every other check on `3a87c2bd1` is green: `Rust`,
`Rust (openhuman, tinymemory)`, `Rust (mongodb)`, `Rust (mail)`, `Desktop`,
`Gated host binary`, all three `Console E2E` lanes, `Actions pinned`,
`Changed areas`, `Markdown line cap`, CodeRabbit, and
`tinysweeper/{review,tests,security,description}`.

`POST /check-runs/100786311748/rerequest` returns 404, so this cannot be re-run
from our side — it needs a maintainer dismissal.

YellowSnnowmann (approved):
Reviewed the three surfaces the PR body calls out: the pin bump (Cargo.lock delta is only tinymemory 1.13.7→1.13.8 and openhuman 0.63.19→0.63.21, nothing unexpected), the catalog/split_engine/driver registration (all three coupling tests satisfied, DRIVER_ID consts used throughout, no string literals), and the conformance double (cx_recall/cx_events asymmetry correctly models the speaker-prefix distinction; cx_forget uses memory_ids and never confirm_all; write visibility is immediate in the double so the driver's wait loop terminates rather than hangs).

One non-blocking note: the console catalog description warns about write latency but doesn't mention that OPENCOMPANY_MEMORY=cortex (the intuitive env var) is still a hard boot refusal — OPENCOMPANY_MEMORY=remote + OPENCOMPANY_MEMORY_DRIVER=cortex is the correct selection. Worth one line in the description field so operators don't hit a cryptic boot failure before they find the spec doc. Happy to land as-is and follow up, or you can fold it in here.

Everything else is correct. Additive, defaults untouched, live tests opt-in.

Issue #2072, opened 2026-09-04
Decide the hosted-Cortex topology, or decline adoption (Phase 0 of #1936)


CodeGhost21:
## The decision

Pick one, and write down why:

- **A — Ratify instance-per-tenant.** Adopt hosted Cortex at the strong
  isolation tier, one CortexDB instance per tenant. Unblocks Phase 2
  (provisioning) and Phase 3 (migration).
- **B — Accept namespace-only separation, explicitly.** Adopt at the weak tier,
  with the tradeoff recorded the way `storage.md` records it for shared-single-DB
  mode.
- **C — Decline adoption.** `cortex` stays registered and selectable, nobody
  selects it, and Phases 2–4 are dropped.

The investigation in #1936 is complete and recommends **C**. This issue exists
so that recommendation gets accepted or overruled by someone who owns memory
strategy, rather than sitting in a doc.

## Why this gates everything

`docs/spec/runtime/memory-engine-cortex.md` — Phase 0:

> Ratify instance-per-tenant, or accept the weak tier explicitly and write down
> why. Licensing is settled (clause 2 permits it); what remains is the topology
> decision, which gates everything below.

Phase 1 (the driver) is done and unbound: merged as
[tinymemory#128](https://github.com/tinyhumansai/tinymemory/pull/128) and
registered in #2065. Phases 2 (provisioning), 3 (migration) and 4 (revisit the
derived layers) have not started and should not, until this is answered. The
infra deliverable on #1936 was left deliberately incomplete for the same reason:
HA, backup/restore and an upgrade path describe operating a deployment the
record recommends against.

## What the evidence says

Measured against a deployed CortexDB instance, not the product page.

**Option A is the only route to the strong tier.** The reachable token minter
does not confine a token to its scope. A token minted for scope A, pointed at
scope B: `POST /v1/recall` is refused `403 POLICY_DENIED`, but `GET
/v1/events?scope=B` returns B's records and `POST /v1/experience` into B is
accepted. Reproduced three times, including with narrowed capabilities. The
isolating alternative (`cortex-auth-ref`) is absent from the v0.9.8 assets, has
no public repository and no published contract.

That leak removes a *shared* instance with real per-tenant credentials from the
table — not instance-per-tenant, which carries its own credential and data
directory per tenant and needs no token scoping at all. So the leak **forces A
over B** rather than blocking A. #1936 required defaulting to the stronger tier,
and A is the only self-hosted way to reach it.

[...]

PR #2076, opened 2026-09-04
Withdraw the Cortex derived-layer finding: enrichment was off

CodeGhost21:
Finding 3 of `docs/spec/runtime/memory-engine-cortex.md` — "the derived fact and
belief tier does not work" — is withdrawn. It was measured on a server with
enrichment switched off, and v0.9.8 had no way to say so.

## What happened

CortexDB **v0.9.9** shipped 2026-09-03, the day after the record landed. Its one
change adds `embeddings`, `enrichment` and `content_processors` checks to
`GET /v1/admin/ready`, because a server could return `{"status":"healthy"}` and
`degraded: false` while quietly underdelivering. Run against our staging
deployment it answered immediately:

```json
"enrichment": { "enabled": false, "mode": "off",
  "detail": "enrichment OFF: facts, beliefs and concepts will stay empty" }
```

Three prerequisites were unset, none discoverable on v0.9.8:

1. `CORTEX_ENRICHMENT_URL` / `_API_KEY` — the enrichment router is a **separate
   lane** from `CORTEX_LLM_*`; `_MODEL` alone gives it a model with no endpoint.
2. `CORTEX_ENTITY_API_KEY` — the binary states "both the primary entity LLM and
   enrichment router are required". Without it, boot logs `LLM router not
   configured` even with `CORTEX_LLM_URL` and `_MODEL` both set.
3. `CORTEX_ENRICHMENT_DELAY_SECONDS` — **the on/off gate**. Unset means
   enrichment is off entirely.

With all three wired, enrichment reports `mode: "deferred:5s"`, both routers
start, and 600 events drain through the pipeline.

## Why this is a withdrawal, not a reversal

The re-measurement is void as well: the embedding provider was returning `402
Insufficient credits` on every call (0 ok / 622 failed, resident vectors
2009 → 0). No embeddings means no vectors, no LLM lanes and no facts — so it
measures our billing, not Cortex. **The derived tier is unproven in both
directions**, pending credit.

## What is unaffected

Finding 2 — the token-scope isolation leak — is untouched, and it is what
independently blocks the instance-per-tenant tier. The concepts lane still fails
with both routers verifiably up. The recommendation therefore still stands, on
the isolation leg alone; its "nothing over the incumbents" leg is suspended.

Docs only. Refs #1936, #2072.

## Summary by CodeRabbit

- **Documentation**
  - Added a correction clarifying that the derived fact/belief tier was evaluated with enrichment disabled.
  - Updated the documented status from a negative finding to “unproven in both directions.”
  - Documented previously undiscoverable prerequisites and revised related findings, layer descriptions, and open-question answers.

### DIFF
--- a/docs/spec/runtime/memory-engine-cortex.md
+++ b/docs/spec/runtime/memory-engine-cortex.md
@@ -13,7 +13,48 @@ proposal and the measurements behind it.**
 [tinymemory#128](https://github.com/tinyhumansai/tinymemory/pull/128) is
 registered, and both live suites pass against a real CortexDB. Selectable is not
 selected: `OPENCOMPANY_MEMORY` defaults exactly as it did, and choosing Cortex is
-the decision this record informs. It argues against it.
+the decision this record informs. It argues against it — with **one leg of that
+argument retracted**; see [Correction](#correction-2026-09-04) below.
+
+## Correction (2026-09-04)
+
+**Finding 3 below is withdrawn. It was measured on a server with enrichment
+switched off, and nothing on the wire said so.**
+
+CortexDB v0.9.9 (released 2026-09-03, after this record was written) adds
+`embeddings`, `enrichment` and `content_processors` checks to
+`GET /v1/admin/ready`, precisely because a server could report
+`{"status":"healthy"}` and `degraded: false` while underdelivering. Run against
+our deployment it answered immediately:
+
+```
+"enrichment": { "enabled": false, "mode": "off",
+  "detail": "enrichment OFF: facts, beliefs and concepts will stay empty" }
+```
+
+Three prerequisites were unset, none of them discoverable on v0.9.8:
+
+1. `CORTEX_ENRICHMENT_URL` / `_API_KEY` — the enrichment router is a separate
+   lane from `CORTEX_LLM_*`. `_MODEL` alone gives it a model with no endpoint.
+2. `CORTEX_ENTITY_API_KEY` — the binary is explicit that "both the primary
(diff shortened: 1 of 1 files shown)


CodeGhost21:
Both Codex findings were right, and the first one was a real error rather than a wording nit. Fixed in `afa2f3c6d`.

**P2 — "Say the scope leak requires instance-per-tenant isolation".** Correct, and I had inverted the record's own architecture conclusion. The topology table in this document marks one-instance-per-tenant **reachable** and recommends it, precisely because each instance carries its own credential and data directory — token scoping never enters into it. What finding 2 removes is the middle row, a *shared* instance using scoped tokens as tenant isolation. So the leak **forces** instance-per-tenant rather than blocking it. Reworded, and corrected in the two places I had propagated it to: #2072's body and my closing comment on #1936.

**P2 — "Qualify the remaining derived-tier conclusions".** Also correct; I qualified the finding and the summary but left three downstream conclusions asserting the withdrawn result:
- *Belief revision is not reachable* — its `no_facts_in_scope` is exactly what an enrichment-off server reports, so it tests the same misconfiguration. Now says so.
- *Phase 4* — was gated on "upstream defects being fixed" when whether there is an upstream defect at all is unproven. Now gated on re-measuring on a correctly configured server.
- The open question that said the derived-layer finding "still decides adoption" — now says it was withdrawn.

**CodeRabbit MD040.** Tagged `text` as suggested; it is a JSON fragment rather than a document, so `json` would be wrong.

Thanks — the first one would have shipped a decision record arguing against its own recommended topology.

Issue #2089, opened 2026-09-04
CortexDB beliefs never supersede: a contradicted claim stays live at full confidence


CodeGhost21:
## Summary

CortexDB builds beliefs but never **supersedes** them. A contradicted belief
stays live at full confidence beside its correction, with nothing marking which
one is current. Measured on staging against CortexDB v0.9.9 with enrichment
correctly configured.

This matters for OpenCompany because `FactStore` is one of the three knowledge
ports the memory seam swaps (`memory-engine.md`), and the decision on #2072 is
to make Cortex the default engine for newly provisioned hosted tenants. An agent
would recall a fact that was true months ago as though it were true now.

## Reproduction

Scope `probe:revision`. Three events establishing an owner:

```text
Priya Raman owns the billing service.
Billing escalations go to Priya Raman.
Priya Raman has owned billing since March.
```

`GET /v1/beliefs?scope=probe:revision` — 2 beliefs, as expected:

| claim | confidence |
|---|---|
| Priya Raman `owns` billing | 1.0 |
| Billing escalations `go_to` Priya Raman | 1.0 |

Then two events handing it over:

```text
Priya Raman handed billing over to Marcus Webb last week.
Marcus Webb now owns the billing service.
```

After the enrichment scanner and auto-layer tick:

| claim | confidence | `valid_to` |
|---|---|---|
| Marcus Webb `owns` billing | 1.0 | `null` |
| **Priya Raman `owns` billing** | **1.0** | **`null`** |
| Billing escalations `go_to` Priya Raman | 1.0 | `null` |

The new belief is appended. The contradicted one is not retired, not
down-weighted, and not stamped with a `valid_to`. `POST /v1/recall` for
"who owns the billing service?" returns all three, undifferentiated.

## Why this is not the same as the old finding

`memory-engine-cortex.md` previously recorded "Belief revision is not
reachable", on the grounds that facts never extracted so beliefs never built.
That was a [misconfiguration](https://github.com/tinyhumansai/opencompany/blob/main/docs/spec/runtime/memory-engine-cortex.md#correction-2026-09-04)
— enrichment was off. Corrected, the derived tier works well: 10 events yield
19 facts and 9 beliefs as resolved subject/predicate/object triples.

So this is the sharper version of that concern. Belief revision is *reachable*
and silently wrong, which is worse than absent: an empty layer is obviously
empty, whereas two contradictory beliefs at `confidence: 1.0` look like data.

## What we do not know yet

[...]

PR #2096, opened 2026-09-05
Correct two findings, and split the record at the seam

CodeGhost21:
Two corrections to the Cortex design record, plus the split the second one forced.
Both came from reading CortexDB's own reference **before** sending them anything.

## 1 — Finding 2 was a policy we did not set

`GET /v1/policy/effective` lists `scope.read.holistic`, `scope.read.descend` and
`scope.write.about_other` among the actor's **allowed** capabilities under the
preset we ran. The cross-scope reads and writes we measured were the deployment
tier doing what it was configured to do. Calling it an enforcement failure was
wrong.

Also corrected: the policy surface is not absent (`GET /v1/policy/deployment`
and `/v1/policy/effective` both answer `200`; only the mutation endpoints `404`,
matching the docs marking them experimental), and the minter we tested is
documented as *"a dev convenience, not a production issuer"* — we enabled it
under a production preset.

What survives is narrower and still decisive: the narrowing API is not in this
build and `cortex-auth-ref` is in no release asset and no doc page, so we can
neither configure the isolation nor test it. **Instance-per-tenant is
unchanged** — it never rested on Cortex's scopes.

## 2 — The belief-revision finding is withdrawn

#2089 said beliefs never revise. `/docs/api-reference/facts` states the rule:

> Newer facts about the same `(subject, predicate)` supersede older ones.

Our two claims were `ent_Priya_Raman | owns` and `ent_Marcus_Webb | owns` —
different subjects, nothing to supersede. The machinery *does* fire on the
documented key: in the same scope a fact carrying temporal language was closed at
`valid_to 2026-08-29`, and `GET /v1/facts/timeline` exposes the chain with
`include_superseded`. Neither had been used before the claim was filed. #2089 is
closed.

What survives is ours: single-valued ownership is not expressible through
natural-language ingest, because the extractor puts the person in the subject
however the sentence is phrased. Only matters if the derived layers reach the
read path, which they do not.

## 3 — The split

The file was at 500/500 lines, so the correction did not fit. The record now
divides at its natural seam:

| File | Holds | Lines |
|---|---|---|
| `memory-engine-cortex.md` | the decision, the plan, the invariants | 394 |
| `memory-engine-cortex-evidence.md` | the measurements behind them | 137 |

Pointers left under the old headings so existing links and anchors still land,
and the new page is indexed from the folder README, per
`assert-md-line-cap.sh`'s own guidance.

## Why this PR keeps growing

Four claims in this investigation have now traced back to our own configuration
or modelling — the derived tier (enrichment off), the ~18 GiB sizing figure (a
configured budget), the cross-scope access (a granted policy), and belief
revision (the documented supersession key). Every one was answered under
`/docs/operations/*` or `/docs/api-reference/*`. Every one was filed from
observed behaviour without reading the page first.

**One defect still stands after checking:** the concepts lane reports no LLM
router while two other lanes start on the same URL and model, has no routable
task in the provider fleet's list, and has no configuration in the docs at all.
Four configurations were tried.

Docs only. Refs #1936, #2089.

## Summary by CodeRabbit

* **Documentation**
  * Added runtime documentation for the Cortex memory engine, including design details and supporting measurements.
  * Clarified that Cortex provides ranked retrieval from embeddings, while derived fact/belief capabilities are not replaced.
  * Corrected findings about belief revision and cross-scope behavior, documenting deployment and policy limitations.
  * Updated the runtime documentation index and corrected links between the design and evidence pages.

### DIFF
--- a/docs/spec/runtime/README.md
+++ b/docs/spec/runtime/README.md
@@ -34,6 +34,9 @@ Supporting docs:
     applied, and how a company created before it keeps working
   - [memory-engine.md](memory-engine.md) — the `OPENCOMPANY_MEMORY` overlay and
     why an ephemeral data root refuses to boot
+  - [memory-engine-cortex.md](memory-engine-cortex.md) — the hosted-Cortex
+    design record for #1936, with its measurements split into
+    [memory-engine-cortex-evidence.md](memory-engine-cortex-evidence.md)
   - [data-root.md](data-root.md) — the root itself: resolution order, ownership,
     and two processes wanting the same directory
   - [offline.md](offline.md) — running with no network at all: the documented
--- a/docs/spec/runtime/memory-engine-cortex-evidence.md
+++ b/docs/spec/runtime/memory-engine-cortex-evidence.md
@@ -14,7 +14,8 @@ it is worth separating the two because conflating them points a driver plan at
 the wrong thing.
 
 Measured on the deployment, with the LLM lanes believed configured and the
-server reporting healthy — see [Correction](#correction-2026-09-04) for why that
+server reporting healthy — see [Correction](memory-engine-cortex.md#correction-2026-09-04)
+for why that
 belief was wrong:
 
 | Cortex layer | Endpoint | Contents |
--- a/docs/spec/runtime/memory-engine-cortex-evidence.md
+++ b/docs/spec/runtime/memory-engine-cortex-evidence.md
@@ -1,0 +1,137 @@
+# What the Cortex evaluation measured
+
+Split out of [`memory-engine-cortex.md`](memory-engine-cortex.md), which
+carries the decision and the plan. This page carries the measurements behind
+them, at the length they need. Read the parent's
+[Correction](memory-engine-cortex.md#correction-2026-09-04) first: three of
+the findings recorded here were later traced to our own configuration, and
+each is marked where it appears.
+
+## Layers are not capability families
+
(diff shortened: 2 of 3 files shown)


CodeGhost21:
All three findings were valid and are fixed in `1059c10ae`. Both reviewers independently caught the same underlying mistake, which is worth naming: I corrected finding 2 and left every conclusion drawn from it in place.

**Topology table** (@coderabbitai and @chatgpt-codex-connector). It still said a shared instance with per-tenant credentials is unreachable because "the reachable minter does not confine a token to its scope" — the exact framing this branch retracts. Now states what is actually missing: the production issuer and the policy-narrowing API, neither of which is in this build.

**The open question** (@chatgpt-codex-connector). It still asked whether the gap "is a defect". It isn't — the deployment tier grants the behaviour — so it now asks the question that is actually open: what should a self-hosted deployment bind actors with, and will `PUT /v1/policy/{tier}` leave experimental.

**The broken anchor** (both). The split left a same-page `#correction-2026-09-04` link on the evidence page while the heading stayed in the parent, so a reader landing directly on measurements that are known-invalid had no route to the explanation. Now points at the parent. I checked the other two cross-file anchors resolve; my first verification script said they didn't, but it was mishandling the em-dash in "Belief revision — finding withdrawn" — the headings are correct.

The general lesson, since this is the fourth correction on this record: a retraction is not finished when the finding is edited. It is finished when everything downstream of it agrees. I had checked the finding and not the summaries.

PR #2109, opened 2026-09-07
feat(hivemind): private asides, tinyhivemind bump, and a vending-machine bundle that exercises agent-to-agent comms

senamakel:
## What this is

Updates the vendored `tinyhivemind` (`e272c84` → `63dfb22`, ~540 commits), migrates the host to its new API, implements **private asides** as a second agent-to-agent seam, and adds `companies/vending_machine_co` — a bundle whose desks talk to each other while running a simulated vending-machine operation over an MCP server.

Everything here has been run live against real models on the ladder router, not only unit-tested. Several of the fixes below exist *because* of that run.

## The submodule bump and its migration

`tinyhivemind` gained private asides, which is a breaking change to the session API:

- `SessionQuery` grew `viewer`. `EpisodeDriver::run` issues `Viewer::Operator` — the true, unelided medium — because `step` must fold one transcript for the whole room. A per-viewer fold would make `quorum::standings`, `attention::bids` and `directory` return well-formed **wrong** answers with no error path: a quorum one member can see and another cannot. `project_for` stays the only thing that narrows, per speaker.
- `project_for` now returns owned rows (it synthesises collapsed aside stubs that exist in no source slice), so `render` / `line_from` / `evidential` widened from `&[&SessionMessage]` to `&[SessionMessage]`.
- `LogMessage` / `SessionMessage` gained `audience` / `elided`; `read_pinboard` takes a viewer — read as the *speaker*, so a pin over an aside cannot quote content to somebody outside it.

## Private asides

Two members of one desk saying something the rest of it cannot read. Spec: `docs/spec/runtime/hivemind-asides.md`.

- **Auditable, not confidential.** A row is elided, never removed: its sequence, author and audience stay, so a `^N` citation still resolves and a peer can see that an exchange it may not read happened. Operators and people read every aside in full. **This is not a security boundary and nothing may be built on it as one.**
- **Information, never support.** A trace in a non-`Desk` row contributes nothing, for every reader — the same rule cross-desk referral already accepts at a channel boundary, applied inside one desk.
- **It costs no turn** (ADR 0011). One authorized turn produces the member's ordinary desk line *and*, optionally, one private row. A refused audience is **dropped, not published** — falling back to the desk would put a second desk-visible contribution on one turn.
- **Off by default**, because upstream measured it and it lost on answer quality. Enabling it is a decision, not a default.

Journal cost is one additive field on `AgentReply` (`audience: Vec<String>`, `skip_serializing_if = "Vec::is_empty"`), so a desk-visible reply serializes byte-for-byte as before and **no stored record needs migrating**.

## Fixes found by running it live

- **`referral_summary` claimed "another desk" for same-desk questions.** `reach` widens strictly, so a desk may legitimately put a question to its own seat — but a live run reported *"asked 2 questions of another desk (@fleet_tech on ops, @field_realist on ops)"* on the ops desk, naming two ops seats. An operator reading that has been told the room reached outside when it did not. `AskedQuestion` now carries `crossed` and the close reports the two separately. Regression test included.
- **Asides were never used until ADR 0011 landed.** A six-day live run with asides enabled used the marker **zero times**, while the same seats repeatedly wrote `!question @peer` — the move already in their list that did not cost them their contribution. After making the aside ride alongside the turn, a real run produced one on its first day.

## The bundle

`companies/vending_machine_co` — eight machines at five host sites, a finite warehouse with a 4-day lead time, contracts that renew. Three hive desks (`ops` 4 seats, `commercial` 3, `intel` 2), all with cross-desk referral, and asides on `ops` only.

The whole work environment is one MCP server (`scripts/vending/`): fleet telemetry, warehouse, margin, clients, incidents, market feed, plus the writes that change them. Every read an agent makes and every change it causes is a tool call. `scripts/vending-sim.py` advances a clock and posts each day's triggers into the desk that owns the *decision* — not the desk holding the most facts, because whether they ask each other is what the run measures.

## Evidence from a live run

A real ops episode, models on the ladder router:

```
route_planner    !propose #ops-day-1 run the van VM-401→VM-301→VM-201→VM-501…
fleet_tech       !evidence ^13 VM-301 (700 footfall) about to blow SANDWICH 0.1d/ENERGY 0.2d…
stock_controller !evidence ^13 warehouse holds 167 ENERGY-250, 63 SANDWICH-1… 4-day lead time
field_realist    !evidence ^13 VM-201/401 need priority, but VM-301/302 are the pressure points
route_planner    !propose #ops-day-1 route VM-301 then VM-302 … → VM-401 → VM-201 → VM-501
fleet_tech       !support ^18 covers every machine at zero or under a day
route_planner    !commit ^18
```

The figures are live MCP reads, only the planner proposed, the realist never supported, and the planner revised its route in response to the objection. And an aside riding alongside a turn — two rows, one turn:

```
[22] fleet_tech  !evidence #ops-day-1 ^13 fleet read shows no faults anywhere…   ← the room's
[23] fleet_tech  !aside @route_planner how many visits can one van make in a day? ← private
```

## Verification

`cargo fmt --all --check` clean · `cargo clippy --all-targets --features openhuman,mcp -- -D warnings` clean · **7113 lib tests pass** (101 hivemind, incl. 11 aside unit + 9 integration) · content tests 26/26 · all integration targets pass.

## Not included, deliberately

`src/store/memory/cortexdb.rs` is **untouched**. A local CortexDB started by `scripts/cortexdb-up.sh` has no LLM router or embedding endpoint wired, so its ranked-recall index lags writes by minutes and every agent turn fails at its first `memory_store`. I tried a fix, verified it did not work, and reverted rather than ship an unverified change to a shared driver. The README documents the trap and how to point at a provisioned instance instead; the underlying issue deserves its own PR.

## Summary by CodeRabbit

* **New Features**
  * Added a complete Northgate Vending company bundle with operational desks, specialized roles, ledgers, seeded tasks, and approval workflows.
  * Added a deterministic vending-machine simulator covering inventory, sales, pricing, incidents, client relationships, supplier orders, and market signals.
  * Added optional private desk asides with bounded, auditable peer messaging and required surfacing.
* **Documentation**
  * Added setup, usage, simulation, and private-aside documentation.
* **Tests**
  * Added coverage for private messaging, referrals, visibility, and vending bundle validation.

### DIFF
--- a/companies/vending_machine_co/README.md
+++ b/companies/vending_machine_co/README.md
@@ -108,8 +108,27 @@ operator later installs.
 
 ```bash
 # 1. Memory. CortexDB is a standalone service on :3141, not the removed in-pod
-#    tinycortex engine. The script starts (or reuses) it and prints the exports.
+#    tinycortex engine. The script starts (or reuses) a local one and prints the
+#    exports:
 eval "$(./scripts/cortexdb-up.sh)"
+#
+#    …but prefer a shared instance that is actually provisioned. A local
+#    container started by that script has no LLM router and no embedding
+#    endpoint wired, and its ranked-recall index then lags the write by
+#    *minutes* — far past the driver's 4s read-after-write check
+#    (`INGEST_RECALL_VISIBILITY_TIMEOUT`), so every agent turn fails at its
+#    first `memory_store` and the desk reports every seat as unfinished. The
+#    failure names `/v1/recall` and looks like a driver bug; it is an
+#    unprovisioned instance. Point at one with an LLM/embedding endpoint
+#    instead — e.g. the shared box on the tailnet:
+#
+#      export OPENCOMPANY_MEMORY=remote
+#      export OPENCOMPANY_MEMORY_DRIVER=cortexdb
+#      export OPENCOMPANY_MEMORY_URL=http://<host>:3141
+#      export OPENCOMPANY_MEMORY_API_KEY=<that instance's CORTEX_API_KEY>
+#      export OPENCOMPANY_MEMORY_ACTOR=service:opencompany
+#
+#    Check before you run: a store→recall round trip has to complete in seconds.
 
 # 2. The company.
 OPENCOMPANY_INFERENCE_URL=http://127.0.0.1:6969/v1 \
--- a/companies/vending_machine_co/README.md
+++ b/companies/vending_machine_co/README.md
@@ -134,10 +134,13 @@ The operational truth lives in the MCP server; duplicating it here would produce
 two records that disagree by the end of the week. These four hold what the
 server cannot:
 
-- **`decisions`** — the *reasoning*. A transcript scrolls out of a
+- **`episodes`** — the *reasoning*. A transcript scrolls out of a
   thirty-message window and nobody re-reads one to learn why the van skipped
   Vulcan for a fortnight. The `not_doing` field is the most useful and the most
-  likely to be left empty.
+  likely to be left empty. (Named `episodes` and not `decisions` because
+  `decisions` is a built-in ledger the runtime ships and a declared slug may not
+  shadow one — and the two are different records anyway: the built-in holds the
+  company's standing calls, this holds what one deliberation concluded.)
 - **`clients`** — what each host was *promised*, and by whom. A promise nobody
   wrote down is one this company will break by accident.
 - **`incidents`** — the *pattern*. One jam is an event; VM-301 jamming four
--- a/companies/vending_machine_co/README.md
+++ b/companies/vending_machine_co/README.md
@@ -1,0 +1,138 @@
+# Northgate Vending
+
+A vending-machine operator: eight machines at five host sites, a finite
+warehouse behind them, and host contracts that renew whether or not anybody
+prepared for it. Three hive-mind desks run it.
+
+Every other hive bundle here answers a **question** — `hive_math_lab` solves a
+stated Project Euler problem and stops. This one runs an **operation**. Work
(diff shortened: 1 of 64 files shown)


senamakel:
The `tinysweeper/critique` `missing-module` finding is a false positive from a stale index — its own summary says so ("The code index is behind this pull request (indexed at `d676c7c1a49a`)"), and the finding text hedges accordingly: "If the module does exist elsewhere in the branch but was not shown in the diff, the finding is moot."

It does exist. `src/hivemind/aside.rs` is added by this PR and `src/hivemind/mod.rs:60` declares `pub mod aside;`. The crate compiles: `cargo clippy --all-targets --features openhuman,mcp -- -D warnings` is clean and 7113 lib tests pass locally, and the Rust CI jobs on this PR build the same tree.

senamakel:
**Merging with two non-blocking reds, both accounted for:**

- `Desktop` — fails on `update::test::the_shipped_placeholder_is_not_configured` (`src/update.rs:161`). This branch does not touch `update.rs` (`git diff origin/main...HEAD -- '*update.rs'` is empty), and the same job fails on `main` (run 34050769896). Pre-existing.
- `tinysweeper/critique` — a stale-index false positive; see the comment above. `src/hivemind/aside.rs` is added by this PR and declared at `src/hivemind/mod.rs:60`.

**Green where it counts:** `Rust` (18m17s), `Rust (openhuman, tinymemory)` — the gated lane that runs the hivemind suite — `Rust (mail)`, `Rust (mongodb)`, `Gated host binary`, `Markdown line cap`, `Console E2E (first run)`, and every `tinysweeper` check except the one above.

PR #2116, opened 2026-09-07
fix(hivemind): two referral wording defects, and retune the signals desk from the benchmark

senamakel:
Follow-up to #2109. Two wording defects a live run printed, and one desk setting changed from measurement rather than from argument.

## Two defects the live run printed

Both were visible in a real `companies/vending_machine_co` transcript, and neither had a test.

**`The room 1 went unanswered.`** — `referral_summary` hung every clause off one `"The room …"` prefix, but the failure and cap counts carry their own subject. An episode whose only referral fact was a failure produced that sentence. The summary is now assembled as sentences: the clauses that continue "the room" join under it, the counted ones stand alone.

**`@route_planner on the Operations desk desk did not answer the question.`** — the note appended `" desk"` unconditionally, and every desk in this repo is *named* "… desk". `named_desk` now adds the word only when the name does not already end in it, so `"Operations desk"` and `"eng"` both read correctly.

## One setting changed, from the benchmark

`companies/vending_machine_co`'s `intel` desk dropped `require_evidential` and went from a 6- to an 8-turn budget. It is the only knob in the bundle set from measurement.

Stacking `require_evidential` on a **two-member** desk whose quorum is already unanimity means a decision needs *both* seats to file a citation that chains to a stated fact. The deliberation benchmark (`vendor/tinyhivemind/crates/tinyhivemind-hive/examples/bench`, 2000 seeded rooms, `--seed 1`):

| `--agents` | `hive+` decided | `hive+ev` decided | `hive+` correct | `hive+ev` correct |
|---|---|---|---|---|
| 4 | 94.6% | 40.0% | 76.8% | 37.7% |
| 3 | 99.8% | 70.0% | 71.6% | 55.8% |
| 2 | 91.6% | **36.2%** | 63.6% | **30.7%** |

Live runs agreed: across 19 episodes the company decided **10**, and the signals desk spent its budget without deciding in nearly every episode it opened.

`ops` and `commercial` keep the rule. They are larger, and their seats read hard figures off the `vending` MCP on almost every turn, so a citation that reaches a fact is cheap there in a way it is not on a desk whose job is to weigh a signal nobody has measured yet. I did **not** blanket-remove it on that table alone — the benchmark's task is averaging noisy private numbers, where evidence deposits are synthetic, and a `--blind-evidence` control (first turns are deposits, closer to how these desks behave) narrows the gap at two members to 44.1% → 41.3%. The change is where the benchmark and the live runs agree.

## Benchmark evaluation of the aside work in #2109

Recorded here because #2109 shipped the mechanism and this is the measurement of it. `hive+along` is ADR 0011 — the aside riding alongside a turn — which #2109 implements.

**Hidden profile, 2000 rooms:**

```
hive+along − hive+:  +0.0 [+0.0, +0.0]
hive+aside − hive+:  -0.1 [-0.4, +0.2]     (turn-spending)
hive+mute  − hive+:  -0.2 [-0.5, +0.0]
hive+fact° − hive+: +27.6 [+25.3, +29.9]   (off the floor)
hive+pooled − hive+: +75.5 [+73.6, +77.4]  (ceiling)
```

The zero-width interval on `hive+along` is the point: riding alongside is **bit-identical** to the baseline, so the episode genuinely cannot see the aside row — ADR 0011's first property, measured rather than asserted. On the uniform profile the turn-spending arms cost 2.2 points and 1.4 turns/episode (76.6% vs 78.8%, 8.26 vs 6.84) while `hive+along` matches `hive+` exactly on both.

So asides are now **free, and still not worth points**. Their claim is auditability and bounded independence, which is what the spec says. The lever this benchmark actually shows is `hive+fact°` (+27.6) — an exchange held *off the floor*, before the episode opens — and that is a different mechanism from an aside, worth its own issue rather than a change here.

Harness self-check (`--stats-check`) passes, so these intervals are the harness's own.

## Verification

`cargo fmt --all --check` clean · `clippy --all-targets --features openhuman,mcp -- -D warnings` clean · **7115 lib tests pass** (103 hivemind, +2 regressions covering exactly the two sentences above).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected referral summaries so unanswered-question counts use proper singular and plural grammar.
  * Improved referral messages to avoid duplicating “desk” in desk names.
  * Clarified episode-close summaries when questions go unanswered or are declined.

* **Configuration**
  * Increased deliberation budgets for the intelligence and signals desks from 6 to 8 turns.
  * Updated evidential-support requirements for the intelligence and signals desks.
  * Increased repetition limits for desk deliberations.
  * Updated operations and commercial dominance limits to use grounded-share percentages.

### DIFF
--- a/companies/vending_machine_co/README.md
+++ b/companies/vending_machine_co/README.md
@@ -56,7 +56,7 @@ cannot see it, would be a desk that never had to be convinced. See
 | --- | --- | --- | --- |
 | `ops` | `route_planner`, `fleet_tech`, `stock_controller`, `field_realist` | 2 of 4, 12 turns | The van's finite day: route, shelves, faults |
 | `commercial` | `account_manager`, `pricing_analyst`, `contract_counsel` | 2 of 3, 9 turns | Margin, prices, host sites, contracts |
-| `intel` | `market_scout`, `demand_analyst` | 2 of 2, 6 turns | What the market did, and whether it changes a plan |
+| `intel` | `market_scout`, `demand_analyst` | 2 of 2, 8 turns | What the market did, and whether it changes a plan |
 
 Each desk restricts its seats' moves (`[group_chat.hive.moves]`), and the
 restrictions carry the design:
@@ -78,10 +78,24 @@ restrictions carry the design:
   of two — unanimity. A room of two that could carry on one supporter is a
   single responder with extra steps.
 
-`require_evidential` is on for all three, so a `!support` with no `^citation`
-adds nothing to quorum. In this company that bites hardest on the stock
-controller: "we have enough stock" is a claim about a number
+`require_evidential` is on for **`ops` and `commercial`**, so a `!support` with
+no `^citation` adds nothing to quorum. In this company that bites hardest on the
+stock controller: "we have enough stock" is a claim about a number
 `warehouse_status` will actually print.
+
+`intel` deliberately does **not** have it, and that is the one setting here
+chosen from measurement rather than argument. Stacking `require_evidential` on a
+two-member desk whose quorum is already unanimity means a decision needs both
+seats to file a citation that chains to a stated fact. The deliberation
+benchmark (`vendor/tinyhivemind/crates/tinyhivemind-hive/examples/bench`) puts
+numbers on it — at `--agents 2` it takes the decided rate from **91.6% to
+36.2%** and correctness from 63.6% to 30.7% over 2000 seeded rooms — and live
+runs agreed: the signals desk spent its budget without deciding in nearly every
+episode it opened. Its budget went from 6 turns to 8 for the same reason. The
+larger desks keep the rule, because their seats read hard figures off the
+`vending` MCP on almost every turn, so a citation that reaches a fact is cheap
+there in a way it is not on a desk whose job is to weigh a signal nobody has
+measured yet.
 
 ## Tool servers
 
--- a/companies/vending_machine_co/company.toml
+++ b/companies/vending_machine_co/company.toml
@@ -248,13 +248,31 @@ members = ["market_scout", "demand_analyst"]
 
 [group_chat.hive]
 enabled = true
-turn_budget = 6
+# Eight, not six. See the evidential note below: this desk was the most
+# constrained in the company and spent its budget without deciding in almost
+# every live run.
+turn_budget = 8
 # Two members, so a quorum of two is unanimity — correct here. A room of two
 # that could carry on one supporter is a room where either member decides
 # alone, which is a single responder with extra steps.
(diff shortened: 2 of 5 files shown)


senamakel:
## Added: a units bug the sweep caught

Running `--sweep --agents 4` (864 policies × 1000 rooms) surfaced something worth more than the tuning it was meant to check.

`EpisodePolicy::dominance_cap` is a **percent of the room's grounded share** — the library damps a member when `share * 100 > total * cap` (`attention::exceeds`), and its default is `50`. The sweep only ever tries `40` and `60`.

Two things had it as a count:

1. **`src/hivemind/types.rs` documented it as "Turns one member may take before the attention market damps its bids."** That is wrong, and it is the source of the error.
2. **Shipped manifests set `dominance_cap = 3`** believing it meant three turns. It means **three percent**: with a handful of grounded contributions on the floor, *every* member that has said anything exceeds the cap and takes the `-3000` dominance penalty on every later bid. The knob meant to damp the loudest seat was damping the whole room.

Fixed here: the doc comment, and `companies/vending_machine_co` (`3` → `40`, the value the sweep's best four-seat policy uses). `repetition_cap` went `1` → `2` on the same evidence — the sweep's best policies all use 2 or 3, and its units were at least correct already.

### Not fixed here, deliberately

**`companies/hive_math_lab` has the same `dominance_cap = 3`**, with a comment describing "how far ahead of the field a proposal's supporter count may run" — which is where this bundle inherited both the value and the misunderstanding. I have not retuned somebody else's lab in a PR they are not expecting; its owner should decide, and the corrected doc comment is now there to read. Happy to do it in a follow-up.

### Sweep result, for the record

Best policy for a four-seat desk over 864 000 simulated episodes: `budget 16, quorum 3, blind yes, dominance 40, repetition 2, refute off, evidential no` → 96.6% decided, 77.6% correct, 6.84 turns/ep, against 74.5% for the matched-budget vote. The worst — quorum 4 (unanimity at four seats) with evidential and refute on — decides **0.0%**.

`cargo fmt --check` clean · `clippy -D warnings` clean · **7115 lib tests pass** · content tests 26/26.

senamakel:
Merging. `Rust`, `Rust (openhuman, tinymemory)` (the gated lane running the hivemind suite), `Rust (mail)`, `Rust (mongodb)`, `Gated host binary`, `Desktop` and `Markdown line cap` all pass.

`Console E2E (live brain)` is red on four tests — `composio-catalog-deadline`, `connections-authority`, `desktop-connections`, `workflow-selection-persist`. None touch hivemind, referral or this bundle, and the same job is red on `main` (run 34096823618). Pre-existing.

Issue #864, opened 2026-08-14
Workflows tab: selected workflow resets to the default when you switch tabs and return


oxoxDev:
## Summary
The Workflows tab does not remember which workflow was selected when you navigate away and come back. Selecting a specific workflow, switching to another tab (e.g. Workspace), then returning to Workflows resets the view to the default workflow instead of the one you were working on.

## Problem
Repro:
1. Open Workflows. The default workflow (say W2) is selected.
2. Select a different workflow (W1 or W3).
3. Switch to another tab (Workspace, Company, etc.).
4. Return to Workflows.

Expected: W1/W3 is still selected. Actual: the view is back on the default workflow, and you have to re-select the one you were using.

The selected workflow lives in `WorkflowsView` local state (`selectedId`), which is dropped when the view unmounts on tab switch. Nothing persists the selection across the tab boundary.

## Solution
Persist the last-selected workflow (and ideally the open run) so returning to the Workflows tab restores the previous selection — e.g. lift the selection to a store/context that outlives the view, or encode it in the URL hash (`#/workflows/<id>`), or otherwise restore it on re-mount.

## Impact
Frequent context loss when an operator is bouncing between tabs mid-work — a recurring, small-but-costly friction on a primary surface.

## Related
- #339 (deep-link `?run=<runId>` — the hash already carries a selected workflow id when a run is linked)

PR #2117, opened 2026-09-07
test(e2e): follow the product-scope flags the console now ships

### LINKED ISSUE #2110: Console E2E is red on main and on every PR: three specs failing since ~2026-09-04
## What happens

`Console E2E` and `Console E2E (live brain)` fail on **main** and on **every open PR** that runs them. The same three specs fail every time:

- `test/e2e/connections-authority.spec.ts:145` › an admin is still offered every control across the three pages
- `test/e2e/desktop-connections.spec.ts:420` › a desktop whose host did not start offers to start it, not a choice of where
- `test/e2e/workflow-selection-persists.spec.ts:176` › a company switch does not reuse the previous company's workflow route (#864)

`Console E2E (live brain)` additionally fails `test/e2e/composio-catalog-deadline.spec.ts:223` › rotating the credential re-reads the catalog instead of warning about it.

The rest of the suite passes (~348 / ~369 green), so this is not an infrastructure outage.

## Failure shapes

```
connections-authority:153   expect(page.locator("#company-credential")).toBeVisible({timeout: 30_000})
                            -> element(s) not found
desktop-connections:432     expect(page.getByTestId("no-connection-run-here")).toBeVisible()
                            -> element(s) not found
workflow-selection-persists locator.click: Test timeout of 60000ms exceeded
```

All four are connections / credential / Composio-shaped surfaces.

## Not attributable to any one PR

- Reproduced on a **re-run** of the same commit, so it is deterministic rather than flake.
- Fails identically on unrelated PRs #2091 (auth/GraphQL, based on the newer `fd3e48f1`), #2088 and #2087.
- Fails on main's own CI: run `34050769896` @ `d304ab8d` (2026-09-06) — `Desktop`, `Console E2E` and `Console E2E (live brain)`.

Because `pull_request` CI builds the merge commit, every PR inherits it, which makes it read as "my branch broke E2E" to each author in turn.

## Timeline

Last main run with these lanes green: `33906023440` @ `e2c790cb`, 2026-09-04 18:27. Failures from roughly 2026-09-04 onward.

Note that a green tick on main is not evidence the breakage is fixed: docs-only merges (e.g. `d6a9cdd4`) **skip** every one of these lanes via `Changed areas`, so the run reports success without running them.

## Not investigated

I have not root-caused this — I ran it down only far enough to establish it was not the PR I was working on (#2075). Filing so the next author does not repeat that.

Related: the `Desktop` lane is red across main and all PRs for a separate and fully diagnosed reason — `src-tauri/src/update.rs::the_shipped_placeholder_is_not_configured` contradicts the pubkey installed by `846913029`. Tracked separately.

CodeGhost21:
Closes #2110. Unbreaks `Console E2E` and `Console E2E (live brain)`, red on main and on every open PR.

## Cause

`f1348aa67` (#2074) added `frontend/src/product-scope.ts`, hiding surfaces "while the product is scoped to one company per install". Three specs still assert those surfaces. That PR updated **ten** other specs for the same change, so these three were missed rather than deliberately left — which is why each is corrected here rather than relaxed or deleted.

Because `pull_request` CI builds the merge commit, every open PR inherits the failure and it reads to each author as though their own branch broke the console.

## The three

**`connections-authority.spec.ts`** — asserted `#company-credential`, which `OAuthView` now hides along with the managed Composio route. That card is gone on *every* build, so it cannot be the invariant any more.

It now asserts the provider grid's `"Sign in"` — precisely the control the member case immediately above asserts a member does **not** get, so the admin/member difference this test exists for is still what is being pinned. Presence, not enabledness: with no credential there is nothing to authorize against, so it renders disabled on a host with no Composio compiled in, which is this lane. Asserting it enabled would pass only on the gated build and quietly turn this into a second Composio test.

**`desktop-connections.spec.ts`** — `canStartHere` is `embedded.roster`, and the bridge stub never implemented `oc_local_instances`, so `localInstances()` answers `null`: a shell predating the roster, which honestly *cannot* offer to start a host. The spec was asserting the offer against a shell too old to make it.

Adds an opt-in `instances` to `BridgeConfig` and uses it in that one case. Deliberately not the default — several cases in this file depend on the pre-roster shape, and flipping it globally would be a silent upgrade for all of them.

**`workflow-selection-persists.spec.ts`** — skipped, not deleted. `COMPANY_SWITCHING_HIDDEN` makes `showCompanies` false, and with `HOSTS_HIDDEN` the trigger has nothing to open and falls to its nameplate branch, so the `"Other"` menu item this drives cannot exist. That is the 60s timeout, not a routing regression.

#864 is still a real guarantee and this is still its spec, so it stays in the tree with the reason attached. Driving the switch some other way would assert a path no operator can take. Skipped at the declaration rather than on the flag because `test/e2e` supplies no `@/*` alias — deliberately, per `tsconfig.e2e.json` — so `COMPANY_SWITCHING_HIDDEN` is not importable there.

## Checks

Run against a real host, not by inspection:

- Full suite: **360 passed, 33 skipped, 0 failed** (the 33rd skip is the one added here).
- The three specs alone: 9 passed, 1 skipped.
- `npm run typecheck:e2e` clean.

No production code changes; three test files, +67/-8.

Note I did **not** run `prettier --write`: it uses an 80-column default against this repo's wider style and reformatted the files wholesale (98 files are already "unformatted" by that config), which would have buried the change in churn.

cc @oxoxDev as the author of #2074 — if any of these surfaces was hidden further than intended, the right fix is in the console rather than here, and I would rather hear that than have a green tick paper over it.

## Summary by CodeRabbit

- **Tests**
  - Updated connection access checks to validate the read-only state and available sign-in action.
  - Expanded desktop test coverage for environments with stopped local instances.
  - Improved coverage for connection-console, company-creation, wallet-mode, and lifecycle reset-button behaviors under gated product configurations.
  - Skipped scenarios for unavailable company switching and legacy managed-route credential rotation, while documenting the affected flows.

### DIFF
--- a/frontend/test/e2e/composio-catalog-deadline.spec.ts
+++ b/frontend/test/e2e/composio-catalog-deadline.spec.ts
@@ -220,7 +220,20 @@ test("a host that never answers at all still gets the honest warning", async ({ 
   await expect(page.getByTestId(PROBE_FAILED)).toBeVisible({ timeout: 60_000 });
 });
 
-test("rotating the credential re-reads the catalog instead of warning about it", async ({
+// Skipped, not deleted: the console behaviour is still worth pinning, but the
+// control this drives is gone. It fills the legacy managed-route token card,
+// which `showManagedTokenCard` gates on `mode === "managed"` — and with
+// `COMPOSIO_MANAGED_HIDDEN` set, `MODE_ORDER` holds only `byok`, so
+// `formModeFor` can never answer `managed` and the card never renders. That is
+// the `locator.fill` timeout on the field below, not a regression in the
+// re-read.
+//
+// Every other case in this file sets the token through the API; only this one
+// needed the UI, because a *rotation through the form* is what it is about.
+// Retargeting it at the BYOK `composio-api-key` field would be a different
+// credential with different eviction semantics, so it is left for whoever owns
+// the managed route to redirect or retire.
+test.skip("rotating the credential re-reads the catalog instead of warning about it", async ({
   page,
 }) => {
   test.setTimeout(120_000);
--- a/frontend/test/e2e/connections-authority.spec.ts
+++ b/frontend/test/e2e/connections-authority.spec.ts
@@ -144,13 +144,25 @@ test("a member sees what is connected but is offered nothing that changes it", a
 
 test("an admin is still offered every control across the three pages", async ({ page }) => {
   await openSettingsPage(page, "oauth");
-  // The member's banner is absent, and the credential surface is present.
-  // The company-credential key is the Apps page's write surface on every
-  // build: the Composio token card only renders when the host reports a
-  // composio credential the admin may override (default-feature hosts never
-  // do), so it is not the invariant to assert here.
+  // The member's banner is absent, and the control it was refused is present.
+  //
+  // This used to assert `#company-credential`, on the reasoning that the
+  // company-credential key is the Apps page's write surface on every build.
+  // `src/product-scope.ts` hides the OpenHuman-managed Composio route and
+  // `OAuthView` hides that card with it — deliberately, since a company
+  // reaching Composio through its own account has nothing to spend that key on.
+  // It is no longer a surface on any build, so it cannot be the invariant.
+  //
+  // What survives is the provider grid's own "Sign in", which is exactly the
+  // control the member case above asserts a member does NOT get. Presence, not
+  // enabledness: with no credential there is nothing to authorize against, so it
+  // renders disabled on a host with no Composio compiled in — which is this
+  // lane. Asserting it is enabled would pass only on the gated build and turn
+  // this into a second Composio test.
   await expect(page.getByTestId("connections-read-only")).toHaveCount(0);
-  await expect(page.locator("#company-credential")).toBeVisible({ timeout: 30_000 });
+  await expect(page.getByRole("button", { name: "Sign in", exact: true })).toHaveCount(1, {
+    timeout: 30_000,
+  });
 
   await openSettingsPage(page, "mcp");
   await expect(page.getByTestId("mcp-read-only")).toHaveCount(0);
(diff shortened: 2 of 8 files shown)


CodeGhost21:
Green across the board, including all three Console E2E lanes — the first clean run in this repo since 2026-09-04.

Two more stragglers turned up after the first push, both the same root cause and neither visible until this PR ran the lanes:

**`Console` (unit) — 10 failures across 4 files.** Pre-existing on main, not introduced here: I checked out `e8123a348` clean and reproduced them. They stayed hidden because `Changed areas` only runs that lane when a PR touches the frontend, and none had since #2074. One cause for all ten — `canCreateCompanies` became `!COMPANY_SWITCHING_HIDDEN && carriesPlatformBearer`, so the create dialog and Settings' Reset never mount.

Neutralised the flag in those four files rather than skipping them, because they cover the dialog's own behaviour once open — the wallet field, the preflight race, the pre-archive guard, the lifecycle disable — which still ships and can still regress. The *gate* has purpose-built coverage in `product-scope-hidden-surfaces.test.ts`, which already names Settings' Reset among the triggers it closes; duplicating it here would only give two places to disagree. If you would rather see those retired than kept alive behind a mock, that is a one-line change either way.

**`composio-catalog-deadline.spec.ts:223`** rotates through the legacy managed-route token card, which `showManagedTokenCard` gates on `mode === "managed"` — unreachable now that `MODE_ORDER` holds only `byok`. Skipped rather than retargeted: every other case in that file sets the token through the API, and this one used the form because rotation *through the form* is the thing under test. Pointing it at the BYOK `composio-api-key` field would be a different credential with different eviction semantics, so I left that call to whoever owns the managed route.

Final tally across #2074's product-scope flags: **3 specs corrected**, **3 skipped** with the flag named so they come back when it clears, **4 unit files** kept covered.

Verified locally at each step rather than by inspection — full unit suite 495 files / 4584 tests, E2E 360 passed / 33 skipped, typechecks and build clean.

One note for review: I did not run `prettier --write`. It formats at an 80-column default against this repo's wider style and rewrote all three files wholesale on the first attempt (98 files are already "unformatted" by that config); I reverted it so the diff stays the change and not the churn.

PR #2118, opened 2026-09-07
feat(console): sign out from General settings

sanil-23:
## Summary

The console could sign a person in and never sign them out. A session ended only by expiring, by an admin revoking it from People, or by clearing cookies — so a shared or borrowed machine had no way to leave a company.

This adds an **Account** card to General settings, between Connection and Lifecycle: that card says where this console is pointed, this one says who it is pointed as, and carries **Sign out**.

Signing out revokes the session server-side first, while the credential still authenticates, then clears whatever this machine was carrying — a cross-origin token, or the desktop core's keychain entry via the already-registered-but-unused `oc_forget_device` — and marks the connection `unauthenticated`, which is the state `ConnectionConsole` already renders `Login` from. Ordering matters both ways: a console that forgot its token and *then* asked would be asking anonymously and the session would outlive the sign-out on the host, and a failed revoke reports and stays put rather than signing out locally anyway — a console on its sign-in screen while the host still honours the session it believes it revoked is the worse of the two states, and only one of them is something a person can see and retry from.

The card renders nothing on a company with no sign-in (`mode: "none"`, the desktop's default). The principal there is resolved from the request rather than from a session, so `/auth/me` answers with the local owner and an identity is there to draw — but `auth/logout` refuses with `auth_mode` and there is no sign-in screen to land on, so the control would be a button whose only outcome is an error.

**Deliberately does not touch running work.** A chat turn is a process-local `tokio::spawn` owned by the host and serialised on the per-company cycle mutex (`runtime/turn_sweep.rs`); the console is only a viewer over SSE. Sessions are per-user, so halting inference on sign-out would let one person closing their laptop stop work the rest of the company depends on. Pause, the emergency stop, and `POST …/workflows/runs/{id}/cancel` remain the controls that stop work. The SSE stream and its 30-second reconciliation timer are torn down on the way out, because `AppShell` unmounts when the connection flips — so there is no retry storm against a revoked session.

## API Or Behavior Changes

- New console affordance only. No new HTTP routes: `POST …/auth/logout` and `GET …/auth/config` already existed and are unchanged.
- New frontend exports: `forgetSession` (`connections/registry.ts`), `forgetSessionInCore` (`api/transport/desktop.ts`), and `AccountCard` (`views/SettingsView.tsx`, exported for its test the same way `LifecycleControls` is).
- Wires the existing `oc_forget_device` Tauri command, which was registered in `lib.rs` but called by nothing.

## Tests

No Rust changed, so the cargo lanes were not run for this diff.

- [x] `npm run typecheck` (`tsc -b`) — clean
- [x] `npm run typecheck:unit` — clean
- [x] `npm test` — `test/unit/settings-sign-out.test.ts` passes (4 cases: the identity is named; the card is absent in `none` mode; sign-out posts the logout and leaves the connection `unauthenticated` with the credential cleared; a refused logout leaves the local credential intact)

Also verified by hand against a local `opencompany serve`: signed in over a magic link, opened Settings → Sign out, landed on the sign-in screen; `auth/me` then answers 401 with no cookie left, and a reload stays signed out. Against a second host in `mode = "none"`, General settings renders normally with no Account card and no Sign out.

⚠️ **The `Console` job is red on this branch, and not because of this change.** `npm test` reports `4 failed | 492 passed` — 10 failures across `settings-lifecycle-reset-button`, `create-company-wallet-mode`, `create-company-dialog-preflight-race` and `connection-console-switch-known-status`. The new `settings-sign-out` test is in the 492 that pass, and all three typecheck steps pass.

Those 10 are pre-existing on `main`, from b8a3e2e97 ("gate company creation where every trigger asks, not per button"). That commit routes the picker's New-company button, the per-card Reset, the no-company trigger and Settings' Reset / Start clean through `canCreateCompanies()` — which is `!COMPANY_SWITCHING_HIDDEN && client.carriesPlatformBearer`, and `COMPANY_SWITCHING_HIDDEN` is `true` in `product-scope.ts`. So those controls no longer render, while the tests still assert that they do; the commit's own file list (`create-company-dialog.tsx`, `SettingsView.tsx`, `app-shell.tsx`) maps one-to-one onto the four failing files. It landed on 2026-09-04, three days before this branch was cut off `efecd6061`, which contains it.

I confirmed the isolation locally as well: reverting this diff against an unchanged dependency tree reproduces exactly the same 10 failures. This change touches `LifecycleControls`' file but not that component — it only adds `AccountCard` and its render site.

Happy to fix those 10 in a separate PR (they need either the gate relaxed for a non-platform client or the tests taught about it) and rebase this one on top, if that's the preference — I left it out of this diff rather than mixing an unrelated fix into a sign-out change.

## Documentation

No docs change. This adds a console control over existing routes; `docs/spec/runtime/users.md` describes the auth modes and session lifecycle, both of which are unchanged.

### DIFF
--- a/frontend/src/api/transport/desktop.ts
+++ b/frontend/src/api/transport/desktop.ts
@@ -140,6 +140,43 @@ export function adoptSessionIntoCore(id: string, session: string): Promise<void>
 }
 
 /**
+ * Drops this machine's stored session for a connection — the sign-out leg of
+ * {@link adoptSessionIntoCore}.
+ *
+ * Local only, and deliberately so: the host's own session record is revoked by
+ * `auth/logout` before this runs, and `oc_forget_device` is documented as not
+ * touching it. Signing out on this machine must not reach into another one.
+ *
+ * Sequenced behind whatever is parked under this id for the same reason the
+ * adoption is: a forget that overtook an in-flight `oc_connect` would be undone
+ * by the registration landing after it, leaving the keychain holding a session
+ * the person has already signed out of.
+ *
+ * Rejects on failure rather than swallowing it, again like the adoption: a
+ * credential that could not be cleared is something the person signing out has
+ * to be told about, not a surprise sign-in on the next launch. The parked
+ * promise is the settled-either-way one so nothing resurfaces.
+ */
+export function forgetSessionInCore(id: string): Promise<void> {
+  const desktop = tauriCore();
+  if (!desktop) return Promise.resolve();
+  const previous = registrations.get(id) ?? Promise.resolve();
+  const forgotten = previous.then(() =>
+    desktop.invoke<void>("oc_forget_device", { connectionId: id }),
+  );
+  registrations.set(
+    id,
+    forgotten.then(
+      () => undefined,
+      (error: unknown) => {
+        console.error(`[desktop] could not forget the session for ${id}`, error);
+      },
+    ),
+  );
+  return forgotten;
+}
+
+/**
  * Drops a host from the core.
  *
  * Sequenced after any registration still in flight. This and
--- a/frontend/src/connections/registry.ts
+++ b/frontend/src/connections/registry.ts
@@ -27,6 +27,7 @@ import type { Transport } from "@/api/transport";
 import {
   closeSshTunnel,
   forgetConnection,
+  forgetSessionInCore,
   openSshTunnel,
   adoptSessionIntoCore,
   registerConnection,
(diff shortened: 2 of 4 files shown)


sanil-23:
The red `Console` job here is not from this change — it is #2121, which fixes ten pre-existing failures on `main` from b8a3e2e97. This branch's own new test passes and all three typecheck steps are clean.

Once #2121 lands I'll rebase this one on top; it should go green with no changes of its own.

sanil-23:
CI finished. Both red jobs are pre-existing on `main`; nothing here is from this change.

**`Console`** — the ten unit failures from b8a3e2e97. #2121 fixes them, and its `Console` job is green (3m9s).

**`Console E2E` / `Console E2E (live brain)`** — three failures:

- `connections-authority.spec.ts:145` — `#company-credential` never visible
- `desktop-connections.spec.ts:420` — `no-connection-run-here` never visible
- `workflow-selection-persists.spec.ts:176` — host-switcher `Other` menuitem never appears

The identical three fail on `main` in run [34108301223](https://github.com/tinyhumansai/opencompany/actions/runs/34108301223) (a87899165, 10:22 UTC) and again in [34096823618](https://github.com/tinyhumansai/opencompany/actions/runs/34096823618) (08:13 UTC) — both before this branch was cut. Same specs, same line numbers. They also touch nothing this diff modifies: the Connections page, the desktop no-connection screen, and the host switcher, against a change confined to General settings' Account card plus two connection-registry helpers.

Green here: `Rust`, `Desktop`, `Gated host binary`, `Console E2E (first run)`, `Actions pinned`, `Changed areas`, `Markdown line cap`, CodeRabbit, tinysweeper.

Happy to take the E2E three in a third PR if that's wanted — they look like the same product-scope gating family as the unit ones, at least for the host-switcher and no-connection cases.

PR #2122, opened 2026-09-07
ci: pass remote-command values as data, not as shell syntax

oxoxDev:
## Summary

Three workflows built their remote command by interpolating values into a quoted string that
`ssh` hands to a shell running as root:

```
ssh root@${SSH_HOST} "NS='${NS}' ENV_ASSIGNMENTS='${ENV_ASSIGNMENTS}' bash -s" <<'REMOTE'
```

Putting the values in `env:` was right; re-interpolating them into the command string afterwards
undoes it. A value containing a quote character ends that quoting, and the remainder reaches the
remote shell as syntax rather than as data.

The ssh command line no longer carries any value at all — it is fixed as
`ssh root@"${SSH_HOST}" bash -s`. Values are base64-encoded locally, piped ahead of the unchanged
quoted heredoc, and decoded back into ordinary shell variables as the remote script's first
statements. Base64's alphabet contains no shell metacharacters, so the transport line cannot be
parsed as syntax, and the decode is a variable assignment via command substitution — never
`eval` — so whatever bytes come out are captured as data.

`ssh-keyscan` had the same shape in all three files and is fixed the same way.

**`actionlint` already detects this** — `SC2029`, "this expands on the client side". It was red on
all three files before this change and is clean on all three after. The linter was installed,
correct, and unread. A follow-up should make it a CI gate; that closes the class rather than
these three instances.

## API Or Behavior Changes

None. The remote scripts receive the same variables with the same values and run unchanged.

Preserved verbatim: the redaction `sed` pipeline, before/after env dumps, `kubectl set env` with
its deliberate `SC2086` word-splitting on `ENV_ASSIGNMENTS`, scale-to-1, rollout wait, the ECR
pull-secret refresh, the containerd pre-pull with its 10×2s retry, the ConfigMap-vs-`set env`
fallback including rollback-on-timeout, the fleet-wide refresh loop, and the landed-state report.

`REFRESH_ALL="${REFRESH_ALL:-false}"` still resolves client-side, in the same place as before.
Traced all three paths: input absent on a push-triggered run → `false`; explicit `false` → `false`;
explicit `true` → `true` and the fleet branch is taken.

On `deploy-staging.yml` specifically: no free-text input reaches that command today. `IMAGE`
derives from the ECR registry output and `github.sha`, three values are repo `env:` constants, and
`refresh_all_tenants` is `type: boolean`. It is included because the shape is an invitation — the
next `type: string` input added to this workflow would inherit the problem silently, and nothing
in the file would warn the author.

## Tests

These are workflow files; the Rust gates below do not exercise them, so they are marked N/A with
the verification that does apply stated instead.

- `actionlint .github/workflows/*.yml` — no `SC2029`/`SC2086` anywhere; the three changed files
  produce no findings at all. Only `ci.yml:759`'s pre-existing, unrelated `SC2034` remains.
- Repo-wide grep: `ssh root@` appears in exactly these three files, and the only value on any ssh
  command line is a single correctly-quoted `"${SSH_HOST}"`. No `${{ }}` expression remains inside
  any `run:` script body in the three files.
- Executable trace with `bash -s` standing in for the remote session: production-shaped values
  decode byte-for-byte; a value carrying quote and semicolon characters arrives as one inert
  literal string and no injected command runs. `ENV_ASSIGNMENTS`'s intentional word-split still
  produces exactly the argv the `kubectl` call expects.

- [ ] N/A: `cargo fmt --all -- --check` — no Rust changed
- [ ] N/A: `cargo clippy --all-targets -- -D warnings` — no Rust changed
- [ ] N/A: `cargo build --all-targets` — no Rust changed
- [ ] N/A: `cargo test` — no Rust changed

## Documentation

None needed. The mechanism is explained where it lives, in the workflow files.

## Summary by CodeRabbit

* **Chores**
  * Updated staging deployment and tenant maintenance workflows to pass configuration more securely during remote execution.
  * Improved handling of SSH host values and encoded deployment parameters.
  * Preserved existing deployment, diagnostics, and tenant environment update behavior.
  * Maintained output redaction for remote command results.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -227,17 +227,38 @@ jobs:
           ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
 
       - name: Add server to known hosts
-        run: ssh-keyscan -H ${{ secrets.SSH_HOST }} >> ~/.ssh/known_hosts
+        env:
+          SSH_HOST: ${{ secrets.SSH_HOST }}
+        run: ssh-keyscan -H "$SSH_HOST" >> ~/.ssh/known_hosts
 
       - name: Point the manager at the new tenant image
         env:
           SSH_HOST: ${{ secrets.SSH_HOST }}
           IMAGE: ${{ steps.ecr-login.outputs.registry }}/${{ env.ECR_REPOSITORY }}:staging-${{ github.sha }}
+          K8S_NAMESPACE: ${{ env.K8S_NAMESPACE }}
+          MANAGER_DEPLOY: ${{ env.MANAGER_DEPLOY }}
+          TENANT_IMAGE_CM: ${{ env.TENANT_IMAGE_CM }}
           REFRESH_ALL: ${{ inputs.refresh_all_tenants }}
         run: |
-          ssh root@${SSH_HOST} \
-            "IMAGE='${IMAGE}' K8S_NAMESPACE='${{ env.K8S_NAMESPACE }}' MANAGER_DEPLOY='${{ env.MANAGER_DEPLOY }}' TENANT_IMAGE_CM='${{ env.TENANT_IMAGE_CM }}' REFRESH_ALL='${REFRESH_ALL:-false}' bash -s" << 'EOF'
+          REFRESH_ALL="${REFRESH_ALL:-false}"
+          IMAGE_B64=$(printf '%s' "$IMAGE" | base64 | tr -d '\n')
+          K8S_NAMESPACE_B64=$(printf '%s' "$K8S_NAMESPACE" | base64 | tr -d '\n')
+          MANAGER_DEPLOY_B64=$(printf '%s' "$MANAGER_DEPLOY" | base64 | tr -d '\n')
+          TENANT_IMAGE_CM_B64=$(printf '%s' "$TENANT_IMAGE_CM" | base64 | tr -d '\n')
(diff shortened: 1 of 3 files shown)


oxoxDev:
Fixed in `a5236535d` — good catch, and it was a real gap rather than a nit.

Only stdout reached the masking `sed`. The remote script's individual `kubectl` calls each carry their own `2>&1`, which is what made this easy to miss: the common paths *were* redacted, so the leak was confined to anything that wrote to stderr without redirecting — a shell error on the remote, or ssh's own messages.

Both files now redirect at the session boundary (`bash -s 2>&1 |`), so the property holds for the session rather than per-command.

Demonstrated rather than assumed, with the same `sed` and a Bearer token on each stream:

```
--- without 2>&1 (the shape before this commit) ---
Bearer secretstderr99 on stderr      <- unmasked
Bearer *** on stdout

--- with 2>&1 ---
Bearer *** on stdout
Bearer *** on stderr
```

Not applied to `deploy-staging.yml`: it pipes to ssh with no redaction filter at all, so there is nothing there to bypass. Your scoping to the two files was right.

`actionlint` clean on both.

oxoxDev:
@coderabbitai review

The stderr redaction finding is fixed in `a5236535d`. Both existing reviews — the approval and the changes-requested — pin to `a56200a3e` and predate that commit, so neither has seen it.

PR #2129, opened 2026-09-07
fix(hivemind): ask a blind member for a deposit, not a position

senamakel:
Fixes the single biggest measured defect in the deliberation protocol, and bumps `tinyhivemind` to pick up the context-budget benchmark that measured it (tinyhumansai/tinyhivemind#25).

## The bug is one sentence in the prompt

```rust
Visibility::Blind => "You cannot yet see your peers' positions. Form your own first.",
```

That instructs every blind member to **advocate a position**. On any question where members hold correlated priors and one member holds the decisive fact — which is what a desk of specialists *is*, since `fleet_tech` holds machine facts nobody else does — it is fatal.

A traced episode, five seats, the truth is `#revert`:

```
planner    blind  !propose #revert     ← the one member holding the fact
critic     blind  !propose #shadow     ← decoy
archivist  blind  !propose #shadow     ← decoy
scout      blind  !propose #shadow     ← decoy
auditor    blind  !propose #shadow     ← decoy, and the room enters Commit
planner    full   !evidence #shadow ^3 The reading I hold rules this one out.
critic     commit !commit #shadow
→ converged on #shadow (but not correct)
```

A proposal counts as its own author's support in `tinyhivemind`, so four coinciding priors reach quorum **inside the blind round** — before the informed member has been able to say anything. The dissenting fact lands one turn later and the Commit phase cannot act on it.

The room is not converging. It is amplifying a shared error and calling the result agreement.

## The fix

Ask a blind member for a **deposit**, not a position:

> You cannot yet see your peers' positions. Put what you *know* on the floor — the fact, figure or reading you hold that the others may not — rather than the option you already favour. If your seat may deposit evidence, deposit it: the room can weigh a fact it has been shown, and cannot weigh one you kept while arguing from it.

## Measured, over 2000 seeded rooms

`vendor/tinyhivemind/crates/tinyhivemind-hive/examples/bench`, `--blind-evidence`:

| profile | advocate first | deposit first |
| --- | --- | --- |
| hidden — one member holds the fact | 16.2% | **66.6%** |
| uniform — everyone holds a noisy copy | 78.8% | 75.1% |

**+50.4 points** where one member knows something the others do not; **−3.7** where every reading is equally good. A desk of specialists is the first case by construction, which is why this is taken for every desk rather than made a knob.

For scale: on the hidden profile the *old* behaviour (16.2%) was worse than a single responder answering alone (35.9%). The deliberating room was actively worse than not deliberating.

## What was tried first and rejected

**Raising `quorum`.** If four coinciding priors carry the decoy, make the bar higher. It does not work — at unanimity the room simply stops deciding:

| quorum | decided | correct |
| --- | --- | --- |
| 3 | 100% | 16.0% |
| 4 | 96.5% | 19.9% |
| 5 (unanimity) | 31.1% | 10.1% |

The bar is not the problem; what the bar counts is.

## Submodule bump

`vendor/tinyhivemind` `63dfb22` → `2eba181`, which adds the context-budget model this measurement leaned on and an off-floor exchange arm. No API change; the crate builds and the whole suite passes against it.

`vendor/openhuman` is untouched at `b7f4e949`.

## Verification

`cargo fmt --all --check` clean · `clippy --all-targets --features openhuman,mcp -- -D warnings` clean · **7115 lib tests pass**. The existing blind-round test is strengthened to assert the deposit instruction rather than merely the sight sentence — it would have passed either way before.

## Summary by CodeRabbit

* **New Features**
  * Blind-round participants are now prompted to share unique facts and evidence before stating their preferred option.
  * Decision quality improves in scenarios where participants have differing information, while maintaining performance when their assessments are equally strong.

* **Documentation**
  * Updated Hive-mind guidance to explain the revised blind-round process, quorum limitations, and benchmark results.

* **Tests**
  * Added coverage confirming that blind-round prompts clearly explain peer positions are hidden and request known information first.


senamakel:
Verified against real models on the ladder router, not only in simulation. The blind round now opens with deposits:

```
route_planner    !propose #ops-day-1 restock VM-301 first…
fleet_tech       !evidence fleet read: zero faults anywhere, every chiller OK…
stock_controller !evidence warehouse read: all three affected lines are full (ENERGY-250 167, SANDWICH-1 …)
field_realist    !evidence fleet read: every machine serviced yesterday but VM-3…
route_planner    !propose #ops-day-1-2 don't reload SANDWICH-1 into VM-101 or VM-201 where it spoiled overnight…
```

Three of four seats deposited what they held, and the planner then **revised its plan** — the desk settled on `#ops-day-1-2`, the second proposal, not the first. That is the room updating on evidence rather than locking in its opening prior, which is the whole point of the change.

Both days of the run decided, none exhausted. Across earlier runs on the old prompt, 7 of 19 episodes spent their budget without deciding.

Every check green, including `Desktop` and `Console E2E (live brain)`, both of which were red on `main` earlier today and have since been fixed upstream.

PR #2132, opened 2026-09-07
test: pin route-principal authority matrix

oxoxDev:
## Outcome

Adds the Workstream A route × principal matrix as a Rust integration target, backed by fixed disposable-state principals and a CI assertion that names the target explicitly.

There is no production behavior change in this PR. The matrix preserves the intended authorization contract even where `main` is currently weaker; those known-red cells are isolated in ignored tests rather than normalized to the defect.

## Coverage and counting

- Source audit at `0056ddf0e`: 184 canonical ops route-method pairs across 142 unique scoped suffixes.
- Both scope forms produce 368 concrete route-method rows; two HMAC routes plus the OAuth callback bring the ops inventory to 371.
- Eight lifecycle/approval authority rows and the overlapping company-status GET bring the complete table to **380 concrete route-method rows across 295 concrete paths**.
- Seven principals per row produce **2,660 committed expectation cells**: 2,605 active and 55 intentionally ignored/red on `main`.
- The real `AdminScopedCompany` slice is **40 route-method pairs across 36 unique suffix paths**, or 80 concrete rows after expanding both address forms. The handoff's 28 was stale; this count comes from parsing the current router method chains, so routes, unique paths, and route-method pairs are kept distinct.
- Warm full-target runtime: **1.56 s wall clock** (`0.99 s` inside the test binary), well below the 45-second budget.

The anti-rot mechanism combines exact source path-set equality, exact runtime method sets from 405 `Allow` headers, anonymous existence checks, explicit feature-lane metadata, widening notes, and the committed 2,660-line snapshot. The source lexer rejects unparseable production registrations and uses exact counted fingerprints for the three intentional nonliteral registrations.

## Red on `main`

The ignored groups preserve **39 concrete route-method rows / 55 principal cells** that fail the intended contract today:

| Group | Rows / cells | Current behavior and intended boundary | Wait |
|---|---:|---|---|
| lifecycle + approval decisions | 8 / 10 | Members reach pause/resume/emergency and approval handlers (`200`, confirmation `400`, or absent-resource `404`) instead of 403; must-change-password admins reach the two `{id}` approval handlers instead of the password-change 403 | `fix/auth-role-dimension-on-authorize-address` |
| custom skill authoring | 2 / 2 | A member reaches body decoding (`415`) instead of admin 403 on both scope forms | `fix/skills-admin-and-bounded-write` |
| seven body-admin methods | 14 / 28 | Tenant-owner and platform principals receive 401 because authorization lives in body-only `require_admin`; the `AdminScopedCompany` contract permits those machine-admin principals | no branch assigned in handoff §7 |
| ledger define + retire | 4 / 4 | A member reaches handler validation/resource lookup (`415`/`404`) instead of admin 403 | no branch assigned in handoff §7 |
| teammate delete | 2 / 2 | A member reaches the absent teammate lookup (`404`) instead of admin 403 | no branch assigned in handoff §7 |
| workflow create/update/delete/run | 8 / 8 | A member reaches handler decoding/validation/resource lookup (`415`/`400`/`404`) instead of admin 403 | no branch assigned in handoff §7 |
| company status temp-password boundary | 1 / 1 | A must-change-password admin receives 200 instead of `403 password_change_required` | no branch assigned in handoff §7 |

Each ignored test and snapshot cell names its real waiting branch where §7 provides one. The remaining five groups explicitly say `none-assigned` rather than inventing branch names; owners/tracking branches are needed before those labels can be made more specific.

## Observed red proofs

All proof edits were copied to `/tmp`, never committed or stashed, restored before continuing, and verified byte-identical afterward (`cmp=0`, source `git diff=0`). The target docblock records the same observations.

1. Removed `clear_policy`'s `require_admin`: member `DELETE /policy` flipped from expected 403 to observed 200 on both address forms; matrix exit 101.
2. Added `scoped("/matrix-canary", ...)`: the source path-set gate named `/matrix-canary` as unexpected; exit 101.
3. Added `.delete(get_activation)` to the existing `/activation` method router: both forms reported runtime `{DELETE, GET}` versus declared `{GET}`; exit 101.

The AdminScoped skeleton was exercised first against the source-derived 40-pair set before the full table was expanded.

## Gates

Run individually and unpiped:

- `cargo fmt --check` — exit 0
- `cargo clippy --locked --no-deps --features openhuman --all-targets -- -D warnings` — exit 0
- `cargo test --features openhuman` — exit 0 (`7,112` library tests passed; all selected integration targets passed)

## Not expressible as one route-level cell

- `POST {scope}/team`: privileged budget/billing/tool fields are conditional on payload content.
- `PATCH {scope}/team/{agent_id}`: tool, model, and harness edits are conditional on payload content and require a real teammate; ordinary profile fields remain member-editable.

Those two rows carry explicit widening notes, but their field-sensitive boundaries remain owned by handler-level tests. Per the requested lane discipline, this PR ran only the `openhuman` feature lane and did not run a union build.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive authorization coverage for operations routes across anonymous, member, admin, tenant, platform, and password-change-required access levels.
  * Added checks to verify route declarations, allowed HTTP methods, and expected authorization responses.
  * Added fixed test authentication scenarios for tenant and platform principals.
  * Added automated CI execution for the authorization test suite, ensuring missing or disabled test targets fail clearly.
  * Documented selected known authorization cases as pending follow-up fixes.

### DIFF
--- a/scripts/ci/assert-auth-matrix.sh
+++ b/scripts/ci/assert-auth-matrix.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "$0")/../.." && pwd)"
+cd "$repo_root"
+
+# Naming the target explicitly makes a missing/disabled integration target an
+# error instead of Cargo's successful "zero tests selected" failure mode.
+cargo test --locked --features openhuman --test auth_matrix
--- a/src/server/test_support.rs
+++ b/src/server/test_support.rs
@@ -5,9 +5,19 @@
 //! the one place that mints it, so the shape of "an authenticated request" is
 //! stated once rather than in every test module.
 
+// Integration targets include this source directly, so the library-test crate
+// cannot see every helper consumer when it performs dead-code analysis.
+#![allow(dead_code)]
+
+use std::collections::HashSet;
+use std::sync::Arc;
+
 use crate::AppState;
 use crate::ports::types::CompanyId;
 use crate::ports::{SessionRecord, UserRecord, UserRole, UserStatus, generate_id, now_millis};
+use crate::server::platform_auth::{
+    PlatformAuthConfig, PlatformClaims, PlatformVerifier, StaticPlatformVerifier,
+};
 use crate::server::users::cookie::session_cookie_name;
 use crate::server::users::token::{OsTokens, mint_session_token, sha256_hex};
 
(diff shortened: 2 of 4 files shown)


oxoxDev:
Both findings were right, and the first was blocking.

**The gate guarded nothing.** `assert-auth-matrix.sh` existed precisely so the target name lives in one place — a bare `cargo test` reports success when it selects no tests at all, which is how a disabled integration target passes as green. But no lane invoked the script. Now wired into the `rust-gated` lane, which already carries the `openhuman` feature the target needs, and `scripts/ci/**` was already in that lane's path filter.

**The provisioning finding turned out to be one level deeper than reported.** `external_authority_router_files_have_no_unclassified_paths` asserts which paths *exist* in `provision.rs`, not that each one is exercised. Four of its eight — `POST /companies`, `GET /companies/provisioning`, `suspend`, `archive` — are `PlatformScope` and carry no matrix row, so no principal ever probes them, and the check stayed green throughout. A source-inventory assertion that reads as coverage is the same shape as the defect this whole harness exists to catch.

Giving them real rows needs an access class the matrix does not have — every `Access` variant today resolves to a company-scoped or admin verdict, and `PlatformScope` is neither — so that is a follow-up rather than a line item here. What this commit does instead is stop the hole being silent: the four are named as the set no principal probes, so a ninth provisioning route, or a fix that gives any of them a row, fails the check rather than quietly joining a set nobody counts.

Red-proved by adding `/pause` to that set: `provisioning paths no principal probes set drifted; missing=["/api/v1/companies/{id}/pause"]`. Reverted, 8 passed.

Gates: `cargo fmt --check` 0, `scripts/ci/assert-auth-matrix.sh` 0 — 8 passed, 7 ignored (the pre-existing named-wait cells).

The suspend-from-suspended boundary and the resume-while-suspended case both need that same access class, so they ride with the follow-up.

PR #2143, opened 2026-09-08
test: drain approval stubs before browser teardown

oxoxDev:
Approval-verdict browser assertions could finish while their company-status stub was still reading a response. Browser teardown then disposed that response and randomly failed otherwise-correct Skip, Cancel, or Retry cases. Track this spec's five route handlers per page, await their complete fetch/read/fulfill callbacks, and only then unregister them and allow context teardown. Callback errors still fail the test; verdict assertions and production behavior are unchanged.

The drain deliberately runs while handlers remain registered: `unrouteAll({ behavior: "wait" })` alone removed Playwright's handler list early and reproduced `Route is already handled!` when another callback finished and disabled interception under a pending fulfill (5/200 cases). This PR owns only `frontend/test/e2e/approval-blocker-verdicts.spec.ts`.

Observed reproduction on fixed main `b4cab3ea3e0d458ae301372a82d9ae2bd4d295f0`:

- Twenty complete file repetitions, one worker, no retries and no trace: **10 failed / 190 passed**, exit **1**, 1.8 minutes. Nine of the twenty repetitions failed. Failures occurred at positions 3, 28, 37, 45, 67, 103, 113, 128, 147, and 148: two-step Skip (3), single Skip (1), Cancel (3), Retry (3). Every failure pointed to the company-status stub's `response.json()` and disposed response/context teardown, rather than a particular verdict or fixed run position.
- Thirty isolated Retry cases with tracing: **30 passed**, exit **0**, 25.6 seconds. The trace showed the company fetch completing 1.6 ms after After Hooks began; the context remained alive another 65 ms with tracing enabled. This is consistent with trace collection masking the teardown window. It is not evidence that the original flake was fixed.
- These observations were reported before implementing the fix. `captureResolve` uses `route.fulfill()` throughout; no cross-spec state explanation or response-reading resolve mechanism is assumed.

Validation on signed commit `67ba85877` (each command unpiped, each command's own exit):

| Command | Exit | Result |
| --- | --- | --- |
| `npm run typecheck` | 0 | passed |
| `npm run typecheck:unit` | 0 | passed |
| `npm run typecheck:e2e` | 0 | passed |
| `npx vitest run` | 0 | 500 files, 4,657 tests; 84.21 seconds |
| Default-host command below | 0 | 200/200 cases; 20/20 repetitions; 1.6 minutes |
| Live-brain command below | 0 | 200/200 cases; 20/20 repetitions; 1.1 minutes |

```bash
PW_SKIP_CONSOLE_BUILD=1 PW_HOST_BINARY=/Users/oxoxdev/Work/Zerolend/tinyhumans.ai/opencompany-approval-flake/target/e2e/bin/opencompany-default-b4cab3ea3 npx playwright test test/e2e/approval-blocker-verdicts.spec.ts --repeat-each=20 --retries=0 --reporter=list

PW_LIVE_BRAIN=1 PW_MOCK_BRAIN_BIND=127.0.0.1:17541 PW_MCP_FIXTURE_BIND=127.0.0.1:17542 PW_SKIP_CONSOLE_BUILD=1 PW_HOST_BINARY=/Users/oxoxdev/Work/Zerolend/tinyhumans.ai/opencompany-approval-flake/target/e2e/bin/opencompany-live-b4cab3ea3 npx playwright test test/e2e/approval-blocker-verdicts.spec.ts --repeat-each=20 --retries=0 --reporter=list
```

Both host binaries were built from the fixed production source above: `cargo build --locked --bin opencompany --target-dir ../opencompany-authmatrix/target` and the same command with `--features openhuman,mcp`, each exit 0. No union build. `npm ci`, `npm run build`, and `npm run build:pages-sdk` also exited 0. Hosts and sessions used only this worktree's disposable `target/e2e` root; no credential or persistent company fixture was changed. No stash was used.

Setup-only failures are excluded from reproduction counts: two premature build attempts before recursive submodule initialization completed exited 101; one sandboxed host bind was refused and exited 1 before any test; an anchored isolated-test filter selected zero tests and exited 1. The reported browser loops ran with local host and browser permissions, and each executed the stated nonzero count.

Main currently has no branch protection: `/branches/main/protection` returns 404 and `/rulesets` returns `[]`. That is why reliable, observed-red gates matter here. Repository settings belong to the maintainer and were not changed. Draft only; no production behavior change.

### DIFF
--- a/frontend/test/e2e/approval-blocker-verdicts.spec.ts
+++ b/frontend/test/e2e/approval-blocker-verdicts.spec.ts
@@ -1,4 +1,4 @@
-import { expect, test, type Page } from "@playwright/test";
+import { expect, test, type Page, type Route } from "@playwright/test";
 
 /**
  * **Issue #2028 — reachability, which only a real click can prove.**
@@ -33,10 +33,10 @@ import { expect, test, type Page } from "@playwright/test";
  * still in flight; context teardown disposed the response underneath it.
  *
  * Thirty isolated Retry runs with tracing passed: a trace showed the company
- * fetch finishing 1.6 ms AFTER After Hooks began, then trace collection kept
- * the context alive another 65 ms. Tracing changed the timing, not the lifetime
- * contract. Drain this spec's routes before context teardown, without ignoring
- * callback errors or adding sleeps to the verdict assertions.
+ * fetch finishing 1.6 ms AFTER After Hooks began; with tracing, the context
+ * stayed alive another 65 ms. This is consistent with tracing masking the
+ * teardown race. Drain this spec's routes before context teardown, without
+ * ignoring callback errors or adding sleeps to the verdict assertions.
  */
 
 const BLOCKER_ID = "e2e-2028-blocker";
@@ -123,7 +123,37 @@ test.beforeEach(async ({ page }) => {
   });
 });
 
+const pendingRoutes = new WeakMap<Page, Set<Promise<void>>>();
+
+async function stubRoute(
+  page: Page,
+  matches: (url: URL) => boolean,
+  handle: (route: Route) => Promise<void>,
+) {
+  let pending = pendingRoutes.get(page);
+  if (!pending) {
+    pending = new Set();
+    pendingRoutes.set(page, pending);
+  }
+  const calls = pending;
+  await page.route(matches, async (route) => {
+    const call = handle(route);
+    calls.add(call);
+    try {
+      await call;
+    } finally {
+      calls.delete(call);
+    }
+  });
+}
+
 test.afterEach(async ({ page }) => {
+  // Drain while handlers remain registered. unrouteAll(wait) alone failed
+  // 5/200 cases: Playwright removed its handler list before waiting,
+  // so one callback finishing could disable interception under another's
+  // pending fulfill, which then failed with "Route is already handled!".
+  const pending = pendingRoutes.get(page);
+  while (pending?.size) await Promise.all(pending);
   await page.unrouteAll({ behavior: "wait" });
 });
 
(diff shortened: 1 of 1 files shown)


oxoxDev:
Verified independently: I reproduced the failure on a fixed base here, then re-ran the same command against this branch.

**Baseline** — the spec restored to its pre-fix content in this same worktree, same host binary, `--repeat-each=20 --retries=0 --workers=1`:

```
3 failed
  approval-blocker-verdicts.spec.ts:411 › a bulk decision never sweeps a question in with the gated calls   (×2)
  approval-blocker-verdicts.spec.ts:375 › Retry sends a retry
197 passed

Error: apiResponse.json: Response has been disposed
  132 |     const response = await route.fetch();
  134 |     const body = await response.json();
```

**This branch**, same command: `200 passed`, exit 0.

3/200 against 10/200 elsewhere is the same mechanism at a different machine's timing, and the failing line is the one this PR guards. `typecheck`, `typecheck:unit`, `typecheck:e2e` each 0, run unpiped.

Two things worth recording so the next person does not re-derive them:

**The earlier exclusion was wrong, and worth knowing why.** This mechanism was ruled out once by inspecting `captureResolve`, which genuinely does use `route.fulfill()` and reads no response. The offending handler is a different one in the same file — the company-read stub in `stubQueue`. Ruling a mechanism out by checking the handler the failing test is *named* after, rather than every registered route, is the error to avoid.

**The lane has a second shape this does not cover.** `chat-attachment-workspace-file.spec.ts:212` fails with `response.body: Protocol error (Network.getResponseBody): No data found for resource with given identifier` — a `page.waitForResponse()` body read after the resource is evicted. Same family, different call, and that spec registers no routes at all. Four further specs (`approval-timeout-honesty` — nine route callsites, the largest exposure — `chat-detached-post-failure`, `composio-provider-detail`, `connections-native-not-offered`) do `route.fetch()` then `response.json()` with no drain, so they carry this shape unguarded. The drain belongs in a shared `test/e2e` helper adopted by all five rather than copied four more times; that is a follow-up, not a reason to hold this.

PR #2151, opened 2026-09-08
ci: run the route authority matrix, and name the routes nothing probes

oxoxDev:
## Summary

Recovers two commits that did not land with #2132. That PR merged at `8cff762a5` while its branch head was `56b8b4fa9` — GitHub held the pull request's head stale for roughly half an hour and merged the older tip, so the last two commits were left behind.

Both were written in response to review on #2132, and both are still needed.

## API Or Behavior Changes

None.

## Tests

`scripts/ci/assert-auth-matrix.sh` — 8 passed, 0 failed, 7 ignored. `cargo fmt --check` clean.

**Wiring the gate to something that runs it.** `scripts/ci/assert-auth-matrix.sh` is on `main` and no workflow calls it, so it guards nothing. The matrix itself does execute today, via the broad `cargo test --locked --features openhuman --tests` in the `rust-gated` lane — so this is not a coverage hole. What the script exists for is naming the target explicitly: a bare `cargo test` reports success when it selects no tests at all, which is how a disabled or renamed integration target passes as green. Calling it from the lane that already carries the `openhuman` feature restores that guard. `scripts/ci/**` is already in that lane's path filter.

**Naming the routes no principal probes.** `external_authority_router_files_have_no_unclassified_paths` asserts which paths exist in `provision.rs`, not that each one is exercised. Four of its eight — `POST /companies`, `GET /companies/provisioning`, `suspend`, `archive` — are `PlatformScope`, carry no matrix row, and no principal probes them, while the check stays green. A source-inventory assertion that reads as coverage is the shape this harness exists to catch.

Giving them real rows needs an access class the matrix does not have: every `Access` variant resolves to a company-scoped or admin verdict, and `PlatformScope` is neither. That is a follow-up. This assertion stops the hole being silent in the meantime — a ninth provisioning route, or a fix granting any of these four a row, fails the check rather than joining a set nobody counts.

Red-proved by adding `/pause` to the named set: `provisioning paths no principal probes set drifted; missing=["/api/v1/companies/{id}/pause"]`. Reverted, 8 passed.

## Documentation

None.

## Summary by CodeRabbit

* **Tests**
  * Strengthened route-authority validation to ensure all provisioning paths and supported HTTP methods are explicitly accounted for.
  * Added coverage for routes that support multiple HTTP methods, improving detection of incomplete authority mappings.
* **Chores**
  * Added an automated CI check that runs route-authority validation during gated builds, helping prevent routing authorization gaps from reaching release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1042,6 +1042,14 @@ jobs:
       - name: Clippy (openhuman, tinymemory)
         run: cargo clippy --locked --no-deps --features openhuman --all-targets -- -D warnings
 
+      # The route-authority matrix names its target explicitly, because a
+      # bare `cargo test` reports success when it selects no tests at all —
+      # which is how a disabled integration target passes as green. The
+      # script exists so the target name lives in one place; a lane that
+      # inlines the command instead drifts the moment the target is renamed.
+      - name: Route authority matrix
+        run: scripts/ci/assert-auth-matrix.sh
+
       # Issue #1524 (PR 1). `pub mod memory` is gated on `tinymemory`
       # (`store/mod.rs`), and neither strict lane enabled that gate: the
       # default-features clippy lints a tree where the module does not exist,
--- a/tests/auth_matrix.rs
+++ b/tests/auth_matrix.rs
@@ -1431,6 +1431,34 @@ fn external_authority_router_files_have_no_unclassified_paths() {
         ]),
         &provision.direct,
     );
+
+    // Knowing a path exists is not the same as exercising it. These four are
+    // `PlatformScope` and the matrix has no platform access class to express
+    // them, so they carry no row and no principal ever probes them. Naming
+    // them here is what stops that being silent: a ninth provisioning route,
+    // or a fix that gives these rows, fails this assertion rather than
+    // quietly joining a set nobody checks.
+    let unexercised: BTreeSet<String> = provision
+        .direct
+        .iter()
+        .filter(|path| {
+            !EXTERNAL_AUTHORITY_ROUTES
+                .iter()
+                .any(|route| route.path == path.as_str())
+        })
+        .cloned()
+        .collect();
+    assert_set_eq(
+        "provisioning paths no principal probes",
+        &string_set(&[
+            "/api/v1/companies",
+            "/api/v1/companies/provisioning",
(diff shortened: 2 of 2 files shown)


oxoxDev:
Both reviews made the same point and it was right: the check was keyed on path alone, so a path already in the inventory could gain a second method and the set would stay green while that method went unprobed.

The scanner now records the verb alongside the path — `.route("/p", post(h))` yields `POST /p` — and the assertion compares method-and-path pairs against `EXTERNAL_AUTHORITY_ROUTES`. An unreadable verb is recorded as `?` rather than skipped, so it fails loudly instead of disappearing from the set.

Red-proved with the reviewer's own example: adding `DELETE /api/v1/companies/{id}/archive` beside the existing POST now fails with `set drifted; missing=["DELETE /api/v1/companies/{id}/archive"]`. Before this commit that addition was invisible.

I also populated the field in the production scanner rather than defaulting it to an empty set there — an empty collection that reads as "no methods found" is the same silent-green shape this whole check exists to catch.

`cargo fmt --check` 0, `scripts/ci/assert-auth-matrix.sh` 0 — 8 passed, 7 ignored.

oxoxDev:
Right again, and it was already biting rather than hypothetical.

Axum's chained form registers two methods on one path, and reading the identifier at a fixed offset after the path recorded the first and dropped the rest. Three routers already use it today — `a2a.rs` (`post(a2a_task).get(agent_card)`), `feedback.rs` (`post(submit).get(list)`) and `setup.rs` (`get(read).post(apply)`) — so the production scanner I touched in the previous commit was under-reporting those three from the moment I wrote it. `provision.rs` happens not to use the idiom, which is the only reason the provisioning assertion still held.

Now walks the whole `.route(...)` call, collecting every verb identifier at depth 1, so a chain of any length is recorded in full. A call that yields no recognisable verb records `?` rather than nothing, so it fails loudly instead of shrinking the set.

Added `chained_route_verbs_are_all_recorded` as a permanent test rather than checking it by hand: it asserts `setup.rs`'s chained pair yields both `GET` and `POST`. Red-proved by degrading the walker to first-verb-only — `must yield both verbs, got {"GET /api/v1/setup", …}`, with the POST absent. Restored, 9 passed.

`cargo fmt --check` 0, full target 9 passed / 7 ignored, rc=0.

Issue #1734, opened 2026-08-25
Chat presents the echo brain's reply as a teammate's, with no sign inference is unconfigured


graycyrus:
## What happens

Send a message in any channel on an instance with no inference configured. A teammate replies:

> **Strategy desk**  14:36
> You said: yo

That is `EchoBrain` — the offline fallback, whose own doc comment says it *"echoes every operator message back"*. It renders in the transcript with the same avatar, name, timestamp and bubble as a real reply. Nothing distinguishes it.

## Why this is a correctness bug, not polish

The product knows this reply is not cognition. Its own tests assert on the string as the marker for a degraded runtime:

```rust
// src/server/operator.rs:4175
assert_ne!(text, "You said: hi", "still routing through the echo brain");
```

So the runtime can tell, the test suite can tell, and the only party who cannot tell is the operator — who concludes the product is stupid rather than unconfigured.

## Root cause

`select_hosted_or_echo` (`src/runtime/builder.rs:3518`) falls through to the echo brain whenever hosted mode has no credential:

```rust
(BrainMode::Hosted, Some(credential)) => …HostedMedullaBrain…
(BrainMode::Sidecar, _)               => …SidecarBrain…
_                                     => Arc::new(EchoBrain::new())
```

Degrading rather than failing is the right call — a company whose brain is unreachable should still boot. The bug is that the degradation is **silent all the way to the UI**.

## What makes this worse than it looks

The remedy is one settings page away. `Settings → Inference` exists (`frontend/src/views/SettingsSection.tsx:129`), backed by `POST …/inference/test` and `…/inference/restart` (`src/server/ops/inference.rs:77-78`). So this is not "unsupported" — it is **trivially fixable and nobody says so**.

Confirmed on a desktop build at `d31e532f`: the data dir contains no inference config, chat answers with echoes, and no surface anywhere mentions it.

## There is already a house pattern for this

The MCP path says exactly the right thing when a capability is present but inert:

> "Added {name}. It is stored, but no teammate can call it until this deployment is rebuilt with the MCP bridge."
> — `frontend/src/lib/mcp-bridge.ts:57`

And `frontend/src/api/setup.ts:58`: *"They exist so the flow can say 'not in this build' instead of offering a switch that does nothing."*

The brain path never got that treatment.

## Proposed behaviour

Chat should be honest before the first echo is ever sent, and route the operator to the fix:

[...]

PR #2160, opened 2026-09-08
test(console): pin the in-flight run bar, and name the Operator feed's real author

graycyrus:
Follow-up to #2159, which **merged at `f9324f5` while its review cycle was still
running** — so the two commits that answer the review are not in `main`. This
carries them, rebased onto `main` @ `b2c8f0fe7`. No behaviour change: one test
and three comments.

## 1. The in-flight run bar was pinned by nothing

Both reviewers on #2159 caught the spec claiming the banner sits directly above
the composer with nothing between them. `InflightRunBar` renders between the two
whenever a run is in flight — deliberately, outside the read-only branch,
because a channel nobody may post in is still a place the company's runs are
visible and stopping one is not posting. #2159 fixed the prose (f9324f5). This
is the assertion half, which CodeRabbit asked for and which did not land.

The reason it was never caught is worth naming: the adjacency assertion in
`chat-readonly-composer.test.ts` — `expect(kids.indexOf(composerRoot)).toBe(kids.indexOf(strip) + 1)`,
from #1984 — passed only because **no test in the file ever passed
`inflightRuns`**, and `ChatView` gates the bar on that prop being *defined*. It
pinned a layout no production shell renders.

So:

- the harness takes an in-flight run (`INFLIGHT_RUN`, opted into per test);
- the idle case additionally asserts the bar is **absent**, so "adjacent" is
  scoped to the condition it is true in rather than reading as a universal claim;
- a new case pins the real order: **banner → run bar → composer**.

Mutation-proved by moving `InflightRunBar` above the banner in `ChatView` and
re-running, then reverting:

```
× lets the in-flight run bar come between it and the composer
  → AssertionError: expected 3 to be less than 2
```

## 2. `#Operator` rows are not under a roster teammate — I had this wrong

codex, P2 on #2159, and correct. Checked against the code rather than taken on
trust:

| Fact | Where |
|---|---|
| Reports are journaled under `WORKFLOW_REPLY_AUTHOR = "workflow-report"` | `src/runtime/channel.rs:52` |
| The owner fallback overrides it to `"owner-fallback-report"` | `src/runtime/channel.rs:73` |
| The read path projects `channel: agent_id` | `src/server/chat_history.rs:573` |
| `senderOf` titleizes an unmatched non-company-voice channel | `frontend/src/views/chat/model.ts:1070-1081` |

So those rows read **"Workflow Report"** and **"Owner Fallback Report"**, with no
roster avatar. #2159's source comment, spec paragraph and test docstring all said
"under a teammate's name and avatar". They now say what is actually there.

**The conclusion is unchanged, and stronger.** `MessageView::project` sets
`by_person: false` on an `AgentReply` *"whichever brain produced it"*
(`chat_history.rs:580`), so `echoMarkerFor` marks every one of those rows — the
pill on a workflow report reads *"Workflow Report did not write this. No agent
harness is available on this host, so the offline echo brain answered instead."*
Without the strip #2159 restored, the reader is left with a "Placeholder" pill
against a name that is not a person, on a feed that takes no replies, and
nothing anywhere saying what did write it. There is nobody to ask even in
principle.

## Deliberately not here

Codex's alternative remedy — show the banner only when the feed contains an
actual echo-authored row — is declined and the reasoning is on
[#2159's thread](https://github.com/tinyhumansai/opencompany/pull/2159#discussion_r3959831478).
Short version: `ChatMessage` carries no provenance, which is why #1734 chose a
company-level state; a contents-gated banner would be a guess presented as a
fact. Report-specific copy for the read-only feed is worth doing and is a copy
change with its own review surface, not a rider on this.

## Verification

```
$ npm run typecheck        # tsc -b --noEmit             → clean
$ npm run typecheck:unit   # tsc -p tsconfig.unit.json   → clean
$ npm run typecheck:e2e    # tsc -p tsconfig.e2e.json    → clean
$ npx vitest run           # Test Files 513 passed · Tests 4820 passed
$ scripts/ci/assert-design-tokens.sh
✓ design tokens: no arbitrary sizes, no palette colours, no stray hex
```

All four run on this branch, on `main` @ `b2c8f0fe7`. No browser pass: nothing
here renders differently — the diff is one unit test and three comment blocks.
#2159's browser evidence, light and dark against a default-feature host, is
[here](https://github.com/tinyhumansai/opencompany/pull/2159#issuecomment-5587939893).

### DIFF
--- a/docs/spec/runtime/ports-cognition.md
+++ b/docs/spec/runtime/ports-cognition.md
@@ -149,7 +149,10 @@ composer rather than being glued to it: `InflightRunBar` renders between the
 two whenever a run is in flight, since stopping a run is not posting. And it is
 rendered on a read-only channel too, where there is no composer at all: the
 sentence is about attribution, not about sending, and `#Operator` is precisely a
-feed of company-authored reports rendered under a teammate's name. `ChatMessage`
+feed of company-authored reports, marked like any other company-side row but
+rendered under the reserved authors `workflow-report` and
+`owner-fallback-report` — names that belong to no person, so the reader has
+nobody to ask even in principle. `ChatMessage`
 carries no provenance, so a company-level state is the only shape that answer
 has — see `MessageRow`'s `cognition` prop for why marking beats suppressing.
 The same state reaches `ThreadPanel`, because a reply read inside a thread is
--- a/frontend/src/views/ChatView.tsx
+++ b/frontend/src/views/ChatView.tsx
@@ -2714,15 +2714,23 @@ export function ChatView({
                     says the replies in this conversation come from the echo brain
                     rather than the teammate they appear under, which is a claim about
                     the messages already on screen. `readOnly` is
-                    `Boolean(channel?.system)`, i.e. the `#Operator` feed, whose
-                    workflow reports are rendered under a teammate's name and avatar
-                    and are stub output whenever the company has no working model.
-                    Those rows carry only `EchoPlaceholder` — a non-focusable `<span>`
-                    whose explanation lives in a `title`, so it reaches neither
-                    keyboard, touch nor screen reader. And a feed nobody can reply to
-                    is where the reader is least able to test the attribution by
-                    asking, so it is the last place to drop the only visible statement
-                    of it.
+                    `Boolean(channel?.system)`, i.e. the `#Operator` feed.
+
+                    Its rows are NOT under a roster teammate, and the difference
+                    matters (codex review on #2159). `DurableOperatorChannel` journals
+                    them under the reserved authors `workflow-report` and
+                    `owner-fallback-report` (`runtime/channel.rs`), which `senderOf`
+                    titleizes into "Workflow Report" and "Owner Fallback Report" —
+                    author lines naming no person at all. That makes the case for the
+                    strip stronger, not weaker: `MessageRow` still marks every one of
+                    those rows, because `project` sets `by_person: false` on an
+                    `AgentReply` whichever brain produced it, and the marker they get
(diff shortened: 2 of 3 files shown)


graycyrus:
@coderabbitai review

graycyrus:
## Status

CI is **green by head SHA** `bae91c0f8` — verified through the runs and jobs API rather than `gh pr checks`:

```
$ gh api "repos/tinyhumansai/opencompany/actions/runs?head_sha=bae91c0f8..."
34249542511 CI completed success

11 success · 3 skipped · 0 failure · 0 pending
  Actions pinned · Changed areas · Console · Console (current Node, advisory)
  Console E2E · Console E2E (first run) · Console E2E (live brain) · Desktop
  Gated host binary · Markdown line cap · Rust
  (skipped: Rust (mail), Rust (mongodb), Rust (openhuman, tinymemory))
```

`Console E2E (first run)` failed once on attempt 1 and passed on attempt 2. The
failing step was **`Download the host binary`** — the artifact fetch, before
Chromium was installed or a single spec ran (steps 6-8 all `skipped`). Nothing
in this diff can reach that step, and the `Gated host binary` job that produces
the artifact was green on both attempts. Recorded rather than glossed: a re-run
that turns a lane green is only evidence if the failure was outside the change,
and here it was.

**CodeRabbit has not reviewed this PR.** It answered *"Review limit reached —
you've used all 2 included reviews currently available"*; both went to #2159. A
re-request is queued for when the window reopens. Flagging it because an absent
CodeRabbit review reads exactly like a clean one.

graycyrus:
@coderabbitai review

Issue #412, opened 2026-08-05
chat: leaving and returning always lands on the first channel, not the one you were reading


oxoxDev:
## What the operator sees

Leave the Chat tab and come back, and you are always in the **engineering** channel — whichever channel you were actually reading. Every return trip costs a re-navigation, and a conversation you were mid-way through is not where you left it.

## Why it happens

The chat view resolves its channel from the hash's second segment. Navigating away and back re-enters without one, so resolution falls through to the first channel in the list — which on this company's board is engineering. It is not remembering anything; it is starting over each time and landing on whatever sorts first.

The shell already tracks the last-viewed channel — that mechanism was added so unread counts could clear correctly, and so a system line could be addressed to the channel the operator was actually looking at. Nothing currently uses it for re-entry.

## What the fix should do

Return the operator to the channel they were last reading, and reflect it in the URL so the state is shareable and survives a reload rather than living only in memory.

Two cases worth deciding rather than assuming:

- **A deep link must still win.** Someone opening a link to a specific channel should land there, not on their own last-viewed one.
- **A last-viewed channel that no longer exists** — a desk removed, or a teammate whose direct message is gone — must fall back cleanly and say so, rather than silently landing somewhere else. The unknown-channel notice added in #370 is the right precedent.

## How to verify

- Read a non-first channel, navigate to Tasks, return to Chat, and land back in that channel.
- The same after a page reload.
- A deep link to another channel still wins over the remembered one.
- A remembered channel that has since disappeared falls back with the notice, not silently.

## Related

Observed on a staging tenant, 2026-08-06.
#370 established the unknown-channel notice and the loading gate this should reuse.
#378 added the last-viewed tracking that already exists but is not consulted on re-entry.
#364 (chat transcripts are console-local) is adjacent and unfixed; this should not depend on it.

PR #2162, opened 2026-09-08
fix(console): mark a failed send on its own row, and route which conversation is open

graycyrus:
Split 2 of 2 from #2052 (`fix/chat-delivery-cluster`), rebased onto current `upstream/main`. This is the **console side**; the host half is #2161. #2052 was a 21-commit, 2487-line, seven-bug cluster whose review churned; splitting it is the point.

## What a person hit

1. **A failed send looked identical to a successful one.** Same avatar, same timestamp, nothing red — just a grey sibling note that scrolled away, after the composer had already cleared. The only copy of what you wrote was pixels.
2. **You wrote in `#general` and it posted into a DM.** The console never *decided* which conversation was open; it re-derived it from data arriving asynchronously, and your draft went with it when the answer changed.
3. **Two live replies with identical text — the second never appeared.**
4. **The typing indicator appeared in an unrelated thread** while the channel actually working showed nothing.

### 1. `sendFailed` is a property of the message, not a line beside it

The failure used to be a separate `system` bubble appended underneath. A sibling line scrolls away from the row it describes, and a long enough message pushes it off screen entirely — leaving something that reads as sent and was not.

`ChatMessage.sendFailed` carries the reason, so **no renderer can draw the bubble without seeing it**. The text is muted, and a `role="status"` notice with a **Retry** button rides the row. `!== undefined`, not truthy: `httpError` keeps an empty `error: ""` envelope as-is, and a truthy check would silently hide the styling, the notice and the Retry for exactly that response.

Retry replays the **wire payload**, kept in a ref keyed by the optimistic id — not rebuilt from the rendered row. Two of the five fields cannot be recovered from the bubble: the wire `mentions` carry targets the rendered chips do not, and an attachment is a workspace node reference rather than the projection the row shows. A Retry rebuilt from the message would quietly send a *different* message — chip-less, or without its file.

It is **manual**. A throw is ambiguous — the host may have journaled the message before the request died — so an automatic resend risks posting the same instruction twice. Whether that is worth it is the operator's call, which is what a button is. `retrySend` no-ops while a POST is in flight, checked *before* the payload and the failed row are consumed, so a mid-send click cannot drop the only copy of the retry.

Both renderers show it. `ThreadPanel`'s `Line` is a second renderer that never received the field, so a failed *threaded* reply read as delivered with no way to send it again; `FailedSendNotice` is exported and shared rather than copied.

`fromHistory` never sets it: a message the host handed back is by definition one it kept, so a reload clears it — correct, because the reload re-reads what was actually journaled.

### 2. Which conversation is open is routed state (B-096)

Two facts made a draft escape into somebody else's DM, and only one is wrong.

The right one: the composer is deliberately **one instance** shared by every channel, and its draft deliberately survives a channel change (`MessageComposer`'s `suppressed` doc, PR #1984). Unmounting it is what discards a half-written message.

The wrong one: a bare `#/chat` — where the magic-link landing route puts you, because `useHashView` canonicalises the *view* and knows nothing about chat's channels — left the open channel as the value of an expression over `members`, `desks`, `transcripts` and `operator`. Each lands asynchronously and can re-order what `firstChannel` answers, so the conversation could change with **no navigation behind it**, and `send` addressed whatever it had become.

The channel id now goes into the hash the moment there is one to name, and `decodedSub` pins it from then on. A deep link short-circuits and always outranks it. This also **subsumes issue #412's restore**, which was a second effect writing the same hash for a bare entry: the two raced and the loser silently won — memory navigated to the remembered channel and the normaliser immediately replaced it with `firstChannel`. One effect, one decision, memory first.

**This is re-authored on top of #2134's `routeOpen` model, not merged into it.** #2052 deleted that effect and rebuilt it with no guard, which would have restored the bug #2134 fixed on 2026-09-07 (clicking **Flows** landing on `#/chat/main`). The guard is the first thing the effect does, and it matters *more* after this change, not less: the old effect only navigated when something was remembered (`if (remembered)`), so an operator with an empty `oc.chat.last-channel` was accidentally spared; this one always resolves — memory, else `channel.id` — and without the guard would bounce **every** such operator out of Flows on the first paint.

Guarded three ways: `room-rail-pinned.test.ts` (#2134's own source guard) is re-anchored on the effect's *body*, since the ref now sits with the other long-lived refs ~170 lines above it, and additionally asserts the unconditional resolve is present; and `chat-bare-hash-normalises.spec.ts` asserts in a real browser that `#/workflows` keeps its own address, **with and without** a remembered channel.

The ref is declared above the three early returns (`desksError`, `!desks`, `!channel`) rather than beside where it is read. A hook below them runs on some renders and not others — React error 310, a blank console with no composer at all on exactly the first paint.

### 3. Live replies dedupe by identity, not text alone

`isDuplicateLiveReply` checks the frame's own durable id (`hostMessageId(String(event.seq))`) first — a row already bearing it is unambiguously this same event, whatever its content. Content matching is kept, but **only against a row not yet reconciled to a durable id** (`!isHostMessageId(m.id)`): the narrow window where the SSE echo of your own turn arrives mid-await, before the POST resolves and names the row. Two genuinely different events that merely share wording — an operator repeating the same ambiguous `@name` produces two identical host notices — both carry durable ids from the start, so content matching never gets a chance to conflate them. The old check dropped the second one outright.

### 4. An unthreaded send keys the channel, not whatever thread is open

`stateKey` derived from `openThreadId` regardless of `parentId`. The channel composer always sends `parentId: undefined`, and a top-level row's Retry lives on that same transcript — so both could fire while an *unrelated* thread was open in the panel, filing the receipt and open-turn recovery under a thread they never touched. The panel showed work it did not start; the channel that ran it showed nothing. Unthreaded now keys `turnStateKey(chatId)` outright. Threaded still keys `openThreadId` rather than `parentId`, because a review reply anchors to a *reply* and the panel's own lookup keys on the open thread.

### 5. A system-attributed live frame

Rendered with `replyVoice`, so it is the centred pill `fromHistory` gives it on reload rather than a named teammate's bubble with an avatar and reply controls — otherwise whoever watched the turn live keeps a reading hydration never corrects. It is also never treated as a turn's completion: it is emitted mid-turn, before the cycle that answers has run, so clearing the live tool trace on it erases a turn that is still running. And `PendingSyncPosts.ended` now takes the settled response's own reply texts, so a held system frame is discarded when the response carries it and released when the response never will — some system lines are folded into `channel_responses` (`system_notice`'s approval overflow, `"Acknowledged."`) and some are deliberately not.

## Verified locally

**Commands run:**

- `npm run typecheck` — clean
- `npx vitest run` — **521 files, 4852 tests, 0 failures**

Browser verification (light and dark, screenshots in the thread below).

Supersedes the console half of #2052.

## Summary by CodeRabbit

- **New Features**
  - Added clear “Not sent” notices for failed messages, including the failure reason.
  - Added a Retry option for failed messages in chats and threads, preserving mentions and attachments.
  - Bare Chat links now resolve to a specific conversation while respecting remembered channels and direct links.

- **Bug Fixes**
  - Improved live-reply handling to prevent duplicate messages.
  - System notices now appear correctly without interrupting active tool traces.
  - Settled responses and live system notices are reconciled more reliably.

### DIFF
--- a/frontend/src/components/app-shell.tsx
+++ b/frontend/src/components/app-shell.tsx
@@ -62,6 +62,7 @@ import { startVisiblePolling } from "@/lib/visible-poll";
 import { withReadTimeout } from "@/lib/read-timeout";
 import {
   hasOtherOpenTurns,
+  isDuplicateLiveReply,
   mergeOpenTurns,
   openTurnsFromRuns,
   PendingSyncPosts,
@@ -2266,6 +2267,14 @@ export function AppShell({
   // resolves, instead of the shell needing a second copy of this logic.
   const renderAgentReply = useCallback(
     (event: AgentReplyEvent) => {
+      // `replyVoice`, not a literal: a live frame attributed to the runtime
+      // itself — `SYSTEM_AUTHOR` on the Rust side, which covers both B-101's
+      // mention-ambiguity note and the iteration-cap pause notice (issue
+      // #2068) — renders as the centred system pill `fromHistory` already
+      // gives it on reload, never as a named teammate's bubble with an
+      // avatar and reply/reaction controls, which is what unconditionally
+      // passing `"company"` here used to produce (Codex review, PR #2052).
+      const from = replyVoice(event.agentId);
       // The event names a thread; `chatChannelByThread` is the only thing that
       // knows which channel renders it. An id no channel owns is a no-op:
       // better silent than in the wrong place.
@@ -2305,20 +2314,32 @@ export function AppShell({
         // history's own order rather than appending to the recent tail this
         // scans. Live-then-hydrate was the one route neither guard covered,
         // and it doubled every reply that arrived while its channel was closed.
-        const dup = existing
-          .slice(-8)
-          .some((m) => m.from === "company" && m.text === event.text);
+        //
+        // **The content check alone is too broad** (Codex review, PR #2052):
+        // two genuinely different events can carry identical text — an
+        // operator repeating the same ambiguous `@name` produces two
+        // B-101 notices with the same wording — and content matching then
+        // suppressed the second one outright, not merely deduped it.
+        // `isDuplicateLiveReply` checks this event's own durable identity
+        // first (`event.seq`) and only falls back to content for a row that
+        // has not yet been reconciled to a durable id — see its own doc for
+        // why that scoping is what keeps two same-text-but-different events
+        // from being conflated. Named and extracted for the same reason
+        // every rule in `live-reply.ts` is: this is exactly the kind of
+        // regression that shows up nowhere but a repeated-mention screenshot.
+        const dup = isDuplicateLiveReply(existing.slice(-8), event, from);
         if (dup) return t;
         return {
           ...t,
           [channelId]: [
             ...existing,
-            // `replyVoice`, not a literal: a host-authored line (the
-            // iteration-cap pause) is projected with `agentId: "system"` and
-            // must render as the same centred row `fromHistory` gives it, or
-            // whoever watched the turn live keeps an agent-style bubble that
-            // hydration will never correct.
-            makeMessage(replyVoice(event.agentId), event.text, {
+            // `from` is `replyVoice(event.agentId)`, computed once above and
+            // reused by the `dup` check too — a host-authored line (B-101's
+            // ambiguity note, the iteration-cap pause notice) must render as
+            // the same centred row `fromHistory` gives it, or whoever watched
+            // the turn live keeps an agent-style bubble hydration never
+            // corrects.
+            makeMessage(from, event.text, {
               channel: event.agentId,
               taskId: event.taskId,
               mentions: event.mentions,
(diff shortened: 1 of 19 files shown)


graycyrus:
## Browser verification — 2026-09-08

Real browser (Chromium via Playwright), local host built from this branch on `127.0.0.1:8591`, company `agentic_software_company`, port claimed and identity-pinned before binding (`wt ports --verify 8591` → `OK 8591 pinned: pid 46315, instance 4ad38c7f65a54e715abb67335347c60c`) and released after. Sends were failed by throwing from `window.fetch` on the chat POST only — nothing else was stubbed.

### 1. A failed send is marked on its own row — light and dark

Sent `@backend_engineer can you confirm the staging deploy went out this morning?` with the mention picked from the `@` list, with the chat POST forced to throw.

The row itself is marked: the message text is muted, the `@backend_engineer` chip is intact, and a red-bordered `Not sent — cannot reach the company host at this origin   Retry` notice rides the row. The composer had cleared, and the text is still on screen — which is the whole point.

Both themes checked and both legible: the destructive border and text read correctly on the dark ground.

### 2. Retry keeps the mention chip **and** the attachment

Second failure with both: `@backend_engineer here are the sprint notes` + a staged `wt8-sprint-notes.txt` (67 B). The failed row carried the mention chip and the attachment card under the Not-sent notice.

Clicking **Retry** replaced the failed row with a delivered one and the host journaled exactly this:

```json
{"text": "@backend_engineer here are the sprint notes",
 "mentions": [{"target": {"kind": "agent", "id": "backend_engineer"}, "text": "@backend_engineer", "offset": 0}],
 "attachments": [{"nodeId": "01a082035922-0000000000fb", "name": "wt8-sprint-notes.txt", "mime": "text/plain", "size": 67}]}
```

Read from the host's own `events.jsonl`, not from the DOM. This is the property the ref-kept wire payload exists to guarantee: the rendered chip does not carry `{kind: agent, id: backend_engineer}` and the bubble does not carry the node id, so a Retry rebuilt from the row would have sent a chip-less, file-less message. Six operator messages are journaled in total — the two failed attempts reached the host **zero** times, so the failure was genuine and Retry was the only delivery.

### 3. The typing indicator lands on the channel, not an open thread

With an unrelated thread open in the panel (the `@backend_engineer` message's thread, reading "0 replies"), a **top-level** send in `#general` shows `Working…` in the channel timeline under the sent row, and the thread panel shows nothing — still "0 replies", no working row. Checked in both themes. Before the `stateKey` fix this send filed its receipt and open-turn recovery under `openThreadId`, so the panel showed work it did not start.

### 4. Clicking Flows still lands on Flows — the #2134 regression test

Both themes, and **both branches of the guard**:

- **With a remembered channel:** read a channel first, then click **Flows** → `#/workflows`, held for 3–4 s (long enough for a stray normalisation to have fired).
- **With nothing remembered** — the case #2052's version would have broken, and the one the old `if (remembered)` effect accidentally survived: cleared every `oc.chat.last-channel*` key, then a **full page load** straight onto `#/workflows` → address unchanged after 5 s, Flows rendered.

And the fix itself works: a bare `#/chat` resolved to `#/chat/main` on landing.

### Gates

`npm run typecheck` · `npm run typecheck:unit` · `npm run typecheck:e2e` — all three clean. `npx vitest run` — 521 files / 4852 tests, 0 failures. `scripts/ci/assert-design-tokens.sh` — clean.

### Inferred, not verified here

CI's own Playwright lanes have not run yet; the two e2e specs in this PR were typechecked but executed only as far as the manual walkthrough above covers the same behaviour.

graycyrus:
@coderabbitai review

PR #2167, opened 2026-09-08
fix(server): require admin authority to resolve or extend an approval

oxoxDev:
## Summary

On `main` today, `resolve_approval` and `extend_approval` take `CompanyAuth` and nothing else: **any signed-in member can settle or extend an approval**. #2047 closed the lifecycle routes yesterday and did not touch approvals, so this gap is still open.

This is #2127's remaining content, rebuilt on current `main`. Its lifecycle commits are now redundant — #2047 landed that work — and only the approvals half survives here. #2127 can be closed once this lands.

## API Or Behavior Changes

`POST …/approvals/{aid}` and `POST …/approvals/{aid}/extend` now require an admin, on both the company-scoped and self-scoped address forms. A member receives `403 forbidden`.

**One decision is deliberately left to the maintainer**: whether `extend` should be admin-only at all, or restored to any member. It is isolated in its own commit so it can be dropped without touching the rest.

## Tests

`cargo test --features openhuman --lib` — 7,214 passed, 0 failed. `--test auth_matrix` — 8 passed. `cargo clippy … -D warnings` 0, `cargo fmt --check` 0.

The route-authority matrix had **already recorded this defect**. Those four cells carried `Wait::AuthRoleBranch` — known-red, waiting for exactly this change — so the matrix was excusing them rather than checking them. Clearing the member cells drops the ignored-red count from **55 to 51**: the four this closes, and no others.

Two things surfaced while rebuilding it on main:

The matrix's completeness check failed with `operator direct path set drifted; missing=[…approvals/{aid}, …/extend]`. The four routes moved from `.route(...)` registration into the `scoped(...)` helper, so the inventory needed to follow them. That check merged yesterday and caught a sibling PR within a day.

Two tests failed on `another user already has the email harness-admin@example.test`. The shared state helper now seeds the fixed admin itself, so seeding again conflicts rather than being idempotent — the second seed is removed.

Also drops the `MemberAndTempPassword` red class, whose only users were these four rows.

## Documentation

`docs/spec/company-brain/approvals.md` states that deciding or extending an approval requires admin authority.

## Summary by CodeRabbit

- **Bug Fixes**
  - Approval updates and deadline extensions now consistently enforce administrative access across supported company URL formats.
  - Members can no longer modify approvals.
  - Anonymous users and administrators using temporary passwords receive consistent authentication and access restrictions.
  - Authorization behavior is now aligned across equivalent approval endpoints.

### DIFF
--- a/src/server/operator.rs
+++ b/src/server/operator.rs
@@ -15346,7 +15346,6 @@ mode = "full"
         for scope in APPROVAL_SCOPES {
             let home_dir = home();
             let state = state_with_company(home_dir.path(), "running").await;
-            crate::server::test_support::seed_fixed_admin(&state, "acme").await;
             let runtime = state.registry().get(&CompanyId::new("acme")).unwrap();
             let id = park_for_extend(&runtime, "appr-admin-resolve", 1_000).await;
             let cookie = crate::server::test_support::fixed_cookie("acme");
@@ -15369,7 +15368,6 @@ mode = "full"
         for scope in APPROVAL_SCOPES {
             let home_dir = home();
             let state = state_with_company(home_dir.path(), "running").await;
-            crate::server::test_support::seed_fixed_admin(&state, "acme").await;
             let runtime = state.registry().get(&CompanyId::new("acme")).unwrap();
             let id = park_for_extend(&runtime, "appr-admin-ext", 1_000).await;
             let cookie = crate::server::test_support::fixed_cookie("acme");
--- a/src/server/operator.rs
+++ b/src/server/operator.rs
@@ -49,7 +49,7 @@ use crate::server::chat_history::{
 use crate::server::error::ApiError;
 use crate::server::graphql::auth::GqlAuth;
 use crate::server::ops::language::{self, DEFAULT_DESK};
-use crate::server::ops::{ScopedCompany, scoped};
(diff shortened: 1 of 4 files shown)


oxoxDev:
Closing as a duplicate of #2127, which is the broader change and already approved.

I built this branch believing #2127 had been superseded except for the approvals half. That was wrong: #2127 still carries the same `resolve_approval` / `extend_approval` move to `AdminScopedCompany` through `scoped()`, plus the lifecycle routes, the temp-password drift between the two address forms, the approvals spec update, and the console affordance note. Everything here is contained in it.

The two findings raised on this PR are addressed there rather than lost:

- the console still offering members Approve / Decline / Extend — landed separately in #2137, which gates those actions on role; #2127 is rebased onto it
- `docs/spec/company-brain/approvals.md` contradicting the implementation — #2127 updates that spec in the same change

The reclassification of the two `{id}` approval rows to `RedCells::None` was right, and it is applied on #2127. I checked it empirically rather than by argument: with those cells promoted out of the ignored wait test, the boundary test passes, so the temp-password admin cell is the expected `Access::Admin` result and not a red cell. Ignored red cells there go 55 → 43.

Branch left in place; nothing to salvage from it.

PR #2201, opened 2026-09-09
feat(companies): retail-co — the tau2-bench retail domain as three desks

sanil-23:
## Summary

`retail-co` — the [tau2-bench](https://github.com/sierra-research/tau2-bench) **retail** domain run as a company of three desks and five seats, plus a runner that replays tau2 tasks through it and grades the result against tau2's own `evaluation_criteria`.

| desk | seats | remedy each seat holds | deliberates |
|---|---|---|---|
| `triage` | `triage` | none — 9 read tools, 0 mutating | no (one seat) |
| `order_ops` | `cancellations`, `amendments` | cancel the whole order / amend in place | yes |
| `returns` | `exchanges`, `refunds` | swap for a variant / take it back | yes |

## Why it is shaped like this

**Scope is enforced below the model.** Each seat gets exactly one `mcp:` grant, and each server registers only its role's tools. A seat reaching outside its role calls a tool that was never registered and fails at the protocol layer — it does not break a policy it was asked to respect. `triage` cannot cancel an order however the conversation goes.

**The write desks are pairs, so the room has something to argue about.** `deliberates()` requires two members, so a desk of one cannot deliberate at all. A delivered-order problem can be answered with an exchange or a refund; a pending-order problem by cancelling or amending. Neither seat can reach the other's tool, so the remedy has to be argued for rather than quietly done both ways, and `quorum = 2` on a two-seat desk makes the one that carries unanimous — the right bar for a write nobody can reverse.

**It asks what tau2 cannot.** tau2's orchestrator wires exactly one agent to one user simulator, with no agent-to-agent path, so it scores whether an agent called the right tool — not whether an *organisation* routed the work to the seat that owns it.

## The servers are not vendored

They live in `opencompany-tau2`, which vendors tau2-bench (~850 MB, mostly benchmark data) and needs its own Python venv. For contrast, `scripts/vending/` — the comparable fixture already in this repo — is 72 KB and stdlib-only.

So `mcp.json` ships five **disabled** entries pointing at placeholder `https` hosts (a bundle here must not point an agent at a host nobody has provisioned), and `scripts/retail-tau2.py` repoints them at loopback at runtime, which is the only layer that accepts an `http://` endpoint. Stdlib only, mirroring `scripts/vending-sim.py`.

## Two findings this surfaced

**`prompt_files` are silently dropped when a bundle is registered into a home.** Registration rewrites agents as inline `[[agent]]` blocks, and `resolve_prompt_files` is only called from the `agents/*.toml` loader — inline blocks have no resolution path. The body vanishes and the seat runs on its one-line `description`, with no warning. `dump-prompt.sh` against the source bundle shows the policy; against the registered copy it shows nothing. This bundle works around it by inlining `prompt`, but the underlying behaviour is worth fixing separately.

**An empty `delegates_to` refuses every crossing as `Unauthorized`, silently.** `authorized()` checks the source *agent* against the target *desk id*. With the list empty, the mention dispatches nothing and the sibling desk simply never takes a turn — the transcript looks normal. Each seat here names its reachable desks explicitly.

## Testing

Local `doctor --company companies/retail_co` ran clean on this bundle. This branch contains **no Rust or TypeScript changes** — nine data files and one stdlib Python script — so I've left compilation to CI rather than to my local vendor setup.

The flow was exercised end-to-end against a live host during development: task 0 (`#W2378156`, a two-item exchange on a delivered order) reached `returns` from `triage`, the room deliberated and referred a question back out to `triage` mid-episode, and the final database state matched tau2's `evaluation_criteria` on all four arguments — `item_ids`, `new_item_ids`, `payment_method_id` and status.

## Summary by CodeRabbit

* **New Features**
  * Added a Tau2 retail environment with triage, order changes, cancellations, exchanges, and refunds capabilities.
  * Added coordinated desk collaboration for resolving customer requests, including handoffs and shared decision-making.
  * Added support for assigning the General channel to a designated desk.
  * Added end-to-end environment startup and shutdown tooling with health checks and optional console access.
  * Added a simulator for replaying and grading retail and airline customer-service tasks.

* **Documentation**
  * Added setup, operating, and workflow documentation for the retail environment.


sanil-23:
@coderabbitai review

PR #2203, opened 2026-09-09
fix: Extend buys another window, a cancel has to be the whole ask, and the cost axis stops saying free

graycyrus:
Three fixes extracted from the abandoned #2054, rebuilt on current `main` and
reviewed as if they were someone else's work — because they were. Everything
else in that PR stays there; the comment on #2054 says which of its fixes this
one takes.

## 1. Extend bought almost nothing

`ManifestApprovalGate::extend` set `parked_at_millis = now`. The deadline is
`anchor + ttl`, so re-anchoring to `now` moved it by *the time that had passed
since the park* and nothing more. On a fresh approval — which is the state an
operator is actually in when they decide they want the weekend to think — that
is a no-op: the route answered `200 {"extended":true}`, a success toast came
up, and the card's countdown read the same "Declines itself in 23h" before and
after. Pressed twice in quick succession it bought tens of milliseconds against
a claimed 24 hours.

The reason written down for that was that an additive extension "would need its
own stored offset and a second place computing the deadline". It does not.
Adding one `ttl` to the anchor moves the deadline by exactly one `ttl`, through
the same single knob the sweeper and the projection already share. What was
needed was to stop reading "the window is measured from here" as "the work was
parked here" — a line the journal had already drawn, keeping
`deadline_anchor_millis` beside `at_millis` so a payload's age does not move
when a deadline does.

`.max(now)` keeps the one guarantee re-anchoring did give: an approval whose
deadline has passed but which the sweeper has not yet reached gets a full window
from now, never a deadline still in the past. Nothing caps it — pressing Extend
twice buys two windows, which is what the word means.

`extend` now returns the anchor it settled on rather than a `bool`, and both
durable halves — the live gate the sweeper reads and the journal the projection
reads and the next boot rehydrates from — are handed that same number instead of
each deriving one. The journal record's own doc already said it "carries the new
anchor rather than an offset"; it was being given `now` while the gate stored
something else, which is how a sweeper and a card come to disagree about one
deadline.

## 2. A message that both cancelled and answered was thrown away whole

A founder sent, in one message:

> @Tomas ok scrap the reminder card. 500 candles worth - jar, lid and label for
> 500 units total across the four scents, cold source it, under 4k. that's the
> answer

`classify_blocker_reply` short-circuited on `mentions_any(&lower, CANCEL_WORDS)`,
which scans the whole message, so the word `scrap` in a subordinate clause
decided a sentence whose main content was an answer. The work was abandoned
**and** the answer it carried — the only thing that could have unblocked the
card — was discarded with it. The card was left paused with no blocker behind
it, so nothing would ever have resumed it.

The verdict ranking above that line is a claim about *two verdicts against each
other* ("cancel it, but retry the other one" is a cancel). It was being applied
as a claim about a verdict against everything. So a cancel now has to be the
whole of the ask: every clause that is not the cancel clause must be filler.
Where it is not, the cancel reading is dropped and the message is judged on its
remaining merits — which for that sentence is an amend, the answer it was
written to give.

The clause split is the existing `clauses`, the one `later_clause_is_imperative`
already uses, so the two places that ask "does a later part of this sentence say
something the first part does not" answer it the same way. A comma is
deliberately not a boundary there, which is what keeps "cancel this one, retry
the other" a cancel.

**Only cancel**, and the asymmetry is the reason: cancel is the one verdict that
does not re-enter. Being wrong about it destroys the work; being wrong about
retry, skip or amend costs a re-run the next message can correct. It is the same
trade `triage_message` already makes between Answer and Chatter.

## 3. …and the P1 that follow-up opened

Because a comma is not a clause boundary, "scrap the reminder card, budget is
$500" is also one clause, and exempting a cancel clause wholesale would throw
the budget away too. A clause carrying a digit is kept in consideration instead
— a digit is the cheap tell that it carries a second fact rather than filler
around the cancel word.

That let `cancel task 123` and `drop option 2` fall through to Amend, which
**re-enters work the operator had just cancelled** — the same destructive
direction the guard exists to prevent, one step further along.

A cancellation that identifies its target numerically carries a digit too, so
the target designator is now removed before the residue is tested: a number
directly beside a target noun (`task 123`, `the 2nd option`, and `#` normalised
to the word it stands for so `cancel #4` reads as `number 4`) is that thing's
identifier, not a fact. Everything else a digit appears in still counts, so
"budget is $500" and "use 500 units" remain amendments, and so does a clause
naming both a target and a fact.

The noun list is short on purpose. Every entry widens what reads as a pure
cancel, so a word that is commonly a verb here — "order 200 units", "run 3
batches", "draft 3 emails" — is left out even though "cancel order 3" would read
fine. Losing a real fact is unrecoverable; failing to spot a numbered target
costs a re-run.

## 4. Money formatted by hand on one chart axis

`AnalyticsLens.tsx:118` formatted its own dollars —
``tickFormatter={(v) => `$${v.toFixed(2)}`}`` — so on the Observatory's "Cost per
agent" chart a real charge smaller than a cent read `$0.00` on the axis while
Finance, one click away, showed the same number as `<$0.01`. A founder looking
at the cost chart saw free.

PR #2050 made `frontend/src/lib/money.ts` the one renderer for a USD dollar
float and swept the console for hand-rolled formatting. This call site was
missed. It is now retargeted at `lib/money.ts`'s `usd`, which already states a
sub-cent amount as a bound instead of as zero and already agrees with every tile
and row that renders money elsewhere. `lib/money.ts` itself is unchanged.

**And the guard #2050 lacked.** The same axis has now been missed twice by a
sweep done by reading, so the sweep is a test:
`frontend/test/unit/money-formatting-sweep.test.ts` walks `frontend/src` and
fails on the three shapes a dollar figure gets rendered without a renderer — a
literal `$` glued to an interpolation, a `"$"` concatenated onto a value, and a
second `Intl` currency formatter. The three legitimate renderers (`lib/money.ts`,
`lib/cost.ts`, `views/finance/money.ts`) are named in an allow-list rather than
matched by filename, so a fourth is a decision written down in that file instead
of a file that happens to be called `money.ts`. Comments are stripped before the
scan, so a doc comment naming the anti-pattern does not count as one.

## Verified

Every claim below names the command that produced it.

| Gate | Command | Result |
|---|---|---|
| Rust format | `cargo fmt --all -- --check` | exit 0 |
| Rust lint | `cargo clippy --all-targets -- -D warnings` | exit 0 |
| Rust tests | `cargo test` | see PR comment |
| Console types | `npm run typecheck` | exit 0 |
| Unit types | `npm run typecheck:unit` | exit 0 |
| E2E types | `npm run typecheck:e2e` | exit 0 |
| Console tests | `npx vitest run` | 598 files passed, 1 skipped · 5179 tests passed, 3 skipped |
| Design tokens | `scripts/ci/assert-design-tokens.sh` | no arbitrary sizes, palette colours or stray hex |

**Proven by mutation**, each one separately:

- restoring the `now` re-anchor in `gate.rs` fails the new route test and both
  gate tests, and nothing else;
- removing the whole-ask requirement makes that founder's sentence a cancel
  again and fails both new parser tests, while
  `a_cancel_word_abandons_the_work` and
  `a_cancel_that_is_the_whole_message_still_cancels` guard the other direction;
- restoring the bare digit check fails
  `a_cancel_that_names_its_target_by_number_still_cancels` and nothing else,
  with `a_numbered_target_does_not_swallow_a_fact_beside_it` guarding the
  reverse;
- restoring ``tickFormatter={(v) => `$${v.toFixed(2)}`}`` fails the new sweep
  test, which names `src/views/observatory/AnalyticsLens.tsx` in its failure.

**Browser**, light and dark. The real console bundle served by a real host
(`opencompany serve`, company `agentic_software_company`), with only the
`ObservatoryRuns` GraphQL read stubbed so every agent's spend is under a cent by
construction — four agents at `$0.0042`, `$0.0031`, `$0.0018` and `$0.0009`.
Before and after were captured from two separately built bundles on two
separately booted hosts, and the tick labels below are read out of the painted
SVG rather than off the picture:

| | Axis of "Cost per agent" |
|---|---|
| **before** (light and dark) | `$0.00` · `$0.00` · `$0.00` · `$0.00` · `$0.01` |
| **after** (light and dark) | `$0.00` · `<$0.01` · `<$0.01` · `<$0.01` · `$0.01` |

Four of the five ticks said the spend was nothing. Every bar on that chart is
real, non-zero money.

## Not in this PR

Deliberately left in #2054: its `lib/money.ts` deletion (that file has nine
importers on `main`), its `resume-claim` scaffolding, its blocker/gated wire
split (landed as `ae9a82c95`) and its answer-on-the-wire work (superseded by
#2038).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USD formatting in the “Cost per agent” chart for consistent currency display.
  * Extending a parked approval now adds a complete time window from its existing schedule, with updated expiration times shown accurately.
  * Improved blocker reply handling so cancellation is recognized only when the entire request is clearly a cancellation; requests containing additional actionable details are handled as amendments.
* **Tests**
  * Added coverage for currency formatting, approval extensions, and blocker reply classification.


graycyrus:
@coderabbitai review

tomas:
@graycyrus why did you mention me?

Issue #1860, opened 2026-08-26
Epic: blockers are conversations — park → DM → reply → resume from the failed step


sanil-23:
## Problem

When a board-created task or a workflow run hits a problem, the operator is not told, cannot answer, and the work cannot pick up where it stopped. Concretely, per failure kind:

| Kind | What happens today | What the operator sees |
|---|---|---|
| Provider/config error (bad model id like #1811, provider 4xx/timeout) | run settles Failed, card returns to `todo`, reason on the note | A wordless column hop; `todo` is both the failure state and the fresh state |
| Tool / MCP / integration failure | failure text appended to the card | Only visible by opening the card and reading the note |
| Planning blocked (missing prerequisite, no valid assignee) | card returns from `planning` to `todo`, reason on the note | A blocked card and an untouched card look identical |
| Approval park | run parks, card → `paused`, approvals feed | Browser-only; with the tab closed, TTL auto-denies in silence |
| Workflow node hard error | run Failed, journaled | SSE toast if a tab is open; a scheduled overnight run: nothing |
| Workflow agent node blocked | parks; resume stash is in-memory, dies on restart (#1816) | A run that can never resume |

Every one of these blockers is **answerable by a person** — a new model id, a refreshed key, a missing prerequisite, a sign-off — but the system has no way to ask, no way to route an answer, and no way to restart the stopped step.

## The mechanism

**Blockers become parks, parks become DMs, replies become resolutions, resolutions resume the stopped step.**

1. **Park instead of settle** — at the failure site, classify: answerable by a person? If yes, park a durable `blocker` effect carrying the step/node id, the failure kind, and what is needed — extending the existing approval-park family (`ApprovalParked` journal + `ContinuationQueue`, rehydrated at boot), so restart-durability comes free. Non-answerable failures settle Failed and go to the passive layer (see the companion surfacing issue).
2. **Deliver as a DM from the stuck step** — the parked blocker renders as a card in the operator's DM with the responsible agent (the task's assignee, or the stuck node's agent; host identity for non-agent nodes), the pattern `approval-card.tsx` already uses in-thread — plus a `NotificationStore` entry so it survives a closed tab.
[...]

PR #2217, opened 2026-09-10
fix(chat): thread the blocker prompt off the message it answers (#2039)

### LINKED ISSUE #2039: chat: a blocker disambiguation prompt is posted unthreaded, though its doc says it threads
## Summary

When several blockers are pending on one desk and the operator's reply is ambiguous, the runtime asks which one they meant. The operator's own message is threaded, but the teammate's question is posted to the channel root — so the exchange splits across two places.

## Where

`post_blocker_prompt` (`src/company/runtime.rs`) takes no parent and hardcodes it:

```rust
pub(crate) async fn post_blocker_prompt(&self, thread: &str, prompt: &str) -> Result<()> {
    // ...
    CompanyEvent::AgentReply {
        parent: None,
        chat_id: thread.to_string(),
        // ...
    }
}
```

The caller has the anchor and drops it on the next statement (`src/server/operator.rs`, the `BlockerReplyPlan::AskWhich` arm):

```rust
let accepted = accept_chat_turn(&runtime, id, &message, by.as_ref(), parent, &desk).await?;
// ...
let posted = runtime.post_blocker_prompt(&desk, &prompt).await.map_err(ApiError);
```

`parent` is in scope, is passed to `accept_chat_turn`, and is not passed here.

## Expected

The disambiguation prompt should carry the same parent the operator's message did, so the question lands in the thread that asked it. Where there is genuinely no parent, `None` is correct — but it should be the caller's answer, not a constant.

## Impact

The operator sees their reply in a thread and the teammate's follow-up question at the channel root. This is the surface #1860 exists to make coherent: a blocker is supposed to read as one conversation.

## Note on the class

The function's own doc comment says it threads and reloads like any transcript line, while the body cannot thread at all — a stated invariant with nothing enforcing it, and the caller already holding the value that would satisfy it. Same shape as #2025 (a card's thread root dropped before the Room jump) and #2008 (`escalate_to_human` writes "the approval's own task link already names it", and no consumer reads it).

Found during the #2008 root-cause work; will be fixed alongside it, since both sit in the blocker resume/prompt path.

oxoxDev:
## Summary

The ask-which question was posted with `parent: None` while the caller held the operator's anchor and dropped it on the next statement. The operator read their own reply inside a thread and the teammate's follow-up question at the channel root — the split this tier exists to close.

`post_blocker_prompt` now takes `parent` as a required argument rather than a constant, which puts the question to every caller, and re-resolves it through `resolvable_parent`: a blocker's anchor can be days old by the time it is answered, and a root that is gone threads nothing.

Scope is the whole family, not the prompt alone. The resume notes thread on the same terms, carrying the parent the approval recorded when it parked, so the acknowledgement lands where the question was asked. That reaches both card resumes and the workflow-node resume, which gain the anchor `drive_blocker_resume` already held.

The prompt takes the caller's own root through `reply_thread` — the convention every other reply in that handler already follows.

Closes #2039
Part of #1860

## API Or Behavior Changes

- `post_blocker_prompt` gains a required `parent: Option<EventSeq>` argument. Internal (`pub(crate)`); the required form is deliberate, so a caller holding an anchor has to decide about it rather than inherit a constant.
- `post_blocker_resume_note`, `cancel_task_card` and `resume_node_blocker` thread the park's recorded parent.
- Behaviour: the disambiguation prompt and the resume acknowledgements now hang off the message they answer instead of the channel root. Where there is genuinely no resolvable root, `None` still applies and the reply sits in the channel as before.

## Tests

Three new cases, each red-proved against the mechanism it pins:

- `the_ask_which_question_threads_off_the_reply_that_was_ambiguous` — two blocked things on one DM, an ambiguous reply, the prompt asserted against the operator's own message sequence. Red: `right: Some(EventSeq(7))`.
- `a_resume_note_threads_off_the_question_it_answers` — a re-entered card and a cancelled card both acknowledge into the thread the question was asked in. Red, with the anchor dropped at the note sites: `left: [None]` against `right: [Some(EventSeq(6))]`.
- `a_resume_note_whose_anchor_is_gone_still_answers` — pins the guard rather than the pass-through. Red, with `resolvable_parent` removed: a dangling `Some(EventSeq(9999))` instead of the channel.

The resumes are driven directly rather than through a park, because `park_blocker` records no parent of its own — only an `escalate_to_human` park carries one, so a park-based test would pass while proving nothing.

- [x] `cargo clippy --features openhuman,tinymemory --lib --all-targets --no-deps -- -D warnings` (exit 0)
- [x] `cargo test --features openhuman,tinymemory --lib -- blocker server::operator::test::` — 339 passed, 0 failed, re-run after the rebase resynced the vendored submodule pins
- [ ] `cargo test` unscoped — not run; the full suite exhausts memory on this machine
- [ ] Console verification — not done. The threading is asserted at the event level only; nothing here proves the console renders the reply inside the thread.

## Documentation

No user-facing docs. The threading rule now lives on `post_blocker_prompt`'s own doc, where the previous constant contradicted it.

---

**Note on CI:** this branch has been rebased onto the restored `main` (#2215 landed the missing `graphql` fixture field), so lanes here should now reflect this change rather than inherited breakage.

### DIFF
--- a/src/company/runtime.rs
+++ b/src/company/runtime.rs
@@ -14225,6 +14225,52 @@ mod tests {
                 .collect()
         }
 
+        /// Journals an operator line in the teammate's DM and hands back its
+        /// sequence, the root a reply in that DM threads off.
+        async fn asked_in_dm(runtime: &Arc<CompanyRuntime>) -> crate::ports::types::EventSeq {
+            runtime
+                .events
+                .append(
+                    &runtime.id,
+                    crate::ports::types::CompanyEvent::OperatorMessage {
+                        text: "which brief is current?".to_string(),
+                        chat: Some("dm:eng".to_string()),
+                        parent: None,
+                        by: None,
+                        deliverable: None,
+                        mentions: Vec::new(),
+                        attachments: Vec::new(),
+                    },
+                )
+                .await
+                .expect("journal the question")
+        }
+
+        async fn dm_replies(
+            runtime: &Arc<CompanyRuntime>,
+        ) -> Vec<(String, Option<crate::ports::types::EventSeq>)> {
+            runtime
+                .events
+                .read_from(
+                    runtime.id(),
+                    crate::ports::types::EventSeq::new(0),
+                    usize::MAX,
+                )
+                .await
+                .expect("read events")
+                .into_iter()
+                .filter_map(|stored| match stored.event {
+                    crate::ports::types::CompanyEvent::AgentReply {
+                        chat_id,
+                        text,
+                        parent,
+                        ..
+                    } if chat_id == "dm:eng" => Some((text, parent)),
+                    _ => None,
+                })
+                .collect()
+        }
+
         fn agent_question() -> BlockerPayload {
             BlockerPayload {
                 kind: BlockerKind::Information,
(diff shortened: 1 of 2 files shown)


oxoxDev:
The red `Console E2E (live brain)` lane here is pre-existing on `main`, not from this change.

I diffed the failing spec sets rather than reading them: 77 failures on this branch, 77 on main's own run of the same lane, and **no spec fails here that does not also fail on main** — identical sets, matching timings. The failures span `agent-detail`, `board-columns`, `brain-memory`, `chat-live-events` and others this diff does not touch, with a uniform 30.5s/1.0m timeout signature rather than assertion mismatches.

The blocker surfaces the change is about pass: all of `approval-blocker-verdicts`, and `blocker-verdicts-live` ("every one of the four verdicts is reachable, and the host acts on the one clicked").

Every other lane is green, including the gated `Rust (openhuman, tinymemory)`.

One gap declared rather than hidden: the threading is asserted at the event level only. Nothing here proves the console renders the reply inside the thread.

PR #2242, opened 2026-09-10
fix(setup): rate-limit propose_roster per company

oxoxDev:
## Summary

`POST {scope}/setup/roster` (`propose_roster`) had no rate limit at all: an authenticated operator could re-trigger the paid model pass behind it unboundedly, and every call is real, metered inference spend. Adds a per-company burst cap and un-ignores the test that documented the gap.

## API Or Behavior Changes

`propose_roster` now refuses with `429 roster_proposal_rate_limited` once a company has made 8 roster-proposal calls inside a trailing 60-second window. The limiter is a process-wide, per-company in-memory registry (`ROSTER_BURSTS`) checked before any work — including the answer-persisting write — so a refused call has no side effects, mirroring how `WorkflowRunLimit` already refuses workflow runs at their concurrency ceiling with the same status and no partial write.

## Tests

- [x] `cargo build --all-targets` (via `cargo check --all-targets`)
- [x] `cargo test` (targeted: `server::ops::setup_test`, `server::ops::` (893 tests), `server::error::` — full `cargo test` not re-run for time, but every module touched by this change was run directly)

Red proof: removed the `admit_roster_proposal(...)?;` call, re-ran `repeated_roster_proposals_are_rate_limited` and `roster_proposals_within_the_burst_cap_succeed_then_429` — both failed on their own assertions (`expected at least one 429 across 20 rapid calls... got: [200, 200, ...]` and `assertion left == right failed / left: 200 / right: 429`), then restored the fix and confirmed both green again.

## Documentation

No docs needed — the cap is an internal safety limit, not a new operator-facing contract; the 429 envelope follows the existing `{ "error", "code" }` shape every other refusal on this route already uses.

## Summary by CodeRabbit

* **New Features**
  * Added rate limiting for roster-proposal requests: up to 8 requests per company within 60 seconds.
  * Requests exceeding the limit now receive a clear 429 rate-limit response with a stable error code.

* **Bug Fixes**
  * Improved handling and reporting of roster-proposal rate-limit errors.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -203,6 +203,19 @@ pub enum OpenCompanyError {
         limit: usize,
     },
 
+    /// A `POST {scope}/setup/roster` call was refused because this company
+    /// already reached its burst cap of roster-proposal calls for the
+    /// current window. Raised before any model pass runs.
+    #[error(
+        "this company already requested {limit} roster proposals in the last {window_secs}s; each call runs a paid model pass — wait before trying again"
+    )]
+    RosterProposalRateLimit {
+        /// The burst ceiling that was hit.
+        limit: usize,
+        /// The rolling window, in seconds, the ceiling applies over.
+        window_secs: u64,
+    },
+
     /// A workflow run failed after some of its nodes had already done durable
     /// work (issue #1008).
     ///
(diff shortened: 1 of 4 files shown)


oxoxDev:
Resolved the unbounded registry finding. `ROSTER_BURSTS` now evicts entries only after their call window is empty and no caller holds the burst, preserves active entries, and refuses insertion once 1,024 live keys remain.

Red proof: temporarily removed only the capacity check. `server::ops::setup::roster_burst_tests::roster_burst_registry_is_bounded_and_evicts_only_idle_entries` failed on its own assertion with `left: 0`, `right: 1` (exit 101). Restored byte-identically; the same test passed (1 passed, 0 failed; exit 0).

Verification: filter listing found exactly 1 test; `cargo fmt --check` exit 0; `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings` exit 0. Deletion scans contained only replacement of the lifetime registry accessor and the assertion form used for the red proof.

@coderabbitai review

PR #2246, opened 2026-09-10
fix(composio): reuse a pending OAuth handoff instead of opening a second one

oxoxDev:
## Summary

`composio_authorize` opened a fresh hosted OAuth handoff on every call. A repeat authorize for a toolkit already mid-connect started a second one.

## Problem

There was no dedup around `client.authorize`. Calling authorize twice for the same toolkit handed the operator a new connect URL while the first handoff was still live, orphaning it. The model has no way to know a handoff is already in flight, so a retry — the most natural thing for it to do — reliably produced this.

An acceptance test for it already existed in the file, `#[ignore]`d, with the defect written into the reason: *"confirms fail-open: repeat authorize is not deduped or short-circuited"*.

## Solution

Key the in-flight handoff on company, credential, toolkit and extra parameters, and reuse it while the connection is still `PENDING`/`INITIATED`/`INITIALIZING` and inside the documented lifetime.

Deliberate about the states that are *not* reused:

- `ACTIVE`, `CONNECTED`, `EXPIRED`, `FAILED`, `ERROR`, `INACTIVE`, `DISCONNECTED` — replaced with a new handoff, since reconnecting is the point
- an unrecognised or unreadable status — **refuses** rather than guessing, and says no new handoff was started
- a failed POST is not cached, so a transient failure does not poison later attempts
- pending entries are pruned at the documented lifetime

The tool description now states the reuse behaviour. A schema that promises something the code does not do is its own defect class, and this repo already has one open.

## Submission Checklist

- [x] Full lib suite under `--features openhuman,composio` — **7701 passed, 0 failed**
- [x] `cargo test --lib --features openhuman,composio authorize` — 23 passed, **0 ignored**
- [x] Red proof: removing the pending-handoff reuse fails the acceptance on its own assertion — `a repeat authorize for the same toolkit must not open a second handoff / left: 2 / right: 1` — then restored to green
- [x] Production-deletion scan: the `#[ignore]`, the old unconditional `client.authorize` call it replaces, and the tool description it supersedes
- [ ] Console E2E — pre-existing failure on main, unrelated

## Impact

A repeat authorize now returns the handoff already in flight instead of orphaning it. Reconnecting after a completed, failed or expired connection is unchanged.

## Related

Seven further tests cover the race between clones, reconnection after a terminal state, unknown status, failed POST caching, lifetime expiry, cache capacity, and key isolation across companies and credentials.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate OAuth authorization handoffs when identical requests occur concurrently.
  * Reuses active authorization attempts for up to 10 minutes and reconnects after completion or expiration.
  * Rejects unreadable connection statuses without starting unnecessary new handoffs.
  * Enforces limits on pending authorization attempts and avoids caching failed requests.

* **Tests**
  * Added coverage for concurrent requests, expiration, reconnection, isolation, capacity limits, and failed authorizations.

### DIFF
--- a/src/harness/built_in/composio.rs
+++ b/src/harness/built_in/composio.rs
@@ -924,7 +924,7 @@ mod live {
         let http = managed_execute_client()?;
 
         let post =
-            async |body: &Value| post_managed_execute(client.inner(), &http, body, &key).await;
+            async |body: &Value| post_managed_execute(client.inner(), http, body, &key).await;
 
         let mut resp = post(&body).await?;
         if is_post_oauth_auth_error(&resp) {
@@ -2237,7 +2237,8 @@ mod live {
             let gmail_result = gmail_call.await.unwrap().unwrap();
             assert!(!gmail_result.is_error, "{}", gmail_result.output());
             assert_eq!(
-                slack_completed, true,
+                usize::from(slack_completed),
+                1,
                 "an authorization for one toolkit must not wait for another toolkit's network call"
             );
             server.abort();
--- a/src/harness/built_in/composio.rs
+++ b/src/harness/built_in/composio.rs
@@ -175,6 +175,8 @@ pub struct TenantComposio {
     /// `composio_execute` sends no connection id, leaving the account to
     /// Composio's own resolution exactly as before.
     defaults: crate::company::composio::ComposioDefaults,
+    #[cfg(feature = "composio")]
+    authorizations: Arc<tokio::sync::Mutex<live::AuthorizeCache>>,
 }
 
 impl TenantComposio {
@@ -196,6 +198,8 @@ impl TenantComposio {
             catalog: Credential::None,
             toolkits,
             defaults: Default::default(),
+            #[cfg(feature = "composio")]
+            authorizations: Default::default(),
         }
     }
 
@@ -238,6 +242,8 @@ impl TenantComposio {
             catalog: Credential::None,
             toolkits,
             defaults: Default::default(),
+            #[cfg(feature = "composio")]
+            authorizations: Default::default(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


oxoxDev:
Resolved the tenant-wide mutex finding with key-scoped single-flight. The tenant cache is now held only to reserve/inspect a request key; both `list_connections()` and `authorize()` run after it is released. Same-key callers still deduplicate, while different toolkits proceed independently.

Red proof: temporarily forced every authorization to use the same single-flight key. `harness::built_in::composio::live::live_tests::a_slow_authorize_does_not_block_an_unrelated_toolkit` failed on its own assertion with `left: 0`, `right: 1` (exit 101). Restored byte-identically; the test passed with `openhuman,composio` (1 passed, 0 failed; exit 0).

Verification: filter listing found exactly 1 test; `cargo fmt --check` exit 0; required `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings` exit 0; additional Composio-inclusive clippy with `openhuman,tinymemory,composio` exit 0. Deletion scans contained only the replaced tenant-wide critical section and two lint-driven expression replacements.

@coderabbitai review

oxoxDev:
Addressed the remaining toolkit-key normalization finding in `7f04f68e0`.

Red proof: I changed only the request-key identity back from `toolkit.trim().to_ascii_lowercase()` to the raw `toolkit`. The case-variant acceptance test failed on its own assertion:

```
assertion `left == right` failed: a repeat authorize for the same toolkit must not open a second handoff
  left: 2
 right: 1
red_rc=101
```

After restoring the fix, the worktree matched the committed tree and the test passed 1/1.

Verification:
- filter listing: 1 test
- `cargo test --locked --lib --features openhuman,tinymemory,composio a_repeated_authorize_for_one_toolkit_is_deduped`: 1 passed, rc 0
- `cargo fmt --check`: rc 0
- `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings`: rc 0
- `cargo clippy --all-targets --no-deps --features openhuman,tinymemory,composio -- -D warnings`: rc 0

@coderabbitai review

PR #2248, opened 2026-09-11
fix(search): keep the slot for a request the backend already received

oxoxDev:
## Summary

A failure decoding the search backend's reply refunded the daily search slot, reopening budget the company had genuinely spent.

## Problem

The refund path did not distinguish *never reached the backend* from *reached it and could not read the answer*. A malformed but **2xx** reply means the request arrived — the backend may already have run and billed the search — yet the slot came back as if nothing had happened.

Two consequences. The daily count undercounts real spend. And because the tool returns an error the caller is invited to retry, a backend returning unreadable replies lets one company reopen its search budget indefinitely.

The test covering this asserted the refund was correct, while naming the problem in its own message: *"this is the HT-098 undercounting gap, not a proof that undercounting is safe"*. It documented the bug rather than guarding against it.

## Solution

Refund only what failed **before dispatch** — a missing credential, a reservation never spent. Once the request is attempted, the slot is retained and the returned error says so, so the model tells the operator search is unavailable rather than inventing sources.

## Submission Checklist

- [x] Full lib suite under `--features openhuman` — **7641 passed, 0 failed**
- [x] `cargo test --lib --features openhuman search::` — 60 passed, 0 ignored
- [x] Red proof: restoring the refund on the post-dispatch path fails both new tests on their own assertions — `a response parsing failure must not refund a request the backend received / left: 0 / right: 1` — then restored to green
- [x] Production-deletion scan: the old refund call, the error text it superseded, and the replaced pinning test
- [ ] Console E2E — pre-existing failure on main, unrelated

## Impact

Daily search counts now reflect requests the backend actually received. A company whose backend returns unreadable replies can no longer reopen its own budget by retrying. Genuine pre-dispatch failures still refund, so nothing is charged for a request that never left.

## Related

The replaced test is joined by `failed_backend_responses_cannot_reopen_the_daily_search_budget` and `a_missing_credential_refunds_the_pre_dispatch_search_reservation`, pinning both sides of the boundary.

## Summary by CodeRabbit

* **Bug Fixes**
  * Search backend failures now provide clearer uncertainty information instead of a generic connectivity error.
  * Usage credits are retained after a search request is sent, including malformed or failed backend responses.
  * Pre-request failures, including missing credentials or blocked requests, refund unused usage reservations.
  * Error details are scrubbed to prevent credential exposure.
  * Failed searches no longer trigger an unintended second dispatch.

### DIFF
--- a/src/harness/built_in/search.rs
+++ b/src/harness/built_in/search.rs
@@ -304,6 +304,7 @@ pub fn search_tools(backend: &SearchBackend, metering: SearchMetering) -> Vec<Bo
     vec![Box::new(WebSearchTool {
         backend: backend.clone(),
         metering,
+        pre_dispatch: search_pre_dispatch,
     })]
 }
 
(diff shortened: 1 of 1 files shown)


oxoxDev:
Resolved the typed pre-dispatch finding. Search now runs the live egress gate through a typed `SearchDispatchError::BeforeDispatch` path before calling `IntegrationClient::post`; only that proven-no-send variant refunds. Errors returned after entering the integration client remain `AfterDispatchAttempt` and retain the slot.

Red proof: temporarily removed only the refund from the typed pre-dispatch arm. `harness::built_in::search::tests::a_proven_pre_dispatch_failure_refunds_the_search_reservation` failed on its own assertion with `left: 1`, `right: 0` (exit 101). Restored byte-identically; the same test passed (1 passed, 0 failed; exit 0).

Verification: filter listing found exactly 1 test; `cargo fmt --check` exit 0; `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings` exit 0. The deletion scan contained only replacement of the direct `IntegrationClient::post` match with the typed dispatch wrapper.

The vendored `IntegrationClient` converts reqwest transport errors to string-only `anyhow` messages, so this crate cannot safely classify DNS/TCP/TLS as pre-dispatch by typed predicates; those remain retained rather than adding a string heuristic.

@coderabbitai review

PR #2251, opened 2026-09-11
fix(delegation): refuse reviews outside in-review state

oxoxDev:
## Summary
Refuse review_task approvals for cards that are not currently in_review.

## Problem
A never-dispatched todo card could be force-moved to done by an approval verdict.

## Solution
Validate the card column before recording the verdict, preserve the card unchanged, and report the refusal while continuing the drain.

## Submission Checklist
- [x] Focused tests listed and passing
- [x] Assertion-level red proof observed and restored byte-identically
- [x] Production-deletion scan reviewed

## Impact
Review writes now require the card state in_review; valid reviews are unchanged.

## Related
HT-078. Main has no branch protection; this gate keeps the state-machine invariant reviewable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Reviews and assignments now use atomic conditional updates, preventing newer task changes from being overwritten.
  - When concurrent reviews target the same card, only one is accepted; the other is refused with a clear explanation.
  - Refused reviews preserve the card’s latest state.
  - Board update events are emitted only when a task update succeeds.
  - Genuine task-store write failures continue to be reported as errors.

### DIFF
--- a/src/company/runtime.rs
+++ b/src/company/runtime.rs
@@ -11968,6 +11968,7 @@ mod tests {
                 &self,
                 _company: &CompanyId,
                 _task: &TaskRecord,
+                _observed: &TaskRecord,
                 _expected_column: &str,
             ) -> crate::Result<bool> {
                 Ok(false)
--- a/src/company/runtime.rs
+++ b/src/company/runtime.rs
@@ -11964,6 +11964,14 @@ mod tests {
             async fn upsert(&self, _company: &CompanyId, _task: &TaskRecord) -> crate::Result<()> {
                 Ok(())
             }
+            async fn update_if_column(
+                &self,
+                _company: &CompanyId,
+                _task: &TaskRecord,
+                _expected_column: &str,
+            ) -> crate::Result<bool> {
+                Ok(false)
+            }
             async fn delete(&self, _company: &CompanyId, _id: &str) -> crate::Result<bool> {
                 Ok(false)
             }
--- a/src/harness/built_in/orchestrator.rs
+++ b/src/harness/built_in/orchestrator.rs
@@ -12426,6 +12426,14 @@ name = "Morning"
         async fn upsert(&self, _company: &CompanyId, _task: &TaskRecord) -> crate::Result<()> {
             unimplemented!("not exercised by these tests")
         }
+        async fn update_if_column(
+            &self,
+            _company: &CompanyId,
+            _task: &TaskRecord,
+            _expected_column: &str,
+        ) -> crate::Result<bool> {
+            unimplemented!("not exercised by these tests")
+        }
         async fn delete(&self, _company: &CompanyId, _id: &str) -> crate::Result<bool> {
             unimplemented!("not exercised by these tests")
         }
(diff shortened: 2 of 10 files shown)


oxoxDev:
Resolved the atomicity finding with a conditional `TaskStore::update_if_column` implemented by each production backend and delegated by the board announcer. `review_task` now refuses a stale/deleted card instead of overwriting or recreating it.

Red proof: temporarily replaced the conditional write with unconditional `upsert`; `runtime::delegation::tests::two_concurrent_reviews_of_the_same_card_admit_exactly_one_writer` failed on its own assertion with `left: 0`, `right: 1` (exit 101). Restored byte-identically; the same test then passed (1 passed, 0 failed; exit 0).

Verification: `cargo fmt --check` exit 0; `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings` exit 0. Deletion scan contained only the intentional replacement of the unconditional review write and bug-pinning test assertions.

@coderabbitai review

oxoxDev:
Addressed the follow-up findings in `b41bfd769` and `293741123`:

- `TaskStore::update_if_column` now carries the caller's observed record through filesystem, SQLite, MongoDB, and `BoardEvents`; each backend compares that observed version before applying the write.
- `assign_task` now uses the same CAS path, so its same-column update cannot invalidate and then be overwritten by a stale review.
- MongoDB deserialization borrows `current_json`, and `cargo check --locked --features mongodb --lib` exits 0.
- Added a deterministic mixed assignment/review interleaving: both calls read the same revision, assignment lands first, then review attempts its stale write.

Red proof: weakening only the filesystem predicate back to `(id, column)` made `a_same_column_assignment_cannot_be_overwritten_by_a_stale_review` fail on its own assertion:

```
assertion `left == right` failed: one operation must refuse the snapshot invalidated by the other
  left: 0
 right: 1
```

The file was restored byte-identically; the test then passed 1/1. The `runtime::delegation::tests::` filter lists 124 tests and passes 124/124 under `openhuman,tinymemory`; `cargo fmt --check` exits 0; required all-target clippy with `openhuman,tinymemory` exits 0.

@coderabbitai review

oxoxDev:
Tightened the ordered mixed-interleaving acceptance in `9e4e81c89`: the assignment must succeed, the stale review must refuse, and the stored card must retain the in-review column, assignee, and assignment note.

Red proof: I weakened only the filesystem CAS predicate from full observed-record equality back to id + column. The target failed on its own ordered assertion:

```
assertion `left == right` failed: the stale review ordered second must refuse
  left: 0
 right: 1
red_rc=101
```

After restoring full observed-record equality, `git diff --exit-code` returned 0 and the test passed 1/1.

Verification:
- filter listing: 1 test
- `cargo test --locked --lib --features openhuman,tinymemory a_same_column_assignment_cannot_be_overwritten_by_a_stale_review`: 1 passed, rc 0
- `cargo fmt --check`: rc 0
- `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings`: rc 0
- pre-commit deletion scan removed only the earlier either-writer assertions; no production deletion

@coderabbitai full review

oxoxDev:
@coderabbitai The ordered test in commit `9e4e81c89` now explicitly requires assignment success, stale-review refusal, `COLUMN_IN_REVIEW`, assignee `chief`, and the assignment note. Is the prior test-coverage objection resolved?

PR #2265, opened 2026-09-11
fix(budget): enforce monthly cap on inference spend

oxoxDev:
## Summary

Enforce `[budget].monthly_usd` before every built-in inference dispatch, including confined turns, so a private company cannot continue model spend after reaching its declared company-wide cap.

## Problem

The monthly budget was projected by Finances and passed to the discoverable-company economy path, but the dominant inference path never consulted it. A non-discoverable company at its cap still called the model. Concurrent turns could also read the same remaining budget and both dispatch before either cost reached the ledger.

The pre-fix acceptance failed on its own assertion:

```
assertion `left == right` failed: a company at its monthly cap must not make another inference call
  left: 1
 right: 0
```

## Solution

Cache the declared company cap with the resident roster, then reload the live ledger under a per-company inference-spend lock before dispatch. The lock is held through model execution and cost journaling, so a competing turn sees spend posted by the preceding turn. Lock registry entries are weak and disappear when no call holds or awaits them.

Ledger read failures fail closed with a visible operator-facing refusal. Companies without a monthly cap take the unchanged fast path.

## Submission Checklist

- [x] Acceptance filter listed exactly one test; green with 1 passed, 0 failed under `openhuman,tinymemory`
- [x] Barrier race filter listed exactly one test; five attempts green with 1 passed, 0 failed under `openhuman,tinymemory`
- [x] Red proof: bypassing the monthly gate made the at-cap test fail with left 1/right 0 and the race fail with left 2/right 1; the production file was restored byte-identically and both returned green
- [x] `cargo fmt --check` — exit 0
- [x] `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings` — exit 0
- [x] Production-deletion scan reviewed: empty for the first commit; the second commit only replaced the lock registry's strong values with weak references

## Impact

Inference spend now obeys the company-wide monthly ceiling for non-discoverable and discoverable companies, and overlapping turns cannot multiply the remaining budget within one process. As with the existing total-token reservation, cross-replica atomicity would require a shared durable reservation service.

## Related

Complements the total-token reservation merged in #2257. It does not change x402 or standing-grant behavior.

## Summary by CodeRabbit

* **New Features**
  * Added a company-wide monthly spending cap for harness activity.
  * Prevents new model requests when the configured monthly budget is exhausted or cannot be verified.
  * Added clear notices explaining monthly budget refusals.

* **Bug Fixes**
  * Prevented concurrent requests from exceeding the remaining monthly budget.
  * Ensured blocked requests do not initiate model inference.

### DIFF
--- a/src/harness/built_in/mod.rs
+++ b/src/harness/built_in/mod.rs
@@ -2629,15 +2629,18 @@ fn policy_ensure_lock(company: &CompanyId) -> Arc<tokio::sync::Mutex<()>> {
 }
 
 static MONTHLY_SPEND_LOCKS: std::sync::LazyLock<
-    std::sync::Mutex<HashMap<CompanyId, Arc<tokio::sync::Mutex<()>>>>,
+    std::sync::Mutex<HashMap<CompanyId, std::sync::Weak<tokio::sync::Mutex<()>>>>,
 > = std::sync::LazyLock::new(|| std::sync::Mutex::new(HashMap::new()));
 
 fn monthly_spend_lock(company: &CompanyId) -> Arc<tokio::sync::Mutex<()>> {
     let mut locks = MONTHLY_SPEND_LOCKS.lock().expect("monthly spend locks");
-    locks
-        .entry(company.clone())
-        .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(())))
-        .clone()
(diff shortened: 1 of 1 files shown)


oxoxDev:
@coderabbitai review

PR #2267, opened 2026-09-11
fix: prevent consumed grant replay after restart

oxoxDev:
## Summary

Make single-use grant dispatch durable before an approved follow-up turn can re-issue its tool call, preventing a crash before the ordinary consumption drain from re-arming spent authority.

## Problem

A grant was removed from the live set synchronously when a tool redeemed it, but `GrantConsumed` was appended only after the cycle finished. A restart inside that window replayed the earlier `ApprovalGranted` record and admitted the identical call again.

The pre-fix journal-layer acceptance failed on its own assertion:

```
assertion `left == right` failed: a grant already consumed by its claimed turn must not re-arm before the drain
  left: 1
 right: 0
```

## Solution

Append a host-durable `GrantDispatched` claim after the continuation owns its execution lock and before the brain runs. Replay folds that claim out of `replayed_grants`. If the completed turn did not redeem the grant, append `ApprovalGranted` again so unused authority still survives an ordinary restart; a crash during an uncertain turn stays fail-closed.

The impossible GrantSet-only ignored test was replaced at the journal layer, where the replayed grant and pending `drain_consumed` state can both be observed. The prior passing cycle test that asserted re-arming was replaced with a deterministic two-barrier test that pauses after real redemption and reloads the journal before the cycle drain.

## Submission Checklist

- [x] Journal acceptance filter listed exactly 1 test; green with 1 passed, 0 failed under `openhuman,tinymemory`
- [x] Crash-window integration filter listed exactly 1 test; green with 1 passed, 0 failed under `openhuman,tinymemory`
- [x] Whole `runtime::` filter listed 1101 tests; green with 1100 passed, 0 failed, 1 ignored under `openhuman,tinymemory`
- [x] Existing unused-grant restart and concurrent-mint tests are green, 1 passed each
- [x] Red proof: removing only the pre-turn `record_grant_dispatched` loop made the crash-window test fail with left 1/right 0 and exit 101; restore matched the saved SHA-256 and the test returned 1 passed
- [x] `cargo fmt --check` — exit 0
- [x] `cargo clippy --all-targets --no-deps --features openhuman,tinymemory -- -D warnings` — exit 0
- [x] Production-deletion scan reviewed: only the superseded broad claim loop and bug-asserting tests were removed, with replacements in the same commits

## Impact

A single-use approval cannot be spent again after a restart in the consume-before-drain window. Unused grants remain restart-durable after a completed follow-up, while a crash during an indeterminate turn chooses loss of authority over duplicate execution.

## Related

PR #2226 owns standing-grant behavior. This change is limited to single-use grants and does not alter standing-grant replay or validation.

### DIFF
--- a/src/runtime/cycle.rs
+++ b/src/runtime/cycle.rs
@@ -661,17 +661,19 @@ impl<'a> CycleRunner<'a> {
             drop(guard);
             return Err(err);
         }
-        for id in events.iter().filter_map(|(_, event)| match event {
-            CompanyEvent::ApprovalResolved { approval_id, .. }
-                if self.rt.grants.peek(approval_id).is_some() =>
-            {
-                Some(approval_id)
-            }
-            _ => None,
-        }) {
+        let claimed_grants: Vec<GrantedCall> = events
+            .iter()
+            .filter_map(|(_, event)| match event {
+                CompanyEvent::ApprovalResolved { approval_id, .. } => {
+                    self.rt.grants.peek(approval_id)
+                }
+                _ => None,
+            })
+            .collect();
+        for grant in &claimed_grants {
             self.rt
                 .journal
-                .record_grant_dispatched(id, now_millis())
+                .record_grant_dispatched(&grant.approval_id, now_millis())
                 .await?;
         }
         let mut claimed: Vec<ApprovalContinuation> = Vec::new();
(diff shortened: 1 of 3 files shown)


oxoxDev:
@coderabbitai review

PR #2283, opened 2026-09-11
fix(desktop): open outward links in the operator's browser

### LINKED ISSUE #2282: Every external link is inert in the desktop build
## Symptom

In the packaged desktop app, **every external link does nothing**. No new window, no error, no console message — the click is swallowed.

Observed on "Manage API keys" and "Top up balance" in the connections/API-key surface, and on "Sign in with OpenRouter" in the first-run setup wizard. Reported as "all buttons from this family aren't working at all".

The same links work in the browser console, which is why this survives: it is desktop-only.

## Why

Every one of them is a plain anchor:

```tsx
<a href={account.manageKeysUrl} target="_blank" rel="noreferrer" data-testid="hub-manage-keys">
```

A browser opens a tab for that. A Tauri webview has no tab to open, and all three things that would rescue it are absent:

1. **The webview is granted no permissions at all.** `src-tauri/gen/schemas/capabilities.json` is literally `{}`, and there is no `src-tauri/capabilities/` directory. So a `shell:allow-open` call would be denied even if one were made.
2. **Nothing makes that call.** The frontend has no `@tauri-apps/plugin-shell` dependency and never invokes `openUrl()`. `tauri_plugin_shell::init()` *is* registered (`src-tauri/src/lib.rs:111`), so the plugin is compiled in and reachable from Rust — but nothing on the JS side can reach it.
3. **No navigation handler.** The builder chain in `src-tauri/src/lib.rs` installs no `on_navigation` / new-window interception, so nothing converts a `_blank` request into an OS-level open.

Net effect: the plugin is present, permitted to nobody, and called by nothing.

## Blast radius

16 anchors across 10 files, all `target="_blank"`:

| File | Lines |
|---|---|
| `views/connections/HubAccountLinks.tsx` | 46, 55 — *Manage API keys, Top up balance* |
| `views/connections/ApiKeyView.tsx` | 211, 256 |
| `views/setup/SetupWizard.tsx` | 1817 — *Sign in with OpenRouter / TinyHumans* |
| `views/Login.tsx` | 525, 534 |
| `views/FeedbackView.tsx` | 170, 203 |
| `components/feedback-form.tsx` | 181, 214, 229 |
| `components/feedback/board-row.tsx` | 87 |
| `components/title-bar-utilities.tsx` | 89 |
| `views/WorkspaceView.tsx` | 2852 |
| `views/finance/InvoicingView.tsx` | 325 |

## Why this one hurts more than a dead link usually does

`SetupWizard.tsx:1817` is on the **first-run wizard's API-key step**. A new operator who does not already hold a key is offered exactly one way to get one, and in the desktop build that button is inert. The `sk-…` field still accepts a pasted key, so anyone who already has one is fine — the escape hatch for everyone else is dead, on the first screen, in a build they just installed.

Same shape as the `OnboardingGate` hash-link problem: a link that reads as a link, changes nothing, and reports nothing.

## Suggested direction

A navigation/new-window handler in `src-tauri/src/lib.rs` fixes all 16 at once and keeps future anchors working without each author remembering a helper. The per-call alternative — add a capability granting `shell:allow-open`, take the `@tauri-apps/plugin-shell` dependency, and route through `openUrl()` with the anchor as web fallback — is more code and leaves the next new anchor dead by default.

Either way this wants a regression test that fails when an external link is inert in the packaged build, since nothing catches it today: the console E2E runs in a browser, where these links work.

## Secondary, unverified

`SetupWizard.tsx:133` sends OpenRouter sign-in to `redirect_url=https://openrouter.ai/workspaces/default/keys`, which assumes a workspace slugged `default` exists on the account. Both that URL and the older `openrouter.ai/keys` return 200, so this is not a dead URL — but if a fresh account has no `default` workspace, a successful sign-in would land somewhere unhelpful. Worth confirming against a real new account; not the cause of this issue.

oxoxDev:
## Summary

Every outward link in the console did nothing in the packaged desktop app — no window, no error, no console message. All 16 of them, across 10 files.

Each is an ordinary anchor:

```tsx
<a href={account.manageKeysUrl} target="_blank" rel="noreferrer">
```

A browser opens a tab for that. A Tauri webview has no tab to open, and all three things that would have rescued it were absent:

1. **The webview held no permissions at all.** There was no `src-tauri/capabilities/` directory, so the generated capability set was literally `{}` and every plugin command was denied.
2. **Nothing asked the shell to open anything.** `tauri_plugin_shell::init()` is registered (`src-tauri/src/lib.rs:111`), so the plugin was compiled in and reachable from Rust — and unreachable from the console.
3. **No navigation handler**, so nothing converted a `_blank` request into an OS-level open.

The plugin was present, permitted to nobody, and called by nothing.

The worst instance is the first-run wizard's API-key step: an operator without a key was offered exactly one route to get one, and in the desktop build that button was inert — on the first screen of a build they had just installed. Anyone who already held a key could paste it and never notice, which is how this shipped.

Also adds the button the Composio BYOK field was missing: its copy read "From your Composio dashboard at app.composio.dev", an address to retype rather than somewhere to go.

Closes #2282

## API Or Behavior Changes

A left-click on an `http`/`https`/`mailto:` anchor in the **desktop shell** now opens it in the operator's browser. In a browser nothing changes: `tauriCore()` is `null` there, so the anchor's own behaviour stands.

The capability grants `shell:allow-open` and nothing else. Capabilities are additive, so adding the first one takes nothing away.

The plugin's own `open` scope needed no configuration: `plugins.shell.open` unset means `Unset`, which the plugin treats as enabled with the default regex `^((mailto:\w+)|(tel:\w+)|(https?://\w+)).+`.

### Why a delegated listener rather than a helper each anchor calls

A helper fixes the anchors whose author remembers to call it. The next one written would be dead again — silently, and only in a packaged build, which is exactly how 16 accumulated. One capture-phase listener covers the anchors that exist and the ones nobody has written yet.

It is installed in `main.tsx` beside `startScrollActivity()`, for the reasons already written there: one capturing listener for the life of the document, outside React so StrictMode cannot arm it twice.

- **`tauriCore()` is probed at click time, not install time.** A listener that decided "web" once, before the bridge was injected, would stay wrong for the life of the window.
- **It reuses the existing bridge** (`api/transport/bridge.ts`) rather than taking `@tauri-apps/plugin-shell`, so no new dependency.

Hash routes, relative paths, `blob:` and `data:` are left alone — handing any of them to the OS would replace a working in-app link with a failed one. Modified clicks pass through. A refusal from the shell is swallowed rather than surfacing as an unhandled rejection.

The Composio button points at the bare `https://app.composio.dev` the field's own copy names, not a deep link to the settings page that mints the key. That path is the vendor's to move, and a stale one strands the operator on a 404 *after* a sign-in that worked.

## Tests

`frontend/test/unit/external-links.test.ts` — 8 cases over both runtimes: opens in desktop, **inert in a browser**, hash route untouched, modified click untouched, refusal swallowed, dispose stops intercepting, and the scheme predicate.

The browser case matters beyond symmetry: the console E2E suite runs in a browser, where all 16 links already worked. That is why nothing caught this, and why the fix must not disturb that path.

```
npm run typecheck        pass
npm run typecheck:unit   pass
npm test                 5361 passed | 3 skipped (613 files)
prettier --check         clean
```

**Verified by hand in a packaged build**, which is the only thing that can prove this: a local release `.app` was built from this branch, installed, and an outward link opened the operator's browser.

One finding worth carrying, because it cost a round trip here: **replacing `/Applications/<app>.app` does not replace a running instance.** macOS keeps executing the already-loaded binary, so the first verification attempt was clicking in the old app and reported the bug as unfixed. Kill the running process before testing a rebuilt bundle.

## Documentation

None needed — no documented behaviour changes.

### DIFF
--- a/frontend/src/lib/external-links.ts
+++ b/frontend/src/lib/external-links.ts
@@ -1,0 +1,73 @@
+import { tauriCore } from "@/api/transport/bridge";
+
+/**
+ * Hand an outward link to the operator's browser when the console is running
+ * inside the desktop shell.
+ *
+ * # Why this exists
+ *
+ * Every outward link in the console is an ordinary anchor —
+ * `<a href={url} target="_blank" rel="noreferrer">`. In a browser that opens a
+ * tab. A Tauri webview has no tab to open and no browser to open it in, so the
+ * click does nothing at all: no window, no error, no console message. The
+ * operator sees a link that is not a link.
+ *
+ * That was every one of them, on every screen, including the first-run wizard's
+ * only route to an API key — so an operator who did not already hold one had no
+ * way forward on the first screen of a build they had just installed.
+ *
+ * # Why a delegated listener rather than a component
+ *
+ * A helper each anchor calls fixes the anchors that remember to call it. The
+ * next one written is dead again, silently, and the failure only shows up in a
+ * packaged build. One capture-phase listener covers the anchors that exist and
+ * the ones nobody has written yet.
+ *
+ * # Why it is inert on the web
+ *
+ * `tauriCore()` is `null` in a browser, so nothing is intercepted and the
+ * anchor's own behaviour stands. The console keeps working in a normal tab
+ * exactly as before, which is also what keeps the E2E suite meaningful.
+ */
+export function installExternalLinkOpener(doc: Document = document): () => void {
+  const onClick = (event: MouseEvent) => {
+    if (event.defaultPrevented || event.button !== 0) return;
+    // A modified click asks for the platform's own behaviour. Nothing to
+    // improve on, and hijacking it would be worse than the bug.
+    if (event.metaKey || event.ctrlKey || event.shiftKey || event.altKey) return;
+
+    const anchor = (event.target as Element | null)?.closest?.("a");
+    const href = anchor?.getAttribute("href");
+    if (!anchor || !href) return;
+    if (!isOutwardHref(href)) return;
+
+    // Probed at click time, not at install time: a listener installed before
+    // the bridge is injected would otherwise decide "web" once and stay wrong
+    // for the life of the window.
+    const core = tauriCore();
+    if (!core) return;
+
+    event.preventDefault();
+    // The command is fire-and-forget on purpose. A rejection here means the
+    // shell refused the open, and there is nothing this layer can do about it
+    // that is better than leaving the operator where they were — but it must
+    // not surface as an unhandled rejection.
+    void core.invoke("plugin:shell|open", { path: href }).catch(() => {});
+  };
+
+  doc.addEventListener("click", onClick, true);
+  return () => doc.removeEventListener("click", onClick, true);
+}
+
+/**
+ * Whether this `href` leaves the console.
+ *
+ * `http`/`https` and `mailto:` go out. An in-app hash route, a relative path
+ * and a `blob:`/`data:` URL do not, and handing any of them to the operating
+ * system would replace a working link with a failed one.
+ */
+export function isOutwardHref(href: string): boolean {
+  const value = href.trim();
+  if (!value || value.startsWith("#") || value.startsWith("/")) return false;
+  return /^(https?:|mailto:)/i.test(value);
+}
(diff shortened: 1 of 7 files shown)


oxoxDev:
@chatgpt-codex-connector, please re review and approve

oxoxDev:
@coderabbitai please approve

PR #2286, opened 2026-09-12
Routing states: stop inferring a mode the company cannot use, and ask for a model on add

graycyrus:
**Stacks on #2262 — merge that first.** This branch is cut from
`origin/docs/inference-rework` @ `e0715e600` and assumes the provider-list
rework and the provider/API-mismatch work are already in. It targets `main`
so the diff is reviewable on its own; the base moves to whatever #2262 lands as.

Implements [`routing-states-design.md` + `routing-states-matrix.md`].

## The report, and what was actually wrong

A fresh company adds Anthropic as its first provider and every turn 404s on
`model: agentic-v1`. The 404 is a symptom of two independent defects:

1. **Routing defaulted to a mode the company cannot use.** An empty route table
   *displayed* as Managed and *resolved* to `Resolution::Primary` — the first
   enabled provider. The screen named one destination and the turn used another.
2. **Nothing ever asked which model the new provider should serve.**
   `add_provider` wrote `models: BTreeMap::new()` and the wire shape had no
   `model` field at all, so there was no way to supply one. `model_for_tier`'s
   `Unknown` arm then put the bare tier on the wire, correctly and honestly.

## The open decision (§4) — how it is implemented, and what to look at

**Guarded auto-route**, with the corrected condition. `auto_route_sole_provider`
(`src/server/ops/inference/providers.rs`) writes all four rows when, and only
when, **all three** hold:

- the route table is **empty** — nothing authored is overwritten;
- the managed chain **does not resolve** — there is no fallback behind the rows;
- after the add there is **exactly one enabled provider**, and it is this one.

"First provider" is not the test, because of entry zero: a company can hold a
provider it never added, so the new one can be the *second* list element and
still be what the operator expects to be used.

**It deliberately does nothing when Managed is available** (row B2). That is
where guessing moves money: a company on Managed adding an OpenRouter key may
want it for one workload, for vision only, or to compare. The table stays empty,
the mode reports `unset`, and the screen asks.

**Auto-routing alone would have traded the 404 for the same 404 with a route
attached**, so §6.1 is closed in the same PR: the draft probe's published
catalogue rides back on `ProbeResultDto` (`models`, `needsModel`), the add dialog
probes **before** anything is written and asks for a model id with that
endpoint's own list in a datalist, and the host refuses an add that would store a
row it knows cannot answer. The rule is read from **what the endpoint published**
(`TierVocabulary::from_catalog_ids`), never from its kind — a self-hosted LiteLLM
publishing `agentic-v1` resolves tiers whoever runs it.

Both halves are visible and reversible: the auto-route reports its tiers in
`affectedTiers` and says so in the note, and every row it writes is editable on
the Routing tab. It also **refuses to write anything the resolver would not read
back as this provider** — a route is persisted as the text an operator would
type, so a slug colliding with a word in that grammar (`local`, `managed`,
`default`, none of which `is_reserved_slug` covers) would round-trip into a
different ref. Writing rows nobody asked for is defensible only while it is
certainly right.

The condition itself is a pure function, `is_the_only_thing_that_can_answer`,
with a test per clause — including the entry-zero case that makes "first
provider" the wrong test.

## What else landed

- **Mode inference** — `infer_routing_mode(routes, managed_resolves)` returns a
  new `Unset` when the table is managed-or-empty and managed resolves to nothing.
  The console renders no selection plus a sentence naming where work actually
  goes, and the Managed mode row is no longer selectable when it cannot answer.
- **The console's `inferRoutingMode` and `orphanedRoutes` are deleted**, with the
  six unit tests that pinned them. Neither was ever called — the rendered mode
  comes from `RoutesDto.mode`.
- **`managedRow("none")` is reachable.** "No credential resolves — agents cannot
  think" is now a constant used by the two dead-end states that need it.
- **`Primary (Managed)`** and **"leaves Managed as the only thing that can
  answer"** stop being printed when managed resolves to nothing.
- **Badges** — `In use` / `Parked` on the provider row, from the `routingMap`
  already in hand.
- **Disable confirmation** as `intent: "disable"` through `removalWarnings`,
  reversible throughout, `Continue` rather than a destructive verb.
- **`parked_tiers` matched on slug only** — so disabling the only Ollama reported
  "Nothing was routed through it", which was false. It, `orphaned_routes` and
  `scrub_removed` now share one matcher; `orphaned_routes` also stops skipping
  slug-less refs entirely.
- **`Resolution::Disabled` and `Resolution::Missing` render** on the workload
  rows; the orphan banner is worded in workload names, not tier ids.
- **`origin` on `ProviderDto`**, so the console stops rendering the three
  controls entry zero answers with a 400.
- **`MODEL_UNAVAILABLE_SIGNATURES`** matches Anthropic's `not_found_error`, so
  the repair advice reaches the provider people connect first.

## What deliberately did not change (§7)

`set_enabled` still does not scrub routes; `Resolution::Disabled` stays a
reported state rather than a silent demotion; the mode stays inferred, never
stored; `model_for_tier` still sends the bare tier for `Unknown`; the default
marker is still cleared rather than moved; entry zero stays where it is.

## Found and not fixed — reported, per the brief

- **§6.2 — Managed's on/off switch is inert on the turn path.** `managed_enabled`
  is read in exactly one place, the status DTO. A route to `managed` reaches
  Managed whether the switch is on or off, and the Providers tab meanwhile prints
  "Managed is switched off, so it is not a fallback". Needs a decision: gate the
  turn path on it, or remove the control.
- **D7 — `revert_config` is not in line with `delete_provider`.** It clears
  `inference/config` and `inference/key` but does not scrub routes naming entry
  zero's slug, clear the default marker, or clear `provider/<slug>/key`.
- **§3.4, the mixed row — `managed` on some rows and unset on others** reads as
  one mode while the rows split two billing destinations: the `managed` rows go
  to Managed, the unset rows to `Primary`. Unchanged here; the mode only stops
  reading Managed when *nothing* in the table can answer.
- **D8 — entry zero and an indexed provider can take the same slug** via
  `PUT …/inference`, which validates only kind and URL. No test covers it.
- **C5 — `route_is_servable` accepts a route naming a disabled provider.** Left
  accepting, which is now consistent: the parked state renders.

## CI, by head SHA

`14c24a2f2b8c1da746a3ac1bac8811afc324cabf`, run
[34689618292](https://github.com/tinyhumansai/opencompany/actions/runs/34689618292):
**12 green, 0 pending**, and the two Console E2E lanes red.

**Those two lanes are already red on the base this stacks on.** Run
[34676447897](https://github.com/tinyhumansai/opencompany/actions/runs/34676447897)
on `e0715e600` (#2262's head) fails `Console E2E` and `Console E2E (live brain)`,
and the failing-test sets are **identical** — the same five, compared by
downloading both runs' `playwright-failures` artifacts:

```
connections-authority-…-control-across-the-four-pages
inference-a-second-provider-holds-a-credential-of-its-own
inference-deleting-a-provider-…-resets-the-routes-that-named-it
inference-Managed-is-always-…-badge-rather-than-a-switch
inference-the-add-dialog-stops-offering-a-provider-once-it-is-connected
```

Two causes, neither of them this branch's: the Managed row's copy no longer says
`"Always on"` (that is #2262's own change, and the assertion was not updated),
and the rest are leftover state in the shared E2E company — `E2E One` and `Groq`
are already connected from previous runs, so the slug collides and the add dialog
no longer offers Groq.

**`Console E2E (first run)` — the lane that runs against a clean company — is
green**, which is the one that actually exercises this branch's add flow, the new
model step included.


graycyrus:
### Three findings routed here from #2289

#2289 is stacked on this branch, so its GitHub base is `main` and its diff shows all of #2262 and #2286. Codex reviewed the whole stack and raised three findings whose code is **this** PR's (`173f7a3a7` and `dc8f8d1b6`), not #2289's. I verified each against the current tree and replied on #2289 rather than acting on them there; recording them here so they are not lost when that PR merges.

**1. `managed_resolves` ignores the Managed switch** — [comment](https://github.com/tinyhumansai/opencompany/pull/2289#discussion_r3996735971)
`managed_resolves` (`src/server/ops/inference.rs`) returns `managed_state(..).configured` only, and `get_routes` passes it to `infer_routing_mode`. So a company whose switch is off can be reported `mode: "managed"` — a mode the turn path now refuses and, as of #2289, the boot path refuses too. Candidate: `configured && enabled`.

**2. An all-default table reports `managed` while turns go to the primary** — [comment](https://github.com/tinyhumansai/opencompany/pull/2289#discussion_r3996773506)
`provider_for_workload` resolves an absent row through `Resolution::Primary`, so a company with Managed configured *and* a BYOK provider connected renders Managed while turns bill the BYOK account. Note this is currently **pinned deliberately**: `the_routing_mode_is_inferred_from_the_routes` asserts `routes["mode"] == "managed"` after adding a custom provider with the table still empty. So it is a design call to re-open rather than a defect to patch — but it is the same "the screen names one destination and the turn uses another" this PR was opened to fix, one case along.

**3. `nothingCanAnswer` and `primaryLabel` read only `configured`** — [comment](https://github.com/tinyhumansai/opencompany/pull/2289#discussion_r3996735973)
A configured-but-switched-off Managed reads as available, so `ProvidersTab` suppresses "agents cannot think" and the disable/removal warnings promise a fallback that will refuse. #2262 has already widened `removalWarnings` to `managed?: { configured?, enabled? }` (`838bef04d`); these five call sites are the rest of that same widening, and they need copy as well as a predicate — "not set up" and "switched off" are different sentences to an operator.

**4.** Separately, `auto_route_sole_provider` ends `store::save_routes(..)?`, so a failed routes write turns a committed add into a 5xx and the retry then reports the slug as taken — [comment](https://github.com/tinyhumansai/opencompany/pull/2289#discussion_r3996753808). Best-effort route, or roll the provider back so the error is truthful.

graycyrus:
### The `inference-save` testid this rewrite retired leaves one spec red and one assertion vacuous

Seen on #2289's `Console E2E (live brain)`, which inherits this branch. `connections-authority.spec.ts` still asserts on `inference-save`, which the per-row Providers tab no longer renders:

- **Line 211 fails.** `expect(getByTestId('inference-save')).toBeVisible()` — element not found, 30s timeout. This is the documented inherited failure on #2289.
- **Line 169 passes for the wrong reason.** The member case asserts `toHaveCount(0)` on the same testid, which is now trivially true for *every* role. The spec's own comment three blocks up says what that costs: *"This is what stops that assertion going vacuous: if the field ever stops rendering here, this fails rather than the member case quietly passing."* That is precisely what has happened, on the one page where it was not paired with a positive assertion.

The write control the pair is about is now `inference-add-open`, which renders for both roles and carries `disabled={!canManage}` — so the replacement is `toBeEnabled()` for the admin and `toBeDisabled()` for the member, rather than a presence check either way.

I have deliberately not made that change on #2289: it is this branch's rewrite that orphaned the testid, and an authority assertion is not something to rewrite speculatively from a stacked bug fix.

---

**Separately**, one E2E failure that #2289 *did* fix, because it only appeared once this branch made the preceding test pass: [`0958906af`](https://github.com/tinyhumansai/opencompany/commit/0958906afb72296c50432a5d592bd3244c4adda3). `inference.spec.ts` → "a custom provider may not take a name the catalogue ships" typed **Groq** and expected the built-in refusal, but `checkSlug` reports `taken` before `reserved` and every test in the file shares one company — so once "the add dialog stops offering a provider once it is connected" succeeds in connecting Groq, the assertion reads the collision rule instead of the reservation rule. It now types Cerebras, a catalogue row no test connects. Worth carrying here too if this branch lands first.

Issue #1737, opened 2026-08-25
Inference card shows one provider in its header and another in its select; probe sends an unauthenticated request


graycyrus:
## What happens

After saving an OpenRouter key and pressing **Test**, the probe fails:

> Inference probe failed: inference returned 401 Unauthorized: {"error":{"message":"Missing Authentication header","code":401}}

`Missing Authentication header` means **no `Authorization` header was sent at all** — not a rejected key. A wrong key returns a different error from OpenRouter.

## The card contradicts itself on the same screen

In one screenshot, simultaneously:

- Header: **`OpenRouter`** · `runtime` · **✓ key set**
- Endpoint: `https://openrouter.ai/api/v1`
- Provider dropdown: **`Managed (TinyHumans)`**
- Status line: `Cognition: echo · no model runs on this path, so Usage stays at zero`

The header says the stored provider is OpenRouter with a key set. The select says the provider is Managed (TinyHumans). Both are rendered from the same card at the same moment.

## Why that likely produces exactly this 401

The card's own help text states the rule:

> "This field is scoped to the provider selected above and is stored against it — a key for any other vendor will fail at the first turn, so it is not the place for one."

So a key stored against one provider is not found when resolution happens under another. The probe resolves via `resolve_effective(runtime.id(), &manifest, None, secrets)` (`src/server/ops/inference.rs:592`) — tenant config, no platform default. If the resolved provider is `managed` while the endpoint is OpenRouter's, there is no key for that pairing, so the request goes out unauthenticated and OpenRouter answers exactly as observed.

**Flagged as inferred, not verified:** I have not confirmed which of the two the resolver actually picked, or whether the select simply failed to rehydrate the saved provider after Save. Either way the visible contradiction is itself the defect — the operator cannot tell which provider their key was stored against.

## Two things to fix, possibly one cause

1. **The Provider select must reflect the saved provider.** If it resets to a default after Save while the header shows something else, a user can store a key against a provider they never chose and get an unauthenticated request with no way to see why.
2. **The probe should refuse rather than send an unauthenticated request.** If resolution yields an endpoint with no key for that provider, that is a configuration error the product can detect locally. Reporting "no key is stored for OpenRouter" is far more useful than relaying a vendor's 401, and it costs the user nothing — note the card warns Test "sends one real message… and your provider may charge for it."

## Acceptance

- The Provider select always shows the provider the stored key is scoped to; header and select can never disagree.
[...]

PR #2287, opened 2026-09-12
fix(inference): the managed card's key rides the platform endpoint

oxoxDev:
## Summary

Selecting the TinyHumans (managed) inference card and saving a TinyHumans API key sent every turn to `openrouter.ai`, which rejected the key. The onboarding probe passed, so the company looked correctly configured right up until an agent was asked to do anything.

## Problem

`resolve_effective_scoped` normalized the provider *before* resolving the endpoint:

```rust
let provider = normalize_provider(&runtime.provider).to_string();   // "managed" -> "openrouter"
let (base_url, credential, proxied) = resolve_endpoint(&provider, ...);
```

`resolve_endpoint`'s first branch is `if is_managed_choice(provider)` — the branch that pins the platform endpoint and lets the operator's own key ride it. Normalization erased `managed` before that branch could ever match, so it was unreachable from runtime resolution.

The next branch (`openrouter && !has_key`) would still have inherited the platform URL, but a key was present, so resolution fell through to `effective_base_url` and produced `https://openrouter.ai/api/v1`. A TinyHumans key then reached OpenRouter, which answers an unparseable credential with `Missing Authentication header` — the phrasing issue #1737 already records as reading like "nothing was sent".

**Why the probe disagreed.** `decl_for_probe` passes the raw kind; only runtime resolution normalizes first. The two paths answered differently for the same configuration, which is why onboarding verified a setup that could not work. The manifest tier had the same fold.

## Solution

Resolve both tiers on the raw provider kind. The decl still reports the normalized value, so storage, telemetry and the console's provider field are unchanged.

The rule this restores: **the card decides the endpoint, not whether a key is set.** The managed card is the TinyHumans one, so a key saved on it is a TinyHumans key and rides the platform endpoint. Going direct to OpenRouter is what selecting the `openrouter` provider does.

A declared `base_url` is now honoured on the managed kind, and the probe drops a stale one from the form before calling. Previously `resolve_endpoint` discarded both — correct for a form where the URL is leftover state from a previously-picked provider, wrong for a manifest where the operator typed it. Splitting them is what lets the manifest tier resolve on the raw kind without losing a real endpoint.

## Submission Checklist

- [x] Full lib suite under `--features opencompany-core/openhuman` — **7657 passed, 0 failed**
- [x] Red proof: restoring the normalize-first order fails the new test on its own assertion — `a saved managed choice must reach the platform endpoint, never openrouter.ai / left: "https://openrouter.ai/api/v1" / right: "https://env.example/openai/v1"` — then restored to green
- [x] `cargo clippy --all-targets --no-deps --features opencompany-core/openhuman,opencompany-core/tinymemory -- -D warnings` clean
- [x] Production-deletion scan: three replaced resolution lines and three assertions that pinned the old reading
- [ ] Console E2E — checked on CI

## Impact

A company that saved the managed card with its own TinyHumans key now reaches the platform endpoint instead of being refused on every turn. Companies on the managed card with no key are unaffected. Selecting `openrouter` with a key still goes direct.

Three existing assertions encoded the opposite reading — that any saved key meant the tenant's own OpenRouter account pays. They are replaced, and the status slug for managed-with-key is now `subscription` rather than `openrouter`, which is what the traffic actually is.

## Related

`a_saved_managed_config_with_an_own_key_still_reaches_the_platform_endpoint` is the runtime twin of the existing `managed_probe_with_own_key_keeps_the_managed_endpoint`. The probe had coverage for this combination; runtime resolution never did, which is how the two paths drifted apart unnoticed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Managed provider configurations now honor explicitly configured endpoints while retaining the appropriate platform default.
  * Managed credentials consistently use the platform endpoint instead of being redirected to an alternate provider endpoint.
  * Provider checks no longer inherit stale endpoint values from previously selected providers.

* **User Interface**
  * Managed provider guidance now identifies the required platform API key.
  * Provider labels accurately reflect the active configuration.
  * An unsaved-changes indicator appears when the form differs from the running configuration.

---

## Console half (added after the host fix was confirmed on a local DMG)

The card named two different routes for one configuration. The header rendered the label while the Provider select seeded from `provider` — the *normalized* kind — so a managed company came back as `openrouter` and the select rested there while the header said TinyHumans.

Both now read `slug`, the host's own statement of which route the traffic takes, so they cannot disagree.

**The key hint asks for a TinyHumans key again.** It had been changed to *"an OpenRouter key (`sk-or-…`) — the managed brain is OpenRouter"*, with a comment recording that it previously read "a TinyHumans API key". The console had been edited to match the routing defect instead of the routing being fixed, so operators were told to paste the one vendor's key that card would never send it to.

**An edited form now says it is unsaved.** The header states what the company runs; the form states what it would switch to. They legitimately differ mid-edit, and without a cue that gap is indistinguishable from the card contradicting itself — which it really did, so the ambiguity was costly.

### Coverage

Unit (`inference-card-honesty.test.ts` — the file whose docstring is *"the card knew a fact about the host and rendered something that contradicted it"*):

- managed proxy → header **and** select both read `Managed (TinyHumans)`
- genuinely direct OpenRouter → both read `OpenRouter` (so this cannot be "fixed" by hardcoding the managed label)
- no unsaved cue when the form matches the host
- unsaved cue once the draft diverges

E2E (`inference.spec.ts`):

- a managed company **with its own key** — the combination that broke — saves, and the host reports `slug: subscription` with a `baseUrl` that is not openrouter.ai, while both card fields read `Managed (TinyHumans)`
- an edited form shows the unsaved cue while the header keeps reporting the running config

Managed *without* a key was never broken, so both new tests set one: the key is what skipped the keyless inheritance arm.

### Verification

- **5373 unit tests passing**, typecheck clean
- Red proofs: reverting the select seeding gives `expected 'OpenRouter' to be 'Managed (TinyHumans)'` — the exact on-screen symptom; dropping the typed-key term from the dirty check fails the unsaved test. Both restored.
- Confirmed on a locally built DMG: endpoint `api.tinyhumans.ai/openai/v1`, `Cognition: harness`, agent responds.

### Known gap, deliberately not in this PR

`savedIsProxied` still derives proxied-ness as `!(provider === "openrouter" && keyConfigured)` — the pre-fix rule, now wrong for managed-with-key. Switching it to `slug` breaks 10 catalog tests whose fixtures omit that field. Practical effect: the model picker may offer raw OpenRouter ids that the platform's tier-native endpoint will not accept; tier ids (`chat-v1`) are unaffected. Worth its own change with the fixtures updated properly.

### DIFF
--- a/frontend/src/views/connections/InferenceSection.tsx
+++ b/frontend/src/views/connections/InferenceSection.tsx
@@ -87,17 +87,12 @@ const PROVIDERS: Record<
     label: "Managed (TinyHumans)",
     acceptsKey: true,
     requiresBaseUrl: false,
-    // The managed brain is OpenRouter with the platform paying: the host treats
-    // `managed` as a legacy alias for `openrouter` (`LEGACY_MANAGED`, since
-    // OpenCompany stopped exposing its own model SKUs), so a key saved here is
-    // resolved onto OpenRouter's own endpoint and sent there.
-    //
-    // This line used to read "a TinyHumans API key" — true when `managed` was a
-    // provider of its own, and never updated when it stopped being one. It is
-    // what issue #1737 actually cost: the operator was asked for one vendor's
-    // key on a card that stored it against another, and every turn and every
-    // Test since has presented it to OpenRouter, which rejects it (issue #1737).
-    keyKind: "an OpenRouter key (`sk-or-…`) — the managed brain is OpenRouter, so that is where a key set here is sent",
+    // The card decides the endpoint: `managed` is the TinyHumans one, so a key
+    // saved here is a TinyHumans key and rides the platform endpoint on this
+    // company's own account. Going direct to OpenRouter is what selecting the
+    // `openrouter` provider does.
+    keyKind:
+      "a TinyHumans API key — the managed card runs on the platform endpoint, so that is where a key set here is sent",
     preset: { baseUrl: "", models: {} },
   },
   openrouter: {
@@ -534,8 +529,16 @@ export function InferenceSection({
    * reasoning as `removeKey` below.
    */
   const seedFromStatus = useCallback((next: InferenceStatus) => {
+    // `provider` is the normalized kind, so a company on the managed card reads
+    // back as `openrouter` and the dropdown would rest on OpenRouter while the
+    // header names the managed route. The host reports which one the traffic
+    // actually takes, so one field decides both and they cannot disagree.
     const stored = (
-      next.provider in PROVIDERS ? next.provider : "openrouter"
+      next.slug === "subscription"
+        ? "managed"
+        : next.provider in PROVIDERS
+          ? next.provider
+          : "openrouter"
     ) as InferenceProvider;
     // The form is a "switch to" form, so it opens on something the operator can
     // actually complete. A company on a route this console does not offer has no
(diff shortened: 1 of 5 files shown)


oxoxDev:
Superseded by work already on main.

The host fix landed independently in `fix(inference): four on the managed and scoped credential paths` (e1494ae77) — same change, same reasoning: resolve on the raw provider kind, because `normalize_provider` folds `managed` onto `openrouter` and skips both managed branches, sending a TinyHumans token to openrouter.ai.

The console half targeted `views/connections/InferenceSection.tsx`, which the inference rework deleted; managed is now its own row rather than a value in a provider select, so the header-vs-select disagreement this fixed cannot occur in that shape.

Coverage is present on main too — `managed_with_a_pasted_inference_key_uses_it` and `managed_falls_back_to_the_company_account_key_and_keeps_the_platform_endpoint`. Nothing here is worth carrying forward.

PR #2316, opened 2026-09-15
feat(server): ask the hub for the connections scope on a key grant

oxoxDev:
## Summary

A desktop or self-hosted OpenCompany install cannot use managed Composio. Every
call it makes to `/agent-integrations/composio/*` is refused with `API key is
missing the required scope: connections` — reproduced live against prod on
2026-09-15. The key the console holds is real and works for inference; it simply
never carried that scope, because the hub withheld `connections` from every
callback origin that is not a provisioned tenant.

`tinyhumansai/backend#1332` opens that path: a loopback console may now be
granted `connections`, but **only when the grant explicitly asks for it**, and
the hub reads the ask from a `scopes` query parameter on `GET /auth/key` that it
pins into the OAuth state beside the PKCE challenge, so the consent screen names
the scope before anything is minted. Nothing on this side was asking, so #1332
is inert without this change and this change is inert without #1332.

**Both must be merged, and the backend must be deployed to prod, before the
desktop 403 clears.** Prod last deployed 2026-09-11 and is stale as of this
writing, so merging both PRs changes nothing a user can see until that deploy
happens.

The ask is added to the shared query builder rather than to a call site. That
split already exists so the API's `GET /auth/key` and the site's `/connect`
provider chooser cannot disagree about the challenge; the scopes ask needs the
same guarantee for a stronger reason — it decides what the key may do, so a
scope present on one path and absent on the other would mint a key whose reach
depended on which page the browser happened to go through.

Only `connections` is requested. Every grant carries the human-mintable set
whether it asks or not, so asking for anything else would widen a key past what
this console spends.

## API Or Behavior Changes

The outbound key-grant authorize URL gains one query parameter. It is built once
in `key_grant_query`, so both readers of that query carry it:

```
…/auth/key?callback_url=…&code_challenge=…&code_challenge_method=S256&name=…&scopes=connections
```

The value is a comma-separated scope list, percent-encoded like every sibling
value, matching the parser `tinyhumansai/backend#1332` added (which also accepts
a repeated `?scopes=a&scopes=b`).

Safe to land before the backend deploys: a backend predating #1332 ignores an
unknown query parameter, so the grant it serves is byte-for-byte the one it
serves today and the key it mints is unchanged. No console behaviour changes
until the backend side is live — no new route, no new stored field, no new
secret-store key, and the credential fan-out is untouched.

Once both sides are live, a key minted through this flow on a non-tenant origin
carries `connections`, and the approver sees it named on the hub's consent
screen before approving. Keys already minted are unaffected; clearing and
re-linking the credential is what picks up the wider scope. `POST /api-keys`
still cannot mint `connections`, so a hand-pasted key does not reach this
surface.

## Tests

`key_grant_query` had no test pinning its output, so the query string is now
pinned whole rather than probed:

- `hub_identity::a_key_grant_query_asks_for_its_scopes_beside_the_challenge` —
  the exact string the hub receives, every value percent-encoded the same way.
- `hub_identity::the_scopes_asked_for_name_connections` — the ask names the
  scope Composio enforces.
- `hub_identity::both_readers_carry_the_same_built_query` — `key_grant_url` and
  `connect_url` both carry the builder's output verbatim, so neither may append
  a parameter of its own.
- `company_key::starting_a_link_asks_the_hub_for_the_connections_scope` — the
  ask survives to the URL the console actually navigates to, once, in the
  builder's position.

The first and last were confirmed to fail with the parameter dropped from the
builder. The existing start-link test, which pins the challenge and asserts the
verifier never leaves the host, is unchanged and still passes — no assertion was
relaxed to accommodate the new parameter.

- [ ] `cargo build --all-targets` — not run as such; `cargo clippy --all-targets`
      above compiled every target in the workspace.
- [ ] `cargo test` — not run in full; a bare `--lib` run exhausts memory on this
      machine. Ran scoped instead, all green:
      `server::hub_identity` (4), `server::hub_account` (6),
      `server::ops::company_key` (45).

Not verified end to end, and this PR does not claim the 403 is fixed: proving it
needs backend #1332 merged **and** deployed to prod. The outbound authorize URL
was not captured from a running desktop app either — the dev script pins a port
already held by another checkout, so attaching to it would have proven nothing
about this branch. The pinned query-string tests above are the evidence that the
parameter is on the wire.

## Documentation

None needed. `key_grant_query` and the new `KEY_GRANT_SCOPES` constant carry
their own doc comments explaining what the ask decides and why it lives in the
shared builder; no spec or module doc in `docs/` describes the grant query
string.

## Summary by CodeRabbit

- **New Features**
  - Authorization links now explicitly request the `connections` scope.
  - Hub, site, and credential-link authorization URLs use a consistent scope parameter.

- **Bug Fixes**
  - Improved URL encoding and query construction for key authorization requests, ensuring the requested scope is included correctly and only once.


oxoxDev:
@coderabbitai review

PR #2318, opened 2026-09-15
fix(skills): report the global baseline from both skill read paths

oxoxDev:
## Summary

The global skill baseline (`globals/globals.toml` → `[skills].always`) is installed into every
company and reaches every agent, but no read path reported it. `EffectiveSkills::materialize`
folded the baseline in as layer 0; `list_skills` and the GraphQL `resolve_company` merged only
company bundles and store deltas. Two user-visible consequences:

- three skills every agent has were invisible in the console, and there was no control to disable
  one — the row that carries the switch never rendered, even though `PUT …/skills/{slug}` with
  `enabled=false` would have worked;
- the Registry tab offered **Install** on those same slugs, because `installedIds` never contained
  them. Clicking wrote a delta byte-identical to what the agent already had and toasted success for
  a no-op.

The same asymmetry hid manifest opt-outs: `[globals].disable = ["skill:…"]` was synthesized into
disabling deltas for the harness only and never reached either reader.

**The fix is one derivation, not a third transcription.** `company::skill_effective::resolve` now
owns "what a company's effective skills are" — the global baseline, the committed bundles, the
operator's deltas, the registry heal, and the manifest opt-outs — and all three callers read it:

- the harness writes out its **enabled** entries (`EffectiveSkills::materialize` no longer re-derives
  the layering; `globals_skill_disables` moved alongside the resolver);
- `GET …/skills` and `Company.skills` project the **whole** set, disabled entries included, because
  a disabled skill's row is the only thing that can turn it back on.

Provenance needed no new `SkillSource` variant and no new DTO field. A global is reported as
`company`: it is a baseline install the operator did not add, it cannot be uninstalled (the
`uninstall` route already refuses a non-`Registry`/`Custom` row), and — decisively — `set_enabled`
already defaults a first toggle of a global to `SkillSource::Company`, so any other choice would
have made a row's source flip the moment someone used its switch.

## API Or Behavior Changes

- **`GET …/skills` and `Company.skills` return rows they never returned before.** The three baseline
  skills now appear in every company, including a platform-provisioned tenant with no checkout. The
  console's Registry tab consequently shows them as Installed rather than offering a no-op Install.
- **A global skill becomes disable-able through the existing `PUT …/skills/{slug}` route**, where
  previously no row existed to carry the switch. No new route, no second write path.
- **A manifest `[globals].disable` entry is honoured by both readers**, reported as a disabled row
  rather than a hidden one.
- **GraphQL now refreshes a skill's display fields from a delta's `custom_doc`** when the slug also
  has a company bundle. REST already did this (`enabled`, `source`, name, description, category,
  version); GraphQL copied only `enabled` and `source`, so the two transports disagreed on the same
  company. They now agree, pinned by a test.
- **A registry delta with no snapshot at all contributes no document.** GraphQL previously fell back
  to the live library for such a row. The harness gives the agent nothing for it, so dressing the row
  up with library text the agent never sees was the dishonest half of the divergence; both readers
  now render it from its slug. This is display-only — no stored state changes.
- **A malformed company bundle now surfaces as an error from `GET …/skills`** instead of a list with
  that bundle silently skipped. `load_dir_skills` refuses the whole directory, so the harness loses
  the catalogue for every agent in that company; a tidy partial list described a set nobody had.
- Added `CompanyRuntime::globals_disable()`, mirroring `enabled_workflow_ids()`, so both readers get
  the manifest opt-outs from one place. No behaviour of its own.

Not changed: what the harness materializes. Every existing materialization test passes untouched.

## Tests

New, and each pinned to a claim in the issue:

- `company::skill_effective::test` (11) — the baseline is the bottom layer; a company bundle
  supersedes a same-slug global; a disabling delta reports disabled rather than hiding; a manifest
  opt-out beats a console re-enable; `custom_doc` supersedes the layer beneath; a malformed
  `custom_doc`/unsafe slug costs that skill, not the resolution; a malformed bundle fails it.
- `server::ops::skills::tests::the_list_includes_the_global_baseline_with_no_bundles_and_no_deltas`
  — pinned against `globals::skills()` itself, so the reader cannot drift from the baseline.
- `…::a_disabled_global_is_listed_as_disabled_rather_than_hidden`,
  `…::a_manifest_opt_out_is_honoured_by_the_reader`.
- `…::the_rest_list_and_the_harness_effective_set_agree` (gated on `openhuman`) — **the test that
  would have caught this**: for one company it materializes the agent's skill tree and asserts the
  directory names on disk equal the enabled rows the REST list reports, with the disabled global
  absent from both and every other global present in both.
- `…::the_rest_list_and_the_graphql_resolver_agree` — field-for-field over the same company,
  including the `custom_doc` refresh that used to diverge.
- `…::http::a_global_can_be_disabled_and_re_enabled_through_the_put_route` — end to end over the real
  route: a global starts enabled, the switch sticks off, the row stays `company`-sourced, and it
  re-enables.
- `…::a_malformed_company_bundle_surfaces_as_an_error`.

Updated: two REST write-route tests asserted "no delta landed" as an empty list, which the baseline
makes wrong; they now compare against a snapshot taken before the refused write. Three GraphQL tests
counted skill rows; they use a new `own_skills` helper (the twin of the existing `own_workflows`) or
assert on rows directly where the fixture installs a baseline slug.

Commands run locally:

- [x] `cargo clippy -p opencompany-core --all-targets --no-deps -- -D warnings`
- [x] `cargo clippy -p opencompany-core --all-targets --features openhuman --no-deps -- -D warnings`
- [ ] `cargo build --all-targets` — not run as such; both `clippy --all-targets` invocations above
      compile every target in the default and `openhuman` feature sets.
- [ ] `cargo test` — not run whole; a bare `--lib` run exhausts memory on this machine. Scoped runs,
      all green: `company::` (1503), `server::` (1867), `harness::` (1844) under `--features
      openhuman`, plus `--test auth_matrix`. Those cover `company::skill_effective`,
      `company::skill_file`, `harness::built_in::skills`, `server::ops::skills`,
      `server::ops::write_test` and `server::graphql`.
- [x] `scripts/ci/assert-feature-lanes.sh` — the one new `openhuman`-gated test sits in an
      already-`tested` lane.

## Documentation

`docs/spec/runtime/globals.md` — the "What each surface does" table named
`EffectiveSkills::materialize` as the skills merge point and the `[globals].disable` paragraph named
`harness::globals_skill_disables`. Both now name `company::skill_effective`, and the table row says
that the harness materializes the enabled entries while the two readers report the whole set so a
global's row carries its own switch.

No frontend change was needed: `Skill.source` is already a free-form string in `@/api/skills`,
`installedIds` is keyed by id, and `InstalledCard` renders the switch for every row while hiding
uninstall for a `company`-sourced one.

## Summary by CodeRabbit

* **New Features**
  * Added a unified effective skills view across the console, REST API, GraphQL, and runtime.
  * Global, company, and custom skills are now combined with consistent precedence and deterministic ordering.
  * Disabled skills remain visible with their current status.
  * Manifest-configured skill opt-outs are reflected across all skill surfaces.
* **Bug Fixes**
  * Improved handling of invalid slugs, malformed documents, registry snapshots, and company bundles.
  * Ensured company-installed skills correctly override matching global skills.


oxoxDev:
@coderabbitai review

Issue #2322, opened 2026-09-15
desktop: analytics wiring is absent — decide whether the desktop reports telemetry at all


oxoxDev:
## What

`serve` calls `with_analytics` plus `install_analytics`; the desktop entry point does neither, and the per-company `with_analytics` (C12) is absent to match.

## Why it is filed rather than fixed

`analytics::config::resolve` returns `Silent(NotHosted)` off a hosted tenant, so this only differs for someone who sets `OPENCOMPANY_ANALYTICS=on` explicitly. Whether a desktop application reports telemetry is a product and privacy decision, not a wiring gap to close by default.

Filed so the absence is a recorded decision rather than an oversight — which is what it is today.

## Found by

Desktop/serve wiring parity audit, 2026-09-15 (items A14, C12). Recorded in the parity PR's `DELIBERATE_DIFFERENCES` as pending this issue.

Issue #2321, opened 2026-09-15
desktop: boards get no baseline setup cards, because task seeding is never enabled


oxoxDev:
## What

`serve` calls `with_task_seeding(true)` on its company builder. The desktop builder does not, so a desktop company's boards start with no baseline setup cards.

`with_task_seeding`'s own doc says to call it from a product entry point. The desktop is one.

## Why it is filed rather than fixed

It is visible: every desktop board gains cards it does not have today. That is a product change worth deciding on its own merits rather than restoring silently inside a wiring PR.

## Found by

Desktop/serve wiring parity audit, 2026-09-15 (item C14). Recorded in the parity PR's `DELIBERATE_DIFFERENCES` as pending this issue.

Issue #2320, opened 2026-09-15
desktop: sqlite storage and the memory overlay are documented but unreachable


oxoxDev:
## What

Two host-wide wirings `serve` performs and the desktop entry point does not:

- `with_stores` + `with_storage_kind` — the desktop never calls `open_storage`, so it can only ever run the `fs` backend. The app manifest states that **sqlite is the desktop's backend**. The doc and the code disagree.
- `with_memory_overlay` — `OPENCOMPANY_MEMORY` and the `[memory]` config section are silently ignored on desktop.

Per-company halves pair with these: `with_storage_kind` (C9) and `with_memory_overlay`/`_cleared` (C18), the latter mattering on rebuild.

## Why it is filed rather than fixed

Wiring these unlocks two code paths that have **never run** on a desktop host. That is a behaviour change with a real risk surface, not a restoration, so it wants its own PR and its own testing rather than riding along with the parity work.

`with_storage_kind`'s current absence is harmless today: the default `fs` is the refusing direction for repository-credential gates.

## Found by

Desktop/serve wiring parity audit, 2026-09-15 (items A7, A8, C9, C18). Recorded in the parity PR's `DELIBERATE_DIFFERENCES` as pending this issue.

Issue #2319, opened 2026-09-15
desktop: setup writes are authorized anonymously forever, because setup_complete is never wired


oxoxDev:
## What

`bin/opencompany.rs` calls `with_setup_complete(setup_completed_at.is_some())`. The desktop entry point (`crates/opencompany-app/src/embedded.rs`) never does.

`setup::authorize` short-circuits on `!setup_complete()`, so on a desktop host the host-level setup surface stays open without a session — not just on first run, but on every relaunch, permanently.

## Why it matters

`server/setup.rs`'s own module doc explains the intent: the call is open without a session in exactly two situations, both meaning "there is nobody who could authorize it" — setup has never completed, or the host has no companies. It then says openness on a *configured* laptop would let any page in the browser rewrite its settings. That is the state every desktop install is in today.

Loopback-gating limits the blast radius to the local machine; it does not close it.

## Fix

Wire `with_setup_complete` on the desktop entry point. Note this **tightens** auth — the behaviour change is that host setup writes stop being anonymous once setup has completed.

## Found by

Desktop/serve wiring parity audit, 2026-09-15. Deferred from the parity PR deliberately because it changes behaviour; recorded in that PR's `DELIBERATE_DIFFERENCES` as pending this issue.

PR #2324, opened 2026-09-15
fix(desktop): wire the embedded host's missing config and hub identity, and fail a test when it diverges from serve

oxoxDev:
## Summary

The desktop builds its `AppState`, `AppConfig` and per-company `RuntimeBuilder` from a hand-written list of calls, and so does `serve`. Nothing compared the two, so every divergence was found by a person hitting it in the console.

This audits both entry points, restores the gaps that are pure wiring, and adds a test that fails when they diverge again. **The test is the point** — fixing today's three gaps without it would let the fourth ship exactly as these did.

One correction to the premise, because it changes where the fix belongs: `serve` does **not** go through `AppConfig::load` either. It also builds field-by-field, but populates every field from the environment and `config.toml` first. The divergence was never "loaded vs hand-built" — it was one hand-written list fed by the config layer and one fed by nothing. Hence a shared resolver, `AppConfig::resolve_host`. `serve` keeps its own list this PR (its build sits inside a `main` match arm with no seam to call); the test binds the two until a follow-up migrates it.

### One finding that is not a wiring row

**A relaunched desktop host authorizes host-level setup writes anonymously, forever.** `setup::authorize` returns early when `is_local_only() && (!setup_complete() || registry().is_empty())`. `serve` seeds `setup_complete` from `config.toml`; the desktop never wires it, so the stamp is `false` on every launch and the first clause is permanently true. Bounded today — loopback plus `AuthMode::None`, where the person at the machine is the principal anyway — but **unbounded the moment an operator picks `email` in the wizard**, which the wizard offers. Not fixed here: wiring it *tightens* authorization, which deserves its own change rather than a side effect of a parity sweep. Tracked in #2319.

### One that is a design question

`builder.with_host_base_url` cannot be fixed by copying the call. The desktop's `config.bind` is the literal `127.0.0.1:0` — the port is chosen by the OS and known only after `server::bind`, which is after every company is registered — so `host_base_url()` answers `http://127.0.0.1:0`. The credential-link flow escapes it only because `callback_origin` prefers the browser's own `Origin`; verified live, a request with one gets a correct callback and a request without one gets `:0`. This needs a seam carrying the bound address, not a copied line.

## API Or Behavior Changes

**Restored — the desktop now does what `serve` always did:**

- Resolves `api_url`, `web_url`, `tinyplace_api_url`, the TinyHumans credential and install-wide MCP defaults from the environment and `config.toml`, instead of taking compiled-in production defaults. A desktop pointed at staging now reports and talks to staging.
- `hub_identity` is wired, so the Account page can read a balance and `credential/link/start` builds a URL instead of refusing with "this host is not part of a TinyHumans ecosystem". It does **not** restore the key-grant UI removed on 2026-09-14; no UI is added back.
- Every desktop company builder now receives the tiny.place URL, MCP defaults, workspace quota and git flag, skill registry, and hub feedback client. The quota and git flag were already resolved onto `AppConfig` and then dropped — those two keys were dead, not merely absent.
- `opencompany-app` enables `opencompany-core/tinyhumans`. It is `["dep:reqwest"]` and `oauth` already pulls reqwest, so no package is added and `Cargo.lock` is byte-identical.

**Not changed:** no new secret-store key, env var or route, and nothing effectively off on the desktop is switched on. The skills registry resolves empty unless a `skills_root` exists, which a packaged install has no checkout to supply — no path is fabricated.

**The full audit** — all 22 rows across `AppState`, `AppConfig` and `RuntimeBuilder`, each classified as fixed, deferred with its tracking issue, or correct-as-is with the reason — lives in `DELIBERATE_DIFFERENCES` in `tests/desktop_serve_parity.rs`, where the test keeps it honest. Duplicating it here would create the second hand-written list this PR exists to remove.

## Tests

- [x] `cargo clippy --all-targets --no-deps -- -D warnings` — clean on both crates
- [x] `cargo test` — scoped: `--test desktop_serve_parity` (2), `--lib app::` (145), `--lib desktop::` (14), `opencompany-app --lib embedded::` (10)

**`tests/desktop_serve_parity.rs`** reads the four source files, extracts the wiring each side performs, and asserts the difference is exactly `DELIBERATE_DIFFERENCES`. It follows `tests/auth_matrix.rs`, which scans `src/server` for route literals on the same reasoning: the comparison one would rather make is impossible, because `serve`'s wiring lives inside a `main` arm that locks a data root and binds a listener, and the desktop's lives in a crate outside this workspace.

The universe of wiring is **derived** from `pub fn with_*` declarations rather than listed, so a builder method added tomorrow enters the comparison with no edit here. It fails in both directions — a declared row that no longer diverges is reported stale, because a stale exception is how the next real gap hides.

Verified it can fail, rather than assumed:

- deleting `.with_workspace_quota(` → fails with `["builder.with_workspace_quota"]`
- adding a new `with_probe_marker` called only from `serve` → fails with `["state.with_probe_marker"]`, the fifth-gap case
- reverting the resolver makes `/spec` report `"api_url":"https://api.tinyhumans.ai"` while `config.toml` names staging — the original bug reproduced

New: five tests on `AppConfig::resolve_host` (env layer, file layer, precedence including a whitespace-only env value that must not beat a real file value, `hub_site()` following the resolved hub, `[workspace]`), and two on the embedded host.

### Verified on a running host

Fresh `OPENCOMPANY_DATA_DIR`, `TINYHUMANS_API_URL` at staging, own console bundle on an OS-chosen port:

- `/spec` → `api_url: https://staging-api.tinyhumans.ai`, `capabilities` includes `hub-identity` — the reported symptom, cleared
- `credential/link/start` → `200` with an authorize URL on **`staging.tinyhumans.ai`**, previously a refusal. That host is `hub_site()` derived from the resolved `api_url`, so the chain works end to end
- `credential/billing` → `{"configured":false}`, the honest no-key branch

**Not verified:** the rendered Account page. The first-run activation funnel would not clear — the company-name field reverts to its placeholder on each confirm — so the page behind it was unreachable. That reproduces independently of this change. The API responses it renders are verified; the page is not.

## Documentation

No user-facing docs change: nothing here adds a surface, flag or key. The reasoning lives where it is load-bearing — `AppConfig::resolve_host` documents which layers it owns, and `DELIBERATE_DIFFERENCES` is a live index of open work rather than a list of unexplained exceptions.

Two stale comments corrected: `embedded.rs` claimed twice that this host "builds its config by hand rather than through `AppConfig::load`" as the reason the `auth_mode` layer was applied inline. That reason no longer holds.

## Related

Follow-ups filed from the audit: #2319 (setup writes authorized anonymously), #2320 (sqlite storage and memory overlay unreachable), #2321 (no baseline setup cards on desktop boards), #2322 (analytics wiring absent). A further follow-up should migrate `serve`'s own `AppConfig` build onto `AppConfig::resolve_host` and delete `resolve_serve_base_url`.

## Summary by CodeRabbit

- **New Features**
  - Host configuration now consistently applies environment variables and root `config.toml` settings, including API endpoints, workspace options, authentication, and MCP servers.
  - Desktop sessions now inherit configured workspace limits, Git settings, shared skills, MCP servers, and service URLs.
  - TinyHumans-enabled deployments can support hub identity and credential-linked feedback.

- **Bug Fixes**
  - Configuration precedence and blank-value handling are now more predictable.
  - Embedded and desktop hosting behavior is better aligned with standard serving behavior.

- **Tests**
  - Added coverage to verify configuration resolution and host feature parity.


oxoxDev:
@coderabbitai review

The previous review was against 772fc1737. Head is now b54c5a1a7, which adds one formatting-only commit: `cargo fmt` for the `opencompany-app` crate, which is a separate workspace that a root-level `cargo fmt --all` does not reach. No behaviour changed.

PR #2325, opened 2026-09-15
feat(setup): skip the model step when the host already supplies one

oxoxDev:
## Summary

First-run setup opened by asking for a model and an API key even on a host that
already reaches one. On a hosted tenant the control plane injects the
credential, `GET /api/v1/setup` reports `inference.ready: true`, and the
operator has no key of their own and no way to obtain one — so the first screen
of the product demanded a secret from the one person who cannot supply it, and
gated Next on probing it.

When the host reports itself ready, the wizard no longer asks. The model step is
absent rather than pre-filled, the progress bar renumbers with it, and the review
screen says once that the model came with the host, so an operator who was never
asked can still find out where theirs came from.

Frontend only — no host change. The signal this reads was already on the wire.

This sits on the same screen as #2317 and should land after it.

## API Or Behavior Changes

- A host reporting `inference.ready` shows **one fewer wizard step**: the model
  step is not rendered and gets no slot in the progress bar, so the flow reads
  "step 1 of 4" where it read "step 1 of 5".
- That path **stores no inference provider**. The apply payload carries no
  `company.inference` block and no `tinyhumans_api_key` field; the host resolves
  the managed endpoint from its own environment at runtime, and writing a
  provider the operator never chose would record a decision nobody made. This
  preserves the property the submit already had — inference is written only when
  a tested credential was the operator's own.
- `inference.ready: false` is unchanged in every respect: same step, same
  ordering, same gate, same providers, same escape.
- One new line on the review screen, shown only on a ready host.
- No new route, no new field on the payload, no new config key.

Note for a follow-up: the `PowerStep` affordances that existed for a ready host —
"using this host's key" and the "use my own" override — are now unreachable from
the wizard, since they were only ever shown when `inference.ready` was true.
They are left in place here rather than widening this diff.

## Tests

`frontend/test/unit/setup-wizard-hosted-model.test.ts` is new. Four claims about
a ready host, all of which fail on unmodified `main`:

- the model step is absent and the bar counts four steps, not five
- the walk reaches Finish with no key field and no connection test rendered
- the apply payload carries no inference block and no key field
- the review screen names the host-provided model exactly once

Three more pin the manual path and pass both before and after — deliberately, as
the regression guard: a host with no model of its own still asks first in a
five-step flow, still holds the step on an untested connection, and still says
nothing about a host-provided model.

`setup-wizard-finish-gate.test.ts` loses four tests that mounted the wizard on a
ready host to assert against the model step's host-key affordances. That host
shows no model step now, so there is no screen left for them to mount. The
invariant they guarded — a house-credential pass is never stored as the
company's own provider — is carried by the new file's payload assertion.

Commands run locally (from `frontend/`):

- [x] `npm run typecheck`
- [x] `npm run typecheck:e2e`
- [x] `npx vitest run` — 5876 passed, 3 skipped
- [x] `npm run build`
- [ ] `cargo fmt --all -- --check` — N/A, no Rust changed
- [ ] `cargo clippy --all-targets -- -D warnings` — N/A, no Rust changed
- [ ] `cargo build --all-targets` — N/A, no Rust changed
- [ ] `cargo test` — N/A, no Rust changed

Walked in a browser against a locally built host serving this bundle, both ways:
with the host reporting no model of its own, and with a platform credential in
its environment.

## Documentation

None needed. The behaviour is described where it is decided — on `visibleSteps`
and on the `tested` verdict — and no public API, route or config key changed.


oxoxDev:
@coderabbitai review

PR #2329, opened 2026-09-15
fix(console): rewrite the room's grammar out of the live feed, without blinding the episode fold

sanil-23:
## Summary

A deliberating desk showed `!support #topic ^16 …` in the operator's chat while
the feed was open, and plain prose after a refresh. The reload path has
rewritten the room's grammar since `readable_moves` existed; the live frame
never did.

**The first attempt at this PR was wrong, and the history keeps it.** Cleaning
`text` on the live frame breaks the deliberation panel: `foldEpisodes` gates on
`moveOf(m.text)`, so with the grammar gone an episode is not detected at all.
Codex caught it (P1, thank you) and the commit is reverted rather than hidden.
The rest of this PR is the fix done in the order it needed.

| # | commit | |
| --- | --- | --- |
| | `Revert "apply the desk's own display rules…"` | back to main's behaviour |
| 1 | carry the agent's own body on the live frame | additive, inert |
| 2 | fold episodes from that body | frontend |
| 3 | rewrite the grammar out of the live reply | the original fix, now safe |

**1 — `cueText` on the live frame.** `MessageView` has shipped two bodies since
it gained `cue_text`: the operator-facing one and the one the model wrote. The
live frame carried only the first, so anything needing the grammar had to scrape
it back out of prose. Inert on its own; byte-equal to `text` on every row
carrying no move.

**2 — the fold reads it.** `moveOf(m.cueText ?? m.text)`, falling back for a host
predating the field. This fixes a **second, older bug**: on reload the host
always rewrites `text`, so the fold saw no moves and rendered no episode. The
deliberation panel has never survived a page refresh on an ordinary episode —
only a cross-desk crossing rescued it, because `HIVE_REFERRAL_AUTHOR` is matched
on the channel rather than the body.

**3 — `text` becomes what its name says.** Cleaning stays at the display edge and
nowhere earlier: the journal keeps the authored line, because a room whose own
transcript had been cleaned could not count itself, and every agent-facing path
(`EpisodePrompt`, `elsewhere_for`, `referral_prompt`, `chat_seed`) still reads
the stored body — which is how a seat cites `^16` against a row it can identify.

## What is deliberately not here

**The closing row still renders as a teammate.** `history_for_desk` drops
`HIVE_REPORT_AUTHOR` and the first attempt did the same live — but `ThreadPanel`
renders no `EpisodeBlock`, and a hive turn parented to its trigger renders
there, so for a threaded episode that row is the only place the verdict appears.
Suppressing it would repeat this PR's original mistake one surface over. It
needs `ThreadPanel` to render episodes, or its own non-teammate treatment —
a UI decision, not a backend flag.

## API Or Behavior Changes

- The live `agent_reply` frame gains `cueText` and rewrites `text`, matching what
  `chat/history` has always returned. A reply with no move is unchanged, which is
  every reply on every desk that does not deliberate.
- No change to the journal, to any fold, or to any agent-facing prompt.

## Tests

- [x] `cargo clippy -p opencompany-core --all-targets -- -D warnings`
- [x] `cargo test` — 5694 passed, 0 failed
- [x] `tsc --noEmit`, `vitest run` — 5873 passed, 0 failed

Backend, paired so the cleaning cannot over-apply:
`projects_the_agents_own_body_beside_the_operators` (grammar survives on
`cueText`, prose on `text`) and `a_reply_with_no_move_carries_the_same_body_twice`.

Frontend: `still finds the room when only cueText carries the grammar`, written
against the **tidied** shape on purpose — grammar in `cueText`, prose in `text`.
Every existing fold test puts grammar in `text`, which passes with or without
this change; that is how the reload gap survived. Confirmed red without the fix
and green with it, rather than assumed.

## Documentation

None needed: `MessageView::cue_text` already documents why a surface that needs
what the model saw must not read what the operator is shown. The fold is such a
surface and is now wired that way; the new call sites cite it.

## Summary by CodeRabbit

* **New Features**
  * Agent replies now preserve both the model’s original wording and a readable prose version.
  * Live and reloaded conversations retain original move syntax for episode tracking.

* **Bug Fixes**
  * Episode folding now recognizes proposals and supports when displayed text has been rewritten into prose.
  * Replies without move syntax remain consistent across live and reloaded views.


sanil-23:
Confirmed, and thank you — both of these are right, and I verified them against the frontend rather than taking them on trust.

`foldEpisodes` gates the entire deliberation UI on:

```ts
const hasRoom = rows.some((m) =>
  m.channel === HIVE_REPORT_AUTHOR ||
  m.channel === HIVE_REFERRAL_AUTHOR ||
  (m.from === "company" && moveOf(m.text) !== null));
if (!hasRoom) return [];
```

This PR removes two of those three disjuncts, so an episode with no referral is not detected at all, and `close(message, parseEndingReport(...))` never fires. `foldEpisodes` feeds `buildTimelineItems` (`RoomView.tsx:1459`, `:1484`), so the cost is the whole episode block, not a detail of it.

The part I had backwards: I treated the live/reload divergence as a defect on the live side. It is the opposite — the live feed carries raw grammar and the closing row **because the fold needs them**, and history can afford to strip both precisely because it is not what the fold runs on. I fixed the wrong side.

The underlying tension is real, though: the fold wants the move, and the operator should not read `!support #topic ^16`. Your suggestion is the right shape — carry the readable rendering *and* typed move metadata, and project the closing row as episode metadata rather than as a teammate reply. That is a bigger change than this PR, spanning both sides of the wire, so I am closing this rather than patching it into something half-done.

Reverting to the status quo restores the episode UI. The operator-facing grammar leak stays open, and I will write it up with this finding so the next attempt starts from "the fold consumes these frames" instead of rediscovering it.

sanil-23:
Both addressed — one fixed, one I believe is a false positive. Details so it is easy to check me.

**Codex, "Forward cueText through the stream dispatcher" (P1) — correct, fixed in `9fbc5d4da`.**

`handleEvent`'s `agent_reply` arm rebuilds the payload field by field rather than spreading the event, and `CompanyStreamEvent` did not declare the field, so the host sent it and the dispatcher dropped it. Live replies would have reached `AppShell` with `cueText` unset, the fold would have fallen back to the rewritten `text`, and a live room would have rendered as ordinary chat — this PR's original failure one layer down. Both the stream variant and the dispatcher now carry it.

The deeper point in that finding is the one worth keeping: `cueText` is optional at every hop, so a hop that forgets it compiles *and* type-checks and simply arrives `undefined`. That is now three boundaries in this change where an optional field let a gap through silently — the history DTO, the stream variant, and the dispatcher. Types cannot guard a chain of optionals, so the guard is a test at the boundary (`agent-reply-cue-text-transport.test.ts`), verified red when the copy is removed again.

**CodeRabbit, "Include `cue_text` in `ChatHistoryMessageDto`" — correct, fixed in `a6d962eb6`** (you marked it addressed). Worth recording that it was not merely a missing mapping: `/chat/history` never carried the field at all, so the previous commit's fold was reading something the endpoint did not send. Confirmed against a running host before fixing.

**CodeRabbit, "Guard demoted rows before parsing `cueText`" (Major) — I do not think this can happen, and I would rather be shown wrong than quietly skip it.**

The premise is a row whose authored body still carries `!support` while the projected body has it stripped. Demotion happens *before journaling*: `moves::demote` is called at `hivemind/episode.rs:839` and its return value is the reply text that gets journaled. So the journal holds the already-demoted line, and `cue_text` is that same body.

`readable_moves` only rewrites a line where `line_kind` is `Some`, which requires a leading `!`. A demoted line has none, so it passes through untouched and `text == cue_text`. The DTO then omits `cueText` (it is only sent when the two differ), `bodyOf` falls back to `text`, and `looksDemoted` runs exactly as before.

So the only rows where the two bodies differ are legitimate moves, which should count. If there is a path where a `!` survives into the journal after demotion I have missed it — that would make this real, and I would take the guard.

PR #2364, opened 2026-09-17
feat: make Hive Mind conversations session-native

senamakel:
## Summary
- route channel and DM messages to one TinyHiveMind-selected responder and preserve per-agent conversation sessions
- make speech default-on with explicit opt-out, and deliver `desk_post` / `desk_dm` replies back into the unified transcript without creating a task for every message
- support concurrent cross-agent DMs while serializing turns for the same agent
- isolate dispatched-card attempts from unrelated chat/task history and keep prior failed output out of retry instructions
- ground agents in their actually wired public-web capabilities, route research directly to `web_search`, and fail fast on credential/session errors
- move tool-call detail into Raw Turns, poll that view quietly, and keep it scrolled to the latest turn
- add focused unit and mock E2E coverage for routing, speech, session isolation, concurrent DMs, research retry behavior, and Raw Turns
- update the OpenHuman gitlink for paged `file_read` support from tinyhumansai/openhuman#6321

## Behavior changes
- ordinary chat and DMs no longer create task cards by default
- mentions and direct greetings route through TinyHiveMind to the addressed teammate
- speech is enabled unless `[speech] disabled = true`; legacy `enabled = false` does not silently preserve the old default-off behavior
- dispatched task attempts start from a clean agent context while operator/reviewer instructions remain authoritative
- long persisted tool results are read through bounded continuation pages instead of repeatedly truncating at byte zero
- completed tool-call detail appears in Raw Turns rather than the normal chat transcript

## Validation
- `git diff --check`
- `cargo test -p opencompany-core --features openhuman,hivemind public_research --lib`
- `cargo test -p opencompany-core --features openhuman,hivemind dispatched_cards_are_isolated_from_an_agents_other_conversations --lib`
- isolated/serial `cargo test -p opencompany-core --features openhuman,hivemind --test hivemind_e2e -- --test-threads=1` (10 passed)
- `pnpm typecheck`
- targeted frontend Raw Turns/live-chat tests (28 passed)
- live desktop run: one agent coordinated two DM recipients concurrently without creating cards; a research retry invoked the real `web_search` tool and stopped with the exact `401 SESSION_EXPIRED` blocker rather than looping

## Dependency
- Depends on tinyhumansai/openhuman#6321. The gitlink intentionally points at that PR's focused feature commit on OpenCompany's recorded compatible OpenHuman base; the OpenHuman PR itself is rebased through current upstream history.

## Test scope
The full OpenCompany suite was intentionally deferred at the operator's request; focused local tests and the complete Hive Mind integration target are green.

## Summary by CodeRabbit

* **New Features**
  * Agents now receive team, desk, role, and hand-off information in their prompts.
  * Team members can hand off work or send bounded direct messages without automatically creating task cards.
  * Speech tools are enabled by default and can wake eligible recipients.
  * Research tasks receive clearer web-search and source-handling guidance.
  * Raw turn views refresh quietly and automatically scroll to the latest entry.

* **Changes**
  * Task cards are created through explicit workflow requests, agent task creation, hand-offs, or human actions.
  * Unrestricted hand-offs are now the default unless limited by configuration.
  * Live activity is shown with a compact working indicator instead of detailed step timelines.


sanil-23:
Reading this closely while working on the referral/a2a side (#2341, #2347) — it lands next to an architecture question, so I have written it up as #2368 rather than as review noise here. Short version, and where I think this PR helps.

**Naming, so nobody talks past each other:** "session-native" here means *agent sessions* — the per-agent continuous session `agent_session.rs` introduced. It is not tinyhivemind's `SessionLog`/`project_session` projection layer, which `chat_history.rs` still never calls. Two different things behind one word; worth saying once in the description.

**What I think is straightforwardly right:**

- **Cards decoupled from conversation.** This is aimed at a real defect I verified on a live journal: a card opened at `seq 206` with `column: in_progress` **before the agent spoke at `seq 208`** — no triage match, and no `delegate_*` call. It comes from `runtime/delegation.rs` opening a card as part of dispatching a turn. That is the right target, and a bigger change than the one-line bullet suggests.
- **`accept_seen`.** Necessary regardless of how the rest settles: a hive turn shows rows outside the session, so the watermark has to record the trigger without jumping past an unread DM below it. Nine lines that anyone would have had to write.

**Where I would push back — the `desk_dm` dispatch edge:**

1. **It is a harness turn.** `ChatTarget::in_thread` means `history_seed: true`, so the recipient gets the session-seeded shape. The referral path deliberately refuses that (`ChatTarget::deliberating`) for reasons its own comment gives: an uncontrolled second copy of a conversation arrives unattributed and in the assistant role. That is why a2a through referral is bounded by `peer_cap`, folded onto the asking row, and carried home; a `desk_dm` wake is none of those.

2. **It defers the thing the tool exists to do.** The library's own description is *"use it to settle a disagreement without spending the room's attention"* — a resolution verb. Staging onto the drain means the sender's turn ends before anything is settled, so the agent that asked to settle something proceeds without the settlement.

3. **Inside a hive turn it is actively harmful.** `turn.reply = String::new()` on a DM-only turn means the seat contributes **no move** — it burns a turn against `turn_budget` and likely trips the malformed-move correction retry. Speech-on-by-default makes it available in exactly that place.

4. **The handoff edge moved.** The library puts it on `post` (*"Mention a teammate with @id to hand them the next turn — only the first mention does that"*); this puts it on `dm` and leaves `post`'s mention inert. Deliberate or not, it is a divergence from the vendored spec.

**Two smaller flags:**

- `legacy enabled = false does not silently preserve the old default-off behavior` — an explicit operator opt-out becomes inert. Acknowledged in the description, but it deserves a migration note rather than a bullet, since companies that wrote `enabled = false` meant it.
- Speech-on-by-default is a large behavioural change on its own: most companies (`retail_co` included) have no `[speech]` block at all, so this hands every agent four new tools and changes how a turn's text is emitted — return text becomes private thinking and is discarded when the turn posts. Combined with `the full OpenCompany suite was intentionally deferred`, that is a lot of surface moving at once.

None of this blocks the parts I called right. If it helps, the split I would suggest is: land the card decoupling and `accept_seen`, and hold the dispatch edge until it can route through the referral mechanism — same trigger, same tool, but opening a bounded exchange that returns what it settled instead of queueing a harness turn.

Happy to be wrong on any of it — I have been on this thread already (I first read the `desk_dm` change as moving away from a stigmergic design, when the library's `post` spec hands turns too, so the asymmetry is defensible).

PR #2366, opened 2026-09-17
fix(console): restore the referral frame fields and 15 tests dropped by #2349

oxoxDev:
## Summary

Restores work that is on `main`'s history but not in `main`'s tree. **One item is a live production regression in the operator console**; the other fifteen are deleted tests. Two hygiene fixes ride along.

### The production regression (live now)

`project_event_for_viewer` stopped emitting `target`, `asker` and `direct` on the `referral` frame. `o["toDesk"]` and `o["sequence"]` survived; only those three went.

The console still reads all three. `app-shell.tsx` picks the shape of a running crossing with:

```ts
event.direct && event.asker && event.target
  ? { direct: true, asker, target, row: event.sequence }
  : !event.direct && event.toDesk
    ? { direct: false, desk: event.toDesk, row: event.sequence }
    : undefined;
```

With the fields never emitted, `event.direct` is `undefined`, so the first branch is unreachable and **every** crossing takes the desk branch — including a person-to-person one. `referralWorkingLabel` then renders "the <desk> desk is answering" instead of "<A> and <B> are talking". On a same-desk pair crossing `to_desk` is the *asking* desk, so the header names the room that asked as the room that is answering. `use-events.ts` still declares `direct: boolean` on the frame, so the type says the field is always there while the host never sends it.

### How it happened

PR #2349 (`test: move inline test modules to sibling *_tests.rs files`) carried two merges from `upstream/main` into its `docs-pass` branch that resolved in favour of the branch's pre-split copy of a file, discarding what `main` had gained in the hours before:

| Merge | Discarded |
|---|---|
| `33caef4af` | #2347 |
| `3b56296fd` | #2353, #2354, #2355 |

The last good tree is `488c02278`; everything here is recovered by diffing against it. The split itself is fine, and #2349's size was not the problem — the overwhelming majority of its `+5,131` is legitimate per-file scaffolding (imports, module headers, `#[path]` declarations) that a per-file split necessarily produces. Reviewing it by line count would not have surfaced this; only the merge resolutions would have.

## API Or Behavior Changes

- **Behavior (fix):** the `referral` SSE frame carries `target`, `asker` and `direct` again. The console can once more tell a person-to-person crossing from a desk crossing, and labels a running pair exchange with both names. No wire field is added or renamed — these are fields the console's types already declare.
- `server::ops::company_key::prober_override` widens from private/`pub(super)` to `pub(crate)`. It was narrowed during the split because the only out-of-module caller — a restored setup test — was no longer there to need it.
- No other public API change.

## Tests

Fifteen deleted tests recovered from `488c02278` and placed in whatever file owns that area post-split (no pre-split file is recreated):

| Test | From | Now in |
|---|---|---|
| `a_direct_crossing_names_both_sides_of_the_exchange` | #2347 | `server/operator_test_group_18.rs` |
| `a_pair_exchange_that_breaks_keeps_the_rows_it_already_has` | #2347 | `hivemind/referral_pair_tests.rs` (new) |
| `a_question_to_a_teammate_does_not_spend_the_peer_desk_allowance` | #2347 | `hivemind/referral_pair_tests.rs` |
| `the_asker_is_handed_the_exchange_it_just_had_and_nothing_older` | #2347 | `hivemind/referral_pair_tests.rs` |
| `a_seat_reads_the_private_lines_it_is_part_of` | #2347 | `hivemind/referral_pair_tests.rs` |
| `the_wizards_account_key_fans_out_onto_the_company_it_seeds` | #2353 | `server/setup/setup_test_group_4.rs` |
| `an_account_key_with_no_company_to_own_it_is_not_written_anywhere` | #2353 | `server/setup/setup_test_group_4.rs` |
| `the_preferred_model_is_probed_before_whatever_the_catalogue_lists_first` | #2354 | `server/setup/setup_test_group_4.rs` |
| `a_catalogue_without_the_preferred_model_keeps_its_own_order` | #2354 | `server/setup/setup_test_group_4.rs` |
| `an_embedding_model_never_outranks_the_preferred_one` | #2354 | `server/setup/setup_test_group_4.rs` |
| `a_refused_key_travels_as_a_reason_and_a_code_never_as_the_hubs_body` | #2355 | `server/ops/company_key_a_companys_own_key_tests.rs` |
| `an_unreachable_hub_is_not_a_refused_key` | #2355 | `server/ops/company_key_a_companys_own_key_tests.rs` |
| `a_host_with_no_hub_says_so_rather_than_blaming_the_key` | #2355 | `server/ops/company_key_a_companys_own_key_tests.rs` |
| `a_company_with_no_key_offers_no_reason_to_judge_one_by` | #2355 | `server/ops/company_key_a_companys_own_key_tests.rs` |
| `only_a_401_is_a_refused_key` | #2355 | `server/ops/company_key_a_companys_own_key_tests.rs` |

Plus the two `FarDesk` break helpers (`failing_after`, `silent_after`) the pair tests need, and the `asker` / `direct == false` assertions that #2347 had added to the existing `a_crossing_names_the_thread_whose_fold_changed`.

The three model-preference tests are `#[cfg(feature = "openhuman")]`; `openhuman` is already a `tested` lane in `scripts/ci/feature-lanes.txt`, so no lane row changes.

**Falsification of the regression guard.** With the three statements reverted and everything else unchanged:

```
a_direct_crossing_names_both_sides_of_the_exchange ... FAILED
  assertion `left == right` failed: a person was asked, not a desk
    left: Null      right: true
a_crossing_names_the_thread_whose_fold_changed ... FAILED
    left: Null      right: "software_engineer"
```

`Null` is the field being absent, which is exactly the shipped fault. Both pass with the fix in.

The other fourteen pass as restored — the production code they cover survived the clobber — so no second regression was found behind them.

### Also in this PR

- `crates/opencompany-app/src/embedded.rs`: deletes a 362-line `/* mod test { … } */` block that duplicated `embedded_tests.rs` function for function. The live module is declared immediately above it.
- `the_folded_aside_reaches_the_wire_as_camel_case` existed in both `operator_aside_dto_tests.rs` and `operator_test_support_1.rs`, and `operator.rs` declares both modules, so it compiled and ran twice. The copy in the support file is removed; the file named for the test keeps it.

### Not in this PR

#2349 also moved some tests from the default lane into the `openhuman`-gated lane. Whether they belong there is a coverage judgement rather than a restore, so it is left for a maintainer to decide separately.

### Commands run

- [x] `cargo clippy --no-deps --all-targets -- -D warnings` (default and `--features openhuman`)
- [x] `cargo test -p opencompany-core` (default features)
- [x] `cargo test -p opencompany-core --features openhuman` for the gated model-preference tests

## Documentation

No docs change. Nothing here alters a documented contract — the wire fields, the routes and the manifest surface are all restored to what the specs already describe.

---

https://claude.ai/code/session_01SpwpaZjyCJvetY7B3KP6to

## Summary by CodeRabbit

- **New Features**
  - Referral event updates now identify the requester, target, and whether an exchange is direct, enabling consoles to distinguish person-to-person conversations from desk referrals.

- **Tests**
  - Expanded coverage for direct referral exchanges, conversation history access, failure handling, credential availability reporting, setup key propagation, model selection, and embedded host behavior.
  - Reorganized embedded host tests without changing their verified behavior.
  - Added coverage for billing credential error classifications and response-data privacy.


oxoxDev:
@coderabbitai review

The findings on this PR have been addressed or answered since your last pass, and the head has moved. Please re-review so the verdict reflects the current commit.

oxoxDev:
On the prober-override finding: I tested it rather than reasoning about it, and it is a false positive. Reverted the guard I had added for it (`f3e4afef9`).

The claim is that `prober_override::set` leaves a thread-local entry a later test can observe. It cannot, because libtest runs every test function on its own thread — including at `--test-threads=1`, which is the case I expected to be the exception. Minimal reproduction, two tests where the second sleeps and then asserts it sees nothing the first wrote:

```
=== default (parallel) ===   test result: ok. 2 passed; 0 failed
=== --test-threads=1 ===     test result: ok. 2 passed; 0 failed
```

The assertion never fires in either mode, so the map dies with the thread that wrote it.

Two further points against the finding, both checkable: `prober_override` exposes no `clear`/`remove` at all, and five pre-existing tests already call `set("acme", …)` with no cleanup — so the pattern predates this PR and is not something it introduces. The suggested precedent, `composio_probe_override`, also has no clear function and its map is process-global rather than thread-local.

No code change for this one. Everything else on this PR stands: the three restored `project_event_for_viewer` statements and 15 restored tests, with the operator test falsified against the shipped fault (`left: Null`).

Issue #2391, opened 2026-09-18
Default model: set DeepSeek v4 Flash (OpenRouter) out of the box and make it reliable in production


Al629176:
## Overview

A fresh OpenCompany install must land the user on a working model with zero manual configuration. Today it doesn't, on two fronts:

1. **No default model is applied at boot.** Onboarding never selects a model, so every new user's agents start on the offline brain and stay there until the user discovers Settings → Model, picks one, and restarts the app. (#2336 tracked the missing onboarding step and is closed, but the underlying "boot with no model" behaviour is still what a fresh 0.2.0 install does.)
2. **The intended default — DeepSeek v4 Flash via OpenRouter — is getting rate-limited on the managed backend.** When it *is* selected, turns fail with `Upstream rate limit exceeded for model 'openrouter/deepseek/deepseek-v4.1-flash'`, and the client gives up after one retry, leaving the user with a dead turn.

Both must be fixed for flows to work end-to-end out of the box. This is a P0 because it blocks every new user before their first agent action.

---

### 1a. Boot-time default
- On first boot with a managed TinyHumans key present, the company's inference config must resolve to `openrouter/deepseek/deepseek-v4.1-flash` (or whatever the product-agreed alias for "DeepSeek v4 Flash" is) with no user interaction
- `/api/v1/company/capabilities` must return `cognition: "configured"` immediately after onboarding completes — not `offline` / unset
- No restart should be required for the default to take effect (0.1.7 reported "booted without a model" and required a full app restart after the model was set manually)

### 1b. Onboarding
- Onboarding either shows the default pre-selected (user can confirm/override) or skips the step entirely and states which model is active — but must never exit with no model bound
- If the managed key lacks the scope needed for inference, onboarding must surface that error at the key step, not silently fall through to the offline brain

### 1c. Observed on fresh installs
- 0.1.7 (16 Sep): API key entered and tested → no model selected → LLM page shows "No providers connected" → agents on offline brain. Restart required after manual fix. No way to change the model afterwards (#2335, #2336)
- 0.2.0 (18 Sep, fresh install after full wipe of `~/.opencompany`): `companies/<slug>/secrets/` and `keys/` are empty on first boot; no model config written anywhere before onboarding

---

### 2a. Rate limiting
- The managed inference path (`api.tinyhumans.ai/openai/v1/chat/completions` → OpenRouter → DeepSeek) is returning upstream 429s under light single-user load:

  ```
[...]

Issue #2336, opened 2026-09-15
Bug: Onboarding has no model selection step — app boots without a model, agents on offline brain until manual fix


Al629176:
## Description

The onboarding flow collects a TinyHumans API key but never asks the user to choose a model. As a result, the app boots without a configured model and agents fall back to the offline brain. The user must manually navigate to Settings → LLM and select a model post-onboarding for anything to work. Additionally, once a model is set, there is currently no option to change it.

**Build:** v0.1.7
**Reproducible:** Yes

## Steps to Reproduce

1. Fresh install of OpenCompany v0.1.7
2. Complete onboarding with TinyHumans API key
3. Enter the app — agents are on offline brain
4. Go to Settings → Connections → API Keys → LLM
5. See: 'Restart required. This company booted without a model, so agents are still on the offline brain and the saved configuration is not yet in effect.'
6. Try to change the model — no option available

## Expected Behaviour

- Onboarding includes a model selection step (e.g. step 2 of 4: 'Choose a model')
- Available models are populated based on the API key entered in step 1
- App boots with the selected model active — no restart required
- Model can be changed post-onboarding from Settings → LLM

## Actual Behaviour

- No model selection in onboarding
- App boots without a model → 'Restart required' banner in LLM settings
- Agents use offline brain until user manually adds a model in Settings
- No option to change the model after the first selection (noted as being patched by Steve)

## Impact

- Every new user lands in the app with non-functional agents after onboarding
- 'Restart required' message on first settings visit is confusing and trust-breaking
- Core agent functionality blocked until user discovers and fixes this manually
- Model lock-in after first selection (no change option) further compounds the issue

## Acceptance Criteria

- [ ] Onboarding includes a model selection step after API key entry
- [ ] Model dropdown is populated with available models for the configured provider
- [ ] App boots with the selected model — no restart required after onboarding
- [ ] 'Restart required. This company booted without a model' banner does not appear post-onboarding
- [ ] User can change the model at any time from Settings → LLM after onboarding
- [ ] Verified on v0.1.7+ with fresh install using TinyHumans managed key

Issue #2335, opened 2026-09-15
Bug: TinyHumans API key entered during onboarding does not persist — LLM page shows 'No providers connected' after setup


Al629176:
## Description

When a user enters their TinyHumans API key during onboarding (Model step 1 of 4), the key is tested successfully but does not persist to the LLM settings page. After completing onboarding, the app shows 'No providers connected yet' and agents are stuck on the offline brain. The user is forced to re-enter their key manually in Settings.

**Build:** v0.1.7
**Reproducible:** Yes

## Steps to Reproduce

1. Fresh install of OpenCompany v0.1.7
2. Go through onboarding — reach 'Model step 1 of 4'
3. Select TinyHumans as provider, enter API key
4. Test connection — receives 'Reached https://api.tinyhumans.ai/openai/v1 using burst-v1 and got a reply'
5. Complete onboarding
6. Navigate to Settings → Connections → API Keys → LLM

## Expected Behaviour

- API key entered during onboarding is saved and appears under LLM Providers as connected
- Agents can think immediately after onboarding completes — no manual reconfiguration needed
- Chat does not show 'Agents can't think yet' banner

## Actual Behaviour

- LLM page shows 'No providers connected yet. Connect one to get started.'
- Chat shows: *'Agents can't think yet. This company has no model configured, so the replies in this conversation come from the offline brain rather than the agent they appear under. Choose a provider in Connections > API Keys > LLM'*
- Settings → Account shows 'Connect to TinyHumans' as if no connection exists
- User must manually re-add the key in Settings to get agents working

## Impact

- Onboarding is completely broken for TinyHumans managed keys — the core happy path fails
- First-time users land in the app with non-functional agents and no clear reason why
- Trust-breaking: user successfully tested the key during onboarding yet it doesn't work
- High drop-off risk at first use

## Acceptance Criteria

- [ ] API key entered and tested during onboarding persists to Settings → LLM Providers
- [ ] After completing onboarding, LLM page shows TinyHumans as connected
- [ ] Agents are functional immediately after onboarding — no offline brain fallback
- [ ] 'Agents can't think yet' banner does not appear after a successful onboarding
- [ ] Settings → Account reflects the connected TinyHumans account correctly
- [ ] Verified on v0.1.7+ with a fresh install and TinyHumans managed key

graycyrus:
covered in #2343

PR #2397, opened 2026-09-18
fix(setup): prefer the chat-v1 tier so a managed install stops defaulting by catalogue position

CodeGhost21:
Track 1 of #2391, managed path: a fresh install stops picking its model by catalogue position.

## What was wrong

`PREFERRED_SETUP_MODEL` exists so a first company does not inherit its model from "whichever the provider happens to list first". On the managed path it never did that job.

TinyHumans' managed endpoint answers `GET /v1/models` with **tier ids** — `reasoning-v1`, `vision-v1`, `chat-v1`, `burst-v1`, … — and never a vendor slug. The OpenRouter passthrough namespace (`openrouter/<author>/<slug>`) is behind an opt-in `?catalog=` query the wizard does not send.

The constant named `deepseek/deepseek-v4-flash`, so in `probe_model_candidates`:

```rust
let preferred = id == PREFERRED_SETUP_MODEL;   // never true on the managed path
```

the sort was a no-op and `candidates.first()` returned whatever came first — `reasoning-v1`, the V4 Pro reasoning SKU, chosen for every fresh managed install by nothing but the order the backend declares its tiers in. `SetupWizard.tsx:875` persists that value as `inferenceModel`, so it is what the company is actually bound to. Exactly the "position in someone else's catalogue rather than a choice" the doc comment says the constant prevents.

## The change

An ordered preference list rather than one string, because a single id cannot cover the three namespaces this probe meets:

| endpoint | what its catalogue lists |
|---|---|
| TinyHumans managed | `chat-v1` (tier ids) |
| self-managed OpenRouter key | `deepseek/deepseek-v4-flash` |
| direct DeepSeek key | `deepseek-v4-flash` |

```rust
const PREFERRED_SETUP_MODELS: &[&str] =
    &["chat-v1", "deepseek/deepseek-v4-flash", "deepseek-v4-flash"];
```

Rank by position in that list. Unlisted models sort after every listed one on an equal key and `sort_by_key` is stable, so a catalogue offering none of them keeps its own order exactly as before — the existing `a_catalogue_without_the_preferred_model_keeps_its_own_order` test is unchanged and still passes.

`chat-v1` leads because it is the short-turn conversational SKU a first company wants. It also keeps a managed install off the passthrough namespace, whose per-model price cap is the other half of #2391 — see tinyhumansai/backend#1356, which fixes the cap that collapsed `openrouter/deepseek/deepseek-v4.1-flash` onto a single upstream endpoint and produced the `Upstream rate limit exceeded` turns.

## Tests

Seven in `setup_model_preference`, three of them new:

- `a_managed_tier_catalogue_lands_on_the_chat_tier_not_catalogue_position` — the regression: the real managed tier listing, in the backend's own declaration order, now lands on `chat-v1` instead of `reasoning-v1`
- `a_vendor_slug_catalogue_still_lands_on_deepseek_v4_flash` — both slug spellings still win in their own namespaces
- `an_earlier_preference_outranks_a_later_one` — an endpoint offering several takes the most preferred

`cargo test -p opencompany-core --features openhuman setup_model_preference` — 7 passed. `cargo fmt` and `cargo clippy -p opencompany-core --features openhuman --all-targets` clean.

## What this does not cover

Still open on #2391, and not addressed here:

- The self-managed branch is untouched — this only changes which model the probe prefers, not whether onboarding can exit without one (#2336) or whether the key persists (#2335)
- No restart-free guarantee, and no `cognition: "configured"` assertion on `/api/v1/company/capabilities`
- Client-side retry/backoff with a visible "model busy" state (Track 2b)
- The 20-consecutive-turn production run the issue asks for

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved first-run model selection to prioritize the preferred chat model across managed tier IDs and vendor model names.
  - Added fallback handling for alternate model namespaces.
  - Ensured preferred models are selected based on priority rather than catalogue order.
  - Updated embedding setup to use the preferred chat model when applicable.


CodeGhost21:
Back to draft — the premise in the description is wrong, found while running the app locally.

The `tinyhumans` provider row does not point at `/openai/v1` (tier ids). It points at `https://api.tinyhumans.ai/agent-integrations/openrouter` (`catalogue.rs:371`), whose `/models` returns **bare OpenRouter slugs** (`listModels.ts:104`, `id: model.slug`) — not `chat-v1`, `reasoning-v1`, ….

So the original constant was never inert on the managed path: `deepseek/deepseek-v4-flash` is a live slug in that catalogue, it matched, and it won. A fresh managed install binds it, not `reasoning-v1`.

Which makes this change a no-op: `chat-v1` is absent from that catalogue, so the ordered list falls through to `deepseek/deepseek-v4-flash` at rank 1 and the outcome is identical. Binding the `chat-v1` tier is not reachable by editing this constant — it would need the managed row repointed at `/openai/v1`, which is a different change.

Holding until the approach is agreed.

CodeGhost21:
Closing — this change is a no-op and its rationale was wrong.

**Why it was wrong.** I read the managed catalogue as tier ids from `GET /openai/v1/models`. It isn't. The `tinyhumans` provider row points at `https://api.tinyhumans.ai/agent-integrations/openrouter` (`catalogue.rs:371`), and that `/models` returns **bare OpenRouter slugs** (`listModels.ts:104`, `id: model.slug`). So `PREFERRED_SETUP_MODEL = "deepseek/deepseek-v4-flash"` is present in that catalogue, matches, and wins. A fresh managed install was already binding deterministically — it was never falling through to catalogue position, and `reasoning-v1` was never in play.

**Why it is a no-op.** `chat-v1` is absent from that catalogue, so the ordered list this PR introduces falls through to `deepseek/deepseek-v4-flash` at rank 1. Same outcome, more code.

**Why it is no longer needed.** The 429s were never about which model was picked — they were the OpenRouter price cap collapsing a model onto its cheapest one or two endpoints. Fixed in tinyhumansai/backend#1356 (merged): the cap now carries 3x headroom, so `deepseek/deepseek-v4-flash` gets its endpoint pool and OpenRouter's health-aware load balancing back.

**What is still open**, and deliberately not done here: making the `chat-v1` *tier* the managed default. That is not reachable from this constant — `chat-v1` is served by `/openai/v1`, not by the OpenRouter proxy this provider row addresses. Repointing the row changes billing (tier rate cards vs at-cost passthrough), which models the picker offers, and failover behaviour. Worth a scoped issue rather than a constant edit.

Branch kept at `CodeGhost21:fix/2391-preferred-setup-model` in case the ordered-list mechanism is useful later.

